PR #450, opened 2017-11-02
add numerically stable bernoulli.batch_log_pdf

dustinvtran:
Following discussion with @neerajprad and @fritzo in https://github.com/uber/pyro/pull/416#discussion_r148127425.

I planned to do the same for Categorical by building off `torch.nn.CrossEntropyLoss`. But not sure if it needed it:
```
batch_log_pdf = logits.masked_select(boolean_mask.byte()).contiguous().view(batch_pdf_shape)
```
seems stable?


dustinvtran:
See #494.

PR #455, opened 2017-11-02
Causal inference example/tutorial

eb8680:
As requested by Zoubin. Everything works with #445 and #459, just need to fill in some text.

Not a real introduction to causal inference, only to Pyro's intervention mechanism `pyro.do` by way of a disease diagnosis example.


karalets:
I also suggest we make sure whenever this is merged that the documentation in advanced features also exists for the do-poutine, currently this is not the case.

robertness:
Great job!  A few comments:

The figure demonstrating that curing the cough doesn't cure the cold might throw people off.  The suggestion that someone might have a cold conflicts with the apparent low probabilities in the figure, and the reader has to backtrack to see the probability of a cold was set to set to low value of .02.  It might be a bit clearer if you _also_ condition on having a fever so the probability of a cold is higher. 

If you do end up making a causal inference tutorial, I suggest including a bandit algorithm (eg [Bareinboim and Pearl 2015](http://ftp.cs.ucla.edu/pub/stat_ser/r460.pdf) or [Lattimore and Reid](http://papers.nips.cc/paper/6195-causal-bandits-learning-good-interventions-via-causal-inference.pdf).  I find that bandits and other sequential decision-making methods are good motivating examples, relative to the canonical Bayes net examples.

talesa:
I know I'm late to the game, but please bear with me - why did this never get merged? Is this example outdated now?

eb8680:
> I know I'm late to the game, but please bear with me - why did this never get merged? Is this example outdated now?

@talesa this code is pretty old and uses a bunch of obsolete APIs. If you're looking for more up to date examples of this sort, see @robertness's lecture notes here: https://github.com/robertness/causalML

talesa:
Thank you @eb8680 @robertness 🙏

PR #492, opened 2017-11-03
Improved Sphinx coloring for docs and examples

OptimusLime:
New image: 
image

From old: 
image


neerajprad:
Just seeing this now. Looks great, but I do not see this take effect yet. Was this not deployed?

PR #500, opened 2017-11-03
Use observations and move data loading upstream

dustinvtran:
See #495.


neerajprad:
Re-triggered the build. Will merge once tests pass.

PR #504, opened 2017-11-03
Fix xfail markers on failing tests

neerajprad:
This fixes a problem with my previous merge when travis was not working.


neerajprad:
Good to merge now.

PR #509, opened 2017-11-03
Make pyro.distributions.util.log_gamma compatible with Tensors

gokceneraslan:
Current implementation of log_gamma is not compatible with Tensors i.e. `log_gamma(torch.ones(1))` throws:

```
RuntimeError: add() received an invalid combination of arguments - got (torch.FloatTensor), but expected one of:
 * (float other, float alpha)
 * (Variable other, float alpha)
```

### DIFF
--- a/pyro/distributions/util.py
+++ b/pyro/distributions/util.py
@@ -6,10 +6,9 @@ from torch.autograd import Variable
 
 
 def log_gamma(xx):
-    if isinstance(xx, Variable):
-        ttype = xx.data.type()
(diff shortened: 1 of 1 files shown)


gokceneraslan:
Btw why are you using `ser = ser + torch.pow(x / c, -1)` instead of `ser = ser + (c / x)`?

fritzo:
> Btw why are you using `ser = ser + torch.pow(x / c, -1)` instead of `ser = ser + (c / x)`?

I'm not sure, @jpchen did you implement the first version of this?

PR #511, opened 2017-11-04
fixes for regression tutorial

karalets:
updated regression tutorial to be uniformly about regression, fixed the code to say prediction instead of latent, made tutorial more consistent with example code

### DIFF
--- a/examples/bayesian_regression.py
+++ b/examples/bayesian_regression.py
@@ -17,7 +17,6 @@ Learning a function of the form:
     y = wx + b
 """
 
-
 # generate toy dataset
 def build_linear_dataset(N, p, noise_std=0.1):
     X = np.linspace(-6, 6, num=N)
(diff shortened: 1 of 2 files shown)


examples/bayesian_regression.py
```diff
@@ -54,18 +54,17 @@ def model(data):
...
     w_prior, b_prior = Normal(mu, sigma), Normal(bias_mu, bias_sigma)
     priors = {'linear.weight': w_prior, 'linear.bias': b_prior}
-    # wrap regression model that lifts module parameters to random variables
-    # sampled from the priors
+    # lift module parameters to random variables sampled from the priors
     lifted_module = pyro.random_module("module", regression_model, priors)
-    # sample a nn
-    lifted_nn = lifted_module()
+    # sample a regressor (which also samples w and b)
+    lifted_rm = lifted_module()
```
jpchen: can you write out regression_model or just `lifted_reg_model` ? i was originally confused what rm was

tutorial/source/bayesian_regression.ipynb
```diff
@@ -182,23 +190,25 @@
     "\n",
     "### `random_module()`\n",
     "\n",
-    "In order to do this, we'll 'lift' the parameters $w$ and $b$ to random variables. We can do this in Pyro via `random_module()`, which effectively takes a `nn.Module` and turns it into a distribution over neural networks. Specifically, each parameter in the original neural net is sampled from the provided prior. This allows us to repurpose vanilla neural nets for use in the Bayesian setting. For example:"
+    "In order to do this, we'll 'lift' the parameters $w$ and $b$ to random variables. We can do this in Pyro via `random_module()`, which effectively takes a `nn.Module` and turns it into a distribution over models. Specifically, each parameter in the original model is sampled from the provided prior. This allows us to repurpose vanilla regression models for use in the Bayesian setting. For example:"
```
jpchen: isnt this conflating a "model" for regression with the `model()` used for svi?  this reads to a beginner like they are the same thing when they are in fact different. or i guess it might be worth connecting `nn.Module` to a "regression model" (eg saying that were using a nn to represent this regression model)

karalets:
ok, both taken care of. Please review and merge if all is good now.

jpchen:
can you run `nbstripout bayesian_regression.ipynb`? i think travis will complain because we are scrubbing output now. 

and while youre at it, can you change the name `sampled_rm` in the tutorial to `sampled_reg_model` (to be consistent with the naming  in the model)? (in the code snippet describing random_module)

otherwise LGTM

karalets:
done, check it.

tutorial/source/bayesian_regression.ipynb
```diff
@@ -408,6 +430,13 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "Bayesian nonlinear regression can also be implemented analogously by lifting neural network modules."
```
jpchen: you mention nns here for the first time, even though theyre used above.  is this to distinguish 'actual' nns as approximators for strictly non-linear functions?  (ie the purpose of this PR)

karalets:
What's used above is only the module nn, not an actual nn. There is no
nonlinearity, just the linear part.

PR #515, opened 2017-11-05
Fix sum variable with dim = -1 torch error

darthsuogles:
From the dev branch:
Using the following code snippet from the tuturial,
```
mu = Variable(torch.zeros(1))   # mean zero
sigma = Variable(torch.ones(1)) # unit variance
x = dist.normal(mu, sigma)      # x is a sample from N(0,1)
log_p_x = dist.normal.log_pdf(x, mu, sigma)
```
We observe a PyTorch error:
`RuntimeError: dimension -1 out of range at pytorch/torch/lib/TH/generic/THTensorMath.c:1622`

### DIFF
--- a/pyro/distributions/normal.py
+++ b/pyro/distributions/normal.py
@@ -88,7 +88,7 @@ class Normal(Distribution):
         # will likely be done in a better/cleaner way in the future
         if self.log_pdf_mask is not None:
             log_pxs = log_pxs * self.log_pdf_mask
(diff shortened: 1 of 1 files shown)


jpchen:
@darthsuogles what version of pytorch are you using?  pytorch >=0.2.0 supports -1 indexing (and broadcasting)

darthsuogles:
Yeah, I just realized I was using `0.10.xxx`, upgraded to `0.20` and -1 indexing works.

Issue #514, opened 2017-11-05
Nondeterministic failures on travis


fritzo:
We've seen a few nondeterministically failing builds post 0.1.0:
- https://travis-ci.org/uber/pyro/builds/297437367
- https://travis-ci.org/uber/pyro/builds/297418157
- https://travis-ci.org/uber/pyro/builds/297417599

Two tests that I've seen fail are:
- test_gmm_elbo_gradient[dense-sample-single]
- test_gmm_elbo_gradient[flat-sample-single]

<details>

```
=================================== FAILURES ===================================
[31m[1m_________________ test_gmm_elbo_gradient[dense-sample-single] __________________[0m

model = <function gmm_model at 0x7f23f41f67b8>
guide = <function gmm_guide at 0x7f23f41f6840>, enum_discrete = False
trace_graph = False

[1m    @pytest.mark.parametrize("model,guide", [[0m
[1m        (gmm_model, gmm_guide),[0m
[1m        (gmm_batch_model, gmm_batch_guide),[0m
[1m    ], ids=["single", "batch"])[0m
[1m    @pytest.mark.parametrize("enum_discrete", [[0m
[1m        pytest.param(True,[0m
[1m                     marks=pytest.mark.xfail(reason="https://github.com/uber/pyro/issues/278")),[0m
[1m        False,[0m
[1m    ], ids=["sum", "sample"])[0m
[1m    @pytest.mark.parametrize("trace_graph", [False, True], ids=["dense", "flat"])[0m
[1m    def test_gmm_elbo_gradient(model, guide, enum_discrete, trace_graph):[0m
[1m        pyro.clear_param_store()[0m
[1m        num_particles = 4000[0m
[1m        data = Variable(torch.Tensor([-1, 1]))[0m
[1m    [0m
[1m        print("Computing gradients using surrogate loss")[0m
[1m        Elbo = TraceGraph_ELBO if trace_graph else Trace_ELBO[0m
[1m        elbo = Elbo(enum_discrete=enum_discrete,[0m
[1m                    num_particles=(1 if enum_discrete else num_particles))[0m
[1m        with xfail_if_not_implemented():[0m
[1m            elbo.loss_and_grads(model, guide, data)[0m
[1m        params = sorted(pyro.get_param_store().get_all_param_names())[0m
[1m        assert params, "no params found"[0m
[1m        actual_grads = {name: pyro.param(name).grad.clone() for name in params}[0m
[1m    [0m
[...]

PR #519, opened 2017-11-06
AIR tutorial: Show example data points as promised in the text.

null-a:
The main text is written with the expectation that the contents of this cell will show up, so this sets `keep_output=True` etc.


null-a:
CI failure appears to be unrelated to this change, error message mentions `tests/infer/test_enum.py`?

eb8680:
I restarted the builds.  Maybe related to #514?  Also, @jpchen what's up with the Travis tutorial failures?

jpchen:
i believe thats because it is trying to run the cells but not all the tutorials are runnable
cc-ing @neerajprad since he would know

neerajprad:
The travis-tutorial stage should not impact the build result. It would fail if any of the notebook cells errors when the tool tries to run it. Since our tutorials are not all runnable (except for bayesian regression), it is expected to fail. 

We should either annotate the non runnable cells by using the comment `#NBVAL_SKIP`, or remove this from travis if it is too much effort to support this in our tutorials. @fritzo - what do you think?

neerajprad:
I am taking a look at #514. The test seems to pass locally, but is failing on travis non-deterministically.

neerajprad:
@null-a, @eb8680 :  we do not need to wait on the test results to merge this, since this is only touching the tutorials.

fritzo:
I think it's reasonable to remove the tutorial stage from Travis tests for now, especially since we're having such difficulties with travis-ci.org. If we remove that, we should at least file an issues to get tutorial tests working, since we've already seen a few bugs in tutorials due to lack of tests.

neerajprad:
> If we remove that, we should at least file an issues to get tutorial tests working, since we've already seen a few bugs in tutorials due to lack of tests.

I am removing this in #520. If the changes seem to work, I will create an issue to reinstate the tutorial stage once our CI issues are resolved.

PR #520, opened 2017-11-06
Minor changes to travis to make build faster

neerajprad:
Makes the following changes to the test utilities:
 - sets `sudo: false`. This allows us to use dockerized instances that have 2 cores available (invoking pytest-xdist with `n=2`), and have reportedly faster build start times. 
 - Removes the tutorials stage (currently failing) - should be reinstated soon #525.

I see a definite, though small improvement in test time. We can merge this and continue to fine-tune travis.yml as needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,10 @@
 language: python
 
-sudo: required
+sudo: false
 
 env:
     global:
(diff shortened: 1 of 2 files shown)


fritzo:
:+1: I hope this works. It feels like we're paralyzed!

PR #533, opened 2017-11-08
Fix gradient scaling when subsampling

### LINKED ISSUE #531: site["scale"] is double counted at non-reparameterized sites
## The Problem

Consider the surrogate elbo at a nonreparameterized site (as [computed in](https://github.com/uber/pyro/blob/ece8ccc167695f81aa7ec1fed74a70c8fc15254b/pyro/infer/trace_elbo.py#L150) `Trace_ELBO`)
```py
log_p = model_site["log_pdf"]
log_q = guide_site["log_pdf"]
surrogate_elbo = log_p + (log_p - log_q).detach() * log_q
```
Now suppose this site occurs in an `irange` where we're subsampling 1% of the data, so that a `ScalePoutine` sets `site["scale"] = 100`. Then `Trace.log_pdf()` [will scale](https://github.com/uber/pyro/blob/ece8ccc167695f81aa7ec1fed74a70c8fc15254b/pyro/poutine/trace.py#L71) both `log_p` and `log_q` by 100. Therefore the `surrogate_elbo`'s first term `log_p` will be scaled by 100, but the second term `(log_p - log_q).detach() * log_q` will be scaled by 100 * 100.

## Scope

This should only affect models in one of the following scenarios:
1. The model has both global and local sites, and the local sites are non-reparameterized and subsampled.
2. The model has local sites that are non-reparameterized and subsampled and includes model parameters at that site (not only guide parameters).

## Possible Solutions

1. Handle `site["scale"]` separately, outside of `Trace.log_pdf()`. This scale is already wired through `Trace_ELBO` and `TraceGraph_ELBO` as `weight`.
  **Pros:** Avoids being too clever. Plays well with scale that varies across a batch (as in `BranchPoutine`).
  **Cons:** May require special treatment of `site["scale"]` in many places outside of `Trace_ELBO` and `TraceGraph_ELBO`.
2. Keep the existing logic of `Trace.log_pdf()`, and remove a factor of `site["scale"]` from the `(log_p - log_q).detach() * log_q` term.
  **Pros:** Minimally invasive.
  **Cons:** Seems like a hack. Does not play well with scales that vary across a batch.

fritzo:
Fixes #531 

This implements a quick-and-dirty fix for double counted `site["scale"]` when subsampling. This corresponds to Solution 2 in #531. The author would like to merge this PR first, and later clean up by implementing Solution 1 in #531.

## Tested

Added a regression test in `tests/infer/test_gradient.py` that compares gradient estimates with and without subsampling. Results are:

| which elbo | distribution | Before this PR | After this PR |
| - | -: | - | - |
| Trace | reparametrized | PASS | PASS |
| Trace | non-reparametrized | FAIL | PASS |
| TraceGraph | reparametrized | PASS | PASS |
| TraceGraph | non-reparametrized | FAIL | PASS |


tests/infer/test_gradient.py
```diff
@@ -0,0 +1,101 @@
...
+    inference.loss_and_grads(model, guide, subsample_size=subsample_size)
+    actual_grads = {name: param.grad.data.clone() for name, param in params.items()}
+
+    for name in sorted(params):
+        print('\nexpected {} = {}'.format(name, expected_grads[name].cpu().numpy()))
+        print('actual   {} = {}'.format(name, actual_grads[name].cpu().numpy()))
+    assert_equal(actual_grads, expected_grads, prec=precision)
+
+
+# The golden values below (mu_q_expected/log_sig_q_expected/) need to be updated each time
```
fritzo: This test was moved from `test_gradient_step.py`. The only changes were to add `ids` to the `@pytest.mark.parameterize` decorators.

PR #538, opened 2017-11-09
Small fix to VAE example when using CUDA

rtqichen:
Previously the network params weren't actually being moved to cuda. This function is now the same as in the tutorial.


martinjankowiak:
yes, thanks, this is right. not sure how that got moved down a few lines...
in order for us to merge this though you'll need to fix linting:

./examples/vae.py:78:1: W293 blank line contains whitespace

rtqichen:
Oops, done. The hazards of directly editing on github. Thanks for this useful and instructive library!

Issue #502, opened 2017-11-03
Alternative design of variable scoping and naming


srush:
Curious why all the params are auto-magically global with a shared namespace. Seems like a very different design choice than the nice cleanly scoped pytorch variables.

eb8680:
We're definitely planning to add a dynamic scoping mechanism of some sort for both `param` and `sample`, it just didn't make it into this release.  See #73 for some discussion about this.

srush:
Sorry, but I don't really get this. The plan seems to be to copy TensorFlow's heavyweight design. PyTorch is so nice because it piggybacks on python's natural scoping rules.

eb8680:
For models that can be written as `nn.Module`s, where all parameters can be created before function evaluation, `pyro.module` will add the correct scope to the parameter names automatically.

Unfortunately, parameters in Pyro don't always work this way; consider e.g. per-datapoint variational parameters that are created on the fly each time a new data point appears.

`sample` statements are also problematic because names are used to identify corresponding random variables in models and guides, which may have arbitrarily different control structure.  Lexical scoping for sample names would break this identification.

srush:
Not sure I understand.

Why should you be able to sample a random variable that you don't have pointer access to? Needing to do that sounds dangerous. 

Why does having non-global scoping prevent per-data point parameters? Presumably whatever object is creating the parameters should own them, not some global store.

srush:
(Sorry, this library is great, and I was thinking of using it for teaching. But I am really worried about this design decision.)

eb8680:
> (Sorry, this library is great, and I was thinking of using it for teaching. But I am really worried about this design decision.)

No problem! The parameter store is definitely a weak point, and this is great feedback.

> Why does having non-global scoping prevent per-data point parameters? Presumably whatever object is creating the parameters should own them, not some global store.

`nn.Module`s expose references to all of the parameters of their child modules, making it possible to completely separate the definition of losses that only interact with output and optimizers that only interact with parameters and gradients.  Since Pyro models can be regular Python functions that call other regular Python functions, you can't in general automatically track parameters through the call stack in the same way, so it seems to me that you end up with something that looks like a parameter store.  I guess we could have a per-model parameter store instead, though, sort of like a dynamically generated `nn.Module` wrapper for a function with `param` calls.

srush:
I guess I get that, but now I'm much more worried about this type of name matching though. This feels very scary though. 

[...]

PR #540, opened 2017-11-09
Add a pyro.contrib.named module

fritzo:
This aims to add syntactic sugar to pyro in the form of an experimental `pyro.contrib.named` module.

## Why?

The proposal for syntactic sugar `anon` in #502 is quite far from Pyro's current language. `anon` is also unsafe and will likely frustrate new users. This PR aims to simplify `anon` to find a compromise between these goals:
- keep close to existing Pyro syntax (so it's easy to mentally translate in either direction)
- avoid string manipulation
- avoid excessive magic that leads to pitfalls and frustrated users
- allow modularity and recursion (in a cleaner way than `with pyro.scope("foo"): ...`)
- avoid modifying any core pyro code

## How?

This PR closely follows the three special pyro statements
```py
pyro.sample("latent.x", fn, *args, **kwargs)
pyro.observe("latent.y", fn, obs, *args, **kwargs)
pyro.param("latent.z", *args, **kwargs)
```
but turns them into in-place methods on `named.Object`s
```py
latent = named.Object("latent")
latent.x.sample_(fn, *args, **kwargs)
latent.y.observe_(fn, obs, *args, **kwargs)
latent.z.param_(*args, **kwargs)
```
Importantly, named sub-objects can be extracted and passed to helper functions
```py
x = latent.x
x = x.sample_(fn, *args, **kwargs)
print(latent.x)  # prints sampled value
```

The `pyro.contrib.named` module introduces three container data structures: `named.Object`, `named.List`, and `named.Dict`. These can contain each other. Each named container knows its own name and this name will be used in Pyro sites.

## Checklist

- [x] Create a module `pyro/contrib/named.py` with new code.
- [x] Create a module docstring describing purpose and authors.
- [x] Create document all public functions and methods.
- [x] Create Sphinx hooks in `docs/source/contrib.rst` and `docs/source/contrib.named.rst`.
- [x] Create tests `tests/contrib/test_named.py`.
- [x] Create examples `examples/contrib/named/*.py`.


jpchen:
Should these kinds of add ons and syntactic sugar be in an extensions module? To let users knows "use at your own risk"

fritzo:
Sure, we could name it `pyro.contrib.sugar.named` or just `pyro.contrib.named`.

eb8680:
> Sure, we could name it pyro.contrib.sugar.named or just pyro.contrib.named

Yeah, until we have several really solid examples I'd say we should just do that.  I'd really prefer to have some new examples included before merging this at all.

srush:
I don't have a vote, but this is a poor compromise in my mind. 

- The use of implicit CPS is very confusing to a standard python user, not versed in previous PPLs (most users this language is targeting). The fact that `named.sample` has vastly different behavior depending on some unseen global is scary. Using `latent.x.sample` corresponds to the expected semantics in python. 

- The current language has been out in alpha for three days, but this modeling language will be used for years. I think it is important to establish one good syntax that matches the semantics of the language. I have yet to hear an argument why the "globals" syntax is better? 

- As I understand it, the proposed #502  syntax is only "unsafe" is because of the current usage of a global store. If you switch to a scoped modeling language, you can eventually transition the backend away from the global store in time. If you do not, there will be lots of legacy code where users do terrible hacky global things. That code will need to be supported for ever.

srush:
(To clarify, my last concern. I think that infer's like SVI, etc should be passing in some object to model() and guide() that conveys to the function what "context" they are being run in. In the other proposed syntax, that was always the first argument `model(latent)`. This object needs to be passed to `.sample`. I guess that this behavior is currently all done with global stateful stacks to changes the meaning of `sample`. In the future though, this syntax allows for a transition to that information being controlled by `latent` and to act only on the members of latent. )

eb8680:
@srush let's try to keep the discussion in this PR about the contents of the PR, and the general discussion in the issue.

fritzo:
> Using `latent.x.sample` corresponds to the expected semantics in python

Done. I've followed [PyTorch conventions](http://pytorch.org/docs/master/tensors.html#torch-tensor) and named these `.sample_()` etc. where the `_` indicates they act in-place.

> proposed #502 syntax is only "unsafe" because ...

Sorry for the confusion, I was narrowly referring to my hacks in `pyro.anon`. This PR is basically a rewrite of `pyro.anon` with better safety checks and using @srush's suggested initialization `latent.x = named.List()`. It should be easy to migrate from `anon` to `named`.

srush:
Thanks @fritzo, this is better. One more issue.

- I think that `model` and `guide` should not name their Object's but have them passed in as an argument either by SVI, or by the user. The idea is that SVI is controlling the objects behavior. So it makes sense that if it passes in a different one it would behave differently (I know this is not how it works now, but I want the user to think that). I made small changes to infer to make this work seamlessly https://github.com/harvardnlp/pyro/blob/anonymous-sites/pyro/anon.py#L271

- Also if it takes `latent` as an argument, that allows the `condition` function to work with easier namespacing. Instead of "latent.x.mu" you can just say "x.mu" since the condition function can check the address of the latent being passed through. This gives the illusion of scoping.

fritzo:
@srush Interesting. Looks like you've gotten pretty far with your [anon.py](https://github.com/harvardnlp/pyro/blob/anonymous-sites/pyro/anon.py#L271). I suggest we
1. Merge this PR (which merely introduces a new data structure), then
2. Distill your branch of `anon.py` to use `pyro.contrib.named`, so that we can focus on the more interesting bits (in #544). For example, your `lift` should simplify to:
```py
def lift(fn, _latent=None):
    @functools.wraps(fn)
    def decorated(*args, **kwargs):
        latent = named.Object('latent') if _latent is None else _latent
        return fn(latent, *args, **kwargs)
    return decorated
```

srush:
Great, let's do it.

fritzo:
@eb8680 @jpchen @neerajprad I'd like this PR to serve as a template for `pyro.contrib.*` modules. Can you think of anything else that should be added to the checklist?

pyro/contrib/named.py
```diff
@@ -0,0 +1,245 @@
...
+    >>> z = state.z.sample_(dist.normal, mu, sigma)
+    >>> state.x.observe_(dist.normal, z, mu, sigma)
+
+For deeper examples of how these can be used in model code, see the
+`Tree Data <https://github.com/uber/pyro/blob/dev/examples/contrib/named/tree_data.py>`_
+and
+`Mixture <https://github.com/uber/pyro/blob/dev/examples/contrib/named/mixture.py>`_
+examples.
+
+Authors: Fritz Obermeyer, Alexander Rush
```
fritzo: @srush I've added you as a coauthor since you suggested most of the ideas and you'll be adding more code in #550 .

pyro/contrib/README.md
```diff
@@ -0,0 +1,12 @@
+# Contributed Code
```
fritzo: For similar organization, see https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib

fritzo:
@eb8680 @neerajprad @jpchen Let me know if I can help with review. I'm happy to walk you through the code on a big monitor.

fritzo:
Bump. Let's try to get this merged this week.

neerajprad:
> @eb8680 @neerajprad @jpchen Let me know if I can help with review. I'm happy to walk you through the code on a big monitor.

@fritzo - that will be great! can we go over this briefly in the afternoon/evening?

PR #549, opened 2017-11-10
[WIP] Revert sudo=false in .travis.yml

fritzo:
https://github.com/travis-ci/travis-ci/issues/5623

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 
-sudo: false
+sudo: true
 
 env:
     global:


neerajprad:
Lets see how the test times look before merging.

fritzo:
did you click on the link?

we're trying desparately to diagnose why our builds are taking so long.

jpchen:
@fritzo that was changing it to true. "revert" suggests there was a PR to make it `False` in the first place

neerajprad:
@jpchen - we changed this in #520where we previously had `sudo: required` (which is the same as `sudo: true`).

fritzo:
@jpchen sorry, I misunderstood your question. When Travis-ci released containers a few years ago, they wanted people to try it out so they claimed "containers are faster, so add sudo=false in your build". We believed them. In practice there is negligible speedup and possibly a slowdown, as in the link.

neerajprad:
Check this out - https://www.traviscistatus.com/. The backlog for open source (container projects) seems to have some spiky behavior which isn't there for non container based projects.

@fritzo - We still have this issue of `no response received within 10 mins` for one of the integration test runs. Also, comparing this with #548 (which removes xdist and sets sudo to true), there seems to be some benefit (small) to parallelizing on both the cores.

fritzo:
Closing in favor of #548

PR #551, opened 2017-11-10
fix documentation for random_module

jpchen:


jpchen:
canceling all tests except docs

pyro/__init__.py
```diff
@@ -387,15 +387,18 @@ def module(name, nn_module, tags="default", update_module_params=False):
...
-    tutorial for an example.
+    See the `Bayesian Regression tutorial <http://pyro.ai/examples/bayesian_regression.html>`_
+    for an example.
 
     :param name: name of pyro module
     :type name: str
     :param nn_module: the module to be registered with pyro
     :type nn_module: torch.nn.Module
-    :param prior: prior distribution or iterable over distributions
+    :param prior: pyro distribution, stochastic function, or python dict with parameter names
```
fritzo: nit: Are these comments < 80 characters wide? I'm just asking because Jupyter notebooks are awful at wrapping docstrings like `help(random_module)`.

jpchen: i aligned it with the `module` method above, so i assumed it was? unless none of the documentation is <80

kjchalup:
Hi, not sure if this is the best place to mention it, but I think random_module's documentation could be improved.

In particular, I don't know what the names of the keys of the `prior` dictionary should be. In the Bayesian Regression tutorial, you call it like this:
```python
dists = {'linear.weight': w_dist, 'linear.bias': b_dist}
lifted_module = pyro.random_module("module", regression_model, dists)
```
I'm confused as to where 'linear.weight' and 'linear.bias' come from. Are these the default names pytorch, or pyro call the torch.nn.Linear layer's parameters? Could this be clarified? Is there a place in pyro (or pytorch) where I can choose the names of these parameters prior to providing them to random_module?

jpchen:
Hi @kjchalup  thanks for the feedback.  `random_module` is also getting a facelift in 0.2 which will support fancier architectures, but to answer your questions:
1. they are the `name`s of pytorch's nn parameters, ie `(name, param)` in `module.named_parameters()`
2. no, the ones that are unspecified will be treated as normal parameters of your neural net

ill bake these into the documentation with the new changes.

PR #553, opened 2017-11-10
Bump to version 0.1.2

fritzo:
This minor patch release contains:
- #533 fix for bug in gradient scaling of subsampled non-reparameterized sites
- misc improvements in documentation
- #530 split LambdaPoutine into ScalePoutine + IndepPoutine
  (this is not strictly backward compatible, but we assume nobody is using `pyro.LambdaPoutine` directly)

PR #557, opened 2017-11-10
Parse pyro.__version__ and README.md in setup.py

fritzo:
1. Move pyro version string from `setup.py` to `pyro.__init__.py` so it is available as `pyro.__version__`. (This is common practice.)
2. Parse `README.md` in `setup.py` so it can display on https://pypi.python.org/pypi/pyro-ppl (Also common practice)

Tests cancelled after linting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,9 @@
 [![Latest Version](https://badge.fury.io/py/pyro-ppl.svg)](https://pypi.python.org/pypi/pyro-ppl)
 
 
-[Getting Started](http://pyro.ai/examples) | [Documentation](http://docs.pyro.ai/) | [Contributing](CONTRIBUTING.md)
+[Getting Started](http://pyro.ai/examples) |
+[Documentation](http://docs.pyro.ai/) |
+[Contributing](https://github.com/uber/pyro/blob/master/CONTRIBUTING.md)
 
 Please also refer to the [Pyro homepage](http://pyro.ai/).
 
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,28 @@
 from __future__ import absolute_import, division, print_function
 
+import sys
+
 from setuptools import find_packages, setup
+
+# Find pyro version.
+for line in open('pyro/__init__.py'):
+    if line.startswith('__version__ = '):
+        version = line.strip().split()[2][1:-1]
+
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -1,11 +1,28 @@
 from __future__ import absolute_import, division, print_function
 
+import sys
+
 from setuptools import find_packages, setup
 
+# Find pyro version.
+for line in open('pyro/__init__.py'):
+    if line.startswith('__version__ = '):
```
neerajprad: Would `pyro.__version__` not work here?

setup.py
```diff
@@ -1,11 +1,28 @@
...
+    sys.stderr.write('Failed to convert README.md to rst:\n  {}\n'.format(e))
+    sys.stderr.flush()
+    long_description = open('README.md').read()
+
 setup(
     name='pyro-ppl',
-    version='0.1.1',
+    version=version,
     description='A Python library for probabilistic modeling and inference',
+    long_description=long_description,
```
neerajprad: Nice.

setup.py
fritzo: This is avoided since the pyro module would requires setup.py would require pyro ...
Also text parsing is faster than loading *all of pyro* as is done in `pyro/__init__.py`.

fritzo:
Ok, I'm ready to merge. I'd like to get this in before the 0.1.2 patch release so we can have a nice pypi page 😃

README.md
```diff
@@ -9,7 +9,9 @@
 [![Latest Version](https://badge.fury.io/py/pyro-ppl.svg)](https://pypi.python.org/pypi/pyro-ppl)
 
 
-[Getting Started](http://pyro.ai/examples) | [Documentation](http://docs.pyro.ai/) | [Contributing](CONTRIBUTING.md)
+[Getting Started](http://pyro.ai/examples) |
+[Documentation](http://docs.pyro.ai/) |
+[Contributing](https://github.com/uber/pyro/blob/master/CONTRIBUTING.md)
```
fritzo: changed this to an absolute link so it still works on pypi.python.org

fritzo:
@neerajprad would you mind merging this so I can rebase #553 ?

setup.py
neerajprad: Makes sense.

fritzo:
Thanks!

PR #558, opened 2017-11-10
Fix tests pytorch v0.3.0

fritzo:
Sorry, one more fix. This won't run on travis, but I've verified that it runs locally. I'll cancel tests after linting.


fritzo:
Cancelling since github is having branch issues

PR #572, opened 2017-11-14
Refactor TraceGraph_ELBO

fritzo:
This just moves code around inside `TraceGraph_ELBO` so as to split up the largest method into smaller helper functions.


fritzo:
Ready to merge.

PR #573, opened 2017-11-14
Implement MAP inference via poutine.lower

### LINKED ISSUE #570: Maximum a posteriori inference
I am considering using pyro for my problem, but I don't quite grok the library yet.  Part of my confusion comes from SVI being the main well documented inference engine in pyro, and I am a beginner to variational inference.

For me, the simplest "inference" methods to understand are maximum a posteriori (MAP) and maximum likelihood, and it would be nice if the `pyro` had some examples of these. I assume `pyro` has the optimization tools necessary to handle MAP, but the documentation is not very clear about how to do it.

### LINKED ISSUE #577: Implement LowerPoutine : sample -> param+observe
MAP estimation requires a `param`+`observe` pattern for each estimated parameter in the `model` (see #573 ). This is awkward. We should instead provide something like a `poutine.lower()` to automatically transform `sample` sites in this way. This will require some refactoring of `Poutine` and `apply_stack`, since current poutines are restricted to 1-1 site transformations, whereas `poutine.lower()` needs to transform one `sample` site into two sites.

fritzo:
Fixes #570 
Fixes #577 

## Why?

@nbren12 requested a MAP wrapper around SVI in #570. @fehiepsi is building Laplace approximation on top of MAP in #590. MAP inference is also useful for beginner tutorials, since MAP is much more widely understood than Black-Box SVI.

## How?

`poutine.lower` transforms each sample site to a param site + a model site (see #577). There is a design choice regarding whether to (1) transform the sample site to a param site and create a new observe site, or (ii) transform the sample site to an observe site and add a new param site. This PR opts for (ii) since it is simpler and a little less intrusive.

This implementation `MAP` uses Maximum Likelihood inference for `pyro.param` sites and Maximum A Posterior inference for `pyro.sample` sites, both conditioned on `pyro.observe` sites.

## Caveats

Discrete distributions cannot be `MAP` optimized. Currently `infer.MAP` raises a `NotImplementedError` if a discrete sample site is encountered. While there are possible workarounds (Search or BranchPoutine), they are out of scope of this first PR.

We currently have no way to restrict optimizer moves to the support of constrained distributions (like `Exponential` or `HalfCauchy`). This suggests exposing some sort of bijector to distributions classes, such as:
```py
class Distribution(object):
    def biject_params(self):
        """
        Returns a bijector between constrained param space and unconstrained space.
        This is often used for optimizing parameters in SVI or Maximum Likelihood.
        """
    def biject_value(self):
        """
        Returns a bjiector between constrained results and an unconstrained space.
        This is often used for MAP inference or Hamiltonian Monte Carlo.
        """
```

## Tested

- Added 3 tests of `infer.MAP`
- Added a test of `poutine.lower`


pyro/infer/map.py
```diff
@@ -0,0 +1,22 @@
+from pyro.infer.svi import SVI
+
+
+def MAP(model, optim, *args, **kwargs):
+    """
+    :param model: the model (callable containing Pyro primitives)
+    :param optim: a wrapper a for a PyTorch optimizer
```
jpchen: typo "a"

pyro/infer/map.py
```diff
@@ -0,0 +1,22 @@
...
+    latent variables, or where all latent variables have been replaced by
+    parameters. In MAP models, the likelihood is specified by ``pyro.observe``
+    statements that observe evidence, and the prior is specified by
+    ``pyro.observe`` statements that observe parameter values.
+    """
+
+    def guide(*args, **kwargs):
+        pass
+
+    return SVI(model, guide, optim, loss="ELBO", *args, **kwargs)
```
jpchen: this could be a one liner right:
```
return SVI(model, lambda *args, **kwargs: None, optim, loss="ELBO", *args, **kwargs)
```
though im sure theres some SO post that says its more pythonic to do it your way

tests/infer/test_map.py
```diff
@@ -0,0 +1,27 @@
...
+from pyro.optim import Adam
+from pyro.util import ng_ones, ng_zeros
+from tests.common import assert_equal
+
+
+def test_map():
+
+    def model(data):
+        mu = pyro.param("mu", Variable(torch.zeros(1), requires_grad=True))
+        pyro.observe("mu_prior", dist.normal, mu, ng_zeros(1), ng_ones(1))
```
fritzo: @martinjankowiak feels that this `param`+`observe` pattern is awkward and that we should instead provide something like a `poutine.lower()` to automatically transform `sample` sites in this way. This will require some refactoring of `Poutine` and `apply_stack`, since current poutines are restricted to 1-1 site transformations, whereas `poutine.lower()` transforms one `sample` site into two sites. See #577 for details.

fritzo:
Waiting on #577 to implement `poutine.lower`.

pyro/infer/map.py
fritzo: Right, but I think it's easier to understand when we write out the trivial guide.

pyro/poutine/lower_poutine.py
```diff
@@ -0,0 +1,36 @@
+from __future__ import absolute_import, division, print_function
+
+from torch.autograd import Variable
+import pyro  # XXX remove this cyclic dependency
+from pyro.params import _PYRO_PARAM_STORE
+
+from .poutine import Poutine
+
+
+class LowerPoutine(Poutine):
```
fritzo: @neerajprad How similar is this is to an HMC poutine? Can we use `poutine.lower` directly for HMC, where we use `LeapFrog` instead of `infer.MAP`?

neerajprad: Discussed with @fritzo. We decided to use ReplayPoutine for HMC where the replayed trace has `requires_grad` set to True for `sample` sites, and the `param` sites detached.

pyro/poutine/lower_poutine.py
```diff
@@ -0,0 +1,36 @@
...
+            msg["is_observed"] = True
+            msg["value"] = _PYRO_PARAM_STORE.get_param(param_name)
+            value = super(LowerPoutine, self)._pyro_sample(msg)
+        else:
+            # Sample an initial param value from the prior.
+            value = super(LowerPoutine, self)._pyro_sample(msg)
+            value = Variable(value.data, requires_grad=True)
+            msg["value"] = value
+            msg["is_observed"] = True
+        return pyro.param(param_name, value)
```
eb8680: Wouldn't it be better if `param_name` were changed and `msg["name"]` were preserved?

fritzo: I tried that at first, but it seemed awkward. We have to decide between pairs of site names, either {"mu", "prior.mu"} or {"mu", "lowered.mu"}. I figure that since the user will almost always want to read the `param` site and almost never want to read the `observe` site, it would be better to preserve the param site. This also has the benefit that the site name will be stored in the param store, which is the same effect that `pyro.param` has (i.e. the statement label ends up in the param store).

tests/infer/test_map.py
```diff
@@ -0,0 +1,75 @@
...
+    for _ in range(100):
+        infer.step(data)
+
+    mu = pyro.param("mu")
+    assert_equal(mu.data[0], 2.0, prec=0.1)
+
+
+def test_map_chain_rule():
+
+    def model():
```
neerajprad: Nice! :)

pyro/poutine/lower_poutine.py
eb8680: >  I figure that since the user will almost always want to read the param site and almost never want to read the observe site, it would be better to preserve the param site.

How should users be getting the results of inference?  In SVI, they can get stuff from the guide, and they can always use `Marginal`, but for `MAP` they're supposed to query the parameter store?

fritzo: As I understand, the results of inference are optimized parameters. For both SVI and MAP, users get those results from the parameter store.

fritzo: This also has the property that `pyro.lower(pyro.lift(fn, prior))` merely inserts `pyro.observe("prior.*", ...)` sites for each parameter (i.e. the param sites are preserved).

pyro/infer/map.py
fehiepsi: Is it better to define MAP as a class (as a sub-class of SVI)? I know the returned value would be the same. But with class, we can inherit to define other inference such as Quadratic approximation (i.e. Laplace approximation), which requires computing hessian after getting the MAP.

neerajprad:
> We currently have no way to restrict optimizer moves to the support of constrained distributions (like Exponential or HalfCauchy).

We will also need this for HMC.

fritzo:
@eb8680 Have you had time to think about the parameter naming thing? My thought was that if `pyro.condition` could condition on noisily observed data (via distributions rather than exact values), and if we use the current naming convention (`mu`, `prior.mu`), then we might satisfy a nice equation like
```py
pyro.lower(pyro.lift(fn, prior)) == pyro.condition(fn, prior)
```

fehiepsi:
We also need to add `from pyro.infer.map import MAP` to `__init__.py` file.

ngoodman:
how come poutine.lower doesn't just convert a sample into a param? that is, why is there an inserted observe? sorry if i'm missing something obvious (and sorry if this is discussed somewhere... couldn't see it in a skim of the many linked issues).

in my initial imagination poutine.lower is a simple one-sided inverse to poutine.lift.

fritzo:
> how come poutine.lower doesn't just convert a sample into a param

The `observe` statement in `poutine.lower` accounts for the prior. Without the `observe`, this would implement Maximum Likelihood rather than MAP.

fritzo:
> We also need to add from pyro.infer.map import MAP to __init__.py file.

Done.

ngoodman:
> The observe statement in poutine.lower accounts for the prior. Without the observe, this would implement Maximum Likelihood rather than MAP.

hmm. i think a nicer idiom (which is a bit closer to the delta-guide version) is to make a guide by lowering the model. ie MAP is something like `optimize(model, lower(model))`, and MLE is `optimize(lower(model), null)`. for this version lower doesn't need to add an observe, and is thus the correct (1-sided) inverse to lift. thoughts?

we could of course have a helper function `MAP(model)` that is equivalent to `optimize(model, lower(model))`.

fritzo:
> a nicer idiom (... closer to the delta-guide version) is to make a guide by lowering the model

Yes, this is how @eb8680 and I have been understanding what the current implementation of `poutine.lower` does. The issue @eb8680 pointed out was that the new guide would duplicate all computation in the model. We're viewing the current implementation as an optimization of the automatic delta-guide version (we're even considering naming sites as `mu` and `guide.mu`) to avoid duplicated computation.

ngoodman:
>  the current implementation as an optimization of the automatic delta-guide version 

that makes sense, but how important do you think that overhead is? at any rate, it doesn't seem worth mucking up the abstraction for poutine.lower.... perhaps build this out of existing lower and an additional poutine ("regularize_param" or something)?

fritzo:
> how important do you think that overhead is

The overhead is often a factor of 2, e.g. when we add a bit of probability to a nn.

I feel like our poutine abstractions should conform to the computational patterns we need, rather than the other way around. This is also an issue in HMC #579 where the poutine mechanism suffices, but where existing basic poutines do not allow us to compute exactly the gradients we want. These are two valuable pieces of evidence that (1) the poutine abstraction is sufficient for many inference algorithms, but (2) our current set of basic poutines is incomplete.

fritzo:
@eb8680 @ngoodman 
Should `MAP` modify only values of latent variables or also values of parameters?

This PR currently optimizes both latent variables and parameters. This is in contrast to Stan's `optimize` which implements posterior mode for fixed params. If we follow Stan and optimize only latent variables, then MAP looks much closer to HMC and acts more like a `Delta` posterior (accessible via `Marginal`). This seems closer to the interface @eb8680 wants. In this version, we could still use `poutine.lift` to convert params to latents amenable to `MAP` inference, and then implement `poutine.lower` closer to @ngoodman 's version that removes `observe` statements. Crucially the `MAP` interface would be decoupled from the `lift : param -> latent` and `lower : latent -> param` interfaces and would not interact with the param store.

fritzo:
Bump. Let's discuss and try to get this merged this week.

tests/infer/test_map.py
```diff
@@ -0,0 +1,75 @@
...
+        infer.step(data)
+
+    mu = pyro.param("mu")
+    assert_equal(mu.data[0], 2.0, prec=0.1)
+
+
+def test_map_param():
+
+    def model(data):
+        mu = pyro.param("mu", Variable(torch.zeros(0), requires_grad=True))
```
fehiepsi: It should be `torch.zeros(1)` here.

tests/infer/test_map.py
```diff
@@ -0,0 +1,75 @@
...
+    mu = pyro.param("mu")
+    assert_equal(mu.data[0], 2.0, prec=0.1)
+
+
+def test_map_chain_rule():
+
+    def model():
+        x = pyro.sample("x", dist.normal, ng_zeros(1), ng_ones(1))
+        y = pyro.sample("y", dist.normal, x, ng_ones(1))
+        pyro.observe("z", dist.normal, y, ng_ones(1) * 3.0, ng_ones(1))
```
neerajprad: Shouldn't the third argument be `ng_ones * 3.0` (observations)?

fehiepsi: Theoretically, it is reasonable. I think that `p(z=y|x,y)` is likelihood and `p(x,y|z=y)` is posterior. Solving MAP by hand, we get to the system of equations `2y - x - 3 = 0` and `2x - y = 0`, which in turn yields `x = 1` and `y = 2`.

neerajprad: @fehiepsi - The third argument to `observe` is the data (followed by the distribution parameters), and I thought the test intended the observed data to be `[3]`. Though you are right that it wouldn't change the solution in this case, since the normal distribution is symmetric.

fritzo: Right, `pyro.observe("x", dist.normal, a, b, c)` is symmetric in `(a,b)`. I was thinking of this `observe` statement as an additional prior that we're putting on `y`, saying that it should be close to 3. The test should produce the same result either way.

fritzo:
Bump. Any chance we can merge this week? If not I'll close this PR.

fritzo:
Closing as too controversial.

PR #575, opened 2017-11-14
Warn about NANs in log_pdf and loss

### LINKED ISSUE #571: How to find cause of nan error in ELBO function
I am trying to modify the `vae.py` example, and the ELBO loss function is returning `nan` before I even do any training steps. My `model` and `guide` functions give sensible output, so I have no way to guess where the error is.  Thanks for your help.

I wouldn't typically open a github issue for this kind of question, but I am not sure what the best place to get support for `pyro` is. Thanks!

fritzo:
Fixes #571 

This adds warnings for:
- NAN and +INF `Trace.*log_pdf()` methods (at each site)
- NAN `.loss()` and `.loss_and_grads()` of in ELBO implementations

Note that this does not print whether errors occurred in the model vs guide, but that information can be seen from the stack trace if a user promotes warnings to errors via
```py
import warnings
warnings.simplefilter("error")
```
or by running a script via `python -W error my_script.py`.

## Tested

I have not added any death tests.

### DIFF
--- a/pyro/poutine/trace.py
+++ b/pyro/poutine/trace.py
@@ -1,8 +1,15 @@
 from __future__ import absolute_import, division, print_function
 
 import collections
+import warnings
 
 import networkx
+import numpy as np
+
+
+def _warn_if_nan(name, variable):
+    if np.isnan(variable.data[0]):
+        warnings.warn("Encountered NAN log_pdf at site '{}'".format(name))
 
 
 class Trace(networkx.DiGraph):
@@ -70,6 +77,7 @@ class Trace(networkx.DiGraph):
                     site_log_p = site["fn"].log_pdf(
                         site["value"], *args, **kwargs) * site["scale"]
                     site["log_pdf"] = site_log_p
+                    _warn_if_nan(name, site_log_p)
                 log_p += site_log_p
         return log_p
 
@@ -91,6 +99,7 @@ class Trace(networkx.DiGraph):
                         site["value"], *args, **kwargs) * site["scale"]
                     site["batch_log_pdf"] = site_log_p
                     site["log_pdf"] = site_log_p.sum()
+                    _warn_if_nan(name, site["log_pdf"])
                 # Here log_p may be broadcast to a larger tensor:
                 log_p = log_p + site_log_p
         return log_p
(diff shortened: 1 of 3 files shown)


fritzo:
@nbren12 Do you think this would have helped catch your error faster?

pyro/poutine/trace.py
```diff
@@ -1,8 +1,19 @@
...
 
 import collections
+import warnings
 
 import networkx
+import numpy as np
+
+
+def _warn_if_nan(name, variable):
+    value = variable.data[0]
```
neerajprad: This seems preferable so as to avoid large tensor checks. I suppose this will leave out some small cases where the NaNs/Infs are in different locations and don't propagate across the tensor?

pyro/poutine/trace.py
```diff
@@ -1,8 +1,19 @@
 from __future__ import absolute_import, division, print_function
 
 import collections
+import warnings
 
 import networkx
+import numpy as np
+
+
+def _warn_if_nan(name, variable):
```
nbren12: Is this `name` argument usually something that is easy to interpret?

fritzo: It is identical to the "site_name" in `pyro.sample("site_name", dist.normal, sigma, mu)` [sic] or `pyro.observe("site_name", dist.normal, mu, sigma, x)` [sic].

nbren12:
Thanks for putting this together so fast! I think this should be pretty nice for me. I think it would be really nice to add a short "troubleshooting" or "debugging" page to the docs with a description of how to track down NaN errors in the loss by elevating warning, looking at stack traces, and maybe doing some step by step debugging.

pyro/poutine/trace.py
fritzo: We're only checking on `log_pdf` which has a single element. The reduction is already done by `batch_log_pdf()` via `.sum()`, so we do indeed check for NANs everywhere.

pyro/poutine/trace.py
nbren12: ok. so it wouldn't tell you whether the problem is in the guide or the model, but that is handled by examining the stack traces as you describe above I guess?

fritzo: Yeah this won't distinguish guide-vs-model unless you examine the stack trace. It would take a lot more plumbing to add that to the warning message. But I do agree that it would help. We'll keep our eye out for better ways to add model-vs-guide info to warning messages.

fritzo:
Ready to merge.

PR #576, opened 2017-11-14
Fix event_shape in TransformedDistribution

### LINKED ISSUE #574: TransformedDistribution's event_shape forwards to incorrect base distribution's method
Probably a typo:

https://github.com/uber/pyro/blob/51a2ccfe9445c7072c3150c4abe1ab1d2ac17246/pyro/distributions/transformed_distribution.py#L62

neerajprad:
The `event_shape` for TransformedDistribution was incorrectly calling `batch_shape` of the base class distribution.

Added a test case to ensure that the sample size drawn from the distribution is consistent with the batch/event shape properties.

Fixes #574.


fritzo:
I'll merge when tests pass.

neerajprad:
Should be good to merge.

Issue #44, opened 2017-07-12
Abstract Trace_kernel class from which MCMC kernels will inherit.


ngoodman:
Kernels take a trace and return a trace.

fritzo:
Am I correct that kernels must take a (model, trace) pair and return a trace? I believe a model is needed because kernels will need to run model code to correctly set up gradients among `Variable`s in the trace.

ngoodman:
yes, the kernel is model specific. but rather than having the kernel take the model each time, i think we might want the kernel constructor to take the model, and the kernel to just take the trace.

eb8680:
First version implemented in #579 by @neerajprad

PR #579, opened 2017-11-15
[Discussion][WIP] HMC implementation for Pyro

neerajprad:
This is to get a discussion started around how the API for HMC should look like in Pyro. I have put in a very bare-bones HMC implementation (not well tested, or modularized) to guide the discussion. I have already had some really helpful discussions with @fritzo. I will create separate issues once we converge on the API and decide on what we would like to support in the very first version.

This version does the following:
 - The HMC class inherits from `TracePosterior` and returns a generator over traces.
 - It currently caches the values of the parameter gradients and resets them once the simulation is over. This is to ensure that none of the parameter values or their gradients are modified by the algorithm.
 -  The posterior is evaluated using all the data points. (see below).

### Discussion - Interface
 - [ ] Interface - Discussed with @fritzo, and one way to design the interface would be to in-place optimize the parameter values (consistent with #573) by placing an improper prior over them. It seem like this is the direction that Stan has taken too (and Tensorflow might take as well). Right now, the param values are treated as fixed and the trace can be used to get the posterior distribution over the sample sites. 
 - [ ] [If HMC inherits off of TracePosterior] We can use this with marginal and draw samples using `marginal(data)`. Since there is no data subsampling, it seems strange to have to supply the data argument each time (though I suppose `_dist_and_values` memoizes the call, so we are probably not rerunning the sim using the same data. Need to verify this!). Are there any other helper functions that we may need in AbstractInfer class? 
 - [ ] Does this satisfy other use cases - e.g. using this from within SVI? 

### Discussion - Near Term
 - [ ] Handling distribution parameters with constrained support. Right now this will just result in a rejection.
 - [ ] Initializing a good value for the mass matrix. Currently this is just an identity matrix.

### Discussion - Longer Term

 - [x] Does it makes sense to have mini-batch / subsampling support in the first version? If so, what are the best algorithms to look. @karalets mentioned about using SGLD for doing this. Update - defer until later, not possible with naive HMC.
 - [ ] Need to benchmark this, but please let me know if you notice potential scaling issues (aside from subsampling).
 - [ ] Other general helper utilities for assessing convergence, etc?

### Discussion (resolved):
 - [x] The leapfrog (verlet) integration is coupled to Pyro internals as it takes the execution trace from the previous step (latent state). This is not ideal, but I found that this was the easiest way to not have to convert a dictionary to a trace and vice-versa, as this can be directly passed to the function that computes the gradient of the potential. Any thoughts on modularizing this? Update - this is decoupled now.

 
Please put in any other issues worth discussing. 

EDIT - After addressing some issues with the previous version, this runs on current PyTorch.

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -23,8 +23,8 @@ def verlet_integrator(z, r, grad_potential, step_size, num_steps):
     :return: (z_next, r_next) having same types as (z, r)
     """
     # deep copy the current state - (z, r)
-    z_next = {key: val.clone() for key, val in z.items()}
-    r_next = {key: val.clone() for key, val in r.items()}
+    z_next = {key: val.clone().detach() for key, val in z.items()}
+    r_next = {key: val.clone().detach() for key, val in r.items()}
     retain_grads(z_next)
     retain_grads(r_next)
     grads = grad_potential(z_next)
@@ -50,6 +50,9 @@ def verlet_integrator(z, r, grad_potential, step_size, num_steps):
 
 def retain_grads(z):
     for key in z:
+        # XXX: can be removed with PyTorch 0.3
+        if z[key].is_leaf and not z[key].requires_grad:
+            z[key].requires_grad = True
         z[key].retain_grad()
 
 
@@ -189,8 +192,8 @@ def test_verlet_integrator():
     def grad(q):
         return {'x': q['x']}
 
-    q = {'x': Variable(torch.Tensor([0.0]))}
-    p = {'x': Variable(torch.Tensor([1.0]))}
+    q = {'x': Variable(torch.Tensor([0.0]), requires_grad=True)}
+    p = {'x': Variable(torch.Tensor([1.0]), requires_grad=True)}
     energy_cur = energy(q, p)
     print("Energy - current: {}".format(energy_cur.data[0]))
     q_new, p_new = verlet_integrator(q, p, grad, 0.01, 100)
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -75,26 +75,26 @@ class HMC(TracePosterior):
         self.warmup_steps = warmup_steps
         # simulation run attributes - will be set in self._setup
         # at start of run
-        self.cached_param_grads = {}
-        self.args = None
-        self.kwargs = None
-        self.accept_cnt = None
-        self.prototype_trace = None
+        self._cached_param_grads = {}
+        self._args = None
+        self._kwargs = None
+        self._accept_cnt = None
+        self._prototype_trace = None
         super(HMC, self).__init__()
 
-    def log_prob(self, z):
+    def _log_prob(self, z):
         """
         Return log pdf of the model with sample sites replayed from z_trace
         """
-        z_trace = self.prototype_trace.copy()
+        z_trace = self._prototype_trace.copy()
         for name, value in z.items():
             z_trace.nodes[name]['value'] = value
         model_trace = poutine.trace(poutine.replay(self.model, trace=z_trace)) \
-            .get_trace(*self.args, **self.kwargs)
+            .get_trace(*self._args, **self._kwargs)
         return model_trace.log_pdf()
 
-    def grad_potential(self, z):
-        log_joint_prob = self.log_prob(z)
+    def _grad_potential(self, z):
+        log_joint_prob = self._log_prob(z)
         log_joint_prob.backward()
         grad_potential = {}
         for name, value in z.items():
@@ -102,34 +102,34 @@ class HMC(TracePosterior):
             grad_potential[name].volatile = False
         return grad_potential
 
+    def _energy(self, z, r):
+        kinetic_energy = 0.5 * torch.sum(torch.stack([r[name] ** 2 for name in r]))
+        potential_energy = - self._log_prob(z)
+        return kinetic_energy + potential_energy
+
     def _setup(self, *args, **kwargs):
-        self.accept_cnt = 0
-        self.args = args
-        self.kwargs = kwargs
+        self._accept_cnt = 0
+        self._args = args
+        self._kwargs = kwargs
         # set the trace prototype to inter-convert between trace object
         # and dict object used by the integrator
-        self.prototype_trace = poutine.trace(self.model).get_trace(*args, **kwargs)
+        self._prototype_trace = poutine.trace(self.model).get_trace(*args, **kwargs)
         # store the current value of param gradients so that they
         # can be reset at the end
-        for name, node in self.prototype_trace.iter_param_nodes():
-            self.cached_param_grads[name] = node['value'].grad
+        for name, node in self._prototype_trace.iter_param_nodes():
+            self._cached_param_grads[name] = node['value'].grad
 
     def _cleanup(self):
         # reset the param values to those stored before the hmc run
-        for name, grad in self.cached_param_grads.items():
+        for name, grad in self._cached_param_grads.items():
             param = pyro.get_param_store().get_param(name)
             param.grad = grad
-
-    def energy(self, z, r):
-        kinetic_energy = 0.5 * torch.sum(torch.stack([r[name] ** 2 for name in r]))
-        potential_energy = - self.log_prob(z)
-        return kinetic_energy + potential_energy
 
     def _traces(self, *args, **kwargs):
         self._setup(*args, **kwargs)
         # sample p's from the distribution given by p_dist
         r_dist = {}
-        z = {name: node['value'] for name, node in self.prototype_trace.iter_stochastic_nodes()}
+        z = {name: node['value'] for name, node in self._prototype_trace.iter_stochastic_nodes()}
         for name, value in z.items():
             r_mu = torch_zeros_like(value)
             r_sigma = torch_ones_like(value)
@@ -145,26 +145,29 @@ class HMC(TracePosterior):
             r = {name: pyro.sample('r_{}_t={}'.format(name, t), r_dist[name]) for name in z}
             z_new, r_new = verlet_integrator(z,
                                              r,
-                                             self.grad_potential,
+                                             self._grad_potential,
                                              self.step_size,
                                              self.num_steps)
-            energy_proposal = self.energy(z_new, r_new)
-            energy_current = self.energy(z, r)
+            energy_proposal = self._energy(z_new, r_new)
+            energy_current = self._energy(z, r)
             # print("Energy - current: {}".format(energy_current))
             # print("Energy - proposal: {}".format(energy_proposal))
             delta_energy = energy_proposal - energy_current
             t += 1
             rand = pyro.sample('rand_t='.format(t), dist.uniform, a=ng_zeros(1), b=ng_ones(1))
             if rand.log().data[0] < -delta_energy.data[0]:
-                self.accept_cnt += 1
+                self._accept_cnt += 1
                 z = z_new
             if t < self.warmup_steps:
                 continue
-            yield (z, self.log_prob(z))
+            yield (z, self._log_prob(z))
         self._cleanup()
 
-    def get_acceptance_ratio(self):
-        return self.accept_cnt / (self.warmup_steps + self.num_samples)
+    @property
+    def acceptance_ratio(self):
+        if self._accept_cnt is None:
+            raise AttributeError("The attribute _accept_cnt is not updated until the first simulation run.")
+        return self._accept_cnt / (self.warmup_steps + self.num_samples)
 
 
 def test_normal_normal():
@@ -178,7 +181,7 @@ def test_normal_normal():
     traces = []
     for t, _ in hmc._traces(data):
         traces.append(t['x'])
-    print('Acceptance ratio: {}'.format(hmc.get_acceptance_ratio()))
+    print('Acceptance ratio: {}'.format(hmc.acceptance_ratio))
     print('Posterior mean:')
     print(torch.mean(torch.stack(traces), 0).data)
     # gradients should not have been back-propagated.
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -1,0 +1,197 @@
+import random
+
+import torch
+from torch.autograd import Variable
+
+import pyro
+import pyro.distributions as dist
+import pyro.poutine as poutine
+from pyro.distributions.util import torch_zeros_like, torch_ones_like
+from pyro.infer import TracePosterior
+from pyro.poutine import Trace
+from tests.common import assert_equal
+
+
+def verlet_integrator(z, r, grad_potential, step_size, num_steps):
+    """
+    Velocity Verlet integrator.
+
+    :param z: trace object containing current values for the sample sites
+    :param r: dictionary of sample site names and corresponding momenta
+    :param grad_potential: function that returns gradient of the potential given z
+        for each sample site
+    :return: (z_next, r_next) having same types as (z, r)
+    """
+    # deep copy the current state - (z, r)
+    z_next = clone_trace(z)
+    r_next = {k: v.clone() for k, v in r.items()}
+    grads = grad_potential(z_next)
+    for _ in range(num_steps):
+        for site_name, node in z_next.iter_stochastic_nodes():
+            # r(n+1/2)
+            r_next[site_name] = r_next[site_name] + 0.5 * step_size * (-grads[site_name])
+            # z(n+1)
+            node['value'] = node['value'] + step_size * r_next[site_name]
+            node['value'].retain_grad()
+        grads = grad_potential(z_next)
+        for site_name in r_next:
+            # r(n+1)
+            r_next[site_name] = r_next[site_name] + 0.5 * step_size * (-grads[site_name])
+    return z_next, r_next
+
+
+def clone_trace(trace):
+    # clone the tensors so as not to mutate the old state (q, p)
+    # if it is not accepted
+    new_trace = trace.copy()
+    for name, node in trace.iter_stochastic_nodes():
+        node['value'].retain_grad()
+        node['value'] = node['value'].clone()
+    return new_trace
+
+
+class HMC(TracePosterior):
+    def __init__(self,
+                 model,
+                 step_size=0.5,
+                 num_steps=3,
+                 warmup_steps=20,
+                 num_samples=1000):
+        self.model = model
+        self.step_size = step_size
+        self.num_steps = num_steps
+        self.num_samples = num_samples
+        self.warmup_steps = warmup_steps
+        self.cached_param_grads = {}
+        self.args = None
+        self.kwargs = None
+        super(HMC, self).__init__()
+
+    def log_prob(self, z_trace):
+        """
+        Return log pdf of the model with sample sites replayed from z_trace
+        """
+        model_trace = poutine.trace(poutine.replay(self.model, trace=z_trace)).get_trace(*self.args, **self.kwargs)
+        return model_trace.log_pdf()
+
+    def grad_potential(self, z):
+        log_joint_prob = self.log_prob(z)
+        log_joint_prob.backward()
+        grad_potential = {}
+        for name, node in z.iter_stochastic_nodes():
+            grad_potential[name] = -node['value'].grad.clone()
+            grad_potential[name].volatile = False
+        return grad_potential
+
+    def _setup(self, z):
+        # store the current value of param gradients so that they
+        # can be reset at the end
+        for name, node in z.iter_param_nodes():
+            self.cached_param_grads[name] = node['value'].grad
+
+    def _cleanup(self):
+        # reset the param values to those stored before the hmc run
+        for name, grad in self.cached_param_grads.items():
+            param = pyro.get_param_store().get_param(name)
+            param.grad = grad
+
+    def energy(self, z, r):
+        kinetic_energy = 0.5 * torch.sum(torch.stack([r[name] ** 2 for name in r]))
+        potential_energy = - self.log_prob(z)
+        return kinetic_energy + potential_energy
+
+    def _traces(self, *args, **kwargs):
+        t = 0
+        self.args = args
+        self.kwargs = kwargs
+        # initialize z with model trace
+        z = poutine.trace(self.model).get_trace(*args, **kwargs)
+        self._setup(z)
+        # sample p's from the distribution given by p_dist
+        r_dist = {}
+        for name, node in z.iter_stochastic_nodes():
+            r_mu = torch_zeros_like(node['value'])
+            r_sigma = torch_ones_like(node['value'])
+            r_dist[name] = dist.Normal(mu=r_mu, sigma=r_sigma)
+        self.accept_cnt = 0
+
+        # Run HMC iterations
+        print('Simulating using HMC...')
+        while t < self.warmup_steps + self.num_samples:
+            if t % 100 == 0:
+                print('Iteration: {}'.format(t))
+            # sample momentum
+            r = {name: r_dist[name].sample() for name in z.stochastic_nodes}
+            z_new, r_new = verlet_integrator(z,
+                                             r,
+                                             self.grad_potential,
+                                             self.step_size,
+                                             self.num_steps)
+            energy_proposal = self.energy(z_new, r_new)
+            energy_current = self.energy(z, r)
+            # print("Energy - current: {}".format(energy_current))
+            # print("Energy - proposal: {}".format(energy_proposal))
+            delta_energy = energy_proposal - energy_current
+            t += 1
+            if random.random() < (-delta_energy).exp().data[0]:
+                self.accept_cnt += 1
+                z = z_new
+            if t < self.warmup_steps:
+                continue
+            yield (z, z.log_pdf())
+        self._cleanup()
+
+    def get_acceptance_ratio(self):
+        return self.accept_cnt / float(self.warmup_steps + self.num_samples)
+
+
+def test_normal_normal():
+    def model(data):
+        mu = pyro.param('mu', Variable(torch.zeros(10), requires_grad=True))
+        x = pyro.sample('x', dist.normal, mu=mu, sigma=Variable(torch.ones(10)))
+        pyro.sample('data', dist.normal, obs=data, mu=x, sigma=Variable(torch.ones(10)))
+
+    data = Variable(torch.ones(1, 10))
+    hmc = HMC(model, step_size=0.4, num_steps=3, num_samples=400, warmup_steps=50)
+    traces = []
+    for t, _ in hmc._traces(data):
+        traces.append(t.nodes['x']['value'])
+    print('Acceptance ratio: {}'.format(hmc.get_acceptance_ratio()))
+    print('Posterior mean:')
+    print(torch.mean(torch.stack(traces), 0).data)
+    # gradients should not have been back-propagated.
+    assert pyro.get_param_store().get_param('mu').grad is None
+
+
+def test_verlet_integrator():
+    def energy(q, p):
+        return 0.5 * p['x'] ** 2 + 0.5 * q.nodes['x']['value'] ** 2
+
+    def grad(q):
+        return {'x': q.nodes['x']['value']}
+
+    q = Trace()
+    q.add_node('x',
+               type='sample',
+               is_observed=False,
+               value=Variable(torch.Tensor([0.0])))
+    p = {'x': Variable(torch.Tensor([1.0]))}
+    energy_cur = energy(q, p)
+    print("Energy - current: {}".format(energy_cur.data[0]))
+    q_new, p_new = verlet_integrator(q, p, grad, 0.01, 100)
+    assert q_new.nodes['x']['value'].data[0] != q.nodes['x']['value'].data[0]
+    energy_new = energy(q_new, p_new)
+    assert_equal(energy_new, energy_cur)
+    print("q_old: {}, p_old: {}".format(q.nodes['x']['value'].data[0], p['x'].data[0]))
+    print("q_new: {}, p_new: {}".format(q_new.nodes['x']['value'].data[0], p_new['x'].data[0]))
+    print("Energy - new: {}".format(energy_new.data[0]))
+    print("-------------------------------------")
+
+
+def main():
+    test_verlet_integrator()
+    test_normal_normal()
+
+
+if __name__ == '__main__':
+    main()
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
+import numpy as np
 import torch
 from torch.autograd import Variable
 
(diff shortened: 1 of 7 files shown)


pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,197 @@
...
+                                             self.grad_potential,
+                                             self.step_size,
+                                             self.num_steps)
+            energy_proposal = self.energy(z_new, r_new)
+            energy_current = self.energy(z, r)
+            # print("Energy - current: {}".format(energy_current))
+            # print("Energy - proposal: {}".format(energy_proposal))
+            delta_energy = energy_proposal - energy_current
+            t += 1
+            if random.random() < (-delta_energy).exp().data[0]:
```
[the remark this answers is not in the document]
neerajprad: Makes sense, will change.

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,197 @@
...
+            r_dist[name] = dist.Normal(mu=r_mu, sigma=r_sigma)
+        self.accept_cnt = 0
+
+        # Run HMC iterations
+        print('Simulating using HMC...')
+        while t < self.warmup_steps + self.num_samples:
+            if t % 100 == 0:
+                print('Iteration: {}'.format(t))
+            # sample momentum
+            r = {name: r_dist[name].sample() for name in z.stochastic_nodes}
```
eb8680: Let's also make these into `pyro.sample` calls

neerajprad: Yes, will change.

neerajprad:
> does this run? i was writing a test but im getting:

You should be able to run this using PyTorch master (and likely the upcoming release). Could you try that? The current release anyways segfaults, and I have not been able to get it to work on that.

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,197 @@
...
+            r_next[site_name] = r_next[site_name] + 0.5 * step_size * (-grads[site_name])
+    return z_next, r_next
+
+
+def clone_trace(trace):
+    # clone the tensors so as not to mutate the old state (q, p)
+    # if it is not accepted
+    new_trace = trace.copy()
+    for name, node in trace.iter_stochastic_nodes():
+        node['value'].retain_grad()
```
fritzo: I'm surprised you call `.retain_grad()` before `.clone()`ing, rather than after.

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,197 @@
...
+from pyro.infer import TracePosterior
+from pyro.poutine import Trace
+from tests.common import assert_equal
+
+
+def verlet_integrator(z, r, grad_potential, step_size, num_steps):
+    """
+    Velocity Verlet integrator.
+
+    :param z: trace object containing current values for the sample sites
```
fritzo: This `verlet_integrator()` would be more valuable if it were less dependent on `Trace` objects, say if `z` and `r` were both simple python dicts of `torch.autograd.Variable`s. If we could eliminate all Pyro dependencies, then `verlet_integrator()` could be reused outside of the context of tracing. Moreover it would be possible to unit test without a `Trace` or `HMC` object.

I believe this could be accomplished by adding a little complexity to `HMC.log_pdf()`, say a `self.prototype_trace` onto which `z` could be spliced prior to `replay`. You've already done much of the work by abstracting out `grad_potential`.

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,197 @@
...
+        self.warmup_steps = warmup_steps
+        self.cached_param_grads = {}
+        self.args = None
+        self.kwargs = None
+        super(HMC, self).__init__()
+
+    def log_prob(self, z_trace):
+        """
+        Return log pdf of the model with sample sites replayed from z_trace
+        """
```
fritzo: To use dict-like `z`s in stead of traces, you could
```py
z_trace = self.prototype_trace.copy()
for name, value in z.items():
    z_trace.nodes['name']['value'] = value
```

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,197 @@
...
+        kinetic_energy = 0.5 * torch.sum(torch.stack([r[name] ** 2 for name in r]))
+        potential_energy = - self.log_prob(z)
+        return kinetic_energy + potential_energy
+
+    def _traces(self, *args, **kwargs):
+        t = 0
+        self.args = args
+        self.kwargs = kwargs
+        # initialize z with model trace
+        z = poutine.trace(self.model).get_trace(*args, **kwargs)
```
fritzo: To use dict-like `z` values, I believe you could
```py
self.prototype_trace = poutine.trace(self.model).get_trace(*args, **kwargs)
z = {name: node['value'] for name, node in self.prototype_trace.iter_stochastic_nodes()}
```

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,197 @@
...
+    # deep copy the current state - (z, r)
+    z_next = clone_trace(z)
+    r_next = {k: v.clone() for k, v in r.items()}
+    grads = grad_potential(z_next)
+    for _ in range(num_steps):
+        for site_name, node in z_next.iter_stochastic_nodes():
+            # r(n+1/2)
+            r_next[site_name] = r_next[site_name] + 0.5 * step_size * (-grads[site_name])
+            # z(n+1)
+            node['value'] = node['value'] + step_size * r_next[site_name]
```
fritzo: I'm surprised you don't `.detach()` here. Wouldn't this implementation propagate gradients far far back in time, through all previous time steps? If so, this might be a possible cause of segfault on PyTorch 0.2.

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        model_trace = poutine.trace(poutine.replay(self.model, trace=z_trace)).get_trace(*self.args, **self.kwargs)
+        return model_trace.log_pdf()
+
+    def grad_potential(self, z):
+        log_joint_prob = self.log_prob(z)
+        log_joint_prob.backward()
+        grad_potential = {}
+        for name, node in z.iter_stochastic_nodes():
+            grad_potential[name] = -node['value'].grad.clone()
```
fritzo: `node['value'].grad.clone().detach()` here?

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,197 @@
...
+            if random.random() < (-delta_energy).exp().data[0]:
+                self.accept_cnt += 1
+                z = z_new
+            if t < self.warmup_steps:
+                continue
+            yield (z, z.log_pdf())
+        self._cleanup()
+
+    def get_acceptance_ratio(self):
+        return self.accept_cnt / float(self.warmup_steps + self.num_samples)
```
fritzo: nit: if you `from __future__ import division`, then you won't need the `float` cast :wink:

pyro/infer/mcmc/hmc.py
neerajprad: Oops..that definitely seems problematic. Let me take a look. Thanks for pointing out!

pyro/infer/mcmc/hmc.py
neerajprad: Cool, will give this a try. I initially had a `dict` for `z`, but had put in some complexity in `ReplayPoutine` itself. I think this should be a relatively light weight wrapper, and we can do this within this class itself.

pyro/infer/mcmc/hmc.py
neerajprad: There are a couple of funny things going on here, will correct it. I need to find a cleaner way of handling when to `detach`, `retain_grad` and `clone`.

neerajprad:
@fritzo - Thanks for the very helpful comments. I have addressed yours and @eb8680's comments, and it now runs on the current PyTorch release. I will focus on adding some good test cases. The integrator  now operates on python `dicts` and not trace objects, and the logic around when to retain, clone and detach should be cleaner.

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,215 @@
...
+                                             self.grad_potential,
+                                             self.step_size,
+                                             self.num_steps)
+            energy_proposal = self.energy(z_new, r_new)
+            energy_current = self.energy(z, r)
+            # print("Energy - current: {}".format(energy_current))
+            # print("Energy - proposal: {}".format(energy_proposal))
+            delta_energy = energy_proposal - energy_current
+            t += 1
+            rand = pyro.sample('rand', dist.uniform, a=ng_zeros(1), b=ng_ones(1))
```
eb8680: Name should contain `t`

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,215 @@
...
+            r_dist[name] = dist.Normal(mu=r_mu, sigma=r_sigma)
+
+        # Run HMC iterations
+        t = 0
+        print('Simulating using HMC...')
+        while t < self.warmup_steps + self.num_samples:
+            if t % 100 == 0:
+                print('Iteration: {}'.format(t))
+            # sample momentum
+            r = {name: pyro.sample('r_{}'.format(name), r_dist[name]) for name in z}
```
eb8680: Sample name should contain `t`

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,215 @@
...
+import pyro
+import pyro.distributions as dist
+import pyro.poutine as poutine
+from pyro.distributions.util import torch_ones_like, torch_zeros_like
+from pyro.infer import TracePosterior
+from pyro.util import ng_ones, ng_zeros
+from tests.common import assert_equal
+
+
+def verlet_integrator(z, r, grad_potential, step_size, num_steps):
```
eb8680: Now that this is using dictionaries, I think we should be able to convert the integrator to a `pyro.optim` optimizer and move it to `pyro.optim`

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,215 @@
...
+            grad_potential[name].volatile = False
+        return grad_potential
+
+    def _setup(self, *args, **kwargs):
+        self.accept_cnt = 0
+        self.args = args
+        self.kwargs = kwargs
+        # set the trace prototype to inter-convert between trace object
+        # and dict object used by the integrator
+        self.prototype_trace = poutine.trace(self.model).get_trace(*args, **kwargs)
```
eb8680: We should have some asserts here that check that there are no non-reparametrizable variables and that `self.prototype_trace.log_pdf() > 0`

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,215 @@
...
+        # simulation run attributes - will be set in self._setup
+        # at start of run
+        self.cached_param_grads = {}
+        self.args = None
+        self.kwargs = None
+        self.accept_cnt = None
+        self.prototype_trace = None
+        super(HMC, self).__init__()
+
+    def log_prob(self, z):
```
eb8680: nit: let's rename to `_log_prob`

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,215 @@
...
+        Return log pdf of the model with sample sites replayed from z_trace
+        """
+        z_trace = self.prototype_trace.copy()
+        for name, value in z.items():
+            z_trace.nodes[name]['value'] = value
+        model_trace = poutine.trace(poutine.replay(self.model, trace=z_trace)) \
+            .get_trace(*self.args, **self.kwargs)
+        return model_trace.log_pdf()
+
+    def grad_potential(self, z):
```
eb8680: nit: rename to `_grad_potential`

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,215 @@
...
+        for name, node in self.prototype_trace.iter_param_nodes():
+            self.cached_param_grads[name] = node['value'].grad
+
+    def _cleanup(self):
+        # reset the param values to those stored before the hmc run
+        for name, grad in self.cached_param_grads.items():
+            param = pyro.get_param_store().get_param(name)
+            param.grad = grad
+
+    def energy(self, z, r):
```
eb8680: nit: rename to `_energy`

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,215 @@
...
+            rand = pyro.sample('rand', dist.uniform, a=ng_zeros(1), b=ng_ones(1))
+            if rand.log().data[0] < -delta_energy.data[0]:
+                self.accept_cnt += 1
+                z = z_new
+            if t < self.warmup_steps:
+                continue
+            yield (z, self.log_prob(z))
+        self._cleanup()
+
+    def get_acceptance_ratio(self):
```
eb8680: nit: should this be a `@property`?

eb8680:
Nice work!  We should think about how this should be structured to comply with @ngoodman's request from #44 for webPPL-like MCMC kernel abstractions.

> Does it makes sense to have mini-batch / subsampling support in the first version?

No - unfortunately you can't subsample in naive HMC without introducing bias.  SGLD, SGHMC, and others are different algorithms that should be implemented separately.

> Does this satisfy other use cases - e.g. using this from within SVI?

This I'm not sure about, but my first guess would be that it doesn't because you're not calling `backward` with `retain_graph=True` to construct the autograd graph through the integrator.

pyro/infer/mcmc/hmc.py
neerajprad: Done.

pyro/infer/mcmc/hmc.py
neerajprad: Done.

pyro/infer/mcmc/hmc.py
neerajprad: Good point, will add this.

pyro/infer/mcmc/hmc.py
neerajprad: Yes, will take a look at the interface and update this.

eb8680: You should be able to crib from `pyro.optim.ClippedAdam`

pyro/infer/mcmc/hmc.py
eb8680: Sorry, I mean `self.prototype_trace.log_pdf() > -infinity`

pyro/infer/mcmc/hmc.py
neerajprad: Done.

pyro/infer/mcmc/hmc.py
neerajprad: Done. Prepended all internal attributes and methods with '_'.

pyro/infer/mcmc/hmc.py
neerajprad: If we do want it to be a PyroOptim object, it should probably return both the newly sampled param values as well as the "loss" (-delta energy). The optimizer would then need an additional `energy` function as argument. That said, the semantics are somewhat different from the other optimizers since we do not in-place modify the parameters based on some objective function. So we could also create a separate class for this in `pyro.optim`. What do you think?

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+    # deep copy the current state - (z, r)
+    z_next = {key: val.clone().detach() for key, val in z.items()}
+    r_next = {key: val.clone().detach() for key, val in r.items()}
+    retain_grads(z_next)
+    retain_grads(r_next)
+    grads = grad_potential(z_next)
+
+    for _ in range(num_steps):
+        # detach graph nodes for next iteration
```
fritzo: I believe you should call `detach_grads()` *after* the update step. Otherwise gradients will be propagated backwards one step in time and never accessed, right?

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,225 @@
...
+        retain_grads(r_next)
+        grads = grad_potential(z_next)
+        for site_name in r_next:
+            # r(n+1)
+            r_next[site_name] = r_next[site_name] + 0.5 * step_size * (-grads[site_name])
+    return z_next, r_next
+
+
+def retain_grads(z):
+    for key in z:
```
fritzo: nit: It's a bit cheaper (and a good habit) to iterate over `.items()` rather than iterate over `.keys()` and call `.__getitem__()` (here and in `detach_nodes()`.

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,225 @@
...
+        detach_nodes(z_next)
+        detach_nodes(r_next)
+        for site_name in z_next:
+            # r(n+1/2)
+            r_next[site_name] = r_next[site_name] + 0.5 * step_size * (-grads[site_name])
+            # z(n+1)
+            z_next[site_name] = z_next[site_name] + step_size * r_next[site_name]
+        # retain gradients for intermediate nodes in backward step
+        retain_grads(z_next)
+        retain_grads(r_next)
```
fritzo: I believe `r_next` need not retain grads, right?

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,225 @@
...
+        grads = grad_potential(z_next)
+        for site_name in r_next:
+            # r(n+1)
+            r_next[site_name] = r_next[site_name] + 0.5 * step_size * (-grads[site_name])
```
fritzo: Call `detach_grads()` *after* this loop?

neerajprad:
> We should think about how this should be structured to comply with @ngoodman's request from #44 for webPPL-like MCMC kernel abstractions.

I think we can implement an abstract MCMC class which passes the previous trace to an instance of TraceKernel class (in this case HMC) and gets a new execution trace. The abstract MCMC class should hold all the state.

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,225 @@
...
+        super(HMC, self).__init__()
+
+    def _log_prob(self, z):
+        """
+        Return log pdf of the model with sample sites replayed from z_trace
+        """
+        z_trace = self._prototype_trace.copy()
+        for name, value in z.items():
+            z_trace.nodes[name]['value'] = value
+        model_trace = poutine.trace(poutine.replay(self.model, trace=z_trace)) \
```
fritzo: I'm pretty sure you'll want a poutine here to `.detach()` params during replay so they won't accumulate gradients. What's implemented now smells bad due to:
- extra work performed during `.backward()` to propagate gradients all the way back to params, and
- dirty `.grad`s on all params in the param store (and it also smells bad to `zero_grads()` on the param store).

One way to get this behavior is to change `poutine.replay()` to replay `param` sites (in addition to `sample` sites). That way you can `.detach()` all param sites in the prototype trace and get the desired behavior. @eb8680 Do you have any objections to `poutine.replay()` behaving this way, so as to avoid yet another poutine?

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,225 @@
...
+        potential_energy = - self._log_prob(z)
+        return kinetic_energy + potential_energy
+
+    def _setup(self, *args, **kwargs):
+        self._accept_cnt = 0
+        self._args = args
+        self._kwargs = kwargs
+        # set the trace prototype to inter-convert between trace object
+        # and dict object used by the integrator
+        self._prototype_trace = poutine.trace(self.model).get_trace(*args, **kwargs)
```
fritzo: Ditto: `detach_grads()` of params during `replay()`.

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,225 @@
...
+        pyro.sample('data', dist.normal, obs=data, mu=x, sigma=Variable(torch.ones(10)))
+
+    data = Variable(torch.ones(1, 10))
+    hmc = HMC(model, step_size=0.6, num_steps=3, num_samples=400, warmup_steps=50)
+    traces = []
+    for t, _ in hmc._traces(data):
+        traces.append(t['x'])
+    print('Acceptance ratio: {}'.format(hmc.acceptance_ratio))
+    print('Posterior mean:')
+    print(torch.mean(torch.stack(traces), 0).data)
```
fritzo: It would be a good end-to-end test of the math to check the true-vs-empirical mean and standard deviation.

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,225 @@
...
+        # set the trace prototype to inter-convert between trace object
+        # and dict object used by the integrator
+        self._prototype_trace = poutine.trace(self.model).get_trace(*args, **kwargs)
+        for name, node in self._prototype_trace.iter_stochastic_nodes():
+            if not node['fn'].reparameterized:
+                raise ValueError('Found non-reparameterized node in the model at site: {}'.format(name))
+        prototype_trace_log_pdf = self._prototype_trace.log_pdf().data[0]
+        if np.isnan(prototype_trace_log_pdf) or np.isinf(prototype_trace_log_pdf):
+            raise ValueError('Model specification incorrect - trace log pdf is NaN, Inf or 0.')
+        # store the current value of param gradients so that they
```
fritzo: See above; this hacking would be simplified if we changed `poutine.replay()` to replay `param` sites.

pyro/infer/mcmc/hmc.py
fritzo: (or does `poutine.replay()` already do the right thing?)

pyro/infer/mcmc/hmc.py
neerajprad: The gradient has been computed right before the start of the loop, cloned and stored in `grads`. This `detach` call is to ensure that we do not propagate that graph further.

pyro/infer/mcmc/hmc.py
neerajprad: You are right, this is not needed. Will remove.

pyro/infer/mcmc/hmc.py
neerajprad: I can override `_pyro_param` in ReplayPoutine and reinstate some of the behavior that I had to begin with, if we all agree with this modification to ReplayPoutine. :)

> extra work performed during .backward() to propagate gradients all the way back to params, and

Detaching will not prevent this extra gradient flow back to the params right?

> dirty .grads on all params in the param store (and it also smells bad to zero_grads() on the param store).

Currently this is handled by the `_cleanup` operation, so that gradients on the params are reset; but I completely agree that detaching would be the cleaner way of doing this.

pyro/infer/mcmc/hmc.py
neerajprad: Yes, will add tests for that. This was only meant for my debugging. :)

pyro/infer/mcmc/hmc.py
neerajprad: I agree; you have me on board. I will make the change to `ReplayPoutine`.

pyro/infer/mcmc/hmc.py
eb8680: Sorry I can't respond to these very quickly, I'm busy working on a workshop submission due tomorrow, but I'm not sure about this `replay` change.

> I'm pretty sure you'll want a poutine here to .detach() params during replay so they won't accumulate gradients

This seems like premature optimization, but if you really want to account for this, just iterate over the parameters in `model_trace` and set `parameter.requires_grad=False` for each one immediately before you call `backward()`, then reset `parameter.requires_grad=True` immediately after.

fritzo: I think you'll want to do something like:
```py
for node in prototype_trace.nodes.values():
    if node['type'] == 'param':
        node['value'] = Variable(node['value'].data)  # detach and set requires_grad = False
    elif node['type'] == 'sample' and not node['is_observed']:
        node['value'] = Variable(node['value'].data, requires_grad=True)
```
which should indeed prevent gradients from being propagated to params.

pyro/infer/mcmc/hmc.py
eb8680: I may be too conservative here, but my inclination has been to have multiple different use cases and unit tests before adding new behavior to `trace` or `replay`, since they're used in almost every algorithm and could therefore absorb any given feature.  Poutines are only useful if we don't have to change them or write new ones for each new algorithm.

This particular behavior will undoubtedly show up in other gradient-based algorithms like SGLD, but my vote would be to wait until we implement those algorithms to incorporate it into `replay`.

pyro/infer/mcmc/hmc.py
neerajprad: @fritzo - There is a subtle issue with this. Since the parameters are installed in site args/kwargs and registered with the param store, this will still propagate the gradients back to the param store.

pyro/infer/mcmc/hmc.py
neerajprad: Makes sense. I will just make a note of it (and create an issue), which we can address after gathering some data points.

tests/infer/test_integrator.py
```diff
@@ -0,0 +1,80 @@
...
+    assert_equal(energy_new, energy_cur)
+    assert_equal(q_new['x'].data[0], np.sin(1.0), prec=1.0e-4)
+    assert_equal(p_new['x'].data[0], np.cos(1.0), prec=1.0e-4)
+    print("q_old: {}, p_old: {}".format(q['x'].data[0], p['x'].data[0]))
+    print("q_new: {}, p_new: {}".format(q_new['x'].data[0], p_new['x'].data[0]))
+    print("Energy - new: {}".format(energy_new.data[0]))
+    print("-------------------------------------")
+
+
+def test_circular_planetary_motion():
```
neerajprad: Thanks for adding these tests in, @martinjankowiak! I have moved them to test_integrator.py. Please feel free to modify/tune them further if needed.

tests/infer/test_hmc.py
```diff
@@ -0,0 +1,116 @@
...
+from pyro.infer.mcmc.mcmc import MCMC
+from pyro.infer.mcmc.verlet_integrator import verlet_integrator
+from tests.common import assert_equal
+
+logging.basicConfig(format='%(levelname)s %(message)s')
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+class GaussianChain(object):
```
neerajprad: These are simply borrowed from @martinjankowiak's conjugate gaussian tests, albeit not testing in all its generality. It will be nice to refactor them so that they are can be reused across different algorithms for comparison.

neerajprad:
I was discussing with @martinjankowiak and we were thinking that it may be best to keep this on a separate `hmc` branch in the main repo (rather than my forked copy), so that we can collectively contribute to the branch. 

@fritzo, @eb8680 - Let me know if there is anything I can address as part of this PR. Once I have addressed all comments, I will push to the `hmc` branch and close this PR. I will also create issues that will need to be addressed before this is release-ready.

neerajprad:
Also, let me know if you have any suggestions for test cases that will sufficiently exercise edge cases (probably having distributions with constrained support), which a minimum viable implementation should satisfy.

stepelu:
For regular HMC (i.e. not biased SGHMC variants), NUTS is a good automatically-tuned variant and the algorithm employed in Stan for MCMC.

In case it helps I implemented it in Lua at:
https://github.com/stepelu/lua-sci/blob/master/mcmc/_nuts.lua

It shouldn't be too difficult to port to Python, the tree is built recursively, and it supports diagonal and full-rank mass matrix estimation.
MIT licensed, feel free to ask if you have questions.

neerajprad:
> For regular HMC (i.e. not biased SGHMC variants), NUTS is a good automatically-tuned variant and the algorithm employed in Stan for MCMC.

> In case it helps I implemented it in Lua at:
https://github.com/stepelu/lua-sci/blob/master/mcmc/_nuts.lua

Thanks @stepelu. I think NUTS is definitely something we would like to have in Pyro. This is more to hash out our API, and think about how HMC integrates with Pyro internals. I will take a look at your implementation of NUTS as I have more time, and get back to you if I have any questions. Did you face any performance issues with your implementation (it may be amplified in python without JIT support)?

pyro/infer/mcmc/mcmc.py
```diff
@@ -0,0 +1,39 @@
...
+        trace = poutine.trace(self.model).get_trace(*args, **kwargs)
+        self.logger.info('Starting MCMC using kernel - {}'.format(self.kernel.__class__.__name__))
+        while self._t < self.warmup_steps + self.num_samples:
+            if self._t % 100 == 0:
+                self.logger.info('MCMC Iteration: {}'.format(self._t))
+            trace = self.kernel.sample(trace, self._t)
+            self._t += 1
+            if self._t < self.warmup_steps:
+                continue
+            yield (trace, trace.log_pdf())
```
eb8680: This should be `(trace, 1)` for MCMC - traces generated by MCMC are unweighted.

eb8680:
Mostly looks good to me, aside from a minor bug I mentioned.  We should also have a convenience class `pyro.infer.HMC` that just wraps MCMC and the HMC kernel.

pyro/infer/mcmc/trace_kernel.py
```diff
@@ -0,0 +1,25 @@
+from __future__ import absolute_import, division, print_function
+
+from abc import abstractmethod, ABCMeta
+
+from six import add_metaclass
+
+
+@add_metaclass(ABCMeta)
+class TraceKernel(object):
```
eb8680: Do you think `TraceKernel` should be a `Distribution`?

pyro/infer/mcmc/mcmc.py
neerajprad: You are right, this will return biased samples otherwise. Will fix.

pyro/infer/mcmc/trace_kernel.py
neerajprad: Hmm..that's an interesting thought. The semantics are somewhat different in that we cannot generate a sample directly, since it needs an existing sample.

neerajprad:
@eb8680 - Will address your comment, and make a few changes to the tests.

stepelu:
> Thanks @stepelu. I think NUTS is definitely something we would like to have in Pyro. This is more to hash out our API, and think about how HMC integrates with Pyro internals. I will take a look at your implementation of NUTS as I have more time, and get back to you if I have any questions. Did you face any performance issues with your implementation (it may be amplified in python without JIT support)?

@neerajprad: it's difficult for me to comment on how it would fare in python, but the LuaJIT performance was competitive with Stan for small dimensional parameter problems.
The limiting factor was the use of forward mode AD in SciLua instead of reverse mode, I never found the time to implement it.
Bonus point: I kept the code implementation very close to the algorithms as listed in the paper, so it should easy to follow.

When time comes it might be worth to check with Bob Carpenter of Stan about some changes they made to the NUTS algorithms after it got published.
I know they made some modifications to the adaptation procedure, which I don't replicate in my code, and possibly something else.

neerajprad:
@eb8680, @fritzo, @martinjankowiak : I have refactored/added some more tests for the HMC implementation, using a 5-10 dimensional gaussian chain model up to a chain length of 9. There are a few things that further need to be implemented for this to be usable in practice:
 - **Diagnostics on convergence** - Certain cases will yield really biased results without fine-tuning the value for the step size and number of steps (as well as the warmup steps).
 - **Faster implementation / profiling** - The current implementation is somewhat slow, and can take several minutes to get a few thousand samples from. I need to do some profiling to figure out the bottlenecks.

I have pushed this to the [hmc](https://github.com/uber/pyro/tree/hmc) branch of the repo, so feel free to push to it. I will keep this PR online for a few days until I have created separate issues for each of the tasks. I will also try to abstract out the interface and the integrator which I suppose can be merged into `dev` as we continue to refine the HMC implementation.

neerajprad:
> When time comes it might be worth to check with Bob Carpenter of Stan about some changes they made to the NUTS algorithms after it got published. I know they made some modifications to the adaptation procedure, which I don't replicate in my code, and possibly something else.

Thanks for the suggestions, @stepelu. There is much to learn from Stan's implementation (and refinement, as you mention) of the NUTS algorithm, and we will check with them once we get started.

neerajprad:
Closing this. All the code on this PR is in `hmc`. I will create separate follow-up issues, which can be addressed separately and merged into `dev`.

PR #581, opened 2017-11-16
Use sample statement instead of observe in tutorials and examples

### LINKED ISSUE #580: Use pyro.sample instead of pyro.observe in the tutorials and examples
We have had some users trip over the unexpected positional obs argument in `pyro.observe`. See #571 for an example. Since the observation shape in many cases is the same as the parameter shape, the model might actually run without warning but give very unexpected results. As a possible solution to this that was discussed in #571, we can encourage the use of `pyro.sample` (with the `obs` argument) in our tutorials and examples. Note that changing the function signature will break existing code.

neerajprad:
Fixes #580.

Note to tutorial authors - I have modified the tutorials, but if there is anything that can be worded better, please let me know and I will make the change. Alternatively, if you would prefer to make this change yourself, please feel free to do so.


neerajprad:
> Have you tested by running any of the tutorials? (it sure would be nice to have tutorial tests for PRs like this)

I ran the examples, but since most tutorial cells are not runnable, I couldn't test them. I agree that it will be nice to be able to run `nbval` on most of the cells. To maintain our tutorials, we definitely need those tests, but that will require refactoring the tutorials to broadly test for any changes in the API, without having to run all cells. Will create a separate task for this.

fritzo:
Does anyone object to merging this?

null-a:
FWIW, the changes to AIR are fine with me. Thanks for doing this!

neerajprad:
> observe is still in the documentation correct

It is there in the documentation, and we are using it internally in our tests.

martinjankowiak:
this is fine with me but i'm pretty sure there are a few more places in the text that need to be updated to reflect the change. let's not merge until i've had a chance to double check.

neerajprad:
> this is fine with me but i'm pretty sure there are a few more places in the text that need to be updated to reflect the change. let's not merge until i've had a chance to double check.

Thanks for taking a look; I will do so as well before we merge this.

tutorial/source/intro_part_ii.ipynb
```diff
@@ -247,8 +247,8 @@
     "def scale_obs(guess):\n",
```
eb8680: Should just delete this function now that it's identical to the example just above it

neerajprad: Good catch. Will remove this.

tutorial/source/dmm.ipynb
```diff
@@ -411,18 +415,18 @@
...
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Packaging the Model and Guide as a Pytorch Module\n",
+    "## Packaging the Model and Guide as a PyTorch Module\n",
     "\n",
-    "At this juncture, we're ready to to proceed to inference. But before we do so let's quickly go over how we packaged the model and guide as a single Pytorch Module. This is generally good practice, especially for larger models."
+    "At this juncture, we're ready to to proceed to inference. But before we do so let's quickly go over how we packaged the model and guide as a single PyTorch Module. This is generally good practice, especially for larger models."
```
neerajprad: Not introduced in this PR, but we have double tos `to to proceed`.

tutorial/source/intro_part_ii.ipynb
```diff
@@ -141,30 +115,9 @@
...
...lU6NGjB5YsWQJra2u5Y9VpwoQJePDgAVQqlf4Sv8YiJiYG6enpKCwshJOT\nE959910MGzYMs2fPRm5uLtq1a4c1a9agTZs2ckcFUHveNm3aYPHixSgoKICdnR169OiBrVu3yh21\n1qzx8fEoLy/Xb8/evXtj0aJFLz1Wky1+IiKqXZM91ENERLVj8RMRCYbFT0QkGBY/EZFgWPxERIJh\n8RO9oLy8PEycOLFBz/Xw8EBxcXGtj8XFxaG8vNyY0YgahJdzEpmQh4cHLl68iNatW7/QY0SmxD1+\nEtKePXuwcOFCAEBGRgY8PDz0t/T45JNPsHfvXly+fBkTJ05EWFgYwsLCcPz4cQBPb6rXv39//bKO\nHTuGkSNHIjQ0FBs3bqyxl79jxw6MHj0aQ4cOxbFjxwBAP/b48eOh0WiMd591ooYwys2diSzM7du3\npREjRkiSJEkbN26Uxo0bp7+nfEBAgHTlyhVJo9FIeXl5kiRJUl5enuTn5yc9fPhQunPnjtSvXz9J\nkiQpPz9f6tevn5SVlSVJkiR9+eWXUrdu3aSioiJJkiSpW7du0o4dOyRJkqTz589Lvr6++gy/fB6R\nOXGPn4TUsWNHlJWV4f79+0hLS8OcOXOQlpaG3NxcVFRUQKfTIScnB9OnT4dGo8H06dOhUCiQnZ1d\nbTmXL1/Gb37zG/2N1EaPHl1jrDfffBPA0/upa7XaRv8XwKjpU8kdgEguAwYMwLfffgudTof+/ftj\n8eLFOH78OPr37w9JkuDh4YFdu3bVmC8nJ+eFxmnevDkAQKlUAnh6G+6fpxHJgXv8JKwBAwZg8+bN\n8PT0BAD07dsXmzdvho+PDzw9PZGdnY1///vf+udnZGRAeuZaiN69e+Pq1av48ccfAQAHDx5s8Pit\nW7fW30GWyJxY/CSsAQMG4O7du/q7X/78/YABA2Bvb4+//vWvWL9+PUaNGoXAwECsW7euRvG3bdsW\nn3zyCaZPn47Q0FAUFBSgWbNmDbqFbmRkJCZNmsSTu2R2vJyT6CUVFRXBxsYGwNM/j7d//37s3r1b\n5lREdeMxfqKXtGPHDvzzn/9EVVUV7O3t8emnn8odiahe3OMnIhIMj/ETEQmGxU9EJBgWPxGRYFj8\nRESCYfETEQnm/wEMxSk3cXS/lQAAAABJRU5ErkJggg==\n",
```
neerajprad: This is removing the image. Is this intentional?

tutorial/source/intro_part_ii.ipynb
```diff
@@ -586,30 +426,9 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 16,
-   "metadata": {},
-   "outputs": [
```
neerajprad: Same here..the output is stripped.

martinjankowiak: nope none of that is intentional... i'm just entirely confused about nbstripout, how/when to keep cell output, etc

neerajprad: We can run `nbstripout --keep-output` to retain the output or `make scrub`. 

@jpchen - Do we need any metadata, like `collapsed` for rendering the tutorials or can we strip out the metadata field completely?

neerajprad: @martinjankowiak - for the time being though, you can revert these hunks manually using your git tool of choice :) (like SourceTree)

martinjankowiak: well is the previous commit ok? the only difference between the two commits is nbstripout

jpchen: no i hardcoded the outputs where need be. just make sure the intro tutorials have that as wel

tutorial/source/intro_part_ii.ipynb
jpchen: can someone make sure the diagrams in there match [the tutorial](http://pyro.ai/examples/intro_part_ii.html) ? i cant seem to pull this branch locally

tutorial/source/intro_part_ii.ipynb
neerajprad: @martinjankowiak - that looks fine except that we have some `collapsed: True` in the metadata.

@jpchen - So would I be correct in assuming that we do not need any metadata, but we should retain the output?

jpchen: yes

martinjankowiak: oh god i tried to revert to 'small text changes' and then do make scrub but it didn't come out right... : (

neerajprad: @martinjankowiak - Did `make scrub` not work? If all your changes are in and we have cruft like `execution count` etc. metadata, I can strip it out.

martinjankowiak: well i think i got rid of all the text edits i had made in the first commit...

martinjankowiak:
ok now it's hopefully ok...

neerajprad:
Looks good!

PR #582, opened 2017-11-16
typo in bayesian_regression.ipynb

manuelhaussmann:
Missing brackets in print statement

### DIFF
--- a/tutorial/source/bayesian_regression.ipynb
+++ b/tutorial/source/bayesian_regression.ipynb
@@ -391,7 +391,7 @@
     "    y_preds = y_preds + sampled_reg_model(x_data)\n",
     "# take the average of the predictions\n",
     "y_preds = y_preds / 20\n",
-    "print \"Loss: \", loss(y_preds, y_data).data[0]"
+    "print (\"Loss: \", loss(y_preds, y_data).data[0])"
    ]
   },
   {


jpchen:
this notebook is built with python2.7 which is why there are no parens. to make this runnable, there would need to be `from __future__ import print_function` and all the other print statements would need to be changed, otherwise we need to reubuild the notebooks with python3. currently this wont be able to run. would you mind if i closed this and opened an issue for py3 support?

manuelhaussmann:
Sure, go ahead!

However are you sure that the notebook is built with python2.7? The other print statements already follow python3 style and the notebook ran without problems (except for that line) when I worked through the example earlier today.

fritzo:
> to make this runnable, there would need to be `from __future__ import print_function`

`print("foo")` runs fine in Python 2. It means the same as `print ("foo")` which means the same as `print "foo"`.

manuelhaussmann:
@fritzo Thanks for the clarification. It's been a while that I looked at python2.7

jpchen:
ah @fritzo is right havent used that syntax. 

@manuelhaussmann do you want to reopen this PR? this is a change worth making so it will run for people using python 3

manuelhaussmann:
@jpchen I have reopened it.

PR #589, opened 2017-11-20
Update to AIR example/tutorial

null-a:
This PR is ready for review, but I think it will conflict with #581, so it probably makes sense to merge that before this.

Summary: The important changes here are:

* Update the tutorial to describe the `log_pdf_mask` stuff.
* Update the full implementation to look more like the thing implemented in the tutorial.

There are a few more minor changes, hopefully the individual commits should be somewhat self explanatory.

I'll open another PR to add some results to the tutorial soon after this is merged.


examples/air/air.py
```diff
@@ -99,15 +109,6 @@ def __init__(self,
             self.cuda()
 
     def prior(self, n, **kwargs):
```
martinjankowiak: i personally find the language choices of prior/model/model_step a bit confusing... also, the order in which the functions appear in the file. can we choose better names?

examples/air/air.py
```diff
@@ -118,13 +119,25 @@ def local_model(self, n, batch=None, **kwargs):
...
+            pyro.observe('obs',
+                         dist.normal,
+                         batch.view(n, -1),
+                         x.view(n, -1),
+                         (self.likelihood_sd * self.ng_ones(1)).expand(n, self.x_size ** 2))
+
+    def model_step(self, t, n, prev, z_pres_prior_p=default_z_pres_prior_p):
 
         # Sample presence indicators.
         if not self.fudge_z_pres:
```
martinjankowiak: some of the trickier things could perhaps use more comments? e.g. fudge_z_pres

examples/air/air.py
null-a: > can we choose better names?

We can try! I'm so familiar with them that I can't see the problem. What names would you have found less confusing?

> also, the order in which the functions appear in the file

Sure. Let me know your preferred order and I'll adjust them. (In a follow up PR.)

examples/air/air.py
null-a: I don't think it's worth keeping `fudge_z_pres` around, so I'll remove that in a follow up PR. I'll check for similar things at the same time, but let me know if there's anything else that needs specific attention.

null-a:
I've fixed the merge conflict. If @martinjankowiak is happy for me to address his comments in follow up PRs, I think this can be merged.

Issue #597, opened 2017-11-21
Linear regression example doesn't run with p > 1


stepelu:
The linear regression example at:
https://github.com/uber/pyro/blob/dev/examples/bayesian_regression.py
doesn't run outside of the 1 dimensional case.
Fixing the data generation part is not enough, an error occur later at `lifted_reg_model(x_data)`.

jpchen:
Closed via #598

PR #598, opened 2017-11-21
(Re)supporting higher dimensional regression

jpchen:
Addresses #597

somewhere between changing the dataset in the anchor model and the one in the tutorial, i broke the functionality of supporting multiple dimensions. This fixes that. 

tutorial does not require the fix, but will change there to make consistent with example

### DIFF
--- a/examples/bayesian_regression.py
+++ b/examples/bayesian_regression.py
@@ -102,12 +102,7 @@ def get_batch_indices(N, batch_size):
     return all_batches
 
 
-def main():
-    parser = argparse.ArgumentParser(description="parse args")
-    parser.add_argument('-n', '--num-epochs', default=1000, type=int)
-    parser.add_argument('-b', '--batch-size', default=N, type=int)
-    parser.add_argument('--cuda', action='store_true')
-    args = parser.parse_args()
+def main(args):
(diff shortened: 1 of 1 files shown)


examples/bayesian_regression.py
```diff
@@ -19,11 +19,13 @@
...
 
 # generate toy dataset
-def build_linear_dataset(N, p, noise_std=0.1):
-    X = np.linspace(-6, 6, num=N)
-    y = 3 * X + 1 + np.random.normal(0, noise_std, size=N)
-    X = X.reshape((N, p))
-    y = y.reshape((N, 1))
+def build_linear_dataset(N, p, noise_std=0.01):
+    X = np.random.rand(N, p)
+    # use random weights from [0, 8]
```
fritzo: nit: `np.random.randint(0, 8)` actually draws from `[0,8) = [0,7]`.

jpchen: makes sense, i picked a random int. ill change the comment

jpchen:
you mean a cla? you know the `p` and `N` a priori which is why it's global. if the toy dataset was swapped out for a real dataset those values would be hardcoded similarly in the code. does it make sense to toggle this value as a cla?

PR #599, opened 2017-11-22
Minor tweaks to the AIR example

null-a:
This PR:

1. Removes the option to `fudge_z_pres`, as promised in #589
2. Adds a CLA for learning rates.

### DIFF
--- a/examples/air/air.py
+++ b/examples/air/air.py
@@ -16,7 +16,6 @@ from torch.autograd import Variable
 import pyro
 import pyro.distributions as dist
 from modules import Identity, Encoder, Decoder, MLP, Predict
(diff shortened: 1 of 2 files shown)


null-a:
CI failed with:

`"The job exceeded the maximum time limit for jobs, and has been terminated."`

> maybe model_step to prior_step?

Oh, gotcha, I think I see the confusion now. If it's OK with you I'll submit a separate PR for this. I'll need to make the same change to the tutorial, and it would be good if we could merge this before I get round to that.

martinjankowiak:
@null-a sure. i'll merge now since that timeout is irrelevant

PR #604, opened 2017-11-24
Add Binomial distribution

fehiepsi:
Based on Multinomial distribution code, I implement Binomial distribution. I also change the code of Multinomial distribution a little bit to fix bugs when parameter `n` is an `int` (e.g. in `expanded_sample()` function, we can not get `self.n.data.dim()`.


fehiepsi:
My mistake, I delete the wrong branch.

fehiepsi:
It seems that travis failed because of exceeding time.

pyro/distributions/multinomial.py
```diff
@@ -25,6 +25,11 @@ class Multinomial(Distribution):
     def __init__(self, ps, n, batch_size=None, *args, **kwargs):
         if ps.dim() not in (1, 2):
             raise ValueError("Parameter `ps` must be either 1 or 2 dimensional.")
+        if isinstance(n, int):
```
fritzo: Prefer `isinstance(n, (int, long))` or `isinstance(n, numbers.Number)`.

fehiepsi: Thanks! I prefer your second suggestion. In python 3, there is no `long` type.

PR #605, opened 2017-11-28
Add helper pyro.distributions.util.broadcast_shape()

fritzo:
This is useful for testing and simplifying broadcasting logic for batching tensors. Note that `np.broadcast()` does not work for torch variables on older PyTorch (e.g. 0.2 release). This prints the same error message as `np.broadcast()` when broadcasting fails.

## Tested

Added a unit test to demonstrate agreement with `np.broadcast()`.

### DIFF
--- a/pyro/distributions/util.py
+++ b/pyro/distributions/util.py
@@ -5,6 +5,28 @@ import torch.nn.functional as F
 from torch.autograd import Variable
 
 
+def broadcast_shape(*shapes):
+    """
+    Similar to ``np.broadcast()`` but for shapes.
+    Equivalent to ``np.broadcast(*map(np.empty, shapes)).shape``.
+
+    :param tuple shapes: shapes of tensors.
+    :returns: broadcasted shape
+    :rtype: tuple
+    :raises: ValueError
+    """
+    reversed_shape = []
+    for shape in shapes:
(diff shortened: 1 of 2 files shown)


pyro/distributions/util.py
```diff
@@ -5,6 +5,28 @@
...
+    """
+    reversed_shape = []
+    for shape in shapes:
+        for i, size in enumerate(reversed(shape)):
+            if i >= len(reversed_shape):
+                reversed_shape.append(size)
+            elif reversed_shape[i] == 1:
+                reversed_shape[i] = size
+            elif size != 1 and reversed_shape[i] != size:
+                raise ValueError('shape mismatch: objects cannot be broadcast to a single shape')
```
neerajprad: Do you think it will be useful to have the tensor sizes in the error statement when broadcasting fails?

fritzo: Maybe. This version prints the `np.broadcast()` error message verbatim. I can change it if you feel strongly.

neerajprad: Lets leave it as is for now. We can put in a more verbose message if we need it later.

neerajprad:
Should I merge this? We shouldn't need the integration test runs for this.

fritzo:
Ready to merge. Thanks for review!

Issue #606, opened 2017-11-28
Wrap torch.distributions.* for use in pyro


fritzo:
While we work on converging pyro.distributions and torch.distributions, we will need a compatibility layer so that torch.distributions can be used inside Pyro. Ideally this layer will become thinner as we move functionality upstream. As torch.distributions matures, we will remove Pyro's older scipy-based implementations.

- [x] #607 Normal
- [x] #630 Gamma
- [x] #630 Beta
- [x] #630 Dirichlet
- [x] #643 Exponential
- [x] #645 Bernoulli
- [x] #649 Categorical
- [x] #649 OneHotCategorical
- [x] Cauchy
- [x] Uniform
- [x] Multinomial
- [ ] HalfCauchy
- [x] Poisson
- [x] LogNormal
- [x] Binomial
- [ ] Laplace
- [ ] Chi2
- [ ] MultivariateNormal
- [ ] Delta

ragulpr:
Sorry if this is a stupid question, but what will the convergence look like? Are you guys working on porting pyro distributions to torch too? As a small example, I would really prefer pyro-style naming like `mu`, `sigma` to the `mean`, `std` naming style in `torch.distributions.normal` and I'm thinking about submitting a pull request. This also goes with log-likelihood functions/losses in torch like `NLLloss` (bernoulli/cross entropy) etc.

Anyone working on this?

fritzo:
Great question! We and the [Prob Torch](https://github.com/probtorch/probtorch) team are actively moving our distributions code upstream to PyTorch. The converged version will aim to maximize compatibility with TensorFlow Distributions (see their excellent [paper](https://arxiv.org/pdf/1711.10604.pdf)). We're actually leaning towards non-greek names (e.g. loc, scale), but we'd appreciate input if you have good reasons for keeping some names greek.

ragulpr:
I discovered Tensorflow Distributions this week and in trying to figure out what you guys where doing it really opened up a can of worm of awesomeness finding Pyro, Edward etc. Great work!

I think naming is the main reason why statistics/probability is hard so I think it's an extremely important question, not to least because this is going to have big effects on future confusions, research etc. I think we should tap into the infinite wisdom of past practices which it seems from the paper like people are doing so I'm happy you guys are working together. 

[...]

Issue #414, opened 2017-10-30
Inconsistent loss values with & without vectorization


hawkrobe:
Consider the [vectorized](https://github.com/hawkrobe/hbnn/blob/master/models/regression/hierarchical_vectorized.py) and [non-vectorized](https://github.com/hawkrobe/hbnn/blob/master/models/regression/hierarchical_direct.py) versions of the same hierarchical regression model. The main differences are that the vectorized version:

* uses `iarange` instead of `irange` 
* uses a single `observe` from a `1 x batch_size` dimensional distribution instead of `batch_size` separate observes from 1 dimensional distributions
* optimizes `1 x k` dimensional params for a single `intercepts` sample site instead of k separate 1 dimensional params for k sample sites

Intuitively, I'd expect these to converge to the same loss; instead the vectorized version converges to ~1900 and the non-vectorized version converges to ~600. [This same model written in webppl](https://github.com/hawkrobe/hbnn/blob/master/models/webppl-comparison/hierarchical-regression.wppl) also converges to about ~600, so this might indicate an issue with scaling in the vectorized version?

Incidentally, the mean-field guide `sigma`s also converge to different values in the vectorized version although the `mu` point estimates are the same. The unvectorized version matches webppl but the vectorized version has much higher certainty. 

It's of course very plausible that there's a bug in my implementation of the vectorized model!

PR #607, opened 2017-11-28
Wrap torch.distributions.Normal for use in Pyro

fritzo:
Addresses #606 

This creates an optional wrapper to use `torch.distributions.Normal` in Pyro. The torch version is only used if all of the following are satisfied:
- The environment variable `PYRO_USE_TORCH_DISTRIBUTIONS=1` is set
- The `torch.distributions` module exists (it is missing in PyTorch 0.2 release)
- The `torch.distributions.Normal` class exists
- All requested features are available (e.g. `torch.distributions.Normal` is reparameterized if `reparameterized=True`, also `log_pdf_mask` is not supported).

If any of the previous conditions are not satisfied, Pyro falls back to the standard implementation.

## Tested

The torch distribution will not be exercised on travis. Tests pass locally except for an [unrelated bug](https://github.com/pytorch/pytorch/issues/3914) in PyTorch master 0.4.0a0+709fcfd.

### DIFF
--- a/pyro/distributions/normal.py
+++ b/pyro/distributions/normal.py
@@ -1,10 +1,13 @@
 from __future__ import absolute_import, division, print_function
 
+import warnings
+
 import numpy as np
 import torch
 from torch.autograd import Variable
 
 from pyro.distributions.distribution import Distribution
+from pyro.distributions.util import broadcast_shape, torch_wrapper
 
 
 class Normal(Distribution):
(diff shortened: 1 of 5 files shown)


pyro/distributions/normal.py
```diff
@@ -105,3 +108,61 @@ def analytic_var(self):
...
+    enumerable = False
+
+    def __init__(self, mu, sigma, *args, **kwargs):
+        self._torch_dist = torch.distributions.Normal(mean=mu, std=sigma)
+        self._param_shape = torch.Size(broadcast_shape(mu.size(), sigma.size(), strict=True))
+        super(TorchNormal, self).__init__(*args, **kwargs)
+
+    def batch_shape(self, x=None):
+        x_shape = [] if x is None else x.size()
+        shape = torch.Size(broadcast_shape(x_shape, self._param_shape, strict=True))
```
neerajprad: This might lead to some hard to find bugs like #414 if the event dimensions do not match between the data and the parameters. Should we allow this kind of broadcasting, or limit it to the batch dimensions only (i.e. any `x`'s rightmost sizes must exactly agree with `sample_shape + event_shape`)?

fritzo: Right, that was actually needed to pass some of the expected-error tests.

pyro/distributions/util.py
```diff
@@ -1,29 +1,48 @@
...
     Similar to ``np.broadcast()`` but for shapes.
     Equivalent to ``np.broadcast(*map(np.empty, shapes)).shape``.
 
     :param tuple shapes: shapes of tensors.
+    :param bool strict: whether to use extend-but-not-resize broadcasting.
     :returns: broadcasted shape
     :rtype: tuple
     :raises: ValueError
     """
+    strict = kwargs.pop('strict', False)
```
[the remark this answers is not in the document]
fritzo: Good idea, will do.

Issue #20, opened 2017-06-27
Visualization in Pyro


eb8680:
Visualization is critical for understanding and debugging models and inference algorithms.  What are our visualization desiderata and what features are necessary to support them?  

It seems to me that eventually, we'll need support for automatic visualization of samples, scores, and summary statistics and their evolution in execution traces, trace and autograd graph structures, losses and individual loss terms and their evolution in inference algorithms, NaN detection, and parameter/parameter gradient summary statistics (variance, norm, etc).  We may also want custom visualizations for specific primitive distributions, and for our model criticism stuff.

Given that we've settled on visdom as our primary visualization tool, what Pyro-specific things do we need to implement, and what sort of interface(s) should we provide?  Should we evaluate other tools as well?

Some references:
- [webppl-viz](https://github.com/probmods/webppl-viz)
- [visdom](https://github.com/facebookresearch/visdom)
- [TensorBoard](https://github.com/tensorflow/tensorflow/blob/r1.2/tensorflow/tensorboard/README.md)
- [TensorBoard and Edward](http://edwardlib.org/tutorials/tensorboard)
- TensorBoard decoupled from the rest of TF: [crayon](https://github.com/torrvision/crayon) and [dmlc's fork](https://github.com/dmlc/tensorboard)
- [TensorBoard for PyTorch](https://github.com/lanpa/tensorboard-pytorch)
- [Visualization in PyMC](http://pymc-devs.github.io/pymc3/api/plots.html)

eb8680:
Closing in favor of more targeted discussion in future

PR #609, opened 2017-11-29
Add default logging support

### LINKED ISSUE #608: Add logging support to Pyro
Using the `print` function is not ideal for multiple reasons:
 - It does not give us enough control and granularity over logging. A mature model will need less logging, whereas we need more of it while debugging or testing a model. Python's logging library provides us this functionality, as we can configure the logging level.
 - The users of Pyro may want to configure their own handler (e.g. email the user if the model throws an error), and we currently do not have a way of doing that.

Right now, we are mostly using the `print` method in the tests. However, since MCMC algorithms will be outputting verbose diagnostic messages or warnings, we should have a logger handle that rather than printing it.

neerajprad:
Fixes #608. Refer to the task for more details. This is needed so that MCMC algorithms can output diagnostic messages to the logger.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -6,7 +6,7 @@ logger = logging.getLogger(__name__)
 
 # create log handler for tests
 handler = logging.StreamHandler()
-formatter = logging.Formatter('%(levelname).1s %(name)s \n %(message)s')
+formatter = logging.Formatter('%(levelname).1s %(name)s \t %(message)s')
 handler.setFormatter(formatter)
 
 # set default logging level for tests
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -6,7 +6,7 @@ logger = logging.getLogger(__name__)
 
 # create log handler for tests
 handler = logging.StreamHandler()
-formatter = logging.Formatter('%(levelname)s - %(message)s')
+formatter = logging.Formatter('%(levelname).1s %(name)s \n %(message)s')
 handler.setFormatter(formatter)
 
 # set default logging level for tests
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,1 +1,14 @@
 from __future__ import absolute_import, division, print_function
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+# create log handler for tests
+handler = logging.StreamHandler()
+formatter = logging.Formatter('%(levelname)s - %(message)s')
+handler.setFormatter(formatter)
+
+# set default logging level for tests
+logger.addHandler(handler)
+logger.setLevel(logging.DEBUG)
--- a/tests/infer/test_gradient.py
+++ b/tests/infer/test_gradient.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
+import logging
+
 import numpy as np
 import pytest
 import torch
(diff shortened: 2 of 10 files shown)


tests/__init__.py
```diff
@@ -1 +1,14 @@
 from __future__ import absolute_import, division, print_function
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+# create log handler for tests
+handler = logging.StreamHandler()
+formatter = logging.Formatter('%(levelname)s - %(message)s')
```
fritzo: I'm fond of `"%(levelname).1s %(name)s %(message)s"` if it's all the same to you.

tests/poutine/test_mapdata.py
```diff
@@ -293,39 +285,45 @@ def irange_cuda_model(subsample_size):
...
     assert tr.batch_log_pdf().is_cuda
 
 
-@pytest.mark.parametrize('model', [
-    iarange_model,
-    irange_model,
-    map_data_vector_model,
-    map_data_iter_model,
-], ids=['iarange', 'irange', 'map_data_vector', 'map_data_iter'])
+@pytest.mark.parametrize(
```
fritzo: Please either revert the autoformat changes or split them into a separate PR.

neerajprad: My bad, this should have been two different commits. Will revert and push again.

tests/__init__.py
neerajprad: Done.

neerajprad: Btw..this is for the root logger for the test. The test output will look as follows (I added a newline before message). Seems a little cluttered to me. What do you think?


```
I tests.nn.test_clipped_nn 
 softmax_ps = Variable containing:
 0.2689  0.7311
[torch.FloatTensor of size 1x2]

.I tests.nn.test_clipped_nn 
 epsilon = 1e-05
I tests.nn.test_clipped_nn 
 softmax_ps = Variable containing:
 0.2689  0.7311
[torch.DoubleTensor of size 1x2]
```

fritzo: It would be cleaner to log `softmax_ps.data.cpu().numpy()`, but that seems to be an issue for the user rather than the log handler.

neerajprad: That was just a random sample that I picked. Changing this to a tab instead of a newline.

eb8680:
Nice!  Also a prerequisite to integrating ideas from #20.

neerajprad:
> this is great and definitely needed, esp as the system gets more complex.
all the assert messages in the code probably should be changed to caught exceptions with a log.error message

We should still raise an exception if we have entered into a completely unexpected state so that we terminate early and throw a meaningful exception, but we can definitely supplement it with logging to output other information that may be useful for debugging.

tests/infer/test_gradient.py
```diff
@@ -1,5 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
+import logging
```
fritzo: I'm used to run `pytest -s` to see this test output. How can I invoke `pytest` after this PR to see the logging output? Is it something like 'PYRO_LOG_LEVEL=1 pytest`?

neerajprad: The test root logger has the default level set to `DEBUG` (we can change this later depending on some environment flag) so you shouldn't need to do anything else. `pytest -s` should work fine.

jpchen:
> We should still raise an exception if we have entered into a completely unexpected state 
Yes sorry I meant replace/add the message not the assert.  Basically exceptions should be greppable in the logs since when you run things on remote machines a lot of times you don't have easy access to the output history

neerajprad:
This is good to merge now.

PR #610, opened 2017-11-29
Add utility classes for MCMC algorithms

neerajprad:
Abstracts the interface for the `MCMC` and `TraceKernel` classes from #579. I have made a few tweaks to make the wrapper class more general, and added to the documentation. This is needed so that @OptimusLime can rebase off of this, rather than working off of the `hmc` branch.

@OptimusLime - Let me know if this is general enough for your use case.

NOTE - Blocked on  #609.


docs/source/mcmc.rst
```diff
@@ -0,0 +1,9 @@
+MCMC
+====
+
+.. automodule:: pyro.infer.mcmc
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+.. include:: pyro.infer.mcmc.txt
```
neerajprad: We can add specific algorithms here.

pyro/infer/mcmc/trace_kernel.py
```diff
@@ -0,0 +1,48 @@
...
+        pass
+
+    def cleanup(self):
+        """
+        Optional method to clean up any residual state on termination.
+        """
+        pass
+
+    def diagnostics(self, time_step):
+        """
```
neerajprad: Removed `num_accepts` from #579 in favor of a general diagnostics method so that the implementing classes are free to provide whatever diagnostic information may be useful.

pyro/infer/mcmc/mcmc.py
```diff
@@ -0,0 +1,56 @@
...
+        trace = poutine.trace(self.model).get_trace(*args, **kwargs)
+        self.logger.info('Starting MCMC using kernel - {} ...'.format(self.kernel.__class__.__name__))
+        logging_interval = int((self.warmup_steps + self.num_samples) / 20)
+        while self._t < self.warmup_steps + self.num_samples:
+            if self._t % logging_interval == 0:
+                self.logger.info('MCMC iteration: {}.'.format(self._t))
+                if self.kernel.diagnostics is not None:
+                    self.logger.info(self.kernel.diagnostics(self._t))
+            trace = self.kernel.sample(trace, self._t, *args, **kwargs)
+            self._t += 1
```
eb8680: Should `TraceKernel` be tracking timesteps internally, or should it use the same counter as the enclosing algorithm?

pyro/infer/mcmc/trace_kernel.py
```diff
@@ -0,0 +1,48 @@
...
+        of the MCMC run. Returns `None` by default.
+
+        :param time_step: Current Monte Carlo time-step.
+        :return: String containing the diagnostic summary. e.g. acceptance rate
+        :rtype: string
+        """
+        return None
+
+    @abstractmethod
+    def sample(self, trace, time_step, *args, **kwargs):
```
eb8680: Should we add a `__call__` alias for `sample` for consistency with other stochastic functions?

fritzo: Do you have a reason for hard-coding `time_step` rather than subsuming it in `*args, **kwargs`?

neerajprad: The only class that currently calls this is the wrapper `MCMC` class which passes this while calling the kernel's sample method (though it is arguable whether we need to do this or have the kernel class track this as part of its state as @eb8680 mentioned). I am thinking of removing this and have the kernel instance track it if it needs to. Then we can remove it from the method signature as well.

pyro/infer/mcmc/mcmc.py
```diff
@@ -0,0 +1,56 @@
...
+    :param int num_samples: The number of samples that need to be generated,
+        excluding the samples discarded during the warmup phase.
+    :param int warmup_steps: Number of warmup iterations. The samples generated
+        during the warmup phase are discarded.
+    :param \*args: Positional arguments for constructing the ``kernel`` instance.
+    :param \*\*kwargs: Arbitrary keyword arguments for constructing the ``kernel``
+        instance.
+    """
+
+    def __init__(self, model, kernel, warmup_steps, num_samples, *args, **kwargs):
```
fritzo: I would recommend putting `num_samples` first, defaulting `warmum_steps=0`. (In the future we might also add a `take_every=1` argument to allow taking only every few samples.) I would also recommend passing in a constructed kernel rather than a cumbersome kernel factory + `*args, **kwargs` (which will inevitably lead to headaches as we change the `MCMC` interface).

pyro/infer/mcmc/mcmc.py
neerajprad: Since the TraceKernel class holds state anyways, I think it will be better if it tracks time-steps internally. That will allow for a clearer method signature too.

pyro/infer/mcmc/mcmc.py
neerajprad: Makes sense, will address.

pyro/infer/mcmc/mcmc.py
```diff
@@ -0,0 +1,55 @@
...
+        given an execution trace returns another sample trace from the target
+        (posterior) distribution.
+    :param int num_samples: The number of samples that need to be generated,
+        excluding the samples discarded during the warmup phase.
+    :param int warmup_steps: Number of warmup iterations. The samples generated
+        during the warmup phase are discarded.
+    """
+
+    def __init__(self, kernel, num_samples, warmup_steps=0):
+        self.kernel = kernel
```
neerajprad: Removed model from the wrapper class.

pyro/infer/mcmc/trace_kernel.py
```diff
@@ -0,0 +1,61 @@
...
+    def initial_trace(self):
+        """
+        Returns an initial trace from the prior to initiate the MCMC run.
+
+        :return: Trace instance.
+        """
+        return NotImplementedError
+
+    @abstractmethod
+    def sample(self, trace):
```
neerajprad: @eb8680, @fritzo - I simplified the method signature for `TraceKernel` and removed the args, kwargs arguments. Since we have a setup method, it should be able to take care of any other arguments that need to be provided to the TraceKernel instance to set up its state.

neerajprad:
@eb8680, @fritzo: Thanks for the review! I have addressed all your comments. I added a small test which really helped clarify some subtle issues with the existing code which are fixed now. I have left comments in the code.

 - Simplified the interfaces so that we do not need to pass the model to both the `TraceKernel` class as well as the `MCMC` base class. Added an `initial_trace` method to `TraceKernel` which is responsible for initializing the chain.
 - The `sample` method only needs to pass the current trace and all the state setup can be done by the `setup` method of the TraceKernel class.

Let me know if you have any further suggestions.

Note - We cannot merge this until #609 is merged.

neerajprad:
> LGTM for now - I suspect things will change (like the nature of the diagnostic information and logging) as we add more MCMC algorithms. We may also want abstract kernel combinators for things like cycle and mixture kernels as in webPPL, but I guess those can be added later with MH.

Thanks. I will read up more on cycle and mixture kernels in webPPL!

tests/infer/mcmc/test_mcmc.py
```diff
@@ -0,0 +1,59 @@
...
+    def setup(self, data):
+        self.data = data
+
+    def cleanup(self):
+        self.data = None
+
+    def initial_trace(self):
+        return poutine.trace(self.model).get_trace(self.data)
+
+    def sample(self, trace):
```
fritzo: Nice! This is exactly the example of stateless kernel I was thinking of when the `time_step` arg seemed weird.

pyro/infer/mcmc/trace_kernel.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        Samples a trace from the approximate posterior distribution, when given an existing trace.
+
+        :param trace: Current execution trace.
+        :param int time_step: Current time step.
+        :return: New trace sampled from the approximate posterior distribution.
+        """
+        raise NotImplementedError
+
+    def __call__(self, trace):
```
fritzo: Just curious, do you have a reason for spelling this out rather than a one-liner?
```py
__call__ = sample
```

neerajprad: Not really. Old habits I guess. :) Will change.

neerajprad: Actually, there is an interesting issue that I encountered. If we do this, `__call__` will become an `abstractmethod` too and will need to be explicitly implemented by the sub class. Which makes sense in a way, but makes the abstract class functionality seem like a hack.

```
Can't instantiate abstract class PriorKernel with abstract methods __call__
```

fritzo: Ah good to know, thanks for explaining.

neerajprad:
@eb8680, @fritzo - This is good to merge, pending any further suggestions.

fritzo:
I'll merge if you change the title to not say "Do not merge". (I accidentally merged a PR recently.)

neerajprad:
> I'll merge if you change the title to not say "Do not merge". (I accidentally merged a PR recently.)

Done! :)

jpchen:
hm the commit message in dev still says [Do Not Merge]. im guessing the title changes but the auto-generated commit message does not.  we should be careful about changing the squashed commit message in the future, or create a Do Not Merge label and use that to track status of PRs

neerajprad:
> hm the commit message in dev still says [Do Not Merge]. im guessing the title changes but the auto-generated commit message does not. 

That's rather unexpected behavior, and I think I saw this once before. Thanks for the heads up. I will use the label instead from next time.

PR #611, opened 2017-11-29
Adding the verlet integrator used in HMC

neerajprad:
This adds the velocity verlet integrator from #579, as a utility function in `pyro.optim`.
 - The integrator has been majorly simplified. It uses the `torch.autograd.grad` method that @martinjankowiak pointed me to, and does not store any gradients on the parameters.
 - Modified the integrator to directly take in a potential function (instead of a callable that computed and returned the gradients), so that the integrator can call `torch.autograd.grad` internally.
 - @fritzo helped point out an issue with the previous implementation which was unnecessarily holding the intermediate integration steps in the computation graph and calling backwards on this. This is fixed.
 - Modified the tests to specify the potential energy function (as per the new method signature) instead of having to specify the function to compute gradients.
 - I have put this in the `pyro.optim` module, but we can also move this to `utils` or `math.utils`, since this will only be internally used.

NOTE - Once #609 is merged, I will change the base branch to run the tests.


pyro/optim/integrator.py
```diff
@@ -0,0 +1,46 @@
...
+        (type ``torch.autograd.Variable``).
+    :param callable potential_fn: function that returns potential energy given z
+        for each sample site. The negative gradient of the function with respect
+        to ``z`` determines the rate of change of the corresponding sites'
+        momenta ``r``.
+    :param float step_size: step size for each time step iteration.
+    :param int num_steps: number of discrete time steps over which to integrate.
+    :return tuple (z_next, r_next): final position and momenta, having same types as (z, r).
+    """
+    z_next = {key: val.clone().data for key, val in z.items()}
```
martinjankowiak: probably doesn't matter much but => val.data.clone()?

pyro/optim/integrator.py
```diff
@@ -0,0 +1,46 @@
...
+            r_next[site_name] = r_next[site_name] + 0.5 * step_size * (-grads[site_name])
+    z_next = {key: Variable(val) for key, val in z_next.items()}
+    r_next = {key: Variable(val) for key, val in r_next.items()}
+    return z_next, r_next
+
+
+def _grad(potential_fn, z):
+    z = {k: Variable(v, requires_grad=True) for k, v in z.items()}
+    z_keys, z_nodes = zip(*z.items())
+    grads = grad(potential_fn(z), z_nodes, create_graph=True)
```
martinjankowiak: do you need create_graph=True?

eb8680:
> lgtm, but i'd probably move out of optim

@martinjankowiak what do you think about implementing the integrator as a `torch.optim` optimizer?

martinjankowiak:
well i think an optimizer is something that consumes gradients and traverses down a loss surface while an integrator like this is something that consumes gradients to move along a vector field so i don't see any reason to mix the two

eb8680:
> well i think an optimizer is something that consumes gradients and traverses down a loss surface while an integrator like this is something that consumes gradients to move along a vector field so i don't see any reason to mix the two

Fair enough, although we may want to revisit this when we add more gradient-based MCMC algorithms.

pyro/optim/integrator.py
neerajprad: Thanks for spotting this. We do not need this now after the refactoring, will remove.

neerajprad:
@martinjankowiak, @eb8680 - Thanks for reviewing. I will address all your comments shortly, and this should be good to merge then.

PR #614, opened 2017-11-30
[Reinstating #611] Verlet integrator for HMC

neerajprad:
Refer to #611 for details. Since the base branch for that was `logging` (as it depended on #609), it was merged into the `logging` branch instead of `dev`. This is the exact PR with the `dev` branch as base.


neerajprad:
Good to merge once tests pass.

PR #615, opened 2017-12-01
reinstate inclined plane example

martinjankowiak:
everyone's favorite example is being reinstated. now with importance sampling. 

(the VI version was finicky and i can't be bothered to figure out how to improve it).

if the PR is accepted i think we should add a link to the example in the 2nd intro tutorial.


examples/inclined_plane.py
```diff
@@ -0,0 +1,130 @@
...
+
+    for i, T_obs in enumerate(observed_data):
+        observe_T(T_obs, "obs_%d" % i)
+
+    return mu
+
+
+def main():
+    parser = argparse.ArgumentParser(description="parse args")
+    parser.add_argument('-n', '--num-samples', default=500, type=int)
```
fritzo: Could you please add a case in `tests/test_examples.py` so that this example will run during testing:
```diff
  def discover_examples():
      for root, dirs, files in os.walk(EXAMPLES_DIR):
          for basename in files:
              ...
              if '--num-epochs' in text:
                  args = ['--num-epochs=1']
              elif '--num-steps' in text:
                  args = ['--num-steps=1']
+             elif '--num-samples' in text:
+                 args = ['--num-samples=1']
              else:
                  # Either this is not a main file, or we don't know how to run it cheaply.
                  continue
              ...
```

examples/inclined_plane.py
```diff
@@ -0,0 +1,130 @@
...
+           "for the inferred (mean) mu as: %.4f seconds") % analytic_T(posterior_mean.data[0]))
+
+    """
+    ################## EXERCISE ###################
+    # vectorize the computations in this example! #
+    ###############################################
+    """
+
+
+if __name__ == '__main__':
```
fritzo: Consider moving arg parsing here and passing `args` to `main(args)`. That would help focus attention on the math in `main()`.

examples/inclined_plane.py
martinjankowiak: thanks, done

examples/inclined_plane.py
martinjankowiak: thanks, done

PR #618, opened 2017-12-10
Expose examples as a package pyro_examples

fritzo:
## Why?

Users often start playing with Pyro by forking pyro examples and modifying the code. Currently the only ways to do this are to (1) edit examples in a branch of Pyro, or (2) copy the `examples/utils.py` file into another directory.

## How?

This PR exposes a `pyro_examples` package so that `pyro_examples.utils` can be imported in examples that have been forked out of the Pyro repo.

Is there a better way? Probably. E.g. we could move examples/utils.py into the main Pyro package.


fritzo:
Closing in favor of https://github.com/uber/pyro/issues/621

PR #622, opened 2017-12-12
Single-site Metropolis-Hastings

OptimusLime:
I've added a simple Metropolis-Hastings implementation, and I'd like to merge in after feedback/review.

Currently passing 1 test, but I'm open to suggestions on others.

### DIFF
--- a/pyro/infer/mcmc/mcmc.py
+++ b/pyro/infer/mcmc/mcmc.py
@@ -48,8 +48,8 @@ class MCMC(TracePosterior):
                 if diagnostic_info is not None:
                     self.logger.info(diagnostic_info)
             trace = self.kernel.sample(trace)
+            self._t += 1
             if self._t < self.warmup_steps:
                 continue
             yield (trace, Variable(torch.Tensor([1.0])))
-            self._t += 1
         self.kernel.cleanup()
--- a/pyro/infer/mcmc/mcmc.py
+++ b/pyro/infer/mcmc/mcmc.py
@@ -24,20 +24,23 @@ class MCMC(TracePosterior):
         during the warmup phase are discarded.
     """
 
-    def __init__(self, kernel, num_samples, warmup_steps=0):
+    def __init__(self, kernel, num_samples, warmup_steps=0, lag_steps=1):
         self.kernel = kernel
         self.warmup_steps = warmup_steps
         self.num_samples = num_samples
+        self.lag_steps = lag_steps
         if warmup_steps >= num_samples:
             raise ValueError('Number of warmup iterations - {} >= Number of MCMC samples - {}'
                              .format(warmup_steps, num_samples))
         self._t = None
+        self._lag = None
         self.logger = logging.getLogger(__name__)
         super(MCMC, self).__init__()
 
     def _traces(self, *args, **kwargs):
         self.kernel.setup(*args, **kwargs)
         self._t = 0
+        self._lag = 0
         trace = self.kernel.initial_trace()
         self.logger.info('Starting MCMC using kernel - {} ...'.format(self.kernel.__class__.__name__))
         logging_interval = int(math.ceil((self.warmup_steps + self.num_samples) / 20))
@@ -48,8 +51,13 @@ class MCMC(TracePosterior):
                 if diagnostic_info is not None:
                     self.logger.info(diagnostic_info)
             trace = self.kernel.sample(trace)
-            self._t += 1
             if self._t < self.warmup_steps:
+                self._t += 1
                 continue
-            yield (trace, Variable(torch.Tensor([1.0])))
+
+            if self._lag % self.lag_steps == 0:
+                self._t += 1
+                yield (trace, Variable(torch.Tensor([1.0])))
+            self._lag += 1
+
         self.kernel.cleanup()
--- a/pyro/infer/mcmc/mcmc.py
+++ b/pyro/infer/mcmc/mcmc.py
@@ -22,42 +22,39 @@ class MCMC(TracePosterior):
         excluding the samples discarded during the warmup phase.
     :param int warmup_steps: Number of warmup iterations. The samples generated
         during the warmup phase are discarded.
+    :param int sample_every: The number of steps between each trace acceptance
+        after the warmup phase. Sometimes described as lag.
     """
 
-    def __init__(self, kernel, num_samples, warmup_steps=0, lag_steps=1):
+    def __init__(self, kernel, num_samples, warmup_steps=0, sample_every=1):
         self.kernel = kernel
         self.warmup_steps = warmup_steps
         self.num_samples = num_samples
-        self.lag_steps = lag_steps
+        self.sample_every = sample_every
         if warmup_steps >= num_samples:
             raise ValueError('Number of warmup iterations - {} >= Number of MCMC samples - {}'
                              .format(warmup_steps, num_samples))
         self._t = None
-        self._lag = None
         self.logger = logging.getLogger(__name__)
         super(MCMC, self).__init__()
 
     def _traces(self, *args, **kwargs):
         self.kernel.setup(*args, **kwargs)
         self._t = 0
-        self._lag = 0
         trace = self.kernel.initial_trace()
         self.logger.info('Starting MCMC using kernel - {} ...'.format(self.kernel.__class__.__name__))
         logging_interval = int(math.ceil((self.warmup_steps + self.num_samples) / 20))
-        while self._t < self.warmup_steps + self.num_samples:
+        while self._t <= self.warmup_steps + self.num_samples:
             if self._t % logging_interval == 0:
                 self.logger.info('Iteration: {}.'.format(self._t))
                 diagnostic_info = self.kernel.diagnostics(self._t)
                 if diagnostic_info is not None:
                     self.logger.info(diagnostic_info)
             trace = self.kernel.sample(trace)
-            if self._t < self.warmup_steps:
-                self._t += 1
-                continue
 
-            if self._lag % self.lag_steps == 0:
-                self._t += 1
-                yield (trace, Variable(torch.Tensor([1.0])))
-            self._lag += 1
+            self._t += 1
+            if self._t <= self.warmup_steps or self._t % self.sample_every != 0:
+                continue
+            yield (trace, Variable(torch.Tensor([1.0])))
 
         self.kernel.cleanup()
--- a/pyro/infer/mcmc/mh.py
+++ b/pyro/infer/mcmc/mh.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import, division, print_function
 
-from pdb import set_trace as bb
 import pyro
 import pyro.distributions as dist
 from pyro import poutine
--- a/pyro/infer/mcmc/mh.py
+++ b/pyro/infer/mcmc/mh.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
-# from pdb import set_trace as bb
+from pdb import set_trace as bb
 import pyro
 import pyro.distributions as dist
 from pyro import poutine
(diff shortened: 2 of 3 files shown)


pyro/infer/mcmc/mcmc.py
```diff
@@ -24,20 +24,23 @@ class MCMC(TracePosterior):
         during the warmup phase are discarded.
     """
 
-    def __init__(self, kernel, num_samples, warmup_steps=0):
+    def __init__(self, kernel, num_samples, warmup_steps=0, lag_steps=1):
```
fritzo: Please describe `lag_steps` in the above docstring. (I think I've also seen this parameter named `sample_every`)

pyro/infer/mcmc/mh.py
```diff
@@ -0,0 +1,255 @@
...
+def VT(val):
+    return V(T(val))
+
+
+def VTA(val):
+    return VT([val])
+
+
+def assert_valid_trace(trace):
+    assert not any(['fn' in n and isinstance(n['fn'], pyro._Subsample)
```
fritzo: Prefer raising `NotImplementedError` here. (In general, we shouldn't use assertions for user-facing errors, only for programmer errors).
```py
if any(...):
   raise NotImplementedError("...")
```

pyro/infer/mcmc/mh.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+# get all the sample sites
+def sample_sites(trace, minus=set()):
+    return [nid for nid, n in trace.nodes(data=True)
+            if nid not in minus and n["type"] == 'sample'
+            and not n["is_observed"]]
+
+
+class NormalProposal():
```
fritzo: `class NormalProposal(object)`

pyro/infer/mcmc/mh.py
```diff
@@ -0,0 +1,255 @@
...
+        return cur_val + dist.Normal(mu, sigma).sample()
+
+    def log_pdf_given(self, y, given_x):
+        # center our distribution at mu = given_x, then
+        # get log pdf of y
+        mu = given_x
+        sigma = self.scale*self.sigma
+        return dist.Normal(mu, sigma).log_pdf(y)
+
+    # borrowed from:
```
fritzo: We need to either paste of their licenses here or find an original source to cite and rewrite the code.

pyro/infer/mcmc/mh.py
```diff
@@ -0,0 +1,255 @@
...
+        self._kernel_traces.append(trace)
+        return trace
+
+    def cleanup(self):
+        for trace in self._kernel_traces:
+            self._cleanup_trace(trace)
+        self._kernel_traces = []
+
+    def _cleanup_trace(self, reject_trace, accept_trace=None):
+        pass
```
fritzo: What is your plan for this method?

pyro/infer/mcmc/mh.py
```diff
@@ -0,0 +1,255 @@
...
+
+    @property
+    def num_accepts(self):
+        return self._accept_cnt
+
+    def _get_trace_proposal(self, trace, r_site, proposal_value):
```
fritzo: This method is a nice conceptual unit that we could reuse. I'd recommend factoring it out into a free standing helper either in this file or in poutine/trace.py:
```py
def truncate_trace(tr, end_site):
    """
    Truncate a trace up until (but not including) an ``end_site``.
    :param Trace tr: A source trace.
    :param str end_site: The name of an end site to truncate after.
    """
    ...your logic here...
```

pyro/infer/mcmc/mh.py
```diff
@@ -0,0 +1,255 @@
...
+from torch import Tensor as T
+from torch.autograd import Variable as V
+from numpy.random import choice
+from numpy import isfinite
+from torch import log as tlog
+
+_R = '_RETURN'
+
+
+def VT(val):
```
fritzo: nit: This is too clever. Just spell out `Variable(Tensor(...))` everywhere.

Rationale: aliases are nice when you're in a jupyter notebook. But we'll be maintaining this code for a long time, including programming with `grep/sed/awk` while refactoring. It's really difficult to make sweeping changes on a maintained codebase when files use aliases like this. In particular, Neeraj and I will have a lot of work to do when PyTorch eliminates Variables, and each nickname in the code will make our lives more difficult.

pyro/infer/mcmc/mh.py
```diff
@@ -0,0 +1,255 @@
...
+    return V(T(val))
+
+
+def VTA(val):
```
fritzo: ditto: Just spell out `Variable(torch.Tensor([val]))`

pyro/infer/mcmc/mh.py
```diff
@@ -0,0 +1,255 @@
...
+from pyro import poutine
+from pyro.infer.mcmc.trace_kernel import TraceKernel
+from pyro.util import ng_ones, ng_zeros
+from torch import Tensor as T
+from torch.autograd import Variable as V
+from numpy.random import choice
+from numpy import isfinite
+from torch import log as tlog
+
+_R = '_RETURN'
```
fritzo: nit: Too clever. Just spell out `'_RETURN'`.

tests/infer/mcmc/test_mh.py
```diff
@@ -0,0 +1,64 @@
...
+        mcmc_run = MCMC(MH(self.model, NormalProposal(mu=ng_zeros(1), sigma=ng_ones(1), tune_frequency=100)),
+                        num_samples=1000, warmup_steps=150)
+
+        marginal = Marginal(mcmc_run)
+        posterior_samples = [marginal() for i in range(1000)]
+
+        posterior_mean = torch.mean(torch.cat(posterior_samples))
+        posterior_stddev = torch.std(torch.cat(posterior_samples), 0)
+
+        self.assertEqual(0, torch.norm(posterior_mean - self.mu_mean).data[0],
```
fritzo: Prefer the nice `assert_equal` helper from test/common.py:
```py
from common import assert_equal
assert_equal(posterior_mean, self.mu_mean, prec=0.03)
```
Also nit, a nice convention I like to follow in testing is to name these `self.expected_mean` and then compute `actual_mean = torch.mean(...)` and check `assert_equal(actual_mean, expected_mean, prec=0.03)`. The advantage of this naming convention is that when a maintainer breaks some upstream piece of code during refactoring, it's clear which value is known truth (`expected`) and which is the value under test (`actual`).

tests/infer/mcmc/test_mh.py
```diff
@@ -0,0 +1,64 @@
...
+
+        def model():
+            mu = pyro.sample("mu", Normal(Variable(torch.zeros(1)),
+                                          Variable(torch.ones(1))))
+            xd = Normal(mu, Variable(torch.ones(1)), batch_size=50)
+            pyro.observe("xs", xd, self.data)
+            return mu
+
+        # data
+        self.data = Variable(torch.zeros(50, 1))
```
fritzo: nit: I would recommend a dataset for which it is more difficult to accidentally guess the parameters (ie with buggy code). For example, put 2 points on +1 and 1 point on +2. Then the posterior mean is 1.0 and stddev is sqrt(0.5). (I just worry that with the current values, 1/50 and 1/51 are indistinguishable, so you can't tell whether the prior has been included)

pyro/infer/mcmc/mh.py
```diff
@@ -0,0 +1,255 @@
...
+import pyro
+import pyro.distributions as dist
+from pyro import poutine
+from pyro.infer.mcmc.trace_kernel import TraceKernel
+from pyro.util import ng_ones, ng_zeros
+from torch import Tensor as T
+from torch.autograd import Variable as V
+from numpy.random import choice
+from numpy import isfinite
+from torch import log as tlog
```
fritzo: nit: Too clever, just spell out `torch.log`.

fritzo (commented):
Looks reasonable overall. Lots o nits :smile:

I think we could factor out some more helper functions to manipulate traces. I'll look through trace logic in more detail on a second review pass.

fritzo:
@neerajprad Do we need a new file `tests/infer/mcmc/conftest.py`, or will the new test `tests/infer/mcmc/test_mh.py` pick up the parent config file `tests/infer/conftest.py`?

pyro/infer/mcmc/mcmc.py
```diff
@@ -49,7 +52,12 @@ def _traces(self, *args, **kwargs):
                     self.logger.info(diagnostic_info)
             trace = self.kernel.sample(trace)
             if self._t < self.warmup_steps:
```
neerajprad: Can we simplify this logic as follows:
```
self._t += 1
if self._t <= self.warmup_steps or self._t % self.lag != 0:
  continue
yield (trace, Variable(torch.Tensor([1.0])))
```

We should also change the while loop above to `<=`.

neerajprad:
> @neerajprad Do we need a new file tests/infer/mcmc/conftest.py, or will the new test tests/infer/mcmc/test_mh.py pick up the parent config file tests/infer/conftest.py?

It should pick up the config from the parent.

pyro/infer/mcmc/mh.py
```diff
@@ -0,0 +1,255 @@
...
+            # increase by ten percent
+            scale *= 1.1
+
+        self.scale = scale
+        # reset our tuning count
+        self._tune_cnt = self.tune_frequency
+        return self
+
+
+class MH(TraceKernel):
```
neerajprad: Lets also add a docstring for this kernel class, and put in any relevant citations here, so as to have it reflect in our documentation.

tests/infer/mcmc/test_mh.py
OptimusLime: Happy to change this, it seemed a bit odd to me as well. However, I remain a bit confused mainly because this is copied directly from an [existing pyro test](https://github.com/uber/pyro/blob/dev/tests/infer/test_sampling.py#L111). 

In other words, I was under the assumption that existing tests were a model for new tests. Is this not the case?

tests/infer/mcmc/test_mh.py
OptimusLime: As this line comes from [an existing test](https://github.com/uber/pyro/blob/dev/tests/infer/test_sampling.py#L59), can you explain the intuitions for why it's okay for an importance sampling test, but isn't very good for MH? I don't think I understand the subtlety, and it might help me to write improved MH tests.

tests/infer/mcmc/test_mh.py
fritzo: Some of Pyro's tests still use older unittest assertions, but newer `assert` statements are now preferred. If you're ever modifying a test file in older style, feel free to clean up and convert everything to newer style.

tests/infer/mcmc/test_mh.py
eb8680: The intuition here is just that the importance sampling test (which I wrote in a hurry a long time ago) isn't a particularly good test.

tests/infer/mcmc/test_mh.py
eb8680: Also, `test_sampling.py` is kind of a mess, at least statistically.  We should rewrite it now that we're adding more sampling algorithms.

ngoodman:
i'll try to take a look at the details sometime soon, but i'd also like to see some of the MH tests from webppl pulled in. especially the ones that exercise trans-dimensional / structure change issues, since these are the subtlest to get right in the fw/bw computation.

tests/infer/mcmc/test_mh.py
```diff
@@ -0,0 +1,64 @@
...
+class NormalNormalSamplingTestCase(TestCase):
+
+    def setUp(self):
+
+        pyro.clear_param_store()
+
+        def model():
+            mu = pyro.sample("mu", Normal(Variable(torch.zeros(1)),
+                                          Variable(torch.ones(1))))
+            xd = Normal(mu, Variable(torch.ones(1)), batch_size=50)
```
fritzo: I believe the syntax `batch_size=50` has been replaced with `sample_shape=(50,)`.

fritzo (commented):
Paul what is the status of this PR? As I understand it just needs docstrings and merge conflict resolution?

FYI This week Neeraj and I are refactoring Pyro to fully support PyTorch distributions. It will save you some work if you get this PR merged soon. If you wait a couple weeks there may be much messier merge conflicts for you to resolve.

PR #624, opened 2017-12-13
Fix GPU device placement when calling tensor.cuda()

### LINKED ISSUE #623: Fix GPU placement when calling tensor.cuda()
Adam pointed out that when matching the `.is_cuda` property of a new tensor to an existing tensor, we should also take care to match the device:
```py
old_tensor = ...
new_tensor = torch.arange(10)
if old_tensor.is_cuda:
    new_tensor = new_tensor.cuda(old_tensor.get_device())
```

fritzo:
Fixes #623 
Suggested by @apaszke

This also fixes the shape of `distributions.Delta.enumerate_support()`, which was returning an element rather than a singleton tensor containing that element (to see the difference, consider `for x in dist.enumerate_support(): ...`).

## Testing

I tested that this runs on a single GPU but I have not tested on a multi-GPU system.

### DIFF
--- a/pyro/distributions/delta.py
+++ b/pyro/distributions/delta.py
@@ -75,4 +75,4 @@ class Delta(Distribution):
         :return: torch variable enumerating the support of the delta distribution.
         :rtype: torch.autograd.Variable.
         """
-        return Variable(self.v.data)
+        return Variable(self.v.data.unsqueeze(0))
--- a/pyro/distributions/bernoulli.py
+++ b/pyro/distributions/bernoulli.py
@@ -106,7 +106,7 @@ class Bernoulli(Distribution):
             sample.
         :rtype: torch.autograd.Variable.
         """
-        result = torch.arange(2).long()
+        result = torch.arange(0, 2).long()
         result = result.view((-1,) + (1,) * self.ps.dim()).expand((-1,) + self.ps.size())
         if self.ps.is_cuda:
             result = result.cuda(self.ps.get_device())
--- a/pyro/distributions/bernoulli.py
+++ b/pyro/distributions/bernoulli.py
@@ -106,7 +106,11 @@ class Bernoulli(Distribution):
             sample.
         :rtype: torch.autograd.Variable.
         """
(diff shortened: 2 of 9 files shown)


tests/infer/mcmc/test_mcmc.py
```diff
@@ -55,5 +55,5 @@ def test_mcmc_interface():
         samples.append(marginal.sample(data))
     sample_mean = torch.mean(torch.stack(samples), 0)
     sample_std = torch.std(torch.stack(samples), 0)
-    assert_equal(sample_mean.data, torch.Tensor([0.0]), prec=5e-2)
-    assert_equal(sample_std.data, torch.Tensor([1.0]), prec=5e-2)
+    assert_equal(sample_mean.data, torch.Tensor([0.0]), prec=0.1)
```
fritzo: I had to loosen this tolerance due to flakiness. I assume the CUDA rng state is slightly different.

pyro/distributions/delta.py
```diff
@@ -75,4 +75,4 @@ def enumerate_support(self, v=None):
         :return: torch variable enumerating the support of the delta distribution.
         :rtype: torch.autograd.Variable.
         """
-        return Variable(self.v.data)
+        return Variable(self.v.data.unsqueeze(0))
```
neerajprad: Thanks for this fix.

neerajprad:
Not sure what's wrong, but all the builds are returning:
```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
```

neerajprad:
Can we remove #627 from this PR, to get this merged earlier?

fritzo:
> Can we remove #627 from this PR

This PR relies on newer functionality of `torch.arange()` that is only available in 0.4. I'll see how difficult it is to use an older invocation of `torch.arange()`...

neerajprad:
> This PR relies on newer functionality of torch.arange() that is only available in 0.4. 

If there is a lot of effort involved, we could just wait for the smaller wheel to be available.

fritzo:
@neerajprad Looks like this is ready to merge

PR #626, opened 2017-12-13
Drop depencency on unittest

fritzo:
Based on https://github.com/uber/pyro/pull/625
Suggested by @eb8680 

This currently fails. I haven't figured out how to get pytest to run setup and teardown methods.

Note that pytest discovery requires naming tests as `TestFoo` rather than `FooTests`.

### DIFF
--- a/pyro/__init__.py
+++ b/pyro/__init__.py
@@ -2,10 +2,10 @@ from __future__ import absolute_import, division, print_function
 
 import contextlib
 import copy
+import logging
 import warnings
 from collections import OrderedDict
 from inspect import isclass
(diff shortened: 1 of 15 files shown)


eb8680:
> This currently fails. I haven't figured out how to get pytest to run setup and teardown methods.

Can this work as currently implemented?  Won't the tests have to be refactored to work correctly with `pytest`?

fritzo:
> Can this work as currently implemented?

Probably not. It looks like this would require more refactoring. I'll close for now. I'm happy to re-open if you see an easy way to do this.

PR #629, opened 2017-12-14
partial implementation of lax gradient estimator in tracegraph_elbo

martinjankowiak:
This implements the LAX part of RELAX from [Grathwohl et al. 2017](https://arxiv.org/abs/1711.00123).

this is pretty simple. this is exactly like the baseline estimator we already have. the only
different is that the baseline is trained via a different loss function: one tries to directly minimize
(gradient) variance instead of minimizing an ad hoc square loss of the form (baseline-cost)**2.

most of the complications have to do with using torch.autograd to setup the new baseline loss (which is necessary because of higher-order derivatives). this is currently a bit of a mess, especially since we're currently mixing autograd.backward() and autograd.grad().

there are still some known issues (e.g. the batch_log_pdf case). so this is very much a wip. putting
this here mostly for @fritzo's benefit.

### DIFF
--- a/pyro/infer/tracegraph_elbo.py
+++ b/pyro/infer/tracegraph_elbo.py
@@ -142,7 +142,6 @@ def _compute_elbo_non_reparam(guide_trace, guide_vec_md_nodes,  #
     # we include only downstream costs to reduce variance
     # optionally include baselines to further reduce variance
     # XXX should the average baseline be in the param store as below?
-    surrogate_elbo = 0.0
     baseline_loss = 0.0
     # the following are needed in case we're using lax baselines
     full_gradients = [None] * len(trainable_params_minus_bl_params)
@@ -186,6 +185,7 @@ def _compute_elbo_non_reparam(guide_trace, guide_vec_md_nodes,  #
                                     create_graph=True, allow_unused=True)
         for k in range(len(grads_log_q)):
             g, p = grads_log_q[k], trainable_params_minus_bl_params[k]
+            # assemble the reinforce-like term
             cost_times_grad = downstream_cost * g.detach() if g is not None else None
             if cost_times_grad is None:
                 continue  # log_q didn't depend on parameter p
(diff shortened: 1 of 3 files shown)


pyro/infer/tracegraph_elbo.py
```diff
@@ -282,28 +323,31 @@ def _loss_and_grads_particle(self, weight, model_trace, guide_trace):
...
-                               for site in trace.nodes.values()
-                               if site["type"] == "param")
+            baseline_loss = _compute_elbo_non_reparam(
+                    guide_trace, guide_vec_md_nodes, non_reparam_nodes, downstream_costs,
+                    trainable_params_minus_bl_params)
 
         if trainable_params:
             surrogate_loss = -surrogate_elbo
-            torch_backward(weight * (surrogate_loss + baseline_loss))
+            # XXX this try block is a hack that will go away when we phase out backward()
```
fritzo: What is the `RuntimeError` that is raised? Can we move this try block into `torch_backward`?

pyro/infer/tracegraph_elbo.py
```diff
@@ -147,24 +164,48 @@ def _compute_elbo_non_reparam(guide_trace, guide_vec_md_nodes,  #
...
+            else:
+                p.grad -= cost_times_grad  # otherwise accumulate p gradient
+            if use_lax:  # if using lax, we keep track of the gradients to compute the lax baseline_loss
+                assert(use_nn_baseline), "lax estimator must use nn_baseline and not baseline_value"
+                if full_gradients[k] is None:
+                    full_gradients[k] = cost_times_grad
+                else:
+                    full_gradients[k] += cost_times_grad
+
+    # if any nodes have use_lax then construct lax baseline_loss
```
fritzo: Punctuate your comments. The lack of period in this comment is confusing and burdens readers with an unnecessary inference task of parsing unstructured text.

pyro/infer/tracegraph_elbo.py
martinjankowiak: it has to do with the fact that backward doesn't have an allow_unused=True flag
(so if there is nothing to differentiate, it fails)

fritzo: Could you add that to the comment in this code?

martinjankowiak: added more comments

eb8680:
This is pretty cool, would be awesome to have a test or example that compares and plots gradient variance for SCG and LAX estimators for a few different models.

martinjankowiak:
going to close this for now given other refactoring that's happening in tracegraph_elbo. will revisit when it's clearer lax is something we want (my initial experiments didn't show any noticeable improvement upon vanilla neural baselines)

PR #630, opened 2017-12-15
Wrap torch.distributions.Gamma, Beta

fritzo:
This refactors the `torch.distribution` wrapping layer and adds a new `TorchDistribution` class to implement common wrapper methods. The new class hierarchy is:

- `pyro.distributions.Distribution`
    - `pyro.distributions.Normal`
    - ...
    - `pyro.distributions.torch_wrapper.TorchDistribution`
        - `pyro.distributions.torch.Normal` -> has a `torch.distributions.Normal`
        - ...
    - `pyro.distributions.RandomPrimitive` -> has a `pyro.distributions.Distribution`

Pytorch implementations are now available side-by-side with older Pyro implementations, e.g,
```py
from pyro.distributions import normal as old_normal
from pyro.distributions.torch import normal as new_normal
```

## Tested

Ran `make test-torch-dist` against the [test-rsample](https://github.com/probtorch/pytorch/tree/test-rsample) branch of pytorch (which implements `Beta`).

### DIFF
--- a/pyro/distributions/__init__.py
+++ b/pyro/distributions/__init__.py
@@ -7,7 +7,7 @@ from pyro.distributions.binomial import Binomial
 from pyro.distributions.categorical import WrapCategorical as Categorical
 from pyro.distributions.cauchy import Cauchy
 from pyro.distributions.delta import Delta
-from pyro.distributions.dirichlet import Dirichlet
+from pyro.distributions.dirichlet import WrapDirichlet as Dirichlet
 from pyro.distributions.distribution import Distribution  # noqa: F401
 from pyro.distributions.exponential import Exponential
 from pyro.distributions.gamma import WrapGamma as Gamma
--- a/pyro/distributions/__init__.py
+++ b/pyro/distributions/__init__.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 # distribution classes
 from pyro.distributions.bernoulli import Bernoulli
-from pyro.distributions.beta import Beta
+from pyro.distributions.beta import WrapBeta as Beta
 from pyro.distributions.binomial import Binomial
 from pyro.distributions.categorical import WrapCategorical as Categorical
 from pyro.distributions.cauchy import Cauchy
@@ -10,7 +10,7 @@ from pyro.distributions.delta import Delta
 from pyro.distributions.dirichlet import Dirichlet
 from pyro.distributions.distribution import Distribution  # noqa: F401
 from pyro.distributions.exponential import Exponential
-from pyro.distributions.gamma import Gamma
+from pyro.distributions.gamma import WrapGamma as Gamma
 from pyro.distributions.half_cauchy import HalfCauchy
 from pyro.distributions.log_normal import LogNormal
 from pyro.distributions.multinomial import Multinomial
--- a/pyro/distributions/__init__.py
+++ b/pyro/distributions/__init__.py
@@ -1,24 +1,37 @@
 from __future__ import absolute_import, division, print_function
 
-# distribution classes
+import os
+
 from pyro.distributions.bernoulli import Bernoulli
-from pyro.distributions.beta import WrapBeta as Beta
 from pyro.distributions.binomial import Binomial
-from pyro.distributions.categorical import WrapCategorical as Categorical
+from pyro.distributions.categorical import Categorical
 from pyro.distributions.cauchy import Cauchy
 from pyro.distributions.delta import Delta
-from pyro.distributions.dirichlet import WrapDirichlet as Dirichlet
 from pyro.distributions.distribution import Distribution  # noqa: F401
 from pyro.distributions.exponential import Exponential
-from pyro.distributions.gamma import WrapGamma as Gamma
 from pyro.distributions.half_cauchy import HalfCauchy
 from pyro.distributions.log_normal import LogNormal
 from pyro.distributions.multinomial import Multinomial
-from pyro.distributions.normal import WrapNormal as Normal
 from pyro.distributions.one_hot_categorical import OneHotCategorical
 from pyro.distributions.poisson import Poisson
 from pyro.distributions.random_primitive import RandomPrimitive
 from pyro.distributions.uniform import Uniform
+
+# TODO Decide based on torch.__version__ once torch.distributions matures.
+USE_TORCH_DISTRIBUTIONS = int(os.environ.get('PYRO_USE_TORCH_DISTRIBUTIONS', 0))
+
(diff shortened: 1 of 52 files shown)


fritzo:
@martinjankowiak `torch.digamma` is not yet in PyTorch master (https://github.com/pytorch/pytorch/pull/3955), so the `Gamma` in this PR weirdly has reparamterized gradients but a non-differentiable score function 🙃

pyro/distributions/gamma.py
```diff
@@ -94,3 +95,34 @@ def analytic_var(self):
...
         """
         return self.alpha / torch.pow(self.beta, 2.0)
+
+
+class TorchGamma(TorchDistribution):
+    """
+    Compatibility wrapper around
+    `torch.distributions.Gamma <http://pytorch.org/docs/master/_modules/torch/distributions.html#Gamma>`_
+    """
+    reparameterized = True
```
fritzo: @martinjankowiak you should be able to test reparameterized `Gamma` on this branch

neerajprad:
@fritzo - In light of what we discussed yesterday, are we looking to get this PR merged? I was thinking that once we start testing against PyTorch master (or some pinned version of master), we can just remove our implementation code and instead wrap around `torch.distribution` classes directly, i.e. `TorchNormal --> Normal`.

fritzo:
> are we looking to get this PR merged?

Yes, I just haven't had a chance to update this PR in light of yesterday's #638 and #640 .

fritzo:
@neerajprad @martinjankowiak could you take another look? I've refactored to support side-by-side use of Pyro and PyTorch distributions:
```py
from pyro.distributions.normal import Normal as OldNormal
from pyro.distributions.torch import Normal as NewNormal
```
The `PYRO_USE_TORCH_DISTRIBUTIONS` environment variable is needed only in testing; at some point we can flip the switch to use those distributions by default.

It would be nice to merge this soon to facilitate experimentation with the PyTorch distributions.

pyro/distributions/__init__.py
```diff
@@ -1,25 +1,38 @@
...
 from pyro.distributions.one_hot_categorical import OneHotCategorical
 from pyro.distributions.poisson import Poisson
 from pyro.distributions.random_primitive import RandomPrimitive
 from pyro.distributions.uniform import Uniform
 
+# TODO Decide based on torch.__version__ once torch.distributions matures.
+USE_TORCH_DISTRIBUTIONS = int(os.environ.get('PYRO_USE_TORCH_DISTRIBUTIONS', 0))
+
+# distribution classes with working torch versions
+if USE_TORCH_DISTRIBUTIONS:
```
neerajprad: Nice!

pyro/distributions/torch/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+from pyro.distributions.distribution import Distribution  # noqa: F401
+from pyro.distributions.random_primitive import RandomPrimitive
+from pyro.distributions.torch.beta import Beta
+from pyro.distributions.torch.categorical import Categorical
+from pyro.distributions.torch.dirichlet import Dirichlet
+from pyro.distributions.torch.gamma import Gamma
+from pyro.distributions.torch.normal import Normal
+
+# function aliases
+beta = RandomPrimitive(Beta)
```
neerajprad: Do we need to expose these functional forms here too (since they should get initialized in `distributions/__init__.py`)?

fritzo: It's convenient to expose those here for side-by-side testing (as Martin is doing today):
```py
from pyro.distributions import normal as old_normal
from pyro.distributions.torch import normal as new_normal
```

fritzo: or
```py
import pyro.distributions as old_dist
import pyro.distributions.torch as new_dist
```

neerajprad: I see, thanks for explaining.

pyro/distributions/util.py
```diff
@@ -1,24 +1,29 @@
...
+
+    def decorator(destin_class):
+        destin_class.__doc__ = source_class.__doc__
+        for name in dir(destin_class):
+            if name.startswith('_'):
+                continue
+            try:
+                source_attr = getattr(source_class, name)
+                destin_attr = getattr(destin_class, name)
+                destin_attr.__doc__ = source_attr.__doc__
```
neerajprad: Should we check if `destin_attr.__doc__` already exists and only change if it doesn't, so that the subclasses don't have the docstrings replaced by the base class?

fritzo: Good idea, I'll do that.

fritzo: Done.

pyro/distributions/util.py
```diff
@@ -1,24 +1,31 @@
...
 import torch
 import torch.nn.functional as F
 from torch.autograd import Variable
 
-# TODO Decide based on torch.__version__ once torch.distributions matures.
-USE_TORCH_DISTRIBUTIONS = int(os.environ.get('PYRO_USE_TORCH_DISTRIBUTIONS', 0))
-
 
-def torch_wrapper(pyro_dist):
+def copy_docs_from(source_class):
```
jpchen: heh this is cute. are we eventually going to migrate them completely so theres only one distribution? will it be `pyro.distributions` or `pyro.distributions.torch`?

pyro/distributions/torch/normal.py
```diff
@@ -0,0 +1,25 @@
...
+from pyro.distributions.torch_wrapper import TorchDistribution
+from pyro.distributions.util import broadcast_shape, copy_docs_from
+
+
+@copy_docs_from(_Normal)
+class Normal(TorchDistribution):
+    reparameterized = True
+
+    def __init__(self, mu, sigma, *args, **kwargs):
+        torch_dist = torch.distributions.Normal(mu, sigma)
```
jpchen: so from what i can see, this returns the pt distribution which only provides sample and score methods.  but batch and event shapes eventually be pushed upstream as well, correct?

neerajprad: We have batch and event shapes in `torch.distributions`, but they are internal attributes and not exposed as an API. Even in Pyro, we only need to use it, outside of the distribution instances, for testing. We still need to add analytical mean/variance to `torch.distributions` classes; otherwise, they should already be at feature parity.

pyro/distributions/util.py
fritzo: We're planning to eventually migrate docs. Thereafter we can still use this decorator to copy docs from `Distributions` to instance classes.

PR #640, opened 2017-12-20
Disallow instance override of .reparameterized property

### LINKED ISSUE #631: Create a MockNormal class; drop instance .reparameterized
## Background

Currently `.reparameterized` is a class-level attribute but can be overridden by instance attributes. This override is only used in testing, where we make sure that `TraceGraph_ELBO` performs the same inference on a reparameterized and non-reparameterized `Normal`.

## The problem

The instance attribute `.reparameterized` causes a few problems:
1. It requires extra plumbing for all `Distribution` classes (they must pass `*args,**kwargs` to `super(...).__init__()`)
2. It is not supported by `torch.distributions` to which we are migrating
3. It is not respected by `RandomPrimitive` which simply ignores the instance attribute and returns the class attribute. (This is necessary because `RandomPrimitive` only lazily constructs instances, and there may be no instance to examine).
4. It complicates the torch.distribution wrapper layer. The wrapper dynamically chooses between pyro vs torch implementations (torch does not support fancy things like `batch_size`). Some torch distributions are reparameterized whereas their Pyro equivalents are not (`Gamma`, `Beta`, `Dirichlet`). Therefore `RandomPrimitive` cannot statically determine whether a distribution is reparameterized.

## Proposed solution

We should eliminate the instance-level `.reparameterized` property. For the few cases where we need a non-reparameterized `Normal` in testing, we should simply implement a `NonreparamNormal` class (or even a `Nonreparameterized` class -> class transform).

fritzo:
Fixes #631 

This removes plumbing that allows Distribution instances to override the class-level `.reparameterized` property. Removing this plumbing simplifies our migration to torch.distributions. This plumbing was only used in tests; to provide the non-reparameterized Normal distribution provided by that override, this PR introduces a `fakes.NonreparameterizedNormal` distribution that is exposed only in tests/.

### DIFF
--- a/tests/fakes.py
+++ b/tests/fakes.py
@@ -1,0 +1,11 @@
+from __future__ import absolute_import, division, print_function
+
+from pyro.distributions.normal import Normal
+from pyro.distributions.random_primitive import RandomPrimitive
+
+
+class NonreparameterizedNormal(Normal):
+    reparameterized = False
+
+
+nonreparameterized_normal = RandomPrimitive(NonreparameterizedNormal)
--- a/pyro/distributions/distribution.py
+++ b/pyro/distributions/distribution.py
@@ -88,16 +88,11 @@ class Distribution(object):
     reparameterized = False
     enumerable = False
 
-    def __init__(self, reparameterized=None):
+    def __init__(self):
         """
         Constructor for base distribution class.
-
-        :param bool reparameterized: Optional argument to override whether
-            instance should be considered reparameterized (by default, this
-            is decided by the class).
         """
-        if reparameterized is not None:
-            self.reparameterized = reparameterized
+        pass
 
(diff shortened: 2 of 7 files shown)


tests/fakes.py
```diff
@@ -0,0 +1,11 @@
...
+
+from pyro.distributions.normal import Normal
+from pyro.distributions.random_primitive import RandomPrimitive
+
+
+class NonreparameterizedNormal(Normal):
+    reparameterized = False
+
+
+nonreparameterized_normal = RandomPrimitive(NonreparameterizedNormal)
```
jpchen: why do you need this? cant you set the class attribute directly in the tests in your conditional?

neerajprad: Once all our distributions are migrated, are we planning to use this as an alias for `has_rsample`?

PR #641, opened 2017-12-21
Remove debug statement

fritzo:
sorry, no need to run tests imo


fritzo:
I'll just include this in another PR.

PR #642, opened 2017-12-21
Use @copy_docs_from on all existing distributions

fritzo:
This simplifies our task of migrating docs from pyro.distributions.* to pyro.distribution.torch.*

## Tested

Ran `make docs` and verified that they look good.

### DIFF
--- a/pyro/distributions/bernoulli.py
+++ b/pyro/distributions/bernoulli.py
@@ -4,9 +4,10 @@ import torch
 from torch.autograd import Variable
 
 from pyro.distributions.distribution import Distribution
-from pyro.distributions.util import get_probs_and_logits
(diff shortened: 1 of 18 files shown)


jpchen:
@neerajprad right now our docs only has one version ie release, which is also hardcoded in the sphinx config... i dont think we ever made that dynamic before launch. do you know the extension [webppl uses](http://webppl.readthedocs.io/en/master/) to suppor multiple versions?

neerajprad:
@jpchen - I don't think we need any special extensions. The read the docs theme supports two versions - a stable one and the latest one (https://docs.readthedocs.io/en/latest/versions.html), but we may need to change our default branch to `dev`. We can try it out tomorrow and see if this works. We also need this for the tutorials, which I am guessing should be the same.

PR #644, opened 2017-12-21
Provide polyfill shim for torch.no_grad()

fritzo:
PyTorch 0.4 is dropping the `.volatile` attribute in favor of a `torch.no_grad()` context manager. Pyro's tests suite currently fails against PyTorch master. This PR fixes the tests by:
- adding a `pyro.shim` module
- addding `torch_no_grad()` and `is_volatile()` methods that work in PyTorch 0.2, 0.3, and 0.4 (master)
- removing all `volatile=True` from codebase (I believe this changes performance but not semantics)

## Tested:
- CI tests run against PyTorch 0.2
- ran `make test` against PyTorch 0.3
- ran `make test` against PyTorch 0.4 master 2017-12-21 (tests failed before this PR)


fritzo:
merge at will

PR #647, opened 2017-12-21
Deprecate Categorical vs argument

### LINKED ISSUE #646: Drop support for vs in Categorical distribution
This issue proposes to separate indexing logic from distribution logic in the `Categorical` distribution.

Currently Pyro's `Categorical` distribution bakes in logic to index against a list of values. This design choice leads to a number of difficulties:
- `Categorical` is difficult to test and requires lots of custom testing logic.
- It is difficult to migrate to `torch.distributions` whose categorical works with `LongTensor`s.

Support for `vs` may be added back in later via a discrete `Bijector`.

fritzo:
Closes #646 
Blocking #606 

This deprecates the `vs` argument to `Categorical`:
- Removes `vs` from tests (so we can swap in `torch.distributions.Categorical`)
- Adds a `UserWarning` indicating that `vs` is deprecated. (In general `UserWarning`s are preferred over `DeprecationWarning`s because Python 2.7 silences `DeprecationWarning` by default whereas Python 3.x does not silence them).

Note that the implementation has been left in place. We will remove the old implementation when moving to PyTorch distributions.

### DIFF
--- a/pyro/distributions/categorical.py
+++ b/pyro/distributions/categorical.py
@@ -1,11 +1,13 @@
 from __future__ import absolute_import, division, print_function
 
-import numpy as np
-import torch
-from torch.autograd import Variable
+import warnings
 
+import numpy as np
+
+import torch
 from pyro.distributions.distribution import Distribution
 from pyro.distributions.util import copy_docs_from, get_probs_and_logits, torch_multinomial, torch_zeros_like
+from torch.autograd import Variable
 
 
 @copy_docs_from(Distribution)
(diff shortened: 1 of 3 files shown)


pyro/distributions/categorical.py
```diff
@@ -41,6 +43,7 @@ def __init__(self, ps=None, vs=None, logits=None, batch_size=None, log_pdf_mask=
         self.vs = self._process_data(vs)
         self.log_pdf_mask = log_pdf_mask
         if vs is not None:
+            warnings.warn('Categorical vs argument is deprecated', UserWarning)
```
neerajprad: Are there existing examples/tutorials that use this argument currently? If not, we might as well just remove it on `dev`, as it will be removed by the next release.

fritzo: I believe `vs` is not used in any examples or tutorials. I looked into removing `vs` code but that turned out to be more work than adding a simple warning :smile:

neerajprad: Looks good! We can definitely get this merged as is, and remove it completely along with the tests soon.

fritzo:
@neerajprad ready to merge

PR #648, opened 2017-12-22
Adding docker support for pyro

### LINKED ISSUE #195: Pyro docker image
As discussed in #114, it will be nice to have a dockerfile that we can use to build a docker container with pyro and all its dependencies packaged in, specially for running on a GPU. I suppose this will need to wait until we make Pyro public (as pip / git clone will not work with restricted access), but opening this task to keep track of the issue.

neerajprad:
This adds docker support for Pyro. This makes it simple to build Pyro over PyTorch using different configurations:
 - released conda/PyPi packages vs. building both from source using any git branch.
 - using python 2/3
 - using CPU/CUDA.

The Makefile has some common recipes to specify how to build the image. The documentation on building docker images and running them is in the [readme](https://github.com/neerajprad/pyro/tree/dockerfile/docker/README.md). It also makes it easy to run notebook or ipython terminal using the built image. To handle data sharing between the container and the user's host system, we use the same uid/guid while building the image and have a shared workspace that can be configured from the Makefile. This is to make it easy to share data / view results, etc.

Tested locally using different configurations. Since I have refactored this quite a bit, I will continue to test that things are working before getting this merged. 

### TODO:
While building PyTorch from source using CUDA 9, we get a few intermediate errors even though the build is finally successful, and functional. This still needs a bit more investigation.

Fixes #195

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,3 +45,7 @@ git clone git@github.com:uber/pyro.git
 cd pyro
 pip install .
 ```
+
+## Running Pyro from a Docker Container
+
+Refer to the instructions [here](docker/README.md).
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,53 @@
+ARG base_img=ubuntu:16.04
+FROM ${base_img}
+
+# Optional args
+ARG cuda=0
+ARG python_version=2.7
+ARG pyro_branch=release
+ARG pytorch_branch=release
+ARG uid=1000
+ARG gid=1000
+
+# Configurable settings
+ENV USER_NAME pyromancer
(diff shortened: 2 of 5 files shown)


neerajprad:
> Ready to merge?

I did a bunch of refactoring, so running a few quick tests. We should be good to merge after that.

neerajprad:
@fritzo - Thanks for reviewing. We can merge this pending @jpchen's approval.

@rohitsingh0812, @OptimusLime - Please feel free to use / modify this for your use case, and let me know what can be improved.

jpchen:
@all will test cpu please hold off on merging until then, thanks

PR #649, opened 2017-12-22
Wrap torch Categorical and OneHotCategorical

fritzo:
Addresses #606 

This wraps `torch.distributions.Categorical` for use in Pyro as both `Categorical` and `OneHotCategorical`. PyTorch does not yet have a native `OneHotCategorical`. The advantage of migrating both distributions now is that the PyTorch version supports full broadcasting and will be easier to update to our new batching semantics.

This also fixes and updates some of the `tests/infer/test_inference.py` tests:
- Correctly splits reparameterized/nonreparameterized tests now that `Gamma` and `Beta` are reparameterized
- Marks one of the `Beta` tests as xfail (@martinjankowiak and @fritzo are debugging it, this is likely due to low-precision numerics in our PyTorch code and will soon be made more precise)
- Eliminates `map_data()` from those tests, reducing cost by ~30%

## Tested

- Ran `make test-torch-dist` against `test-rsample` branch of PyTorch.

### DIFF
--- a/pyro/distributions/torch/one_hot_categorical.py
+++ b/pyro/distributions/torch/one_hot_categorical.py
@@ -5,7 +5,7 @@ from torch.autograd import Variable
 
 from pyro.distributions.one_hot_categorical import OneHotCategorical as _OneHotCategorical
 from pyro.distributions.torch_wrapper import TorchDistribution
-from pyro.distributions.util import copy_docs_from
+from pyro.distributions.util import copy_docs_from, get_probs_and_logits
 
 
 @copy_docs_from(_OneHotCategorical)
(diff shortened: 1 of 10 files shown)


fritzo:
@neerajprad the `Categorical` and `OneHotCategorical` wrappers appear to work now.

@martinjankowiak could you please review changes to `test_inference.py`?

pyro/distributions/torch/one_hot_categorical.py
```diff
@@ -0,0 +1,44 @@
...
+        torch_dist = torch.distributions.Categorical(ps)  # TODO switch to OneHotCategorical
+        x_shape = ps.shape
+        event_dim = 1
+        super(OneHotCategorical, self).__init__(torch_dist, x_shape, event_dim, *args, **kwargs)
+
+    def sample(self):
+        ps = self.torch_dist.probs.data
+        zero = ps.new(self._sample_shape + ps.shape).zero_()
+        indices = super(OneHotCategorical, self).sample().data
+        if indices.dim() < zero.dim():
```
neerajprad: Is this for when we have scalar support in PyTorch? `sample` currently should return a non-zero dim tensor/variable.

fritzo: I believe this is to compensate for Pyro distributions adding an extra dim in all cases (for consistency), whereas I believe PyTorch currently adds an extra dim only to tensors that would be scalars but can't yet be. But honestly I just type random stuff until tests pass 😊

tests/infer/test_inference.py
```diff
@@ -169,26 +160,29 @@ def setUp(self):
...
+        data_sum = self.data.sum(0)
+        self.alpha_n = self.alpha0 + data_sum  # posterior alpha
         self.beta_n = self.beta0 + \
             Variable(torch.Tensor([self.n_data]))  # posterior beta
         self.log_alpha_n = torch.log(self.alpha_n)
         self.log_beta_n = torch.log(self.beta_n)
 
+    @pytest.mark.skipif(not dist.gamma.reparameterized, reason='not implemented')
+    def test_elbo_reparameterized(self):
+        self.do_elbo_test(True, 10000)
```
martinjankowiak: did you change any of these step counts?

tests/infer/test_inference.py
```diff
@@ -304,18 +307,16 @@ def guide():
...
             alpha_q, beta_q = torch.exp(alpha_q_log), torch.exp(beta_q_log)
-            pyro.sample("p_latent", dist.beta, alpha_q, beta_q)
-            pyro.map_data("aaa", self.data, lambda i, x: None, batch_size=self.batch_size)
+            pyro.sample("p_latent", beta, alpha_q, beta_q)
 
         adam = optim.Adam({"lr": .001, "betas": (0.97, 0.999)})
         svi = SVI(model, guide, adam, loss="ELBO", trace_graph=False)
 
-        for k in range(10001):
+        for k in range(n_steps):
```
fritzo: @martinjankowiak The only change I made to step counts was to change this from 10001 to 10000. What was the extra 1 for?

tests/infer/test_inference.py
fritzo: see below, the only change was a 10001 -> 10000 in another test

tests/infer/test_valid_models.py
```diff
@@ -487,6 +487,8 @@ def guide():
 
 
 def test_enum_discrete_global_local_error():
+    if dist.USE_TORCH_DISTRIBUTIONS:
+        pytest.xfail(reason="torch Bernoulli is too permissive?")
```
neerajprad: Is this failing because of some broadcasting issue?

tests/infer/test_inference.py
martinjankowiak: nothing important just checking. the 1 was so it'd print on the last step.... (so irrelevant)

tests/infer/test_valid_models.py
fritzo: I believe so. This is a failing death test, so the failure is due to **lack** of an error being raised. Pyro's old distributions did not allow broadcasting and PyTorch does. I think it's safe to put this off until just before release.

neerajprad: Sounds good! Just want to know all broadcasting related failures / unexpected outcomes, so that if there is a need to revisit it on the PyTorch side, we can do that since its still early days.

fritzo:
Thanks fore reviewing!

PR #650, opened 2017-12-22
Gaussian Process tutorial

fehiepsi:
This pull request is a WIP tutorial on how to create a (simple) GP regression model using pyro. Definitely, there should be added more explanations, comments, plots,... Any comments are really helpful for me to complete it.

- [x] Add an introduction, formulas, remarks, comments to code.
- [x] Explain how to define a `Distribution`. (skip as suggested by @fritzo)
- [x] Explain how to use `random_module` primitive. (skip)
- [x] Explain Pyro's `model-guide` + `SVI` approach. (skip)
- [x] Implement MAP inference.
- [x] Simplify the tutorial after moving some parts to `pyro.contrib.gp`.


fritzo:
cc @ysaatchi @karalets

fritzo:
If you separate the `MultivariateNormalTri` and `GPRegressor` into a separate PR, then you can avoid explaining some of the coding details in this PR (`Distributions`, `nn.random_module`) and focus more attention on the interesting parts of Gaussian Processes.

fehiepsi:
@fritzo Thank you for your suggestions! `contrib.gp` is a good place to start making a skeleton for GP. I will try to finish this tutorial first. After my vacation (next week), I will focus on that.

fritzo:
@fehiepsi If it's easier for you, it would be fine to split this up into two PRs by:
1. update this tutorial to use @dwd31415's `MultivariateNormal` which is now merged;
2. merge this PR
3. factor out `GaussianProcess` into `pyro.congrib.gp` in a second PR.

fehiepsi:
@fritzo : I am on a vacation and will return in a week. Your suggestion is good for me. I will follow it when I come back to my home.

fehiepsi:
@fritzo I have updated the tutorial after moving GPR model and RBF kernel to `pyro.contrib.gp` and using the MultivariateNormal distribution which is currently implemented.

pyro/contrib/gp/kernels/kernel.py
```diff
@@ -0,0 +1,49 @@
...
+        super(Kernel, self).__init__()
+        self.active_dims = active_dims
+        self.name = name
+
+    def K(self, X, Z=None):
+        """
+        Calculate covariance matrix of inputs on active dimensionals.
+
+        :param torch.autograd.Variable X: A 2D tensor of size `N x input_dim`.
+        :param torch.autograd.Variable X2: A 2D tensor of size `N x input_dim`.
```
fritzo: replace `X2` with `Z` here and below

pyro/contrib/gp/kernels/rbf.py
```diff
@@ -0,0 +1,33 @@
...
+
+import torch
+from torch.nn import Parameter
+
+from .kernel import Kernel
+
+
+class RBF(Kernel):
+    """
+    Implementation of RBF kernel.
```
fritzo: Could you expand this to "Radial Basis Function kernel"?

.gitignore
```diff
@@ -1,3 +1,4 @@
+# temp
```
fritzo: nit: remove this line

pyro/contrib/gp/models/gpr.py
```diff
@@ -0,0 +1,63 @@
...
+        self.kernel = kernel
+        # TODO: define noise as a nn.Module, so we can train/set prior to it
+        self.noise = Variable(noise.type_as(X.data))
+        self.priors = priors
+
+    def model(self):
+        kernel_fn = pyro.random_module(self.kernel.name, self.kernel, self.priors)
+        kernel = kernel_fn()
+        K = kernel.K(self.X) + self.noise.repeat(self.input_dim).diag()
+        zero_loc = Variable(torch.zeros(self.input_dim).type_as(K.data))
```
fritzo: I think to get correct GPU device placement you'll need to
```py
zero_loc = K.new([0]).expand(self.input_dim)
```

fehiepsi: I see no problem with using `type_as` method (in pytorch ver 3). However, I just have 1 GPU to test. Did you mean that this will fail for multi-GPU?

fritzo: Yes, `K.new(...)` ensures that the result is on the same GPU as `K`, whereas I believe `.type_as()` only ensures that the result is on some GPU. I only learned this recently, and we still have a few bugs in Pyro relating to GPU placement.

fritzo: Now I try to use `.new()` whenever I create a new tensor.

fehiepsi: Aha, thank you a lot for the tip! Will change soon.

fehiepsi:
@fritzo Could you please remove the label "awaiting response" so I can get reviews to make necessary changes?

pyro/contrib/gp/kernels/kernel.py
```diff
@@ -0,0 +1,49 @@
...
+    Every inherited class should implement the forward pass which
+        take inputs X, X2 and return their covariance matrix.
+    """
+
+    def __init__(self, active_dims=None, name=None):
+        super(Kernel, self).__init__()
+        self.active_dims = active_dims
+        self.name = name
+
+    def K(self, X, Z=None):
```
fritzo: nit: Do you think this method would be better named `__call__`?

pyro/contrib/gp/models/gpr.py
```diff
@@ -0,0 +1,63 @@
...
+
+import pyro
+import pyro.distributions as dist
+
+
+class GPRegression(nn.Module):
+    """
+    Gaussian Process regression module.
+    """
+    def __init__(self, X, y, kernel, noise=torch.ones(1), priors={}):
```
fritzo: nit: It is bit dangerous to use mutable values for default arguments, since all instances then share these mutable values. Prefer `None` with `is None` checks:
```py
def __init__(self, X, y, kernel, noise=torch.ones(1), priors={}):
   if noise is None:
        noise = X.new([1])  # using .new() also ensure noise is on correct cuda device
   if priors is None:
        priors = {}
```

pyro/contrib/gp/models/gpr.py
```diff
@@ -0,0 +1,63 @@
...
+from torch.autograd import Variable
+import torch.nn as nn
+
+import pyro
+import pyro.distributions as dist
+
+
+class GPRegression(nn.Module):
+    """
+    Gaussian Process regression module.
```
fritzo: Could you add `:param:` descriptions for the inputs? Something like
```
Gaussian Process regression module.

:param torch.autograd.Variable X: A tensor of inputs.
:param torch.autograd.Variable y: A tensor of outputs for training.
:param pyro.gp.kernel.Kernel kernel: A Pyro kernel object.
:param torch.Tensor noise: An optional noise tensor.
:param dict priors: A mapping from ??? to priors.
```

fritzo:
@karalets Could you please review the high-level approach? I've already reviewed for software details.

pyro/contrib/gp/kernels/kernel.py
fehiepsi: Yes, it is unnecessary right now. So I remove it. From now on, each forward method in derived kernels should include `_slice_X` statements.

pyro/contrib/gp/kernels/rbf.py
```diff
@@ -0,0 +1,31 @@
...
+
+class RBF(Kernel):
+    """
+    Implementation of Radial Basis Function kernel.
+    """
+
+    def __init__(self, variance=torch.ones(1), lengthscale=torch.ones(1), active_dims=None, name="RBF"):
+        super(RBF, self).__init__(active_dims=active_dims, name=name)
+        self.variance = Parameter(variance)
+        self.lengthscale = Parameter(lengthscale)
```
ysaatchi: Need to add a description of how lengthscale works, typically you have one independent lengthscale per dimension, but you are assuming that lengthscales across all dimensions are the same (?)

pyro/contrib/gp/models/gpr.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    def __init__(self, X, y, kernel, noise=None, priors=None):
+        super(GPRegression, self).__init__()
+        self.X = X
+        self.y = y
+        self.input_dim = X.size(0)
+        self.kernel = kernel
+        # TODO: define noise as a nn.Module, so we can train/set prior to it
+        if noise is None:
+            self.noise = Variable(X.data.new([1]))
```
ysaatchi: What does this do? Needs a comment

pyro/contrib/gp/models/gpr.py
```diff
@@ -0,0 +1,73 @@
...
+    :param torch.Tensor noise: An optional noise tensor.
+    :param dict priors: A mapping from kernel parameter's names to priors.
+    """
+    def __init__(self, X, y, kernel, noise=None, priors=None):
+        super(GPRegression, self).__init__()
+        self.X = X
+        self.y = y
+        self.input_dim = X.size(0)
+        self.kernel = kernel
+        # TODO: define noise as a nn.Module, so we can train/set prior to it
```
ysaatchi: Not the best idea, you should define noise as a *kernel* with its own hypers and optimize it that way

pyro/contrib/gp/models/gpr.py
```diff
@@ -0,0 +1,73 @@
...
+            self.priors = {}
+
+    def model(self):
+        kernel_fn = pyro.random_module(self.kernel.name, self.kernel, self.priors)
+        kernel = kernel_fn()
+        K = kernel(self.X) + self.noise.repeat(self.input_dim).diag()
+        zero_loc = Variable(K.data.new([0]).expand(self.input_dim))
+        pyro.sample("f", dist.MultivariateNormal(zero_loc, K), obs=self.y)
+
+    def guide(self):
```
ysaatchi: What is the purpose of the guide in this context? It seems like you are doing inference in forward(), so what is the point of the guide?

pyro/contrib/gp/models/gpr.py
```diff
@@ -0,0 +1,73 @@
...
+            assert ValueError("Train data and test data should have the same feature sizes.")
+        if Z.dim() == 1:
+            Z = Z.unsqueeze(1)
+        kernel = self.guide()
+        K = kernel(self.X) + self.noise.repeat(self.input_dim).diag()
+        K_xz = kernel(self.X, Z)
+        K_zx = K_xz.t()
+        K_zz = kernel(Z)
+        loc = K_zx.matmul(self.y.gesv(K)[0]).squeeze(1)
+        covariance_matrix = K_zz - K_zx.matmul(K_xz.gesv(K)[0])
```
ysaatchi: This is very inefficient (calling gesv twice on an NxN matrix), see GPML book (Gaussian Processes for Machine Learning) for correct pseudocode for doing this.

ysaatchi: Out of interest, does gesv play well with autograd -- quite cool if so :)

pyro/contrib/gp/models/gpr.py
fehiepsi: Originally, I want to put some constraints on parameters but it needs to write a wrapper of `nn.Parameter` (support transform methods). Then I found that setting priors and using guide for MAP inference might be a simpler idea.  Do you find a better way of using the guide?

pyro/contrib/gp/models/gpr.py
fehiepsi: Noise plays the role of Gaussian likelihood. I intend to add Likelihood, Mean function modules in a future pull request. So temporarily, I let it be constant. See #681 for a plan I have in mind. The pull request is served for the purpose of simplifying the original tutorial code.

pyro/contrib/gp/models/gpr.py
fehiepsi: Yes, I agree!

pyro/contrib/gp/kernels/rbf.py
fehiepsi: @ysaatchi You are right (originally, I just wrote this module for 1-dimensional data). We need to refactor it a bit. I will solve it by introducing `input_dim` parameter (similar to GPy or GPflow) so we can set correct shape for initial lengthscale.

pyro/contrib/gp/models/gpr.py
fehiepsi: @ysaatchi Originally, I put `noise` as hyperparameter and use Cholesky decomposition. Then when `noise` is small, the Lapack error "the leading minor of order ... is not positive definite" annoyed me. In addition, somehow, I find `torch.trtrs` is not stable (https://github.com/pytorch/pytorch/issues/4296). So I use `gesv` instead. Of course, these problems might come from some bugs in my code at that time.

Anyway, using `gesv` might be not a good way, so I will use Cholesky decomposition again. 

p/s: `gesv` supports autograd, but not supports batch yet. :)

fritzo:
@fehiepsi We plan to merge this early next week. We're about to do a big refactoring and I want to make sure your PR can come along for the ride.

fehiepsi:
@fritzo @ysaatchi I have made several changes reflect your reviews. For the change from `noise` to Gaussian Likelihood (a nn.Module), I think it is better to do in another pull request. My main concern is not about its implementation, but about naming: we should have a good way to distinguish parameters of kernels and likelihoods/mean_functions when setting priors to them.

fehiepsi:
@ysaatchi I make a benchmark to compare the following three methods on 1000x1000 matrices.
```python
def method1(L, K):
    return K.t().matmul(K.potrs(L, upper=False))

def method2(L, K):
    v = K.trtrs(L, upper=False)[0]
    return v.t().matmul(v)

def method3(L, K):
    v = L.inverse().matmul(K)
    return v.t().matmul(v)
```
On my CPU: method2 is fastest (10ms), method1 is slower (13ms), method3 is slowest (20ms).
On my GPU: method1 is much faster than method3 (2ms comparing to 10ms), method2 throws an error.

The drawbacks of method2 are: `torch.trtrs()` does not support cuda tensor. In addition, it does not support gradients. So I think that using method1 is the best option for GPR.

fehiepsi:
@fritzo I think that calling `K.inverse()` is better than `L.inverse()`. `L` is just helpful when it is used together with `torch.trtrs()`. So in my implementation, I choose the method1 among the following 4 methods (previously I chose method2, but @ysaatchi  points out that it is ineffective to call `torch.gesv` two times on the same kernel matrix`). Anyway, I think that this is ready to merge.
```
def method1(K, A, y):
    K_inv = K.inverse()
    A_t = A.t()
    loc = A_t.matmul(K_inv.matmul(y))
    scale = A_t.matmul(K_inv.matmul(A))
    return loc, scale
    
def method2(K, A, y):
    A_t = A.t()
    loc = A_t.matmul(y.gesv(K)[0])
    scale = A_t.matmul(A.gesv(K)[0])
    return loc, scale

def method3(K, A, y):
    L = K.potrf(upper=False)
    L_inv = L.inverse()
    L_inv_A = L_inv.matmul(A)
    L_inv_y = L_inv.matmul(y)
    L_inv_A_t = L_inv_A.t()
    loc = L_inv_A_t.matmul(L_inv_y)
    scale = L_inv_A_t.matmul(L_inv_A)
    return loc, scale
    
def method4(K, A, y):
    L = K.potrf(upper=False)
    L_inv_A = A.trtrs(L, upper=False)[0]
    L_inv_y = y.trtrs(L, upper=False)[0]
    L_inv_A_t = L_inv_A.t()
    loc = L_inv_A_t.matmul(L_inv_y)
    scale = L_inv_A_t.matmul(L_inv_A)
    return loc, scale
```

fritzo:
Hi @fehiepsi I'm going to merge this now so it can follow our refactoring work, but feel free to keep submitting updates in follow-up PRs. Thanks for contributing this!

martinjankowiak:
@fehiepsi is it ok if i submit a PR to clean up the language/organization of this tutorial a bit to make it a bit easier to follow?

fehiepsi:
@martinjankowiak Sure, it would be great! Given that the API for GP is stable now (current PRs does not affect the API), it is a good time to revise it. I am happy to see the changes from you.

PR #655, opened 2017-12-28
Empirical distribution

karalets:
Here we add (with @fritzo ) a first version of an empirical distribution adapted from `abstract_infer.Histogram` and a test. 

## Remaining Tasks
- [x] do a test that will work with objects
- [x] add documentation
- [ ] automatically create logits from dataset
- [ ] check for replicates and increase counts on probs


fritzo:
Closing since this is very out-of-date.

Issue #63, opened 2017-07-21
Variance reduction for reparametrized ELBo


ngoodman:
For the reparameterizable case (courtesy @eb8680):
- [Reducing reparameterization gradient variance](https://arxiv.org/abs/1705.07880)
- [Reparameterization gradients through acceptance-rejection algorithms](https://arxiv.org/abs/1610.05683)

jpchen:
@martinjankowiak what is the status of this? i dont know if youve looked back at this since your fancy variance-reduced estimators.. otherwise can we close this in favor of concrete tasks?

cavaunpeu:
FWIW, I'm working on a [PR](https://github.com/blei-lab/edward/pull/819/files) for almost this in Edward. Perhaps I can do a port to Pyro when done. Would be 🌴 💯 ☀️ to be back working in PyTorch land...

fritzo:
@cavaunpeu we've implemented RSVI in #659 which should be merged within the next week or two (just needs some clean-up and tests).

cavaunpeu:
Super!

fritzo:
RSVI and Sticking the Landing are already in Pyro. I'm closing this issue in favor of more targeted issues.

PR #659, opened 2018-01-03
Implement Rejection Sampling Variational Inference

fritzo:
Addresses #63

This implements Rejection Sampling Variational Inference (RSVI) of [Naesseth et al. (2017)](https://arxiv.org/pdf/1610.05683.pdf).

This implements a new class `Rejector` as an abstract base class derived from `Distribution`. The crux to integrating these with Pyro's SVI in `TraceGraph_ELBO` is a new method `Distribution.score_function_term()` that can be overridden to compute `g_cor` in the paper. This method computes a partial score function for a partially reparameterized distribution such that: for an unreparameterized distribution it computes the entire score function, and for a fully reparameterized distribution it computes zero.

## Tasks

- [x] integrate with `Trace_ELBO`
- [x] implement `RejectionGamma` example, e.g. see @slinderman's [notebook](https://github.com/blei-lab/ars-reparameterization/blob/master/gamma/demo.ipynb)
- [x] implement @naesseth's shape augmentation, e.g. see [code](https://github.com/blei-lab/ars-reparameterization)
- [x] write convergence tests
- [x] write smoke tests
- [x] clean up PR to remove irrelevant testing stuff
- [x] fix unit tests

This was joint work by @martinjankowiak and @fritzo


fritzo:
cc @rachtsingh @naesseth

slinderman:
Glad to see this incorporated into pyro, @fritzo and @martinjankowiak!   One of the points discussed in the paper is that the score function term tends to be higher variance, and in practice it can often be safely ignored at the cost of a small bias.  

You can also ameliorate it to some extent by "shape augmentation," i.e. introducing auxiliary uniform r.v.'s in order to increase the shape parameter and thereby increase the acceptance probability.  I did not implement this in the demo notebook, but we used it in our paper.  The code for that is in this repo: https://github.com/blei-lab/ars-reparameterization and @naesseth can probably provide more details on its implementation.

fritzo:
@martinjankowiak This is almost done, there's just one more test to fix.

Could you please review the changes to `Rejector`? I've simplified and added a `log_scale` argument that is used to correctly compute the `.batch_log_pdf()` scale.

martinjankowiak:
cool awesome. i'll do so in the morning when i have a fresh brain

pyro/distributions/distribution.py
```diff
@@ -187,6 +191,27 @@ def batch_log_pdf(self, x, *args, **kwargs):
...
+        distributions should override this method to compute correct
+        `.score_function` and `.entropy_term` parts.
+
+        :param torch.autograd.Variable x: A single value or batch of values.
+        :return: A `ScoreParts` object containing parts of the ELBO estimator.
+        :rtype: ScoreParts
+        """
+        log_pdf = self.batch_log_pdf(x, *args, **kwargs)
+        if self.reparameterized:
+            return ScoreParts(log_pdf, 0, log_pdf)
```
martinjankowiak: maybe explicit names in namedtuple here?

fritzo:
@martinjankowiak I believe I've fixed `Rejector` (sorry for the earlier bug!), but `ShapeAugmentedGamma` gradients still appear to be wrong. Do you think it's worth merging this as-is and debugging `ShapeAugmentedGamma` in a follow-up PR that also adds `ShapeAugmentedBeta` etc?

fritzo:
Ready to merge, finally.

rmehta1987:
@fritzo are there any examples of using the accept-reject dirichlet or gamma?

Issue #656, opened 2017-12-29
Improve Gradients of MultivariateNormal.batch_log_pdf 


dwd31415:
The results of MultivariateNormal.batch_log_pdf can not be differentiated with respect to the covariance matrix yet if normalized=False. I am opening this issue to keep the discussion regrading this issue alive. See #651 for the details on this.

dwd31415:
cc @fritzo

dwd31415:
I have started to implement this using a `autograd.Function` as you suggested. I think I will be able to finish it rather soon. I am posting this code as WIP but it seems to be working already:
```python
class _NormalizationConstant(Function):
    @staticmethod
    def forward(ctx, sigma, sigma_cholesky, inverse, dimension, return_zero):
        ctx.save_for_backward(inverse)
        results = torch.Tensor(
            [torch.log(sigma_cholesky.diag()).sum() + (dimension / 2) * np.log(2 * np.pi)]).type_as(sigma_cholesky)
        return results if not return_zero else torch.zeros(1).type_as(sigma_cholesky)

    @staticmethod
    def backward(ctx, grad_output):
        inverse, = ctx.saved_variables
        grad = 2 * inverse - torch.diag(torch.diag(inverse))
        return grad_output * grad, None, None, None, None

```
I figured that passing the inverse would be a good idea since it is always being computed right now anyway and the procedure is rather expensive, so doing it twice would really be a waste of time.
However differentiation wrt. to the Cholesky  decomposition does not work yet, but adding that gradient should not really be much of a problem (or so I think at least). 
If you agree with this approach I will open a pull request as soon as I have implemented a gradient wrt. to `sigma_cholesky`.

fritzo:
Your implementation seems reasonable to me. However I would recommend branching on `return_zero` outside of `_NormalizationConstant`, since I would guess that `torch.log(sigma_cholesky.diag()).sum() ...` leads to a more numerically stable gradient than the on you compute in `_NormalizationConstant`.

dwd31415:
I don’t exactly follow, why should there be any difference in results? If I see it correctly the two expressions are exactly the same.

fritzo:
What I was imagining was
```py
class MultivariateNormal(Distribution):
    def batch_log_pdf(self):
        ...
        if self.normalized:
            # Gradient is computed automatically by PyTorch.
            normalization_factor = (torch.log(self.sigma_cholesky.diag()).sum() +
[...]

PR #663, opened 2018-01-04
Add Gradients of MultivariateNormal.batch_log_pdf 

dwd31415:
This is a fix for #656

### DIFF
--- a/pyro/distributions/multivariate_normal.py
+++ b/pyro/distributions/multivariate_normal.py
@@ -38,12 +38,17 @@ def matrix_inverse_compat(matrix, matrix_chol):
 
 
 class _NormalizationConstant(Function):
+    """
+    This computes either zero or the true normalization constant depending on normalized,
+    but always computes the true gradient.
+    """
     @staticmethod
-    def forward(ctx, sigma, sigma_cholesky, inverse, dimension, return_zero):
+    def forward(ctx, sigma, sigma_cholesky, inverse, dimension, normalized):
         ctx.save_for_backward(inverse, sigma_cholesky)
-        results = torch.Tensor(
-            [torch.log(sigma_cholesky.diag()).sum() + (dimension / 2) * np.log(2 * np.pi)]).type_as(sigma_cholesky)
-        return results if not return_zero else torch.zeros(1).type_as(sigma_cholesky)
+        if normalized:
+            return torch.log(sigma_cholesky.diag()).sum(-1) + (dimension / 2) * np.log(2 * np.pi)
+        else:
+            return torch.zeros(1).type_as(sigma_cholesky)
 
     @staticmethod
     def backward(ctx, grad_output):
@@ -153,7 +158,7 @@ class MultivariateNormal(Distribution):
         # batch size and the size of the covariance matrix
         sigma_inverse = matrix_inverse_compat(self.sigma, self.sigma_cholesky)
         normalization_factor = _NormalizationConstant.apply(self.sigma, self.sigma_cholesky, sigma_inverse,
-                                                            self.mu.size()[0], not self.normalized)
+                                                            self.mu.size()[0], self.normalized)
         return -(normalization_factor + 0.5 * torch.sum((x - self.mu).unsqueeze(2) * torch.bmm(
             sigma_inverse.expand(batch_size, *self.sigma_cholesky.size()),
             (x - self.mu).unsqueeze(-1)), 1)).view(batch_log_pdf_shape)
--- a/pyro/distributions/multivariate_normal.py
+++ b/pyro/distributions/multivariate_normal.py
@@ -6,6 +6,7 @@ from torch.autograd import Variable
 
 from pyro.distributions.util import copy_docs_from
 from pyro.distributions.distribution import Distribution
+from torch.autograd import Function
 
 
 def potri_compat(var):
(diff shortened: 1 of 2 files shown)


pyro/distributions/multivariate_normal.py
```diff
@@ -29,15 +29,30 @@ def linear_solve_compat(matrix, matrix_chol, y):
 
 def matrix_inverse_compat(matrix, matrix_chol):
     """Computes the inverse of a positive semidefinite square matrix."""
-    assert matrix.requires_grad == matrix_chol.requires_grad
-    if matrix.requires_grad:
+    if matrix.requires_grad or matrix_chol.requires_grad:
```
fritzo: Could you help me understand under what conditions the assertion would fail?

tests/distributions/test_multivariate_normal.py
```diff
@@ -0,0 +1,52 @@
...
+        self.sample = Variable(torch.randn(N))
+
+        self.cholesky_mv_normalized = MultivariateNormal(self.mu, scale_tril=self.L, normalized=True)
+        self.cholesky_mv = MultivariateNormal(self.mu, scale_tril=self.L, normalized=False)
+
+        self.full_mv_normalized = MultivariateNormal(self.mu, self.sigma, normalized=True)
+        self.full_mv = MultivariateNormal(self.mu, self.sigma, normalized=False)
+
+    def test_log_pdf_gradients_cholesky(self):
+        self.cholesky_mv.log_pdf(self.sample).backward()
```
fritzo: I would recommend using `torch.autograd.grad()` in these tests. We're finding it much more convenient than `.backward()`:
```py
from torch.autograd import grad

class TestMultivariateNormal(TestCase):
    ...
    def test_log_pdf_gradiennts_cholesky(self):
        grad1 = grad([self.cholesky_mv.log_pdf(self.sample)], [self.L])[0].data
        grad2 = grad([self.cholesky_mv_normalized.log_pdf(self.sample)], [self.L])[0].data
        assert_equal(grad1, grad2)
  ```

File: pyro/distributions/multivariate_normal.py [hunk 3]
```diff
@@ -29,15 +29,30 @@ def linear_solve_compat(matrix, matrix_chol, y):
...
         return potri_compat(matrix_chol)
 
 
+class _NormalizationConstant(Function):
+    @staticmethod
+    def forward(ctx, sigma, sigma_cholesky, inverse, dimension, return_zero):
+        ctx.save_for_backward(inverse, sigma_cholesky)
+        results = torch.Tensor(
+            [torch.log(sigma_cholesky.diag()).sum() + (dimension / 2) * np.log(2 * np.pi)]).type_as(sigma_cholesky)
+        return results if not return_zero else torch.zeros(1).type_as(sigma_cholesky)
```
fritzo: I would recommend (1) preserving the parameter name `normalized` throughout, and (2) avoiding unnecessary computation
```py
if not normalized:
    return torch.zeros(1).type_as(sigma_cholesky)
return sigma_cholesky.diag().log().sum() + ...  # no need for .type_as() here
```

dwd31415: I will split the if statement, but `return sigma_cholesky.diag().log().sum()` sadly does not work. The `.sum()` does not return a Tensor but a float (or whatever the type is) and pytorch won't accept that return type. That is the reason for the `torch.Tensor(...).type_as`.

pyro/distributions/multivariate_normal.py
```diff
@@ -29,15 +29,30 @@ def linear_solve_compat(matrix, matrix_chol, y):
...
-    if matrix.requires_grad:
+    if matrix.requires_grad or matrix_chol.requires_grad:
         # If derivatives are required, use the more expensive inverse.
         return torch.inverse(matrix)
     else:
         # Use the cheaper Cholesky based potri.
         return potri_compat(matrix_chol)
 
 
+class _NormalizationConstant(Function):
```
fritzo: nit: It would be nice to have a one-line comment explaining the purpose of this class, something like:
> This computes either zero or the true normalization constant depending on `normalized`, but always computes the true tradient.

pyro/distributions/multivariate_normal.py
dwd31415: The problem here is potrf_compat on PyTorch < 0.3.0: In that case `Variable(torch.potrf(sigma.data))` is returned and if `sigma.requires_grad` was true then they don't match up any more. The unit tests didn't catch this until now because they did not need any gradients.

pyro/distributions/multivariate_normal.py
fritzo: Gah, I forgot about that `.sum()` behavior, but I'm not sure your version supports batching. What about this?
```py
sigma_cholesky.diag().log().sum(0) + ...
```

tests/distributions/test_multivariate_normal.py
dwd31415: Thanks, it is much cleaner like this.

pyro/distributions/multivariate_normal.py
fritzo: Hmm, even `.diag()` doesn't support batching...

dwd31415: That seems to work.

dwd31415: Yeah, but this whole thing does not support batched parameters, so that does not really matter for now.

PR #666, opened 2018-01-05
Move Trace.log_pdf(), .batch_log_pdf() methods to trace_*() functions

### LINKED ISSUE #661: Move .log_pdf(), .batch_log_pdf() methods out of Trace class
Pyro's [Trace](https://github.com/uber/pyro/blob/dev/pyro/poutine/trace.py) data structure currently includes logic to compute `.batch_log_pdf()` and `.log_pdf()` methods. These methods should be moved to free-standing functions ideally in pyro.infer.

## Why?

First, the `.log_pdf()` methods have been a source of many bugs involving batching, broadcasting, device placement, and type errors (`float` vs `Variable`). The logic in `.log_pdf()` and `.batch_log_pdf()` should be moved out to helpers so that `Trace` objects are merely containers where inference quantities can be cached (even that is debatable).

Second, `.log_pdf()` methods are blocking implementation of RSVI in #659 which needs an additional `.score_function_term()` method whose result is cached just like `log_pdf`. It seems awkward and intrusive to bake this logic into `Trace`, but it seems reasonable to add logic to a free standing function like `pyro.infer.util.compute_score_terms(trace, batch_shape_info)`.

fritzo:
Fixes #661

### DIFF
--- a/pyro/infer/enum.py
+++ b/pyro/infer/enum.py
@@ -3,11 +3,11 @@ from __future__ import absolute_import, division, print_function
 import functools
 
 import torch
-from torch.autograd import Variable
-
 from pyro import poutine, util
+from pyro.infer.util import trace_batch_log_pdf
 from pyro.poutine.trace import Trace
 from six.moves.queue import LifoQueue
+from torch.autograd import Variable
 
 
 def site_is_discrete(name, site):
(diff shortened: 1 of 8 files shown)


fritzo:
Closing in favor of site-specific methods.

PR #678, opened 2018-01-13
Test tutorials on travis

fritzo:
This adds support for testing some of our tutorials on travis.

## Why?

We're changing Pyro and we want to avoid bit rot of tutorials.

## How?

This uses a kludgey mechanism of defining a boolean `smoke_test = 'CI' in os.environ` in each tutorial. This is a little ugly, so we might something cleaner before Pyro 0.2 release, e.g. replacing all strings of the form `num_iters = [0-9]*` with the string `num_iters = 1`.


fritzo:
@neerajprad looks like build is failing due to wheel issues

neerajprad:
@fritzo - I had updated the wheel files for Multinomial, but forgot to overrride the permissions to public. Restarted the tests, it should work now.

tutorial/source/gmm.ipynb
```diff
@@ -29,7 +30,8 @@
     "from pyro.optim import Adam\n",
     "from pyro.infer.svi import SVI\n",
     "from pyro.distributions import categorical, normal\n",
-    "softmax = Softmax(dim=-1)"
+    "softmax = Softmax(dim=-1)\n",
+    "smoke_test = ('CI' in os.environ)"
```
neerajprad: Were the changes in this file due to just running the `nbval` command?

null-a:
This seems reasonable to me. Once #680 is merged the AIR tutorial should be working again and could be added to these smoke tests. I could do this as a follow-up PR if you like.

fritzo:
@neerajprad sorry it took me so long to fix this PR. We'll need this before **The Big Refactoring** next week.

PR #679, opened 2018-01-14
Wrap PyTorch multinomial distribution for use in Pyro

fritzo:


pyro/distributions/torch/multinomial.py
```diff
@@ -0,0 +1,36 @@
...
+
+@copy_docs_from(_Multinomial)
+class Multinomial(TorchDistribution):
+    enumerable = True
+
+    def __init__(self, ps, n, *args, **kwargs):
+        if isinstance(n, Variable):
+            n = n.data
+        if not isinstance(n, numbers.Number):
+            if n.max() != n.min():
```
neerajprad: Do you think we should change our API to have `n` be `int` as well?

fritzo: That seems reasonable. I'd like to switch all our distributions to look more like PyTorch distributions. I'd also like to eliminate `.log_pdf()` and rename `.batch_log_pdf()` to `.log_prob()`.

neerajprad: I want to start making these changes on the Pyro side too. Removing all Pyro distributions (that do not have a corresponding PyTorch implementation) will probably help make this faster. What do you think?

fritzo: I'm a bit wary of removing all Pyro distributions. I think we may want to keep Pyro versions of `Delta`, `Empirical`, and `MultivariateNormal` long-term. The other Pyro-only distributions are currently being worked on by various PyTorch folks, so I think it's safe to remove those.

PR #681, opened 2018-01-15
Add skeleton for Gaussian Process module

fehiepsi:
This pull request implements a simple version of Gaussian Process Regression and RBF kernel under the `pyro.contrib.gp` module. Comparing to other GP packages, it lacks a huge number of features. So things to do in the future might be
+ Add more kernels, support adding or multiplying kernels together.
+ Add likelihood, mean_function modules.
+ Add other models (such as classification, Poisson regression, dimensional reduction).
+ Support spare methods (using inducing points).
+ Support multidimensional output learning.
+ ... (due to the lack of my knowledge, I can't name them here).


fehiepsi:
This pull request will be reviewed at #650, so I close.

PR #682, opened 2018-01-16
add adagradrmsprop, a mash-up optimizer

martinjankowiak:
i needed to implement the optimizer used in 

'Automatic Differentiation Variational Inference', Alp Kucukelbir,
Dustin Tran, Rajesh Ranganath, Andrew Gelman, David M. Blei

to replicate some experiments. the optimizer is a mash-up of adagrad and rmsprop.
so i thought i might as well clean up the code a bit and add it to pyro.

@karalets perhaps you might like try plopping this into some of your experiments?

### DIFF
--- a/pyro/optim/adagrad_rmsprop.py
+++ b/pyro/optim/adagrad_rmsprop.py
@@ -1,0 +1,75 @@
+from __future__ import absolute_import, division, print_function
+
+import torch
+from torch.optim.optimizer import Optimizer
+
+
+class AdagradRMSProp(Optimizer):
+    """
+    Implements a mash-up of the Adagrad algorithm and RMSProp. See ref. [1].
+
+    References:
+    [1] 'Automatic Differentiation Variational Inference', Alp Kucukelbir,
+        Dustin Tran, Rajesh Ranganath, Andrew Gelman, David M. Blei
+    [2] 'Lecture 6.5 RmsProp: Divide the gradient by a running average
+        of its recent magnitude', Tieleman, T. and Hinton, G.,
+        COURSERA: Neural Networks for Machine Learning.
+    [3] 'Adaptive subgradient methods for online learning and stochastic optimization',
+        Duchi, John, Hazan, E and Singer, Y.
+
+    Arguments:
+        params (iterable): iterable of parameters to optimize or dicts defining
+            parameter groups
+        eta (float, optional): sets the step size scale (default: 1.0)
+        t (float, optional): momentum parameter (default: 0.1)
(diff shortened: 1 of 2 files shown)


eb8680:
Seems fine, does it need tests?  Are there any optimizer tests from PyTorch we should borrow from?

pyro/optim/adagrad_rmsprop.py
```diff
@@ -0,0 +1,75 @@
+from __future__ import absolute_import, division, print_function
+
+import torch
+from torch.optim.optimizer import Optimizer
+
+
+class AdagradRMSProp(Optimizer):
+    """
+    Implements a mash-up of the Adagrad algorithm and RMSProp. See ref. [1].
```
eb8680: Is it worth adding the update equation to the docstring?

martinjankowiak:
pytorch has some tests. although i think it'd be a wasted engineering effort on our end to replicate. for these kinds of optimizers i figure we should do one of the following:
1. put a red flag in the docs warning that it's not fully vetted code
2. try to migrate to pytorch
3. put it contrib (basically another version of #1)

pyro/optim/adagrad_rmsprop.py
martinjankowiak: instead i'll put a more direct reference in the docstring

pyro/optim/adagrad_rmsprop.py
```diff
@@ -0,0 +1,77 @@
...
+        URL: https://arxiv.org/abs/1603.00788
+    [2] 'Lecture 6.5 RmsProp: Divide the gradient by a running average
+        of its recent magnitude', Tieleman, T. and Hinton, G.,
+        COURSERA: Neural Networks for Machine Learning.
+    [3] 'Adaptive subgradient methods for online learning and stochastic optimization',
+        Duchi, John, Hazan, E and Singer, Y.
+
+    Arguments:
+        params (iterable): iterable of parameters to optimize or dicts defining
+            parameter groups
```
jpchen: arent the docs all still in sphinx format? i dont think this will render properly

pyro/optim/adagrad_rmsprop.py
```diff
@@ -0,0 +1,77 @@
...
+            for p in group['params']:
+                state = self.state[p]
+                state['sum'].share_memory_()
+
+    def step(self, closure=None):
+        """Performs a single optimization step.
+
+        Arguments:
+            closure (callable, optional): A closure that reevaluates the model
+                and returns the loss.
```
jpchen: same comment as above

Issue #638, opened 2017-12-19
Master Task - Update broadcasting and batching semantics for Pyro Distributions


neerajprad:
We have made some major changes to the `torch.distributions` API, and would like to make this available in Pyro in its full generality. However, the current Pyro shape semantics are incompatible with this new API. 

Some of these incompatibilities are as follows:
 - The `batch` and `sample` dimensions in Pyro are conflated, which makes it confusing for the user to determine the shape of the distribution parameters to be supplied, i.e. it is common to provide identical distribution parameters only to impose a shape on the `sample`.
 - Any batching (i.e. `sample` dimensions) is restricted to a single dimension, and does not support arbitrary shaped sample sizes.
 - In the absence of scalar support, `batch_log_pdf` upcasts the resulting score by appending an additional `(1,)` rightmost dimension. In the case of PyTorch, we have tried to localize this behavior to only upcast when any dimension reduction results in a 0 sized tensor. This makes the migration to `torch.Scalar` simpler.

By the next PyTorch release, we would like to have this rich and updated API available to Pyro. @fritzo and I decided that it will simplify the transition if Pyro's `dev` branch only tries to maintain compatibility with `PyTorch's` master instead of the `release` version which it does now (refer to Phase 1.5).

The transition plan is as follows:

### Phase 1 - Changes in PyTorch:
- [x] Create implementations for all the Pyro distributions in PyTorch. These are being tracked at https://github.com/probtorch/pytorch/projects/1
- [x] #606 Build light wrappers around `pyro.distributions` in Pyro which preserve our current shape semantics.
- [x] #684 Delete Pyro distributions and distribution tests. Migrate all major testing to PyTorch.

### Phase 1.5 - Infra / internal changes to help with migration:
Can be done simultaneously with Phase 1. 
- [x] Create a lightweight torch wheel that can be periodically updated to reflect PyTorch master. This will be useful for CI testing of Pyro dev against PyTorch master. - Refer to [pytorch#4178](https://github.com/pytorch/pytorch/issues/4178), #670 
- [x] Make tutorials testable - tutorials need to be written using a format that allows for easy and fast testing of API related changes.
- [ ] Add versioning to tutorials - Given that some of our changes will not be backwards compatible, we should version our tutorials to make it easy for our users to refer to the relevant version depending on their version of Pyro install. (Ref. #639)
 
### Phase 2 - Internal change to Distributions API:
[...]

PR #684, opened 2018-01-17
Remove Pyro distributions migrated to PyTorch

neerajprad:
Addresses #638

This simply removes all the distributions that are already migrated to PyTorch. In subsequent PRs, we will rename the method names to align with `torch.distributions`, and finally change the batch semantics and remove/update most of our existing test suite.


jpchen:
this only removes the ones that have _already been migrated_ correct? so we shouldnt be losing support for any existing distributions in this PR?

neerajprad:
> this only removes the ones that have already been migrated correct? so we shouldnt be losing support for any existing distributions in this PR?

That's right, not yet. :) What changes with this PR is that you need to build PyTorch master to be able to use Pyro's `dev` branch.

neerajprad:
There is some travis CI slowdown, so it might take a while for the tests to start running.

neerajprad:
Should be good to merge once tests pass.

PR #689, opened 2018-01-17
Wrap torch multinomial

fritzo:


fritzo:
oops, this appears to be stale

PR #690, opened 2018-01-18
Adding simple HMC Kernel 

neerajprad:
This cleans up, and reinstates the prior HMC implementation:
 - Most of the math and API was discussed and reviewed in https://github.com/uber/pyro/pull/579.
 - Uses the integrator that was added in https://github.com/uber/pyro/pull/611.
 - A big part of the time was taken up in `trace.copy()`, which is not needed for the implementation. As such, I have tried to eliminate the copy calls. Profiling results attached below.
 
**Testing:** Tests for chained gaussians and logistic regression are added. The same can be adapted for other MCMC algorithms if needed, with some minor modification.


neerajprad:
**Profiling results:**
This is based on the CPU runs for the third test case in `test_hmc_conjugate_gaussian` (See. https://github.com/uber/pyro/pull/690). The python file with the standalone code and the profiling results are included in the zip file.

 - Barebones PyTorch implementation - ~100 seconds (underapprox. as this is restricted to just iterating the integrator loop, exc. the Metropolis step). [file - `profile_barebones_torch.prof`]
 - Old implementation with trace copies and `pyro.distributions` ~ 187 s. [file - `profile_torch_dist_orig.prof`]
 - Current implementation without trace copies and `pyro.distributions` ~ 169 s. [file - profile_wo_trace_copies.prof]
 - Current implementation without trace copies and using `torch.distributions` directly ~ 145 s. [file profile_torch_dist_wo_trace_copies.prof]

The best way to visualize the profiler results is by using `snakeviz`. e.g. `snakeviz profile_torch_dist_orig.prof`.

Given that the overhead of running through the poutine stack (check `_get_trace` method) is around  15%, I am happy with the current numbers. 

A simple improvement that can shave off around 14s (~9%) is by simply reusing distribution instances, i.e. using the object form of distributions `pyro.sample(dist_instance)`, rather than the functional form `pyro.sample(dist.normal, mu=x, sigma=y)`. This is understandable as there is a lot of validation and broadcasting logic in the constructor for `torch.distributions` and those microseconds add up quickly if we initialize a new distribution per sample/score call. cc. @fritzo 

**zipped results**
[Archive.zip](https://github.com/uber/pyro/files/1641306/Archive.zip)

eb8680:
> Given that the overhead of running through the poutine stack (check _get_trace method) is around 15%, I am happy with the current numbers.

What's the root cause of this slowdown?  Can it be improved?

neerajprad:
> What's the root cause of this slowdown? Can it be improved?

I do not see any obvious improvements or steps that can be be eliminated (except for `trace.copy` which I removed). I think this is just the unavoidable overhead of running the model and generating traces through a generic mechanism. You could try to visualize using `snakeviz` in case something catches your eye.

jpchen:
was there anything changed in the core hmc implementation from the PR's you linked? (as far as i can tell the implementation looks the same.)

>A simple improvement that can shave off around 14s ...

interesting, 9% is a lot more than id expect. would this have been different if you used pyro distributions (instance vs functional)?

neerajprad:
> was there anything changed in the core hmc implementation from the PR's you linked? (as far as i can tell the implementation looks the same.)

I think the core should be pretty much the same - this is just wrapping over the updated interface and integrator classes (and has some minor perf fixes, like avoiding copying traces). It has some additional tests that were not there in the early PRs. 

> interesting, 9% is a lot more than id expect. would this have been different if you used pyro distributions (instance vs functional)?

Our internal implementations were anyways less performant (took 20s more). There would definitely be an additional cost to creating new instances in our implementation as well. I have not profiled how much is that difference though (the code is deleted as well), and the magnitude would depend on how much work is happening in the constructors. If there is not much validation / broadcasting happening then creating new instances may not be as expensive. PyTorch distributions have more flexible broadcasting support, and do a bit more work in the constructor. As such reusing those instances makes sense. Separately, we are looking to write the broadcasting logic in C, and that might make creating new instances cheaper.

tests/infer/mcmc/test_hmc.py
```diff
@@ -0,0 +1,176 @@
...
+        num_samples=1200,
+        warmup_steps=300,
+        hmc_params={'step_size': 0.46,
+                    'num_steps': 5},
+        expected_means=[0.20, 0.40, 0.60, 0.80],
+        expected_precs=[1.25, 0.83, 0.83, 1.25],
+        mean_tol=0.06,
+        std_tol=0.06,
+    ),
+    # XXX: Very sensitive to HMC parameters. Biased estimate is obtained
```
fritzo: Are the other tests run during integration tests?

pyro/infer/mcmc/hmc.py
```diff
@@ -0,0 +1,108 @@
...
+        Hamiltonian dynamics. The state at the end of the trajectory is
+        returned as the proposal.
+    """
+
+    def __init__(self, model, step_size=0.5, num_steps=3):
+        self.model = model
+        self.step_size = step_size
+        self.num_steps = num_steps
+        # simulation run attributes - will be set in self._setup at start of run
+        self._t = 0
```
fritzo: I'd recommend calling `self.cleanup()` here rather than duplicating it's logic. That eliminates the need to keep these two pieces of code synchronized. Also consider renaming `cleanup` to `reset` so it makes more sense to call here?

pyro/poutine/trace_poutine.py
```diff
@@ -158,7 +158,10 @@ def get_trace(self, *args, **kwargs):
         Helper method for a very common use case.
         Calls this poutine and returns its trace instead of the function's return value.
         """
+        clone = kwargs.pop('clone', True)
```
fritzo: This makes me a bit uncomfortable, but I'm not sure how to work around so it seems fine for now.

tests/infer/mcmc/test_hmc.py
neerajprad: They are not. I have only kept them so that they can be run locally, if needed. The third one was used for profiling, for instance.

pyro/infer/mcmc/hmc.py
neerajprad: Makes sense. The only thing is that `setup` and `cleanup` are methods in the abstract `TraceKernel` class. I suppose we could just change the interface to have a single `reset` method. I don't currently foresee us needing different methods at this point.

pyro/poutine/trace_poutine.py
neerajprad: Yeah, that's true. Can we just default to returning the current trace, instead of making a defensive copy? It seems to me that it should be safe since we are anyways calling `trace_poutine` right before the return call. cc. @eb8680?

eb8680: The current behavior is better as a default because if you return a reference to `TracePoutine.trace` and then call the `TracePoutine` again, it will silently mutate the trace and may cause other problems.  If you don't want to make a defensive copy it's better to be explicit about the effectful behavior:
```python
tp = poutine.trace(fn)
tp(...)
tr = tp.trace
```

neerajprad: I think it shouldn't be an issue, because this calls self, and we instantiate a new trace object in `__call__`. I will make the change to access the attribute directly, but we should eliminate the `copy` call if it is not needed. Let me know.

PR #692, opened 2018-01-18
Default to n=1 in Multinomial distribution

fritzo:
This sets a default value of 1 for the `n` argument of the Multinomial distribution. This is useful when the distribution appears in `pyro.observe` statements where `n` can be inferred from observations:
```py
pyro.sample("documents", dist.multinomial, probs, obs=word_counts)
```

This change makes it clearer that although `Multinomial.sample()` is only implemented for `n` an integer or constant `Tensor`, `Multinomial.log_pdf()` is implemented for heterogeneous `n` and is useful e.g. in observing word counts in documents of different length.

### DIFF
--- a/pyro/distributions/torch/multinomial.py
+++ b/pyro/distributions/torch/multinomial.py
@@ -15,17 +15,19 @@ class Multinomial(TorchDistribution):
     Multinomial distribution.
 
     Distribution over counts for `n` independent `Categorical(ps)` trials.
+    Note that `n` need not be specified if only :meth:`log_pdf` is used, e.g.
+    if this distribution is used only in :meth:`pyro.observe` statements.
 
     This is often used in conjunction with `torch.nn.Softmax` to ensure
     probabilites `ps` are normalized.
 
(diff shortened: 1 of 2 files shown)


fritzo:
cc @1Reinier

pyro/distributions/torch/multinomial.py
```diff
@@ -15,17 +15,19 @@ class Multinomial(TorchDistribution):
...
     Distribution over counts for `n` independent `Categorical(ps)` trials.
+    Note that `n` need not be specified if only :meth:`log_pdf` is used, e.g.
+    if this distribution is used only in :meth:`pyro.observe` statements.
 
     This is often used in conjunction with `torch.nn.Softmax` to ensure
     probabilites `ps` are normalized.
 
     :param torch.autograd.Variable ps: Probabilities (real). Should be positive
         and should normalized over the rightmost axis.
-    :param int n: Number of trials. Should be positive.
```
jpchen: can you put "Defaults to 1" here?

fritzo:
@martinjankowiak would you please merge this when you get a chance?

PR #693, opened 2018-01-19
Refactor and simplify MultivariateNormal distribution

fritzo:
This refactors and simplifies Pyro's `MultivariateNormal` distribution into a PyTorch-style `TorchMultivariateNormal` distribution plus a wrapper. This makes it easy to swap in a PyTorch `MultivariateNormal` as soon as it is available (as soon as https://github.com/probtorch/pytorch/pull/52 is merged).

This PR is needed to support @fehiepsi 's Gaussian Process tutorial #650 as Pyro migrates to PyTorch distributions.


jpchen:
failing tests though:
```
E           RuntimeError: Lapack Error in potrf : the leading minor of order 2 is not positive definite at /remote/pytorch/aten/src/TH/generic/THTensorLapack.c:617
```

fehiepsi:
@fritzo My GPRegression tutorial runs fine with this pull request. However, for a general case (when I use other likelihoods such as Bernoulli for GPClassification), `torch.potrs` will throw errors `RuntimeError: Lapack Error in potrf : the leading minor of order ... is not positive definite at ...`.

Here is a simple script to replicate the problem
```
import torch

import pyro

K = Variable(torch.rand((1000, 1000))).double()
K = K @ K.t()

y = Variable(torch.zeros(1000)).double()

z = pyro.sample("z", dist.MultivariateNormal(y, K))
```

Of course, the problem is more related to "how to deal with such RuntimeError", rather than this implementation.

fehiepsi:
I think that it is better to raise an issue to discuss, rather than comment here.

PR #694, opened 2018-01-22
Include AIR tutorial in smoke tests.

null-a:
This is an attempt to include the AIR tutorial in new (#678) tutorial smoke tests. I'm not making use of the `smoke_test` flag within the tutorial as none of its cells do much work. This works locally, but I'll keep an eye on travis in case of problems.

### DIFF
--- a/tutorial/source/air.ipynb
+++ b/tutorial/source/air.ipynb
@@ -14,7 +14,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -46,7 +46,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
    "metadata": {
     "keep_output": true
    },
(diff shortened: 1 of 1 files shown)


null-a:
I've added the execution counts to the repo by running the notebook locally, and then running `nbstripout --keep-count`. Let's see if the tests pass with this.

fritzo:
Well I'm not opposed to pinning to an old "working" version of `nbval==0.7` if that's the easiest solution :smile:

fritzo:
Actually I recall being unable to load a tutorial notebook due to a `cell_count=None` error at some point, so `--keep-count` sounds like a good idea.

fritzo:
@null-a Do you want me to merge this as-is, or do you want to add `--keep-count` to the `scrub` command in `Makefile`?

null-a:
@fritzo I don't have a preference, happy to go with whatever seems best to you. If you'd like the `Makefile` updating, I can do it as part of this.

fritzo:
We can do it in a late PR. merging. Thanks for adding this!

null-a:
@fritzo: no worries, thanks for your help.

PR #695, opened 2018-01-23
Refactoring poutines into Messengers and handlers

adscib:
Refactors poutines into messengers that only transform messages and poutines that wrap a computation and contain a messenger.


eb8680:
Some tests related to enumeration and `IndepPoutine` are failing.  I'll investigate.

adscib:
Looks like this was caused by some unnecessary changes I made to the code. Fixed now.

eb8680:
> Looks like this was caused by some unnecessary changes I made to the code. Fixed now.

I thought the behavior you removed (reverting the scale and independence contexts in the second pass) was what we wanted?

adscib:
I think so, but it doesn't matter for now. I would recommend to leave it as is for now and sort it out later as a part of a more thorough effort to systematise how messengers behave.

adscib:
Sweet! Let me know if you want to switch to using greenlets at some point, I'd be happy to code it.

pyro/poutine/escape_poutine.py
```diff
@@ -32,6 +31,27 @@ def _pyro_sample(self, msg):
         """
         if self.escape_fn(msg):
             msg["done"] = True
-            raise NonlocalExit(msg)
-        else:
-            return super(EscapePoutine, self)._pyro_sample(msg)
+            msg["stop"] = True
+
+            def cont(m):
+                # for enumeration, we need a value at the site to get the shape
```
fritzo: Why not just call `.shape()`?

eb8680: I wasn't sure if that was correct, but if so I can change it

pyro/poutine/indep_poutine.py
```diff
@@ -2,40 +2,34 @@
...
-    def _prepare_site(self, msg):
-        """
-        Construct the message that is consumed by ``TracePoutine``;
-        ``cond_indep_stack`` encodes the nested sequence of ``irange`` branches
-        that the site at name is within.
-        """
-        msg["cond_indep_stack"].append(CondIndepStackFrame(self.name, self.counter, self.vectorized))
-        return msg
+    def _process_message(self, msg):
+        msg["cond_indep_stack"].insert(0, CondIndepStackFrame(self.name, self.counter, self.vectorized))
```
fritzo: This erroneously reverses the direction of `cond_indep_stack`. It worries me that this bug was not caught by tests.

fritzo: cc @martinjankowiak

eb8680:
Can you write a test that proves that? I'm pretty sure it reverses both the
order of traversal of the poutine stack and the order of insertion into the
independence stack, so that the independence stack ordering is preserved.

pyro/poutine/indep_poutine.py
fritzo: > it reverses both the order of traversal of the poutine stack and the order of insertion into the
independence stack

Oh, that makes sense, thanks for explaining.

adscib: I actually first just reversed the order of the poutine stack traversal and then your tests caught the problem with the independence stack order so we reversed that too.

PR #698, opened 2018-01-23
Allow for sample_shape argument in distribution.sample

neerajprad:
- Changes the TorchDistribution class to include the `sample_shape` parameter. This deprecates the `batch_size` parameter given to the constructor, and replaces those instances with `sample_shape=(batch_size,)`. 
 - Removes `poisson`, `binomial` and `half_cauchy`. The old unused pyro `beta.py` file is removed too. These will be reinstated once the migration is complete and these implementations are available in PyTorch master (refer to #699).


pyro/distributions/torch_wrapper.py
```diff
@@ -13,13 +13,12 @@ class TorchDistribution(Distribution):
     `torch.distributions.Distribution <http://pytorch.org/docs/master/_modules/torch/distributions.html#Distribution>`_
     """
 
-    def __init__(self, torch_dist, x_shape, event_dim, batch_size=None, log_pdf_mask=None):
+    def __init__(self, torch_dist, x_shape, event_dim, log_pdf_mask=None, *args, **kwargs):
```
fritzo: Can we remove the `*args, **kwargs`? Otherwise `batch_size` may be silently ignored.

pyro/distributions/random_primitive.py
```diff
@@ -34,7 +36,8 @@ def shape(self, x=None, *args, **kwargs):
         return self.dist_class(*args, **kwargs).shape(x)
 
     def sample(self, *args, **kwargs):
-        return self.dist_class(*args, **kwargs).sample()
+        sample_shape = kwargs.pop('sample_shape', torch.Size())
```
fritzo: I think you'll need to `kwargs.pop('sample_shape', None)` in all other methods too, otherwise this will be passed to the distribution constructor.

pyro/distributions/torch_wrapper.py
neerajprad: Actually, your comment above is related to this. This is basically to absorb the `sample_shape` argument, but I think that it will be cleaner to remove that in `random_primitive` than adding args/kwargs here. Will change.

pyro/distributions/random_primitive.py
neerajprad: That's right, this is being absorbed into args/kwargs in the constructor, but will change to remove `batch_shape` in the methods here.

pyro/distributions/torch_wrapper.py
fritzo: Thanks, I just want to make sure that users will see errors in their existing code that uses `batch_size` in distribution constructors. If the constructor silently swallows the arg then users will only later see a mysterious broadcasting error, or worse a scaled gradient with no error.

pyro/distributions/random_primitive.py
neerajprad: Makes sense.

pyro/distributions/torch_wrapper.py
neerajprad: Makes sense, will change.

neerajprad:
@fritzo - can we merge this?

PR #705, opened 2018-01-24
Fix GP tutorial and add to CI tests

fritzo:
- Fixes the GP tutorial's broken references to moved probability distributions
- Adds the tutorial to `make test-tutorials` and our travis-CI tests


fritzo:
cc @fehiepsi

fehiepsi:
In the new version of PyTorch, there is no `volatile` Variable anymore. To get rid of the error, in the plot function,  we can replace
```
Z = Variable(torch.linspace(-0.5, 5.5, 100), volatile=True)
mean, cov = model(Z)
```
by
```
Z = Variable(torch.linspace(-0.5, 5.5, 100))
with torch.no_grad():
    mean, cov = model(Z)
```

fritzo:
Thanks, I've updated to remove `volatile`.

fritzo:
@neerajprad would you mind merging this before other Distributions PRs, so that the GP tutorial will be tested in CI? Thanks!

PR #706, opened 2018-01-24
Thin out wrappers around TorchDistribution

neerajprad:
Makes the following changes so as to make the migration to torch distributions easier.
 - Removes most of the customized logic in `batch_log_pdf` and moves them to the torch wrapper `TorchDistribution` class.
 - Changes `sample`, `enumerate_support`, `batch_shape` and `event_shape` definitions to match those in PyTorch distributions.
 - Changes the shape of Categorical samples to match `torch.distributions.categorical` by removing the prepended event dim.


pyro/distributions/distribution.py
```diff
@@ -88,7 +88,7 @@ class Distribution(object):
     reparameterized = False
     enumerable = False
 
-    def batch_shape(self, x=None, *args, **kwargs):
+    def batch_shape(self, *args, **kwargs):
```
fritzo: Nice! It might be nice to drop random primitive and make this an actual `property`, but that can wait.

pyro/distributions/torch/dirichlet.py
```diff
@@ -52,7 +50,4 @@ def batch_log_pdf(self, x):
         :return: log probability densities of each element in the batch.
         :rtype: torch.autograd.Variable of torch.Tensor of dimension 1.
         """
-        batch_log_pdf = self.torch_dist.log_prob(x).view(self.batch_shape(x) + (1,))
-        if self.log_pdf_mask is not None:
-            batch_log_pdf = batch_log_pdf * self.log_pdf_mask
-        return batch_log_pdf
+        return super(Dirichlet, self).batch_log_pdf(x)
```
fritzo: Do you think it's worth simply removing these docs-only methods, now that we have `copy_docs_from`? It would certainly reduce our codebase.

pyro/distributions/torch_wrapper.py
```diff
@@ -13,22 +13,16 @@ class TorchDistribution(Distribution):
...
         self.log_pdf_mask = log_pdf_mask
-        self._x_shape = x_shape
-        self._event_dim = event_dim
 
-    def batch_shape(self, x=None):
-        x_shape = [] if x is None else x.size()
-        shape = torch.Size(broadcast_shape(x_shape, self._x_shape, strict=True))
-        event_start = len(shape) - self._event_dim
-        return shape[:event_start]
+    def batch_shape(self):
```
fritzo: Nice!

tests/distributions/test_categorical.py
```diff
@@ -80,19 +80,19 @@ def modify_params_using_dims(ps, dim):
...
 
 
-def test_batch_log_dims(dim, ps):
-    batch_pdf_shape = (3,) + (1,) * dim
-    expected_log_pdf = np.array(wrap_nested(list(np.log(ps)), dim-1)).reshape(*batch_pdf_shape)
-    ps = modify_params_using_dims(ps, dim)
-    support = dist.categorical.enumerate_support(ps)
-    batch_log_pdf = dist.categorical.batch_log_pdf(support, ps)
-    assert_equal(batch_log_pdf.data.cpu().numpy(), expected_log_pdf)
+# def test_batch_log_dims(dim, ps):
```
fritzo: Remove or mark xfail.

neerajprad: Yup, will do that. This is mostly to see if the integration tests pass.

pyro/distributions/torch/dirichlet.py
neerajprad: Yes, that's the plan finally, i.e. only class level documentation if needed. The remaining should just be replicated from Distribution or TorchDistribution.

fritzo: Great. I'm also trying to streamline our docs in #708 .

neerajprad:
> Looks great! I bet it was tricky working out that batching logic 😄

It induced a mild headache. :) Can't wait to remove all these temporary wrappers.

pyro/distributions/log_normal.py
```diff
@@ -32,24 +32,11 @@ def __init__(self, mu, sigma, *args, **kwargs):
...
-                raise ValueError("Parameter `mu` with shape {} is not broadcastable to "
-                                 "the data shape {}. \nError: {}".format(mu.size(), x.size(), str(e)))
-        return mu.size()[:-event_dim]
+    def batch_shape(self):
+        return self.mu.size()
 
     def event_shape(self):
-        event_dim = 1
-        return self.mu.size()[-event_dim:]
+        return ()
```
fritzo: It's probably a little safer to return `torch.Size(())` here, since `torch.Tensor(())` treats its argument as an iterable, whereas `torch.Tensor(torch.Size(()))` treats its argument as a shape.

pyro/distributions/log_normal.py
```diff
@@ -58,13 +45,14 @@ def sample(self, sample_shape=torch.Size()):
...
-        mu = self.mu.expand(self.shape(x))
-        sigma = self.sigma.expand(self.shape(x))
+        mu = self.mu
+        sigma = self.sigma
         ll_1 = Variable(torch.Tensor([-0.5 * np.log(2.0 * np.pi)]).type_as(mu.data).expand_as(x))
         ll_2 = -torch.log(sigma * x)
         ll_3 = -0.5 * torch.pow((torch.log(x) - mu) / sigma, 2.0)
         batch_log_pdf = torch.sum(ll_1 + ll_2 + ll_3, -1)
-        batch_log_pdf_shape = self.batch_shape(x) + (1,)
+        batch_log_pdf_shape = broadcast_shape(self.batch_shape() + self.event_shape(),
```
fritzo: nit: Can you use `self.shape()` in place of `self.batch_shape() + self.event_shape()`? I'm not sure what our semantics is.

pyro/distributions/random_primitive.py
```diff
@@ -35,9 +35,9 @@ def event_dim(self, *args, **kwargs):
         kwargs.pop('sample_shape', None)
         return self.dist_class(*args, **kwargs).event_dim()
 
-    def shape(self, x=None, *args, **kwargs):
+    def shape(self, *args, **kwargs):
         kwargs.pop('sample_shape', None)
-        return self.dist_class(*args, **kwargs).shape(x)
+        return self.dist_class(*args, **kwargs).shape()
```
fritzo: Now is this the correct behavior, or should `.shape()` also include the `sample_shape` that was popped on the previous line? Again I'm not sure what our semantics is.

pyro/distributions/torch_wrapper.py
```diff
@@ -37,8 +31,11 @@ def sample(self, sample_shape=torch.Size()):
             return self.torch_dist.sample(sample_shape)
 
     def batch_log_pdf(self, x):
-        batch_log_pdf_shape = self.batch_shape(x) + (1,)
+        sample_shape = broadcast_shape(self.batch_shape() + self.event_shape(), x.size(), strict=True)
```
fritzo: This is surprising, is it correct? I thought
```py
shape = sample_shape + batch_shape + event_shape
```
whereas this suggests that
```py
sample_shape >= shape = batch_shape + event_shape
```

neerajprad: I think the variable name is confusing, will rename this to `shape` instead. This is basically the broadcasted shape of `x` and the distribution's batch + event shape.

pyro/distributions/log_normal.py
neerajprad: Will change.

pyro/distributions/random_primitive.py
neerajprad: I was considering removing even the `shape` method later, and replacing with `batch_shape + event_shape` instead. But I think it will be useful to retain it and prepend `sample_shape` before. Will clean this up.

fritzo: Yeah I'm fine removing the ambiguously named `.shape()` method.

neerajprad:
Closing, as this was already included in #710.

PR #707, opened 2018-01-24
Nested inference example: Schelling coordination game

eb8680:
This PR adds two small examples (ported from [this Forest example](http://forestdb.org/models/schelling.html) and [this one](http://forestdb.org/models/schelling-falsebelief.html)) illustrating nested inference with enumeration in Pyro.  The examples implement two variants of a Schelling coordination game, in which two agents who want to meet decide where to go independently by reasoning recursively about the other's decision process.


examples/schelling.py
```diff
@@ -0,0 +1,56 @@
...
+                           obs=bob_prior)
+    else:
+        return bob_prior
+
+
+# We sample Bob's choice of location by marginalizing
+# over his decision process.
+bob_rec = Marginal(Search(bob))
+
+print(sum([bob_rec(Variable(torch.Tensor([0.6])), 2)
```
jpchen: can you print an informative output here? like `print(bob's marginal probability, sum(...).data[0])`

examples/schelling_false.py
```diff
@@ -0,0 +1,72 @@
...
+                           obs=bob_prior)
+    else:
+        return bob_prior
+
+
+# We sample Alice's true choice of location
+# by marginalizing over her decision process
+alice_decision = Marginal(Search(alice_fb))
+
+print(sum([alice_decision(Variable(torch.Tensor([0.55])), 3)
```
jpchen: same as above, can you put some message like `print(alice's true marginal probability, sum(...).data[0])`

eb8680:
> also example takes surprisingly long for a depth of only 2 :(

Not much we can do about that with naive implementations of `trace` and `replay`, unfortunately.

fritzo:
> Not much we can do about that with naive implementations of `trace` and `replay`

I see you're planting seeds for a "sped up Schelling example by 1000x" PR here :wink:

eb8680:
> can you also modify the test script so that this example gets tested?

Done.

> I see you're planting seeds for a "sped up Schelling example by 1000x" PR here 

Stay tuned...

eb8680:
Hmm, looks like `Marginal` needs to be updated after the latest distributions changes.

fritzo:
@eb8680 let me know if you want any help debugging. I'm not sure what broke, since `Marginal.batch_log_pdf()` also used to raise `NotImplementedError`.

eb8680:
@fritzo there was a change to `Marginal` in #710.  It's not immediately obvious to me what broadcasting semantics `Marginal` and `Histogram` should have.  I suspect the real solution is to rewrite them as kernel mean embeddings, with the current version as a special case with a delta kernel.

fritzo:
@eb8680 What I don't understand is that #710's change to `Marginal` looks like a no-op:
```diff
- def batch_log_pdf(self, ...):
-     raise NotImplementedError
+ def log_prob(self, ...):
+     raise NotImplementedError
```
I suspect the culprit is somewhere else in that PR.

neerajprad:
I can take a look to see what's broken.

neerajprad:
@eb8680, @fritzo - There is definitely a problem in the way that's handling batches. I will put up a fix for this shortly.

eb8680:
@neerajprad I think it's as simple as deleting `Marginal.log_prob` so that it uses `Histogram.log_prob`.

neerajprad:
> @neerajprad I think it's as simple as deleting Marginal.log_prob so that it uses Histogram.log_prob.

That's true! Please feel free to make that change. I wanted `log_prob` to accept an arbitrary batch of samples though (like other distributions), and this assumes that we are only using `log_prob` to score a single sample.

eb8680:
> That's true! Please feel free to make that change. I wanted log_prob to accept an arbitrary batch of samples though (like other distributions), and this assumes that we are only using log_prob to score a single sample.

`Marginal` is a bit of a weird special case since it can have non-Tensor return type and sampling and scoring aren't batched tensor operations internally.  We can revisit this soon along with `BranchPoutine`.

eb8680:
Fixed the merge conflict, @jpchen please merge after tests pass unless you have other requests for changes.

PR #708, opened 2018-01-24
Clean up docs: @copy_docs_from, GP docs

fritzo:
- Use links rather than full copied text when using the `@copy_docs_from` decorator. I've followed this pattern from PyTorch, since it nicely prevents duplication.
- Hook up @fehiepsi 's GP documentation with our Sphinx docs. It now appears after `make docs`.

### DIFF
--- a/docs/source/contrib.gp.rst
+++ b/docs/source/contrib.gp.rst
@@ -1,4 +1,33 @@
+Gaussian Processes
+==================
+
+See the `Gaussian Processes tutorial <http://pyro.ai/examples/gp.html>`_ for an introduction.
+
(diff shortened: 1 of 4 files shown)


docs/source/contrib.gp.rst
```diff
@@ -1,5 +1,34 @@
+Gaussian Processes
+==================
+
+See the `Gaussian Processes tutorial <http://pyro.ai/examples/gp.html>`_ for an introduction.
```
fritzo: Of course this link will only work after we publish the 0.2 release.

neerajprad:
Nice, will start using it in our distribution classes too!

PR #710, opened 2018-01-26
Deprecate batch log pdf in favor of PyTorch batching semantics

### LINKED ISSUE #712: Changes to tracegraphELBO to support arbitrary batch dims
I am trying to deprecate `batch_log_pdf` in distributions and have a single `log_prob` with the same batching semantics as PyTorch distributions, i.e. supporting arbitrary batch shape. I am finding that the air tutorial breaks at the following lines for #710:

```py
    for node in topo_sort_guide_nodes:
        node_log_pdf_key = 'batch_log_pdf' if node in guide_vec_md_nodes else 'log_pdf'
        downstream_costs[node] = model_trace.nodes[node][node_log_pdf_key] - \
            guide_trace.nodes[node][node_log_pdf_key]
        nodes_included_in_sum = set([node])
        downstream_guide_cost_nodes[node] = set([node])
        for child in guide_trace.successors(node):
            child_cost_nodes = downstream_guide_cost_nodes[child]
            downstream_guide_cost_nodes[node].update(child_cost_nodes)
            if nodes_included_in_sum.isdisjoint(child_cost_nodes):  # avoid duplicates
                if node_log_pdf_key == 'log_pdf':
                    downstream_costs[node] += downstream_costs[child].sum()
                else:
                    **downstream_costs[node] += downstream_costs[child] ** errors here
                nodes_included_in_sum.update(child_cost_nodes)
```

throwing an error, 
<details>
<summary> Error trace </summary>

```
  File "/Users/npradhan/workspace/pyro_dev/pyro/examples/air/main.py", line 328, in <module>
    main(**vars(parser.parse_args()))
  File "/Users/npradhan/workspace/pyro_dev/pyro/examples/air/main.py", line 215, in main
    loss = svi.step(X, args.batch_size, z_pres_prior_p=partial(z_pres_prior_p, i))
  File "/Users/npradhan/workspace/pyro_dev/pyro/pyro/infer/svi.py", line 98, in step
  File "/Users/npradhan/workspace/pyro_dev/pyro/pyro/infer/elbo.py", line 65, in loss_and_grads
    return self.which_elbo.loss_and_grads(model, guide, *args, **kwargs)
  File "/Users/npradhan/workspace/pyro_dev/pyro/pyro/infer/tracegraph_elbo.py", line 257, in loss_and_grads
    loss += self._loss_and_grads_particle(weight, model_trace, guide_trace)
  File "/Users/npradhan/workspace/pyro_dev/pyro/pyro/infer/tracegraph_elbo.py", line 291, in _loss_and_grads_particle
    model_trace, guide_trace,  model_vec_md_nodes, guide_vec_md_nodes, non_reparam_nodes)
  File "/Users/npradhan/workspace/pyro_dev/pyro/pyro/infer/tracegraph_elbo.py", line 62, in _compute_downstream_costs
    downstream_costs[node] += downstream_costs[child]
RuntimeError: The expanded size of the tensor (3) must match the existing size (50) at non-singleton dimension 1
```

</details>

Upon investigating, I found that the reason this error is thrown is that it tries to sum up the `batch_log_pdf` for `node` and `child` which have shapes `(60, 3)` and `(60, 5)`. Our `batch_log_pdf` implementation just preserves the leftmost dimension, and hence the shape of the pdf score was `(60, 1)` for both of these sites and they could be summed up. With our new batching semantics, there is no concept of a single batch dimension and hence `log_prob` returns the same sized tensor* which cannot be summed up as earlier. Does this necessitate a change in tracegraphELBO, or is there a correct way of handling this inside AIR? 

cc. @fritzo, @martinjankowiak, @null-a.

\* for this distribution there is no `event_dim`, hence the shape of the scored tensor remains the same.

neerajprad:
This deprecates `batch_log_pdf`, so that we can adopt the batching/broadcasting semantics from PyTorch distributions. All tests and tutorials currently work.

**Note for reviewers:**

 - This is stacked on top of PR #706. I have squashed commits to organize this - the first commit has all changes from #706, the next commit has all major changes to deprecate `batch_log_pdf`, and the final commit introduces `extra_event_dims` and modifications to the AIR tutorial (inc. the changes from @fritzo). 
 - `extra_event_dims` is used to indicate additional event dimensions that should be summed out in `log_prob` computation. e.g. for univariate normal `log_prob` returns the same sized tensor as the data to be scored. However, if we specify the `extra_event_dims=1`, the last dimension would be summed out and stored in `batch_log_pdf` field for that stochastic node. This is needed to have a canonical way of specifying distributions with diagonal covariance structure in `traceGraph_ELBO`. After discussions with @fritzo, this seemed like the most non-invasive way forward, but this is an easily reversible decision, and can be revisited in the future.
 - Note that `mapdata` (/`iarange`) still only supports a single batch dimension, and is outside the scope of this PR.

Fixes #712.


neerajprad:
Currently blocked on #712.

neerajprad:
> i think this requires changes in the tutorials as well

I will make the changes to all failing tutorials in the tests. If there are tutorials that are not under test coverage, they will be ignored for now. I will create tasks for the tutorials owners to update their tutorials and add test coverage like #694.

pyro/distributions/log_normal.py
```diff
@@ -32,40 +32,23 @@ def __init__(self, mu, sigma, *args, **kwargs):
...
-        mu = self.mu.expand(self.shape(x))
-        sigma = self.sigma.expand(self.shape(x))
-        ll_1 = Variable(torch.Tensor([-0.5 * np.log(2.0 * np.pi)]).type_as(mu.data).expand_as(x))
-        ll_2 = -torch.log(sigma * x)
-        ll_3 = -0.5 * torch.pow((torch.log(x) - mu) / sigma, 2.0)
-        batch_log_pdf = torch.sum(ll_1 + ll_2 + ll_3, -1)
-        batch_log_pdf_shape = self.batch_shape(x) + (1,)
-        return batch_log_pdf.contiguous().view(batch_log_pdf_shape)
+    def log_prob(self, x):
+        ll_1 = Variable(torch.Tensor([-0.5 * np.log(2.0 * np.pi)]).type_as(self.mu.data).expand_as(x))
```
fritzo: This simpler version works correctly on multi-GPU systems:
```py
ll_1 = -0.5 * np.log(2.0 * np.pi)
```
It's fine to address this in a later PR though.

pyro/distributions/torch_wrapper.py
```diff
@@ -13,36 +13,41 @@ class TorchDistribution(Distribution):
...
+    def log_prob(self, x):
+        log_prob = self.torch_dist.log_prob(x)
+        for _ in range(self.extra_event_dims):
+            log_prob = log_prob.sum(-1)
         if self.log_pdf_mask is not None:
-            batch_log_pdf = batch_log_pdf * self.log_pdf_mask
-        return batch_log_pdf
+            # Prevent accidental broadcasting of log_prob tensor,
+            # e.g. (64, 1), (64,) --> (64, 64)
+            assert len(self.log_pdf_mask.size()) <= len(log_prob.size())
```
fritzo: :+1: Good idea!

pyro/util.py
```diff
@@ -205,7 +207,8 @@ def enum_extend(trace, msg, num_samples=None):
         num_samples = -1
 
     # Batched .enumerate_support() assumes batched values are independent.
-    batch_shape = msg["fn"].batch_shape(msg["value"], *msg["args"], **msg["kwargs"])
+    batch_shape = broadcast_shape(msg["fn"].shape(*msg["args"], **msg["kwargs"]),
+                                  msg["value"].size())[:-1]
```
fritzo: What is the `[:-1]` slice for?

neerajprad:
@null-a - Could you take a look at the AIR example and tutorial, and let me know if the changes look good to you?

pyro/util.py
neerajprad: I seem to have missed updating this in the second part of the PR. Thanks for spotting this! Will change to using the `event_dim` arg and update.

pyro/distributions/torch_wrapper.py
neerajprad: This was an issue with the AIR tutorial, so thought of adding this as a check.

pyro/distributions/log_normal.py
neerajprad: Oddly, enough this threw an error `AttributeError: type object 'Variable' has no attribute 'from_numpy'` when we add this to a Variable later, but changing to python numeric type works.

neerajprad:
> Looks great. What is your plan for merge order?

I think in this case it might just be better to do one atomic update rather than staggered merges that are likely to have some broken semantics. If there are any major PRs out, we should try to get them merged though. What do you think?

pyro/distributions/log_normal.py
fritzo: Oh right, I've had that problem too.

fritzo:
@jpchen Do you want to take a second look at this before we merge it?

neerajprad:
@null-a : If you have any comments on the AIR part, I will address that in a separate PR.

null-a:
@neerajprad Thanks. I've not kept up with recent changes to batching semantics, so I don't think I can add anything, but if you're all happy, then so am I! I'll take another peek once I've caught up though.

Issue #547, opened 2017-11-10
multiple versions of documentation, one per release (+dev)


dustinvtran:
Having docs for varying versions, especially dev and master, sounds useful for development. I don't know if Pytorch has a good solution but I like that TF has multiple versions. FWIW, read the docs does this automatically: https://docs.readthedocs.io/en/latest/index.html.

relevant: https://github.com/uber/pyro/pull/545#issuecomment-343333425

ngoodman:
we did this in webppl (using readthedocs) and it worked nicely. (all though it meant i pointed people to the dev docs most of the time ;)

jpchen:
unfortunately you can only do that if it is hosted on readthedocs.org. pytorch does this manually which is why they have this [hacked together page](http://pytorch.org/docs/versions.html). the problem with this is that, our doc files will scale with each release, which will be a lot of overhead when a user clones pyro.

neerajprad:
How about hosting documentation / tutorials on readthedocs, and retaining our domain name - http://docs.readthedocs.io/en/latest/alternate_domains.html?

jpchen:
oo hm this seems new ill look into this

jpchen:
docs.pyro.ai (give 24 hours for CNAME to propagate)

neerajprad:
Great! 
 - Can we opt out of the ads (http://docs.readthedocs.io/en/latest/ethical-advertising.html#opting-out)? They seem kind of pesky.
 - Can we get our old CSS back?

jpchen:
> Can we opt out of the ads

nope - i already opted out of paid ads but thats all we can do on the free tier

>Can we get our old CSS back?

Go for it :)

PR #713, opened 2018-01-26
get Pyro version automatically

jpchen:
partially addresses #547, #638

this makes the version in the docs consistent with the version deployed. it is the first step in having togglable versions (those changes would be on a different branch; see discussion in issue)


jpchen:
killed tests except lint and docs

docs/source/conf.py
```diff
@@ -66,9 +68,9 @@
 #
 # The short X.Y version.
 # version = pkg_resources.require('pyro-ppl')[0].version
-version = u'0.1.2'
+version = __version__
```
fritzo: It seems like we'd want to be git aware here, something like
```py
if git_head_is_tag():
    version = git_tag()  # for major versions
else:
    version = '{}+{}'.format(__version__, git_commit()[:8])  # for dev
```
WDYT?

tutorial/source/conf.py
```diff
@@ -68,9 +71,10 @@
 # built documents.
 #
 # The short X.Y version.
-version = u''
-# The full version, including alpha/beta/rc tags.
-release = u''
+# get the latest git tag, otherwise get the latest commit
+version = re.sub('^v', '', os.popen('git describe --tags').read().strip())
```
fritzo: I usually use `check_call('git describe --tags'.split()).strip()` from `subprocess` module. Is there any difference between this and `os.popen()`.

fritzo: Sorry, could you add the comment with expected output below this line?

jpchen: > When I run it I get pytorch-0.3-28-g4c52460 and I'm not sure whether this is expected

it is. basically the last tag and the commit in which it corresponds to (for those bleeding-edge users)

> Is there any difference between this and os.popen()

not anything significant.. i can change it if you have a reason that incantation is better

>  could you add the comment

done

fritzo:
Why are tests failing?

fritzo:
If you just let tests run to completion, then reviewers won't need to navigate to travis-ci.org to figure out why tests are failing.

jpchen:
>  If you just let tests run to completion, then reviewers won't need to navigate to travis-ci.org to figure out why tests are failing.

sorry will do. back in my day we were very short on cores so being environmentally conscious was lauded not berated

fritzo:
Fix lint failures.

PR #715, opened 2018-01-27
Add two tests of diag normal shape

fritzo:
This PR adds two simple tests that should help us preserve distribution behavior in #706 and #710. In particular, I've added TODOs to indicate how these tests will change through those PRs.

### DIFF
--- a/tests/distributions/test_shapes.py
+++ b/tests/distributions/test_shapes.py
@@ -75,6 +75,7 @@ def test_diag_normal_batch_log_pdf_shape():
     mu1 = ng_zeros(2, 3)
     mu2 = ng_zeros(2, 4)
     sigma = ng_zeros(2, 1)
+    # TODO these will need extra_event_dims=1
     d1 = dist.Normal(mu1, sigma.expand_as(mu1))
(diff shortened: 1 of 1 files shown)


tests/distributions/test_shapes.py
```diff
@@ -69,3 +69,25 @@ def test_normal_batch_log_pdf_shape():
...
+    x2 = d2.sample()
+    assert d1.batch_log_pdf(x1).size() == (2, 1)  # TODO will become (2,)
+    assert d2.batch_log_pdf(x2).size() == (2, 1)  # TODO will become (2,)
+    d1.batch_log_pdf(x1) + d2.batch_log_pdf(x2)  # this line should not error
+
+
+def test_diag_normal_log_pdf_mask_shape():
+    mu = ng_zeros(4, 3)
+    sigma = ng_zeros(4, 3)
+    mask = ng_ones(4, 1)  # note that 3 was summed out  TODO will become (2,)
```
neerajprad: This should be become `(4,)` right (not `(2,)`?

fritzo: right, fixed.

Issue #702, opened 2018-01-24
Implement variational GP where pseudo inputs are learned using SVI


fritzo:
@fehiepsi 's first version of Gaussian Processes uses the full dataset. For large scale problems it is cheaper to use s small set of inducing points. Let's try to learn these points this using Pyro.

## Tasks

- [x] Implement an inducing point guide
- [x] Add more covariance functions to pyro.contrib.gp
- [x] Implement differentiable version of `torch.potrs` or `torch.trtrs`.
  See https://github.com/pytorch/pytorch/issues/4669 and https://github.com/pytorch/pytorch/issues/3073

fritzo:
cc @ysaatchi @fehiepsi

ssequeira:
I'm searching for simple tasks to begin contributing to Pyro. I'd be happy
to add more kernels; should I start with Matern 3/2 and 5/2/

Best,
Shannon Sequeira

ysaatchi:
Yes, that's a good place to start! Other ones could include linear, periodic, neural network and spline. 
Another family which is super useful in practice are the "meta kernels", e.g. additive (kernel is sum of constituent kernels), product (kernel is sum of constituent kernels). Thanks!

fehiepsi:
I will take care of implement the spare version of GPR.

ysaatchi:
@fehiepsi your setup is very conducive to VGP, you can add pseudo-inputs and targets to your hyperpriors dict and use SVI to learn them via pyro machinery (I hope!)

fehiepsi:
@ysaatchi The spare models require using "Woodbury matrix identity" to reduce cost to inverse matrices and calculate logdet. I have written down all calculations for SpareGPR with DTC and FITC assumptions.  Implement them is quite straightforward and similar to the current GPR implementation.

However, for SpareGPR with VFE, there is an additional regularizer term `Trace(K_theta)` (formula (2.33) in [Damianou's thesis](http://etheses.whiterose.ac.uk/9968/1/Damianou_Thesis.pdf)). One option is to add a callback function to SVI to add this additional loss to ELBO. A better option is to write `guide` in a suitable way (I am getting stuck at this point) such that ELBO will have this additional term.

I am starting to implement them and hope to have a preliminary version this weekend so we can start to discuss more.

ysaatchi:
Interesting, I think your intuition RE finding a good guide makes sense. Let me think about it and get back to you.

eb8680:
[...]

Issue #662, opened 2018-01-04
Point to literature in algorithm implementations


fritzo:
[Suggested](https://forum.pyro.ai/t/research-papers-on-variational-inference-algorithms-pyro-implements-under-the-hood/76) by Aleksandr Reshytko:
> Can you please refer the research papers on variational inference algorithms and techniques that pyro uses under the hood. Without good understanding of the actual inference happens it's hard to debug and understand the models.

Implementations like `TraceGraph_ELBO` should add comments pointing to literature to aid readers understand and adapt the code. See for example @OptimusLime's excellent commenting in #622 e.g. in [HM.sample](https://github.com/uber/pyro/pull/622/files#diff-6179a7b6f32c720f1654ebd0f4f4c9e5R166).

- [x] Trace_ELBO
- [x] TraceGraph_ELBO
- [x] ...other algorithms...?

PR #722, opened 2018-01-30
[Done] Implement spare gaussian process regression

fehiepsi:
@fritzo @ysaatchi This pull request is an attempt to implement sparse approximation versions of GPR (#702). The main reference for notations is [Damianou's thesis](http://etheses.whiterose.ac.uk/9968/1/Damianou_Thesis.pdf). The supported methods are DTC, FITC, and VFE.

- [x] Implement model and guide for DTC, FITC, VFE.
- [x] Implement inference for the above sparse methods.
- [x] Add support for Kernel to quickly calculate diagonal of covariance matrices.
- [x] Add options to predict full cov matrix or just diagonal form, with noise or not noise (y or f), jitter.
- [x] Move SpareMultivariateNormal to distribution section (make another pull request for this to easy to test).
- [x] Add tests to detect if there are errors in the implementation.
- [x] Fix bug in tutorial and ~add a section on sparse regression~ (move this part to another pull request).

Note that the implementations of Likelihood module and Mean function module will be in some future pull requests.


pyro/contrib/gp/models/sgpr.py
```diff
@@ -0,0 +1,182 @@
...
+from torch.autograd import Variable
+import torch.nn as nn
+
+import pyro
+import pyro.distributions as dist
+
+from .util import _matrix_triangular_solve_compat
+
+
+class _SparseMultivariateNormal(dist.MultivariateNormal):
```
fritzo: Feel free to move this into `pyro.distributions.sparse_mvn` or something. That would make it easier to add distribution tests for it.

fehiepsi: Sure, after making things work, I will refactor it.

fehiepsi:
@fritzo @ysaatchi I have completed the implementation of sparse regression models. There are several tests for `forward` pass of these models. And I add a sparse section in the GP tutorial to illustrate the `inference` step. Would you mind helping me review this pull request?

p/s: I don't know if the way I unify all three models "DTC", "FITC", "VFE" is standard or not (I don't have much background, just learn about GP around two months ago, and this implementation comes from what I understand from various literature - especially [this note](https://github.com/GPflow/GPflow/blob/master/doc/source/notebooks/SGPR_notes.ipynb) - after writing down all computation details). As far as I can see, it is pretty fast to inference. The math in the implementation is quite clear to follow because I use the standard notions.

fehiepsi:
@fritzo Travis fails at running the tutorial due to not detecting my "pyro" kernel. I don't know how to resolve it. :(

neerajprad:
> @fritzo Travis fails at running the tutorial due to not detecting my "pyro" kernel. I don't know how to resolve it. :(

Could you try reverting the changes to the metadata, in particular removing the kernel name `pyro` from kernelspec? It's trying to activate the named ipython kernel `pyro`, but that is only on your local system. I think you can also just run `python -m nbstripout --keep-output tutorial/source/gp.ipynb` which will revert the non-essential changes to the notebook:

```
  "metadata": {
    "kernelspec": {		    "kernelspec": {
 -   "display_name": "Python 2",		 +   "display_name": "Python (pyro)",
     "language": "python",		     "language": "python",
 -   "name": "python2"		 +   "name": "pyro"
    },		    },
```

fehiepsi:
@neerajprad Thanks for the tip!!! I just modified it.

fehiepsi:
After https://github.com/pytorch/pytorch/pull/4923, we can use `mean`, `variance` in Distributions. I suggest we update Travis with a newer build of pytorch.

fritzo:
@fehiepsi we're having issues building a new PyTorch wheel for testing, but we'll update as soon as we can.

neerajprad:
> After pytorch/pytorch#4923, we can use mean, variance in Distributions. I suggest we update Travis with a newer build of pytorch.

The wheels are updated, and these methods should be available now. I restarted the tests that were failing earlier.

pyro/contrib/gp/models/sgpr.py
```diff
@@ -0,0 +1,172 @@
...
+        if Xnew.dim() == 1:
+            Xnew = Xnew.unsqueeze(1)
+
+        kernel, Xu = self.guide()
+        Kuu = kernel(Xu) + self.jitter.expand(self.num_inducing)
+        Kus = kernel(Xu, Xnew)
+        Kuf = kernel(Xu, self.X)
+        Luu = Kuu.potrf(upper=False)
+
+        # Ref: "A Unifying View of Sparse Approximate Gaussian Process Regression"
```
fritzo: Thanks for pointing to literature #662

tutorial/source/gp.ipynb
```diff
@@ -433,44 +436,188 @@
...
...PFVulFXPZDuAZJ+BKO4IWydYr9elGJb+eeZRkKGiMwRXx/H9Xk6+8mQdU7RXTSYp2zNmvUmLdcxz8uLDFpG/S7lY+0KdCQFsw/VnulJIgcruACcnJ6WbFbsVuRLshwpe7ucyPaHtuzaxaR5CVOXrrLU/8DClTZmfTckDz5/h8pv/76pt78vY0vYiWr5zbm9ybq1b1jrvxdfW69RTjoB7vsZQdgAhijvC1gmm0+mWT/FoNNLn5+cb341xT4gtg/zZ1qFM9wnOS2zz8Q05DSjHxBcbSW/6GDYx2XHZjo+PS6sZC4jr62t9dnZWfm7XJjMb+zJR9wHZB1w+t7x7xNbkyWRS5iOvCiqz9RtbP7SdtBU7kZq7TzHW4FzZJ5hQf2pbXZnHw3e52AsdM/vcJ3MsolNdNmItn6457+zsLHv7dJ1om+I+KduuS/yHZJeZTqd6PB5vuRr2oR01AURxh5iTAVtrzM7n2hYJDWRz4RtUYiYxVyBgUWy6Wrj8kVJWwqaFy1xZ+7aSpHVL1qWZ57HupFIUmzkleVJmd5OUiaBp+jyRtkEfnj+0DGz95DzFtgVfqGjle8htZe63sRO0q6/JRanpOpEijlLeVcp2PX+H69n13TbbTsiYmUN4xhDjbpKjrlKez1zsSWOFqywspOXuqJm+M/T+vrLPZjN9enq6tZg03Serni/U4FPVhvi55eFftqOq9w2I4o6JsahI5ISYO9WQz8JrduaqbUk5CMnAwap6qML0D+N/2/ITVw2ernvnmFTk1hNPqrHWqhByTchtT6R9w7XYYncF0/dWBp8ybN2VliSXOKhbTl4Yv/nmm3oymWxM9r7AONcYY9tWNvtY1URqGx/W67XVHaSpLWff55uyFLfdd0LGzJRnTiVGkJmktgMWk6FucDy3yfpgY4XrvRWF27hTp37N75onzbIIjT1wpiheGWPM8wFknbBV2ndd6d4o+8F0Oq30w95lIIo7xOwUoStPc0J0DdSpA40t4pUDElwCwCYW+N8232hfPcQ+R53vh07yVYNe1XbtwcFBKYybyNAQMyHXrU8bfbCw5oIXgTK7CZ8aJY9R5j7ryhzBQrAJcXB1dVUm8OdxYDqd6qOjIz2dTp3bwr52wv9mH2VzYWk+p2siNe/LsRJNpEYLbaspglW+49FopN97772N8q/X69INKrY8dYm5T5sZYWIWEbbv
```
fritzo: Why is the envelope so narrow around the inducing points? I would expect a plot of the estimated total variance at each point x, whereas it looks like you're plotting only the uncertainty in the mean (not the variance due to measurement error). Do you think it would be clearer to plot total uncertainty rather than uncertainty in the mean?

fritzo: (this comment refers to the final figure)

pyro/contrib/gp/models/gpr.py
```diff
@@ -5,81 +5,94 @@
...
+        if Xnew.dim() == 1:
+            Xnew = Xnew.unsqueeze(1)
+
         kernel = self.guide()
-        K = kernel(self.X) + self.noise.repeat(self.input_dim).diag()
+        Kff = kernel(self.X)
+        if noiseless:
+            Kff = Kff + self.jitter.expand(self.num_data).diag()
+        else:
+            Kff = Kff + self.noise.expand(self.num_data).diag()
```
ysaatchi: This is correct, however, it is safer to add jitter regardless, as sometimes the MAP solution may be attracted to short lengthscales and very low noise (aka Level-2 overfitting :))

pyro/contrib/gp/models/gpr.py
```diff
@@ -5,81 +5,94 @@
...
+        Lffinv_y = matrix_triangular_solve_compat(self.y, Lff, upper=False)
+        Lffinv_Kfs = matrix_triangular_solve_compat(Kfs, Lff, upper=False)
 
-        # TODO Use torch.trtrs or torch.potrs when it supports cuda tensors
-        # and is differentiable.
-        # Refer to Gaussian processes for machine learning main gpr algorithm
-        # L = torch.potrf(K, upper=False)
-        # alpha = torch.potrs(L, self.y, upper=False)
+        # loc = Kfs.T @ inv(Kff) @ y
+        loc = Lffinv_Kfs.t().matmul(Lffinv_y)
```
ysaatchi: wow, cool, this works now?

pyro/contrib/gp/models/sgpr.py
```diff
@@ -0,0 +1,172 @@
...
+class SparseGPRegression(nn.Module):
+    """
+    Sparse Gaussian Process Regression module.
+
+    :param torch.autograd.Variable X: A tensor of inputs.
+    :param torch.autograd.Variable y: A tensor of outputs for training.
+    :param pyro.contrib.gp.kernels.Kernel kernel: A Pyro kernel object.
+    :param pyro.contrib.gp.InducingPoints Xu: An inducing-point module for spare approximation.
+    :param torch.Tensor noise: An optional noise tensor.
+    :param str approx: One of approximation methods: 'DTC', 'FITC', and 'VFE' (default).
```
ysaatchi: may be nice for newcomers to expand these acronyms

tests/contrib/gp/test_models.py
```diff
@@ -4,21 +4,72 @@
...
+    kernel = RBF(input_dim=3, variance=torch.ones(2), lengthscale=torch.ones(2))
+    X = Variable(torch.Tensor([[2, 5, 3], [4, 3, 7]]))
+    y = Variable(torch.Tensor([0, 1]))
+    Xu = InducingPoints(X.data.clone())  # must be set to compare
+    Xnew = Variable(torch.Tensor([[3, 1, 4], [1, 3, 1]]))
+    noise = torch.Tensor([1])
+
+    gpr = GPRegression(X, y, kernel, noise=noise)
+    sgpr = SparseGPRegression(X, y, kernel, Xu, noise=noise)
+    sgpr_ftic = SparseGPRegression(X, y, kernel, Xu, approx="FITC", noise=noise)
```
ysaatchi: typo: call it sgpr_fitc

pyro/contrib/gp/models/gpr.py
fehiepsi: For `torch.trtrs` cuda and grad: not yet. We make a wrapper here: https://github.com/uber/pyro/blob/dev/pyro/distributions/util.py#L274

pyro/contrib/gp/models/gpr.py
fehiepsi: I think that it is better to address this when we start to put constraints on parameters (lengthscale, variance, noise,...).

tutorial/source/gp.ipynb
fehiepsi: Nice catch! The current implementation predicts $f*$, not $y*$. Although I have put the option: `noiseless=False`, I forgot to include it in the implementation. Will fix this soon!

tests/contrib/gp/test_models.py
fehiepsi: yup! thanks!

pyro/contrib/gp/models/sgpr.py
fehiepsi: I address this in the description of the model. :)

fehiepsi:
@ysaatchi @fritzo I addressed your reviews in the last commit. Especially, I fixed a mistake while introducing the `noiseless` parameter for prediction. Now, we get a reasonable plot for sparse version.

@ysaatchi I already made a preliminary version of the variational approach. Will make a pull request for it soon. It is much cleaner than the exact implementation here and is able to address general likelihood! :)

PR #727, opened 2018-01-30
Fix profiling scripts after distributions refactor

neerajprad:
Due to distribution API changes, the profiling script was broken. This fixes the script to work with the changed API.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,13 +36,18 @@ jobs:
               - make lint
               #- make scrub;
               #  git diff-index --quiet HEAD
-        - stage: unit test
+        - stage: auxiliary modules
           python: 2.7
           env: STAGE=docs
-          script: 
+          script:
               - pip install -r docs/requirements.txt
               - make docs
         - python: 2.7
+          env: STAGE=profiler
+          script:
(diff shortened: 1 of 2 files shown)


fritzo:
Should we add the profiling script to any unit tests? I've usually done this in the past, where the script is parameterized by `num_samples` or some other cost parameter and tests run with small cost.

neerajprad:
> Should we add the profiling script to any unit tests? I've usually done this in the past, where the script is parameterized by num_samples or some other cost parameter and tests run with small cost.

If it is with an aim to prevent breakages - we can just add `python -m profiler.distributions` to .travis.yml`, to quickly check if the script works as expected (shouldn't take any time). I will do that actually. If it is to actually test perf regression, that'll need some more thinking.

fritzo:
Let's add `python -m profiler.distributions` to `.travis.yml` as you suggest, that sounds like a good idea.

neerajprad:
> Let's add python -m profiler.distributions to .travis.yml as you suggest, that sounds like a good idea.

Done.

fritzo:
Looks great. I'll merge as soon as tests pass.

PR #729, opened 2018-01-31
Use torch.distributions.kl_divergence() in Trace_ELBO

### LINKED ISSUE #688: Use kl_divergence() in SVI
PyTorch distributions implements a `kl_divergence()` function. We should use this when available in SVI. First we need to account for Pyro distribution wrappers:
```py
def kl_divergence(p, q):
    try:
        p = p.torch_dist
        q = q.torch_dist
    except AttributeError:
        raise NotImplementedError
    return torch.distributions.kl_divergence(p, q)
```
then in SVI we can `try...except`:
```py
assert self.assume_mean_field and self.use_analytic_kl
try:
    kl = kl_divergence(guide_site['fn'], model_site['fn'])
    elbo += kl.detach().sum()
    surrogate_elbo += kl
except NotImplementedError:
    # fall back to monte carlo algorithm
```

fritzo:
Closes #688 

This attempts to integrate @vishwakftw's `kl_divergence()` implementation into Pyro's `Trace_ELBO` inference algorithm.

`torch.distributions.kl_divergence()` currently has some numerical issues for extreme values of distributions; this is being addressed in https://github.com/probtorch/pytorch/issues/117

### DIFF
--- a/pyro/infer/trace_elbo.py
+++ b/pyro/infer/trace_elbo.py
@@ -8,6 +8,7 @@ from torch.autograd import Variable
 
 import pyro
 import pyro.poutine as poutine
+from pyro.distributions import kl_divergence
 from pyro.distributions.util import torch_zeros_like
 from pyro.infer.enum import iter_discrete_traces
 from pyro.infer.util import torch_backward, torch_data_sum, torch_sum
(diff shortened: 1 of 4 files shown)


pyro/infer/trace_elbo.py
```diff
@@ -161,21 +162,31 @@ def loss_and_grads(self, model, guide, *args, **kwargs):
...
-                        elbo_particle += model_log_pdf - guide_log_pdf
-                        surrogate_elbo_particle += model_log_pdf
-
-                        if not is_identically_zero(entropy_term):
-                            if not batched:
-                                entropy_term = entropy_term.sum()
-                            surrogate_elbo_particle -= entropy_term
 
-                        if not is_identically_zero(score_function_term):
+                        try:
```
fritzo: @martinjankowiak Does this look right?

eb8680:
This can't be applied to arbitrary model-guide pairs, so it should live in another loss function or be explicitly enabled with a kwarg.

martinjankowiak:
my feeling is that we if want something like this in trace_elbo it should be enabled by something like
`mean_field = True, analytic_kl = True`

fritzo:
> @martinjankowiak `mean_field = True`

@eb8680 suggested that there may be a wider setting in which analytic KLs can be used, perhaps if the cliques of the factor graphs of the model and guide agree. We'll have to work out details...

martinjankowiak:
i already worked out the details months ago...

1. The parents of z_n in the p-DAG must not include any nodes that are descendants of z_n in the q-DAG
2. The distributions library must have the relevant analytic formulas available 

so another simple case is p-DAG = q-DAG

karalets:
Yeah we really have had this discussion when I pitched the kl divergences
and Martin's comments there should still apply the same.

fritzo:
closing since this is very out-of-date

PR #730, opened 2018-01-31
Add sparse multivariate normal distribution

fehiepsi:
To simplify the review for #722, which is almost finished, I move part of that pull request which adds a SparseMultivariateNormal distribution to here.

This pull request implements a "sparse" (fast calculation) for log prob of a Multivariate Normal distribution whose covariance matrix has the form D + W.T @ W. Here D is a diagonal matrix and W is a matrix of size `M x N` (in the sparse situation: M << N).

### DIFF
--- a/pyro/distributions/sparse_multivariate_normal.py
+++ b/pyro/distributions/sparse_multivariate_normal.py
@@ -47,7 +47,7 @@ class SparseMultivariateNormal(Distribution):
 
         self.trace_term = trace_term if trace_term is not None else 0
         super(SparseMultivariateNormal, self).__init__(*args, **kwargs)
-        
+
     @property
     def mean(self):
         return self.loc
--- a/pyro/distributions/sparse_multivariate_normal.py
+++ b/pyro/distributions/sparse_multivariate_normal.py
@@ -1,0 +1,125 @@
+from __future__ import absolute_import, division, print_function
+
+import math
+
+import torch
+from torch.autograd import Variable
+from torch.distributions.utils import lazy_property
+
+from pyro.distributions.distribution import Distribution
+from pyro.distributions.util import copy_docs_from, matrix_triangular_solve_compat
+
+
+@copy_docs_from(Distribution)
+class SparseMultivariateNormal(Distribution):
+    """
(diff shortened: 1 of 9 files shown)


fehiepsi:
Oops, there is something change in Pyro distribution recently. I will make a fix soon.

pyro/distributions/sparse_multivariate_normal.py
```diff
@@ -0,0 +1,127 @@
...
+import torch
+from torch.autograd import Variable
+from torch.distributions.utils import lazy_property
+
+from pyro.distributions.distribution import Distribution
+from pyro.distributions.util import copy_docs_from, matrix_triangular_solve_compat
+
+
+@copy_docs_from(Distribution)
+class SparseMultivariateNormal(Distribution):
```
fritzo: It would be nice to implement this as a PyTorch distribution + wrapper as we've done with MVN
```py
# in pyro/distributions/torch/sparse_multivariate_normal.py
class TorchSparseMultivariateNormal(torch.distributions.Distribution):
    has_rsample = True
    ...your implementation here...

class SparseMultivariateNormal(TorchDistribution):
    reparameterized = True
    def __init__(self, loc, D_term, W_term, trace_term=None, *args, **kwargs):
        torch_dist = TorchSparseMultivariateNormal(loc, D_term, W_term, trace_term)
        super(SparseMultivariateNormal, self).__init__(torch_dist, *args, **kwargs)
```
This would have a few advantages:
- it makes it easier to move this implementation upstream to PyTorch at some time
- it supports `log_pdf_mask` and `extra_event_dim`
- it makes it easier to register a `kl_divergence()` implementation
- it makes it easier for us to refactor towards PyTorch distributions interface

fehiepsi:
@fritzo Thanks a lot for your review! I have made changes relevant to what you suggested. There are  several bugs that I detected along the way, so I include them on this pull request:
+ docs hook for binomial, poisson
+ change from `scale_triu` to `scale_tril` in MultivariateNormal (because we put constraint on `scale_tril`, not `scale_triu`).

This should pass all tests now.

Issue #711, opened 2018-01-26
DMM example returns NAN on Pyro dev


neerajprad:
While refactoring to move to PyTorch distributions, I noticed that the `dmm.py` example has been throwing warnings after one epoch when it starts encountering NANs at multiple sites. 

This is on the current `dev` branch with Pytorch `master`. @martinjankowiak - Could you help take a look at what's happening? 

<details>

```
/Users/npradhan/miniconda2/envs/pytorch/bin/python /Users/npradhan/workspace/pyro_dev/pyro/examples/dmm/dmm.py
Namespace(annealing_epochs=1000, beta1=0.96, beta2=0.999, checkpoint_freq=0, clip_norm=20.0, cuda=False, iaf_dim=100, learning_rate=0.0004, load_model='', load_opt='', log='dmm.log', lr_decay=0.99996, mini_batch_size=20, minimum_annealing_factor=0.1, num_epochs=5000, num_iafs=0, rnn_dropout_rate=0.1, save_model='', save_opt='', weight_decay=0.6)
N_train_data: 229     avg. training seq. length: 60.29    N_mini_batches: 12
[training epoch 0000]  61.6189 				(dt = 16.419 sec)
/Users/npradhan/workspace/pyro_dev/pyro/pyro/poutine/trace.py:13: UserWarning: Encountered NAN log_pdf at site 'z_70'
  warnings.warn("Encountered NAN log_pdf at site '{}'".format(name))
/Users/npradhan/workspace/pyro_dev/pyro/pyro/poutine/trace.py:13: UserWarning: Encountered NAN log_pdf at site 'z_71'
  warnings.warn("Encountered NAN log_pdf at site '{}'".format(name))
/Users/npradhan/workspace/pyro_dev/pyro/pyro/poutine/trace.py:13: UserWarning: Encountered NAN log_pdf at site 'z_73'
  warnings.warn("Encountered NAN log_pdf at site '{}'".format(name))
/Users/npradhan/workspace/pyro_dev/pyro/pyro/poutine/trace.py:13: UserWarning: Encountered NAN log_pdf at site 'z_74'
  warnings.warn("Encountered NAN log_pdf at site '{}'".format(name))
/Users/npradhan/workspace/pyro_dev/pyro/pyro/poutine/trace.py:13: UserWarning: Encountered NAN log_pdf at site 'z_75'
```

</details>

fehiepsi:
[...]

PR #734, opened 2018-01-31
Fix DMM example with --num-iafs > 0

### LINKED ISSUE #721: DMM Example  Model and guide dims disagree when IAF is applied
When the number of IAF is non zero, the latent dimension changes to the same value as the batch size.

neerajprad:
We needed some shape fixes to TransformedDistribution to fix the DMM example with IAF. There is still the issue with NAN loss - #711 that seems to have been introduced much earlier (I checked until November 2017). 

Fixes #721.

### DIFF
--- a/pyro/distributions/transformed_distribution.py
+++ b/pyro/distributions/transformed_distribution.py
@@ -55,8 +55,8 @@ class TransformedDistribution(Distribution):
             next_input = y
         return next_input
 
-    def shape(self, sample_shape=torch.Size(), *args, **kwargs):
-        return self.base_dist.shape(sample_shape, *args, **kwargs)
+    def shape(self, *args, **kwargs):
+        return self.base_dist.shape(*args, **kwargs)
 
     def batch_shape(self, *args, **kwargs):
         return self.base_dist.batch_shape(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


neerajprad:
@fritzo - Can I just remove the pyro `TransformedDistribution` and `Bijector` classes and use the corresponding PyTorch implementations? I will modify the `InverseAutoregressiveFlow` class to inherit off of torch `Bijector`. I will probably need that for HMC too.

neerajprad:
> I agree the automated example discovery logic is getting kludgey.

Its not that bad yet. :) Just added a TODO to keep in mind that if we are messing around with that part too much, it might be worth revisiting.

fritzo:
> Can I just remove the pyro `TransformedDistribution`

I am working on that in this branch https://github.com/uber/pyro/tree/torch-transformed-and-iaf
Feel free to take over the work if you like. I can open a PR if you want to discuss.

neerajprad:
> I am working on that in this branch https://github.com/uber/pyro/compare/torch-transformed-and-iaf Feel free to take over the work if you like. I can open a PR if you want to discuss.

Yeah, I will need this for HMC. Unfortunately, we can't mix `RandomPrimitive` and other distributions, since `RandomPrimitive` instances can't support properties (we need to pass in distribution parameters to those methods). As such they are not interchangeable with torch distributions.

PR #736, opened 2018-02-02
Make distribution API consistent between Pyro and PyTorch

neerajprad:
**Changes:**
 - Removes `args/kwargs` piping from the Distribution base class, and makes all batch/event shape methods into properties. 
 - Changes the `copy_docs_from` annotation to also incorporate properties which have a read-only `__doc__` attribute.
 - `RandomPrimitive` works for most cases (see below) due to python duck typing, but this is an intermediate step before its deprecation.
 - `TransformedDistribution` uses the light PyTorch wrapper, and does not accept RandomPrimitive instances.

**Future Work:**
 - Deprecating `RandomPrimitive`.
 - Adding other properties like `support`, `params`, `mean`, `variance` etc. that already exist in `torch.distributions`.


pyro/distributions/util.py
```diff
@@ -22,10 +22,23 @@ def decorator(destin_class):
...
-                    destin_attr.__doc__ = source_doc
+                    destin_doc = source_doc
                 else:
-                    destin_attr.__doc__ = 'See :meth:`{}.{}.{}`'.format(
-                        source_class.__module__, source_class.__name__, source_attr.__name__)
+                    attr_name = source_attr.fget.__name__ if isinstance(source_attr, property) \
+                        else source_attr.__name__
+                    destin_doc = 'See :meth:`{}.{}.{}`'.format(
+                        source_class.__module__, source_class.__name__, attr_name)
+                if isinstance(destin_attr, property):
```
fritzo: Nicely done, this is becoming a pretty clever decorator :+1:

pyro/infer/abstract_infer.py
```diff
@@ -72,10 +72,15 @@ def _gen_weighted_samples(self, *args, **kwargs):
         raise NotImplementedError("_gen_weighted_samples is abstract method")
 
     def sample(self, *args, **kwargs):
+        sample_shape = kwargs.pop("sample_shape", None)
+        if sample_shape:
+            raise ValueError("Arbitrary `sample_shape` not supported by Distribution class.")
```
fritzo: nit: replace `Distribution` with `Histrogram`, right?

pyro/util.py
```diff
@@ -207,9 +208,14 @@ def enum_extend(trace, msg, num_samples=None):
         num_samples = -1
 
     # Batched .enumerate_support() assumes batched values are independent.
-    shape = broadcast_shape(msg["fn"].shape(*msg["args"], **msg["kwargs"]),
-                            msg["value"].size())
-    batch_dims = len(shape) - msg['fn'].event_dim(*msg["args"], **msg["kwargs"])
+    if isinstance(msg["fn"], RandomPrimitive):
```
fritzo: Wow I admit to creating this logic, and I see it is gross. I'll try to remove it asap.

pyro/infer/abstract_infer.py
neerajprad: I meant Distribution subclass so that I could use it uniformly in other places..but its better to be explicit. Will change to actual class names.

fritzo: But `Distribution`s do support arbitrary sample shape, don't they?

neerajprad: Yes, Distributions do; meant this particular Distribution sub-class does not. :) Never mind, changed to actual class names. We should fix the `sample_shape` in Rejectors separately.

pyro/util.py
neerajprad: I was using it myself in the tests, but fortunately, Pyro internals otherwise look fairly decoupled.

Issue #592, opened 2017-11-20
Add .transform property to constrained distributions for HMC, MAP


fritzo:
Blocked by #293 (needs vectorized `TransformedDistribution.log_det_jacobian()`) 

Algorithms like HMC and MAP inference need to be able to automatically update latent variables by some small step. When latent variables are constrained (e.g. drawn from a `Beta`, `Dirichlet`, `Uniform`, or `Exponential`) these algorithms need to be able to transform to an unconstrained parameter space. Must PPL systems do this by registering a `Bijector` or `Bijector` factory that can transform the constrained variable to an unconstrained space.

fritzo:
To avoid duplicating work, we could do this in the `torch.distributions` layer and wait until the wrapping layer matures.

fehiepsi:
It would be a nice feature. Waiting for this!

PR #740, opened 2018-02-03
Using torch transforms to run HMC on distributions with constrained support

neerajprad:
Addresses #592

This adds support for constrained distributions to Pyro's HMC implementation. This task is made simple by leveraging PyTorch 0.4's libraries for [Constraints](http://pytorch.org/docs/master/distributions.html#module-torch.distributions.constraints) and [Bijective Transforms](http://pytorch.org/docs/master/distributions.html#module-torch.distributions.transforms).

## WIP
 - [x] Existing tests need to use Distribution instances rather than RandomPrimitive instances.
 - [x] ~[Needs further investigation] The Bernoulli-Beta test in `test_hmc.py` currently gives a slightly biased estimate for the posterior latent mean.~ This gives expected results with higher samples. The earlier results were due to a bad seed initialization, and a smaller observation dataset.
 - [x] More tests for constrained distributions: `Exponential` or `Gamma` for `positive`, ~`Dirichlet` for `simplex`~ (Needs `.log_abs_det_jacobian` implementation for the StickingBreakingTransform).
 - [x] Investigate and fix slowness in NUTS - current implementation is unusable with NUTS.


neerajprad:
This PR will need updated PyTorch wheels for the test runs.

neerajprad:
@fritzo - I have uploaded the new CI wheels for PyTorch based on commit https://github.com/pytorch/pytorch/commit/ca5071d0721767fcfeb226b5c695dfd5d0671072. The one after this breaks the build scripts (issue being tracked on https://github.com/pytorch/pytorch/issues/5038).

pyro/ops/integrator.py
```diff
@@ -44,3 +34,38 @@ def _grad(potential_fn, z):
...
+        to ``z`` determines the rate of change of the corresponding sites'
+        momenta ``r``.
+    :param float step_size: step size for each time step iteration.
+    :param int num_steps: number of discrete time steps over which to integrate.
+    :param dict transforms: Optional dictionary that specifies a transform
+        for a sample site with constrained support to unconstrained space. The
+        transform should be invertible, and implement `log_abs_det_jacobian`.
+    :return tuple (z_next, r_next): final position and momenta, having same types as (z, r).
+    """
+    if transforms is None:
```
fritzo: I kind of liked the `transoforms={}` which would make this check become `if not transforms: ...`. Do you have a reason for preferring None?

pyro/ops/integrator.py
```diff
@@ -44,3 +34,38 @@ def _grad(potential_fn, z):
...
+        return unconstrained_velocity_verlet(z, r, potential_fn, step_size=step_size, num_steps=num_steps)
+
+    def unconstrained_potential_fn(u):
+        z = {name: transforms.get(name, identity_transform).inv(u_name) for name, u_name in u.items()}
```
fritzo: Do you think it's worth eliminating the dependency on `identity_transform` by expanding our logic a bit here and below?
```py
z = u.copy()
for name, transform in transforms.items():
    z[name] = transform(z[name])
```

pyro/ops/integrator.py
neerajprad: Using `{}` is fine since we are not mutating the default argument. My linter just gives me a warning when we use a mutable default argument, so I tend to not use them, even though the correct null representation for dict types should be an empty dict. Will change.

fritzo: OK, thanks.

And stop working on your vacation :smile:

pyro/ops/integrator.py
neerajprad: Unless we have some perf considerations (both seem like they should be about the same), I like the pattern of using the default identity transform above. We should change `u_name` to `unconstrained_value` or just `value` to make it clear that it is a tensor object.

neerajprad:
@fritzo - Thanks to all the awesome work on the PyTorch transforms, this was smooth. Please feel free to make changes to the PR, I think we might just need to add a few more tests. Thanks a lot for starting this, and all the help along the way!

fritzo:
@neerajprad Great, I'll add more tests this week. This was some really productive pair coding.

neerajprad:
The reason for why NUTS was slow is that we need to evaluate the energy (and the acceptance probability) in terms of the unconstrained potential energy. Earlier, we were using the unconstrained potential energy only inside the integrator while computing the potential gradient. However, this results in very different scales for `z` and `r` (besides other issues), and the stopping criterion for NUTS is never met.

As such, I have changed the design to keep the integrator the same, but instead modified the `potential_energy` function in the main `HMC` class itself. This ensures that we are doing minimal amount of back and forth transformations, and the `NUTS` class can just inherit this. The tests speeds are comparable between the two now (and transformations are quite fast).

pyro/infer/mcmc/hmc.py
```diff
@@ -28,31 +29,49 @@ class HMC(TraceKernel):
...
+    :param dict transforms: Optional dictionary that specifies a transform
+        for a sample site with constrained support to unconstrained space. The
+        transform should be invertible, and implement `log_abs_det_jacobian`.
+        If not specified and the model has sites with constrained support,
+        automatic transformations will be applied, as specified in
+        ``torch.distributions.constraint_registry``.
     """
 
-    def __init__(self, model, step_size=0.5, num_steps=3):
+    def __init__(self, model, step_size=0.5, num_steps=3, transforms=None):
```
neerajprad: Had to set this to None, as otherwise this was getting cached on different instances of NUTS in `test_nuts.py`.

neerajprad:
@fritzo - This is almost the same as what we had coded earlier, but just moving it to the HMC base class due to the reasons mentioned above. The test with the Dirichlet latent site does not yet work because we don't have the `log_abs_det_jacobian` implementation for the StickBreakingTransform. The other tests for NUTS and HMC work well. Let me know your thoughts.

cc. @fehiepsi

pyro/infer/mcmc/hmc.py
```diff
@@ -28,31 +29,49 @@ class HMC(TraceKernel):
     :param int num_steps: The number of discrete steps over which to simulate
         Hamiltonian dynamics. The state at the end of the trajectory is
         returned as the proposal.
+    :param dict transforms: Optional dictionary that specifies a transform
+        for a sample site with constrained support to unconstrained space. The
+        transform should be invertible, and implement `log_abs_det_jacobian`.
+        If not specified and the model has sites with constrained support,
+        automatic transformations will be applied, as specified in
+        ``torch.distributions.constraint_registry``.
```
fritzo: Pyro docs now link directly to PyTorch documentation. Can you update this link to
```
:module:`torch.distributions.constraint_registry`
```
(and maybe `make docs` and ensure the link works)

neerajprad: Great! Will do.

pyro/infer/mcmc/hmc.py
```diff
@@ -28,31 +29,49 @@ class HMC(TraceKernel):
...
+        for name, transform in self.transforms.items():
+            z_constrained[name] = transform.inv(z_constrained[name])
         z_trace = self._prototype_trace
-        for name, value in z.items():
+        for name, value in z_constrained.items():
             z_trace.nodes[name]['value'] = value
         trace_poutine = poutine.trace(poutine.replay(self.model, trace=z_trace))
         trace_poutine(*self._args, **self._kwargs)
-        return trace_poutine.trace
+        return trace_poutine.trace, z_constrained
```
fehiepsi: I think that it is better to move this calculation for z_constrained to `potential_energy` method for three reasons:
- Keep the old API of `_get_trace`
- Avoid unnecessarily calculate `z_constrained` in the last step of `sample`.
- We can safely change the name `unconstrained_potential_energy` to `_potential_energy` without misunderstanding.

fritzo: @neerajprad do you want to make this change in this PR, or should I merge now and wait for a follow-up PR?

neerajprad: @fehiepsi - We will be storing unconstrained values in the traces, in that case. So we will need to run the inverse transforms before we return the trace. I am happy to make that change, it might make things a bit more explicit.

neerajprad: @fritzo - It's a small change; let me address it right now itself.

fehiepsi: Look great now!

fehiepsi:
@neerajprad When I change `step_size` of `test_hmc.test_normal_gamma` to 0.03, I encounter `NaN` error. I wonder if this phenomenon is expected?

related: #849

neerajprad:
> @neerajprad When I change step_size of test_hmc.test_normal_gamma from 0.01 to 0.03, I encounter NaN error. I wonder if this phenomenon is expected?

Taking a look, @fehiepsi.

neerajprad:
@fehiepsi - Unfortunately, it does seem like with large step-size we can get a NaN, due to finite precision. In this case the gradient for the `latent_p[1]` oscillates to very high values (as it is initially far from the true posterior), and when the unconstrained value is ~ `-1400`, it gets mapped to `0` by the exp transform, which gives `NaN` when we try to compute its log probability under Gamma. Please feel free to skip/adjust this test for dual averaging in #849, but we should look into how best to handle cases like these.

Can we get this PR merged in the meantime? Lets handle edge cases / further refinements in a separate PR.

neerajprad:
@fritzo - Does it make sense to add some clamping factors to ensure that the transform specified by `biject_to` when applied to any real in the [min, max] range for the tensor type returns a scorable sample? I think this may be useful more generally.

fritzo:
> Does it make sense to add some clamping factors

What error are you trying to fix? I'm not sure where it would be best to clamp. On the one hand we want to correctly handle endpoints for e.g. categorical `probs=[0,1]`; on the other hand we want to be able to solve ODEs in a constrained space.Maybe it makes most sense to add these clamps to HMC rather than upstream in torch.distributions.

Also, can why we're getting NANs rather than `inf`? It seems like proper behavior should be to get `inf` and reject the MH move. This seems like a cleaner solution IMHO.

neerajprad:
> It may be cleanest to fix Gamma(alpha,beta).log_prob(0) to be either 0 or inf depending on alpha

I'll need to check if this will solve the issue, as we need the gradient to also not be NaN. 

@fehiepsi - If we return something like `Inf`, that should be sufficient for us to inform the correct step_size in #849, right?

neerajprad:
@fritzo - Can we get this merged? That'll make reviewing #849 easier. We can address the numerical stability issue separately.

fehiepsi:
@neerajprad I think we can merge this and deal with edge cases in a separate pull request to make things clean as you suggested. For `Inf`, I guess it is a valid value in MCMC.

PR #743, opened 2018-02-06
Fix decision rule in BlockPoutine

### LINKED ISSUE #619: Default guide for `Importance` gives non-zero `log_weight`
The default `guide` in `Importance` blocks _all_ sites (and not just those observed). This leads to the `guide_trace` being empty (except for the input and return values). As a result, no samples are reused when replayed on the model _and_ the `guide_trace.log_pdf()` evaluates to `0.0`. The `log_weight` is then equal to the `model_trace.log_pdf()` (which is also evaluated on different samples), which I believe is unintended and incorrect.

The program below illustrates this for a simple univariate Gaussian, where my proposal and target distribution are identical and I would expect the log of the weights to be `0.0`. The latter is only the case when the sites are _explicitly_ exposed.

```python
import torch

import pyro
from pyro import distributions as dist
from pyro import infer
from pyro import poutine


def gaussian():
    return pyro.sample('x', dist.normal,
                       Variable(torch.Tensor([0.0])),
                       Variable(torch.Tensor([1.0])))


# Using `Importance` with the default `guide`, the `log_weight` is equal to the
# `model_trace.log_pdf()`. That is, the `guide_trace.log_pdf()` (evaluated
# internally) is incorrectly `0.0`.
print('importance_default_guide:')
importance_default_guide = infer.Importance(gaussian, num_samples=10)
for model_trace, log_weight in importance_default_guide._traces():
    model_trace_log_pdf = model_trace.log_pdf()
    are_equal = log_weight.data[0] == model_trace_log_pdf.data[0]
    print(log_weight.data[0], are_equal)

# However, setting the `guide` to expose `x` ensures that it is replayed so
# that the `log_weight` is exactly zero for each sample.
print('importance_exposed_guide:')
importance_exposed_guide = infer.Importance(
    gaussian,
    guide=poutine.block(gaussian, expose=['x']),
    num_samples=10)
for model_trace, log_weight in importance_exposed_guide._traces():
    print(log_weight.data[0])
```
```
importance_default_guide:
-0.9368391633033752 True
-1.3421428203582764 True
-0.9189755320549011 True
-2.1423826217651367 True
-2.301940679550171 True
-1.142196774482727 True
-0.9449963569641113 True
-2.7146053314208984 True
-3.420013904571533 True
-1.7994171380996704 True
importance_exposed_guide:
0.0
0.0
0.0
0.0
0.0
0.0
0.0
0.0
0.0
0.0
```

eb8680:
`BlockPoutine` was incorrectly hiding sample sites when given `hide_types=["observe"]`, breaking the default guide in `infer.Importance`.  This fixes that bug and adds a test that would have caught it.

Fixes #619

### DIFF
--- a/pyro/poutine/block_poutine.py
+++ b/pyro/poutine/block_poutine.py
@@ -49,7 +49,7 @@ class BlockPoutine(Poutine):
         1. msg["name"] in hide
         2. msg["type"] in hide_types
         3. msg["name"] not in expose and msg["type"] not in expose_types
-        4. hide_all == True
+        4. hide_all == True and hide, hide_types, and expose_types are all None
         """
         super(BlockPoutine, self).__init__(fn)
         # first, some sanity checks:
@@ -71,8 +71,12 @@ class BlockPoutine(Poutine):
         # hide_types and expose_types intersect?
         if hide_types is None:
             hide_types = []
+        else:
+            hide_all = False
         if expose_types is None:
             expose_types = []
+        else:
+            hide_all = False
         assert set(hide_types).isdisjoint(set(expose_types)), \
             "cannot hide and expose a site type"
 
--- a/pyro/poutine/block_poutine.py
+++ b/pyro/poutine/block_poutine.py
@@ -65,6 +65,9 @@ class BlockPoutine(Poutine):
 
         if expose is None:
             expose = []
+        else:
+            hide_all = True
+
         assert set(hide).isdisjoint(set(expose)), \
             "cannot hide and expose a site"
 
@@ -73,10 +76,12 @@ class BlockPoutine(Poutine):
             hide_types = []
         else:
             hide_all = False
+
         if expose_types is None:
             expose_types = []
         else:
-            hide_all = False
+            hide_all = True
+
         assert set(hide_types).isdisjoint(set(expose_types)), \
             "cannot hide and expose a site type"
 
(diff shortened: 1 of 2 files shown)


pyro/poutine/block_poutine.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, fn,
         1. msg["name"] in hide
         2. msg["type"] in hide_types
         3. msg["name"] not in expose and msg["type"] not in expose_types
-        4. hide_all == True
+        4. hide_all == True and hide, hide_types, and expose_types are all None
```
fritzo: I have to think really hard to guess what combinations of arguments are valid. Do you think it would be reasonable to eliminate some args? E.g. `hide_all=True` should be equivalent to `expose=[]`, and `expose_all=True` should be equivalent to `hide=[]`. Maybe `hide` would be better named `hide_only`?

Alternatively, what do you think of adding a stricter sanity checking block at the the start of the function, e.g.
```py
if hide_all and expose_all:
    raise ValueError('Cannot set both hide_all=True and expose_all=True")
if hide_all:
    if hide is not None:
        raise ValueError("Cannot set both hide_all and hide")
    if hide_types is not None:
        raise ValueError("Cannot set both hide_all and hide_types")
if expose_all:
    if expose is not None:
        raise ValueError("Cannot set both hide_all and hide")
    if expose_types is not None:
        raise ValueError("Cannot set both hide_all and hide_types")
```
This would at least make it clearer to developers which arg combos are valid.

eb8680:
> Do you think it's worth clarifying the meaning of arguments?

Yes, and it probably makes more sense to move the arguments and logic into a separate function that returns a boolean, like the way `EscapePoutine` works.  I'll leave that for a separate PR.

fritzo:
Ready to merge?

eb8680:
Yep

PR #744, opened 2018-02-06
Ensure that TransformedDistribution is reparameterized

fritzo:
This fixes a bug in Pyro's `TransformedDistribution` that uses PyTorch `TransformedDistribution` in a clever way that happens to work. It fixes plumbing to correctly translate between PyTorch properties and Pyro properties (e.g. `.reparameterized` vs `.has_rsample`).

This is needed especially after https://github.com/pytorch/pytorch/pull/5086 which ensures that PyTorch `TransformedDistrubution.sample()` is not accidentally reparameterized.

This also fixes a bug in `@copy_docs_from()` and fixes docstring formatting in `SparseMultivariateNormal`.


pyro/distributions/torch/sparse_multivariate_normal.py
```diff
@@ -70,8 +70,9 @@ def log_prob(self, value):
     def _compute_logdet_and_mahalanobis(self, D, W, y, trace_term=0):
         """
         Calculates log determinant and (squared) Mahalanobis term of covariance
-        matrix (D + Wt.W), where D is a diagonal matrix, based on the
-        "Woodbury matrix identity" and "matrix determinant lemma":
+        matrix ``(D + Wt.W)``, where ``D`` is a diagonal matrix, based on the
+        "Woodbury matrix identity" and "matrix determinant lemma"::
```
jpchen: nit: itd be nice to put a link here to the [wiki](https://en.wikipedia.org/wiki/Woodbury_matrix_identity)

fritzo:
@jpchen would you mind merging this?

pyro/distributions/torch/sparse_multivariate_normal.py
fritzo: meh it's a mere google search away

PR #747, opened 2018-02-07
[Done] Implement sparse variational GP

fehiepsi:
I am trying to implement the variational versions of GP models in this pull request. I will try to keep the same interface with GPR and SGPR models.

- [x] Implement sparse variational GP
- [x] Implement variational GP
- [x] Add Likelihood module
- [x] Add some tests / examples

To reviewers: During implementing GP models using Pyro, I understand more about both GP and Pyro. After finishing this pull request, I think it is a good moment to refactor/remove redundant codes. Below are something I will change (at least). So please ignore those points while reviewing, they will be addressed in the next pull request. 
- Remove `kernel_prior`, `Xu_prior` arguments in models. Instead, it is better to use a method `set_prior(param_name, param_distribution)`.
- Remove MAP guide in model, move MAP guide to `Kernel` module (if that param has prior).
- Remove InducingPoints module. I created this module to be able to set prior to the parameter `Xu`. With the `set_prior` method, we can remove it.


fehiepsi:
After reading some literature carefully, I think that the current implementation is different. I will change it.

fehiepsi:
@fritzo @ysaatchi I have finished implementing the models VGP and SVGP. It is fine to review **the approach** now if you have time, or you can wait until I finish all tasks.

In upcoming days, I will add likelihood module with a simple Gaussian likelihood to test them. Especially, I expect the performance of SVGP with Gaussian likelihood is similar to SGPR using VFE approximation.

fehiepsi:
With #767, I will refactor the code for svgp and vgp a bit. Then which is a good place to add examples @fritzo? Maybe at examples/contrib/gp or another notebook?

fehiepsi:
@fritzo @eb8680 I have finished the implementation of variational models. I also added a section on the tutorial on how to use sparse variational model. It is better to merge #767 first to make this pull request clean.

@ysaatchi Would you mind checking the algorithmic side of `svgp.py` file? I just wonder if using Monte Carlo with **one** sample for $\int_{p(f|u)} p(y|f)$ is good or not. In an above comment, I posted some papers which uses Monte Carlo to estimate this expectation.

If things are good, I will start to add more kernels and likelihoods in other pull requests. Maybe I will start to do it from the next week (I have a holiday this week).

pyro/contrib/gp/models/sgpr.py
```diff
@@ -80,64 +84,44 @@ def model(self):
...
-        :param bool full_cov: Predict full covariance matrix or just its diagonal.
-        :param bool noiseless: Include noise in the prediction or not.
-        :return: loc and covariance matrix of p(y*|Xnew).
+        :param bool full_cov: Predicts full covariance matrix or just its diagonal.
+        :param bool noiseless: Includes noise in the prediction or not.
+        :return: loc and covariance matrix of ``p(y*|Xnew)``.
         :rtype: torch.autograd.Variable and torch.autograd.Variable
         """
         if Xnew.dim() == 2 and self.X.size(1) != Xnew.size(1):
             assert ValueError("Train data and test data should have the same feature sizes.")
```
fritzo: This should be `raise ValueError`; assert will never raise.

fehiepsi: Ah, my mistake, thanks!

pyro/contrib/gp/models/svgp.py
```diff
@@ -0,0 +1,146 @@
...
+class SparseVariationalGP(Model):
+    """
+    Sparse Variational Gaussian Process module.
+
+    References
+
+    [1] `Scalable variational Gaussian process classification`,
+    James Hensman, Alexander G. de G. Matthews, Zoubin Ghahramani
+
+    :param torch.autograd.Variable X: A tensor of inputs.
```
martinjankowiak: there needs to be more info about the dimensions of things.... it seems you only allow the input to be 1 dimensional?

pyro/contrib/gp/models/vgp.py
```diff
@@ -0,0 +1,108 @@
...
+import pyro
+import pyro.distributions as dist
+from pyro.distributions.util import matrix_triangular_solve_compat
+
+
+class VariationalGP(nn.Module):
+    """
+    Variational Gaussian Process module.
+
+    This model can be seen as a special version of SparseVariationalGP model
```
martinjankowiak: since this is a special case why not have more code reuse?

pyro/contrib/gp/models/svgp.py
```diff
@@ -0,0 +1,146 @@
...
+        Kuf = kernel(Xu, self.X)
+        Luu = Kuu.potrf(upper=False)
+
+        zero_loc = Variable(Xu.data.new([0])).expand(self.num_inducing)
+        # TODO: use scale_tril=Luu
+        u = pyro.sample("u", dist.MultivariateNormal(loc=zero_loc, covariance_matrix=Kuu))
+
+        # p(f|u) ~ N(f|mf, Kf)
+        # mf = Kfu @ inv(Kuu) @ u; Kf = Kff - Kfu @ inv(Kuu) @ Kuf
+        pack = torch.cat((u.unsqueeze(1), Kuf), dim=1)
```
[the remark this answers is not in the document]
fehiepsi: Right now, `torch.trtrs` does not support grad or cuda, so we have to use `.inverse()` to solve. Combining will be much faster (because we don't have to take inverse two times). In case we use `trtrs`, the performance is similar.

pyro/contrib/gp/models/svgp.py
```diff
@@ -0,0 +1,146 @@
...
+
+    References
+
+    [1] `Scalable variational Gaussian process classification`,
+    James Hensman, Alexander G. de G. Matthews, Zoubin Ghahramani
+
+    :param torch.autograd.Variable X: A tensor of inputs.
+    :param torch.autograd.Variable y: A tensor of outputs for training.
+    :param pyro.contrib.gp.kernels.Kernel kernel: A Pyro kernel object.
+    :param torch.Tensor Xu: An inducing-point parameter.
```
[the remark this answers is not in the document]
fehiepsi: I agree. It would be helpful to provide some default initializations. We will address it in a future pull request.

pyro/contrib/gp/models/svgp.py
```diff
@@ -0,0 +1,146 @@
...
+        self.jitter = Variable(self.X.data.new([jitter]))
+
+    def model(self):
+        self.set_mode("model")
+
+        kernel = self.kernel
+        likelihood = self.likelihood
+        Xu = self.get_param("Xu")
+
+        Kuu = kernel(Xu) + self.jitter.expand(self.num_inducing).diag()
```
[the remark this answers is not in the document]
fehiepsi: If we use `torch.eye`, then we have to multiply with `jitter`. It would be much slower.

pyro/contrib/gp/models/svgp.py
fehiepsi: Yes, I will add some info there. Right now, the input is 2D (though 1D is also fine). However, the output is just a 1D tensor. Supporting multidimensional output will be addressed this in a future pull request.

pyro/contrib/gp/models/vgp.py
fehiepsi: Thanks, I will make vgp as a subclass of svgp.

fehiepsi:
@martinjankowiak  I just addressed some of your reviews. There might be some bugs because the inference in the tutorial for svgp is different after the change. I will check back when returns home.

pyro/contrib/gp/models/svgp.py
```diff
@@ -0,0 +1,104 @@
...
+        Luuinv_pack = matrix_triangular_solve_compat(pack, Luu, upper=False)
+        Luuinv_u = Luuinv_pack[:, 0]
+        Luuinv_Kuf = Luuinv_pack[:, 1:]
+
+        mf = Luuinv_Kuf.t().matmul(Luuinv_u)
+        Kffdiag = kernel(self.X, diag=True)
+        Qffdiag = (Luuinv_Kuf ** 2).sum(dim=0)
+        Kfdiag = Kffdiag - Qffdiag
+
+        f = dist.Normal(mf, Kfdiag).sample()
```
fehiepsi: This is the place I mention about Monte Carlo with 1 sample. Did I make it right?

pyro/contrib/gp/models/svgp.py
```diff
@@ -0,0 +1,105 @@
...
+        Luuinv_u = Luuinv_pack[:, 0]
+        Luuinv_Kuf = Luuinv_pack[:, 1:]
+
+        mf = Luuinv_Kuf.t().matmul(Luuinv_u)
+        Kffdiag = kernel(self.X, diag=True)
+        Qffdiag = (Luuinv_Kuf ** 2).sum(dim=0)
+        Kfdiag = Kffdiag - Qffdiag
+
+        # get 1 sample for f
+        f = dist.Normal(mf, Kfdiag)()
```
fehiepsi: This is the place I mentioned about Monte Carlo in the above comments. Did I make it right?

fehiepsi: This is the place I mention about Monte Carlo with 1 sample. Did I make it right?

fritzo:
> which is a good place to add examples ...? Maybe at examples/contrib/gp or another notebook?

Yes either of those are good places to add examples.

pyro/contrib/gp/models/svgp.py
fritzo: Looks correct to me.

martinjankowiak: yeah it's fine to take one sample... it's not guaranteed to work (just like for any stochastic optimization problem) but certainly it's a reasonable thing to do and should often be ok

pyro/contrib/gp/models/svgp.py
```diff
@@ -0,0 +1,105 @@
...
+from torch.nn import Parameter
+
+import pyro
+import pyro.distributions as dist
+from pyro.distributions.util import matrix_triangular_solve_compat
+
+from .vgp import VariationalGP
+
+
+class SparseVariationalGP(VariationalGP):
```
martinjankowiak: isn't your class hierarchy backwards here?

tutorial/source/gp.ipynb
```diff
@@ -263,14 +268,14 @@
...
      "text": [
-      "RBF$$$variance_unconstrained \n",
-      " 1.4176\n",
-      "[torch.FloatTensor of size 1]\n",
-      "\n",
       "RBF$$$lengthscale_unconstrained \n",
       " 2.7259\n",
       "[torch.FloatTensor of size 1]\n",
       "\n",
+      "RBF$$$variance_unconstrained \n",
```
martinjankowiak: this plot doesn't seem to have converged to a reasonable result?

fehiepsi: Yes, because I initiate lengthscale too large (comparing to the range of our data `(0, 5)`). I just thought that it is boring to see similar plots (although we use different ways: set priors / not set priors), so I spent that section mainly to introduce the notion of constraint and how to get unconstraint value using `get_param()`. How about I just remove the plots in that section?

martinjankowiak: alternatively you can just add a comment that the lengthscale is too large etc etc

pyro/contrib/gp/models/svgp.py
fehiepsi: Yes, although VariationGP is a special case of sparse version (`Xu = X`), there is a different point (`Xu` is a parameter in sparse version). We can think that Sparse method is an **extended** version of VariationGP (by introducing a new parameter `Xu`). I found that it is simpler to implement this way. Either way, we have to rewrite `model` and `guide`.

tutorial/source/gp.ipynb
fehiepsi: @martinjankowiak thanks, I just added a comment about that :)

pyro/contrib/gp/models/svgp.py
martinjankowiak: did you end up with much of any actual code reuse?

pyro/contrib/gp/models/svgp.py
```diff
@@ -0,0 +1,105 @@
...
+
+    References
+
+    [1] `Scalable variational Gaussian process classification`,
+    James Hensman, Alexander G. de G. Matthews, Zoubin Ghahramani
+
+    :param torch.autograd.Variable X: A 1D or 2D tensor of inputs.
+    :param torch.autograd.Variable y: A 1D tensor of outputs for training.
+    :param pyro.contrib.gp.kernels.Kernel kernel: A Pyro kernel object.
+    :param torch.Tensor Xu: An inducing-point parameter.
```
martinjankowiak: you should make it clear this is an initial value

tutorial/source/gp.ipynb
```diff
@@ -636,7 +699,10 @@
     "## Reference\n",
     "\n",
     "[1] `Deep Gaussian processes and variational propagation of uncertainty`,<br />&nbsp;&nbsp;&nbsp;&nbsp;\n",
-    "Andreas Damianou"
+    "Andreas Damianou\n",
+    "\n",
+    "[2] `A unifying framework for sparse Gaussian process approximation using power expectation propagation`,<br />&nbsp;&nbsp;&nbsp;&nbsp;\n",
+    "Thang D. Bui, Josiah Yan, and Richard E. Turner"
```
martinjankowiak: can you please add more references?

fehiepsi: yup :D

fehiepsi: @martinjankowiak I have added the main references I used for implementations. If there are more, I will add later.

Issue #752, opened 2018-02-08
Remove RandomPrimitive


fritzo:
## Why?

- Motivated by moving closer to PyTorch distributions https://github.com/uber/pyro/issues/732#issuecomment-362454322
- Would allow a fix for #750 i.e. calling `site["fn"].size()`
- PyTorch distributions now cache things, and RandomPrimitive discards the cache
- RandomPrimitive is not compatible with properties
- PyTorch distributions are expensive to construct; `RandomPrimitive` constructs them repeatedly

## Tasks

- [x] #754 Make `sample_shape` a parameter of `Distribution.log_prob()` and `.enumerate_support()`
- [x] #759 Update profiling scripts
- [x] #757 Update tests (except distribution fixture)
- [x] #761 Update all examples
- [x] #762 Update distribution test fixture
- [x] #755 Update all tutorials
- [ ] #764 Remove  `RandomPrimitive`

eb8680:
Do we still need our distribution tests after this?

fritzo:
> Do we still need our distribution tests after this?

Some are redundant and can safely be removed. Others should remain, e.g. tests for `Rejector`, `Delta`, our `TorchDistribution` wrapper, our `SparseMultivariateNormal`, ...

PR #753, opened 2018-02-08
Support sample_shape arg in .log_prob, .score_parts, .enumerate_support

fritzo:
Addresses #752 

## Why?

@eb8680 convinced me that `sample_shape` should at least be an argument to `.enumerate_support()` since those methods return almost the same kind of object. Additionally, we want `**kwargs` to apply equally to `.sample()`, `.log_prob()` and `.score_parts()`, so for symmetry `sample_shape` should be supported. This makes some sense because `.log_prob()` and `.score_parts()` can conceivably `.expand()` their results by the given `.sample_shape` for proper broadcasting.

More immediately, this is required to replace `RandomPrimitive`s with `Distribution` objects in some of our tests.

## Tested

- [ ] Add tests for `.log_prob()`
- [ ] Add tests for `.score_parts()`
- [ ] Add tests for `.enumerate_support()`

### DIFF
--- a/pyro/__init__.py
+++ b/pyro/__init__.py
@@ -132,7 +132,7 @@ class _Subsample(Distribution):
             result = Variable(torch.randperm(self.size)[:self.subsample_size])
         return result.cuda() if self.use_cuda else result
 
-    def log_prob(self, x):
+    def log_prob(self, x, sample_shape=torch.Size()):
         # This is zero so that iarange can provide an unbiased estimate of
         # the non-subsampled log_prob.
         result = Variable(torch.zeros(1))
--- a/pyro/distributions/delta.py
+++ b/pyro/distributions/delta.py
@@ -38,18 +38,15 @@ class Delta(Distribution):
         shape = sample_shape + self.v.size()
         return self.v.expand(shape)
 
-    def log_prob(self, x):
-        v = self.v
-        v = v.expand(broadcast_shape(self.shape(), x.size()))
+    def log_prob(self, x, sample_shape=torch.Size()):
+        v = self.v.expand(broadcast_shape(sample_shape + self.v.shape, x.shape))
         return torch.eq(x, v).float().log()
 
-    def enumerate_support(self, v=None):
+    def enumerate_support(self, sample_shape=torch.Size()):
         """
         Returns the delta distribution's support, as a tensor along the first dimension.
 
-        :param v: torch variable where each element of the tensor represents the point at
-            which the delta distribution is concentrated.
         :return: torch variable enumerating the support of the delta distribution.
         :rtype: torch.autograd.Variable.
         """
-        return Variable(self.v.data.unsqueeze(0))
+        return Variable(self.sample(sample_shape).data.unsqueeze(0))
--- a/pyro/distributions/distribution.py
+++ b/pyro/distributions/distribution.py
@@ -170,7 +170,7 @@ class Distribution(object):
         raise NotImplementedError
 
     @abstractmethod
-    def log_prob(self, x):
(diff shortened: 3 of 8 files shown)


neerajprad:
> Additionally, we want **kwargs to apply equally to .sample(), .log_prob() and .score_parts(), so for symmetry sample_shape should be supported. 

Are there other work-arounds that we can explore than having to pass these to all methods? Or maybe we can even pass these as no-op args for the time being (until we can remove it), so as to prevent people writing code that starts making use of this argument. 

> This makes some sense because .log_prob() and .score_parts() can conceivably .expand() their results by the given .sample_shape for proper broadcasting.

`.log_prob` does not need to know `sample_shape` and should be able to broadcast the parameters as needed to score any sample. From what I understand, this then allows us to generate a single or a batch of samples and expand the resulting `log_prob` score to some arbitrary `sample_shape`. Do you think this is something that the distribution class needs to support? If it is not a common use case, this is something that the users can do through a one liner instead, so that we do not need to change the Distribution API, and break consistency with PyTorch. Unless we have some motivating examples for why we need this and can explain it easily, I think this will make for a confusing API. What do you think?


```
dist = Normal(mean=torch.ones(2, 2), std=torch.ones(2, 2))
x = dist.sample(sample_shape=torch.Size((3, 3))   # returns tensor of size (3, 3, 2, 2)
log_prob = dist.log_prob(x)   # returns tensor of size (3, 3, 2, 2)

## Current proposal - do we need this as part of the Distribution API?
x = dist.sample()
log_prob = dist.sample(x, sample_shape=torch.Size((3, 3))  # copies the value 
 # above and expands it to shape (3, 3) 
```

eb8680:
> Are there other work-arounds that we can explore than having to pass these to all methods?

One of the basic assumptions in Pyro's current design is that distributions at sample sites take the same arguments when calling `sample` and `log_prob`.  This is currently enforced by `RandomPrimitive`, whose removal creates an asymmetry between `sample` and the other `Distribution methods.  This asymmetry must be dealt with somewhere in Pyro - it can either be in `TorchDistribution`, in `pyro.sample`, spread throughout `poutine`, in `Trace` somewhere, or some combination of those.  Handling it in `TorchDistribution` by adding null arguments to preserve the symmetry seemed like one reasonable solution when I discussed it with @fritzo but I'm certainly open to alternatives.

neerajprad:
> This is currently enforced by RandomPrimitive, whose removal creates an asymmetry between sample and the other Distribution methods. 

I thought that the deprecation of `RandomPrimitive` would make it easier to not have to pipe in arbitrary *args/**kwargs to distribution method calls, since the method signatures (while different between `sample` or `log_prob`), are consistent for all distribution classes in the absence of distribution parameters. I could be missing some finer points that necessitate this though. 

> Handling it in TorchDistribution by adding null arguments to preserve the symmetry seemed like one reasonable solution when I discussed it with @fritzo but I'm certainly open to alternatives.

I am on board with whatever you and @fritzo think is a good choice to make progress on #752. My concern was that one of the motivations for deprecating `RandomPrimitive` was so that we could slowly move to start using `torch.distributions` directly in Pyro (i.e. remove the `TorchDistribution` wrapper eventually, or have a very light wrapper with maximal compatibility) and this is taking us a bit in the other direction by tacking on some of the incompatibilities to the `TorchDistribution` wrapper. 

> Another good option is dealing with sample_shape in Trace as part of a refactor to systematize the various helper methods there.

This sounds like a good option. I will need to take a deeper look at our internals to have a more informed view of the alternatives available, so I would defer to you and @fritzo for any changes that we need to make for the 0.2 release. I will take a look at this as I get some time.

fritzo:
Closing in favor of #754

Issue #23, opened 2017-06-27
Update Distribution interface


ndgAtUber:
per (pleasant, unheated) discussion with @eb8680, we could consider upgrading distribution objects to be callables (i.e. functions) that return a sample, and also have a scoring function attached. in particular, we could allow them to have args:

~~~
>>binomial(ps)  
[1]
>>binomial.sample(ps)
[1]
>>binomial.log_pdf(ps,[1])
ln(0.5)
~~~

there are reasons to like this: it simplifies distributions (don't need to construct and then use), and makes it more clear that stateful dists (eg CRP) are the ones that require a constructor. (presumably if we go this route, we will change the `Binomial` class to a provided `binomial` function.)

on the other hand, it requires making the signature to `pyro.sample` and `pyro.observe` more complex: they must take the fn args as input. 

are there other complications to worry about?

if we do make this change, should a distribution automatically call pyro.sample when it is called in the context of a pyro program? i.e. should `pyro.sample(binomial, args)` be writable as `binomial(args)`?

PR #754, opened 2018-02-08
Prepare to drop RandomPrimitive in Trace methods

fritzo:
Addresses #752 more cleanly than #753 

This introduces the assumption in Pyro that: if Pyro ever uses the `.log_prob()` or `.score_parts()` methods of a stochastic function, then that function agrees to follow the `Distribution` api. Specifically, `.log_prob()` and `.score_parts()` take a `sample` and no `sample_shape` or other `*args, **kwargs`.

A special exception is made for `RandomPrimitive` since that does not conform to the `Distribution` api; this exception will be removed as soon as `RandomPrimitive` is removed.

This PR is needed to replace `dist.normal` etc. with `dist.Normal` etc. in tests.

### DIFF
--- a/pyro/poutine/trace.py
+++ b/pyro/poutine/trace.py
@@ -7,7 +7,7 @@ import networkx
 import numpy as np
 from torch.autograd import Variable
 
-from pyro.distributions import RandomPrimitive
+from pyro.distributions.torch_wrapper import TorchDistribution
 
 
 def _warn_if_nan(name, value):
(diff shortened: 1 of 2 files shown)


pyro/poutine/trace.py
```diff
@@ -81,8 +82,10 @@ def log_pdf(self, site_filter=lambda name, site: True):
                     site_log_p = site["log_pdf"]
                 except KeyError:
                     args, kwargs = site["args"], site["kwargs"]
-                    site_log_p = site["fn"].log_prob(
-                        site["value"], *args, **kwargs).sum() * site["scale"]
+                    if isinstance(site["fn"], RandomPrimitive):
+                        site_log_p = site["fn"].log_prob(site["value"], *args, **kwargs).sum() * site["scale"]
+                    else:
+                        site_log_p = site["fn"].log_prob(site["value"]).sum() * site["scale"]
```
eb8680: `TorchDistribution` is the special case here.  This should be:
```
if isinstance(site["fn"], TorchDistribution):
    site_log_p = site["fn"].log_prob(site["value"]).sum() * site["scale"]
else:
    site_log_p = site["fn"].log_prob(site["value"], *args, **kwargs).sum() * site["scale"]
```

fritzo: Ok, Done.

I was hoping to remove this kludgey `isinstance()` logic when we remove `RandomPrimitive`. I think this would be possible if you refactored `Histogram` to be a distribution factory rather than an `Distribution`. That way our `Distribution` interface could be narrower and closer to the PyTorch `Distribution` interface, and `Histogram` would only differ from it's current behavior by a `functools.partial`. This would make Neeraj's and my lives easier. What do you think?

fritzo:
Thanks for reviewing! Ready to merge if tests pass :crossed_fingers:

pyro/poutine/trace.py
eb8680: I'd have to see the proposed changes to `Histogram` but it's probably fine

eb8680: See also #23 for a much older discussion about this

Issue #756, opened 2018-02-09
Sparse GP tests fail nondeterministically


fritzo:
The following test fails when run alone:
```sh
pytest -v tests/contrib/gp/test_models.py::test_forward_sgpr   # FAILS
```
however it passes when run as part of the whole file
```sh
pytest -v tests/contrib/gp/test_models.py  # PASSES
```
As a result travis-ci tests pass but local `make test` fails (because it parallelizes)

fritzo:
@fehiepsi could you take a look if you have time? I'm not sure why the test fails when run alone.

fehiepsi:
@fritzo I fixed the bug at #758. Have no idea why it passes when run as a whole @@. For the bug: I have used `torch.ones(2)` instead of `torch.Tensor([2])` >"<.

fritzo:
Closed by #758

PR #758, opened 2018-02-09
Fix the bug #756 at sparse test

fehiepsi:
This pull request fixes the bug #756.

It also fixes another bug: not `clear_param_store` after each test.

@fritzo


fritzo:
@fehiepsi would you mind also adding to this PR a comprehensive `pyro.clear_param_store()` as follows:
```diff
diff --git a/tests/conftest.py b/tests/conftest.py
index 2f718020..848b7325 100644
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,6 +13,7 @@ def pytest_configure(config):
 
 
 def pytest_runtest_setup(item):
+    pyro.clear_param_store()
     test_initialize_marker = item.get_marker("init")
     if test_initialize_marker:
         rng_seed = test_initialize_marker.kwargs["rng_seed"]
```
I believe if we had had this in place, you would not have faced this difficult-to-diagnose error.

fehiepsi:
@fritzo You are right. It took me a couple of hours to realize that I didn't clear the store. The test passed for each individual test but not for the whole file.

PR #676, opened 2018-01-12
Switch to using PyTorch distributions by default

fritzo:
This flips the switch to start using PyTorch distributions by default, only if they are available. If they are not available (i.e. if an older PyTorch version is installed), then the old Pyro distributions will be used instead.

Note that a few Pyro distributions are currently being implemented in PyTorch, and are not switched in this PR:
- Binomial: Alican
- Multinomial: Alican
- MultivariateNormal: Brooks
- HalfCauchy: JP
- Poisson: Rachit
- LogNormal: Fritz
- Delta: may not need a PyTorch wrapper at all

PR #763, opened 2018-02-10
Thin-out the Pyro wrappers around torch distributions

fritzo:
Addresses #676 

1. Simplify `TransformedDistribution` to only work with wrapped `TorchDistributions`. After this PR `TransformedDistribution` simply unwraps the `base_dist` rather than overriding methods.
2. Add `.analytic_mean()` and `.analytic_var()` methods that alias PyTorch `.mean`, `.variance` properites
3. Remove unused imports and overrides


fehiepsi:
@fritzo Would you mind adding `.analytic_mean` to Delta distribution in this pull request? I need to use this method to get the mean of a distribution (instead of `.torch_dist.mean`, which is not available for Delta distribution).

PR #765, opened 2018-02-10
Add Parameterized module for GP

fehiepsi:
This pull request adds a Parameterized class for GP module. With this class, we can remove "lengthy" blocks using `pyro.random_module(...)` in GP models. In addition, from now on, we can set `constraints` to parameters of any (kernel/model/likelihood) module by using `set_constraint(...)` method (as suggested by @fritzo at https://github.com/uber/pyro/issues/702#issuecomment-363473322).

I make this separated from https://github.com/uber/pyro/pull/747 for easy to have a discussion. While creating Likelihood module there, I don't want to repeatedly require `likelihood_prior`, `pyro.random_model("likelihood", self.likelihood, ...)`... as the way I did for `self.kernel` and `self.Xu`. With this pull request, the logic in `self.model` and `self.guide` will be controlled by using `self.kernel.set_mode("model")` and `self.kernel.set_model("guide")` e.g.!

### DIFF
--- a/pyro/contrib/gp/__init__.py
+++ b/pyro/contrib/gp/__init__.py
@@ -23,7 +23,7 @@ class InducingPoints(nn.Module):
 
 class Parameterized(nn.Module):
     """
-    Parameterized module.
+    Parameterized class.
 
     This is a base class for other classes in this Gaussian Process module.
     """
@@ -49,7 +49,7 @@ class Parameterized(nn.Module):
         Sets a constraint to a parameter.
 
         :param str param: Name of a parameter.
-        :param torch.distributions.constraints.Constraint constraint: A pytorch constraint.
+        :param torch.distributions.constraints.Constraint constraint: A Pytorch constraint.
             See `Pytorch's docs
             <http://pytorch.org/docs/master/distributions.html#module-torch.distributions.constraints>`_
             for a list of constraints.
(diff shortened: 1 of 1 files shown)


pyro/contrib/gp/__init__.py
```diff
@@ -13,3 +19,64 @@ def __init__(self, Xu, name="inducing_points"):
...
     def forward(self):
         return self.inducing_points
+
+
+class Parameterized(nn.Module):
+    """
+    Parameterized module.
+
+    This is a base class for other classes in this Gaussian Process module.
+    The method `set_prior(param, prior)` will be used to set priors to parameters.
```
fritzo: Could you move these descriptions into docstrings of methods? I think this will look better in the html docs. When writing new docs I usually `make docs && open docs/build/html/index.html` to see how they look.

fehiepsi: Ah, thanks for the tip! Previously, I just look at docs of other codes to figure out how things will be rendered.

fehiepsi:
@fritzo I am glad that you like this class. So I think it is a good chance to refactor GP module to demonstrate how to use `set_mode()`. For #747,  it is better to be served for algorithms only.

fehiepsi:
@fritzo I close this in favor of #767.

PR #767, opened 2018-02-11
Refactor GP module

fehiepsi:
This pull request refactors GP module in favor of #765. Lots of codes are removed in models, the tutorial is updated to newest changes. I also make docs look better. Especially, use Parameterized class introduced at #765 , we can
- Fix a param (so it will not be learned)
- Set prior to a param
- Set constraint to a param.

The only problem I got is I can't replicate the result of the Sparse model in the tutorial. I have checked things carefully, the problem is not lying at this refactoring, but at Pyro's internal changes. I checked each pull request and saw that the behavior of tutorial is different after #695 get merged. I guess there is some things change in that pull request?

@fritzo


eb8680:
@fehiepsi I can help troubleshoot, can you explain what you think is going wrong?

fehiepsi:
@eb8680 Thanks! In the current dev version of [GP tutorial](https://github.com/uber/pyro/blob/dev/tutorial/source/gp.ipynb), at sparse approximation part, we can see that variance is small at inducing points and large at the other points. When I run the tutorial again with the dev branch, the plot is similar to what it looks in this pull request (we can't recognize the differences of variances between inducing points and remaining points). So I checkouted each commit from git log, installed Pyro again and again to see which commit made things change. It turns out that the change happened at the pull request #695. I am not sure that things work better or not after that pull request. I just can't replicate the old behavior. Maybe the inference works better now (because the model learnt the mean pretty well!!).

eb8680:
Thanks, I'll look into it, things shouldn't have changed.

fehiepsi:
@eb8680 I am really sorry if I have wasted your time to diagnose the problem. After checking the refactor of Poutine, I think that there is nothing change on the inference due to that pull request. It's a great refactoring (thanks to it, I understand more about the internal process of Pyro). The problem lies at the random nature of stochastic inference. I just gave a random seed at the beginning of the notebook, so when I delete some cells, change num_steps,... the later inference will change. I can replicate the old behavior of SGPR with some "random" changes (num_steps of GPR, which are run before SGPR in the notebook).

Reason for the good prediction of the mean: Before this pull request, I missed a `trace_term` for SparseMultivariateNormal', so VFE is just the same as DTC. With this pull request, we can see that VFE performs pretty well.

The problem is solved! Is there someone would like to review this?

pyro/contrib/gp/__init__.py
```diff
@@ -1,15 +1,136 @@
...
+        else:
+            return self._registered_params[param]
+
+    def fix_param(self, param, value=None):
+        """
+        Fixes a parameter to a specic value. If ``value=None``, fixes the parameter to the
+        default value.
+
+        :param str param: Name of a parameter.
+        :param torch.Tensor value: A tensor to be fixed to ``param``.
```
fritzo: PyTorch tensors will soon be deprecated. Could you change the input to be a detached Variable?
```py
def fix_param(self, param, value=None):
    if value is None:
        value = getattr(self, param).detach()
    self._fixed_params[param] = value
```

pyro/contrib/gp/__init__.py
```diff
@@ -1,15 +1,136 @@
...
+            self._registered_params[param] = self._fixed_params[param]
+            return
+        prior = self._priors[param] if param in self._priors else None
+        if self._name is None:
+            param_name = param
+        else:
+            param_name = pyro.param_with_module_name(self._name, param)
+
+        if prior is None:
+            constraint = self._constraints[param] if param in self._constraints else None
```
fritzo: nit: In a few places you can simplify conditional dict lookup using `.get()`:
```py
constraint = self._constraints.get(param)
```

pyro/contrib/gp/__init__.py
```diff
@@ -1,15 +1,136 @@
...
+        """
+        Sets mode for the module to be able to use its parameters in stochastic functions.
+
+        :param str mode: Either "model" or "guide".
+        """
+        if mode not in ["model", "guide"]:
+            raise ValueError("Mode should be either 'model' or 'guide', but got {}.".format(mode))
+        for param in self._parameters:
+            self._register_param(param, mode)
+        return self
```
fritzo: nit: It's kind of weird to `return self` here. I realize some codebases use method chaining heavily, but the patter is seldom used in Pyro.

pyro/contrib/gp/__init__.py
```diff
@@ -1,15 +1,136 @@
...
+        default value.
+
+        :param str param: Name of a parameter.
+        :param torch.Tensor value: A tensor to be fixed to ``param``.
+        """
+        if value is None:
+            value = getattr(self, param).data
+        self._fixed_params[param] = Variable(value)
+
+    def unfix_param(self, param):
```
fritzo: Is there an actual use case for this? If there's nothing immediate, maybe it's simpler to add it when we need it later on.

pyro/contrib/gp/__init__.py
```diff
@@ -1,15 +1,136 @@
...
+        super(Parameterized, self).__init__()
+        self._priors = {}
+        self._constraints = {}
+        self._name = None
+        self._registered_params = {}
+        self._fixed_params = {}
+
+    def set_prior(self, param, prior):
+        """
+        Sets a prior to a parameter. Use `set_prior(param, None)` to remove the prior distribution of
```
fritzo: Is there a use case for unsetting a prior? If there's no foreseeable use case, it would be better to not support this functionality.

pyro/contrib/gp/__init__.py
```diff
@@ -1,15 +1,136 @@
...
+        Registers a parameter to Pyro. It can be seen as a wrapper for `pyro.param()` and
+        `pyro.sample()` calls.
+
+        :param str param: Name of a parameter.
+        :param str mode: Either "model" or "guide".
+        """
+        if param in self._fixed_params:
+            self._registered_params[param] = self._fixed_params[param]
+            return
+        prior = self._priors[param] if param in self._priors else None
```
fritzo:
```py
prior = self._priors.get(param)
```

pyro/contrib/gp/__init__.py
```diff
@@ -1,15 +1,136 @@
...
-        return self.inducing_points
+    This is a base class for other classes in Gaussian Process.
+    By default, a parameter will be a ``torch.nn.Parameter`` containing ``torch.FloatTensor``.
+    To cast them to the correct data type or GPU device, we can call methods such as
+    ``.double()``, ``.cuda(device=0)``,...
+    See `torch.nn.Module
+    <http://pytorch.org/docs/master/nn.html#torch.nn.Module>`_ for more information.
+    """
+
+    def __init__(self):
```
fritzo: nit: What do you think of adding `name` as an `__init__` arg rather than setting `self._name = ...` in derived classes? This seems a little safer to me.

pyro/contrib/gp/__init__.py
fehiepsi: I agree, it just provides me a shortcut for `kernel = self.kernel.set_mode(...)`, which is unnecessary.

pyro/contrib/gp/__init__.py
fehiepsi: I didn't use it ^^! will remove it soon

pyro/contrib/gp/__init__.py
fehiepsi: The use case is when we want to reuse some modules, but are lazy to redefine it. ^^! Because it is not important (in my case, I don't use it), I will delete that part of docs.

pyro/contrib/gp/__init__.py
fehiepsi: yes, a detached variable doesn't require grad, which is the behavior I want. Thanks

fehiepsi:
@fritzo I have addressed your reviews. There is an additional change: when we call `set_mode()` method, it also automatically sets `mode` for submodules which are derived from `Parameterize` class. This behavior is similar to other Pytorch module's behaviors (`.train()`, `.eval()`, `.double()`, `.cuda()`,... So in the model, we just simply call `self.set_mode(...)` instead of several calls for its submodule (kernel, likelihood).

pyro/contrib/gp/__init__.py
```diff
@@ -1,15 +1,132 @@
...
+                unconstrained_param_0 = Variable(
+                    transform_to(constraint).inv(default_value).data.clone(),
+                    requires_grad=True)
+                p = transform_to(constraint)(pyro.param(unconstrained_param_name,
+                                                        unconstrained_param_0))
+        elif mode == "model":
+            p = pyro.sample(param_name, prior)
+        else:  # prior != None and mode = "guide"
+            MAP_param_name = param_name + "_MAP"
+            MAP_param_0 = Variable(prior.torch_dist.mean.data.clone(), requires_grad=True)
```
fritzo: This should be
```diff
- prior.torch_dist.mean
+ prior.analytic_mean()
```
but I'll address that in a follow-up PR.

Issue #732, opened 2018-01-31
Use PyTorch distributions directly in Pyro


fritzo:
@eb8680 suggested entirely removing the `TorchDistribution` wrapper class from Pyro. Here's a proposal for how we might do that:

## Refactoring Plan

- [x] #736 Change `.batch_shape` and `.event_shape` from methods to properties. (This is currently violated by `RandomPrimitive`, which requires workarounds).
- [x] #752 Remove `RandomPrimitive` and prefer `Dist(**params)` over `dist, **params`. This makes sense since more of our distributions are becoming stateful (`TransformedDistribution`, `Rejector`). See [detailed comment](https://github.com/uber/pyro/issues/732#issuecomment-362454322) in discussion thread below.
- [x] #768 Remove `log_pdf_mask` since this can be achieved via `with poutine.scale(None, log_pdf_mask)`.
- [x] #769 Simplify `Distribution` to be a mixin class that provides compatibility with PyTorch distributions. This should provide `.reparameterized`, `.enumerable`, `.__call__()`, and `.score_parts()`.
- [x] #769 Move `extra_event_dims` and `sample_shape` args to a `Reshape` distribution.
- [x] #769 Remove `TorchDistribution` wrapper class
- [x] #794 Programmatically wrap PyTorch distributions, as Edward does with Tensorflow distributions.
- [ ] Rename all distribution parameters to match PyTorch names.

### Pros

- less code for Pyro to maintain
- allows Pyro to use any new PyTorch distributions that users create
- avoids duplicated of effort in implementing new distributions
- removing `RandomPrimitive` would speed up Pyro (especially important now that PyTorch distributions cache intermediate results and are more expensive to construct)

### Cons

- refactoring requires effort
- breaks existing users
- makes Pyro more susceptible to API changes upstream

See also #638

neerajprad:
As I was working to incorporate transforms of distribution with constrained support in HMC, I wanted to start using

neerajprad:
I think deprecation of `RandomPrimitive` might become important for the following reason - currently, the Pyro distribution API is inconsistent with respect to `torch.distributions` with regard to certain attributes like `batch_shape`, `event_shape` etc. which are properties in `torch.distributions` but methods in `pyro.distributions`. The reason why we have them as methods in Pyro is to be able to support `RandomPrimitive` instances which cannot support properties (as we are required to pipe in args/kwargs). 

[...]

PR #769, opened 2018-02-11
Replace Distribution+TorchDistribution with a thin Distribution mixin

fritzo:
Addresses #732 

This refactors our PyTorch wrapping layer from a *has-a* to an *is-a* relationship, thereby allowing Pyro to use PyTorch-compatible distributions directly in Pyro sample statements.

## Why?

- This brings us much closer to being able to programmatically import all PyTorch distributions.
- This makes it easier to implement new PyTorch-compatible distributions for use in Pyro.

## How?

This PR extends Pyro's `Distribution` class hierarchy with a small mixin class
```
   Distribution
        |
TorchDistributionMixin    torch.distributions.Distribution
                 |           |
               TorchDistribution
```
The mixin allows PyTorch distributions to be subclassed and used directly in Pyro by:
- renaming `.reparameterized`, `.enumerable`, and param names, and
- providing default logic for `.__call__()` and `.score_parts()`
- providing a `.reshape()` property and `Reshape` wrapper class to finely control distribution shapes, e.g.
```diff
- pyro.sample("x", dist.Normal(mu, sigma, extra_event_dims=1), sample_shape=(2,3))
+ pyro.sample("x", dist.Normal(mu, sigma)
+                      .reshape(sample_shape=(2,3), extra_event_dims=1))
```

## Tasks

- [x] Implement a `TorchDistributionMixin` class
- [x] Add a `Reshape` distribution
- [x] Refactor `Normal`
- [x] Refactor `TransformedDistribution`
- [x] Refactor other distributions
- [x] Update all code to use `Reshape` instead of `extra_event_dims`
- [x] Update all code to use `Reshape` instead of `sample_shape`
- [x] Clean up docs


pyro/poutine/trace.py
```diff
@@ -82,10 +80,7 @@ def log_pdf(self, site_filter=lambda name, site: True):
                     site_log_p = site["log_pdf"]
                 except KeyError:
                     args, kwargs = site["args"], site["kwargs"]
-                    if isinstance(site["fn"], TorchDistribution):
-                        site_log_p = site["fn"].log_prob(site["value"]).sum() * site["scale"]
-                    else:
-                        site_log_p = site["fn"].log_prob(site["value"], *args, **kwargs).sum() * site["scale"]
+                    site_log_p = site["fn"].log_prob(site["value"], *args, **kwargs).sum() * site["scale"]
```
fritzo: @eb8680 Now that `sample_shape` is handled by `Reshape`, we no longer need this `isinstance` kludge.

eb8680:
I like it! Now if we can just automate the `TorchDistribution` wrapper generation then this will be really clean.

> Refactor Histogram to not inherit from `Distribution`

Do we want to allow stochastic functions that aren't `Distribution`s to expose a `log_prob` method?

fritzo:
> Do we want to allow stochastic functions that aren't Distributions to expose a log_prob method?

Yes. This PR reappropriates the name `Distribution` to denote a mixin class that provides Pyro functionality to PyTorch distributions. Due to Python's method resolution order, this pattern prevents us from having a common base class to declare the `.__call__()`+`.log_prob()` interface. In the lack of such a common base class (formerly named `Distribution`), I think the cleanest thing to do is allow duck-typed stochastic functions to have `.log_prob()` methods.

fritzo:
@eb8680 I was unable to eliminate the `Distribution` abstract base class from Pyro due to `isinstance()` logic in lift_poutine.py. Instead I've inserted the `TorchDistributionMixin` class in the existing hierarchy.

pyro/distributions/torch/__init__.py
```diff
@@ -0,0 +1,91 @@
+from __future__ import absolute_import, division, print_function
```
fritzo: This file is now much closer to being a simple loop over `dir(torch.distributions)`. It still remains to rename parameters and fix `Multinomial` parameter types.

fritzo:
Thanks for reviewing, @eb8680!

Issue #771, opened 2018-02-13
Support PyTorch scalars


fritzo:
PyTorch master now supports scalars. Pyro currently breaks with PyTorch master.

fritzo:
I think this works now. We don't use scalars in all our examples, but many test and examples already use scalars.

fritzo:
@fehiepsi Do you know whether your gp module supports scalars? I'm starting to use it and I noticed that your tutorial still uses 1-dimensional tensors like `variance = torch.Tensor([10])` rather than scalars like `variance = torch.tensor(10.0)`.

fehiepsi:
@fritzo I think that it should be fine. We use `length_scale` with `X / length_scale` and `variance` to multiply with the final computation of covariance matrix. At some place we do `variance.expand(...)` but the behavior of this method is the same for both scalar and non-scalar tensor. I will fix if you detect any error related to scalar tensors.

PR #772, opened 2018-02-14
RSA+CCG example

eb8680:
This is a quick port of [this webPPL/DIPPL example](http://dippl.org/examples/zSemanticPragmaticMashup.html), that I did today while playing around with some variants of search.  The direct translation of the example is done and it seems to behave mostly identically to the original, but it needs to be cleaned up and made to work with PyTorch `master` (#771 ).  It's also really slow...

cc @karalets

### DIFF
--- a/examples/semantic_parsing.py
+++ b/examples/semantic_parsing.py
@@ -358,7 +358,7 @@ def speaker(world):
 
 
 def rsa_listener(utterance, qud):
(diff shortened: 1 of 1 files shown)


ngoodman:
is this abandoned? or moved elsewhere?

eb8680:
> is this abandoned? or moved elsewhere?

Abandoned, at least for now

PR #777, opened 2018-02-15
add omt mvn distribution

martinjankowiak:
multivariate normal with omt gradients. @fritzo, you can assume the backwards math is correct (there's a test!), but i'm mostly looking for CR for distributions interface etc. also i'm always confused about ctx/save_for_backward....

### DIFF
--- a/pyro/distributions/omt_mvn.py
+++ b/pyro/distributions/omt_mvn.py
@@ -6,6 +6,7 @@ from torch.autograd.function import once_differentiable
 from torch.distributions import constraints
 
 from pyro.distributions.torch.multivariate_normal import MultivariateNormal
+from pyro.distributions.util import sum_leftmost
 
 
 class OMTMultivariateNormal(MultivariateNormal):
@@ -20,31 +21,16 @@ class OMTMultivariateNormal(MultivariateNormal):
     :param torch.autograd.Variable scale_tril: Cholesky of Covariance matrix.
     """
     params = {"loc": constraints.real, "scale_tril": constraints.lower_triangular}
-    support = constraints.real
-    has_rsample = True
-    reparameterized = True
 
     def __init__(self, loc, scale_tril):
-        assert(len(loc.size()) == 1), "OMTMultivariateNormal loc must be 1-dimensional"
-        assert(len(scale_tril.size()) == 2), "OMTMultivariateNormal scale_tril must be 2-dimensional"
+        assert(loc.dim() == 1), "OMTMultivariateNormal loc must be 1-dimensional"
+        assert(scale_tril.dim() == 2), "OMTMultivariateNormal scale_tril must be 2-dimensional"
         covariance_matrix = torch.mm(scale_tril, scale_tril.t())
         super(OMTMultivariateNormal, self).__init__(loc, covariance_matrix)
         self.scale_tril = scale_tril
 
     def rsample(self, sample_shape=torch.Size()):
         return _OMTMVNSample.apply(self.loc, self.scale_tril, sample_shape + self.loc.shape)
-
-
-def sum_all_but_rightmost_dim(x):
-    if len(x.size()) == 1:
-        return x
-    return x.view(-1, x.size(-1)).sum(0)
-
-
-def sum_all_but_rightmost_two_dims(x):
-    if len(x.size()) == 2:
-        return x
-    return x.view(-1, x.size(-2), x.size(-1)).sum(0)
 
 
 class _OMTMVNSample(Function):
@@ -65,18 +51,18 @@ class _OMTMVNSample(Function):
 
         dim = L.size(0)
         g = grad_output
-        loc_grad = sum_all_but_rightmost_dim(grad_output)
+        loc_grad = sum_leftmost(grad_output, -1)
 
         z_ja = z.unsqueeze(-1)
         g_transpose = torch.transpose(g, 0, -1)
         g_ij = g_transpose.contiguous().view(dim, -1)
         L_inv_bi_g_ji = torch.trtrs(g_ij, L, upper=False)[0].view(g_transpose.size())
         L_inv_bi_g_ji = torch.transpose(L_inv_bi_g_ji, 0, -1).unsqueeze(-2)
-        diff_L_ab = 0.5 * sum_all_but_rightmost_two_dims(L_inv_bi_g_ji * z_ja)
+        diff_L_ab = 0.5 * sum_leftmost(L_inv_bi_g_ji * z_ja, -2)
 
         epsilon_jb = epsilon.unsqueeze(-2)
         g_ja = g.unsqueeze(-1)
-        diff_L_ab += 0.5 * sum_all_but_rightmost_two_dims(g_ja * epsilon_jb)
+        diff_L_ab += 0.5 * sum_leftmost(g_ja * epsilon_jb, -2)
 
         identity = torch.eye(dim).type_as(g)
         R_inv = torch.trtrs(identity, L, transpose=True, upper=False)[0]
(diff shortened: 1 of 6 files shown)


pyro/distributions/omt_mvn.py
```diff
@@ -0,0 +1,112 @@
...
+    :param torch.autograd.Variable loc: Mean.
+    :param torch.autograd.Variable scale_tril: Cholesky of Covariance matrix.
+    """
+    params = {"loc": constraints.real, "scale_tril": constraints.lower_triangular}
+    support = constraints.real
+    has_rsample = True
+    reparameterized = True
+
+    def __init__(self, loc, scale_tril):
+        self.loc = loc
```
fritzo: It looks like you're inheriting from `MultivariateNormal` to get its `.log_prob` method. If this is the case, then you'll need to call its constructor
```py
def __init__(self, loc, scale_tril):
    covariance_matrix = torch.mm(scale_tril, scale_tril.t())
    super(OMTMultivariateNormal, self).__init__(loc, covariance_matrix)
    self.scale_tril = scale_tril
```

pyro/distributions/omt_mvn.py
```diff
@@ -0,0 +1,112 @@
...
+        self.scale_tril = scale_tril
+        assert(len(loc.size()) == 1), "OMTMultivariateNormal loc must be 1-dimensional"
+        assert(len(scale_tril.size()) == 2), "OMTMultivariateNormal scale_tril must be 2-dimensional"
+
+    @property
```
fritzo: Remove this; it is provided by base class.

pyro/distributions/omt_mvn.py
```diff
@@ -0,0 +1,112 @@
...
+    def variance(self):
+        return self.covariance_matrix.diag()
+
+    @lazy_property
```
fritzo: Remove this; the computation was already needed to create the base class.

martinjankowiak: thanks! done

pyro/distributions/omt_mvn.py
martinjankowiak: thanks! done

pyro/distributions/omt_mvn.py
martinjankowiak: thanks! done

pyro/distributions/omt_mvn.py
```diff
@@ -0,0 +1,104 @@
...
+    :param torch.autograd.Variable loc: Mean.
+    :param torch.autograd.Variable scale_tril: Cholesky of Covariance matrix.
+    """
+    params = {"loc": constraints.real, "scale_tril": constraints.lower_triangular}
+    support = constraints.real
+    has_rsample = True
+    reparameterized = True
+
+    def __init__(self, loc, scale_tril):
+        assert(len(loc.size()) == 1), "OMTMultivariateNormal loc must be 1-dimensional"
```
fritzo: Note you can use `x.dim()` instead of `len(x.size())`

martinjankowiak:
added test_log_prob. do i need to add anything to pyro `__init__`s?

pyro/distributions/omt_mvn.py
```diff
@@ -0,0 +1,104 @@
...
+        assert(len(scale_tril.size()) == 2), "OMTMultivariateNormal scale_tril must be 2-dimensional"
+        covariance_matrix = torch.mm(scale_tril, scale_tril.t())
+        super(OMTMultivariateNormal, self).__init__(loc, covariance_matrix)
+        self.scale_tril = scale_tril
+
+    def rsample(self, sample_shape=torch.Size()):
+        return _OMTMVNSample.apply(self.loc, self.scale_tril, sample_shape + self.loc.shape)
+
+
+def sum_all_but_rightmost_dim(x):
```
fritzo: Consider adding a generalized version to pyro/distributions/util.py
```py
def sum_leftmost(x, dim):
    if dim < 0:
        dim = x.dim() - dim
    if dim == 0:
        return x
    return x.contiguous().view(-1, *x.shape[dim:]).sum(0)

x = torch.ones(2,3,4)
assert sum_leftmost(x, -2).shape == (3, 4)
assert sum_leftmost(x, -1).shape == (4,)
```

pyro/distributions/omt_mvn.py
```diff
@@ -0,0 +1,104 @@
...
+    O(D^3), although the resulting gradient variance is generally expected to be lower.
+
+    A distribution over vectors in which all the elements have a joint Gaussian
+    density.
+
+    :param torch.autograd.Variable loc: Mean.
+    :param torch.autograd.Variable scale_tril: Cholesky of Covariance matrix.
+    """
+    params = {"loc": constraints.real, "scale_tril": constraints.lower_triangular}
+    support = constraints.real
```
fritzo: Omit the `.support`, `.reparameterized`, and `.has_rsample` properties; these are already correct.

fritzo:
> do i need to add anything to pyro __init__s?

Could you import this in pyro/distributions/__init__.py and add a section to docs/source/distributions.rst?

pyro/distributions/omt_mvn.py
fritzo: Actually this would be useful for me as well :smile:

martinjankowiak: done, thanks

pyro/distributions/omt_mvn.py
martinjankowiak: done, thanks

pyro/distributions/omt_mvn.py
martinjankowiak: done, thanks

pyro/distributions/util.py
```diff
@@ -98,6 +98,31 @@ def sum_rightmost(value, dim):
...
+    ```
+    x = torch.ones(2,3,4)
+    assert sum_leftmost(x, -2).shape == (3, 4)
+    assert sum_leftmost(x, -1).shape == (4,)
+    ```
+
+    :param torch.autograd.Variable x: A tensor
+    :param int dim: Specifies the number of dims to sum out (a non-positive integer)
+    """
+    assert(dim <= 0), "dim must be less than or equal to zero"
```
fritzo: Could you please implement the general interface I suggested.
```py
def sum_leftmost(x, dim):
    if dim < 0:
        dim += x.dim()
    ...
```
and add a test
```py
def test_sum_leftmost():
    x = torch.ones(2,3,4)
    assert sum_leftmost(x, 0).shape == (2, 3, 4)
    assert sum_leftmost(x, 1).shape == (3, 4)
    assert sum_leftmost(x, 2).shape == (4,)
    assert sum_leftmost(x, -1).shape == (4,)
    assert sum_leftmost(x, -2).shape == (3, 4)
```

martinjankowiak: i didn't understand your interface

martinjankowiak: since ``dim += x.dim()`` was different. i guess now i see it

fritzo: Sorry for being unclear. The interface should match `sum_rightmost()`: for positive integers `n`, `sum_leftmost(x, n)` sums out the `n` leftmost dimensions. For negative integers `-n`, `sum_leftmost(x, -n)` sums out all but the `n` rightmost dimensions, i.e. the leftmost `x.dim() - n` dimensions. This matches the behavior of negative indexing for many other operations.

fritzo: Thanks, I'm only requesting this because we'll be needing it in our work on `iarange` and parallel enumeration, so your efforts in this PR will pay off.

martinjankowiak:
either way, tests are pretty thorough, would be hard to fool

PR #781, opened 2018-02-16
Make convenient functions `is_nan`, `is_inf`

fehiepsi:
Besides the above convenient functions, I also clean HMC a bit.

### DIFF
--- a/pyro/util.py
+++ b/pyro/util.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import functools
+import numbers
 import warnings
 
 import graphviz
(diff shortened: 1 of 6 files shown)


pyro/util.py
```diff
@@ -181,6 +182,26 @@ def ng_zeros(*args, **kwargs):
...
 
 
+def is_nan(x):
+    """
+    A convenient function to check if a Tensor contains all nan; also works with numbers
+    and torch.autograd.Variable
+    """
+    if isinstance(x, numbers.Number):
+        return x != x
+    return all(x != x)
```
fritzo: I believe this should use PyTorch via
```py
return (x !=  x).all()
```

pyro/util.py
```diff
@@ -181,6 +182,26 @@ def ng_zeros(*args, **kwargs):
...
+
+
+def is_inf(x):
+    """
+    A convenient function to check if a Tensor contains all inf; also works with numbers
+    and torch.autograd.Variable
+    """
+    if isinstance(x, numbers.Number):
+        return x == float('inf')
+    return all(x == float('inf'))
```
fritzo: ditto:
```py
return (x == float('inf')).all()
```

pyro/util.py
fehiepsi: @fritzo It seems that they are the same in my system. Anyway, I made the changes based on my your suggestions.

fritzo:
```py
>>> all(torch.ones(2))
True
>>> all(torch.ones(2, 2))
  File "/Users/fritzobermeyer/pytorch-dev/2.7/torch/tensor.py", line 168, in __bool__
    " containing more than one value is ambiguous")
RuntimeError: bool value of torch.FloatTensor containing more than one value is ambiguous
```

PR #784, opened 2018-02-17
Add instructions on installing dev branch

fritzo:
This adds instructions for installing Pyro dev branch with a recent (but not master) PyTorch release. This is needed because Pyro dev branch currently requires a PyTorch version that is neither a release nor PyTorch master. These docs will need to be updated as part of #771 when we fix Pyro to use PyTorch master.

See #547 for supporting multiple documentation version per release

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Pyro is a flexible, scalable deep probabilistic programming library built on PyT
 Pyro is in an alpha release.  It is developed and used by [Uber AI Labs](http://uber.ai).
 For more information, check out our [blog post](http://eng.uber.com/pyro).
 
-## Installation
+## Installing Pyro stable release
 
 First install [PyTorch](http://pytorch.org/).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ Pyro is a flexible, scalable deep probabilistic programming library built on PyT
 Pyro is in an alpha release.  It is developed and used by [Uber AI Labs](http://uber.ai).
 For more information, check out our [blog post](http://eng.uber.com/pyro).
 
-## Installation
+## Installing Pyro stable release
```
jpchen: nit: i think it would be cleaner to have
```
## Installation
### Installing stable ...
### Installing from ...
``` 
so you can get the hyperlink to a specific section or the entire installation

README.md
```diff
@@ -43,13 +43,29 @@ pip3 install pyro-ppl
...
 
-Note that to use Pyro's `dev` branch, you will need to build PyTorch's master branch. 
-Refer to the [README](https://github.com/pytorch/pytorch/blob/master/README.md) for more details
-on how to build PyTorch from source. Otherwise, to use the latest version of Pyro that 
-is compatible with PyTorch release, use the `master` branch of Pyro.
+## Installing Pyro dev branch
+
+For recent features you can install Pyro from source.
+
+First install a recent PyTorch, currently PyTorch commit `ca5071d07`.
```
jpchen: the most recent commit will change, do you want to just put a placeholder `[commit hash]`?

neerajprad: I think it would be better if we specify the exact commit that was used to build the PyTorch CPU wheels used in travis, so as to have a certain level of assurance that the integration isn't broken. Every time we update the wheels, we can update this README.

README.md
fritzo: done

README.md
fritzo: > the exact commit that was used to build the PyTorch CPU wheels

Yes that is my intention. The commit in this PR was the latest wheel commit.

PR #785, opened 2018-02-18
Correct diagnostic behavior of hmc

fehiepsi:
In the current implementation, `mcmc._t` and `hmc._t` are different each time we call `.diagnostic`, so the logging info is not quite true. This pull request removes the property `self._t` of `MCMC` (if `_accept_cnt` is a property of HMC kernel, then `_t` should also be). As a consequence, we can remove the `time_step` argument of `.diagnostic`.

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -87,7 +87,7 @@ class HMC(TraceKernel):
 
     def sample(self, trace):
         z = {name: node['value'] for name, node in trace.iter_stochastic_nodes()}
-        r = {name: pyro.sample('r_{}_t={}'.format(name, self._t), self._r_dist[name]) for name in z}
+        r = {name: pyro.sample('r_{}_t={}'.format(name, self._t), self._r_dist[name]) for name in sorted(z)}
         z_new, r_new = velocity_verlet(z, r, self._potential_energy, self.step_size, self.num_steps)
         # apply Metropolis correction
         energy_proposal = self._energy(z_new, r_new)
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -12,7 +12,7 @@ from pyro.util import ng_ones, ng_zeros, is_nan, is_inf
 
 class HMC(TraceKernel):
     """
-    Simple Hamiltonian Monte Carlo kernel, where `step_size` and `num_steps`
+    Simple Hamiltonian Monte Carlo kernel, where ``step_size`` and ``num_steps``
     need to be explicitly specified by the user.
 
     **Reference**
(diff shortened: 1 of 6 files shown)


fehiepsi:
It seems that the tests failed due to #786.

pyro/infer/mcmc/hmc.py
```diff
@@ -40,8 +40,7 @@ def _get_trace(self, z):
         for name, value in z.items():
             z_trace.nodes[name]['value'] = value
         trace_poutine = poutine.trace(poutine.replay(self.model, trace=z_trace))
-        trace_poutine(*self._args, **self._kwargs)
-        return trace_poutine.trace
+        return trace_poutine.get_trace(*self._args, **self._kwargs)
```
neerajprad: The reason for avoiding the `.get_trace` method is to avoid a trace copy which turns out to be quite expensive (given the number of calls involved) and unnecessary for HMC. Refer to https://github.com/uber/pyro/pull/690#discussion_r163117016 for a discussion on this (you may have to uncollapse the comments).

pyro/poutine/trace_poutine.py
```diff
@@ -129,13 +129,13 @@ def __exit__(self, *args, **kwargs):
                 get_vectorized_map_data_info(self.trace)
         return super(TraceMessenger, self).__exit__(*args, **kwargs)
 
-    def get_trace(self, *args, **kwargs):
```
neerajprad: This does not seem to be the latest `dev`; could you merge and update the PR?

```
    def get_trace(self, *args, **kwargs):
        """
        :returns: data structure
        :rtype: pyro.poutine.Trace

        Helper method for a very common use case.
        Calls this poutine and returns its trace instead of the function's return value.
        """
        self(*args, **kwargs)
        return self.msngr.get_trace()
```

pyro/infer/mcmc/hmc.py
fehiepsi: Yeah, I can confirm this! Will change back to the old behavior.

pyro/poutine/trace_poutine.py
fehiepsi: @neerajprad This change happens at the messenger class, not the poutine class. I use the dev branch.

fehiepsi:
Just a notice, the bug at #786 is solved in this pull request.

pyro/infer/mcmc/hmc.py
```diff
@@ -88,7 +87,7 @@ def cleanup(self):
 
     def sample(self, trace):
         z = {name: node['value'] for name, node in trace.iter_stochastic_nodes()}
-        r = {name: pyro.sample('r_{}_t={}'.format(name, self._t), self._r_dist[name]) for name in z}
+        r = {name: pyro.sample('r_{}_t={}'.format(name, self._t), self._r_dist[name]) for name in sorted(z)}
```
neerajprad: Thanks for fixing this hard to diagnose source of randomness! :)

Issue #786, opened 2018-02-18
Test for HMC fail randomly


fehiepsi:
In the current dev branch, when I tried to run `pytest tests/infer/mcmc/test_hmc.py::test_hmc_conjugate_gaussian` with `TEST_CASES  = TEST_CASES[:1]` (just test the first case among four cases) repeatedly about 10 times, the test failed. So I guess that: despite we already had `@pytest.mark.init(rng_seed=34)`, somehow the results still cannot be reproduced.

neerajprad:
What is the pyro/pytorch versions you are on? Could you also paste the logs?

fehiepsi:
@neerajprad I believe the bug is fixed in #785 . The reason for randomness is: we sample `r`s based on keys of a dict.

neerajprad:
Thanks for fixing this in #785. For the records, the change was to iterate over the keys of the dictionary in order when sampling the momenta -
 https://github.com/uber/pyro/pull/785#discussion_r169121588.

PR #787, opened 2018-02-18
Implement No-U-turn sampler

fehiepsi:
This pull request adds NUTS kernel for mcmc. The implement is quite similar to Algorithm 3 of the [original paper](http://jmlr.org/papers/volume15/hoffman14a/hoffman14a.pdf).

- [x] Implement NUTS
- [x] Add test file

### DIFF
--- a/pyro/infer/mcmc/nuts.py
+++ b/pyro/infer/mcmc/nuts.py
@@ -129,7 +129,7 @@ class NUTS(HMC):
             if int(is_other_half_tree.data[0]) == 1:
                 z_proposal = other_half_tree.z_proposal
 
-        # leaves of the full tree is determined by the direction
+        # leaves of the full tree are determined by the direction
         if direction == 1:
             z_left = half_tree.z_left
             r_left = half_tree.r_left
--- a/pyro/infer/mcmc/nuts.py
+++ b/pyro/infer/mcmc/nuts.py
@@ -2,8 +2,11 @@ from __future__ import absolute_import, division, print_function
 
 from collections import namedtuple
 
+import torch
+
 import pyro
 import pyro.distributions as dist
+from pyro.infer.util import torch_data_sum
 from pyro.ops.integrator import velocity_verlet
 from pyro.util import ng_ones, ng_zeros
 
@@ -35,8 +38,13 @@ class NUTS(HMC):
         self._dE_max = 1000
 
     def _is_turning(self, z_left, r_left, z_right, r_right):
+        nodes = sorted(z_left)
+        z_left = torch.stack([z_left[name] for name in nodes])
+        r_left = torch.stack([r_left[name] for name in nodes])
+        z_right = torch.stack([z_right[name] for name in nodes])
+        r_right = torch.stack([r_right[name] for name in nodes])
         dz = z_right - z_left
-        return (dz.dot(r_left) < 0).all() or (dz.dot(r_right) < 0).all()
+        return (torch_data_sum(dz * r_left) < 0) or (torch_data_sum(dz * r_right) < 0)
 
     def _build_basetree(self, z, r, log_slice):
         z_new, r_new = velocity_verlet(z, r, self._potential_energy, self.step_size)
@@ -64,8 +72,8 @@ class NUTS(HMC):
         other_half_tree = self._build_tree(z, r, log_slice, direction, tree_depth-1)
 
         tree_size = half_tree.size + other_half_tree.size
-        prob_other = other_half_tree.size / tree_size
-        rand = int(dist.Bernoulli(probs=ng_ones(1) * prob_other)().data[0])
+        prob_other = other_half_tree.size / tree_size if tree_size != 0 else 1
+        rand = int(dist.Bernoulli(ps=ng_ones(1) * prob_other)().data[0])
         z_proposal = other_half_tree.z_proposal if rand == 1 else half_tree.z_proposal
 
         if direction == 1:
@@ -86,8 +94,8 @@ class NUTS(HMC):
 
     def sample(self, trace):
         z = {name: node['value'] for name, node in trace.iter_stochastic_nodes()}
-        r = {name: pyro.sample('r_{}_t={}'.format(name, self._t), self._r_dist[name]) for name in z}
-        slice_var = dist.Uniform(ng_zeros(1), self._energy(z, r))()
+        r = {name: pyro.sample('r_{}_t={}'.format(name, self._t), self._r_dist[name]) for name in sorted(z)}
+        slice_var = dist.Uniform(ng_zeros(1), torch.exp(-self._energy(z, r)))()
         log_slice = slice_var.log()
         z_left = z_right = z
         r_left = r_right = r
@@ -98,7 +106,7 @@ class NUTS(HMC):
 
         while not (turning or diverging):
             direction = pyro.sample("direction_t={}_depth={}".format(self._t, tree_depth),
-                                    dist.Bernoulli(probs=ng_ones(1) * 0.5))
+                                    dist.Bernoulli(ps=ng_ones(1) * 0.5))
             direction = int(direction.data[0])
             if direction == 1:
                 tree = self._build_tree(z_left, r_left, log_slice, direction, tree_depth)
--- a/pyro/infer/mcmc/nuts.py
+++ b/pyro/infer/mcmc/nuts.py
@@ -123,7 +123,7 @@ class NUTS(HMC):
         #     we choose the proposal from the first half
         #     (any is fine, because the probability of picking it at the end is 0!).
         if tree_size != 0:
-            other_half_tree_prob = other_half_tree.size / tree_size if tree_size != 0 else 0.5
+            other_half_tree_prob = other_half_tree.size / tree_size
             is_other_half_tree = pyro.sample("is_other_halftree",
                                              dist.Bernoulli(ps=ng_ones(1) * other_half_tree_prob))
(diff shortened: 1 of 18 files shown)


neerajprad:
This has been on our todo list, so cheers for starting work on this! I will be taking a look at this soon.

eb8680:
Thanks @fehiepsi this is great, will review when you add tests.  Did you consider implementing the dual averaging version (algorithm 6 in the paper)?

fehiepsi:
@eb8680 For dual averaging, I think that it applies to HMC too, so we can do it in another pull request. I intend to add kernels for Gaussian Process this week, so will consider to do it in a near future (after making GP more versatile).

fehiepsi:
@neerajprad @eb8680 I have used tests of HMC with a little modification for NUTS. This pull request is ready to review now. :)

pyro/infer/mcmc/nuts.py
```diff
@@ -0,0 +1,135 @@
...
+    :param model: python callable containing pyro primitives.
+    :param float step_size: determines the size of a single step taken by the
+        verlet integrator while computing the trajectory using Hamiltonian
+        dynamics.
+    """
+
+    def __init__(self, model, step_size=0.5):
+        super(NUTS, self).__init__(model, step_size, num_steps=None)
+        self._reset()
+        self._dE_max = 1000
```
eb8680: Add a comment explaining this attribute

pyro/infer/mcmc/nuts.py
```diff
@@ -0,0 +1,135 @@
...
+from .hmc import HMC
+
+
+_TreeInfo = namedtuple("TreeInfo", ["z_left", "r_left", "z_right", "r_right",
+                                    "z_proposal", "size", "turning", "diverging"])
+
+
+class NUTS(HMC):
+    """
+    No-U-Turn Sampler kernel, where ``step_size`` need to be explicitly specified by the user.
```
eb8680: Can you add a little more information here, and maybe a small instantiation example?  I know the HMC documentation is a bit slim as well but since I suspect this will get a fair amount of use once it's mature it should be made as accessible as possible.

pyro/infer/mcmc/nuts.py
```diff
@@ -0,0 +1,135 @@
...
+        return (torch_data_sum(dz * r_left) < 0) or (torch_data_sum(dz * r_right) < 0)
+
+    def _build_basetree(self, z, r, log_slice):
+        z_new, r_new = velocity_verlet(z, r, self._potential_energy, self.step_size)
+        dE = (log_slice + self._energy(z_new, r_new)).data[0]
+        tree_size = 1 if dE <= 0 else 0
+        diverging = dE >= self._dE_max
+        return _TreeInfo(z_new, r_new, z_new, r_new, z_new, tree_size, False, diverging)
+
+    def _build_tree(self, z, r, log_slice, direction, tree_depth):
```
eb8680: Since this is the complicated bit, can you add a few more comments explaining what's happening?

pyro/infer/mcmc/nuts.py
fehiepsi: Yes, it is important so I will explain it in the implementation.

pyro/infer/mcmc/nuts.py
fehiepsi: Yes, I'll do.

fehiepsi:
@eb8680 I have to admit that it is easy to have bugs without comments. Now, I have added much more comments based on my understanding of the paper. I just started to study NUTS from the last week so please correct me if I got something wrong. For the example, I took from the test of HMC.

During the process of adding comments, I discover that I missed a crucial point: change the sign of `step_size` if we go to the left. After I fixed that bug, things seem good: we get better performance. As you can see from the test file, `mean_tol` and `std_tol` for both test case 1 and 2 are smaller than in HMC.  The tests 3 and 4 also passed for NUTS while they are failed in HMC. In addition, the acceptance rates of NUTS are much higher than HMC case (in my personal view, it suggests that we can use a smaller `num_samples` but still get good performance). For the speed, I don't think that it is a fair comparison because in HMC there are a fixed num_steps while in NUTS it is a free parameter (FYI, the first test took 12.27s for NUTS while it took 3.28s for HMC).

In addition, I have included a parameter `max_tree_depth` as in Stan package. This replaces the `while` loops (keep doubling until some conditions fail) by a `for` loop (when the tree is so big, by default it is 2**10 = 1024, we stop).

neerajprad:
> I don't have any major comments about the code. One of the reasons it took us a while to add HMC was profiling to make sure Pyro wasn't adding unacceptable overhead. 

Since the runtime was primarily dominated by log prob and gradient computation, the additional NUTS book-keeping shouldn't add too much of an overhead. As @fehiepsi noted, doing a comparison on the test examples is not straightforward due to the adaptive nature of the `num_steps` parameter, but I think we can do some rough checks to ensure that the overhead isn't too much. I will try to run, and validate that. Also, due to the better effective sample size, we might not need as many samples to get a good posterior estimate with NUTS. 

@fehiepsi - I was on vacation, so sorry about the delay in reviewing. Overall, this is looking great! Going through this in more detail now.

pyro/infer/mcmc/nuts.py
```diff
@@ -0,0 +1,211 @@
...
+        z_left = torch.stack([z_left[name] for name in nodes])
+        r_left = torch.stack([r_left[name] for name in nodes])
+        z_right = torch.stack([z_right[name] for name in nodes])
+        r_right = torch.stack([r_right[name] for name in nodes])
+        dz = z_right - z_left
+        return (torch_data_sum(dz * r_left) < 0) or (torch_data_sum(dz * r_right) < 0)
+
+    def _build_basetree(self, z, r, log_slice, direction):
+        step_size = self.step_size if direction == 1 else -self.step_size
+        z_new, r_new = velocity_verlet(z, r, self._potential_energy, step_size)
```
neerajprad: It does seem that for the same number of steps, we will be doing double the number of gradient (and model log prob eval) computations in `velocity_verlet` for NUTS. The for loop in the integrator reuses the computation in the previous step which we cannot do with `num_steps=1` and end up repeating the computation. Correct me if I am mistaken. I don't immediately see a clean way of caching this at the moment, but will think about it so that we can avoid this double cost.

neerajprad:
This is a very clean and well-documented implementation of NUTS. Love it! :) 

One way to reuse the gradient computation could be to define a new integrator interface as follows for NUTS to allow for reusing gradients in a recursive setting, augment `_TreeInfo` with `z_left_grad` and `z_right_grad`, and pipe these in in `_build_tree`. 


```
def single_step_velocity_verlet(z, r, potential_fn, step_size, z_grads=None):
    z_next = {key: val.data.clone() for key, val in z.items()}
    r_next = {key: val.data.clone() for key, val in r.items()}
    grads = _grad(potential_fn, z_next) if z_grads is None else z_grads

    for site_name in z_next:
        # r(n+1/2)
        r_next[site_name] = r_next[site_name] + 0.5 * step_size * (-grads[site_name])
        # z(n+1)
        z_next[site_name] = z_next[site_name] + step_size * r_next[site_name]
    grads = _grad(potential_fn, z_next)
    for site_name in r_next:
        # r(n+1)
        r_next[site_name] = r_next[site_name] + 0.5 * step_size * (-grads[site_name])
    z_next = {key: Variable(val) for key, val in z_next.items()}
    r_next = {key: Variable(val) for key, val in r_next.items()}
    return z_next, r_next, grads
```

We could also reuse the potential energy term instead of calling `self._energy` for each of the steps (the number of call could be `num_steps` those in the naive implementation). This will ensure that the NUTS overhead is very minimal. In its current form, it could add almost double the time as compared to the naive HMC implementation for the same number of steps.

I really like the simplicity and readability of the current approach though, so we can investigate if this is worth it in a later PR, and how best to do it.

pyro/infer/mcmc/nuts.py
fehiepsi: Whoa, a crucial point! Thanks for pointing it.

fehiepsi:
@neerajprad You have pointed out a very important point for performance. Thanks a lot for that! I will try to address it in the next commit based on your idea.

neerajprad:
> @neerajprad You have pointed out a very important point for performance. Thanks a lot for that! I will try to address it in the next commit based on your idea.

We can try addressing any perf issues in a separate PR, and discuss the pros/cons of different approaches. This PR is looking great as is!

fehiepsi:
@neerajprad Yes, we can do it and discuss it in another pull request. I just addressed your comment about sorting dicts by making `self._r_dist` be an OrderDict. Now, we don't have sorted keys anymore.

**Edit:** In Trace, we use `node_dict_factory = OrderedDict`, so no need of the last commit. I will remove it.

fehiepsi:
Oopps, the randomness problem at #786 happened again. I am wondering `.iter_stochastic_nodes()` yield items from an OrderDict?

fehiepsi:
Yeah, it is strange. Despite that we use `node_dict_factory = collections.OrderedDict` for trace, the `iter_stochastic_nodes()` methods yield results randomly (maybe what I understand about `node_dict_factory` is wrong). After I sort it, I get consistent results. It should be fine now.

neerajprad:
> Despite that we use node_dict_factory = collections.OrderedDict for trace, the iter_stochastic_nodes() methods yield results randomly (maybe what I understand about node_dict_factory is wrong).

I think sorting this is needed, as the order from trace poutine is not by key's alphabetic order, but is probably related to the order of encountering the nodes in the execution.

PR #794, opened 2018-02-21
Automatically import all standard torch distributions

fritzo:
This adds a `for ... in torch.distributions.__dict_...` loop that automatcially wraps all available PyTorch distributions. Only `TransformedDistribution` is moved into this loop since most distributions require custom param renaming. However this PR does give Pyro access to all other PyTorch distributions e.g. `Geometric` and `Pareto`.

## Tested

- existing tests
- checked that documentation is correctly generated

### DIFF
--- a/pyro/distributions/__init__.py
+++ b/pyro/distributions/__init__.py
@@ -2,14 +2,12 @@ from __future__ import absolute_import, division, print_function
 
 from pyro.distributions.delta import Delta
 from pyro.distributions.distribution import Distribution
+from pyro.distributions.omt_mvn import OMTMultivariateNormal
 from pyro.distributions.rejector import Rejector
-from pyro.distributions.torch import (Bernoulli, Beta, Binomial, Categorical, Cauchy, Dirichlet, Exponential, Gamma,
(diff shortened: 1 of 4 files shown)


pyro/distributions/__init__.py
```diff
@@ -2,14 +2,12 @@
 
 from pyro.distributions.delta import Delta
 from pyro.distributions.distribution import Distribution
+from pyro.distributions.omt_mvn import OMTMultivariateNormal
 from pyro.distributions.rejector import Rejector
-from pyro.distributions.torch import (Bernoulli, Beta, Binomial, Categorical, Cauchy, Dirichlet, Exponential, Gamma,
-                                      LogNormal, Multinomial, Normal, OneHotCategorical, Poisson,
-                                      TransformedDistribution, Uniform)
+from pyro.distributions.torch import *
```
jpchen: didnt you just berate me a few days ago to never `import *`!

fritzo: This is the exception that makes the rule :wink: In pyro.distributions.torch we are dynamically populating the module, therefore explicit import is impossible. Instead we do the next best: we explicitly construct the `__all__` list.

PR #795, opened 2018-02-21
Reuse grad calculation for NUTS

fehiepsi:
As @neerajprad pointed out [here](https://github.com/uber/pyro/pull/787#issuecomment-367146820), we can reuse gradient calculations to increase the speed of NUTS. In my opinion, @neerajprad 's solution is optimal (assume we want to reuse grads, after each doubling step, because we don't know which direction we will go, it is necessary to save grads of both left and right leaves of the current tree), so I follow that solution. FYI the speed of the last implementation for the first two tests is `43s`  while it is `32s` for the current one. I have also checked the inference results: both versions give identical solutions.

Addresses #799.

### DIFF
--- a/pyro/ops/integrator.py
+++ b/pyro/ops/integrator.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import, division, print_function
 from torch.autograd import Variable, grad
 
 
-def velocity_verlet(z, r, potential_fn, step_size, num_steps=1, z_grads=None):
+def velocity_verlet(z, r, potential_fn, step_size, num_steps=1):
     """
     Second order symplectic integrator that uses the velocity verlet algorithm.
 
@@ -22,7 +22,7 @@ def velocity_verlet(z, r, potential_fn, step_size, num_steps=1, z_grads=None):
     """
     z_next = {key: val.data.clone() for key, val in z.items()}
     r_next = {key: val.data.clone() for key, val in r.items()}
-    grads = _grad(potential_fn, z_next) if z_grads is None else z_grads
+    grads = _grad(potential_fn, z_next)
 
     for _ in range(num_steps):
         for site_name in z_next:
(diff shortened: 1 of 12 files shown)


fehiepsi:
Another solution is to introduce `self._grads` and `self._grads_tmp`, where `self._grads` take the job of storing grads travelling along the current direction and `self._grads_tmp` just stores the static site leaf.

It might save some spaces of code but a little bit hacky.

pyro/ops/integrator.py
```diff
@@ -35,7 +36,7 @@ def velocity_verlet(z, r, potential_fn, step_size, num_steps=1):
             r_next[site_name] = r_next[site_name] + 0.5 * step_size * (-grads[site_name])
     z_next = {key: Variable(val) for key, val in z_next.items()}
     r_next = {key: Variable(val) for key, val in r_next.items()}
-    return z_next, r_next
+    return z_next, r_next, grads
```
neerajprad: nit: Rather than changing the interface of this integrator, I would suggest creating another method called `single_step_velocity_verlet` and call that from here. This ensures that we can use either of the methods without having to change the return type for the default method. 

For instance, we might want to return the computed potential energy value for `z_next` too in `single_step_velocity_verlet` so as to not have to recompute this using `self._energy`, and it will be cumbersome to change the default method's return type again.

```
    z_next = {key: val.data.clone() for key, val in z.items()}
    r_next = {key: val.data.clone() for key, val in r.items()}
    grads = _grad(potential_fn, z_next)
    for _ in range(num_steps):
        z_next, r_next, grads = single_step_velocity_verlet(z_next, r_next, grads)
    z_next = {key: Variable(val) for key, val in z_next.items()}
    r_next = {key: Variable(val) for key, val in r_next.items()}
    return z_next, r_next
```

neerajprad:
> FYI the speed of the last implementation for the first two tests is 43s while it is 32s for the current one. I have also checked the inference results: both versions give identical solutions.

I think we can get a further benefit by reusing the potential energy term computed in the integrator instead of having to recompute that in `self._energy`, but it is only worth investigating that if it results in a signficant decrease in runtime.

fehiepsi:
I record time spending for the first two tests:

Not reuse grad and potential, before @fritzo 's latest commit to fix memory leak: 62s
Not reuse grad and potential, after @fritzo 's latest commit to fix memory leak: 55s
Reuse grad, after latest commit: 42s
Reuse grad + potential, after latest commit: 29s.

I must say: whoa, huge change! @@

Thanks a lot! @neerajprad @fritzo

PR #797, opened 2018-02-22
Resolve code breakage due to PyTorch scalar introduction

neerajprad:
Addresses #771 

Some minor changes to bring Pyro `dev` in sync with PyTorch master. Tests will fail unless we update the CI wheels, though I have tested locally that all tests pass.

@fritzo - let me know if this looks reasonable. I'll update the CI wheels and re-run the tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ pip install .
 
 For recent features you can install Pyro from source.
 
-First install a recent PyTorch, currently PyTorch commit `ca5071d07`.
+First install a recent PyTorch, currently PyTorch commit `853dba8`.
 ```sh
 git clone git@github.com:pytorch/pytorch
 cd pytorch
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ First install a recent PyTorch, currently PyTorch commit `853dba8`.
 ```sh
 git clone git@github.com:pytorch/pytorch
 cd pytorch
-git checkout ca5071d07
+git checkout [commit hash above]
 ```
 Then build PyTorch following instructions in the PyTorch
 [README](https://github.com/pytorch/pytorch/blob/master/README.md).
--- a/pyro/infer/abstract_infer.py
+++ b/pyro/infer/abstract_infer.py
@@ -56,14 +56,16 @@ class Histogram(dist.Distribution):
                 logits.append(logit)
             else:
                 # Value has already been seen.
-                logits[ix] = util.log_sum_exp(torch.stack([logits[ix], logit]).squeeze())
+                logits[ix] = util.log_sum_exp(torch.stack([logits[ix], logit]))
 
         logits = torch.stack(logits).squeeze()
         logits -= util.log_sum_exp(logits)
         if not isinstance(logits, torch.autograd.Variable):
             logits = Variable(logits)
         logits = logits - util.log_sum_exp(logits)
-
+        # Categorical distribution must have at least one event dimension
+        if logits.dim() == 0:
+            logits = logits.unsqueeze(-1)
         d = dist.Categorical(logits=logits)
         return d, values
 
--- a/pyro/infer/abstract_infer.py
+++ b/pyro/infer/abstract_infer.py
@@ -58,14 +58,11 @@ class Histogram(dist.Distribution):
                 # Value has already been seen.
                 logits[ix] = util.log_sum_exp(torch.stack([logits[ix], logit]))
 
-        logits = torch.stack(logits).squeeze()
+        logits = torch.stack(logits).contiguous().view(-1)
         logits -= util.log_sum_exp(logits)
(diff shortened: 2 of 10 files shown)


tutorial/source/bayesian_regression.ipynb
```diff
@@ -123,7 +123,7 @@
     "    y_data = data[:, -1]\n",
     "    for j in range(num_iterations):\n",
     "        # run the model forward on the data\n",
-    "        y_pred = regression_model(x_data)\n",
+    "        y_pred = regression_model(x_data).squeeze(-1)\n",
```
neerajprad: @jpchen - `y_pred` returns a tensor of size `100x1` on PyTorch master now, whereas `y_data` has size `100` and hence it errors when we compute the loss. Also, removed the other `squeeze` operations without which things seem to run fine.

pyro/infer/tracegraph_elbo.py
```diff
@@ -155,7 +155,8 @@ def _compute_elbo_non_reparam(guide_trace, guide_vec_md_nodes,  #
         if log_pdf_key == 'log_pdf':
             score_function_term = score_function_term.sum()
         if use_nn_baseline or use_decaying_avg_baseline or use_baseline_value:
-            if downstream_cost.size() != baseline.size():
+            if (downstream_cost.dim() == 0 and baseline.dim() > 1) or \
```
neerajprad: @fritzo, @martinjankowiak - This works fine now, but please feel free to modify this logic later. The issue is that `downstream_cost` can be a dim 0 tensor now whereas the baseline will always have dim > 0 as the output from a nn module can't be dim 0. I tried a couple of other approaches, but this seemed like the cleanest way to get all the tests working.

pyro/infer/abstract_infer.py
```diff
@@ -56,14 +56,16 @@ def _dist_and_values(self, *args, **kwargs):
...
-                logits[ix] = util.log_sum_exp(torch.stack([logits[ix], logit]).squeeze())
+                logits[ix] = util.log_sum_exp(torch.stack([logits[ix], logit]))
 
         logits = torch.stack(logits).squeeze()
         logits -= util.log_sum_exp(logits)
         if not isinstance(logits, torch.autograd.Variable):
             logits = Variable(logits)
         logits = logits - util.log_sum_exp(logits)
-
+        # Categorical distribution must have at least one event dimension
```
fritzo: Would it be cleaner to replace the `.squeeze()` above with a `.view()`?
```diff
- logits = torch.stack(logits).squeeze()
+ logits = torch.stack(logits).contiguous().view(-1)
```

tutorial/source/bayesian_regression.ipynb
jpchen: this looks fine, can you please change this in the example as well? and test with python3 also if you havent, since there was a squeeze bug in pytorch when doing `backward()`.

pyro/util.py
```diff
@@ -202,16 +202,9 @@ def is_inf(x):
...
-        vecs = vecs.view(1, -1)
-    _, idx = torch.max(vecs, 1)
-    max_score = torch.index_select(vecs, 1, idx.view(-1))
-    ret = max_score + torch.log(torch.sum(torch.exp(vecs - max_score.expand_as(vecs))))
-    if n == 1:
-        return ret.view(-1)
-    return ret
+def log_sum_exp(tensor):
+    max_val = tensor.max(dim=-1)[0]
+    return max_val + (tensor - max_val).exp().sum(dim=-1).log()
```
jpchen: +1. this kind of code simplification is great.

fritzo: Don't you need to `.unsqueeze(-1)` the `max_val` to broadcast correctly?
```py
return max_val + (tensor - max_val.unsqueeze(-1)).exp().sum(dim=-1).log()
```

neerajprad: Completely my bad, will change.

pyro/infer/abstract_infer.py
neerajprad: Done

neerajprad:
Wheels are updated, so this should be good to merge if all tests pass.

neerajprad:
> Oh I forgot, could you update the version hash in the README.md?

I forgot about that too! Will do.

README.md
```diff
@@ -53,7 +53,7 @@ pip install .
 
 For recent features you can install Pyro from source.
 
-First install a recent PyTorch, currently PyTorch commit `ca5071d07`.
+First install a recent PyTorch, currently PyTorch commit `853dba8`.
```
fritzo: Sorry, you also need to replace the commit below in `git checkout`. Feel free to reword so the commit only appears once.

neerajprad:
This should be good to merge now.

Issue #799, opened 2018-02-22
Resolve perf issues with NUTS


neerajprad:
NUTS is currently slower than base HMC, mostly because of some repeated calls to compute the potential energy term and its gradient, that arise out of the recursive nature of the algorithm. 
 - [x] The base HMC reuses gradient computation from the previous step in each call to the integrator which the NUTS implementation currently does not, and needs to recalculate this for the same set of nodes, `z`.
- [x] The potential energy evaluation is further repeated alongside each call to the integrator (same as the base implementation, but the number of calls in the base case are `num_steps` less).
- [x] Enable tests 2, 3 in `test_nuts_conjugate_gaussian` to run locally preferably within 1-2 mins.

neerajprad:
Closing major perf issues. Will run some more profiling and create issues if needed.

PR #800, opened 2018-02-22
Skip slow NUTS tests for local runs

neerajprad:
These are already skipped on CI, but we can be re-enable these to run locally once #799 is resolved. Currently these tests take up a lot of time when testing locally.


fritzo:
wow this test is really slow :smile:

Issue #802, opened 2018-02-22
OMTMultivariateNormal errors with PyTorch master


neerajprad:
The `test_omt_mvn` tests fail with PyTorch master, because in the backward step, at the following line:

```
torch.trtrs(identity, L.t(), transpose=False, upper=True)
```

`identity` has type `Tensor` but `L.t()` has type `Variable`. cc. @martinjankowiak, @fritzo 

<details>
<summary> stack trace </summary>

```

tests/distributions/test_omt_mvn.py:23 (test_mean_gradient[sample_shape0-0.5-0.6-0.4])
sample_shape = torch.Size([1000, 1000]), L21 = 0.4, omega1 = 0.5, L11 = 0.6
L22 = 0.8, L33 = 0.9, omega2 = 0.75

    @pytest.mark.parametrize('L21', [0.4, 1.1])
    @pytest.mark.parametrize('L11', [0.6, 0.95])
    @pytest.mark.parametrize('omega1', [0.5, 0.9])
    @pytest.mark.parametrize('sample_shape', [torch.Size([1000, 1000]), torch.Size([100000])])
    #@pytest.mark.skip(reason="Errors on PyTorch master")
    def test_mean_gradient(sample_shape, L21, omega1, L11, L22=0.8, L33=0.9, omega2=0.75):
        omega = Variable(torch.Tensor([omega1, omega2, 0.0]))
        mu = Variable(torch.zeros(3), requires_grad=True)
        zero_vec = [0.0, 0.0, 0.0]
        off_diag = Variable(torch.Tensor([zero_vec, [L21, 0.0, 0.0], zero_vec]), requires_grad=True)
        L = Variable(torch.diag(torch.Tensor([L11, L22, L33]))) + off_diag
    
        dist = OMTMultivariateNormal(mu, L)
        z = dist.rsample(sample_shape)
>       torch.cos((omega*z).sum(-1)).mean().backward()

test_omt_mvn.py:38: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../../../miniconda2/envs/pytorch/lib/python2.7/site-packages/torch/autograd/variable.py:128: in backward
    torch.autograd.backward(self, gradient, retain_graph, create_graph)
../../../../../miniconda2/envs/pytorch/lib/python2.7/site-packages/torch/autograd/__init__.py:83: in backward
[...]

PR #803, opened 2018-02-23
Restore OMTMultivariateNormal test

neerajprad:
Based on @fehiepsi's diagnosis of the issue in #802.

### DIFF
--- a/pyro/distributions/omt_mvn.py
+++ b/pyro/distributions/omt_mvn.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import torch
-from torch.autograd import Function
+from torch.autograd import Function, variable
 from torch.autograd.function import once_differentiable
 from torch.distributions import constraints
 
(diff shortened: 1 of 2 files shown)


pyro/distributions/omt_mvn.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import torch
-from torch.autograd import Function
+from torch.autograd import Function, variable
```
fehiepsi: @neerajprad is this new style in Pytorch to create variables?

neerajprad: @fehiepsi - `variable` was introduced sometime back to incorporate support for scalars, e.g. using `variable(1)` will give a 0 dim tensor wrapped inside a `Variable`. In that sense, it is more general than `Variable`. I think soon Variable and Tensor APIs will be fused into a single `tensor` type.

neerajprad:
> great thanks! i would have fixed it myself but i don't want to switch to master quite yet.

Np at all, this was a really small fix. Thanks to @fehiepsi for debugging.

PR #804, opened 2018-02-23
Reduce time for NUTS tests

fehiepsi:
With this change, time for test 2 is 131s and test 3 is 31s on my local machine.

@neerajprad 
https://github.com/uber/pyro/issues/799

### DIFF
--- a/tests/infer/mcmc/test_nuts.py
+++ b/tests/infer/mcmc/test_nuts.py
@@ -1,7 +1,8 @@
 from __future__ import absolute_import, division, print_function
 
-import logging
 from collections import defaultdict
+import logging
+import os
 
 import pytest
 import torch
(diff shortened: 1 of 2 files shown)


tests/infer/mcmc/test_nuts.py
```diff
@@ -13,18 +14,19 @@
...
 
-from .test_hmc import rmse, TEST_CASES, TEST_IDS
+from .test_hmc import rmse, T, TEST_CASES, TEST_IDS
 
 logging.basicConfig(format='%(levelname)s %(message)s')
 logger = logging.getLogger('pyro')
 logger.setLevel(logging.INFO)
 
 TEST_CASES[0] = TEST_CASES[0]._replace(mean_tol=0.04, std_tol=0.04)
 TEST_CASES[1] = TEST_CASES[1]._replace(mean_tol=0.04, std_tol=0.04)
```
neerajprad: At this point, I feel that it might just be better to copy over the test cases. :) That will be more readable.

fehiepsi: Do you want to make that change now? I think that it is better to do it if there are more samplers come. This way makes the difference between HMC and NUTS easier to see: more precise for test 0, test 1, fewer samples for test 2 test 3.

neerajprad: Makes sense. Let's punt on that now, but it might be worth thinking later how best to refactor these cases to support different variants like hmc, nuts or nuts with dual averaging.

neerajprad:
Thanks for tuning these tests for NUTS!

PR #805, opened 2018-02-23
[Done] Add various kernels for GP module

fehiepsi:
This pull request adds various kernels for GP module, which addresses #702. Most of the implementations and names follow GPML book, chapter 4.

- [x] Isotropic kernels: Exponential (Matern12), Matern32, Matern52, RationalQuadratic, RBF(SquaredExponential).
- [x] Static kernels: Constant (Bias), WhiteNoise
- [x] Dot-product kernels: Polynomial, Linear.
- [x] Periodic kernels: Cosine, SineSquaredExponential (Periodic).
- [x] Brownian kernel.
- [x] Add tests for all kernels

This pull request does not implement Combination (Add, Product) kernels. The main reason is: it is not straightforward to get parameters of a combination kernel. For example, when we define `k = k1 + k2 + k3`, then k is understood as `Add(Add(k1, k2), k3)`. It is a Combination of Combinations, so it is not clear how to set priors to parameters of k1 (e.g.). Due to this problem, I will address it in another pull request. One idea I came up with is to save an infix expression. As a consequence, `k` is a Combination of a OrderDict of kernels together with an infix expression (which is saved as a string `"({} + {}) + {}"`. The forward pass will be processed based on this infix expression.

If anyone has requirements for adding more kernels or has other ideas to solve the above problem of Combination kernels, please let me know in comments. Thanks in advance!

@fritzo @ysaatchi


fehiepsi:
@fritzo  @ysaatchi  I include here the corresponding results for tests in scikit-learn and GPy to make comparision.
https://gist.github.com/fehiepsi/cf55756ea44dbf87bbb51ab331738a87

eb8680:
The kernel implementations seem fine since the test results match scikit-learn and GPy.  You might want to double check their numerical stability under the various solve operations used in the GP models.

> it is not straightforward to get parameters of a combination kernel

Can you just use `Add(Add(k1, k2), k3).named_parameters()`, as in the implementation of `pyro.module`?  Since all your kernels are `nn.Module`s, that will list the parameters of `k1`, `k2`, and `k3` with properly scoped names as long as `Add` stores its arguments as attributes.

fehiepsi:
@eb8680 Nice idea! I will test each kernel for some GP model. :)

For combination kernels, the situation is a little bit different.

Assume I define
```
k1 = RBF(1)
k2 = Matern12(1)
k3 = Matern32(1)
k4 = Matern52(1)
k = ((k1 + k2) + k3) * k4
```
Then it is easy to set priors to, set constraints to, or fix parameters of `k1`, `k2`, `k3`, `k4` by calling e.g. `k1.set_constraint("lengthscale", constraints.postive)`. However, people usually define kernels as:
```
k = (RBF(1) + Matern12(1) + Matern32(1)) * Matern52
```
Now, the question is: from `k`, how to access parameters of RBF kernel (in the sense that we not only want to get values of these parameters, but also want to set priors/constraints on them or fix them) (I already create [a derived class](https://github.com/uber/pyro/blob/dev/pyro/contrib/gp/__init__.py#L11) from nn.Module to be able to do these jobs).

My idea is to use: `k.get_kernel("RBF")` to get the RBF kernel, and then we can access its parameters easily. To achieve that, all four sub-kernels will be saved as a list of sub-modules (rather than a hierarchy of modules) together with an infix expression for the forward pass.

eb8680:
> Now, the question is: from k, how to access parameters of RBF kernel (in the sense that we not only want to get values of these parameters, but also want to set priors/constraints on them or fix them)

What about using `k.modules()`/`k.named_modules()` then?  That will get you the RBF kernel plus a name indicating its place in the expression.

fehiepsi:
@eb8680 I didn't know about `.named_modules()`. After a look at Pytorch docs, I think that we can achieve the target by using it. I will try to explore and get back to you soon. Thanks a lot!

fehiepsi:
@eb8680 I tested all implemented kernels with GPR and detected a big issue with Isotropic kernels which require `torch.sqrt()`. The formula for distance is `r(x, z) = sqrt((x - z)**2)`. So in case x = z then we have to take `sqrt` at 0. However, the backward of `sqrt` at 0 is [NaN in pytorch](https://github.com/pytorch/pytorch/issues/2421). As suggested by a comment in that thread, I have made a convenient function: `_torch_sqrt(x, eps=1e-18)` which return `(x + eps).sqrt()`. This helps avoid that issue. As I can see from the test, all kernels work well:
 https://gist.github.com/fehiepsi/f97a131a15d6e5cb00d30706d67b6729

For polynomial kernel with high degree, I often see the Cholesky decomposition error: the matrix is not positive definite. This is understandable because with the polynomial kernel `k(x, z) = (1 + x.z)^4` (for example), the values of `K(X, X)` are large. Hence the noise or jitter cannot help avoid such "non-positive definite" error. Beyond that, I didn't detect any else issue.

fehiepsi:
@eb8680 This is ready for final review. I will add combination kernels in the next pull request. :)

ysaatchi:
This looks awesome @fehiepsi lgtm! Can you not use abs() to sidestep the sqrt() issue? It's quite minor and your fix is fine I think.

fehiepsi:
@ysaatchi The NaN issue does not happen for `torch.abs` or `torch.norm`. We can use `torch.norm` to calculate distance matrix. However, it is computational ineffective. For a test
```
import torch
X = torch.rand(1000, 1000)
Z = torch.rand(1000, 1000)
def method1(X, Z):
    X2 = (X**2).sum(dim=1, keepdim=True)
    Z2 = (Z**2).sum(dim=1, keepdim=True)
    XZ = X.matmul(Z.t())
    r2 = X2 - 2*XZ + Z2.t()
    return r2.sqrt()
def method2(X, Z):
    X_Z = X.unsqueeze(1) - Z.unsqueeze(0)
    return torch.norm(X_Z, dim=2)
# check if they return the same result
print((method1(X, Z) - method2(X,Z)).norm())
```
It tooks 17ms for method1 and more than 6s for method2. In addition, we need to store a matrix of size $NxMxD$, which is memorical ineffectively.

ysaatchi:
@fehiepsi understood, method2 is actually quite cool, unfortunate that it is so much slower!

PR #812, opened 2018-02-25
Organize distributions module; fix make docs

fritzo:
This moves remaining Pyro distributions from distributions.torch.* to distributions.*, and makes pyro.distributions.torch a single-file module rather. This is just cleanup after our migration to PyTorch distributions, and makes our distributions file locations more consistent.

This PR also fixes `make docs` to work locally. I've locally run `make docs` to ensure the reorganized documentation is still generated correctly. The fix is to guard our readthedocs hacks by `if 'READTHECOCS' in os.environ: ...` which [should only be run on readthedocs servers](http://docs.readthedocs.io/en/latest/builds.html#build-environment).

Finally this rebalances some unit tests to avoid timeout on travis.


docs/source/conf.py
```diff
@@ -186,9 +186,10 @@ def setup(app):
     app.connect("autodoc-skip-member", skip)
 """
 
-# awful hack to get rtd builder to install pytorch
-os.system('pip install awscli')
-os.system('aws s3 --no-sign-request sync s3://pyro-ppl/ci tmp --exclude "*" --include "*-cp27-*";')
-os.system('pip install tmp/*')
-os.system('rm -r tmp')
-os.system('pip install -e .')
+# @jpchen's hack to get rtd builder to install pytorch
```
jpchen: thats what `git blame` is for :P

fritzo: I always find it useful when modifying config files to add a comment on who added a section, especially if it was I who modified the section. That way later on I can see how my config differs from the original. While `git blame` can do some of this, it can easily get lost due to a file move.

fritzo:
> isnt make docs run in travis?

you're right, and I don't know why it's failing locally for me yet passing on travis. Anyway this PR should speed up `make docs` everywhere except readthedocs.

PR #813, opened 2018-02-25
Add tests for Reshape distribution; fix some bugs

fritzo:
This adds unit tests for `.reshape(sample_shape, extra_event_dims)` and fixes three bugs:
- `.reshape(sample_shape2).reshape(sample_shape2)` is now applied in the correct order
- `.reshape(sample_shape).mean` now has the same shape as `...sample()`
- `.reshape(sample_shape).variance` now has the same shape as `...sample()`

### DIFF
--- a/pyro/distributions/torch_distribution.py
+++ b/pyro/distributions/torch_distribution.py
@@ -167,10 +167,10 @@ class Reshape(TorchDistribution):
         return self.base_dist.has_enumerate_support
 
     def sample(self, sample_shape=torch.Size()):
-        return self.base_dist.sample(self.sample_shape + sample_shape)
+        return self.base_dist.sample(sample_shape + self.sample_shape)
 
     def rsample(self, sample_shape=torch.Size()):
-        return self.base_dist.rsample(self.sample_shape + sample_shape)
+        return self.base_dist.rsample(sample_shape + self.sample_shape)
 
     def log_prob(self, value):
         return sum_rightmost(self.base_dist.log_prob(value), self.extra_event_dims)
@@ -194,8 +194,8 @@ class Reshape(TorchDistribution):
 
     @property
     def mean(self):
-        return self.base_dist.mean
+        return self.base_dist.mean.expand(self.batch_shape + self.event_shape)
 
     @property
     def variance(self):
-        return self.base_dist.variance
+        return self.base_dist.variance.expand(self.batch_shape + self.event_shape)
--- a/tests/distributions/test_reshape.py
+++ b/tests/distributions/test_reshape.py
@@ -1,0 +1,53 @@
+from __future__ import absolute_import, division, print_function
+
+import pytest
+import torch
+from torch.autograd import Variable
+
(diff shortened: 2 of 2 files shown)


pyro/distributions/torch_distribution.py
```diff
@@ -167,10 +167,10 @@ def has_enumerate_support(self):
         return self.base_dist.has_enumerate_support
 
     def sample(self, sample_shape=torch.Size()):
-        return self.base_dist.sample(self.sample_shape + sample_shape)
+        return self.base_dist.sample(sample_shape + self.sample_shape)
```
jpchen: what is the difference between this change and the way it was?

tests/distributions/test_reshape.py
```diff
@@ -0,0 +1,53 @@
...
+
+import pytest
+import torch
+from torch.autograd import Variable
+
+from pyro.distributions.torch import Bernoulli
+
+
+@pytest.mark.parametrize('sample_dim,extra_event_dims',
+                         [(s, e) for s in range(4) for e in range(4 + s)])
```
neerajprad: Not exactly related to this PR, but it seems like right now we do not place any constraints on `extra_event_dims`. Would it be the case that `extra_event_dims <= len(self.base_dist.batch_shape)`? I am guessing that it should at least be `<= sample_shape + self.base_dist.batch_shape`, but correct me if I am mistaken.

fritzo: Good point. The requirement is that `extra_event_dims < len(sample_shape + batch_shape)`. In particular it's fine if `extra_event_dims > len(batch_shape)`. I've added a check to this PR and added death tests to verify the check.

pyro/distributions/torch_distribution.py
fritzo: The desired behavior is for `.reshape()` to add shape on the left, and `.sample()` to add shape on the left of that. Thus the following should be equivalent:
```py
assert dist.reshape(shape1).sample(shape2).shape \
    == dist.sample(shape2 + shape1).shape
```
For example if you want to replace a 3-dimensional multivariate normal with a diag normal, the following should have the same shape:
```py
dist1 = MultivariateNormal(variable([0, 0, 0]), ...)
dist2 = Normal(0, 1).reshape(sample_shape=3, extra_event_dims=3)  # diag normal
assert dist1.event_shape == dist2.event_shape
assert dist1.batch_shape == dist2.batch_shape
assert dist1.sample([100]).shape == (100, 3)
assert dist2.sample([100]).shape == (100, 3)
```
I've added a simple test to verify this behavior explicitly.

neerajprad:
Thanks for adding in the new checks! Some of the enum tests are failing though due to the new assertion, you may need to tweak the test fixtures. Looks good otherwise.

pyro/distributions/torch_distribution.py
martinjankowiak: thanks this is helpful for my understanding. i would have expected ```extra_event_dims=1```?

PR #806, opened 2018-02-23
Adopt strict batch shape semantics for distributions

fritzo:
See [Design Doc](https://docs.google.com/document/d/1EpAnAIotJj24-yS_8h_5KEiYWYop7zHcAErSeO7XAZc/edit) | Blocking #780

This PR adds checks for stricter use of batch dimensions in Pyro. This is only recently possible since it relies on PyTorch support for scalars.

## Why?

Pyro 0.2 will be able to automatically introduce batch dimensions for parallelizing `enum_discrete` and `num_particles`. To give Pyro space to create new batch dimensions, we need to restrict how dimensions are introduced by user code.

## How?

The new requirements are:
- users must declare non-batch dimensions using `.reshape(extra_event_dims=n)`
- batch dimensions will only be allowed inside `iarange` (and maybe `irange`?)
- to use auto-parallelization, users must request dims via `SVI(..., max_iarange_nesting=n)`

These requirements are checked by the new helper `check_site_shape` which uses the new `.size` field in `cond_indep_stack` frames.

## Tasks

- [x] check sites in `Trace_ELBO`
- [x] check sites in `TraceGraph_ELBO`
- [x] update tests/infer/test_valid_models.py
- [x] update other tests
- [x] update examples
- [x] update tutorials
- [ ] document new batching semantics

PR #821, opened 2018-02-26
Implement TorchDistribution.mask() for dependent masks

fritzo:
Blocking #806

This adds a `TorchDistribution.mask(-)` method that returns a `MaskedDistribution` wrapper.

This is needed to apply masks to dependent variables, since `poutine.scale` can only mask out independent variables. This is typically only used just before declaring `extra_event_dims`, as in
```py
pyro.sample("x",
            dist.Normal(loc, scale)
                .mask(mask)                    # first apply mask
                .reshape(extra_event_dims=1))  # then sum out the rightmost dim
```
This is needed for the DMM example for #806 .

### DIFF
--- a/examples/dmm/dmm.py
+++ b/examples/dmm/dmm.py
@@ -193,16 +193,19 @@ class DMM(nn.Module):
             # first compute the parameters of the diagonal gaussian distribution p(z_t | z_{t-1})
             z_mu, z_sigma = self.trans(z_prev)
             # then sample z_t according to dist.Normal(z_mu, z_sigma)
-            with poutine.scale(None, annealing_factor * mini_batch_mask[:, t - 1:t]):
-                z_t = pyro.sample("z_%d" % t, dist.Normal(z_mu, z_sigma))
+            with poutine.scale(None, annealing_factor):
+                z_t = pyro.sample("z_%d" % t,
+                                  dist.Normal(z_mu, z_sigma)
+                                      .mask(mini_batch_mask[:, t - 1:t]))
 
             # compute the probabilities that parameterize the bernoulli likelihood
             emission_probs_t = self.emitter(z_t)
             # the next statement instructs pyro to observe x_t according to the
             # bernoulli distribution p(x_t|z_t)
-            with poutine.scale(None, mini_batch_mask[:, t - 1:t]):
-                pyro.sample("obs_x_%d" % t, dist.Bernoulli(emission_probs_t),
-                            obs=mini_batch[:, t - 1, :])
+            pyro.sample("obs_x_%d" % t,
+                        dist.Bernoulli(emission_probs_t)
+                            .mask(mini_batch_mask[:, t - 1:t]),
+                        obs=mini_batch[:, t - 1, :])
             # the latent sampled at this time step will be conditioned upon
             # in the next time step so keep track of it
             z_prev = z_t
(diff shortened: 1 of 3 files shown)


pyro/distributions/torch_distribution.py
```diff
@@ -71,9 +71,20 @@ def reshape(self, sample_shape=torch.Size(), extra_event_dims=0):
...
+        Masks a distribution by a zero-one tensor that is broadcastable to the
+        distributions ``batch_shape``.
+
+        :param Variable mask: A zero-one valued float tensor.
+        :return: A masked copy of this distribution.
+        :rtype: :class:`MaskedDistribution`
+        """
+        return MaskedDistribution(self, mask)
 
     def analytic_mean(self):
```
martinjankowiak: nit: but masking probably kills analytic_mean etc.?

pyro/distributions/torch_distribution.py
```diff
@@ -202,3 +213,50 @@ def mean(self):
...
+class MaskedDistribution(TorchDistribution):
+    """
+    Masks a distribution by a zero-one tensor that is broadcastable to the
+    distributions ``batch_shape``.
+
+    :param Variable mask: A zero-one valued float tensor.
+    """
+    def __init__(self, base_dist, mask):
+        if broadcast_shape(mask.shape, base_dist.batch_shape) != base_dist.batch_shape:
+            raise ValueError("Expectewd mask.shape to be broadcastable to base_dist.batch_shape, "
```
martinjankowiak: typo: Expectewd

examples/dmm/dmm.py
```diff
@@ -193,16 +193,19 @@ def model(self, mini_batch, mini_batch_reversed, mini_batch_mask,
             # first compute the parameters of the diagonal gaussian distribution p(z_t | z_{t-1})
             z_mu, z_sigma = self.trans(z_prev)
             # then sample z_t according to dist.Normal(z_mu, z_sigma)
-            with poutine.scale(None, annealing_factor * mini_batch_mask[:, t - 1:t]):
```
neerajprad: Do we also need to change the notebook?

martinjankowiak: i think we should hold off on cleaning up tutorials until we've finalized all the breaking 0.2 changes, semantics, etc

pyro/distributions/torch_distribution.py
```diff
@@ -71,9 +71,20 @@ def reshape(self, sample_shape=torch.Size(), extra_event_dims=0):
...
         """
-        return Reshape(self, sample_shape, extra_event_dims)
+        return ReshapedDistribution(self, sample_shape, extra_event_dims)
+
+    def mask(self, mask):
+        """
+        Masks a distribution by a zero-one tensor that is broadcastable to the
+        distributions ``batch_shape``.
+
+        :param Variable mask: A zero-one valued float tensor.
```
neerajprad: I am liking this pattern!

examples/dmm/dmm.py
neerajprad: Sounds good, as long as our tutorials test pass.

pyro/distributions/torch_distribution.py
fritzo: Just as with `.sample()`, it is up to the user to ignore masked-out values of `.mean`.

pyro/distributions/torch_distribution.py
fritzo: Fixed. ugh, now for another travis cycle.

fritzo:
Feel free to merge before tests pass, I've made only a single-character change since last time they passed.

Issue #818, opened 2018-02-26
Remove networkx dependency from Trace


eb8680:
Having briefly tried and failed to fix the memory leak (#798) at its source, I propose rewriting `Trace` as a simplified custom object (effectively a dict of `OrderedDict`s) with an option to export to a networkx graph.

fritzo:
Sounds reasonable. I'd like to hear your networkx war story.

eb8680:
Will reopen if/when we revisit.

PR #822, opened 2018-02-27
Remove networkx from Trace

eb8680:
This PR, the first in a series of `Trace` cleanup PRs, replaces the old `networkx.DiGraph`-derived `Trace` with a new class that duplicates the parts of the old API used in Pyro.  It seems to (mostly) work, and just needs documentation and a couple of additional tests.

Addresses #818


pyro/poutine/trace.py
```diff
@@ -20,103 +19,83 @@ def _warn_if_nan(name, value):
...
-
-    @property
-    def remove_node(self):
-        return self._graph.remove_node
-
-    @property
-    def add_edge(self):
-        return self._graph.add_edge
+        # metadata
+        self.graph = collections.OrderedDict()
```
fritzo: what is `self.graph`?

eb8680: It's a collection of global metadata in a `networkx.DiGraph`.  I couldn't remember if it was used in Pyro, so I kept that part of the API, but will check now that things work.

neerajprad:
Nice! We can also remove the dependency from `setup.py` now and remove the outdated tests from `test_memory.py`. 

@fritzo - Should we update the wheels and get #824 merged first?

fritzo:
> Should we update the wheels?

Yes, I think #824 is blocking some refactoring in #806. I think it's highest priority to catch up with PyTorch master. (if only so I can get my questions answered on PyTorch slack 😉 ).

fritzo:
@eb8680 I suspect your changes are orthogonal to #824, and should lead to no merge conflict.

eb8680:
This PR isn't super urgent or blocking anything, so feel free to introduce merge conflicts.

neerajprad:
> @eb8680 I suspect your changes are orthogonal to #824, and should lead to no merge conflict.

It won't lead to merge conflicts, but the tests here will fail in the interim.

fehiepsi:
Just a quick observation: With this pull request, it seems that main Pyro only requires `torch` and `six` (for compatible py2 and py3). Other packages such as `observations` is only used in examples; `cloudpickle` can be replaced by `torch.save`, `torch.load` (?); `numpy` is only used in examples and tests (and it is a dependence of torch); `graphviz` is only used in `save_visualization` util function (so can be made as an optional package). ^^

fritzo:
> numpy is only used in ...

we'd need to replace `np.isnan` with `math.isnan` (@martinjankowiak let's avoid `np.isnan`)

martinjankowiak:
thanks for doing this! can you please add a descendants method? i need this to make testing more robust.

eb8680:
> can you please add a descendants method?

Done

pyro/poutine/trace.py
```diff
@@ -20,103 +19,112 @@ def _warn_if_nan(name, value):
...
+
+    def iter_edges(self):
+        """
+        Iterates (parent, child) pairs corresponding to edges in the trace.
+        """
+        for node_from in self.edges:
+            for node_to in self.edges[node_from]:
+                yield (node_from, node_to)
+
+    def topological_sort(self):
```
fritzo: We could use a test for `topological_sort()`

pyro/poutine/trace.py
```diff
@@ -20,103 +19,112 @@ def _warn_if_nan(name, value):
...
-
-
 class Trace(object):
     """
-    Execution trace data structure
+    Execution trace data structure.
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, graph_type=None):
```
fritzo: why not default to `graph_type="flat"`?

pyro/poutine/trace.py
```diff
@@ -20,103 +19,112 @@ def _warn_if_nan(name, value):
...
+        new_trace.graph.update(self.graph.copy())
+        new_trace.nodes.update(self.nodes.copy())
+        new_trace.edges.update(self.edges.copy())
+        return new_trace
+
+    def iter_edges(self):
+        """
+        Iterates (parent, child) pairs corresponding to edges in the trace.
+        """
+        for node_from in self.edges:
```
fritzo: nit: it's a tiny bit cheaper to iterate over dictionary items rather than keys, because it avoids a hash lookup:
```py
for node_from, edges_from in self.edges.items():
    for node_to in edges_from:
        yield node_from, node_to
```

pyro/poutine/trace.py
```diff
@@ -20,103 +19,112 @@ def _warn_if_nan(name, value):
...
+                yield (node_from, node_to)
+
+    def topological_sort(self):
+        """
+        Computes topological ordering of sites in the trace.
+        """
+        indegree_map = {node: 0 for node in self.nodes}
+        for node_from, node_to in self.iter_edges():
+            indegree_map[node_to] += 1
+        zero_indegree = [node for node in indegree_map if indegree_map[node] == 0]
```
fritzo: To avoid nondeterminism, you should either make `indegree_map` an `OrderedDict` or sort `zero_indegree` before entering the `while` loop.

martinjankowiak:
great thanks!

martinjankowiak:
ok once we get this merged i'll have a test that compares a brute force non-recursive downstream cost computation to the one that's done in tracegraph_elbo.... yay!

eb8680:
I left the old `Trace` API intact but replaced the internals with a more uniform functional API.  Once this gets merged, I'll get rid of the old API entirely.

pyro/poutine/trace.py
```diff
@@ -20,105 +19,265 @@ def _warn_if_nan(name, value):
...
+    def descendants(self, site_name):
+        """
+        :param string site_name: name of site to get descendants of
 
-    def __len__(self):
-        return len(self._graph)
+        Returns a list of all sites downstream of site_name
+        """
+        desc = set()
+        for succ in self.edges[site_name]:
```
fritzo: This implementation is exponential in number of nodes. Why not make it linear via depth-first search?
```py
frontier = set([site_name])
desc = set()
while frontier:
    node = frontier.pop()
    if node not in desc:
        desc.add(node)
        frontier.update(set(self.edges[node]) - desc)
desc.remove(site_name)
return list(sorted(desc))
```

pyro/poutine/trace.py
```diff
@@ -20,105 +19,265 @@ def _warn_if_nan(name, value):
...
-        but raises an error when attempting to add a duplicate node
-        instead of silently overwriting.
+        Adds a site with metadata in the kwargs to the trace.
+        Raises an error when attempting to add a duplicate sample node.
         """
         # XXX should do more validation than this
         if kwargs["type"] != "param":
             assert site_name not in self, \
                 "site {} already in trace".format(site_name)
+        self.nodes[site_name] = dict(**kwargs)
```
fritzo: nit: it's cheaper to `dict(kwargs)`

pyro/poutine/trace.py
```diff
@@ -20,105 +19,265 @@ def _warn_if_nan(name, value):
...
+        Adds an edge with optional metadata to the trace.
+        Both sites must already be in the trace.
+        """
+        assert node_from in self, \
+            "node_from {} not in trace".format(node_from)
+        assert node_to in self, \
+            "node_to {} not in trace".format(node_to)
+        assert node_to not in self.edges[node_from], \
+            "edge from {} to {} already exists".format(node_from, node_to)
+        self.edges[node_from][node_to] = dict(**kwargs)
```
fritzo: nit: it's cheaper to `dict(kwargs)`

pyro/poutine/trace.py
```diff
@@ -20,105 +19,265 @@ def _warn_if_nan(name, value):
...
+    ret = collections.OrderedDict()
+    for name, node in nodes.items():
+        if fn(node):
+            ret[name] = node
+    return ret
+
+
+def log_pdf(site):
+    try:
+        site_log_p = site["log_pdf"]
```
fritzo: `log_pdf()` would be a great name if we were using Haskell. Since we're using Python it would be better to name this `site_log_pdf()` or `get_log_pdf()` something narrower and verby.

Also, it's simpler to
```py
def log_pdf(site):
    try:
        return site["log_pdf"]
    except KeyError:
        ...compute and store...

File: pyro/poutine/trace.py [hunk 10]
```diff
@@ -127,93 +286,58 @@ def log_pdf(self, site_filter=lambda name, site: True):
...
-                    args, kwargs = site["args"], site["kwargs"]
-                    site_log_p = site["fn"].log_prob(site["value"], *args, **kwargs)
-                    site_log_p = scale_tensor(site_log_p, site["scale"]).sum()
-                    site["log_pdf"] = site_log_p
-                    _warn_if_nan(name, site_log_p)
-                log_p += site_log_p
-        return log_p
-
-    def compute_batch_log_pdf(self, site_filter=lambda name, site: True):
+        return fold_trace(lambda log_p, msg: log_p + log_pdf(msg),
```
fritzo: As a lover of lambda calculus, I still have to :-1: this use of lambdas in Python. A for loop would be clearer than `fold`.

pyro/poutine/trace.py
```diff
@@ -127,93 +286,58 @@ def log_pdf(self, site_filter=lambda name, site: True):
...
-                try:
-                    site["batch_log_pdf"]
-                except KeyError:
-                    args, kwargs = site["args"], site["kwargs"]
-                    site_log_p = site["fn"].log_prob(site["value"], *args, **kwargs)
-                    site_log_p = scale_tensor(site_log_p, site["scale"])
-                    site["batch_log_pdf"] = site_log_p
-                    site["log_pdf"] = site_log_p.sum()
-                    _warn_if_nan(name, site["log_pdf"])
+        map_trace(batch_log_pdf,
```
fritzo: This would be clearer as a for loop.

pyro/poutine/trace.py
```diff
@@ -20,105 +19,265 @@ def _warn_if_nan(name, value):
...
+        args, kwargs = site["args"], site["kwargs"]
+        site_log_p = site["fn"].log_prob(site["value"], *args, **kwargs)
+        site_log_p = scale_tensor(site_log_p, site["scale"])
+        site["batch_log_pdf"] = site_log_p
+        site["log_pdf"] = site_log_p.sum()
+        _warn_if_nan(site["name"], site["log_pdf"])
+    return site["batch_log_pdf"]
+
+
+def score_parts(site):
```
fritzo: Rename to `get_score_parts()` or similar verb

pyro/poutine/trace.py
```diff
@@ -20,105 +19,265 @@ def _warn_if_nan(name, value):
...
+    except KeyError:
+        args, kwargs = site["args"], site["kwargs"]
+        site_log_p = site["fn"].log_prob(site["value"], *args, **kwargs)
+        site_log_p = scale_tensor(site_log_p, site["scale"]).sum()
+        site["log_pdf"] = site_log_p
+        _warn_if_nan(site["name"], site_log_p)
+    return site["log_pdf"]
+
+
+def batch_log_pdf(site):
```
fritzo: rename to `get_batch_log_pdf()` or similar verb

fritzo:
I look forward to the two or three PRs into which this is split.

I also look forward to having more profiling tooling in place to show the benefits of ditching networkx and to ensure that functionalizing our low-level operators avoids performance regressions. @neerajprad and I have been chatting about possible profiling tests for SVI and HMC.

eb8680:
> I look forward to the two or three PRs into which this is split.

Since it's a fairly lightweight installation dependency, unless networkx turns out to be a performance bottleneck this PR isn't really urgent.

PR #780, opened 2018-02-15
working towards nested iarange in tracegraph_elbo

martinjankowiak:
add nested iarange support to tracegraph_elbo. clean-up to follow in #816

PR #824, opened 2018-02-27
Resolve changes in tensor/variable API with PyTorch master

### LINKED ISSUE #815: Support PyTorch after Variable+Tensor were merged
Pyro is behind PyTorch master and needs to be updated now that `Tensor` and `Variable` were merged. This is logically orthogonal to the rest of our refactoring, although there will be lots of superficial merge conflicts.

neerajprad:
This resolves a few minor issues so as to keep in sync with PyTorch master.
 - `tensor[0]` returns a `Variable` scalar instead of a python numeric type. We need to use `.item()` to convert a PyTorch scalar into a python numeric type.
 - `.data` returns an instance of type `Variable` and not `torch.Tensor`. Instead of using `.data.cpu().numpy()`, we are now using `.detach().cpu().numpy()`.

NOTE: Tests will fail until we update the PyTorch wheels for travis. The core changes are really small, so would suggest reviewing only once the tests pass.

Only partially resolves #815. We will need more renaming changes later.


neerajprad:
PyTorch wheels are updated; running tests against the updated wheel now

examples/air/viz.py
```diff
@@ -35,7 +35,7 @@ def draw_one(imgarr, z_arr):
     # Note that this clipping makes the visualisation somewhat
     # misleading, as it incorrectly suggests objects occlude one
     # another.
-    clipped = np.clip(imgarr.data.cpu().numpy(), 0, 1)
+    clipped = np.clip(imgarr.detach().cpu().numpy(), 0, 1)
```
fritzo: Is it safe to replace `.detach().cpu().numpy()` with `.cpu().numpy()` everywhere?

pyro/infer/abstract_infer.py
```diff
@@ -151,4 +151,4 @@ def __call__(self, *args, **kwargs):
         if not isinstance(logits, torch.autograd.Variable):
             logits = Variable(logits)
         ix = dist.Categorical(logits=logits).sample()
-        return traces[ix.data[0]]
+        return traces[ix]
```
fritzo: Why not `ix.item()` here? (This may not be perfectly tested)

tests/distributions/test_rejector.py
```diff
@@ -149,8 +149,8 @@ def test_shape_augmented_beta(alpha, beta):
     z = dist.rsample()
     cost = z.sum()
     (cost + cost.detach() * dist.score_parts(z)[1]).backward()
-    mean_alpha_grad = alphas.grad.data.mean()
-    mean_beta_grad = betas.grad.data.mean()
+    mean_alpha_grad = alphas.grad.data.mean().item()
```
fritzo: Why not simply `alphas.grad.mean().item()`?

fritzo:
@neerajprad has #780 been merged into this branch? I believe that introduces some new helpers like `MultiViewTensor` that will need to be updated.

examples/air/viz.py
neerajprad: That's what I started with. If the node has `requires_grad=True`, it will throw an exception when we convert it to numpy without detaching.

pyro/infer/abstract_infer.py
neerajprad: Because a scalar is a valid index value. We can change it `ix.item()` but it's not necessary.

fritzo:
Update README.md to recommend installing PyTorch commit `05269b5`?

tests/distributions/test_rejector.py
neerajprad: Yup, in this case we can just use that. Will change here and other patterns that I see.

pyro/infer/abstract_infer.py
fritzo: This is great! It will be so much easier to write mixture models now.

neerajprad:
>> Update README.md to recommend installing PyTorch commit 05269b5?

Eeks..I always forget that. Thanks for the reminder!

neerajprad:
>> @neerajprad has #780 been merged into this branch? I believe that introduces some new helpers like MultiViewTensor that will need to be updated.

Thanks for the heads up. Will take a look and update.

pyro/infer/abstract_infer.py
neerajprad: Completely; scalars will simplify a number of clunky looking stuff.

neerajprad:
Ready to merge, unless there are any further pending comments.

fehiepsi:
@neerajprad I would like to know about the changes on this. As a summary, we will:
- Use `.item()` to get Python value of a scalar.
- Use `.tolist()` to convert a tensor to Python list.
- Use `.data` to make `requires_grad=False`.
- Use `.detach()` to separate a tensor from the graph and returns a non-grad tensor. It works like a combination of `.data` and `.clone` (`.clone` alone still requires grad).
- Use `.numpy()` on a non-grad non-cuda tensor to convert it to numpy array. They use the same memory.
- No need to use pyro's utils `ng_zeros`, `ng_ones`. Use `torch.ones`, `torch.zeros` instead.
- Use `torch.tensor(..., dtype=..., requires_grad=...)` to create a tensor with the corresponding type, grad. The previous ways `torch.FloatTensor(...)`, `torch.cuda.DoubleTensor(...)` also work.
- `b = torch.tensor(...).type_as(a)` is the same as `b = torch.tensor(..., dtype=a.dtype)`.
- `.sum()` always return a scalar tensor.

neerajprad:
Thanks for the summary, @fehiepsi.

> Use .tolist() to convert a tensor to Python list.

Interesting, didn't know about this.

> Use .detach() to separate a tensor from the graph and returns a non-grad tensor. It works like a combination of .data and .clone (.clone alone on a has-grad tensor still requires grad).

Is this new behavior, or planned for the future? `.detach()` used to share the same underlying data, and still does on the PyTorch branch that I am using. This is also what PyTorch recommends in the error that it throws `(RuntimeError: Can't call numpy() on Variable that requires grad. Use var.detach().numpy() instead.)`

> Use torch.tensor(..., dtype=..., requires_grad=...) to create a tensor with the corresponding type, grad. The previous ways torch.FloatTensor(...), torch.cuda.DoubleTensor(...) also work.

Still waiting on these changes in PyTorch master. We will use `torch.tensor` everywhere once the changes are merged.

fehiepsi:
@neerajprad That it my mistake. :( I will ask on PyTorch's slack what is the difference between `x.data` and `x.detach()` and get back to you. They seem identical now. For `torch.tensor`, it is already committed to Pytorch master. :)

fehiepsi:
@neerajprad As answered by Adam, they are very similar except that `x.detach()` has some additional checks for `in-place` operators (I am not clear how these checks work on a detached Variable though). So in my opinion, they can be used interchangeably. :)

PR #830, opened 2018-03-01
add brute force comparison for downstream costs test

martinjankowiak:
compares the recursive computation to the quadratic one

### DIFF
--- a/tests/infer/test_compute_downstream_costs.py
+++ b/tests/infer/test_compute_downstream_costs.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 import math
 
+import networkx
 import pytest
 import torch
 from torch.autograd import Variable, variable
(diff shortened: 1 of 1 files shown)


tests/infer/test_compute_downstream_costs.py
```diff
@@ -10,10 +11,59 @@
 import pyro.distributions as dist
 import pyro.poutine as poutine
 from pyro.infer.tracegraph_elbo import _compute_downstream_costs
+from pyro.infer.util import MultiViewTensor
 from pyro.poutine.util import prune_subsample_sites
 from tests.common import assert_equal
 
 
+def _brute_force_compute_downstream_costs(model_trace, guide_trace,  #
```
fritzo: note: there's no need to prefix with underscore since this is a test file

fritzo:
ready to merge?

martinjankowiak:
think so yes

PR #831, opened 2018-03-01
[Ready] Implement composed kernels

fehiepsi:
This pull request implements methods to construct new kernels from old kernels. Most of the methods (and their names) are taken from GPML book, chapter 4. The exponent method is taken from Chapter 6 of Bishop's book.

- [x] Add sum/direct sum, product/tensor product kernels.
- [x] Add exponent kernel: k_new = e^k.
- [x] Add vertical scaling kernel: k_new(x, z) = f(x) k(x, z) f(z).
- [x] Add warping kernel (including warping input by a function and output by a polynomial with positive coefficient): k_new(x, z) = q(k(f(x), f(z))).
- [x] Make tests for these methods.
- [x] Add sphinx documentation.


fehiepsi:
@ysaatchi Could you please help me review the approaches to these methods? I don't have experience with using these methods so I might have made something wrong. The implementation could be fine when the tests pass.

pyro/contrib/gp/kernels/kernel.py
```diff
@@ -52,3 +68,238 @@ def _slice_input(self, X):
...
+class Combination(Kernel):
+    """
+    Base class for Sum and Product kernels.
+
+    :param pyro.contrib.gp.kernels.Kernel kernel0: First kernel to combine.
+    :param pyro.contrib.gp.kernels.Kernel or numbers.Number kernel1: Second kernel to combine.
+    """
+
+    def __init__(self, kern0, kern1, name=None):
+        if not (isinstance(kern0, Kernel) and
```
fritzo: nit: This would be clearer as a pair of checks:
```py
if not isinstance(kern0, Kernel):
    raise TypeError(...)
if not isinstance(kern1, (Kernel, numbers.Number)):
    raise TypeError(...)
```

pyro/contrib/gp/kernels/kernel.py
```diff
@@ -10,6 +13,16 @@ class Kernel(Parameterized):
     Every inherited class should implement the forward pass which
     take inputs X, Z and return their covariance matrix.
 
+    To construct a new kernel from the old ones, we can use Python operators
+    `+` or `*` with another kernel/constant. Or we can use methods `.add(...)`,
```
fritzo: You could use sphinx notation:
```
:meth:`add`, :meth:`mul`, ...
```

pyro/contrib/gp/kernels/kernel.py
```diff
@@ -52,3 +68,238 @@ def _slice_input(self, X):
...
+        where ``f := vscaling_fn`` is a vertical scaling function.
+        """
+        return VerticalScaling(self, vscaling_fn, name=name)
+
+    def warp(self, iwarping_fn=None, owarping_coef=None, name=None):
+        """
+        Creates a new kernel, which is an instance of Warping kernel, according to:
+        ``k_new(x, z) = q(k(f(x), f(z)))``,
+        where ``f := iwarping_fn`` is a input warping function, and ``q`` is a polynomial
+        with non-negative coefficients ``owarping_coef``.
```
fritzo: Could you add sphinx documentation
```
:param callable iwarping_fn: An input warping function.
:param list owarping_coeff: A list of polynomial coefficients by which to warp the output.
:returns: A warped kernel.
:rtype: Warping
```

pyro/contrib/gp/kernels/kernel.py
```diff
@@ -52,3 +68,238 @@ def _slice_input(self, X):
...
+        else:
+            raise KeyError("There is no subkernel with the name '{}'.".format(name))
+
+
+class Combination(Kernel):
+    """
+    Base class for Sum and Product kernels.
+
+    :param pyro.contrib.gp.kernels.Kernel kernel0: First kernel to combine.
+    :param pyro.contrib.gp.kernels.Kernel or numbers.Number kernel1: Second kernel to combine.
```
fritzo: Can you confirm that this generates proper documentation? I'm not sure that `X or Y` types work inside `:param:` and I think you might need to move that out to a separate line `:type kernel1: X or Y`.

pyro/contrib/gp/kernels/kernel.py
```diff
@@ -52,3 +68,238 @@ def _slice_input(self, X):
...
+        self.vscaling_fn = vscaling_fn
+
+    def forward(self, X, Z=None, diag=False):
+        if diag:
+            return self.vscaling_fn(X) * self.kern(X, Z, diag) * self.vscaling_fn(X)
+        elif Z is None:
+            vscaled_X = self.vscaling_fn(X).unsqueeze(1)
+            return vscaled_X * self.kern(X, Z, diag) * vscaled_X.t()
+        else:
+            return self.vscaling_fn(X).unsqueeze(1) * self.kern(X, Z, diag) * self.vscaling_fn(Z).unsqueeze(0)
```
fritzo: Will we ever support batching? If so we should index from the right: `...unsqueeze(-1) * ...unsqueeze(-2)`.

pyro/contrib/gp/kernels/kernel.py
```diff
@@ -52,3 +68,238 @@ def _slice_input(self, X):
...
+    """
+
+    def forward(self, X, Z=None, diag=False):
+        if isinstance(self.kern1, Kernel):
+            return self.kern0(X, Z, diag) * self.kern1(X, Z, diag)
+        else:  # constant
+            return self.kern0(X, Z, diag) * self.kern1
+
+
+class Deriving(Kernel):
```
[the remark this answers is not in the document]
fehiepsi: It is not a common name. I originally used the name `Transform` but then saw that it was available in `torch.distributions` too so I changed it to `Deriving`. Use the name `Transforming` seems good to me.

pyro/contrib/gp/kernels/kernel.py
fehiepsi: Thanks, @fritzo , didn't know about that. Just discovered that we can type some math with :math:`...` too.

pyro/contrib/gp/kernels/kernel.py
fehiepsi: Oops, I forgot to make docs for this pull request. >"<

pyro/contrib/gp/kernels/kernel.py
fehiepsi: @fritzo I think that I need to find some references to GP with batch. Will address this issue in a future pull request, when I figure out how batching works. Right now I think that it is good to leave as it is, so the tests for GP with batch will throw wrong answers.

fehiepsi:
@fritzo I have addressed your comments and used `:meth:` `:math:` `:class:` in all docs of GP modules. It looks much better now. Thanks for your recommendation!

fehiepsi:
The fail is due to "The job exceeded the maximum time limit..."

pyro/contrib/gp/kernels/kernel.py
```diff
@@ -52,3 +67,267 @@ def _slice_input(self, X):
             return X.unsqueeze(1)
         else:
             raise ValueError("Input X must be either 1 or 2 dimensional.")
+
+    def __add__(self, other):
+        return self.add(other)
```
eb8680: I'm not a huge fan of operator overloading; in my experience it tends to significantly reduce readability.  I'd prefer keeping combination kernels explicit: users should always have to write `Sum(k1, k2)`.

fehiepsi: So I will uncomment them and remove the relevant part in docs. We can enable them again in the future if necessary.

pyro/contrib/gp/kernels/kernel.py
```diff
@@ -52,3 +66,267 @@ def _slice_input(self, X):
...
+    # def __mul__(self, other):
+    #     return self.mul(other)
+    #
+    # def __radd__(self, other):
+    #     return self.add(other)
+    #
+    # def __rmul__(self, other):
+    #     return self.mul(other)
+
+    def add(self, other, name=None):
```
eb8680: I'd say go ahead and remove these methods as well.

fehiepsi: Ya, I get your point!

neerajprad:
@fehiepsi - the build time out issue should be fixed with #843. You may find a couple of tests that timeout and you may need to restart them though.

pyro/contrib/gp/kernels/kernel.py
```diff
@@ -52,3 +63,267 @@ def _slice_input(self, X):
...
+    # def __mul__(self, other):
+    #     return self.mul(other)
+    #
+    # def __radd__(self, other):
+    #     return self.add(other)
+    #
+    # def __rmul__(self, other):
+    #     return self.mul(other)
+    #
+    # def add(self, other, name=None):
```
fritzo: I kind of like the method forms, and we're using this pattern for `Distribution.reshape()` and `Distribution.mask()`. @eb8680 would you be ok with `.add()` and `.mul()` methods being the default usage?

@fehiepsi Let's keep commented code out of the codebase, since that leads to lots of false positives when grepping.

fehiepsi: Yup, I'll remove them or enable them again based on the outcome of this discussion. :)

eb8680: > would you be ok with .add() and .mul() methods being the default usage?

Since you don't mind, it's fine with me as well

fehiepsi: So I reversed behaviors. :)

pyro/contrib/gp/kernels/isotropic.py
```diff
@@ -104,9 +104,10 @@ def __init__(self, input_dim, variance=None, lengthscale=None, active_dims=None,
 
 class RationalQuadratic(Isotropy):
     """
-    Implementation of Rational Quadratic kernel: ``(1 + 0.5 * r^2 / alpha l^2)^(-alpha)``.
+    Implementation of RationalQuadratic kernel
+    :math:`\\left(1 + 0.5 \\times \\frac{r^2}{\\alpha l^2}\\right)^{-\\alpha}`.
```
fritzo: nit: It's cleaner to use raw string literals for docstrings with backslashes, e.g. this is what PyTorch does:
```py
r"""
Implementation of ...
:math:`\left(...\alpha}`.
"""
```
Could you use raw docstring literals throughout and remove the double backslashes? (this will help avoid accidental escapes like in `:math:\exp\\left`above :wink:.

PR #837, opened 2018-03-01
Improve documentation for distributions

fritzo:
This makes a few improvements to our distributions documentation, as promised to @karalets .
- uses intersphinx to link to PyTorch documentation (master)
- adds all PyTorch distributions to the TOC
- adds the `__call__` method to sphinx documentation
- documents a couple properties

See the results at http://fritzo.org/temp/pyro/html/distributions.html

### DIFF
--- a/docs/source/distributions.rst
+++ b/docs/source/distributions.rst
@@ -6,16 +6,22 @@ Distributions
    :maxdepth: 2
    :caption: Contents:
 
-Primitive Distributions
-~~~~~~~~~~~~~~~~~~~~~~~~~
-
 PyTorch Distributions
----------------------
+~~~~~~~~~~~~~~~~~~~~~
+
+Most distributions in Pyro are thin wrappers around PyTorch distributions.
+For details on the PyTorch distribution interface, see
+:class:`torch.distributions.Distribution`.
+For differences between the Pyro and PyTorch interfaces, see
+:class:`pyro.distributions.torch_distribution.TorchDistributionMixin`.
 
 .. automodule:: pyro.distributions.torch
     :members:
     :undoc-members:
-    :show-inheritance:
+    :special-members: __call__
+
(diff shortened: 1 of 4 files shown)


pyro/distributions/torch.py
```diff
@@ -101,13 +100,31 @@ def __init__(self, a, b):
...
-        continue
 
-    class _PyroDist(_Dist, TorchDistributionMixin):
-        pass
 
-    _PyroDist.__name__ = _name
-    locals()[_name] = _PyroDist
-    del _PyroDist, _name, _Dist
+# Create sphinx documentation.
+__doc__ = '\n\n'.join([
```
fritzo: This was the only way I was able to get these distributions to show up in our sphinx toc.

fritzo:
@jpchen will this update our readthedocs page as soon as it's merged?

docs/source/distributions.rst
```diff
@@ -6,23 +6,30 @@ Distributions
...
+For differences between the Pyro and PyTorch interfaces, see
+:class:`pyro.distributions.torch_distribution.TorchDistributionMixin`.
 
 .. automodule:: pyro.distributions.torch
     :members:
     :undoc-members:
-    :show-inheritance:
+    :special-members: __call__
+
+Primitive Distributions
```
neerajprad: Is it okay to have multivariate normal implementations under Primitive Distributions, or should we create a header like `Pyro Distributions` as a parking space for distributions that are not yet migrated to PyTorch (or better kept in Pyro like the Delta distribution)?

jpchen:
> will this update our readthedocs page as soon as it's merged

yes.

taking a look now

pyro/distributions/torch.py
jpchen: oof. gross

PR #838, opened 2018-03-02
Minor cleanup to verlet integrator

neerajprad:
This is no longer needed as we are passing regular tensors inside the integrator, and lifting them to variables only when we need to compute the potential.


fritzo:
Can I trust that unit test coverage is high enough to detect an error if this were incorrect?

neerajprad:
> Can I trust that unit test coverage is high enough to detect an error if this were incorrect?

Yes, you can! One of the integration batches seem to have timed out which is unrelated.

fehiepsi:
I think that this is correct, because we don't modify items of the dict.

Issue #840, opened 2018-03-02
Slow test clean up time


neerajprad:
Here is our winning list of tests taking longer than 2 minutes. Let us figure out a plan for these - remove them (I can run them as a cron job on my workstation weekly), or modify them to run within 2 minutes ideally. 

I am taking the liberty to remove the first one, which was being skipped but was added back by me in #806. Addressing these slowest tests will give a respite to all Pyro developers who have been facing build timeouts due to slow tests.

#### More than 5 mins - should definitely be removed, or changed
- [ ] **519.81s** tests/infer/test_inference.py::PoissonGammaTests::test_elbo_nonreparameterized
- [ ] **406.33s** tests/infer/test_inference.py::NormalNormalTests::test_elbo_nonreparameterized
- [x] **341.14s** tests/infer/test_sampling.py::ImportanceTest::test_importance_prior
- [ ] **328.93s** tests/infer/test_sampling.py::ImportanceTest::test_importance_guide
- [ ] **302.56s** tests/infer/test_inference.py::LogNormalNormalTests::test_elbo_nonreparameterized

#### More than 3 mins
- [ ] **274.57s** tests/integration_tests/test_conjugate_gaussian_models.py::GaussianPyramidTests::test_elbo_nonreparameterized_three_layers
 - [ ] **268.61s** tests/infer/test_inference.py::LogNormalNormalTests::test_elbo_with_transformed_distribution
 - [ ] **264.82s** tests/infer/test_inference.py::LogNormalNormalTests::test_elbo_reparameterized
 - [ ] **264.01s** tests/infer/test_inference.py::ExponentialGammaTests::test_elbo_nonreparameterized
 - [ ] **229.62s** tests/infer/test_inference.py::BernoulliBetaTests::test_elbo_reparameterized
 - [ ] **223.50s** tests/infer/test_inference.py::PoissonGammaTests::test_elbo_reparameterized
 - [ ] **206.49s** tests/infer/test_inference.py::BernoulliBetaTests::test_elbo_nonreparameterized
 - [ ] **181.70s** tests/infer/test_inference.py::ExponentialGammaTests::test_elbo_rsvi

#### More than 2 mins
 - [ ] **153.72s** tests/integration_tests/test_conjugate_gaussian_models.py::GaussianPyramidTests::test_elbo_reparameterized_three_layers
 - [ ] **136.95s** tests/infer/test_inference.py::ExponentialGammaTests::test_elbo_reparameterized
[...]

PR #841, opened 2018-03-02
Remove slow tests resulting in build timeout

neerajprad:
Addresses #840.


fritzo:
Why don't we just fix Pyro to not be so slow?

neerajprad:
The test that this removes took more than 8 mins (it was only recently reinstated by me). Will close if #842 fixes this (it does seem to take care of it in python 3.5 from what I saw earlier).

neerajprad:
Just skips the `test_importance_prior` test as discussed in #840.

Issue #839, opened 2018-03-02
Investigate Pyro slowness in Python 2.7 


neerajprad:
Python 2.7 seems to be an order of magnitude slower than 3.6 for certain models. One of the potential sources of slowness seems to be in `enum.py` queue which takes almost 11 seconds inside `poutine.__init__.queue` method in python 2.7 but only takes 0.5 seconds in python 3.6 for the non-reparametrized Poisson Gamma test (being removed in #841). Profiles are 
attached and can be visualized using `snakeviz {*.prof}`.

[profile.zip](https://github.com/uber/pyro/files/1774145/Archive.zip)

fritzo:
This points to @eb8680 's refactoring #775 which causes `queue` to create a `range` of 1e6 items each time it is called. `range` creates a list in Python 2 and an iterator in Python 3.

@eb8680 the reason I had in-lined that logic in the first place was to avoid the bad `max_tries` logic.

fritzo:
@neerajprad would it be easy to add a script in `profiler/` to generate the profiles we used here (and maybe open snakeviz)? I found it very helpful to see what function was taking all the time. I think it would benefit the whole Pyro team if we could all use a push-button profiling tool to ensure our refactorings are safe. (I've certainly found such tools useful in the past.)

neerajprad:
> @neerajprad would it be easy to add a script in profiler/ to generate the profiles we used here (and maybe open snakeviz)? I found it very helpful to see what function was taking all the time. I think it would benefit the whole Pyro team if we could all use a push-button profiling tool to ensure our refactorings are safe. (I've certainly found such tools useful in the past.)

I will be adding the model as a regression test, and I can create a `make` command to run cProfile on any of these models and start a snakeviz session. 

On a personal note, I have not found automated profiling that useful simply because it is exploratory in nature - profiling requirements are usually motivated by very specific questions (e.g. in this case, why is this model so much slower in 2.7 vs 3.6), and it is more effort maintaining a set of global examples than just running cProfile on the specific instance (I literally just copied that test, ran cProfile and snakeviz). The important thing is knowing that a regression has occurred, and an example that triggered it, which can motivate the right profiling questions to ask.

PR #842, opened 2018-03-02
Fix performance regression in queue

fritzo:
Addresses #839

This attempts to fix a performance regression in #775 by removing `max_tries` logic from `queue` and using cheaper `max_tries` logic in `Sample` (the only place `max_tries` is be used).


eb8680:
This logic is incorrect.  `max_tries` is the maximum number of times to try returning a single value from `queue`, not the number of times to invoke it.

eb8680:
Oops sorry didn't mean to close

fritzo:
@eb8680 Can you create a simpler PR that fixes this by
```py
from six.moves import xrange

for i in xrange(max_tries):
   ...
```

PR #843, opened 2018-03-02
Replace range with xrange in poutine.queue

eb8680:
Addresses #839 and #842


neerajprad:
This one liner has certainly fixed almost all our tests - there are only a very few that take more than 2 minutes now!

Issue #825, opened 2018-02-27
Parallelize computation of Marginal and Posterior


eb8680:
Along the lines of #791, we should allow global independence dimensions in `Marginal` and `TracePosterior`.

ngoodman:
great idea. (related: make importance parallel.)

this is possible only when the underlying computation is batchable. it might be sometimes hard to tell if this is so. is it possible to dynamically detect a failure of batching and bail out to the serial method?

fritzo:
`Search` was removed in #1044 . As I understand, `Search` would be reimplemented as part of this issue.

fritzo:
@eb8680 Does this issue also cover paralleization / batch-support for `SVI`? I'd like to implement batched inference to expose our EM solvers (BP + Newton).

eb8680:
> Does this issue also cover paralleization / batch-support for SVI?

What did you have in mind?

fritzo:
> @eb8680 What did you have in mind?

I've written some details in #1213

fritzo:
@eb8680 is this still on track for our 0.3 release with PyTorch 1.0?

eb8680:
> is this still on track for our 0.3 release with PyTorch 1.0?

We don't have an immediate use case, so probably not

PR #845, opened 2018-03-03
[Only for discussion] Explore possible ways to make Histogram faster

neerajprad:
While taking a look at why some sampling tests take so much time, I found that the source of the bottleneck is the `_eq` method that tries to repeatedly compare tensors. 

I replaced that by hashing the byte representation of the underlying `tensor.numpy()` arrays instead and that results in much faster tests. e.g. the `test_importance_guide` now finishes in 5 secs vs 341 secs earlier. I did some basic checks that seem to bear out, but I need to dig deeper into what guarantees we have by using this, and what bases it doesn't cover. Is something like this good enough for our use case? Can we do something more robust?

Note that there is currently an issue tracking hash functions for tensor data in PyTorch - https://github.com/pytorch/pytorch/issues/2569. 

cc. @eb8680

### DIFF
--- a/pyro/infer/abstract_infer.py
+++ b/pyro/infer/abstract_infer.py
@@ -1,40 +1,13 @@
 from __future__ import absolute_import, division, print_function
 
+from collections import OrderedDict
+
 import torch
 from torch.autograd import Variable
 
 import pyro.distributions as dist
 import pyro.poutine as poutine
 import pyro.util as util
-
-
-def _eq(x, y):
-    """
-    Equality comparison for nested data structures with tensors.
-    """
-    if type(x) is not type(y):
-        return False
-    elif isinstance(x, dict):
-        if set(x.keys()) != set(y.keys()):
-            return False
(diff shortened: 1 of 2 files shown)


eb8680:
> Is something like this good enough for our use case? Can we do something more robust?

There are three ways the current proposal could be problematic:

1. Batched `Marginal` construction #825 
2. Extending `Marginal` to use kernels other than the equality kernel, important for scoring continuous values
3. Unhashable return types like lists

I think if we can just make this a preprocessing step within `Marginal` that returns unique values and their weights then it should be gracefully usable and would definitely be a big improvement over what we have.

neerajprad:
Good points, @eb8680. Closing this PR; and look forward to the sampling refactors! I will take another look at this once the refactoring is finished. There is also an externally exposed `Empirical` distribution (in #655), which looks to be very similar to what `Histogram` is doing (except that values are provided all at once), and could benefit from similar optimizations.

PR #849, opened 2018-03-05
Support dual averaging for HMC and NUTS

fehiepsi:
This pull request implements dual averaging versions for HMC and NUTS. We can implement Algorithms 5 and 6 of NUTS paper directly but I found it a little bit counter-intuitive when looking at the pseudo-code. So I created a separate DualAveraging class to make things easier to follow (the implementation based on [Nesterov's paper on primal-dual subgradient methods](http://ium.mccme.ru/postscript/s12/GS-Nesterov%20Primal-dual.pdf)). In addition, it can be used for both HMC and NUTS. So in my view, this separation is useful.

- [x] Implement DualAveraging
- [x] Implement find_reasonable_epsilon method for HMC
- [x] Use DualAveraging for HMC
- [x] Use DualAveraging for NUTS
- [x] Create tests.


fehiepsi:
@eb8680 @neerajprad It seems that with duo averaging, if the model is sensible to step_size, then it is easy to get NaN. The reason seems lie at `mu = log(10 step_size)`. With this setting, early steps of duo averaging scheme will give big step_size. I don't know how to solve this problem. :(

neerajprad:
> @eb8680 @neerajprad It seems that with duo averaging, if the model is sensible to step_size, then it is easy to get NaN. The reason seems lie at mu = log(10 step_size). With this setting, early steps of duo averaging scheme will return big step_size, which in turns causes NaN due to the sensibility of the model. I don't know how to solve this problem without modifying it. :(

Replied in #740. I'll take a look at this PR in detail today.

fehiepsi:
@neerajprad this is ready to review. Travis failed due to an unrelated issue (cannot download a file).

pyro/infer/mcmc/hmc.py
```diff
@@ -94,45 +149,65 @@ def setup(self, *args, **kwargs):
...
-            r_mu = torch.zeros_like(node['value'])
-            r_sigma = torch.ones_like(node['value'])
+        for name, node in sorted(trace.iter_stochastic_nodes(), key=lambda x: x[0]):
+            r_mu = torch.zeros_like(node["value"])
+            r_sigma = torch.ones_like(node["value"])
             self._r_dist[name] = dist.Normal(mu=r_mu, sigma=r_sigma)
-            if node['fn'].support is not constraints.real and self._automatic_transform_enabled:
-                self.transforms[name] = biject_to(node['fn'].support).inv
-        self._validate_trace(self._prototype_trace)
+        if node["fn"].support is not constraints.real and self._automatic_transform_enabled:
```
neerajprad: This should be inside the `for` block as we want to populate transforms for each of the nodes, if applicable.

As an aside, it is always nice to separate out the refactoring related changes from the issue-specific changes, and address them in separate PRs. That makes it easier to review both of them. No need to worry about this one though, as I can see that it will be tough to do with all the merges. :)

pyro/infer/mcmc/mcmc.py
```diff
@@ -34,16 +34,18 @@ def __init__(self, kernel, num_samples, warmup_steps=0):
...
             trace = self.kernel.sample(trace)
             if t % logging_interval == 0:
-                self.logger.info('Iteration: {}.'.format(t))
+                self.logger.info("Iteration: {}.".format(t))
                 diagnostic_info = self.kernel.diagnostics()
                 if diagnostic_info is not None:
                     self.logger.info(diagnostic_info)
             if t <= self.warmup_steps:
+                if t == self.warmup_steps:
+                    self.kernel._warmup = False
```
neerajprad: If we are accessing the `._warmup` kernel attribute, we might as well rename it to `.is_warmup`.

tests/infer/mcmc/test_hmc.py
```diff
@@ -234,3 +234,69 @@ def model(data):
...
+    labels = dist.Bernoulli(logits=(true_coefs * data).sum(-1)).sample()
+
+    def model(data):
+        coefs_mean = Variable(torch.zeros(dim), requires_grad=True)
+        coefs = pyro.sample('beta', dist.Normal(coefs_mean, Variable(torch.ones(dim))))
+        y = pyro.sample('y', dist.Bernoulli(logits=(coefs * data).sum(-1)), obs=labels)
+        return y
+
+    hmc_kernel = HMC(model, step_size=None)
+    hmc_kernel._adapted = True
```
neerajprad: I think a nice API to have for HMC would be as follows:
 - `HMC(model)`: uses NUTS kernel with dual averaging.
 - `HMC(model, step_size=x)`: uses NUTS kernel without dual averaging.
 - `HMC(model, num_steps=y)`: uses HMC kernel with dual averaging.
 - `HMC(model, step_size=x, num_steps=y)`: uses vanilla HMC kernel.

With this API, we should also think about the default `warmup_steps` for the cases that we are estimating the `step_size` (it shouldn't be 0). However, we can handle this in a separate PR that cleans out the API. For the time being, we should definitely remove the `_adapted` and `_trajectory_length` attributes, and pass them into the constructor if needed.

pyro/infer/mcmc/nuts.py
```diff
@@ -180,49 +197,53 @@ def sample(self, trace):
...
                 z = new_tree.z_proposal
 
             if self._is_turning(z_left, r_left, z_right, r_right):  # stop doubling
                 break
             else:  # update tree_size
                 tree_size += new_tree.size
 
-        if is_accepted:
+        if self._adapted:
+            accept_prob = new_tree.sum_accept_probs.item() / new_tree.num_proposals
```
neerajprad: Once the warmup phase is over, shouldn't we set `self._adapted` back to False?

pyro/infer/mcmc/hmc.py
```diff
@@ -75,16 +78,68 @@ def _energy(self, z, r):
 
     def _reset(self):
         self._t = 0
+        self._accept_cnt = 0
+        # TODO: move these parameters to self.defaults dict or self.config dict of init method
+        self._target_accept_prob = 0.8  # from Stan
+        self._adapted_step_size = self.step_size if self.step_size is not None else 1
+        self._trajectory_length = 2 * math.pi  # from Stan
+        self._t0 = 10
+        self._kappa = 0.75
```
neerajprad: Can we remove these parameters as they are effectively constants (and we are not setting them via the constructor) that are replicated in `DualAveraging`? If for some reason, we need them to be different for HMC and NUTS, we can put them back in later.

pyro/infer/mcmc/hmc.py
```diff
@@ -94,45 +149,65 @@ def setup(self, *args, **kwargs):
...
-            r_sigma = torch.ones_like(node['value'])
+        for name, node in sorted(trace.iter_stochastic_nodes(), key=lambda x: x[0]):
+            r_mu = torch.zeros_like(node["value"])
+            r_sigma = torch.ones_like(node["value"])
             self._r_dist[name] = dist.Normal(mu=r_mu, sigma=r_sigma)
-            if node['fn'].support is not constraints.real and self._automatic_transform_enabled:
-                self.transforms[name] = biject_to(node['fn'].support).inv
-        self._validate_trace(self._prototype_trace)
+        if node["fn"].support is not constraints.real and self._automatic_transform_enabled:
+            self.transforms[name] = biject_to(node["fn"].support).inv
```
eb8680: Reindent these lines

tests/infer/mcmc/test_hmc.py
```diff
@@ -234,3 +234,69 @@ def model(data):
...
+    hmc_kernel._trajectory_length = 1
```
neerajprad: Why do we need to change this from its default value of `2\pi`? If we have to specify this for certain problems, we should pass it in the constructor directly, and document this argument. Also, we will need to think carefully about how some of these arguments will clash - e.g. we should not specify both the `trajectory_length` and `num_steps`. Any ideas how Stan models are able to run by just using the default value?

fehiepsi:
@eb8680 @neerajprad I will address your reviews after lunch. Just a quick reply for some points:
+ @eb8680 `find_reasonable_stepsize` is used in stan with target_accept_prob = 0.8. It seems that pymc3 doesn't use it (they might just set some initial stepsize). It is mainly used for setting `mu` (x0) for DualAveraging.
+ @neerajprad related to API, it is my intention to make them all `_foo`. Most of them should be moved into a better place (for example: `self.config` or `self.defaults`). I let these decisions for pyro's devs ^^. For the test, if we set `trajectory` as default (2pi) then it is slow (the tests pass anyway). HMC with dual averaging uses trajectory_length instead of num_steps.

pyro/infer/mcmc/mcmc.py
fehiepsi: @neerajprad I think that it is better to have a method to do it. What do you think about: `end_warmup`?

fehiepsi: Yeah, making `end_warmup` method is also useful to set step_size at the end of DA scheme (it takes the average of previous step_sizes). Previously, I have to make `warmup` and `adapted` flag to deal with that logic.

fehiepsi:
@neerajprad @eb8680 Thanks for your reviews! I have addressed your comments and the code turns out to be cleaner. The tests are identical to those of the previous version (on the non-transformed model: logistic regression). I added an additional mechanism for setting default parameters:
+ step_size=None will be set to 1
+ trajectory_length=None will be set to num_steps x step_size. If num_steps is None then set it to 2pi.
+ num_steps=int(trajectory_length/step_size). In Stan, they don't have num_steps in HMC's parameters.

pyro/infer/mcmc/hmc.py
```diff
@@ -18,17 +20,24 @@ class HMC(TraceKernel):
...
+        dynamics. If not specified, it will be set to 1.
+    :param float trajectory_length: Length of a MCMC trajectory. If not
+        specified, it will be set to ``step_size x num_steps``. In case
+        ``num_steps`` is not specified, it will be set to :math:`2\pi`.
     :param int num_steps: The number of discrete steps over which to simulate
         Hamiltonian dynamics. The state at the end of the trajectory is
-        returned as the proposal.
+        returned as the proposal. This value is always equal to
+        ``int(trajectory_length / step_size)``.
+    :param bool adapt_step_size: A flag to decide if we want to adapt step_size
```
neerajprad: Can we internally set `self._adapt_step_size = True if step_size is None else False`, for NUTS and HMC? That way we will have one less argument to worry about, and avoid invalid configs (like `adapt_step_size=True, step_size=0.2`).

pyro/infer/mcmc/hmc.py
```diff
@@ -18,17 +20,24 @@ class HMC(TraceKernel):
...
+    Radford M. Neal
 
-    :param model: python callable containing pyro primitives.
-    :param float step_size: determines the size of a single step taken by the
+    :param model: Python callable containing pyro primitives.
+    :param float step_size: Determines the size of a single step taken by the
         verlet integrator while computing the trajectory using Hamiltonian
-        dynamics.
+        dynamics. If not specified, it will be set to 1.
+    :param float trajectory_length: Length of a MCMC trajectory. If not
```
neerajprad: It will also be nice to avoid interactions between `trajectory_length`, `num_steps` and `step_size`. If we need to be able to specify `trajectory_length` for adapting the `step_size`, can we just remove `step_size` as a parameter. So, our parameters are just `HMC(model, num_steps=None, trajectory_length)`. Then if `num_steps` is not specified we just use the dual averaging version.

tests/infer/mcmc/test_nuts.py
```diff
@@ -129,3 +129,44 @@ def model(data):
...
+    nuts_kernel = NUTS(model, adapt_step_size=True)
+    mcmc_run = MCMC(nuts_kernel, num_samples=500, warmup_steps=100)
+    posterior = []
+    for trace, _ in mcmc_run._traces(data):
+        posterior.append(trace.nodes['beta']['value'])
+    posterior_mean = torch.mean(torch.stack(posterior), 0)
+    assert_equal(rmse(true_coefs, posterior_mean).item(), 0.0, prec=0.05)
+
+
+@pytest.mark.xfail(reason='the model is sensible to NaN log_pdf')
```
neerajprad: nit: `sensible` -> `sensitive`.

neerajprad:
@eb8680 - Any other comments before we get this merged?

PR #852, opened 2018-03-05
reuse conjugate gaussian chain for gradient tests

martinjankowiak:
a step in converting some (but not all) of our (expensive) integration tests into cheaper unit tests (although this will only actually be cheap once we have `num_particles` parallelized properly)

note: the changes in test_conjugate_gaussian_models.py were basically the minimum changes needed to make the test reusable (making the model and guide attributes of self and changing the argument piping accordingly)

### DIFF
--- a/tests/infer/test_conjugate_gradients.py
+++ b/tests/infer/test_conjugate_gradients.py
@@ -1,6 +1,3 @@
-from __future__ import absolute_import, division, print_function
-
-import pytest
 
 import pyro
 from pyro.infer.tracegraph_elbo import TraceGraph_ELBO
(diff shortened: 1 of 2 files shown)


tests/infer/test_conjugate_gradients.py
```diff
@@ -0,0 +1,31 @@
...
+import pyro
+from pyro.infer.tracegraph_elbo import TraceGraph_ELBO
+from tests.common import assert_equal
+from tests.integration_tests.test_conjugate_gaussian_models import GaussianChain
+
+
+# TODO increase precision and number of particles once latter is parallelized properly
+class ConjugateChainGradientTests(GaussianChain):
+
+    def test_gradients(self):
```
neerajprad: Unfortunately, `pytest.parametrize` doesn't mix well with `unittest.TestCase`. There are some work-arounds mentioned in https://docs.pytest.org/en/latest/unittest.html#mixing-fixtures, but your solution here is preferable. We should just consider migrating all `unittest.TestCase` tests to pytest at some point.

tests/integration_tests/test_conjugate_gaussian_models.py
```diff
@@ -73,6 +71,48 @@ def setup_reparam_mask(self, N):
...
+        for k in reversed(range(1, self.N + 1)):
+            mu_q = pyro.param("mu_q_%d" % k, Variable(self.target_mus[k].data +
+                                                      difficulty * (0.1 * torch.randn(1) - 0.53),
+                                                      requires_grad=True))
+            log_sig_q = pyro.param("log_sig_q_%d" % k,
+                                   Variable(-0.5 * torch.log(self.lambda_posts[k]).data +
+                                            difficulty * (0.1 * torch.randn(1) - 0.53),
+                                            requires_grad=True))
+            sig_q = torch.exp(log_sig_q)
+            kappa_q = None if k == self.N \
```
fritzo: Avoid multiline inline `if else` expressions. Just make this an `if else` statement
```py
kappa_q = None
if k == self.N:
    kappa_q = pyro.param(...)
```

tests/infer/test_conjugate_gradients.py
neerajprad: It doesn't work because `GaussianChain` inherits off of `unittest.TestCase`, and that can only run tests that do not have any arguments (pytest internally just calls `TestCase.runTest()`). We can change `GaussianChain` to inherit off of `object`, and then this would run, but we will need to manually call the `setUp` method here and in the other tests in `test_conjugate_gaussian_models`.

tests/integration_tests/test_conjugate_gaussian_models.py
martinjankowiak: done

tests/integration_tests/test_conjugate_gaussian_models.py
```diff
@@ -73,6 +71,48 @@ def setup_reparam_mask(self, N):
...
 
+    def model(self, reparameterized, difficulty=0.0):
+        next_mean = self.mu0
+        for k in range(1, self.N + 1):
+            latent_dist = dist.Normal(next_mean, torch.pow(self.lambdas[k - 1], -0.5))
+            mu_latent = pyro.sample("mu_latent_%d" % k, latent_dist)
+            next_mean = mu_latent
+
+        mu_N = next_mean
+        for i, x in enumerate(self.data):
```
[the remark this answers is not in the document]
martinjankowiak: done

PR #828, opened 2018-02-28
Fix enumerate + iarange

fritzo:
Addresses #846

This attempts to fix bugs in enumerate+iarange introduced by #776.

- [x] Fix combination of `iarange` + sequential enumeration with no broadcasting or subsampling
- [x] Fix parallel enumeration
- [x] Fix subsampling
- [ ] Fix `iarange` + sequential enumeration + broadcasting (via weights as a `MultiViewTensor`)

PR #854, opened 2018-03-06
Make cond_indep_stack a tuple (hence hashable)

fritzo:
This makes `site["cond_indep_stack"]` a tuple rather than a list.

## Why?

This makes `cond_indep_stack` hashable and therefore usable as keys in a dictionary. In #828 I'm converting `cond_indep_stack` to a tuple almost everywhere I use it. In an effort to simplify review of #828 I'm pulling out this change as an easy PR.


fritzo:
Thanks for reviewing!

PR #855, opened 2018-03-06
more downstream cost tests

martinjankowiak:
added a irange in iarange test. while at it i also fattened the ```test_compute_downstream_costs_big_model_guide_pair``` test a bit


martinjankowiak:
what would be the point of such a change? afaict the downstream formulation is more efficient in this case

fritzo:
> what would be the point of [refactoring in terms of `local_cost * upstream_grad`?

No point, I'm just curious what it would look like. I thought that if both `TraceEnum_ELBO` and `DiCE` use the `local_cost * upstream_grads` formulation it might be nice to explore how that would work in `TraceGraph_ELBO`, but if there's no benefit then indeed we should avoid unnecessary change.

PR #856, opened 2018-03-06
Fix docs links to PyTorch documentation

fritzo:
This updates Pyro docs to take advantage of intersphinx to link Pyro docs to PyTorch docs.

## Tested

- ran `make docs` locally and clicked to verify links work.

### DIFF
--- a/pyro/distributions/torch_distribution.py
+++ b/pyro/distributions/torch_distribution.py
@@ -13,7 +13,7 @@ class TorchDistributionMixin(Distribution):
 
     This is mainly useful for wrapping existing PyTorch distributions for
     use in Pyro.  Derived classes must first inherit from
-    :class:`torch.distributions.Distribution` and then inherit from
+    :class:`torch.distributions.distribution.Distribution` and then inherit from
     :class:`TorchDistributionMixin`.
     """
     @property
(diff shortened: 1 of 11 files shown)


pyro/distributions/torch_distribution.py
```diff
@@ -78,7 +81,8 @@ def shape(self, sample_shape=torch.Size()):
     def reshape(self, sample_shape=torch.Size(), extra_event_dims=0):
         """
         Reshapes a distribution by adding ``sample_shape`` to its total shape
-        and adding ``extra_event_dims`` to its ``event_shape``.
+        and adding ``extra_event_dims`` to its
+        :attr:`~torch.distributions.distribution.Distribution.event_shape`.
```
fehiepsi: Hi @fritzo, good to know that we have :func: :attr: to make docs better! Just wonder what is the role of `~`? ^^!

fritzo: The tilde says to print only the last part `event_shape` instead of the full `torch.distributions.distribution.Distribution.event_shape`. I've used this for a number of long links.

fehiepsi: Got it, thanks!

neerajprad:
Nice, LGTM! @jpchen - any comments, or should I get this merged (has some conflicting changes with #857)?

fritzo:
No need for JP's review. Ready to merge if it helps @neerajprad resolve conflicts.

PR #857, opened 2018-03-06
Using torch.tensor API in distributions module 

neerajprad:
This makes the `distributions` module compatible with PyTorch master. 
 - Removing `variable` usage as it throws a UserError.
 - Remove `ng_ones`, `ng_zeros`.

Note that this needs an intermediate PyTorch version (commit - `8327982`) which supports both the old tensor/variable constructors as well as the new; so that we can refactor the modules one by one and then update to PyTorch master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ pip install .
 
 For recent features you can install Pyro from source.
 
-First install a recent PyTorch, currently PyTorch commit `05269b5`.
+First install a recent PyTorch, currently PyTorch commit `8327982`.
 ```sh
 git clone git@github.com:pytorch/pytorch
 cd pytorch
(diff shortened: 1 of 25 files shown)


neerajprad:
@fritzo - do you mind if I update the wheels to `8327982` for travis CI? It shouldn't affect the remaining tests.

fritzo:
> do you mind if I update the wheels

Sure, if this PR is nearly ready to merge

fritzo:
@neerajprad This sounds like a very reasonable migration plan.

fehiepsi:
Yeah, after this, we won't have to worry about Pytorch's API changes. Using `torch.ones`, `torch.zeros` directly would make things much clearer!!! Thank @neerajprad !

fehiepsi:
To be honest, people will love to use Pyro with these changes. All things come down to `torch.tensor`, just like `numpy.array`.

neerajprad:
This should be good to merge now.

PR #859, opened 2018-03-07
Cleanup things a bit

fehiepsi:
This pull request cleans up things a bit. I also move things around a bit: remove cloudpickle in favor of torch.save/torch.load and move observations out of the main setup.

### DIFF
--- a/pyro/util.py
+++ b/pyro/util.py
@@ -135,7 +135,7 @@ def set_rng_seed(rng_seed):
         torch.cuda.manual_seed_all(rng_seed)
     random.seed(rng_seed)
     try:
-        import numpy
+        import numpy as np
 
         np.random.seed(rng_seed)
     except ImportError:
--- a/pyro/util.py
+++ b/pyro/util.py
@@ -5,7 +5,6 @@ import numbers
 import warnings
 
 import graphviz
-import numpy as np
 import torch
 from six.moves import zip_longest
 from torch.autograd import Variable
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -55,8 +52,10 @@
             'pytest',
             'pytest-cov',
             'nbval',
+            'numpy>=1.7',
             'scipy>=0.19.0',
             # examples/tutorials
+            'observations>=0.1.4',
```
fritzo: I believe we should also include `observations` in `notebooks` since some examples and tutorials depend on them? Or maybe we should keep it in `install_requires`? I'm not sure whether to cater to new users for whom it would be easier to include `observations` in `install_requires`, or to power users who might want to deploy Pyro with minimal dependencies. @neerajprad do you have an opinion?

pyro/util.py
```diff
@@ -128,13 +128,17 @@ def _fn(*args, **kwargs):
...
     """
-    Sets seeds of torch, numpy, and torch.cuda (if available).
+    Sets seeds of torch and torch.cuda (if available).
     :param int rng_seed: The seed value.
     """
     torch.manual_seed(rng_seed)
-    if torch.cuda.is_available():
-        torch.cuda.manual_seed_all(rng_seed)
-    np.random.seed(rng_seed)
+    random.seed(rng_seed)
```
fritzo: :+1:

setup.py
neerajprad: I think we should just add it to `notebooks` to keep core Pyro requirements separate (and small), and not have to add in ad-hoc libraries that might be needed by some example. Even `matplotlib` is required for running some of the notebooks and is not listed as a `notebook` requirement, which we should likewise add.

fehiepsi: I failed to find a good way to include `extras` into `test`, so I make a new variable EXTRAS_REQUIRE. We can keep the old way but have to modify travis config (like `pip install -e [extras,test]`).

PR #863, opened 2018-03-08
Make iarange contexts reusable

### LINKED ISSUE #370: Support irange/iarange nested inside iarange
See [Design Doc](https://docs.google.com/document/d/1EpAnAIotJj24-yS_8h_5KEiYWYop7zHcAErSeO7XAZc/edit?usp=sharing)

This proposes generalizing `TraceGraph_ELBO` to apply variance reduction techniques when an `irange` or `iarange` is nested beneath an `iarange`.

Now that Pyro distributions support arbitrarily many batch dimensions, `TraceGraph_ELBO` should take care to sum out only those dimensions of each sample site that are controlled by an `iarange`. For example, consider this model (say with a mean field guide)
```py
def model(time_series):
    N = time_series.size(-1)
    with iarange("series", N):
        z = pyro.sample("z", Bernoulli(0.5 * ng_ones(365, N)))
        pyro.sample("obs", Bernoulli(0.5 * ng_ones(365, N)),
                    obs=(time_series * x))
```
We need to ensure that `TraceGraph_ELBO` sums out dimension 0 of size 365 but does not sum out dimension 1 of size N (since that dimension is declared independent by the `iarange`).

fritzo:
This enhances `iarange` to be a reusable context manager.

## Why?

Reusable `iarange` contexts are needed to support overlapping but non-nested independence. For example, consider an image noise model with independent noise components for each row, column, and pixel:
```py
x_axis = iarange('outer', 320, dim=-1)
y_axis = iarange('outer', 200, dim=-2)
with x_axis:
    x_noise = sample("x_noise", Normal(mu, sigma).reshape([320]))
with y_axis:
    y_noise = sample("y_noise", Normal(mu, sigma).reshape([200, 1]))
with x_axis, y_axis:
    xy_noise = sample("xy_noise", Normal(mu, sigma).reshape([200, 320]))
```
This corresponds to non-nested the plate diagram
```
+----------------+
| x_noise    320 |
|                |
| +--------------------------------+
| | xy_noise     |  y_noise    200 |                
| +--------------------------------+
+----------------+
```

This PR also fixes a subsampling bug whereby variables such as `y_noise` used to be scaled by both `x_axis` and `y_axis` subsampling, but henceforth will be scaled by only `y_axis` subsampling.

## How?

This changes `iarange` to implement the full context manager interface. We now need a new `dim` arg to keep track of which tensor dim will be used later on (since this is not know at time of construction).

This functionality allows us to disallow broadcasting of log_prob within `iarange` contexts. Henceforth you can specify for each sampled tensor exactly the independence dims of that tensor.

## Tested

- added new shape checks to `check_site_shape`
- added reuse examples to `tests/infer/test_enum.py`
- added reuse examples to `tests/infer/test_valid_models.py`
- added tests for broadcasting errors to `tests/infer/test_valid_models.py`

Closes #370 (although @martinjankowiak's #780 did most of the work)

### DIFF
--- a/pyro/__init__.py
+++ b/pyro/__init__.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import, division, print_function
 
-import contextlib
 import copy
 import logging
 import warnings
@@ -165,23 +164,22 @@ def _subsample(name, size=None, subsample_size=None, subsample=None, use_cuda=No
     return size, subsample_size, subsample
 
 
-@contextlib.contextmanager
-def iarange(name, size=None, subsample_size=None, subsample=None, use_cuda=None):
+class iarange(object):
     """
     Context manager for conditionally independent ranges of variables.
 
-    :func:`iarange` is similar to :func:`torch.arange` in that it yields an
-    array of indices by which other tensors can be indexed. :func:`iarange`
+    :class:`iarange` is similar to :func:`torch.arange` in that it yields an
+    array of indices by which other tensors can be indexed. :class:`iarange`
     differs from :func:`torch.arange` in that it also informs inference
     algorithms that the variables being indexed are conditionally independent.
-    To do this, :func:`iarange` is a provided as context manager rather than a
+    To do this, :class:`iarange` is a provided as context manager rather than a
     function, and users must guarantee that all computation within an
-    :func:`iarange` context is conditionally independent::
+    :class:`iarange` context is conditionally independent::
 
         with iarange("name", size) as ind:
             # ...do conditionally independent stuff with ind...
 
-    Additionally, :func:`iarange` can take advantage of the conditional
+    Additionally, :class:`iarange` can take advantage of the conditional
     independence assumptions by subsampling the indices and informing inference
     algorithms to scale various computed values. This is typically used to
     subsample minibatches of data::
@@ -199,7 +197,7 @@ def iarange(name, size=None, subsample_size=None, subsample=None, use_cuda=None)
         independent within the context.
 
     :param str name: A unique name to help inference algorithms match
-        ``iarange`` sites between models and guides.
+        :class:`iarange` sites between models and guides.
     :param int size: Optional size of the collection being subsampled
         (like `stop` in builtin `range`).
     :param int subsample_size: Size of minibatches used in subsampling.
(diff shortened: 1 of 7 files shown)


pyro/util.py
```diff
@@ -365,10 +384,10 @@ def check_site_shape(site, max_iarange_nesting):
 
     # Check for incorrect iarange placement on the right of max_iarange_nesting.
     for actual_size, expected_size in zip_longest(reversed(actual_shape), reversed(expected_shape), fillvalue=1):
-        if expected_size != -1 and actual_size not in (1, expected_size):
+        if expected_size != -1 and expected_size != actual_size:
```
fritzo: This one line motivates the entire PR:
```diff
- if ... and actual_size not in (1, expected_size)  # allow broadcasting
+ if ... and actual_size != expected_size           # disallow broadcasting
```
With this new stricter requirement, we can more precisely track each `sample` statement's `iarange`s, and thereby support correct scaling (from subsampling) and parallel enumeration.

ngoodman:
this seems like a nice extension. 

having to be explicit about batch dimension for a context adds more work for the user, but i think it might actually be clearer in the long run than having that be magic?

pyro/__init__.py
```diff
@@ -208,36 +206,64 @@ def iarange(name, size=None, subsample_size=None, subsample=None, use_cuda=None)
...
+            raise ValueError('Expected dim < 0 to index from the right, actual {}'.format(dim))
+        self.name = name
+        self.dim = dim
+        self.size, self.subsample_size, self.subsample = _subsample(name, size, subsample_size, subsample, use_cuda)
+
+    def __enter__(self):
+        self._wrapped = am_i_wrapped()
+        if self._wrapped:
+            dim = 'auto' if self.dim is None else self.dim
+            self._scale_poutine = poutine.scale(None, self.size / self.subsample_size)
```
eb8680: You can change `scale` to `ScaleMessenger` too

pyro/__init__.py
```diff
@@ -208,36 +206,64 @@ def iarange(name, size=None, subsample_size=None, subsample=None, use_cuda=None)
...
-    else:
-        with poutine.scale(None, size / subsample_size):
-            with poutine.indep(name, vectorized=True, size=subsample_size):
-                yield subsample
+    def __init__(self, name, size=None, subsample_size=None, subsample=None, dim=None, use_cuda=None):
+        if dim is not None and dim >= 0:
+            raise ValueError('Expected dim < 0 to index from the right, actual {}'.format(dim))
+        self.name = name
+        self.dim = dim
+        self.size, self.subsample_size, self.subsample = _subsample(name, size, subsample_size, subsample, use_cuda)
```
eb8680: Is it always the case that we'll want an `iarange` instance to use the same subsampled indices?  I'm pretty sure it is, just want to double-check.

pyro/poutine/indep_poutine.py
```diff
@@ -4,7 +4,11 @@
 
 from .poutine import Messenger
 
-CondIndepStackFrame = namedtuple("CondIndepStackFrame", ["name", "counter", "vectorized", "size"])
+
+class CondIndepStackFrame(namedtuple("CondIndepStackFrame", ["name", "dim", "size", "counter"])):
```
eb8680: Very slick!

pyro/__init__.py
fritzo: Yes, we want each `iarange` to fix a single subsample for each trace, so that subsampling agrees among tensors. For example, consider adding `noise = x_noise + y_noise + xy_noise` below:
```py
def model(image):
    x_axis = iarange('outer', 320, dim=-1)
    y_axis = iarange('outer', 200, dim=-2)
    with x_axis:
        x_noise = sample("x_noise", Normal(mu, sigma).reshape([320]))
    with y_axis:
        y_noise = sample("y_noise", Normal(mu, sigma).reshape([200, 1]))
    with x_axis, y_axis:
        xy_noise = sample("xy_noise", Normal(mu, sigma).reshape([200, 320]))
        noise = x_noise + y_noise + xy_noise  # <--- these need to refer to the same indices
        pyro.observe("obs", dist.Normal(noise, 0), data=image)
```
These already broadcast correctly, but we need the subsampling to agree so that the `x` agrees between `x_noise[x]` and `xy_noise[:, x]`.

fritzo:
>  I'll push a quick PR today that splits them up.

Great idea, I agree that `pyro/__init__.py` is getting unwieldy.

eb8680:
@martinjankowiak you can merge this when you're ready

martinjankowiak:
i'd like to discuss some of the consequences of this PR w/ @fritzo before doing so

Issue #832, opened 2018-03-01
Make GP module more versatile


fehiepsi:
Continue to #702, the following steps are necessary to make GP module more accessible.

- [x] Support multidimensional output (currently, the output `y` is assumed to be 1D). This should not be complicated I guess.

- [x] Add more likelihoods: Bernoulli, Multiclass-classification,...

- [x] Support batching (need reference for this, is an option to change X, y enough?)

- [x] Use new style of `transform` with `pyro.param`

- [ ] Split/revise the current tutorial into GPR and SGPR

- [x] Go over all the docs of GP module, polish it!

- [x] Add tutorial on Bayesian Optimization

After 0.2.0 release,

- [x] Support mean_function

- [x] ~Add a tutorial on creating kernels, [warping with neural network](https://arxiv.org/pdf/1511.02222.pdf)~ Already add docs for this, and have an example which uses it on MNIST.

- [x] ~~Add a tutorial on classification (SVGP) (banana data, replicate [original paper](http://proceedings.mlr.press/v38/hensman15.pdf)).~~ Instead, we have an example on MNIST

- [x] Implement [whitening transform](https://en.wikipedia.org/wiki/Whitening_transformation), which is said to [aid optimization](https://github.com/GPflow/GPflow/blob/master/gpflow/models/vgp.py) from GPflow authors. Yes, it helps optimization!

- [x] ~Implement Gaussian-Hermite quadrature to compute the expectation of likelihood.~ This might be not necessary because doing MC is a better fit for Pyro framework.

- [x] Benchmark with other frameworks: performance is same, for speed, gpflow CPU < PyroGP CPU ~ gpflow GPU < PyroGP GPU (x < y means x is faster than y)

- [x] Add a regression test.

- [x] Implement MC version of GP models.

- [x] Implement GPLVM for dimensional reduction.

Future,

- [x] Explore deep GP.

ysaatchi:
Another extension which ought to be straightforward with pyro is transforming the inputs using neural networks before passing them to the GP. This allows for much more flexible kernel learning (cf "Deep Kernel Learning" papers)

fehiepsi:
@ysaatchi Can it be obtained by using method `.warp` in #831?

ysaatchi:
Yes, but the f(.) should have trainable parameters. What does q(.) do?

fehiepsi:
[...]

PR #866, opened 2018-03-08
[WIP] Fix 2D output for gaussian process

### LINKED ISSUE #862: Broadcasting issue with Gaussian Process
I was trying to fit some 2D data with Gaussian Process.
During the training at multivariate normal, calculating log prob fails as broadcast doesn't work. 
adding unsqueeze followed by expand_as and diag seems to fix the issue.
Code to reproduced 
```py
#%%
import torch
import pyro
from pyro.contrib.gp.kernels import RBF
from pyro.contrib.gp.models import GPRegression
from pyro.infer import SVI
from pyro.optim import Adam
from pyro.util import ng_ones, ng_zeros

pyro.set_rng_seed(0)

N = [20, 500]
X = dist.Uniform(ng_zeros(N), ng_ones(N)*5).sample()
y = 0.5 * torch.sin(3*X) + dist.Normal(ng_zeros(N), ng_ones(N)*0.5).sample()

kernel = RBF(input_dim=500, variance=torch.Tensor([10]), lengthscale=torch.Tensor([10]))
gpr = GPRegression(X, y, kernel, noise=torch.Tensor([1]))


optim = Adam({"lr": 0.001})
svi = SVI(gpr.model, gpr.guide, optim, loss="ELBO")
losses = []
num_steps = 1000
for i in range(num_steps):
    losses.append(svi.step())
```

tobyclh:
This attempts to fix #862, and also is a part of #832 that @fehiepsi 's GPR breaks when 2d output data is provided, tested on both 1d,2d input output. 
In order to fix that I change the equation of multivariate normal according to [wiki](https://en.wikipedia.org/wiki/Multivariate_normal_distribution#Likelihood_function)

### DIFF
--- a/pyro/contrib/gp/models/gpr.py
+++ b/pyro/contrib/gp/models/gpr.py
@@ -44,7 +44,7 @@ class GPRegression(Model):
         noise = self.get_param("noise")
 
         K = kernel(self.X) + noise.expand(self.num_data).diag()
-        zero_loc = torch.tensor(K.data.new([0])).expand(self.num_data)
+        zero_loc = torch.zeros_like(self.y)
         pyro.sample("y", dist.MultivariateNormal(zero_loc, K), obs=self.y)
 
     def guide(self):
--- a/pyro/distributions/multivariate_normal.py
+++ b/pyro/distributions/multivariate_normal.py
@@ -58,8 +58,9 @@ class MultivariateNormal(TorchDistribution):
 
     def log_prob(self, value):
         delta = value - self.loc
+        if len(delta.shape) == 1:
+            delta = delta.unsqueeze(-1)
         sigma_inverse = _matrix_inverse_compat(self.covariance_matrix, self.scale_tril)
-        normalization_const = ((0.5 * self.event_shape[-1]) * math.log(2 * math.pi) +
-                               self.scale_tril.diag().log().sum(-1))
-        mahalanobis_squared = (delta * torch.matmul(delta, sigma_inverse)).sum(-1)
+        normalization_const = (0.5 * self.event_shape[-1] * math.log(2 * math.pi) + self.scale_tril.diag().log().sum())
+        mahalanobis_squared = (delta.t() @ sigma_inverse @ delta).sum()
         return -(normalization_const + 0.5 * mahalanobis_squared)


pyro/contrib/gp/models/gpr.py
```diff
@@ -44,7 +44,7 @@ def model(self):
         noise = self.get_param("noise")
 
         K = kernel(self.X) + noise.expand(self.num_data).diag()
-        zero_loc = torch.tensor(K.data.new([0])).expand(self.num_data)
+        zero_loc = torch.zeros_like(self.y)
```
fehiepsi: You might change `y` shape like this: `y = self.y if self.y.dim() == 1 else self.y.t()`

pyro/distributions/multivariate_normal.py
```diff
@@ -58,8 +58,9 @@ def rsample(self, sample_shape=torch.Size()):
...
     def log_prob(self, value):
         delta = value - self.loc
+        if len(delta.shape) == 1:
+            delta = delta.unsqueeze(-1)
         sigma_inverse = _matrix_inverse_compat(self.covariance_matrix, self.scale_tril)
-        normalization_const = ((0.5 * self.event_shape[-1]) * math.log(2 * math.pi) +
-                               self.scale_tril.diag().log().sum(-1))
-        mahalanobis_squared = (delta * torch.matmul(delta, sigma_inverse)).sum(-1)
+        normalization_const = (0.5 * self.event_shape[-1] * math.log(2 * math.pi) + self.scale_tril.diag().log().sum())
+        mahalanobis_squared = (delta.t() @ sigma_inverse @ delta).sum()
```
fehiepsi: @tobyclh The implementation of `log_prob` should follow Pytorch/Pyro's `batch_shape`, `event_shape` semantic. In MultivariateNormal distribution, `event_shape` is the last dimension of `loc` (number of samples in GP). We should not change this implementation. It is better to change the observation `y` to match this semantic.

fehiepsi:
@tobyclh I think that it is better to ask users give output shape `latent_shape x num_data` rather than `num_data x latent_shape` (in that case, we only have to modify the forward pass a bit to support this change). It has several benefits:
- Implementation for likelihoods will be simplified a lot (due to the output matches Pytorch/Pyro semantic: `batch_shape x event_shape`). 
- With this change, it will be clearer to see that the latent dimensions are independent.
- Output tensor can be of any size as long as the last dimension is equal to num_data.
- We don't have to reshape the data when using outputs of Gaussian Process for other distributions such as Normal or MultivariateNormal.

The downside is it will break the users of other libraries such as GPy (which requires the first dimension of output tensor must be equal to `num_data`). We can state this difference in docs anyway.

@fritzo @martinjankowiak @ysaatchi Would you mind giving your opinions on this?

fehiepsi:
Although the proposal in my last comment fits my intuition about GP (a distribution over functions with different output dimensions use a "same" kernel) and PyTorch/Pyro distribution shape semantic, another downside is it will give an extra-effort for users to prepare data. I take a look at other libraries: scikit-learn, GPy, GPflow, and they don't prepare data that way. So I think it is better to use `num_data x latent_shape` and give a note to users that they have to transpose the output mean of GP to be able to use it for other Pyro distributions (to draw and plot samples).

@tobyclh How about letting me address this problem? There are other models to support multioutput too. And we need to add tests, docs, etc.

tobyclh:
@fehiepsi 
No problem, appreciate that :)
Then I will work on Bayesian Optimisation

PR #865, opened 2018-03-08
add partial rao-blackwellization to trace_elbo

martinjankowiak:
a coarser level of rao-blackwellization than the one in tracegraph_elbo, namely the one that obtains from only taking advantage of some iarange info

PR #867, opened 2018-03-08
Implement MultiFrameTensor data structure for Trace*_ELBO

fritzo:
This implements a `MultiFrameTensor` data structure for internal use by `Trace*_ELBO` implementations. This will replace `MultiViewTensor` in the more flexible `iarange` setting since #863.

## Tasks

- [x] Allocate `iarange` dims at the first `__enter__()` rather than later at each `pyro.sample()` statement. This is needed so that iarange `CondIndepStackFrame`s always have their `.dim` field set (no more 'auto').
- [x] Implement `MulitFrameTensor` keyed on `cond_indep_stack` tuples.
- [x] Add unit tests.
- [x] Integrate into `TraceGraph_ELBO`. 
- #865 Integrate into `Trace_ELBO`.

Pair coded by @fritzo and @martinjankowiak.

### DIFF
--- a/pyro/__init__.py
+++ b/pyro/__init__.py
@@ -13,7 +13,8 @@ import pyro.poutine as poutine
 from pyro.distributions.distribution import Distribution
 from pyro.params import _MODULE_NAMESPACE_DIVIDER, _PYRO_PARAM_STORE, param_with_module_name
 from pyro.poutine import _PYRO_STACK, condition, do  # noqa: F401
-from pyro.util import apply_stack, deep_getattr, get_tensor_data, ones, set_rng_seed, zeros, am_i_wrapped  # noqa: F401
+from pyro.poutine.indep_poutine import _DIM_ALLOCATOR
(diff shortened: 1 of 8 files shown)


martinjankowiak:
@fritzo yup. i can integrate into trace_elbo too (to replace #865). added a task accordingly

pyro/__init__.py
```diff
@@ -164,6 +164,53 @@ def _subsample(name, size=None, subsample_size=None, subsample=None, use_cuda=No
...
+        """
+        Free a dimension.
+        """
+        assert self._stack[-1 - dim] == name
+        self._stack[-1 - dim] = None
+        while self._stack and self._stack[-1] is None:
+            self._stack.pop()
+
+
+_DIM_ALLOCATOR = _DimAllocator()
```
fritzo: @eb8680 I'm a little uncomfortable introducing this extra bit of global state (with lifetime similar to `_PYRO_STACK`). Do you see any clean alternatives?

PR #869, opened 2018-03-09
Using torch.tensor API in other Pyro modules

neerajprad:
Same as #868 for `nn`, `params`, `optim` and `ops`. After this and #868, we will only need to make changes to the `infer` module.

### DIFF
--- a/pyro/nn/auto_reg_nn.py
+++ b/pyro/nn/auto_reg_nn.py
@@ -2,7 +2,6 @@ from __future__ import absolute_import, division, print_function
 
 import torch
 import torch.nn as nn
-from torch.autograd import Variable
 from torch.nn import functional as F
 
 from pyro.distributions.util import torch_multinomial
@@ -17,7 +16,7 @@ class MaskedLinear(nn.Linear):
     :param out_features: the number of output features
     :type out_features: int
     :param mask: the mask to apply to the in_features x out_features weight matrix
-    :type mask: torch.autograd.Variable
+    :type mask: torch.Tensor
     :param bias: whether or not `MaskedLinear` should include a bias term. defaults to `True`
     :type bias: bool
     """
@@ -29,7 +28,7 @@ class MaskedLinear(nn.Linear):
         """
         the forward method that does the masked linear computation and returns the result
         """
-        masked_weight = self.weight * torch.autograd.Variable(self.mask)
+        masked_weight = self.weight * self.mask
         return F.linear(_input, masked_weight, self.bias)
 
 
@@ -82,8 +81,8 @@ class AutoRegressiveNN(nn.Module):
             self.permutation = permutation
 
         # these masks control the autoregressive structure
-        self.mask1 = Variable(torch.zeros(hidden_dim, input_dim))
-        self.mask2 = Variable(torch.zeros(input_dim * self.output_dim_multiplier, hidden_dim))
+        self.mask1 = torch.zeros(hidden_dim, input_dim)
+        self.mask2 = torch.zeros(input_dim * self.output_dim_multiplier, hidden_dim)
 
         for k in range(hidden_dim):
             # fill in mask1
--- a/pyro/ops/integrator.py
+++ b/pyro/ops/integrator.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
-from torch.autograd import Variable, grad
+import torch
+from torch.autograd import grad
 
 
 def velocity_verlet(z, r, potential_fn, step_size, num_steps=1):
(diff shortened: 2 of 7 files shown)


fritzo:
It looks like this led to a performance regression :smile: Why don't we try merging #858 before this and see if it catches the error?

neerajprad:
> It looks like this led to a performance regression 😄 Why don't we try merging #858 before this and see if it catches the error?

Oh wow! :) That's rather unexpected. For sure, that'll be a good test for the profiler.

neerajprad:
There is indeed a perf regression, which was caught by the tests:
```
----------------------------------------------------------------------------------------------------------------------------------------------------------------------
Performance has regressed:
	benchmark[BernoulliBeta::HMC] (0006) - Field 'min' has failed PercentageRegressionCheck: 89.010744628 > 10.000000000
	benchmark[BernoulliBeta::NUTS] (0006) - Field 'min' has failed PercentageRegressionCheck: 155.709616561 > 10.000000000
---------------------------------------------------------------------------------------------------------------------------------
```

My guess is that we are creating new tensors when we call `torch.tensor(T, requires_grad=True)`, which was previously `Variable(T, requires_grad=True)`, and that is what was causing the regression here. It's a little worrisome that its an easy trap to fall into. The only solution I could find was to use the original `Variable` API, since we can't set `requires_grad` attribute on an already constructed tensor otherwise.

apaszke:
I don't think that's expected. What kind of copies do you have in mind? cc @colesbury @gchanan

gchanan:
Which calls exactly are you comparing?  Passing in python lists, tuples, etc?

gchanan:
Or if you have a specific benchmark I can run before and after this, I can take a look.

neerajprad:
@gchanan, @apaszke : This might just be expected behavior, but it was nevertheless something that we easily missed. We were previously using something like `Variable(t, requires_grad=True)` (`t` is a `torch.Tensor`) which was being wrapped to compute some derivaties) in a loop that is called very frequently. We changed this to `torch.tensor(t, requires_grad=True)` and saw that the runtime increased quite a bit. I suppose this is simply because `torch.tensor` copies the underlying data. 

Is wrapping inside a `Variable` still the expected way to set `requires_grad=True` on an already defined tensor so as to avoid this copying?

Issue:
```
In [3]: t = torch.ones(10000)

In [4]: %timeit Variable(t, requires_grad=True)
The slowest run took 63.79 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 927 ns per loop

In [5]: %timeit torch.tensor(t, requires_grad=True)
The slowest run took 391.60 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 7.66 us per loop
```

gchanan:
@neerajprad you should just be able to do `var.requires_grad=True` on an already defined tensor.  You are correct that `torch.tensor` and `Tensor.new_tensor` copy; that was intentional because the previous API is complex wrt copying semantics.

neerajprad:
> you should just be able to do var.requires_grad=True on an already defined tensor. 

That throws a runtime error like `RuntimeError: you can only change requires_grad flags of leaf variables`. Since these were declared as tensors in the first place, it is tough to say which of thse comprise of the leaf nodes. I then tried the following which is also a factor of 5-6 slower. It seems like using `Variable` is the best option here.

```
    for v in z.values():
        v.detach().requires_grad = True
```

apaszke:
NB: if you kept them all not requiring grad in the first place (the default for tensors), then they should already be leaves. If it complains about it in there, it means that it might be trying to think that you will want to differentiate some of paths that create `z.values()`, and this can incur extra overhead too.

fritzo:
@neerajprad It sounds like we could avoid copy overhead by entirely avoiding the `.detach()` operation and replacing `.backward()` with `torch.autograd.grad()`. Then we can avoid `Variable` entirely. (Thanks for explaining offline, @apaszke).

gchanan:
@fritzo or @apaszke can you write out the logic here for future reference?

neerajprad:
@fritzo, @apaszke - We are already using `torch.autograd.grad` and not `Variable.backward`. I tried out the following:
 - The following is still 5-6 times slower than the code below. z_nodes have `requires_grad=True` set when the tensors are constructed (otherwise this would throw an error).
```
grads = grad(potential_energy, z_nodes)
```
 - The following (in spirit) is what we have now:
```
    for i in range(z_nodes):
        z_nodes[i] = Variable(z_nodes[i].data, requires_grad=True)
    grads = grad(potential_energy, z_nodes)
```
 - The following snippet where we set `requires_grad=True` right before calling `torch.autograd.grad` and unsetting it right after, is also fast:
```
    for node in z_nodes:
        node.requires_grad = True
    grads = grad(potential_energy, z_nodes)
    for node in z_nodes:
        node.requires_grad = False
```

neerajprad:
@gchanan - I will create a PR and post the logic there.

Issue #816, opened 2018-02-26
tracegraph_elbo related cleanup


martinjankowiak:
- [x] refactor vectorized_map_data_info and related data structures
- [x] refactor way prune_subsample_sites etc. is used to setup a trace for computation
- [x] rethink any needed asserts/warnings in tracegraph_elbo or trace_poutine
- [x] go through test_valid_models and make sure we have good coverage
- [x] rethink integration/gradient test coverage (where possibly prefer cheaper tests)

PR #871, opened 2018-03-09
miscellaneous trace*_elbo cleanup

martinjankowiak:
addresses part of #816


pyro/poutine/trace_poutine.py
```diff
@@ -5,84 +5,6 @@
 from .util import site_is_subsample
 
 
-def get_iarange_info(trace):
```
fritzo: :+1: to deleted code!

PR #872, opened 2018-03-09
Make irange iterators reusable

fritzo:
This is a follow-up to #863 that makes `irange` iterators reusable.

## Why?

The main purpose of this PR is to avoid name mangling in `_Subsample` sites, and thereby to simplify reasoning about `cond_indep_stack`s in `Trace*_ELBO` implementations. Additionally this PR allows finer control of synchronized-vs-independent subsampling in nested `irange`. Synchronized subsampling is often the desired behavior but is not supported before this PR.

Consider the nested `irange` sequence that subsamples 10x10 pixels from a 320x200 image
```py
for i in irange('outer', 200, 10):
    for j in irange('inner', 320, 10):
        ...
```
Before this PR, each `irange('inner')` would be name-mangled and subsample an independent 10 pixels. After this PR this code will error, and users need to be explicit about whether they want synchronized or independent inner subsamples:
```py
def synchronized_model():
    inner_irange = irange('inner', 320, 10)
    for i in irange('outer', 200, 10):
        for j in inner_irange:  # <---- irange is reused
            ...

def independent_model():
    for i in irange('outer', 200, 10):
        for j in irange('inner_%d' % i, 320, 10):  # <---- user must give distinct names
            ...
```

## Tested

- updated many existing tests with new idiom
- added a gradient tests that relies on lack of name mangling

### DIFF
--- a/pyro/__init__.py
+++ b/pyro/__init__.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 import copy
 import logging
+import numbers
 import warnings
 from collections import OrderedDict
 from inspect import isclass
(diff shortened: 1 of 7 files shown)


pyro/__init__.py
```diff
@@ -151,9 +152,7 @@ def _subsample(name, size=None, subsample_size=None, subsample=None, use_cuda=No
         size = -1  # This is PyTorch convention for "arbitrary size"
         subsample_size = -1
     elif subsample is None:
-        names = [name]
-        names += [str(f.counter) for f in _PYRO_STACK if isinstance(f, poutine.IndepMessenger)]
-        subsample = sample("_".join(names), _Subsample(size, subsample_size, use_cuda))
+        subsample = sample(name, _Subsample(size, subsample_size, use_cuda))
```
fritzo: The dropped name-mangling logic here is the main motivation for this PR.

PR #873, opened 2018-03-09
Fix brittleness in profiling scripts

neerajprad:
This makes the perf regression and profiling utils from #858 more robust as follows:
 - Instead of checking out some reference branch, we create a `tmp` directory pulling in the single branch from github, based on @fritzo's suggestion. This has the benefit that we do not need to worry about any interactions in the current working directory.
 - A `_cleanup` function is used as an exit trap to ensure that any temporary directory created is removed when the script exits.
 - Increased number of replications.

~NOTE: If the script runs, I will make changes to run it as a cron job instead.~

### DIFF
--- a/scripts/perf_test.sh
+++ b/scripts/perf_test.sh
@@ -1,44 +1,39 @@
 #!/usr/bin/env bash
-set -x
+
+set -xe
+
+function _cleanup() {
+    [[ ${#DIRSTACK[@]} -gt 1 ]] && popd
+    [[ -d ${REF_TMP_DIR} ]] && rm -rf ${REF_TMP_DIR}
+}
+
+trap _cleanup EXIT
+
+# Reference is with respect to the `dev` branch, by default.
+REF_HEAD=${1:-dev}
+BENCHMARK_FILE=tests/perf/test_benchmark.py
+IS_BENCHMARK_FILE_IN_DEV=1
+REF_TMP_DIR=.tmp_test_dir
 
 CURRENT_HEAD=$(git rev-parse --abbrev-ref HEAD)
 # If the current state is detached head (e.g. travis), store current commit info instead.
 if [ ${CURRENT_HEAD} = 'HEAD' ]; then
-    git checkout -b tmp
     CURRENT_HEAD=$(git rev-parse HEAD)
 fi
-# Reference is with respect to the `dev` branch.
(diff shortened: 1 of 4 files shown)


neerajprad:
@fritzo, @jpchen - This is now delegated to the cron job.

jpchen:
where is the actual cron tab located?

scripts/profile_model.sh
```diff
@@ -1,44 +1,40 @@
...
-IS_BENCHMARK_FILE_IN_DEV=1
-REF_EXIT_CODE=0
-IS_DIRTY=0
 
-# If uncommitted changes exist, stash these and set a flag.
-git diff-index --quiet HEAD -- || IS_DIRTY=1
-git stash
-git checkout ${REF_HEAD}
+# clone the repo into the temporary directory and run benchmark tests
+git clone -b ${REF_HEAD} --single-branch https://github.com/uber/pyro.git ${REF_TMP_DIR}
```
fritzo: git local clones are super quick. Could this be changed to use a local clone?
```sh
git clone -b ${REF_HEAD} --single-branch . ${REF_TMP_DIR}
```

scripts/perf_test.sh
```diff
@@ -1,50 +1,47 @@
...
-IS_BENCHMARK_FILE_IN_DEV=1
-REF_EXIT_CODE=0
-IS_DIRTY=0
 
-# If uncommitted changes exist, stash these and set a flag.
-git diff-index --quiet HEAD -- || IS_DIRTY=1
-git stash
-git checkout ${REF_HEAD}
+# clone the repo into the temporary directory and run benchmark tests
+git clone -b ${REF_HEAD} --single-branch https://github.com/uber/pyro.git ${REF_TMP_DIR}
```
fritzo: ditto: can this use a local clone?

fritzo:
I'm happy to merge this now if you'd like, the local git clone is a minor comment.

scripts/profile_model.sh
neerajprad: Good idea. Let me modify this for travis vs. local runs a bit.

neerajprad:
> I'm happy to merge this now if you'd like, the local git clone is a minor comment.

Let me quickly take care of that. I think we can also do lower number of replications when we run locally.

PR #881, opened 2018-03-12
Clean up variable API usage in HMC and NUTS

neerajprad:
This cleans up the HMC/NUTS implementations:
 - Removes usage of `Variable/variable` in favor of `torch.tensor`.
 - Removes `requires_grad=True` from latent sites. Instead, we set the flag right before taking derivatives inside `integrator._grad`, and unset it right after.

NOTE: This implements @gchanan's suggestion from #869.

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -189,7 +189,7 @@ class HMC(TraceKernel):
         self._reset()
 
     def sample(self, trace):
-        z = {name: node["value"] for name, node in trace.iter_stochastic_nodes()}
+        z = {name: node["value"].data for name, node in trace.iter_stochastic_nodes()}
         # automatically transform `z` to unconstrained space, if needed.
         for name, transform in self.transforms.items():
             z[name] = transform(z[name])
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -189,7 +189,7 @@ class HMC(TraceKernel):
         self._reset()
 
     def sample(self, trace):
-        z = {name: node["value"].data for name, node in trace.iter_stochastic_nodes()}
+        z = {name: node["value"].detach() for name, node in trace.iter_stochastic_nodes()}
         # automatically transform `z` to unconstrained space, if needed.
         for name, transform in self.transforms.items():
             z[name] = transform(z[name])
--- a/pyro/ops/integrator.py
+++ b/pyro/ops/integrator.py
@@ -59,9 +59,11 @@ def single_step_velocity_verlet(z, r, potential_fn, step_size, z_grads=None):
 
 
 def _grad(potential_fn, z):
-    z = {k: Variable(v, requires_grad=True) for k, v in z.items()}
     z_keys, z_nodes = zip(*z.items())
+    for node in z_nodes:
+        node.requires_grad = True
     potential_energy = potential_fn(z)
     grads = grad(potential_energy, z_nodes)
-    grads = [v.data for v in grads]
+    for node in z_nodes:
+        node.requires_grad = False
     return dict(zip(z_keys, grads)), potential_energy
--- a/pyro/ops/integrator.py
+++ b/pyro/ops/integrator.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
-from torch.autograd import grad, Variable
+from torch.autograd import grad
 
 
 def velocity_verlet(z, r, potential_fn, step_size, num_steps=1):
--- a/tests/infer/mcmc/test_hmc.py
+++ b/tests/infer/mcmc/test_hmc.py
@@ -24,8 +24,8 @@ class GaussianChain(object):
         self.dim = dim
         self.chain_len = chain_len
         self.num_obs = num_obs
-        self.mu_0 = torch.zeros_like(torch.tensor(self.dim))
-        self.lambda_prec = torch.ones_like(torch.Tensor(self.dim))
+        self.mu_0 = torch.zeros(self.dim)
+        self.lambda_prec = torch.ones(self.dim)
 
     def model(self, data):
         mu = pyro.param('mu_0', self.mu_0)
@@ -37,7 +37,7 @@ class GaussianChain(object):
 
     @property
     def data(self):
-        return torch.ones_like(torch.Tensor(self.num_obs, self.dim))
+        return torch.ones(self.num_obs, self.dim)
 
     def id_fn(self):
         return 'dim={}_chain-len={}_num_obs={}'.format(self.dim, self.chain_len, self.num_obs)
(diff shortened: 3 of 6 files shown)


pyro/ops/integrator.py
```diff
@@ -59,9 +59,11 @@ def single_step_velocity_verlet(z, r, potential_fn, step_size, z_grads=None):
 
 
 def _grad(potential_fn, z):
-    z = {k: Variable(v, requires_grad=True) for k, v in z.items()}
     z_keys, z_nodes = zip(*z.items())
+    for node in z_nodes:
```
neerajprad: @apaszke, @gchanan - As mentioned in #869, I am setting `requires_grad=True` here right before calling `.grad` and then unsetting it later. If we set this to `True` at the time of tensor declaration, we can just call `.grad` directly here without having to set/unset these flags, but that ends up recording a lot of other tensor operations that we don't need to and is much slower as a result.

pyro/infer/mcmc/hmc.py
```diff
@@ -189,7 +189,7 @@ def cleanup(self):
         self._reset()
 
     def sample(self, trace):
-        z = {name: node["value"] for name, node in trace.iter_stochastic_nodes()}
+        z = {name: node["value"].data for name, node in trace.iter_stochastic_nodes()}
```
fritzo: Should this be `.detach()` rather than `.data`? I'm not sure the difference, and I thought we should always use `.detach()` now.

tests/infer/mcmc/test_hmc.py
```diff
@@ -25,20 +24,20 @@ def __init__(self, dim, chain_len, num_obs):
         self.dim = dim
         self.chain_len = chain_len
         self.num_obs = num_obs
-        self.mu_0 = Variable(torch.zeros_like(torch.Tensor(self.dim)), requires_grad=True)
-        self.lambda_prec = Variable(torch.ones_like(torch.Tensor(self.dim)))
+        self.mu_0 = torch.zeros_like(torch.tensor(self.dim))
```
fritzo: Why not simply `torch.zeros(self.dim)`?

tests/infer/mcmc/test_hmc.py
```diff
@@ -25,20 +24,20 @@ def __init__(self, dim, chain_len, num_obs):
...
+        self.lambda_prec = torch.ones_like(torch.Tensor(self.dim))
```
fritzo: Why not simply `torch.ones(self.dim)`?

tests/infer/mcmc/test_hmc.py
```diff
@@ -140,8 +139,8 @@ def test_hmc_conjugate_gaussian(fixture,
         param_name = 'mu_' + str(i)
         latent_mu = torch.mean(torch.stack(post_trace[param_name]), 0)
         latent_std = torch.std(torch.stack(post_trace[param_name]), 0)
-        expected_mean = Variable(torch.ones_like(torch.Tensor(fixture.dim)) * expected_means[i - 1])
-        expected_std = 1 / torch.sqrt(Variable(torch.ones_like(torch.Tensor(fixture.dim)) * expected_precs[i - 1]))
+        expected_mean = torch.ones_like(torch.tensor(fixture.dim)) * expected_means[i - 1]
```
fritzo: Consider using the cheaper `torch.Tensor(fixture.dim).fill_(expected_means[i - 1])`

tests/infer/mcmc/test_hmc.py
```diff
@@ -25,20 +24,20 @@ def __init__(self, dim, chain_len, num_obs):
...
             mu = pyro.sample('mu_{}'.format(i),
-                             dist.Normal(mu=mu, sigma=Variable(lambda_prec.data)))
-        pyro.sample('obs', dist.Normal(mu, Variable(lambda_prec.data)), obs=data)
+                             dist.Normal(mu=mu, sigma=lambda_prec))
+        pyro.sample('obs', dist.Normal(mu, lambda_prec), obs=data)
 
     @property
     def data(self):
-        return Variable(torch.ones_like(torch.Tensor(self.num_obs, self.dim)))
+        return torch.ones_like(torch.Tensor(self.num_obs, self.dim))
```
fritzo: Why not simply `torch.ones(self.num_obs, self.dim))`?

tests/infer/mcmc/test_hmc.py
```diff
@@ -181,16 +180,16 @@ def model(data):
 
 def test_bernoulli_beta():
     def model(data):
-        alpha = pyro.param('alpha', variable([1.1, 1.1], requires_grad=True))
-        beta = pyro.param('beta', variable([1.1, 1.1], requires_grad=True))
+        alpha = pyro.param('alpha', torch.tensor([1.1, 1.1]))
```
fritzo: Hmm it seems a bit un-idiomatic to call `pyro.param` with `requires_grad=False`. Can we discuss to determine whether `with torch.no_grad()` would be a suitable alternative solution that could (1) preserve our existing speed, but (2) allow more idiomatic use of grad-requiring params?

pyro/infer/mcmc/hmc.py
neerajprad: Both are actually the same, and return a new instance with requires_grad=True, sharing the same underlying data. Will change to `.detach()` for consistency.

tests/infer/mcmc/test_hmc.py
neerajprad: My bad, all of this can be simplified now, since we `torch.zeros/ones` support dytes.

tests/infer/mcmc/test_hmc.py
neerajprad: Actually, this was written in the spirit of other SVI models, but there really is no need to use the `pyro.param` primitive here. We could just use regular variables, and that might make things clearer since we are not interacting with the param store at all. Does that sound reasonable?

```
        alpha = torch.tensor([1.1, 1.1])
        beta = torch.tensor([1.1, 1.1])
```

Additionally, we are just generating samples from the latent sites and do not need `requires_grad` to be set to true on these parameter sites at all. So overall, using `pyro.param` just seems to be adding confusion here.

The only down-side (arguable) is that we can't just swap out a model in SVI, and use `HMC` to generate samples from it, but the semantics are sufficiently different anyhow. What do you think?

neerajprad:
> Let's chat about our desired idiom for pyro.param

@fritzo - I removed `pyro.param` from the HMC tests since we don't need to interact with the param store at all for sampling. Lets discuss tomorrow if you have any thoughts or reservations.

pyro/ops/integrator.py
apaszke: Yup that looks good

tests/infer/mcmc/test_hmc.py
fritzo: Thanks for addressing this. Let's merge and then discuss how to make it easy to swap out SVI for HMC.

PR #882, opened 2018-03-13
Support GP models with multidimensional output

### LINKED ISSUE #862: Broadcasting issue with Gaussian Process

fehiepsi:
This pull request attempts to fix #862 (and #832). Several changes are:
- Forward passes of variational models no longer return `likelihood`: to make consistency with API of non-variational models.
- Add `set_data` method: to support Bayesian optimization, batch training,... As a consequence, `num_data`, `num_inducing` are removed.
- Add a convenience `optimize` method to train GP models: to simplify Bayesian optimization process (which I intend to write a tutorial this week).
- Support 2D output.
- Improve tests for GP models.

Currently, 4/16 tests failed due to an issue related to `iarange`: "at site "u", invalid log_prob shape. Expected [] but got [4]". Here `4` is latent_shape of output `y`, which plays the role of `batch_shape` for MultivariteNormal distribution. It seems that Pyro does not automatically take `sum` for `log_prob`? I think that it can be solved by using `irange` directly but I guess there is a better way to handle this issue. The error message suggests the following ways:
- enclosed with `iarange`
- using `.reshape`
- using `.permute`

TODO:
- [x] Fix the above issue.


fehiepsi:
@fritzo I think that you understand the above issue much better than me. Could you help me to solve it?

pyro/contrib/gp/models/svgp.py
```diff
@@ -73,34 +77,42 @@ def guide(self):
...
         unconstrained_Lu = pyro.param("unconstrained_u_tril", unconstrained_Lu_0)
         Lu = transform_to(constraints.lower_cholesky)(unconstrained_Lu)
 
         # TODO: use scale_tril=Lu
-        Ku = Lu.t().matmul(Lu) + self.jitter.expand(self.num_inducing).diag()
-        pyro.sample("u", dist.MultivariateNormal(loc=mu, covariance_matrix=Ku))
+        Ku = Lu.t().matmul(Lu) + self.jitter.expand(Lu.size(0)).diag()
+        # correct event_shape for mu
+        mu_t = mu.t() if mu.dim() == 2 else mu
+        pyro.sample("u", dist.MultivariateNormal(loc=mu_t, covariance_matrix=Ku))
```
fritzo: If `mu_t` is batched, you'll need to either enclose in an `iarange` context
```py
if mu.dim() == 1:
    pyro.sample("u", dist.MultivariateNormal(loc=mu, covariance_matrix=Ku))
else:
    with iarange("u", mu.shape[0]):  # <---- should be minibatch size, did I get this right?
        pyro.sample("u", dist.MultivariateNormal(loc=mu.t(), covariance_matrix=Ku))
```
 or reshape with `.reshape(extra_event_dims=1)`
```py
mu_t = mu.t() if mu.dim() == 2 else mu
pyro.sample("u",
            dist.MultivariateNormal(loc=mu_t, covariance_matrix=Ku)
                .reshape(extra_event_dims=mu.dim() - 1))
```
Since everything is reparameterized here, the two forms are equivalent. If you ever foresee discrete parameters (e.g. a mixture of kernels), then the `iarange` form would be preferred.

pyro/contrib/gp/models/gpr.py
```diff
@@ -43,9 +41,11 @@ def model(self):
...
         noise = self.get_param("noise")
 
-        K = kernel(self.X) + noise.expand(self.num_data).diag()
-        zero_loc = torch.tensor(K.data.new([0])).expand(self.num_data)
-        pyro.sample("y", dist.MultivariateNormal(zero_loc, K), obs=self.y)
+        K = kernel(self.X) + noise.expand(self.X.size(0)).diag()
+        zero_loc = K.new([0]).expand(K.size(0))
+        # correct event_shape for y
+        y = self.y.t() if self.y.dim() == 2 else self.y
+        pyro.sample("y", dist.MultivariateNormal(zero_loc, K), obs=y)
```
fritzo: ditto: you'll need to either enclose in an `iarange` or `.reshape(extra_event_dims=y.dim() - 1)`

pyro/contrib/gp/models/svgp.py
fehiepsi: Thank you! I like the `reshape` style more than `iarange` in this case. :D

fehiepsi:
Thanks @fritzo ! I was able to fix that issue by using your suggestion. This pull request is ready to review now. :)

pyro/contrib/gp/models/model.py
```diff
@@ -23,6 +40,24 @@ def guide(self):
         """
         raise NotImplementedError
 
+    def optimize(self, optimizer, num_steps=1000):
```
eb8680: Add a default value for `optimizer`

tests/contrib/gp/test_models.py
```diff
@@ -1,75 +1,100 @@
...
+        pass
+    else:
+        loc, cov = gp(X, full_cov=True)
+        assert_equal(loc, y)
+        assert_equal(cov.abs().sum().item(), 0)
+
+
+@pytest.mark.parametrize("model, X, y, kernel, likelihood", TEST_CASES, ids=TEST_IDS)
+def test_inference(model, X, y, kernel, likelihood):
+    if "Sparse" in model.__name__:
```
eb8680: nit: use `isinstance` instead

pyro/distributions/sparse_multivariate_normal.py
```diff
@@ -104,6 +105,9 @@ def _compute_logdet_and_mahalanobis(self, D, W, y, trace_term=0):
             W_Dinv_y = W_Dinv.matmul(y)
         elif y.dim() == 2:
             W_Dinv_y = W_Dinv.matmul(y.t())
+        else:
+            raise ValueError("This distribution does not support computing log_prob "
```
eb8680: nit: `"This distribution..."` -> `"SparseMultivariateNormal..."`

tests/contrib/gp/test_models.py
```diff
@@ -1,75 +1,100 @@
...
+    assert loc0.size()[1:] == y.size()[1:]  # test latent shape
+    assert cov0.dim() == 2
+    assert cov0.size(0) == cov0.size(1)
+    assert cov0.size(0) == Xnew.size(0)
+    assert_equal(loc0, loc1)
+    assert_equal(cov0.diag(), var1)
+
+    # test trivial forward
+    # for variational models, inferences depend on variational parameters, so skip
+    if "Variational" in model.__name__:
```
eb8680: nit: use `isinstance` instead

tests/contrib/gp/test_models.py
fehiepsi: @eb8680 model is indeed model_class, so I use `model_class is Sparse...` instead.

Issue #836, opened 2018-03-01
Bayesian Optimization


ngoodman:
We should make a module on top of pyro for bayesian optimization. (This could initially live in contrib.)

Now that we have working GPs it should be pretty straightforward to implement standard BO. We could also do BO based on adaptive basis function regression as the function distribution.

@ysaatchi @fehiepsi (etal) please edit to add details as you see fit!

fehiepsi:
Yeah, this is an interesting direction. I would like to work on this (might be two weeks later) after implementing Duo Averaging for HMC. It would be great if someone takes an initial step. :)

eb8680:
There are quite a few GP features that we should add first - more documentation/examples/tutorials, batching, NUTS with dual averaging for hyperparameters, a simpler interface for learning, etc.  I'd recommend focusing on those at least until after the Pyro 0.2.0 release.

ysaatchi:
This is an awesome idea and definitely lives on top of pyro as @ngoodman said! 
One challenge is optimizing acquisition functions with derivatives. This requires computing derivatives w.r.t. GP inputs. @fehiepsi would this be straightforward? 
I also agree with @eb8680 RE priorities. I think we would also like some additional likelihoods before Pyro 0.2.0, especially for classification.

tobyclh:
Hello, I am here to share about what I learn so far trying to do BO with pyro :)
regarding derivatives w.r.t. to GP inputs, theoretically derivatives of GP can be written as they do [here](http://mlg.eng.cam.ac.uk/mchutchon/DifferentiatingGPs.pdf), but I haven't tested it myself.
And for the acquisition functions, folks in [Robo](https://github.com/automl/RoBO/tree/master/robo/acquisition_functions) did a great job implementing the most common ones so I assume that will be a good start.

fehiepsi:
@ysaatchi @tobyclh It seems to me that computing derivatives is quite straightforward (just make the input variables requires_grad when we need to compute). Let me some time to catch up with literature before I can make any claim.

For more features of GP, I made a tracking list at #832. Will start to implement them after duo averaging stuff.

tobyclh:
I have reviewed some of frameworks for API design idea, including
[AutoML's RoBO](https://github.com/automl/RoBO) (a top level BO object with input string to specify acq func. to use)
[GPFlowOPT](https://github.com/GPflow/GPflowOpt) (separate acq, bo object
[...]

PR #883, opened 2018-03-13
[Ready] Add tutorial for Bayesian Optimization

fehiepsi:
After reading related literature, I think that GP module is ready for Bayesian Optimization. An issue like gradient w.r.t. input, as mentioned in #836, is solved quite naturally within Pytorch/Pyro framework. So this week, I would like to make a tutorial for BO. Hope that this tutorial will help other contributors in designing API for BO using Pyro. Plans are:
- [x] Add ULB acquisition
- [x] Step by step solves a simple objective using GPR model with ULB.
- [x] ~~Replicate an experiment in the popular [paper](https://arxiv.org/pdf/1206.2944.pdf) which uses BO to find a minimum of Branin-Hoo function.~~ This part can be added in a future tutorial which @tobyclh is working on.
- [x] Fix an issue of getting `nan` while finding minimizer of acquisition function. It seems that this is an issue of LBFGS. I might change to use Adam to find the minimizer.
- [x] Add more texts / references.

cc @ngoodman @ysaatchi @tobyclh


tobyclh:
Thank you for putting it up, I was about to do the same actually :)
here is the init code for BO
Let me know how I can help!

fehiepsi:
@tobyclh Great!!! Would you mind making a pull request when you think it is ready (even in some preliminary form)? :)

My intention for making this tutorial is to just use current Pyro components. When API for BO is ready, it would be great if we have a tutorial to introduce it.

tobyclh:
@fehiepsi 
No problem. Currently I am having issue with kernel getting negative and fail to decompose, once I resolve it I will clean up and make a PR.

fehiepsi:
I was able to make it work but I can't explain why RBF kernel works better than Matern in this case. So I thought that it is better to let this tutorial as a reference for other contributors on BO, rather than for users. If Pyro's devs want to merge this, I will add more texts to explain other things (for example: explain the implementation of LCB acquisition).

fehiepsi:
I think that the tutorial is in a good shape now. Is there anyone would like to review? :D

eb8680:
@fehiepsi I'll review after you fix merge conflicts, looks like this and #882 got a little out of sync with dev

fehiepsi:
@eb8680 this pull request is only different to #882 at `bo.ipynb`, `gp/util.py` (I put LowerConfidenceBound class there) and toc for `bo`.

eb8680:
@tobyclh feel free to add your thoughts on the tutorial as well

pyro/contrib/gp/util.py
```diff
@@ -128,3 +130,82 @@ def _register_param(self, param, mode="model"):
             p = pyro.sample(param_name, dist.Delta(MAP_param))
 
         self._registered_params[param] = p
+
+
+class LowerConfidenceBound(nn.Module):
```
eb8680: This doesn't need to be a class, you can just split it into stateless functions and move it into the tutorial.

tobyclh: Agree, I think it makes more sense to simply provide a collection of functions to do the optimization. Rather than embedding the optimizer into the class.
In practices people have quite different optimization preference like random start, DIRECT etc.

fehiepsi: I see. The reason I made it a class was to have a universal Acquisition class (with all helper methods) which only requires implementation of `forward` pass for derived classes. However, I agree that it is not necessary for this single tutorial, especially when we decide to put its implementation into the notebook. About @tobyclh points about "in practices, ...", it is one of my most worrisome things with designing API for BO. I am willing to see news with your work! ;)

fehiepsi:
Good points @eb8680 ! I should have made these plots condensed into a single one. Will do it and move LowerConfidenceBound into notebook later today. I am now stuck at dealing with multi-class likelihood. >"<

pyro/contrib/gp/util.py
ysaatchi: A better way to do this is (I think) is to call this class something like "OptimizingGP" or "BOGP" which takes an acquisition function as an argument. I do like the fact that it is a class though. It should be.

pyro/contrib/gp/util.py
```diff
@@ -128,3 +130,82 @@ def _register_param(self, param, mode="model"):
...
+            # last minimum point will be an init point for first minimum candidate,
+            # other minimum candidates will get uniform random initialization
+            unconstrained_x0 = transform_to(self.constraint).inv(self._get_init_point(random=(i != 0)))
+            unconstrained_x = torch.tensor(unconstrained_x0, requires_grad=True)
+            minimizer = optim.Adam([unconstrained_x], lr=1)
+
+            for j in range(self.num_steps):
+                minimizer.zero_grad()
+                x = transform_to(self.constraint)(unconstrained_x)
+                y = self(x)
```
ysaatchi: this call should be to the acquistion function which is a member of this class (see comment above)

pyro/contrib/gp/util.py
```diff
@@ -128,3 +130,82 @@ def _register_param(self, param, mode="model"):
...
+        Gets minimum points of current acquisition posterior.
+        """
+        candidates = []
+        values = []
+        for i in range(self.num_restarts):
+            # last minimum point will be an init point for first minimum candidate,
+            # other minimum candidates will get uniform random initialization
+            unconstrained_x0 = transform_to(self.constraint).inv(self._get_init_point(random=(i != 0)))
+            unconstrained_x = torch.tensor(unconstrained_x0, requires_grad=True)
+            minimizer = optim.Adam([unconstrained_x], lr=1)
```
ysaatchi: I'd use LBFGS here, especially when you're not using stochastic VI (for stochastic VI may be nice to use Adam)

ysaatchi: Adam with lr=1, courageous :)

pyro/contrib/gp/util.py
```diff
@@ -128,3 +130,82 @@ def _register_param(self, param, mode="model"):
...
+                minimizer.step()
+            x = transform_to(self.constraint)(unconstrained_x)
+            y = self(x)
+            candidates.append(x.data)
+            values.append(y.data)
+
+        argmin = torch.min(torch.cat(values), dim=0)[1].item()
+        return candidates[argmin]
+
+    def forward(self, x):
```
ysaatchi: Yes, here have a switch statement for different acquisition functions

pyro/contrib/gp/util.py
```diff
@@ -128,3 +130,82 @@ def _register_param(self, param, mode="model"):
...
+    def minimum(self):
+        """
+        Gets minimum points of current acquisition posterior.
+        """
+        candidates = []
+        values = []
+        for i in range(self.num_restarts):
+            # last minimum point will be an init point for first minimum candidate,
+            # other minimum candidates will get uniform random initialization
+            unconstrained_x0 = transform_to(self.constraint).inv(self._get_init_point(random=(i != 0)))
```
ysaatchi: This solution will not scale well in higher dimensional BO, maybe allow users to pass in different initialization strategies?

pyro/contrib/gp/util.py
fehiepsi: I will try to use LBFGS again, it stuck at local minima. >"<

pyro/contrib/gp/util.py
fehiepsi: Indeed, I am going to move this class into the notebook and split it into several functions. :)

pyro/contrib/gp/util.py
fehiepsi: @ysaatchi This class was created for just this particular tutorial. For a BO module with various features, @tobyclh is working on it.

pyro/contrib/gp/util.py
eb8680: @fehiepsi @ysaatchi I think moving everything into functions in the tutorial is fine for now.  I'm writing a design doc for a Pyro Bayesian optimization package where we can discuss what the actual API should look like.  Unfortunately I'm away from work for 3 days starting tomorrow so I'll try to finish adding my thoughts and post it on #836 on Monday for others to comment/edit/extend.

pyro/contrib/gp/util.py
eb8680: @ysaatchi do you have a custom PyTorch implementation of LBFGS you're using that you could share as a gist?

pyro/contrib/gp/util.py
ysaatchi: got it, cool :)

fehiepsi:
@eb8680 I have moved LCB to the notebook, split it out into several functions, and used `autograd.grad`+`autograd.backward` (to modify `.grad` attribute of the input tensor).
@ysaatchi I used again the LBFGS with Matern32 kernel. It seems good.

@eb8680 @ysaatchi @tobyclh  If you have more comment on the tutorial, please let me know. I will try to address them. :)

tobyclh:
@fehiepsi 
As pointed out [here](https://discuss.pytorch.org/t/lbfgs-doesnt-seem-to-work-well/9195), the current implementation of L-BFGS in torch does not have line search implemented hurts its performance significantly, if you want to test it scipy's version is better. I'm don't think its a good idea to introduce extra dependency in tutorial, but a reminder to the reader about the issue should be nice.

fehiepsi:
@tobyclh Thanks for your comment! I think that all gradient-based methods have problems with local minima. Using L-BFGS, we don't have to specify `num_steps`, which is good to me. For performance, I can't justify which one (Adam or L-BFGS) is better. :( Mentioning that pytorch's L-BFGS is not effective might be not a good idea. We'd rather suggest ways to deal with it. In this tutorial, I used multi-started approach. In [this paper](https://www.cs.ox.ac.uk/people/nando.defreitas/publications/BayesOptLoop.pdf), the authors mentioned about other recent approaches such as SOO and BamSOO, but I am not familiar with them.

eb8680:
Sorry for the delay @fehiepsi - here's another round of feedback on the content, ordered by section title.  Overall, you should add more context, background and motivation; the tutorial should be as self-contained as possible.

## Bayesian optimization ##
- The problem still needs motivation.  Give three or four application areas or examples, preferably with references.
- The description of BO is still somewhat imprecise.  Instead of the paragraph you have, add three or four lines of pseudocode for the basic sequential BO algorithm and explain the details and variable names.
- Try to introduce BO without specifically referring to GPs, except in the last sentence, since the introduction doesn't depend on them.  Prefer phrasing like "we build a probabilistic model of f" and "the model will be updated"

## Define a problem ##
- Add a sentence of motivation for this toy problem: maybe because it has local minima?  Because it appears in other libraries or papers?

## Set a Gaussian Process prior ##
- Introduce GP regression very briefly
- Explain why GPs in particular are useful models for BO
- Why is choosing a kernel important? Why pick Matern32 for this problem?  Add a sentence or two of motivation.

## Define an acquisition function ##
- Add a sentence of motivation for this choice of acquisition function
- Give a simple, informal abstract definition of the LCB acquisition function, then the GP-specific mathematical definition

## Find `x_n` ##
- Mostly fine, but you could shorten it by removing the beginning and end frames of the middle figure and replacing them with the first figure and the last figure.

fehiepsi:
Hi @eb8680 , thanks for your review! I spent quite amount of time to deal with `latent_shape` in #888. Now, I only need to add tests and fix bugs for it (and wait for Pyro uses Pytorch master). So it is a good time to address your review. After taking a look at various functions at https://www.sfu.ca/~ssurjano/optimization.html, I think that [FORRESTER ET AL. (2008) FUNCTION](https://www.sfu.ca/~ssurjano/forretal08.html) might be a good choice for this tutorial (the current function was made by myself). GPyOpt devs also used Forrester function for their [introduction tutorial]. I will make other corresponding changes based on your review soon.

fehiepsi:
@eb8680 I have revised bo tutorial based on your great feedback. If there is any suggestion else, please let me know so I can improve the tutorial. :)

eb8680:
Hmm... I fixed a typo to try to get GitHub to show the correct diff after merging #955 but I'm still not seeing this branch diffed correctly against dev.

eb8680:
@fehiepsi this is ready to merge but something is wrong with the diff, can you open a new PR with this branch?

PR #885, opened 2018-03-13
Fix active_dims handling in pyro.contrib.gp

fritzo:
Simple fix to use `|` rather than `+` for set operations.

## Tested

- added a regression test
- fixed a nearby death test


fritzo:
cc @fehiepsi

fehiepsi:
LGTM. It was a bug and @fritzo already fixed it.

PR #886, opened 2018-03-14
Fix examples and tutorials to use the updated tensor API

neerajprad:
This is the last remaining update to move our tutorials/examples to the updated torch API. Once this and #884 are merged, I'll make any final (small) changes and update to PyTorch master.

@martinjankowiak, @jpchen, @karalets , @rohitsingh0812, @fritzo, @null-a, @fehiepsi  : I would highly recommend that you guys take a look at your corresponding tutorial, before the release. This refactor has been done with the aim of moving our codebase to use the latest PyTorch tensor API, but there are some other things we need to check: 
 - I am sure that with the updated syntax, there might be things that you can do in a more idiomatic fashion, e.g. by using torch scalars instead of constructing singleton tensors. 
 - This is doing a very basic check that the cells and examples are are runnable, but if you find that the tutorials are converging to a different answer or giving unexpected results (possible, though unlikely), we can debug that together.


fehiepsi:
@neerajprad Thanks for updating gp tutorial! I will make corresponding changes after this pull request merged.

neerajprad:
@fritzo - Can we get this merged?

neerajprad:
Thanks, @martinjankowiak!

PR #888, opened 2018-03-14
[Ready] Add more gp likelihoods

fehiepsi:
This pull request addresses part of #832 . Plans are:

- [x] Add Binary likelihood
- [x] Add a parameter `latent_shape` to models, correct the implementation with `latent_shape`
- [x] Add MultiClass likelihood
- [x] Add test

~~Blocked until pyro use Pytorch master and its MultivariateNormal distribution.~~


fehiepsi:
The implementation of MultivariateNormal in Pyro does not fully support for `batch`, which is required for MultiClass likelihood. Let's wait until https://github.com/pytorch/pytorch/pull/4950 get merged. I also need to modify variational models a bit to support this likelihood.

fehiepsi:
Happy to make it work. One important thing under the hood is: add a property `latent_shape` for gp models, which controls the shape of latent function: `f_shape = latent_shape x num_data`. I see that it is easier to control things if we assume output tensor has arbitrary dimensional.
+ Assume output tensor for Binary/Gaussian prediction is `N x D` and latent_shape is `1`, then `f_shape=1xN`, `y_shape=DxN`. Then in the likelihood, f_shape will be broadcast to y_shape.
+ Assume output tensor for Multiclass prediction is `N x D` and latent_shape is `num_classes`, then `f_shape=num_classes x N`, `y_shape = D x N`. In the likelihood, we convert `f_shape = N x num_classes` and broadcast to `y_shape x num_classes`.

With the introduction of this property, new likelihoods will be implemented easily: we just need to handle f_shape a bit to make `f` compatible with the corresponding likelihood.

For reviewers, I used `permute`, `reshape`, `view` at many places to handle `latent_shape`. Nothing changes with algorithms.

fehiepsi:
@fritzo Would you mind reviewing this pull request when you have time? This pull request heavily related to new batch shape semantics in Pyro. :)

pyro/contrib/gp/likelihoods/binary.py
```diff
@@ -0,0 +1,29 @@
...
+        By default, we use `sigmoid` function.
+    """
+    def __init__(self, response_function=None):
+        super(Binary, self).__init__()
+        self.response_function = (response_function if response_function is not None
+                                  else F.sigmoid)
+
+    def forward(self, f, y=None):
+        f_res = self.response_function(f)
+        if y is None:
```
fritzo: @fehiepsi I'm surprised that this cannot be simplified to take advantage of broadcasting in `torch.distributions.Distribution.log_prob()`:
```py
def forward(self, f, y=None):
    f_res = self.response_function(f)
    return pyro.sample("y", dist.Bernoulli(f_res), obs=y)
```
Were you seeing errors in this simplified version? If so maybe we can fix those instead of working around them here.

pyro/contrib/gp/likelihoods/gaussian.py
```diff
@@ -16,16 +16,17 @@ class Gaussian(Likelihood):
...
         self.set_constraint("variance", constraints.positive)
 
-    def forward(self, f, obs=None):
-        variance = self.get_param("variance").expand_as(f)
-        event_dims = f.dim()
-        return pyro.sample("y", dist.Normal(f, variance).reshape(extra_event_dims=event_dims),
-                           obs=obs)
+    def forward(self, f, y=None):
+        variance = self.get_param("variance")
+        if y is None:
```
fritzo: ditto: I'm surprised this cannot be simplified.

pyro/contrib/gp/likelihoods/multi_class.py
```diff
@@ -0,0 +1,45 @@
...
+            raise ValueError("Latent function output should be at least 2 dimensional: "
+                             "one for number of classes and one for number of data.")
+        # swap class dimension and data dimension
+        f_swap = f.transpose(-2, -1)  # -> num_data x num_classes
+        if f.size(-1) != self.num_classes:
+            raise ValueError("Number of Gaussian processes should be equal to the "
+                             "number of classes. Expected {} but got {}."
+                             .format(self.num_classes, f.size(-1)))
+        f_res = self.response_function(f_swap)
+        if y is None:
```
fritzo: ditto: I'm surprised this cannot be simplified to
```py
def forward(self, f, y=None):
   # ...error checking...
    f_res = self.response_function(f)  # <--- no swap needed?
    return pyro.sample("y", dist.Categorical(f_res), obs=y)
```

pyro/contrib/gp/models/model.py
```diff
@@ -8,20 +8,34 @@
...
 
         :param torch.Tensor X: A 1D or 2D tensor of input data for training.
-        :param torch.Tensor y: A 1D or 2D tensor of output data for training.
+        :param torch.Tensor y: A tensor of output data for training with
+            ``y.size(0)`` equals to number of data points.
         """
-        if X.dim() > 2 or y.dim() > 2:
-            raise ValueError("Input tensor and output tensor should be of 1 or 2 dimensionals.")
+        if X.dim() > 2:
+            raise ValueError("Input tensorshould be of 1 or 2 dimensionals.")
```
fritzo: It would be nice for users if we print the actual dim:
```py
raise ValueError("Expected input tensor of 1 or 2 dimensions, actual dim = {}".format(X.dim()))
```

pyro/contrib/gp/models/sgpr.py
```diff
@@ -83,14 +81,14 @@ def model(self):
...
 
-        # correct event_shape for y
-        y_t = self.y.t() if self.y.dim() == 2 else self.y
-        zero_loc = y_t.new([0]).expand(y_t.size())
         # DTC: cov = Qff + noise, trace_term = 0
         # FITC: cov = Qff + diag(Kff - Qff) + noise, trace_term = 0
         # VFE: cov = Qff + noise, trace_term = tr(Kff - Qff) / noise
-        pyro.sample("y", dist.SparseMultivariateNormal(zero_loc, D, W, trace_term).reshape(
-            extra_event_dims=zero_loc.dim() - 1), obs=y_t)
+        # convert y_shape from N x D to D x N
```
fritzo: I'm surprised this is needed, since we generally try to keep batch dimensions on the left. Can you explain a bit more what issues you were seeing? (Feel free to ping me on PyTorch slack, or we could talk over Zoom)

pyro/contrib/gp/models/svgp.py
```diff
@@ -78,25 +88,18 @@ def guide(self):
...
 
-        # TODO: use scale_tril=Lu
-        Ku = Lu.t().matmul(Lu) + self.jitter.expand(Lu.size(0)).diag()
-        # correct event_shape for mu
-        mu_t = mu.t() if mu.dim() == 2 else mu
-        pyro.sample("u", dist.MultivariateNormal(loc=mu_t, covariance_matrix=Ku).reshape(
-            extra_event_dims=mu_t.dim() - 1))
-        return kernel, likelihood, Xu, mu, Lu
+        pyro.sample("u", dist.MultivariateNormal(loc=mu, scale_tril=Lu)
+                    .reshape(extra_event_dims=mu.dim()-1))
```
fritzo: Is `mu` batched or is it multidimensional? If it is multidimensional then you've correctly used `extra_event_dims`, but if its batched, then it may be more appropriate to use `iarange`.

pyro/contrib/gp/likelihoods/multi_class.py
fehiepsi: Here I want to make the last dimension of `f` is num_classes (event of Categorical distribution).
So if f_shape = f_extra x num_classes x N, y_shape = y_extra x N (I swap y in models), its shape for Categorical distribution will be y_extra x N x num_classes (broadcasted from f_extra x N x num_classes).

pyro/contrib/gp/likelihoods/binary.py
fehiepsi: I got invalid log_prob shape, expected [], actually [2 , 3] for y_binary2D case.

pyro/contrib/gp/models/sgpr.py
fehiepsi: Here `N` is event_shape, and `D` is batch_shape (but people traditionally set N x D for y, I don't want to break that). This swap is to guarantee y_shape follows Pyro batch semantics: batch_shape x event_shape.

pyro/contrib/gp/models/svgp.py
fehiepsi: I think mu batched: `u_shape = D x N`, D is batch_shape, N is event_shape. Is there any difference if we use `iarange` here instead of `extra_event_dims`? Here D can be arbitrary shape: `torch.Size([])` or `torch.Size([3, 9, 7])`.

fehiepsi:
I want to close this pull request in favor of a new one: to avoid confusion about shape-related things. Really sorry for this! I will state all the changes clearly in that pull request.

PR #890, opened 2018-03-14
Upgrade to latest pytorch master

neerajprad:
PyTorch CI wheels are updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ pip install pyro-ppl[extras]  # for running examples/tutorials
 
 For recent features you can install Pyro from source.
 
-First install a recent PyTorch, currently PyTorch commit `8327982`.
+First install a recent PyTorch, currently PyTorch commit `e40425f`.
 ```sh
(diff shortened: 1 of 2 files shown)


neerajprad:
@fritzo - By upgrading to PyTorch master, the `torch.poisson` function only accepts `DoubleTensor` now.

```
>>> import torch
>>> torch.poisson(torch.tensor([1]))
RuntimeError: expected scalar type Double but found Float

>>> torch.poisson(torch.DoubleTensor([1]))
 2
[torch.DoubleTensor of size (1,)]
```

Is that expected, or some regression? cc. @apaszke

apaszke:
I don't think that is expected

neerajprad:
> I don't think that is expected

Should I create an issue in PyTorch to investigate further?

apaszke:
Yes, please!

neerajprad:
@martinjankowiak, @jpchen - Can we merge this? This will fix the currently broken build due to the change in CI wheels (issue mentioned above).

PR #891, opened 2018-03-14
Robustify dev install instructions in README

fritzo:
- Use http protocol rather than git protocol (which was failing for some internal users)
- Move commit hash down to script so we can copy-and-paste install instructions
- Provide easy suggestion for installing `PyTorch` which often works fine (w/o GPU)

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,18 +59,22 @@ pip install pyro-ppl[extras]  # for running examples/tutorials
 
 For recent features you can install Pyro from source.
 
-First install a recent PyTorch, currently PyTorch commit `e40425f`.
 ```sh
-git clone git@github.com:pytorch/pytorch
+git clone --recursive https://github.com/pytorch/pytorch
 cd pytorch
-git checkout [commit hash above]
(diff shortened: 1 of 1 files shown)


fritzo:
No need to let tests pass, since this is README only

README.md
```diff
@@ -59,18 +59,22 @@ pip install pyro-ppl[extras]  # for running examples/tutorials
...
 cd pytorch
-git checkout [commit hash above]
+git checkout e40425f  # <---- a well-tested commit
 ```
 Then build PyTorch following instructions in the PyTorch
 [README](https://github.com/pytorch/pytorch/blob/master/README.md).
+Depending on your system, this may be as simple as
+```sh
+pip install ninja
+python setup.py install
```
jpchen: for me the install build fails if i have `ninja`, i had to uninstall `ninja` for build install (vs build develop)

fritzo: @jpchen can you point me to your PR that fixes this by removing this line?

jpchen:
Yep I'll cut a PR I wasn't sure if it was just my machine if you tested it

PR #892, opened 2018-03-15
Implement half-Cauchy distribution

fritzo:
This resurrects @jpchen's earlier implementation of `HalfCauchy` from https://github.com/probtorch/pytorch/pull/97. We could conceivably move this upstream to PyTorch, but it is so simple to implement that maybe it better belongs in Pyro.

I've also removed duplicate doc entries from the `automodule:: pyro.distributions.torch`, since those are auto-generated by Pyro code rather than sphinx.

## Tested

- added a fixture to distribution tests
- generated docs and confirmed they look correct


pyro/distributions/half_cauchy.py
```diff
@@ -0,0 +1,48 @@
...
+
+    :param torch.Tensor loc: lower bound of the distribution.
+    :param torch.Tensor scale: half width at half maximum.
+    """
+    params = Cauchy.params
+
+    def __init__(self, loc, scale):
+        loc, scale = broadcast_all(loc, scale)
+        base_dist = Cauchy(0, scale)
+        transforms = [AbsTransform(), AffineTransform(loc, 0)]
```
jpchen: ` AffineTransform(loc, 0)` -> `AffineTransform(loc, 1)`

tests/distributions/conftest.py
```diff
@@ -200,6 +200,17 @@
...
             ],
             scipy_arg_fn=lambda mu, gamma: ((), {"loc": np.array(mu), "scale": np.array(gamma)})),
+    Fixture(pyro_dist=dist.HalfCauchy,
+            scipy_dist=sp.halfcauchy,
+            examples=[
+                {'loc': [0.5], 'scale': [1.2],
+                 'test_data': [1.0]},
+                {'loc': [0.5, 0.5], 'scale': [1.2, 1.2],
+                 'test_data': [[1.0, 1.0], [1.0, 1.0]]},
+                {'loc': [[0.5], [0.3]], 'scale': [[1.2], [1.0]],
```
jpchen: can you add a test `loc < 0`?

fritzo: Done

fehiepsi:
Yeah, waiting to use this! It is so useful for models which we know apriori that the parameters are positive.

pyro/distributions/half_cauchy.py
```diff
@@ -0,0 +1,48 @@
...
+    def loc(self):
+        return self.transforms[1].loc
+
+    @property
+    def scale(self):
+        return self.base_dist.scale
+
+    @constraints.dependent_property
+    def support(self):
+        return constraints.greater_than(self.loc)
```
fehiepsi: I just wonder if we set support `greater_than(0)` then is `0` still a valid value for this distribution?

jpchen: yeah it technically should be "greater than or equal to" though im not clear it makes a difference for continuous distributions

fritzo:
Thanks, JP!

Issue #847, opened 2018-03-04
Refactor and extend ELBO implementations


fritzo:
This issue documents refactoring and enhancement of `Trace_ELBO` and `TraceGraph_ELBO` over a series of PRs. See also the related #816 and the motivating #846 and #828 .

## Why?

It has become clear while diagnosing #846 that `enum_discrete` requires stricter dependency structure than `Trace_ELBO` or `TraceGraph_ELBO` require, namely that variables in an `iarange` cannot have downstream dependents outside of that `iarange`. This motivates splitting out a new class `TraceEnum_ELBO` from the simpler `Trace_ELBO`, so that `Trace_ELBO` can be simpler, more flexible, more auditable as a reference implementation, and easier to for users to fork when defining their own `ELBO` implementations.

Furthermore, the `TensorTree` data structure we're using to implement `TraceEnum_ELBO` in #828 can later be used in `Trace_ELBO` to provide full support for batching (i.e. for Rao-Blackwellization based on coarse structure of `irange` and `iarange`, rather than fine dependency structure as used in `TraceGraph_ELBO`). This motivates an overhaul of `Trace_ELBO` before 0.2 release.

Note that `Trace_ELBO` has lost support for minibatch Rao-Blackwellization during our refactoring to support multiple batch dimensions. This functionality will be added back in before 0.2 release, and is already supported in #828.

## Steps

- [x] #848 Refactor to split `Trace_ELBO` into `TraceEnum_ELBO` + `Trace_ELBO`.
- [x] #828 Fix `TraceEnum_ELBO` to work correctly with `iarange`.
- [x] #863 Allow mixed reuse of `iarange` contexts (e.g. `context1 = pyro.iarange(...)`). This is needed to fix subsample scale bugs in `iarange`.
- [x] #872 Allow mixed reuse of `irange` iterators.
- [x] #865 Add nested `iarange` support to `Trace_ELBO` (using `MultiFrameTensor`), e.g. to Rao-Blackwellize over minibatches.
- [x] #867 Fix broadcasting in `TraceGraph_ELBO` (using `MultiFrameTensor`).
- [x] #894 Fix broadcasting in `TraceEnum_ELBO` (using `MultiFrameDice`).
- [x] #902 Clean up `iter_discrete_traces` after its implementation settles down.
- [x] #902 Add more tests
- [x] #903 Expose `Trace*_ELBO` in sphinx documentation.
[...]

PR #894, opened 2018-03-16
Fix enumerate parallel via DiCE

### LINKED ISSUE #846: Errors when combining iarange + discrete enumeration
This issue tracks progress of combining enumeration #742 and nested `iarange` #780. Currently `TraceEnum_ELBO` gives incorrect results when enumerating any `iarange`, even in SS-VAE (as discovered by @rohitsingh0812).

fritzo:
Fixes #846 :tada:
Addresses #847

- Simplifies `TraceEnum_ELBO` to use the [DiCE](https://arxiv.org/abs/1802.05098) gradient estimator.
- Completely fixes parallel + sequential + MCMC enumeration logic (in all combinations).

Joint work with @eb8680 who wrote the [first version](https://github.com/uber/pyro/blob/dice-elbo/pyro/infer/dice_elbo.py) of `Dice_ELBO` from which much of this is copied.


eb8680:
Nice!  FYI my less-tested version with extra Rao-Blackwellisation is [here](https://github.com/uber/pyro/blob/dice-elbo/pyro/infer/dice_elbo.py).

pyro/infer/util.py
```diff
@@ -124,68 +125,71 @@ def __repr__(self):
...
-    Data structure to compute cumulative costs along paths in a tree.
-    Typically keys are ``cond_indep_stack``s.
+    An implementation of the DiCE operator compatible with Pyro features.
+
+    This implementation correctly handles:
+    - scaled log-probability due to subsampling
+    - independence in different contexts due to iarange
+    - weights due to parallel and sequential enumeration
+
+    This assumes restricted dependency structure on the model and guide:
```
martinjankowiak: "This assumes restricted dependency..."

are we going to add checks for this?

pyro/infer/util.py
```diff
@@ -124,68 +125,71 @@ def __repr__(self):
...
-    def copy(self):
-        result = TreeSum()
-        result._terms = self._terms.copy()
-        result._upstream = self._upstream.copy()
-        result._frozen = self._frozen
-        return result
+        for site in guide_trace.nodes.values():
+            if site["type"] != "sample":
+                continue
+            log_prob = site['score_parts'].score_function  # not scaled by subsampling
```
martinjankowiak: can we add a test with enumeration with rejection_gamma?

pyro/infer/util.py
fritzo: We can add some partial checks before release, but some cases are impossible to detect. E.g. in
```py
x_axis = iarange('x')
y_axis = iarange('y')
with x_axis:
    x = pyro.sample('x', ...)
with y_axis:
    y = pyro.sample('y', ...)
with x_axis, y_axis:
    xy = pyro.sample("xy", ...)
```
it is impossible to determine whether `y` depends on `x`. (we could examine the autograd graph, but the user could get around this via `x.detach()`).

martinjankowiak:
another question i have is the following;

let's say i only wanted it to be once differentiable. would it be simple to support that case and if so would it lead to a a performance gain? seems like the answer should be yes on both points?

fritzo:
> let's say i only wanted it to be once differentiable

I think you could wrap it in a `torch.autograd.Function` with a `@once_differentiable` `.backward()`.

Actually I think we could be more efficient by wrapping the entire `(log_prob - log_prob.detach()).exp()` op in a `torch.autograd.Function` that avoids computing `log_prob`, but that seems relatively low priority.

fritzo:
> i'm unclear if rsvi will work correctly

I believe RSVI will work correctly, but will fail to take advantage of sticking-the-landing.

martinjankowiak:
wouldn't it be sufficient to add a detach in one or two places if ```only_once_differentiable=True```?

fritzo:
> let's say i only wanted it to be once differentiable
> wouldn't it be sufficient to add a detach

Actually the way we're using this in `loss_and_grads` should already be nearly as cheap as a once-differentiable version, since we're not passing `create_graph=True` to `.backward()`. In that sense, we're only using it in a once-differentiable way.

PR #895, opened 2018-03-16
fix dev build instructions

jpchen:
#891

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,17 +59,20 @@ pip install pyro-ppl[extras]  # for running examples/tutorials
 
 For recent features you can install Pyro from source.
 
+First, build PyTorch following instructions in the PyTorch
+[README](https://github.com/pytorch/pytorch/blob/master/README.md).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,18 +59,21 @@ pip install pyro-ppl[extras]  # for running examples/tutorials
...
-[README](https://github.com/pytorch/pytorch/blob/master/README.md).
-Depending on your system, this may be as simple as
+On Linux:
 ```sh
-pip install ninja
 python setup.py install
 ```
+On OSX:
+```sh
+MACOSX_DEPLOYMENT_TARGET=10.9 CC=clang CXX=clang++ python setup.py install
```
neerajprad: Thanks for the fix! I somehow missed this, and assumed a linux install. :)

neerajprad: You might still need `ninja` though for python 3.6 install (if I remember correctly).

PR #898, opened 2018-03-16
Add pytorch installation script for mac and linux

### LINKED ISSUE #896: Make PyTorch builds available for common use cases
Now that we have a CloudFront distribution backed by S3, we should make it easier for Pyro consumers to work off of Pyro `dev`:
 - [ ] make CPU binaries for Mac and Linux installable via a script.
 - [ ] Migrate to using the same infra for CI.

neerajprad:
Fixes #896.
 - Includes a PyTorch installation script that can be used to download and install the CPU wheels for PyTorch version compatible with Pyro `dev`. This should make local Pyro development easy since users on `dev` will not have to build PyTorch from source.
 - Uses the same script in `.travis.yml` to install the linux wheels for CI testing.

**Testing:** Ran script locally for osx wheels. Travis will check the wheels for Linux.

### DIFF
--- a/scripts/install_pytorch.sh
+++ b/scripts/install_pytorch.sh
@@ -17,7 +17,7 @@ PYTORCH_BUILD_COMMIT=e40425f
 
 
 # Detect OS and Python version numbers.
-OS=$(uname -s)
+OS=$(uname -s | cut -d " " -f1)
 PYTHON_VERSION=$(python -c 'import sys; version=sys.version_info[:3]; print("{0}{1}".format(*version))')
 
 # Lookup wheel names to download
--- a/scripts/install_pytorch.sh
+++ b/scripts/install_pytorch.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+
+set -xe
+
+
+TMP_DIR=tmp
+
+function _cleanup() {
+    [[ -d ${TMP_DIR} ]] && rm -rf ${TMP_DIR}
+}
+
+trap _cleanup EXIT
+
+# Adjust as per the version used in CI.
+PYTORCH_VERSION=0.4.0a0
+PYTORCH_BUILD_COMMIT=e40425f
+
+
+# Detect OS and Python version numbers.
+OS=$(uname -s)
+PYTHON_VERSION=$(python -c 'import sys; version=sys.version_info[:3]; print("{0}{1}".format(*version)')
+
+# Lookup wheel names to download
+WHL_VERSION=${PYTORCH_VERSION}%2B${PYTORCH_BUILD_COMMIT}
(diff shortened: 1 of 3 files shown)


scripts/install_pytorch.sh
```diff
@@ -0,0 +1,52 @@
...
+    WHL_PREFIX=${PYTORCH_LINUX_PREFIX}
+    WHL_LOOKUP=PYTORCH_LINUX_PY_${PYTHON_VERSION}_WHL
+else
+    echo "OS - ${OS} is not supported by the install script."
+    exit 1
+fi
+
+# Download wheel and install
+wget -P tmp/ "${WHL_PREFIX}/${!WHL_LOOKUP}.whl"
+pip install tmp/*
```
jpchen: `rm -rf tmp` after

scripts/install_pytorch.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
```
jpchen: will this work on non-POSIX terminals ? as opposed to `#!/usr/bin/sh`?

scripts/install_pytorch.sh
neerajprad: That will be handled the exit trap `_cleanup` (see top of script).

scripts/install_pytorch.sh
```diff
@@ -0,0 +1,52 @@
...
+
+function _cleanup() {
+    [[ -d ${TMP_DIR} ]] && rm -rf ${TMP_DIR}
+}
+
+trap _cleanup EXIT
+
+# Adjust as per the version used in CI.
+PYTORCH_VERSION=0.4.0a0
+PYTORCH_BUILD_COMMIT=e40425f
```
fritzo: Do you think we're likely to get these commits out-of-sync by forgetting to update this file? Or are more likely to shoot ourselves in the foot being too clever via
```sh
PYTORCH_BUILD_COMMIT=$(grep 'git checkout .* a well-tested commit' README.md | cut -f3 -d' ')
```

neerajprad: I like that. :) And if anyone changes the README in a way that breaks the script, we 'll catch it in CI.

neerajprad:
> Do I understand correctly that this should "just work" on both Linux and OS X?

Yes, this should just work but only for the CPU versions. Fro the GPU stuff, it's probably best to use the docker scripts.

scripts/install_pytorch.sh
neerajprad: This invokes the `bash` shell explicitly (so it is not exactly POSIX compliant, or it might be but I haven't tested that), but shouldn't be an issue with either mac or linux which is what it caters to.

PR #902, opened 2018-03-17
Simplify TraceEnum_ELBO and add more tests

fritzo:
Addresses #847 

- Refactors `iter_discrete_traces()` to use `poutine.queue()` again. This appears to cost on 3% overhead on tests.
- Adds an RSVI test of `TraceEnum_ELBO` as requested in #894 . This also simplifies the test to allow use of `@pytest.mark.parametrize`.

### DIFF
--- a/tests/infer/test_inference.py
+++ b/tests/infer/test_inference.py
@@ -218,59 +218,51 @@ class PoissonGammaTests(TestCase):
 
 
 @pytest.mark.stage("integration", "integration_batch_1")
-class ExponentialGammaTests(TestCase):
-    def setUp(self):
-        # exponential-gamma model
-        # gamma prior hyperparameter
-        self.alpha0 = torch.tensor(1.0)
-        # gamma prior hyperparameter
-        self.beta0 = torch.tensor(1.0)
-        self.n_data = 2
-        self.data = torch.tensor([3.0, 2.0])  # two observations
-        self.alpha_n = self.alpha0 + torch.tensor(self.n_data)  # posterior alpha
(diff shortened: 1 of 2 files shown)


tests/infer/test_inference.py
```diff
@@ -218,59 +218,51 @@ def guide():
...
+    def guide():
+        alpha_q_log = pyro.param("alpha_q_log", torch.tensor(log_alpha_n.data + 0.17, requires_grad=True))
+        beta_q_log = pyro.param("beta_q_log", torch.tensor(log_beta_n.data - 0.143, requires_grad=True))
+        alpha_q, beta_q = torch.exp(alpha_q_log), torch.exp(beta_q_log)
+        pyro.sample("lambda_latent", gamma_dist(alpha_q, beta_q))
+
+    adam = optim.Adam({"lr": .0003, "betas": (0.97, 0.999)})
+    svi = SVI(model, guide, adam, loss="ELBO",
+              trace_graph=(elbo_impl == "TraceGraph"),
+              enum_discrete=(elbo_impl == "TraceEnum"),
```
fritzo: Note that there's nothing discrete here; we're simply testing that DiCE correctly handles RSVI's score_parts.

martinjankowiak: is it not worth having a KL gradient test that mixes rsvi with discretes?

fritzo: Sure, I'll add some more test with rsvi + discrete enumeration

eb8680: Do you want this merged now or after you add the tests?

PR #903, opened 2018-03-19
simple automatic guide generation for advi

### LINKED ISSUE #897: Implement Gaussian ADVI
We'd like to be able to automatically create a guide from a model via something like
```py
def advi_multivariate(model, *args, **kwargs):
    """Assumes model structure and dimension are fixed."""
    prototype_trace = poutine.trace(mode).get_trace(*args ,**kwargs)
    dim = sum(site["value"].view(-1).size(0)
              for site in prototype_trace.nodes.values()
              if site["type"] == "sample" and not site["is_observed"])

    def sample_advi_latent():
        loc = pyro.param("_advi_loc", torch.zeros(dim))
        cov = pyro.param("_advi_cov", torch.zeros(dim, dim))
        return pyro.sample("_advi_latent", dist.MultivariateNormal(loc, cov))

    def guide(*args, **kwargs):
        latent = sample_advi_latent()
        pos = 0
        for name, site in prototype_trace.nodes.values():
            shape = site["fn"].shape()
            size = product(shape)
            unconstrained_value = latent[pos, pos + size].view(shape)
            pos += size
            value = transform_to(site["fn"].support)(unconstrained_value)
            pyro.sample(name, dist.Delta(value))

    def wrapped_model(*args, **kwargs):
        with poutine.scale(0):  # <--- avoid counting log p(latent)
            sample_advi_latent()
        return model(*args, **kwargs)

    return wrapped_model, guide
```

martinjankowiak:
this builds on fritz's initial sketch in #897.

this simple prototype actually runs and seems to "do the right thing" at least on the end-to-end conjugate gaussian chain test that is included. the multivariate normal uses a cholesky parameterization or a diagonal covariance (both options are available).

this is actually really simple because of poutine magic. the main improvement left for the future is better initialization.

pair coded with @fritzo 

closes #897


pyro/infer/advi.py
```diff
@@ -0,0 +1,118 @@
...
+    Base class for implementations of ADVI. Each derived class implements its own
+    :meth:`sample_latent` method. Assumes model structure and latent dimension are fixed.
+    """
+    def __init__(self, model):
+        self.prototype_trace = None
+        self.base_model = model
+
+    def setup_prototype(self, *args, **kwargs):
+        # run the model so we can inspect its structure
+        self.prototype_trace = poutine.block(poutine.trace(self.base_model).get_trace)(*args, **kwargs)
```
martinjankowiak: @eb8680 i think you'll appreciate this use of poutine.block

fritzo:
Could you also add a documentation section to https://github.com/uber/pyro/blob/dev/docs/source/pyro.infer.txt, then `make docs` and ensure that your well-written docstrings appear in the generated docs?

pyro/infer/advi.py
```diff
@@ -0,0 +1,138 @@
...
+                pyro.sample(name, dist.Delta(value))
+        assert pos == len(latent)
+
+    def model(self, *args, **kwargs):
+        """
+        A wrapped model with the same ``*args, **kwargs`` as the base ``model``.
+        """
+        # wrap sample statement with a 0.0 poutine.scale to zero out unwanted score
+        with poutine.scale("advi_scope", 0.0):
+            self.sample_latent(*args, **kwargs)
```
eb8680: Since `self.model` and `self.guide` don't share any state, it seems like the right solution here would be to allow auxiliary guide variables in SVI?  That way `ADVI` could just return a guide.

That can be left for a potential followup PR, though.

fritzo:
@martinjankowiak Summary of changes I've pushed:
- Reshape the guide sites via `dist.Delta(...).reshape(extra_event_dims=site["fn"].event_dim)`
- Add `.support` properties to `ReshapedDistribution` and `MaskedDistribution`
- Added docstrings to a couple methods

tests/integration_tests/test_advi.py
```diff
@@ -0,0 +1,78 @@
...
+            self.target_advi_mus[n] += self.target_mus[n].item()
+            self.target_advi_mus[n] += self.target_kappas[n].item() * self.target_advi_mus[n + 1]
+            self.target_advi_diag_cov[n] += 1.0 / self.lambda_posts[n].item()
+            self.target_advi_diag_cov[n] += (self.target_kappas[n].item() ** 2) * self.target_advi_diag_cov[n + 1]
+
+    def test_multivariatate_normal_advi(self):
+        self.do_test_advi(3, reparameterized=True, n_steps=10001)
+
+    def do_test_advi(self, N, reparameterized, n_steps):
+        print("\nGoing to do ADVIGaussianChain test...")
```
fritzo: nit: in tests we try to use `logger.debug(...)` rather than `print(...)`

fritzo:
> @eb8680 Maybe we should add some way for users to mark a model as static

Alternatively we could use a `poutine.trace` to ensure that the `base_model` results in a trace with same site names and shapes as the `prototype_trace`.

pyro/infer/advi.py
```diff
@@ -81,7 +82,10 @@ def model(self, *args, **kwargs):
         with poutine.scale("advi_scope", 0.0):
             self.sample_latent(*args, **kwargs)
         # actual model sample statements shouldn't be zeroed out
-        return self.base_model(*args, **kwargs)
+        base_trace = poutine.trace(self.base_model).get_trace(*args, **kwargs)
```
martinjankowiak: if you're gonna do this might you not as well block all the other sites?

martinjankowiak: shouldn't it error if there exist any _subsample sites?

tests/integration_tests/test_advi.py
```diff
@@ -0,0 +1,109 @@
...
+
+        assert_equal(pyro.param("advi_loc"), self.target_advi_mus[1:], prec=0.05,
+                     msg="advi mean off")
+        assert_equal(diag_cov, self.target_advi_diag_cov[1:], prec=0.07,
+                     msg="advi covariance off")
+
+
+@pytest.mark.parametrize('advi_class', [ADVIDiagonalNormal, ADVIMultivariateNormal])
+def test_diagonal_gaussians(advi_class):
+    n_steps = 3501 if advi_class == ADVIDiagonalNormal else 6001
```
fritzo: tip: another way you could write this is
```py
@pytest.mark.parametrize('advi_class,n_steps', [
    (ADVIDiagonalNormal, 3501),
    (ADVIMultivariateNormal, 6001),
])
def test_diagonal_gaussians(advi_class, n_steps):
    ...
```
which scales a bit better as we add new ADVI classes. no need to change though

pyro/infer/advi.py
fritzo: > if you're gonna do this might you not as well block all the other sites?

What other sites? `base_trace` should contain exactly the same sites as the `prototype_trace`.

> shouldn't it error if there exist any _subsample sites?

Why is that? I have pruned the subsample sites exactly as you did when constructing the `prototype_trace`. Indeed when I failed to prune those sites, the tests failed.

martinjankowiak: well doesn't a subsample hint at being outside of the workable model class?

fritzo: Can you clarify whether you believe this check to be too lax or too strict? I'm having a hard time understanding what you're requesting.

martinjankowiak: i'm fine with the check (assuming it will later be gated).

i'm just suggesting that certain kinds of subsampling (e.g. subsampling local random variables) will break this. maybe a check of some sort is possible. maybe there is no need for a check and we should just add some more textual warnings to the docstring

fritzo: Thanks, that makes sense. I think we can even simply check once in the `prototype_trace`. I'm not sure exactly what the check should be, so maybe we should chat and add it to a follow-up PR.

For example, I'm not sure whether it's ok to subsample observations, since we might allow the `latent` to depend on `*args, **kwargs` through which a subsample might be passed.

martinjankowiak:
yeah let's discuss tomorrow i think this PR is fine to be merged as is

2018-03-20 20:58 GMT-07:00 Fritz Obermeyer <notifications@github.com>:

fehiepsi:
I guess MAP inference can be a straight follow-up of this pull request? :D

beyondpie:
Hi, I cannot find the ADVI code in the latest dev branch. Do I miss something?

fritzo:
Hi @beyondpie, this now lives in [pyro.infer.autoguide](https://docs.pyro.ai/en/stable/infer.autoguide.html)

PR #906, opened 2018-03-19
Add check for TraceEnum_ELBO requirements

fritzo:
This adds a conservative warning to `TraceEnum_ELBO` in cases where the model or guide might contain dependencies on enumerated variables. The warning is only triggered in cases where the model or guide could easily be rewritten to be obviously correct.

### DIFF
--- a/pyro/util.py
+++ b/pyro/util.py
@@ -396,7 +396,7 @@ def check_traceenum_requirements(model_trace, guide_trace):
         for name, site in trace.nodes.items():
             if site["type"] != "sample":
                 continue
-            iranges[name] = tuple(f for f in  site["cond_indep_stack"] if not f.vectorized)
+            iranges[name] = tuple(f for f in site["cond_indep_stack"] if not f.vectorized)
             context = frozenset(f for f in site["cond_indep_stack"] if f.vectorized)
             for restricted_context, names in restricted_contexts.items():
 
--- a/pyro/util.py
+++ b/pyro/util.py
@@ -387,32 +387,54 @@ def check_site_shape(site, max_iarange_nesting):
     # TODO Check parallel dimensions on the left of max_iarange_nesting.
 
 
+def _are_independent(counters1, counters2):
+    for name, counter1 in counters1.items():
+        if name in counters2:
+            if counters2[name] != counter1:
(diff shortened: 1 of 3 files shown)


pyro/util.py
```diff
@@ -386,6 +387,56 @@ def check_site_shape(site, max_iarange_nesting):
...
+    enumerated_sites = set(name for name, site in guide_trace.nodes.items()
+                           if site["type"] == "sample" and site["infer"].get("enumerate"))
+    for role, trace in [('model', model_trace), ('guide', guide_trace)]:
+        irange_counters = {}
+        enumerated_contexts = defaultdict(set)
+        for name, site in trace.nodes.items():
+            if site["type"] != "sample":
+                continue
+            irange_counter = {f.name: f.counter for f in site["cond_indep_stack"] if not f.vectorized}
+            context = frozenset(f for f in site["cond_indep_stack"] if f.vectorized)
```
fritzo: We only care about vectorized frames, since only vectorized variables are enumerated in lock-step.

pyro/util.py
```diff
@@ -386,6 +387,56 @@ def check_site_shape(site, max_iarange_nesting):
...
+
+def check_traceenum_requirements(model_trace, guide_trace):
+    """
+    Warn if user could easily rewrite the model or guide in a way that would
+    clearly avoid invalid dependencies on enumerated variables.
+
+    :class:`~pyro.infer.traceenum_elbo.TraceEnum_ELBO` enumerates over
+    synchronized products rather than full cartesian products. Therefore models
+    must ensure that no variable outside of an iarange depends on an enumerated
+    variable inside that iarange. Since full dependency checking is impossible,
```
martinjankowiak: ... is impossible without ____ ....

fritzo: full dependency checking is impossible without *an oracle to the halting problem*

fritzo:
Any additional comments, or is this ready to merge?

martinjankowiak:
yup, just waiting on a commitment from you to explain to any confused users why their model generated this warning : )

fritzo:
yes, i hereby commit to responding to users about why warnings were generated.

PR #907, opened 2018-03-19
Demonstrate how to monitor gradients in gmm tutorial

fritzo:
This adds a tiny bit of code to monitor gradient norms during convergence in the GMM tutorial. The relevant logging snippet is:
```py
gradient_norms = defaultdict(list)
inference.loss(model, guide, data)  # Initializes param store.
for name, value in pyro.get_param_store().named_parameters():
    value.register_hook(lambda g, name=name: gradient_norms[name].append(g.norm().item()))
```
after which we can plot `gradient_norms`.

This is in response to @yhalk's question https://github.com/uber/pyro/issues/583#issuecomment-374185289


jpchen:
> ```for name, value in pyro.get_param_store().named_parameters():
    value.register_hook(lambda g, name=name: gradient_norms[name].append(g.norm().item()))```

ive been using variations of this snippet a lot in my code to inspect gradients. would this be useful to expose in pyro?

fritzo:
> @jpchen would this be useful to expose in pyro?

Do you have a suggestion for a nice api? Rather than build Pyro functionality around `.register_hook()`, it might be better to simply demonstrate that functionality in various tutorials. Maybe you could add a different way to use `.register_hook()` in your Bayesian regression tutorial?

jpchen:
> Maybe you could add a different way to use `.register_hook()`

i can do that, i just find myself using the line to register hook over a list of variables a lot, so itd be helpful to wrap that in some sugar to reuse. like
```
from pyro.debug import inspect_grad

inspect_grad(...) # '...' = list or dict or variable
```
or a debugger flag you can turn on that would register hooks for all params in the paramstore. is your concern the clunkiness of such helpers?

fritzo:
> is your concern the clunkiness of such helpers?

yeah i'm just concerned with keeping pyro "Minimal"

PR #908, opened 2018-03-20
Implement constraints in pyro.param statements

### LINKED ISSUE #749: Support constrain_to=___ argument in pyro.param()
The usual way to do constrained optimization in Pyro is to transform unconstrained parameters, e.g.
```py
log_variance = pyro.param("log_variance", Variable(torch.zeros(10), requires_grad=True))
variance = log_variance.exp()
```
PyTorch 0.4 introduces a generic way to do this using `transform_to()`:
```py
from torch.distributions import constraints, transform_to

variance = transform_to(constraints.positive)(pyro.param("unconstrained_variance", ...))
```
It would be nice to use `transform_to()` to provide more first-class support for constraints in `pyro.param`, something like
```py
# possible future syntax:
variance = pyro.param("variance", Variable(...), constrain_to=constraints.positive)
```

See also discussion in #702

fritzo:
Closes #749 

This adds new syntax
```py
sigma = pyro.param('sigma', torch.tensor(1.0),  # <--- no need to requires_grad=True

assert pyro.param('sigma') > 0  # <--- no need to specify constraint
```
Note that constraints are not yet compatible with param store load/save operations. This is blocked by https://github.com/probtorch/pytorch/issues/138, after which load/save should work fine.

## Tested

- [x] Convert integration tests to use `constraints.positive`
- [x] Convert some tests to use `constraints.unit_interval`
- [x] Measure performance overhead

## Benchmarks

| | before (sec) | after (sec) |
| --- | --- | ---
| test_exponential_gamma | 107.52 | 107.43 |
| NormalNormalTests | 32.06 | 31.54 | 

Pair coded with @eb8680.


pyro/params/param_store.py
```diff
@@ -223,19 +228,29 @@ def get_param(self, name, init_tensor=None, tags="default"):
...
             if callable(init_tensor):
-                self._params[name] = init_tensor()
-            else:
-                # from the memory passed in
-                self._params[name] = init_tensor
+                init_tensor = init_tensor()
+
+            # store the unconstrained value and constraint
+            with torch.no_grad():
+                unconstrained_param = transform_to(constraint).inv(init_tensor)
```
neerajprad: Is it worth skipping this for `constraints.real`? I suppose it should be fine unless there is some non-negligible cost to using the identity transform. It might be good to do a perf comparison (we can just use one of the tests that you modified, but we can do it separately).

eb8680: > It might also be good to do a perf comparison

Yeah, we should definitely add one, and there's a clear perf regression since the transform is applied every time `get_param` is called.  This could be avoided with a bit more bookkeeping, but we thought we'd address that in a followup PR.

fritzo: `transform_to(constraints.real).inv` simply returns its argument and is very cheap.

neerajprad:
Looks great, can we merge this?

pyro/params/param_store.py
```diff
@@ -51,11 +54,13 @@ def clear(self):
         self._active_params = set()
         self._param_tags = defaultdict(lambda: set())
         self._tag_params = defaultdict(lambda: set())
+        self._constraints = {}
 
     def named_parameters(self):
         """
         Returns an iterator over tuples of the form (name, parameter) for each parameter in the ParamStore
         """
+        # TODO consider returing constrained
```
jpchen: 'returning'

fritzo:
OK, ready to merge. I added some benchmarking of some tests and found no speed penalty.

PR #909, opened 2018-03-20
Delete tagging mechanism from parameter store

eb8680:
Another param store cleanup PR.  This feature wasn't being used much and was significantly bloating the parameter store.

Blocked by #908


tests/infer/test_inference.py
```diff
@@ -98,16 +98,16 @@ def setUp(self):
...
 
-    def do_test_fixedness(self, fixed_tags):
+    def do_test_fixedness(self, fixed_parts):
         pyro.clear_param_store()
 
         def model():
             alpha_p_log = pyro.param(
                 "alpha_p_log", torch.tensor(
-                    self.alpha_p_log_0.clone(), requires_grad=True), tags="model")
+                    self.alpha_p_log_0.clone()))
```
jpchen: does this need `requires_grad=True` for pyro params

martinjankowiak:
i don't have any objection to removing this but if it "wasn't being used" then the changes you made to air wouldn't have been required....

eb8680:
> i don't have any objection to removing this but if it "wasn't being used" then the changes you made to air wouldn't have been required....

OK, I've added a "much" to my description.  Does that help clear things up?

tests/infer/test_inference.py
```diff
@@ -89,7 +89,7 @@ def guide():
         assert_equal(0.0, log_sig_error, prec=0.05)
 
 
-@pytest.mark.skip("Reinstate once poisson is migrated to PyTorch - https://github.com/uber/pyro/issues/699")
+# @pytest.mark.skip("Reinstate once poisson is migrated to PyTorch - https://github.com/uber/pyro/issues/699")
```
fritzo: We should either delete this line or uncomment it.

tests/integration_tests/test_tracegraph_elbo.py
```diff
@@ -568,8 +568,8 @@ def guide():
                                             infer=dict(baseline=dict(baseline_value=baseline_value)))
                         assert z_i_k.shape == (4 - i,)
 
-        def per_param_callable(module_name, param_name, tags):
-            if 'baseline' in tags:
+        def per_param_callable(module_name, param_name):
+            if 'baseline' in param_name or 'baseline' in module_name:
```
fritzo: :+1: I think string matching is a more maintainable mechanism than a tags dict + code to access it. Especially as we refactor towards local param stores.

Issue #912, opened 2018-03-21
Test failure of enum-parallel gradients after PyTorch #5776


neerajprad:
With the latest pytorch master, many parallel enum tests in `test_enum` are failing due to mismatch in the gradient computation.

To replicate, checkout the commit `5fa3aac610ee234338dbc11eb5b6d4a133cb483d` in PyTorch master (https://github.com/pytorch/pytorch/pull/5776), build PyTorch and run these tests
```sh
pytest -v --tb=short tests/infer/test_enum.py
```
Example of a failing test - `test_elbo_iarange_iarange  2-2-None-None-parallel-None`.

@fritzo, @eb8680 - I thought that there could be some unexpected interactions between the dice elbo change and upstream PyTorch. Turns out that is not exactly the case as 11 of our tests fail even before the dice elbo change, but there are more failures (79) with dice elbo. Could you guys take a look? 

This could either be a Pyro bug or something in PyTorch upstream.

cpuhrsch:
Hey @neerajprad, thank you for your post, I'm looking into this now.

cpuhrsch:
I found the bug, I'll send a patch soon.

neerajprad:
Thanks, @cpuhrsch! Curious to see where the bug was.

cpuhrsch:
@neerajprad please see PR https://github.com/pytorch/pytorch/pull/5926

fritzo:
Fixed upstream by pytorch/pytorch#5926 and in Pyro by #917.

Issue #910, opened 2018-03-20
Allow configurable error-checking level


fritzo:
As of https://github.com/pytorch/pytorch/pull/5358 PyTorch supports globally configurable error checking of distribution parameters via `Distribution.set_default_validate_args(-)`. It would be helpful (e.g in diagnosing #875) if Pyro similarly had a globally configuration function like `pyro.set_validation_level(-)` to throttle model/guide sanity checks (`check_model_guide_match`, `check_site_shape`, etc.) in `Trace*_ELBO` (and possibly to also throttle `Distribution` checks, though @jpchen feels this is cumbersome).

It would also be nice to implement some checks with higher false-positive rate, e.g. detecting possible violation of `irange`/`iarange` variable dependency in `Trace*_ELBO` implementations (as suggested by @martinjankowiak). Ideally we not log these warnings by default, but only if the user requested extra validation logic.

Possible policies:
- allow configuration levels 0, 1, 2, 3, 4.
- allow configuration frequencies "never", "once", "always". (maybe it's fine to let users control this across iterations)

## Benchmarks

| validation_level = | 0 | 1 | 2 | 3 |
| -: | - | - | - | - |
| test_valid_models.py | ? | ? | ? | ? |
| test_enum.py | ? | ? | ? | ? |
| test_inference.py::test_exponential_gamma | ? | ? | ? | ? |

## Possible Tasks

- [x] update to PyTorch after https://github.com/pytorch/pytorch/pull/5358
- [x] add a global variable and setter for validation configuration
- [x] guard checks in `Trace*_ELBO` by validation level
- [x] add death tests to `test_valid_models.py` to test high validation level
- [ ] start using `Distributions.set_default_validate_args(True)` in some tests
- [x] turn on validation in all of our smoke-tested tutorials
- [ ] turn on validation in most tests (maybe only the first iteration of our expensive inference tests) - Not required yet.
- [x] informally benchmark a couple tests to determine overhead (see above table) - Done in #913

fritzo:
@jpchen @neerajprad Which do you think is cleanest?
1.  Separate validation toggles
    ```py
    torch.distributions.Distribution.set_default_validate_args(True)
    pyro.set_validate_model(True)
    ```
2.  A combined validation knob
    ```py
    pyro.set_validation_level(2)
    ```
3. Per-module validation toggles
    ```py
    torch.distributions.Distribution.set_default_validate_args(True)
    pyro.infer.set_validation(True)
[...]

PR #913, opened 2018-03-21
Introduce validation toggle flags for pyro.infer module

neerajprad:
Addresses some of the tasks from #910.
 - [x] Introduces a global and module level toggles for setting validation. Most of the checks in `param_store.py` are very light-weight assertions, and I did not see the need of gating them. I will introduce the distributions toggle once we upgrade to PyTorch master (blocked by #912). 
 - [x] Checks that the validation flags are working as expected through some tests in `test_valid_models.py`.
 - [x] Add `pyro.set_validation(True)` to tutorials that are smoke tested.
 - [x] Run profiling to see the impact of toggling validation in the infer module.


neerajprad:
Toggling on the checks in the infer module marginally increases the test times. I found no substantial different in `test_enum.py`. For `test_exponential_gamma`, it increases by ~4 seconds. The overall test times in `test_inference` did go down by 8-10 seconds from 5mins 30 secs to 5 mins 20 secs. 
 - I would still propose that we add this to gate more expensive checks in the infer module.
 - I don't think we need to add anything to `pyro.params` yet, but I'll add another flag for distributions module, since distributions arg checking can get expensive for large tensors.

pyro/util.py
```diff
@@ -346,6 +349,8 @@ def check_model_guide_match(model_trace, guide_trace):
 
 
 def check_site_shape(site, max_iarange_nesting):
+    if not pyro.infer.is_validation_enabled():
```
fritzo: I think it's cleaner to move the gating to outside of the checking logic. E.g `check_site_shape` is called inside loops in `Trace_ELBO` ([code](https://github.com/uber/pyro/blob/dev/pyro/infer/trace_elbo.py#L46)), and those loops can be entirely avoided:
```py
if not pyro.infer.is_validation_enabled():
    for site in ...:
       if ...:
            check_site_shape(site, ...)
    for site in ...:
       if ...:
            check_site_shape(site, ...)
```

pyro/infer/__init__.py
```diff
@@ -8,3 +8,14 @@
 from pyro.infer.svi import SVI
 
 # flake8: noqa
+
+_IS_VALIDATE = False
```
fritzo: nit: maybe call this `'_VALIDATION_ENABLED`?

pyro/__init__.py
```diff
@@ -420,3 +422,17 @@ def _fn():
...
         return lifted_fn(name, nn_copy, update_module_params=True, *args, **kwargs)
     return _fn
+
+
+def set_validation(is_validate):
+    pyro.infer.set_validation(is_validate)
+
+
+@contextmanager
+def validate(is_validate):
```
fritzo: nit: maybe call this `validation_enabled` and default to `True`, this reads nicely
```py
with validation_enabled():
    # do stuff
```

pyro/util.py
```diff
@@ -311,6 +312,8 @@ def check_model_guide_match(model_trace, guide_trace):
     also appears in the model, (3) at each sample site that appears in both
     the model and guide, the model and guide agree on sample shape.
     """
+    if not pyro.infer.is_validation_enabled():
```
fritzo: ditto: It seems cleaner to gate outside the `check_*()` functions

pyro/__init__.py
```diff
@@ -420,3 +422,17 @@ def _fn():
         # update_module_params must be True or the lifted module will not update local params
         return lifted_fn(name, nn_copy, update_module_params=True, *args, **kwargs)
     return _fn
+
+
+def set_validation(is_validate):
```
fritzo: nit: maybe call this `enable_validation()`?

fritzo:
@jpchen Are you ok with this additional complexity? The performance improvement is less than I thought, but this PR would give @martinjankowiak and I more freedom to implement stricter and more expensive checks (i.e. this PR would make it easier for users to ignore annoying false positives).

pyro/__init__.py
```diff
@@ -5,8 +5,10 @@
 import numbers
 import warnings
 from collections import OrderedDict
+from contextlib import contextmanager
 from inspect import isclass
 
+import pyro
 import torch
 
 import pyro.poutine as poutine
```
jpchen: nit: for consistency we should use the full namespace or not. `poutine.method()` vs `pyro.infer.method()` are both used in this file.

pyro/util.py
jpchen: if site shapes dont match, it will always result in an exception correct? if so, you need this check regardless of validation level

fritzo: I think the idea is to trade off computation for clarity of error messages: if `_VALIDATION_ENABLED` then perform possibly-expensive checks to ensure errors are caught as early as possible. Note that some shape errors might result in silent failure due to broadcasting.

neerajprad:
> this is fine. in the future we may want levels of validation other than true/false and maybe wrap up nan checks in there.

Agreed, we can start here and add more functionality as needed. I would caution against over-engineering though, and lets use these flags only when we are doing potentially expensive validation - like checking tensor values. NaN checks and simple asserts (like those in param_store) are really cheap and don't need this gating.

neerajprad:
> if site shapes dont match, it will always result in an exception correct? if so, you need this check regardless of validation level

In general, the gating is to prevent repetitive checks for models that are already validated. Not checking is no guarantee that your model will run fine. For those cases, this will just give a more readable error message. That said, tensor shape checks are really cheap, so we can remove this gating.

pyro/__init__.py
jpchen: no strong preference which is used, though we should agree on a convention

pyro/util.py
neerajprad: My only (minor) quibble with that is that we will need to remember to gate checks inside each of the `trace*_ELBO` classes any time we are using these checks, and for any subsequent checks that we add. e.g. (note the four checks here)

```
if infer.validation_enabled():
    check_model_guide_match(model_trace, guide_trace)
guide_trace = prune_subsample_sites(guide_trace)
model_trace = prune_subsample_sites(model_trace)
if infer.validation_enabled():
    check_traceenum_requirements(model_trace, guide_trace)

...
if infer.validation_enabled():
    check_site_shape(site, self.max_iarange_nesting)
...
if infer.validation_enabled():
    check_site_shape(site, self.max_iarange_nesting)
```

Centralizing this then seems more cleaner, at the cost of a little less control and some additional looping (which should be negligible cost for all practical purposes). I don't have a strong opinion on this though.

fritzo: I do feel strongly that `check_site_site()` should check the site's shape, rather than *maybe* check the site's shape, depending on some global state. E.g. I might want to call `check_site_shape()` in a debugger or in a custom inference algorithm, and in those use case I don't want to think about global state.

pyro/__init__.py
neerajprad: Yup, will change the names.

fritzo: i don't feel strongly about the names tho :smile:

pyro/util.py
neerajprad: > I do feel strongly that check_site_site() should check the site's shape, rather than maybe check the site's shape

That's a fair point.

neerajprad:
@fritzo, @jpchen - Let me know if this looks good to merge.

PR #914, opened 2018-03-22
address transform bug in advi

martinjankowiak:
fixes a bug @fritzo noticed in ADVI. adds a test that would have caught it.


fritzo:
@martinjankowiak may I merge once tests pass?

martinjankowiak:
fine by me

PR #916, opened 2018-03-22
Rename sparse_mvn and support batch for mvn variance

fehiepsi:
MultivariateNormal distribution is recently merged into Pytorch master. However, we are temporarily blocked by #912 . As soon as we start to use Pytorch master, I wish to use its MultivariateNormal distribution to test various changes of https://github.com/uber/pyro/pull/888. This pull request does two things:
+ Use Pytorch's MultivariateNormal
+ Rename filename of SparseMultivariateNormal to sparse_mvn.py


fritzo:
@fehiepsi this should now be unblocked

fehiepsi:
@fritzo So this pull request just renamed sparse_mvn and added batch calculation for `variance` property of mvn. :)

PR #917, opened 2018-03-22
Upgrading to PyTorch master

neerajprad:


fritzo:
It looks like the test failures are related to an import looking up the wrong `MultivariateNormal`. I think it's safe to simply remove the test since we'll be switching to upstream `MultivariateNormal` immediately in #916.

neerajprad:
> It looks like the test failures are related to an import looking up the wrong MultivariateNormal. I think it's safe to simply remove the test since we'll be switching to upstream MultivariateNormal immediately in #916.

I made changes to move to `torch.distributions.multivariate_normal`, and skipping this test. The issue is that the upstream pytorch implementation hasn't implemented `.variance` (or `.mean`), which causes this failure.

fritzo:
@neerajprad We could implement `.variance` and `.mean` in a wrapper in pyro/distributions/torch.py if those are needed elsewhere in Pyro. We can push those implementations upstream later.

neerajprad:
> @neerajprad We could implement .variance and .mean in a wrapper in pyro/distributions/torch.py if those are needed elsewhere in Pyro. We can push those implementations upstream later.

Done.

pyro/distributions/torch.py
```diff
@@ -67,6 +67,19 @@ def __init__(self, ps, n=1):
         super(Multinomial, self).__init__(n, probs=ps)
 
 
+class MultivariateNormal(torch.distributions.MultivariateNormal, TorchDistributionMixin):
+    def __init__(self, loc, cov):
```
fritzo: Can we omit the `__init__` method and replace `cov` with `covariance_matrix` (rather than replace `covariance_matrix` with `cov` as I see you've done in a few tests)?

fritzo: There are two reasons to omit the `__init__` method:
- this saves us a renaming later on
- we will have immedate access to the `scale_tril` arg, which we want in a few places

neerajprad: My bad..I thought that we are using `cov` in `torch.distributions`, but I was mistaken. Will revert.

fritzo: Yeah, I'm fixing the broken `MultivariateNormal` docs upstream right now :smile:

neerajprad: > we will have immedate access to the scale_tril arg, which we want in a few places

Makes sense. Removed.

PR #918, opened 2018-03-22
Increase failure threshold for perf testing in travis

neerajprad:
The travis environment is a bit noisy for running high fidelity regression tests (I have been seeing a couple of failures lately). This changes the regression percentage when `CI` is in the environment, so that we increase our tolerance when testing in travis.

### DIFF
--- a/scripts/perf_test.sh
+++ b/scripts/perf_test.sh
@@ -15,6 +15,12 @@ BENCHMARK_FILE=tests/perf/test_benchmark.py
 IS_BENCHMARK_FILE_IN_DEV=1
 REF_TMP_DIR=.tmp_test_dir
 
+if [ -z ${CI} ]; then
+    PERCENT_REGRESSION_FAILURE=10
+else
+    PERCENT_REGRESSION_FAILURE=20
+fi
+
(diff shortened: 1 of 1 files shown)


neerajprad:
> i assume most people will rely on travis to catch perf regression rather than testing locally, so just acknowledging the risk of increasing threshold

Travis is for catching the worst offenders, as any benchmarking on these VMs will have a higher variance than testing locally. The other thing to do would be to increase the test sizes or the number of replications (or both). 

Lets not merge this PR yet. Let me investigate a few possibilities first.

scripts/perf_test.sh
```diff
@@ -15,6 +15,12 @@ BENCHMARK_FILE=tests/perf/test_benchmark.py
 IS_BENCHMARK_FILE_IN_DEV=1
 REF_TMP_DIR=.tmp_test_dir
 
+if [ -z ${CI} ]; then
+    PERCENT_REGRESSION_FAILURE=10
+else
+    PERCENT_REGRESSION_FAILURE=20
```
fritzo: Consider adding a comment like "Allow looser comparisons on CI, since machines are more variable"

Issue #915, opened 2018-03-22
Dependent Enumeration


karalets:
We have been discussing with @fritzo about getting a version of enumeration working that can do dependent variables, such as in an HMM. It seems that the current enumeration routines are pretty close to what is needed and this should be doable.

Also, @ngoodman had an example he has been using for how he solved that in webppl. 
It would be nice to take inspiration from that functionality for pyro.
Here are example-programs where this is tackled that I would love to write down in pyro:
http://dippl.org/chapters/04-factorseq.html

Leaving this here as a reminder for interested people to discuss.

## Related work

- Laurence Aitchison (2018) "Tensor monte carlo" [pdf](http://www.gatsby.ucl.ac.uk/~balaji/udl-camera-ready/UDL-26.pdf)

## Tasks

- [x] #1279 Avoid expanding in `.enumerate_support()`
- [x] #1289 Support `expand=False` in SVI
- [x] #1301 Support `expand=False` in HMC
- [x] #1302 Support local `num_particles` in `TracEnum_ELBO`
- [x] #1313 Support message passing in [Dice](https://github.com/uber/pyro/blob/ed58ec6/pyro/infer/util.py#L210)
- [x] #1315 Support `iarange` + message passing in `Dice`
- [x] #1318 Memoize the opt_einsum contract path
- [x] #1328 Marginalize out discrete variables in the model
- [x] Work around `torch.einsum`/`torch.bmm` performance issues https://github.com/pytorch/pytorch/issues/10661
- [x] #1340 Set `expand=False` by default
- [x] #1322 Support message passing in [EnumTraceProbEvaluator](https://github.com/uber/pyro/blob/6aca9eb/pyro/infer/mcmc/util.py#L8)
- [x] #1343 Support model enumeration at different `iarange`s in SVI
- [x] Expose model enumerated random variables for prediction in SVI
    - [x] #1359 via local marginals
    - [x] #1367 via forward-sample backward-filter

## Triaged Tasks

- Allow guide enumeration for prediction in SVI
- Support parallelization in `Empirical` (required by `Search`)
- Support enumeration in `Search` (#1226)
- ([advi-expand](https://github.com/uber/pyro/compare/trace-enum-expand...advi-expand?expand=1)) Support `expand=False` in `AutoDiscreteParallel`

eb8680:
[...]

PR #919, opened 2018-03-23
Add unit tests with HMMs; weaken checks so enumeration works

fritzo:
Addresses #915 

This is a first step towards allowing Pyro to take advantage of Markov structure in enumeration. This PR adds the HMM example from #915 to tests/infer/test_enum.py and weakens `check_mode_guide_match()` to allow a weak form of site-shape-mismatch: sites may only match up to broadcasting, for dims governed by parallel enumeration (other dims must match exactly). The test passes but takes exponential time & space in the number of time steps. Eventually we'd like to take only linear time & space.

This relaxation in this PR is a step towards us being able to implement a junction tree algorithm that can efficiently compute `ELBO` on a factor graph of discrete variables, and to do this cheaply by avoiding broadcasting where unnecessary. Much more work remains.


fritzo:
After discussion with @martinjankowiak we've decided to add more tests to this PR.

fritzo:
Ok I've added one more HMM test, this time with markov structure in the guide. After a conversation with @eb8680 I am again convinced that `TraceEnum_ELBO` is correct. I think this is ready to merge as soon as tests pass.

PR #920, opened 2018-03-23
Update gmm tutorial to use PyTorch master

fritzo:
This converts one `softmax()` to a `constraints.simplex` now that an upstream gradient bug has been fixed in PyTorch https://github.com/pytorch/pytorch/pull/5842


fritzo:
> itd be nice to make these figures embedded images

I agree it's nice to have readable diffs, but embedded images don't show up on Github. I like that this tutorial can be read directly on Github right now and can be downloaded locally. I think the embedded image boilerplate makes it more difficult for users to fork-and-modify notebooks. Let's try to find a common practice before 0.2 launch.

PR #922, opened 2018-03-23
Controlling distribution argument validation via a global flag

### LINKED ISSUE #910: Allow configurable error-checking level

neerajprad:
Resolves #910, #928. 

Adds a flag for toggling distribution argument validation in `torch.distributions`, that can additionally be controlled via `pyro.set_validation(flag)`.
 - Sets the flag to true for all tests. Introduces a pytest marker `disable_validation` for expensive tests - currently only needed by HMC and NUTS which run model/guide a number of times. e.g. for HMC the test time for conjugate gaussian tests increased by 17 secs with args validation enabled.
 - small fixes to existing tests/code to ensure that we are passing correct parameters to torch distributions.
 - Ensures that compositions over `TorchDistribution` - e.g. ReshapedDistribution do not throw an exception when the global arg validation flag is set. This is accomplished by setting `arg_contraints={}` as the argument validation would have already been done on `base_dist`. The issue is that the `ReshapedDistribution` does not have `arg_constraints` attribute set and this throws an error when we do argument validation.


fritzo:
Looks like only the gp tutorial needs to be fixed.

neerajprad:
@fehiepsi - I see that in `svgp.py`, the `Kfdiag` term can be negative for certain indices, and this throws an error when we run pyro with distributions argument validation enabled (negative variance). I will currently remove argument validation for `gp.ipynb` so as to get this PR merged, but let us try to address this issue and re-enable validation.

fehiepsi:
@neerajprad Mathematically, Kfdiag should be positive, even when we add jitter to Kuu term (formula 2.24 GPML [book](http://www.gaussianprocess.org/gpml/chapters/RW2.pdf)). It seems that the problem is related to computational precision issues. I will see if other GP libraries deal with this problem. The worst scenario is we have to clip Kfdiag by small positive number (jitter).

fritzo:
It looks like tests are failing due to a bug/weirdness in the definition of `TransformedDistribution.params`. There is already an upstream PR under review that fixes this bug: https://github.com/pytorch/pytorch/pull/5989

fehiepsi:
I increased `jitter` from 1e-6 to 1e-5/1e-4 on that example, it passed all the check. So I can confirm that this is a precision issue. Should w add a check for `Kfdiag.min() > 0` and throw a warning to tell users to "try to increase jitter setting", or we handle it by ourselves (by clipping Kfdiag)?

neerajprad:
Thanks for looking into it, @fehiepsi. I have increased the value of `jitter` to `1e-5`, and that seems to pass the tests locally at least.

fritzo:
@neerajprad I'll try to clean up in #928 this afternoon, but feel free to take the issue if you're in a hurry. I'll self-assign it once I start.

tests/integration_tests/test_tracegraph_elbo.py
```diff
@@ -396,6 +396,7 @@ def guide():
         assert_equal(0.0, mu_error, prec=0.05)
         assert_equal(0.0, tau_error, prec=0.05)
 
+    @pytest.mark.xfail(reason="awaiting https://github.com/pytorch/pytorch/pull/5989")
```
fritzo: This merged this morning :smile: see #928

tests/distributions/test_transformed_distribution.py
```diff
@@ -51,7 +52,9 @@ def AffineExp(affine_b, affine_a):
 
 
 def get_transformed_dist(distribution, affine_a, affine_b):
-    return TransformedDistribution(distribution, [AffineExp(affine_b, affine_a)])
+    # Set `validate_args=False` as base distribution would have been validated, if needed.
```
fritzo: I'd really like to avoid this workaround after #928 fixes the bug in `TransformedDistribution`.

neerajprad: Makes sense. Let us merge the fixes from #928 first, and then update this PR.

neerajprad:
@fritzo - Updated the PR based on the new wheels. Ready to merge, unless there are more comments.

PR #923, opened 2018-03-24
Update bayesian regression/tutorial example for 0.2

jpchen:
mostly supporting the new batching semantics which broke the example. updated the tutorial and example.


examples/bayesian_regression.py
```diff
@@ -20,7 +20,8 @@
 def build_linear_dataset(N, p, noise_std=0.01):
     X = np.random.rand(N, p)
     # use random integer weights from [0, 7]
-    w = np.random.randint(8, size=p)
+    w = np.random.randint(4, size=p)
+    print('w = {}'.format(w))
```
fritzo: Was this debug statement accidentally committed?

examples/bayesian_regression.py
```diff
@@ -42,7 +43,7 @@ def forward(self, x):
 
 
 N = 100  # size of toy data
-p = 1  # number of features
+p = 2  # number of features
```
fritzo: :+1:

examples/bayesian_regression.py
```diff
@@ -51,10 +52,11 @@ def forward(self, x):
 def model(data):
     # Create unit normal priors over the parameters
     mu = torch.zeros(1, p).type_as(data)
-    sigma = torch.ones(1, p).type_as(data)
+    sigma = 2 * torch.ones(1, p).type_as(data)
```
fritzo: Would you mind modernizing all of this initialization code to avoid `.type_as()` (since `.type_as()` does not work on multi-gpu systems):
```py
mu = data.new_zeros(1, p)
sigma = 2 * data.new_ones(1, p)
bias_mu = data.new_zeros(1)
bias_sigma = 2 * data.new_ones(1)
```

examples/bayesian_regression.py
jpchen: since `w` is selected randomly, i thought it would be helpful to output the `w` selected, otherwise the user would have no way of comparing to the ground truth

examples/bayesian_regression.py
fritzo: I guess we can fix these in a later PR since many examples need to be fixed.

jpchen: i was actually in the process of doing that but the mvn pr broke my local env since i think it needs a newer pt version so i couldnt test cuda. will open another pr with this fix

PR #924, opened 2018-03-24
[blocked] Use new style of param transform, pre-support for MCMC with GP

fehiepsi:
This pull request uses the new style of param transform after https://github.com/uber/pyro/pull/908 (thanks @fritzo ). Other related changes are:
+ Clean guide for all models. Now, they are very simple!
+ Move variational parameter to model definition. This does not change the logic of inference but has benefit for MCMC inference. After getting parameters' samples from HMC (trained on `self.model`), we can simply call various `.fix_param()` methods and then, we can use the model to give predict on `Xnew`.
+ Remove implementation of batch_cholesky_transform in favor of https://github.com/pytorch/pytorch/pull/5980 

This pull request will be blocked until https://github.com/pytorch/pytorch/pull/5980 merged and pyro uses that pytorch commit. Changes will be clean after #888 merged.


fehiepsi:
Closed in favor of 933

PR #925, opened 2018-03-24
[blocked] support for deep GP

fehiepsi:
This pull request adds a support for creating deep GP models by letting `self.y` accept None value. I also clean methods `.model()` to make them clearer to read.

Together with #924, these pull requests do not modify algorithms. They just clean code, add comments, use better variable names, make various models have a uniform structure, and support for general cases (MCMC and deep GP). The `.forward()` pass will be cleaned in the next pull request together with rewriting docs for all GP module.

Blocked by #924.


fehiepsi:
closed in favor of #933

PR #926, opened 2018-03-25
Remove ng_ones(), ng_zeros() in favor of .new_zeros(), .new_ones()

fritzo:
PyTorch 0.4 now has more convenient helper functions:
- `torch.zeros()` and `torch.ones()` now return variables (since variables and tensors were merged)
- `x.new_zeros()` and `x.new_ones()` now returns a tensor with same type and *gpu placement* as `x`. This is now preferred because it supports multi-gpu systems. (note that type_as ignored gpu device placement and often created tensors on the wrong gpu in multi gpu systems)

### DIFF
--- a/examples/air/air.py
+++ b/examples/air/air.py
@@ -16,7 +16,7 @@ import pyro
 import pyro.distributions as dist
 import pyro.poutine as poutine
 from modules import Identity, Encoder, Decoder, MLP, Predict
-from pyro.util import ng_zeros, ng_ones, zeros
+from pyro.util import zeros
 
 
 # Default prior success probability for z_pres.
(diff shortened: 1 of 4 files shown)


fritzo:
> is `new_*` like `Variable` and `torch.tensor` where `requires_grad` defaults to false unless set to true

Yes. Even this new fancy syntax is supported: `torch.zeros(5,5, requires_grad=True)` :smile:

> is `tensor.news_zeros.cuda` equivalent to `tensor.cuda.new_zeros`? i recall `Variable`s were picky about this but not sure how these new `Tensors` are

I'm not sure. I do know that `x = x.cuda(); y = x.new()` is preferred over `y = x.new().cuda(); x = x.cuda()` because the former ensures x and y are on the same device.

PR #933, opened 2018-03-26
Revise GP codes

fehiepsi:
It seems to me that combining all the refactoring work into a pull request will make things easier to follow. This pull request combines #924, #925, #888, together with some new refactoring. I will make a summary for all of them here:
+ Add new likelihood functions and tests for: Binary and MultiClass. Other likelihoods (based on user requests) can be added in the same manner.
+ Regarding `latent_shape` which is introduced in #888 (and some opinions about it in #866), after having a discussion with @fritzo, I have decided to make things consistency by forcing GP output follows the Pyro batch semantics: `y_shape = y_batch_shape x num_data`. This way makes GP in Pyro different to GPflow, GPy but it fits everything about `y_shape` in my mind. In addition, codes are simplified a lot thanks to that (no need to permute f, y, scale_tril,...). For example, for a multiclass classification with 4 samples, if `y = [[2, 0, 1, 2], [1, 1, 0, 2]]` then `[2, 0, 1, 2]` is a output of 1 GP, `[1, 1, 0, 2]` is a output of another GP (or both of them are outputs of the same GP). In short, `X.shape[0]` must be equal to `y.shape[-1]`.
+ Use new `pyro.param(..., constraint)` syntax.
+ Allow `y` get None value. This is useful for creating deep GP models (which hidden GP output needs not to observe data).
+ Move variational parameters `u_loc`, `u_scale_tril` to model definition. This fits the syntax of other parameters in Reparameterize class and simplify variational `.guide`. In addition, it is helpful for MCMC. With MCMC trained on `self.model`, we don't need to work with these variational parameters but after we get samples of `u`, we can use `fix_param("u_loc", posterior["u"].mean())` (of course, we have to set `u_scale_tril = None`), then the `guide + forward` is ready to give predictions based on results from MCMC. There still needs to add some tests  or tutorials for this support in a future pull request.
+ Model / guide / forward are simplified a lot thanks to a new `_conditional(...)` method in GPModel. It is ported from the old `predict_f` but is more general (handle both `p(f | X, Xu, u_loc, u_cov)` and `p(f | X, Xu, u)`. I also added a detailed formula in comments together with reference to the [post](https://www.prowler.io/sparse-gps-approximate-the-posterior-not-the-model/) which is suggested by @eb8680 .
+ There are minor things such as: `size()` ->`shape`, `view()` -> `reshape()` to be more numpy friendly, `Model` -> `GPModel`, reorder `W_term`, `D_term` of SparseMVN distribution (because W is more important than D), add `jitter` for GPRegression

I think that GP API is in a good form for Pyro 0.2 now. I run on the notebook and saw that it generate sample figures. @eb8680 I will revise GP docs in the next pull request (to not distract reviewers of this pull request). After that, will work on tutorials.


fehiepsi:
@eb8680 I have added some tests for hmc and deepgp (to check if all the changes are really working, not to check their effectiveness). During that, I see that it is necessary to make several more changes (all are reasonable):
+ Give names for likelihoods, models (for deepGP, we combine many gps together, so we need a way to distinguish `pyro.sample("y", ...)` in different models).
+ Likelihood requires `f_loc`, `f_var` (instead of f) to estimate `E_p(f) [p(y | f)]`. This is helpful for Gaussian likelihood because we don't need to sample f, then sample y at separate steps (instead, just simply sample y with y_var = f_var + variance). On the other hand, if in the future, we support [Gauss-Hermite quadrature](https://en.wikipedia.org/wiki/Gauss%E2%80%93Hermite_quadrature) (which requires f_loc, f_var) to estimate that expectation, then it can be handled by default from Likelihood module. This is the approach of GPflow.
+ In case `self.y = None`, we return f_loc, f_var (in previous commits, I returned a sample for `y`, which is wrong because if we sample a latent "y", then we have to define a guide for it).
+ It seems that we can remove the confusing `extra_event_dims`. I have posted the report about the error I got in #936 . After things are investigated (cc @fritzo ), I believe that we can remove it.

Now, the API is quite similar to GPflow (I have tried to follow literatures instead of available codes but this convergence is unavoidable). We still lack three features: Gauss-Hermite quadrature for likelihood expectation, whitening transform (which aids optimization), and GPLVM. They will be supported after Pyro 0.2.

With the tests showed that things were working as expected, I will focus on docs and tutorials. A problem is that I have to prepare for my graduation from next week, so I don't have much time then. This might affect the plan which you suggested in #832. Anyway, I will try to complete docs, BO tutorial, and split current GP tutorial as soon as possible but might not have enough time to write new tutorials for kernels, classifications,... :(

pyro/contrib/gp/models/model.py
```diff
@@ -64,11 +82,97 @@ def forward(self, *args, **kwargs):
...
         Checks the correction of the shape of new data.
         """
-        if Xnew.dim() != X.dim():
+        if Xnew.dim() != self.X.dim():
             raise ValueError("Train data and test data should have the same number of dimensions.")
-        if Xnew.dim() == 2 and X.size(1) != Xnew.size(1):
+        if Xnew.dim() == 2 and self.X.shape[1] != Xnew.shape[1]:
             raise ValueError("Train data and test data should have the same feature sizes.")
+
+    def _conditional(self, Xnew, X, kernel, f_loc, f_scale_tril=None, Lff=None, full_cov=False):
```
fritzo: Could you add a unit test for `GPModel._conditional()` for both cases of `full_cov`? It's easier for us to review tests than to review math (and I trust unit tests more than I trust myself to catch errors in math :wink:)

pyro/contrib/gp/models/sgpr.py
```diff
@@ -29,77 +29,83 @@ class SparseGPRegression(Model):
...
             self.approx = approx
         else:
-            raise ValueError("The sparse approximation method should be one of 'DTC', "
-                             "'FITC', 'VFE'.")
-
-        self.jitter = self.X.new([jitter])
+            raise ValueError("The sparse approximation method should be one of "
+                             "'DTC', 'FITC', 'VFE'.")
 
     def model(self):
```
fritzo: Can you remind me where this is tested?

pyro/contrib/gp/models/sgpr.py
```diff
@@ -29,77 +29,83 @@ class SparseGPRegression(Model):
...
         self.set_mode("guide")
 
-        kernel = self.kernel
         noise = self.get_param("noise")
         Xu = self.get_param("Xu")
 
-        return kernel, noise, Xu
+        return self.kernel, noise, Xu
 
     def forward(self, Xnew, full_cov=False, noiseless=True):
```
fritzo: Can you remind me where this is tested?

tests/contrib/gp/test_likelihoods.py
```diff
@@ -0,0 +1,75 @@
...
+import torch
+
+from pyro.contrib.gp.kernels import RBF
+from pyro.contrib.gp.likelihoods import Binary, MultiClass
+from pyro.contrib.gp.models import SparseVariationalGP, VariationalGP
+
+T = namedtuple("TestGPLikelihood", ["model_class", "X", "y", "kernel", "likelihood"])
+
+X = torch.tensor([[1, 5, 3], [4, 3, 7], [3, 4, 6]])
+kernel = RBF(input_dim=3, variance=torch.tensor([1]), lengthscale=torch.tensor([3]))
```
fritzo: Can we make these scalars? `variance=torch.tensor(1.), lengthscale=torch.tensor(3.)`

tests/contrib/gp/test_likelihoods.py
```diff
@@ -0,0 +1,75 @@
...
+noise = torch.tensor([1e-6])
```
fritzo: Can this be a scalar `torch.tensor(1e-6)`?

tests/contrib/gp/test_likelihoods.py
```diff
@@ -0,0 +1,75 @@
...
+    if isinstance(likelihood, MultiClass):
+        latent_shape = y.shape[:-1] + (likelihood.num_classes,)
+    else:
+        latent_shape = y.shape[:-1]
+    if model_class is SparseVariationalGP:
+        gp = model_class(X, y, kernel, X, likelihood, latent_shape)
+    else:
+        gp = model_class(X, y, kernel, likelihood, latent_shape)
+
+    gp.optimize(num_steps=1)
```
fritzo: This appears to be a smoke test. Do you have any tests that would catch math errors?

tests/contrib/gp/test_likelihoods.py
fehiepsi: Yes, we can use scalars after HMC supports it :D

tests/contrib/gp/test_likelihoods.py
fehiepsi: Yes.

pyro/contrib/gp/models/model.py
fehiepsi: Yup

tests/contrib/gp/test_likelihoods.py
fehiepsi: In the past, it is a little complicated to test models. Now, things should be easier to test. I will do it soon.

tests/contrib/gp/test_models.py
```diff
@@ -60,7 +71,23 @@
...
+    else:
+        gp = model_class(X, None, kernel, likelihood)
+
+    loc, var = gp.model()
+    if model_class is VariationalGP or model_class is SparseVariationalGP:
+        assert_equal(loc.norm().item(), 0)
+        assert_equal(var, torch.ones(var.shape[-1]).expand(var.shape))
+    else:
+        assert_equal(loc.norm().item(), 0)
+        assert_equal(var, kernel(X).diag())
```
fehiepsi: Here are model tests. `loc`, `var` of variational model come from variational parameters.

tests/contrib/gp/test_models.py
```diff
@@ -71,29 +98,195 @@ def test_model_forward(model_class, X, y, kernel, likelihood):
...
+    assert_equal(loc_diff.norm().item(), 0)
+    cov_diff = cov - cov[..., :1, :1].expand(y.shape[:-1] + (10, 10))
+    assert_equal(cov_diff.norm().item(), 0)
+
+    # test noise kernel forward: kernel = WhiteNoise
+    gp.kernel = WhiteNoise(input_dim=3, variance=torch.tensor(10))
+    Xnew = torch.tensor([[2, 3, 1]])
+    loc, cov = gp(X, full_cov=True)
+    assert_equal(loc.norm().item(), 0)
+    assert_equal(cov, torch.eye(cov.shape[-1]).expand(cov.shape) * 10)
```
fehiepsi: Here are forward tests. There are three cases which are easy to calculate: Xnew = X, Xnew = constant, kernel = WhiteNoise.

tests/contrib/gp/test_models.py
```diff
@@ -71,29 +98,195 @@ def test_model_forward(model_class, X, y, kernel, likelihood):
...
+    gp.set_data(X, target_y)
+
+    gp.optimize(optim.Adam({"lr": 0.01}), num_steps=1000)
+
+    variance = gp.kernel.get_param("variance")
+    lengthscale = gp.kernel.get_param("lengthscale")
+    target_variance = kernel.get_param("variance")
+    target_lengthscale = kernel.get_param("lengthscale")
+    assert_equal(variance, target_variance, prec=0.2)
+    assert_equal(lengthscale, target_lengthscale, prec=0.2)
```
fehiepsi: Inference tests for regression models. Skip variational models.

tests/contrib/gp/test_models.py
```diff
@@ -71,29 +98,195 @@ def test_model_forward(model_class, X, y, kernel, likelihood):
...
+    Xu = torch.linspace(0, 5, 10)
+
+    sgpr = SparseGPRegression(X, y, kernel, Xu)
+    sgpr.optimize(optim.Adam({"lr": 0.01}), num_steps=1000)
+
+    Xnew = torch.linspace(0, 5, 100)
+    loc, var = sgpr(Xnew, full_cov=False)
+    target = 0.5 * torch.sin(3*Xnew)
+
+    assert_equal((loc - target).abs().mean().item(), 0, prec=0.05)
```
fehiepsi: Test inference for sgpr model. Try to learn the target function.

tests/contrib/gp/test_models.py
```diff
@@ -71,29 +98,195 @@ def test_model_forward(model_class, X, y, kernel, likelihood):
...
+            f_name = pyro.param_with_module_name(gp.name, "f")
+            post_trace["f"].append(trace.nodes[f_name]["value"])
+        if model_class is SparseVariationalGP:
+            u_name = pyro.param_with_module_name(gp.name, "u")
+            post_trace["u"].append(trace.nodes[u_name]["value"])
+
+    for param in post_trace:
+        param_mean = torch.mean(torch.stack(post_trace[param]), 0)
+        logger.info("Posterior mean - {}".format(param))
+        logger.info(param_mean)
```
fehiepsi: hmc test with different shapes of sites

tests/contrib/gp/test_conditional.py
```diff
@@ -0,0 +1,66 @@
...
+    loc1, var1 = conditional(Xnew, X, kernel, f_loc, f_scale_tril, full_cov=False)
+
+    if loc is not None:
+        assert_equal(loc0, loc)
+        assert_equal(loc1, loc)
+    n = cov0.shape[-1]
+    var0 = torch.stack([mat.diag() for mat in cov0.view(-1, n, n)]).reshape(cov0.shape[:-1])
+    assert_equal(var0, var1)
+    if cov is not None:
+        assert_equal(cov0, cov)
```
fehiepsi: conditional tests

pyro/contrib/gp/kernels/brownian.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, input_dim, variance=None, active_dims=None, name="Brownian"):
         super(Brownian, self).__init__(input_dim, active_dims, name)
 
         if variance is None:
-            variance = torch.ones(1)
+            variance = torch.tensor(1)
```
fritzo: We will need to make this a float tensor here and elsewhere `torch.tensor(1.)`. This is necessary after https://github.com/pytorch/pytorch/pull/5997. See #944 for our migration plans.

tests/contrib/gp/test_models.py
```diff
@@ -71,29 +98,195 @@ def test_model_forward(model_class, X, y, kernel, likelihood):
...
+
+
+@pytest.mark.disable_validation()
```
fritzo: Why does this disable validation?

tutorial/source/gp.ipynb
```diff
@@ -649,6 +649,7 @@
     }
    ],
    "source": [
+    "pyro.enable_validation(False)\n",
```
fritzo: Why do you disable validation here? Could you instead
```py
pyro.enable_validation(smoke_test)
```

tests/contrib/gp/test_models.py
```diff
@@ -71,29 +98,195 @@ def test_model_forward(model_class, X, y, kernel, likelihood):
...
+
+    Xnew = torch.linspace(0, 5, 100)
+    loc, var = svgp(Xnew, full_cov=False)
+    target = 0.5 * torch.sin(3*Xnew)
+
+    assert_equal((loc - target).abs().mean().item(), 0, prec=0.05)
+
+
+@pytest.mark.parametrize("model_class, X, y, kernel, likelihood", TEST_CASES, ids=TEST_IDS)
+@pytest.mark.disable_validation()
```
fritzo: Why do you disable validation in a smoke test?

tests/contrib/gp/test_models.py
```diff
@@ -71,29 +98,195 @@ def test_model_forward(model_class, X, y, kernel, likelihood):
...
+
+    Xnew = torch.linspace(0, 5, 100)
+    loc, var = sgpr(Xnew, full_cov=False)
+    target = 0.5 * torch.sin(3*Xnew)
+
+    assert_equal((loc - target).abs().mean().item(), 0, prec=0.05)
+
+
+@pytest.mark.init(rng_seed=0)
+@pytest.mark.disable_validation()
```
fritzo: Why do you disable validation in a smoke test?

tests/contrib/gp/test_models.py
```diff
@@ -71,29 +98,195 @@ def test_model_forward(model_class, X, y, kernel, likelihood):
...
+    variance = gp.kernel.get_param("variance")
+    lengthscale = gp.kernel.get_param("lengthscale")
+    target_variance = kernel.get_param("variance")
+    target_lengthscale = kernel.get_param("lengthscale")
+    assert_equal(variance, target_variance, prec=0.2)
+    assert_equal(lengthscale, target_lengthscale, prec=0.2)
+
+
+@pytest.mark.init(rng_seed=0)
+@pytest.mark.disable_validation()
```
fritzo: Can you run with validation for one step, then disable and continue for 1000 steps?

fritzo:
@fehiepsi To clarify, all Pyro code should be able to run with `pyro.enable_validation(True)`. All tests should run with validation on, except a few very expensive inference tests. For each expensive test that disables validation, there should be a corresponding cheap smoke test that runs the same model with validation enabled.

@neerajprad Should we remove the `disable_validation` decorator from our codebase and ensure that all tests are run with validation for at least one iteration?

tests/contrib/gp/test_models.py
fehiepsi: @fritzo Enable it will give me [error](https://gist.github.com/fehiepsi/00f7b2fdfcc7bf39ee25968bad40f6db) which I mentioned in https://github.com/uber/pyro/pull/936 . It seems that I have to use `pyro.iarange` for a bunch of independent dimensions? How can I handle its ndim automatically?
```
pyro.sample(dist.Normal(torch.zeros(3), torch.ones(3)), obs=torch.rand(9, 8, 7, 6, 3))
```
Another way is to use extra_event_dim, but you put many questions why so at #888. So I thought what I did was wrong. :(

fritzo: @fehiepsi You may have done it right in #888. We're still working out best practices. Let's enable validation and figure out whether you need to use `iarange` or `extra_event_dims`.

fritzo: > How can I handle its ndim automatically?

Good question. A manual answer to your example is:
```py
with iarange('a'), iarange('b'), iarange('c'), iarange('d'):
    pyro.sample(dist.Normal(torch.zeros(3), torch.ones(3)), obs=torch.rand(9, 8, 7, 6, 3))
```
I guess one answer as to how to do this automatically is to put the responsibility with the code that introduced each such independent dimension. Would this work for you, e.g. by putting `iarange`s in your tests that call GP code, and expecting that users of pyro.contrib.gp would use `iarange` when needed, outside of pyro.contrib.gp?

fehiepsi: Yes, let's temporarily use "extra_event_dims". I will mark TODOs at points needed to be revised (if we follow `iarange` approach).

fehiepsi:
@fritzo Thanks a lot for your review! I have removed disable validation marks and used `.reshape` again. If we decide to use `iarange`, then I think it can be obtained by requires `y.dim()` <= 2. This is reasonable and only requires users reshape their multidimensional outputs into 2D (1 for num_data, 1 for batch).

pyro/contrib/gp/likelihoods/binary.py
```diff
@@ -0,0 +1,32 @@
...
+                                  else F.sigmoid)
+
+    def forward(self, f_loc, f_var, y):
+        # calculates Monte Carlo estimate for E_q(f) [logp(y | f)]
+        f = dist.Normal(f_loc, f_var)()
+        f_res = self.response_function(f)
+        print(f_res.shape, y.shape)
+        return pyro.sample(self.y_name,
+                           dist.Bernoulli(f_res)
+                               .reshape(sample_shape=y.shape[:-f_res.dim()],
```
fritzo: Are you sure that `f_res.dim()` is always positive? If it is ever zero, then this will accidentally slice
```py
y.shape[:-0] = y.shape[:0] = ()
```
As a habit I use the safer positive indexing as follows:
```py
y.shape[:y.dim()-f_res.dim()]
```

fehiepsi: I will never get zero, at least `f_res.shape[-1] = num_data`.

Issue #931, opened 2018-03-26
Dev docs are broken on pyro.ai


fritzo:
Currently this page is blank http://docs.pyro.ai/en/dev/primitives.html
even though docs correctly build locally.

fritzo:
@jpchen Do you think this could be a pytorch wheel mismatch when docs are built?

jpchen:
the logs confirm your hunch, we need to upgrade to the new pytorch wheel. @neerajprad did you bake a new wheel we can use?

fritzo:
@neerajprad IIRC we use some special grepping to pull out the blessed commit from README.md when we build a docker or something. Can we use the same technique to automatically grab the latest wheel when generating docs on readthedocs (inside JP's doc generation script)?

neerajprad:
> Can we use the same technique to automatically grab the latest wheel when generating docs on readthedocs (inside JP's doc generation script)

Could you point me to the code that does this? Can we just run `bash scripts/install_pytorch.sh` to install pytorch (that would do the grepping)?

fritzo:
Thanks for fixing this, @jpchen

PR #934, opened 2018-03-27
use install script to install pytorch for docs

jpchen:
addresses #931

rtd cannot get new branches without a dev merge. so either we can merge this to test or merge another pr

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -189,10 +189,6 @@ def setup(app):
     app.connect("autodoc-skip-member", skip)
 """
 
-# @jpchen's hack to get rtd builder to install pytorch
+# @jpchen's hack to get rtd builder to install latest pytorch
 if 'READTHEDOCS' in os.environ:
-    os.system('pip install awscli')
-    os.system('aws s3 --no-sign-request sync s3://pyro-ppl/ci tmp --exclude "*" --include "*-cp27-*";')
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -189,10 +189,6 @@ def setup(app):
...
 
-# @jpchen's hack to get rtd builder to install pytorch
+# @jpchen's hack to get rtd builder to install latest pytorch
 if 'READTHEDOCS' in os.environ:
-    os.system('pip install awscli')
-    os.system('aws s3 --no-sign-request sync s3://pyro-ppl/ci tmp --exclude "*" --include "*-cp27-*";')
-    os.system('pip install tmp/*')
-    os.system('rm -r tmp')
-    os.system('pip install -e .')
+    os.system('sh scripts/install_pytorch.sh')
```
neerajprad: If this doesn't work, it could be because the installation script assumes bash and I'm not sure what `sh` points to here.

neerajprad:
So should we merge this and give it a try?

jpchen:
yeah I'll patch it if there's an issue. seems to be no other way

jpchen:
```
sh: 0: Can't open scripts/install_pytorch.sh
```
might be a permissions issue. now that the branch exists ill test and follow up

Issue #921, opened 2018-03-23
Including build metadata for Pyro versioning


neerajprad:
@fritzo suggested that it might be a good idea to include build metadata in Pyro version, as PyTorch does. This will easily help pin the Pyro version used, say in jupyter notebook, by running `pyro.__version__` at the top of the file. 

A reasonable format for this could be what we already have, plus some other information, that follows the [semantic versioning](https://semver.org/) specs. In particular, we could use the format:
`major.minor.patch-{tag}+{build-metadata}`.
 - `major.minor.patch` is the same as what we have right now and can just be specified as a string in the `pyro/__init__.py` file.
 - `tag` denotes optional pre-release version and may indicate that the compatibility requirements expected from `major.minor.patch` may not be satisfied (if its an alpha release). Could be `alpha`. PyTorch uses `a0`.
 - `build-metadata` - This has additional build specific metadata, and what we might want to include in our version info. It could just be the shortened sha commit id. Additionally, it could have the following information:
   - `.pytorch-{pytorch-sha-commit}` - this may be debatable, as `torch.__version__` should give us this info.
   - `.dirty` to indicate that the user has uncommitted changes. e.g. if a notebook has `.dirty`, it will indicate that you may not be able to replicate the results by simply using the sha commit id indicated by the version, as the user has modified pyro source code.

The full version string might look like - `0.1.2-a0+e581401.pytorch-a58f2d2.dirty`. Let me know your thoughts, preferences etc.?

neerajprad:
cc. @fritzo, @jpchen.

fritzo:
@jpchen What do you think? I think this would be useful for two reasons:
- Users could paste a useful `pyro.__version__` in github issues
- We could add a version header to the top of all internal notebooks (e.g. our pyro_model.ipynb):
    ```py
    import torch
    import pyro
    print(torch.__version__)
    print(pyro.__version__)
    ```

jpchen:
this is a great idea! ive personally found it cumbersome to check which pyro version im using since they all default to the latest release tag. for some of your specific suggestions:
> tag denotes optional pre-release version

why is this needed? cant all that information be inferred from the semver version? seems like these prerelease tags are informal and dont convey additional information

>`.pytorch-{pytorch-sha-commit}`

i would prefer not to have this since it overloads the version and like you said, can be elicited by running `torch.__version__`. i like something like:
```
> import pyro
> pyro.__version__
0.1.2+e581401[-dirty]
```
[...]

PR #935, opened 2018-03-27
Add semver compliant build metadata to Pyro version

neerajprad:
Addresses #921 (with a caveat. See below). 

After looking at how a few projects have implemented including the commit sha in their version info, I decided on the following approach (same as that of PyTorch):
 - Only the `version_prefix` needs to be changed anytime we are doing a public release.
 - When `setup.py` is called we read the `version_prefix`, append the current commit information and write to `_version.py`.
 - `__init__.py` reads `__version__` from the `_version.py`.

The caveat is that this will not be sufficient when Pyro is an editable install (i.e. installed via `-e / --editable`). In that case the source may have been modified beyond what is indicated by the commit sha in `_version.py`. For the case of an editable install, we will need a dynamic `__version__`, which first detects (via pip helper utilities) that `pyro-ppl` is an editable installation, and then runs the git incantations to get the current commit information. We will need to be a bit careful with this, but it can be included, if we need.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,11 +19,16 @@ for line in open(os.path.join(PROJECT_PATH, 'pyro', '__init__.py')):
         version = line.strip().split()[2][1:-1]
 
 # Append current commit sha to version
-commit_sha = subprocess.check_output(['git', 'rev-parse', '--short', 'HEAD'],
-                                     cwd=PROJECT_PATH).strip()
+commit_sha = ''
+try:
+    commit_sha = subprocess.check_output(['git', 'rev-parse', '--short', 'HEAD'],
+                                         cwd=PROJECT_PATH).decode('ascii').strip()
+except OSError:
+    pass
 
 # Write version to _version.py
-version += '+{}'.format(commit_sha)
+if commit_sha:
+    version += '+{}'.format(commit_sha)
 with open(os.path.join(PROJECT_PATH, 'pyro', '_version.py'), 'w') as f:
     f.write(VERSION.format(version))
 
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,31 @@
 from __future__ import absolute_import, division, print_function
 
+import subprocess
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -1,14 +1,32 @@
...
+
 # Find pyro version.
-for line in open('pyro/__init__.py'):
-    if line.startswith('__version__ = '):
+for line in open(os.path.join(PROJECT_PATH, 'pyro', '__init__.py')):
+    if line.startswith('version_prefix = '):
         version = line.strip().split()[2][1:-1]
 
+# Append current commit sha to version
+commit_sha = subprocess.check_output(['git', 'rev-parse', '--short', 'HEAD'],
```
fritzo: Do you think it's work putting this in a `try-except` in case some use case isn't using a git clone?
```py
try:
    comit_sha = subprocess.check_output(...).strip()
    version += '+{}'.format(commit_sha)
except OSError:
    pass
```

fritzo:
I'll wait for @jpchen's review. I realize this adds complexity, but I think it could help reduce confusion on forum.pyro.ai, on internal chat channels, and in possibly stale notebooks.

setup.py
neerajprad: Yes, absolutely.

neerajprad:
>  I don't see how to support pip install -e without calling git rev-parse each time pyro is loaded.

Yeah, we will definitely have to call `rev-parse` and use some pip helper utils to detect that the install is editable. Since pyro will be loaded only once though, it may not be noticeable. If we are willing to take on this additional complexity, I can make that change in this PR or a separate one.

fritzo:
@neerajprad May I merge this?

neerajprad:
Yes, this should be good to merge! Thanks for the review.

neerajprad:
I think we can hold off on the dynamic `__version__` for now.

Issue #827, opened 2018-02-27
Rename things before Pyro 0.2 release


fritzo:
Here is a list of candidate renamings, feel free to edit or add:

### confirmed:

- [x] #932 `.enumerable` -> `.has_enumerate_support`
- [x] #932 `.analytic_mean()` -> `.mean`
- [x] #932 `.analytic_var()` -> `.variance`
- [x] #940 `batch_log_pdf` -> `log_prob`;   `log_pdf` -> `log_prob_sum`
- [x] #947 drop `pyro.observe` in favor of `pyro.sample`
- [x] #961 distribution parameter names (see also #732)
- [x] #959 `.reparameterized` -> `.has_rsample`
- [x] #1000 `Poutine` -> `Handler` (assigned to @eb8680 )
- [x] #1016 `.reshape(sample_shape)` -> `.expand_by(sample_shape)`
- [x] #1016 `.reshape(extra_event_dims=n)` -> `.independent(n)`
  (Tensorflow calls this [Independent(-, reinterpreted_batch_ndims=n)](https://www.tensorflow.org/api_docs/python/tf/contrib/distributions/Independent))

### triaged:
- ~~`config_enumerate` -> `mark_sites_for_enumeration`, or maybe `config_infer`~~

eb8680:
Except for the `sample`/`rsample` distinction, we should rename all distribution methods and parameters to match `torch.distributions`.  We should also name all `log_pdf` things to `log_prob` in `Trace` and elsewhere, and `Poutine` to `Handler`.

PR #936, opened 2018-03-27
Tutorial on tensor shapes

### LINKED ISSUE #814: Add a tutorial on batch shaping in Pyro 0.2
Pyro 0.2 will have stricter batching requirements (after #806) to allow new features like parallel enumeration #742 and nested `iarange` #780. This will require more careful use of `pyro.iarange`, `TorchDistribution.reshape()` and `pyro.scope()` for masking. The tutorial should also cover how to write broadcastable PyTorch code, such as using `Ellipsis` as in `x[..., i, j]` and setting flexible `sample_shape` in sample statements.

fritzo:
Resolves #814
 
Do you find tensor shapes in Pyro 0.2 confusing?
Have you ever wondered what you could do with *two* nested `iarange`s?
Have you wanted to enumerate discrete latent variables, but been afraid to ask under what conditions Rao-Blackwellization is valid?

This PR is your venue to ask!
Don't wait. Comment now.

(to see a draft, click the `[View]` button just below `[Unified|Split]` on the [Files changed page](https://github.com/uber/pyro/pull/936/files))


fehiepsi:
@fritzo After reading your tutorial, I think that what you wondered is right: no need for `extra_event_dims` in GP models. So I tried to run the test code in a notebook: no error! I guess there is a setting in `tests` which makes things wrong. Gist file shows the error: https://gist.github.com/fehiepsi/00f7b2fdfcc7bf39ee25968bad40f6db

tutorial/source/tensor_shapes.ipynb
```diff
@@ -0,0 +1,292 @@
...
+    "x = d.sample()\n",
+    "assert x.shape == (3,)            # == batch_shape + event_shape\n",
+    "assert d.log_prob(x).shape == ()  # == batch_shape"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In Pyro you can treat a univariate distribution as multivariate by calling the `.reshape(extra_event_dims=_)` property."
```
jpchen: separate this and the next section from the previous one with a header like "Declaring dependence" or something

tutorial/source/tensor_shapes.ipynb
```diff
@@ -0,0 +1,292 @@
...
+    "def model():\n",
+    "    a = sample(\"a\", Normal(0, 1))\n",
+    "    b = sample(\"b\", Normal(zeros(2), 1).reshape(extra_event_dims=1))\n",
+    "    with iarange(\"c\", 2):\n",
+    "        c = sample(\"c\", Normal(zeros(2), 1))\n",
+    "    with iarange(\"d\", 3):\n",
+    "        d = sample(\"d\", Normal(zeros(3,4,5), 1).reshape(extra_event_dims=2))\n",
+    "    assert a.shape == ()\n",
+    "    assert b.shape == (2,)\n",
+    "    assert c.shape == (2,)\n",
```
jpchen: add some comments for what these shapes are and their independence along what dimensions

tutorial/source/tensor_shapes.ipynb
```diff
@@ -0,0 +1,292 @@
...
+    "            |<--->|\n",
+    "enumeration |batch| event\n",
+    "------------+-----+--------\n",
+    "            |. . .|        a = sample(\"a\", Normal(0, 1))\n",
+    "            |. . .|2       b = sample(\"b\", Normal(zeros(2), 1)\n",
+    "            |     |                        .reshape(extra_event_dims=1))\n",
+    "            |     |        with iarange(\"c\", 2):\n",
+    "            |. . 2|            c = sample(\"c\", Normal(zeros(2), 1))\n",
+    "            |     |        with iarange(\"d\", 3):\n",
+    "            |. . 3|4 5         d = sample(\"d\", Normal(zeros(3,4,5), 1)\n",
```
jpchen: 👍

tutorial/source/tensor_shapes.ipynb
```diff
@@ -0,0 +1,292 @@
...
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def model():\n",
+    "    a = sample(\"a\", Normal(0, 1))\n",
+    "    b = sample(\"b\", Normal(zeros(2), 1).reshape(extra_event_dims=1))\n",
+    "    with iarange(\"c\", 2):\n",
+    "        c = sample(\"c\", Normal(zeros(2), 1))\n",
```
jpchen: do these `sample`s need to be aliased to `pyro.sample` for this code block to run?

fritzo: Yes, I'm only abbreviating because of the severe width-constraints in the ascii-art tables

fritzo:
@jpchen Thanks for reviewing!

fehiepsi:
@fritzo Do you know any setting of pyro in `tests` (related to shape stuff) but not enable by default in the main pyro?

fritzo:
@fehiepsi Yes we recently turned on full distribution shape validation in tests. You can enable outside of tests via `pyro.enable_validation(True)`. Grep around to see how it's used.

fehiepsi:
@fritzo Thank you so much!!! Disable it let me pass the tests. So I am going to remove `extra_dim` stuff now. :D

fritzo:
Pasting @katherine-h's review of the tutorial:

> I’m a little confused about the `reshape` terminology. Ordinarily (e.g. in numpy, torch, etc.), applying `reshape` to an array preserves the number of elements, whereas here it seems to be adding elements? If this is right, it might be worth explicitly stating in the tutorial, as it seems pretty counterintuitive relative to other usages.

> My second question is about `extra_event_dims`. I would have thought there’d always be an objective fact of the matter about whether things are dependent or independent, but if I understand this operation correctly, it seems like we can declare some things dependent by fiat? If so, I think some more discussion about why we’re allowed to do this, and in particular when we’d want to do it, would be helpful. 

> For instance, here:

```py
d = Bernoulli(0.5 * torch.ones(3,4)).reshape(extra_event_dims=1)
assert d.batch_shape == (3,)
assert d.event_shape == (4,)
x = d.sample()
assert x.shape == (3, 4)
assert d.log_prob(x).shape == (3,)
```

> I would have thought that all 12 samples are in fact independent and that we therefore shouldn’t be able to say that some are dependent, unless this is actually somehow changing the underlying distribution to some kind of multivariate analog of the Bernoulli? 

> Similarly, in `d = dist.Normal(torch.zeros(1, 3), torch.ones(1, 3)).reshape(extra_event_dims=1)`, does using `extra_event_dims=1` cause Pyro to transform the variance tensor into a square covariance matrix?

> And definitely, for both of these, or at least one of them, a motivating example to show why we’d want to use extra event dims would definitely be helpful.

fritzo:
> @katherine-h I’m a little confused about the reshape terminology.

Good point, I hadn't thought of the `np.reshape` behavior when naming this, and naming is even more important now that `torch.Tensor` also has a `.reshape` method. If you can think of a clearer name, we're happy to change it as part of #827.

> @katherine-h  I would have thought there’d always be an objective fact of the matter about whether things are dependent or independent

Great point. I do have an answer, but I'll put that answer in the tutorial 😄

fehiepsi:
> If you can think of a clearer name, we're happy to change it

@fritzo I propose `enshape` ^^

fritzo:
Let's move renaming discussion to the renaming issue #827 so it's not lost when this PR merges.

katherine-h:
@fritzo I see, that makes more sense.

fritzo:
Any additional requests before we merge this?

martinjankowiak:
let me take a quick look

tutorial/source/tensor_shapes.ipynb
```diff
@@ -0,0 +1,555 @@
...
+    "import pyro\n",
+    "from torch.distributions import constraints\n",
+    "from pyro.distributions import Bernoulli, Categorical, MultivariateNormal, Normal\n",
+    "from pyro.infer import ELBO, config_enumerate\n",
+    "from pyro.optim import Adam\n",
+    "\n",
+    "smoke_test = ('CI' in os.environ)\n",
+    "pyro.enable_validation(True)    # <---- This is always a good idea!\n",
+    "\n",
+    "def test_model(model, guide=None, **kwargs):\n",
```
martinjankowiak: add a comment about the purpose of this helper function?

tutorial/source/tensor_shapes.ipynb
```diff
@@ -0,0 +1,555 @@
...
+    "x2 = d.sample(sample_shape)\n",
+    "assert x2.shape == sample_shape + batch_shape + event_shape\n",
+    "```\n",
+    "In summary\n",
+    "```\n",
+    "      |      iid     | independent | dependent\n",
+    "------+--------------+-------------+------------\n",
+    "shape = sample_shape + batch_shape + event_shape\n",
+    "```\n",
+    "For example univariate distributions have empty event shape (because each number is an independent event). Distributions over vectors like `MultivariateNormal` have `len(event_shape) == 1`. Distributions over matrices like `InverseWishart` have `len(event_sahpe) == 2`.\n",
```
martinjankowiak: typo: len(event_sahpe)

tutorial/source/tensor_shapes.ipynb
```diff
@@ -0,0 +1,555 @@
...
+    "```\n",
+    "This is useful for two reasons: First it allows us to easily swap in a `MultivariateNormal` distribution later. Second it simplifies the code a bit since we don't need an `iarange` (see below) as in\n",
+    "```py\n",
+    "with pyro.iarange(\"x_iarange\", 10):\n",
+    "    pyro.sample(\"x\", dist.Normal(0, 1).reshape([10]))\n",
+    "```\n",
+    "The difference between these two versions is that the second version with `iarange` informs Pyro that it can make use of independence information when estimating gradients, whereas in the first version Pyro must assume they are dependent (even though the normals are in fact independent). This is analogous to d-separation in graphical models: it is always safe to add edges and assume variables *may* be dependent (i.e. to widen the model class), but it is unsafe to assume independence when variables are actually dependent (i.e. narrowing the model class so the true model lies outside of the class, as in mean field). In practice Pyro uses reparameterized gradient estimators for `Normal` distributions so both gradient estimators have the same performance. "
```
martinjankowiak: " In practice Pyro uses reparameterized gradient estimators..."

+ ... when using SVI ... or the like

tutorial/source/tensor_shapes.ipynb
```diff
@@ -0,0 +1,555 @@
...
+    "As an exercise, try to tabulate the shapes of sample sites in one of your own programs."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Subsampling tensors inside an `iarange`    <a class=\"anchor\" id=\"subsampling\"></a>\n",
+    "\n",
+    "One of the main uses of [iarange](http://docs.pyro.ai/en/dev/primitives.html#pyro.__init__.iarange) is to subsample data. This is possible within an `iarange` because data are independent, so the expected value of the loss on half the data should be half the expected value of the loss on the full data.\n",
```
martinjankowiak: "so the expected value of the loss on half the data " ==>
so the expected value of the loss on, say, half the data

fritzo:
Thanks for reviewing, @martinjankowiak

PR #937, opened 2018-03-27
automatically get latest pytorch version in docs

jpchen:
workarounds because we cant install `wget` on rtd servers
sorry for the commit history had to force push to trigger rebuilds


jpchen:
@neerajprad 
using `torch-0.4.0a0+2df578a-cp27-cp27mu-linux_x86_64.whl`, im getting this error:
```
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>7DB591CA63AE7F14</RequestId><HostId>Pn5mlTh9BW42AmkM78EcL06UVQ5Sic0MNU5ZnCqXGi+m2QmpW97VWfU2c2Txh+9PsjLwR89NxYM=</HostId></Error>Invalid requirement: '.whl'
  File "/home/docs/checkouts/readthedocs.org/user_builds/pyro-ppl/envs/docs-upgrade/local/lib/python2.7/site-packages/pip/req/req_install.py", line 82, in __init__
    req = Requirement(req)
  File "/home/docs/checkouts/readthedocs.org/user_builds/pyro-ppl/envs/docs-upgrade/local/lib/python2.7/site-packages/pip/_vendor/packaging/requirements.py", line 96, in __init__
    requirement_string[e.loc:e.loc + 8]))
```
can i not curl into an output the same way wget downloads it? i checked and they should support the same protocols. any ideas?

neerajprad:
@jpchen - `curl` works fine, I just checked. There is probably something wrong with your url, try to print it out. In any case, I think if `curl` is the only constraint we should simply modify the install script rather than putting in any hacks here. What do you think?

jpchen:
yes i thought you had a reason for using `wget`. i will modify the script then that couldve saved me all that history of commits :P

neerajprad:
Yeah, I think we should be good as travis should have `curl`.

scripts/install_pytorch.sh
```diff
@@ -48,5 +48,7 @@ else
 fi
 
 # Download wheel and install
-wget -P tmp/ "${WHL_PREFIX}/${!WHL_LOOKUP}.whl"
+curl -o tmp/"${!WHL_LOOKUP}.whl" "${WHL_PREFIX}/${!WHL_LOOKUP}.whl"
+
 pip install tmp/*
+rm -rf tmp
```
neerajprad: This is not needed. See the exit trap above - that should handle the cleanup.

jpchen:
yes I know I was just testing something locally earlier and commented the
cleanup out. And I don't have the commit hash before it's pushed so I'm
testing on a script elsewhere ha

jpchen:
i can finally install torch, but there are errors in `Cauchy` 
```
AttributeError: type object 'Cauchy' has no attribute 'params'
```
appears throughout the code base everywhere its imported. @fritzo is this due to a recent change?

fritzo:
`AttributeError: type object 'Cauchy' has no attribute 'params'` I'm seeing this too.

jpchen:
ready for merge

docs/source/conf.py
```diff
@@ -191,4 +191,7 @@ def setup(app):
 
 # @jpchen's hack to get rtd builder to install latest pytorch
 if 'READTHEDOCS' in os.environ:
-    os.system('sh scripts/install_pytorch.sh')
+    os.system('curl -o install.sh https://raw.githubusercontent.com/uber/pyro/dev/scripts/install_pytorch.sh')
```
neerajprad: Could you explain why this hack is required when the same files are present locally?

PR #938, opened 2018-03-27
use cat instead of stack in HMC

fehiepsi:
Using `torch.stack` for difference size latent variables throw errors because it requires latent variables have same shapes. This pull request changes it to `torch.cat` which covers more cases (still cannot cover arbitrary latent shapes).

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -75,7 +75,7 @@ class HMC(TraceKernel):
         return trace_poutine.trace
 
     def _kinetic_energy(self, r):
-        return 0.5 * torch.sum(torch.stack([r[name]**2 for name in r]))
+        return 0.5 * torch.sum(torch.cat([r[name]**2 for name in r]))
 
     def _potential_energy(self, z):
         # Since the model is specified in the constrained space, transform the
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -75,7 +75,7 @@ class HMC(TraceKernel):
         return trace_poutine.trace
 
     def _kinetic_energy(self, r):
-        return 0.5 * torch.sum(torch.cat([r[name]**2 for name in r]))
+        return 0.5 * sum(x.pow(2).sum() for x in r.values())
 
     def _potential_energy(self, z):
         # Since the model is specified in the constrained space, transform the
--- a/pyro/infer/mcmc/nuts.py
+++ b/pyro/infer/mcmc/nuts.py
@@ -6,7 +6,6 @@ import torch
 
 import pyro
 import pyro.distributions as dist
-from pyro.infer.util import torch_data_sum
 from pyro.ops.integrator import single_step_velocity_verlet
 
 from .hmc import HMC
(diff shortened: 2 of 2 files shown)


pyro/infer/mcmc/hmc.py
```diff
@@ -75,7 +75,7 @@ def _get_trace(self, z):
         return trace_poutine.trace
 
     def _kinetic_energy(self, r):
-        return 0.5 * torch.sum(torch.stack([r[name]**2 for name in r]))
+        return 0.5 * torch.sum(torch.cat([r[name]**2 for name in r]))
```
neerajprad: Can we just change this to `0.5 * torch.sum(torch.cat([r[name].view(-1)**2 for name in r]))`? That should work with arbitrary sized tensors.

pyro/infer/mcmc/nuts.py
```diff
@@ -81,10 +81,10 @@ def __init__(self, model, step_size=None, adapt_step_size=False, transforms=None
         self._max_sliced_energy = 1000
 
     def _is_turning(self, z_left, r_left, z_right, r_right):
-        z_left = torch.stack([z_left[name] for name in self._r_dist])
-        r_left = torch.stack([r_left[name] for name in self._r_dist])
-        z_right = torch.stack([z_right[name] for name in self._r_dist])
-        r_right = torch.stack([r_right[name] for name in self._r_dist])
+        z_left = torch.cat([z_left[name] for name in self._r_dist])
```
neerajprad: Same here. Is there a way to flatten this so that we do not need to have this rather stringent requirement of latent site shapes matching up?

neerajprad:
Let us also add a test with differently shaped latent sites, and ensure that it works! I can add that too in a separate PR.

pyro/infer/mcmc/hmc.py
fritzo: Actually both of these look memory inefficient. Is there a good reason not to use
```py
return 0.5 * sum(x**2 for x in r.values())
```

neerajprad: hey @fritzo, I removed your comment by mistake. Wanted to thumbs up. That was a good idea!

neerajprad: @fritzo's comment:
Actually both of these look memory inefficient. Is there a good reason not to use
```
return 0.5 * sum(x**2 for x in r.values())
```

pyro/infer/mcmc/nuts.py
fehiepsi: I wonder it too. ^^

fritzo: I was going to mention in your GP pr that this might better be done in a loop of accumulators, something like:
```py
diff_left = 0
diff_right = 0
for name in self._r_dist:
    dz = z_right[name] - z_left[name]
    diff_left += (dz * r_left[name]).sum()
    diff_right += (dz * r_right[name]).sum()
return diff_left < 0 or diff_right < 0
```

fehiepsi:
@neerajprad Would you mind adding some tests after merging this pull request? I tested it with GP models, where latent variables are a mix of 1D, 2D tensors.

neerajprad:
> Would you mind adding some tests after merging this pull request? 

Sure, I’ll do that!

Issue #942, opened 2018-03-27
Empirical distribution for MCMC


neerajprad:
As sketched in #655 (but not merged), this is to have an empirical distribution with a limited scope (mentioned below). Amongst other possible use cases, this will be useful as a convenience utility for MCMC algorithms. For instance, to not require end users to have to manipulate traces. e.g. from HMC tests:

```
    for trace, _ in mcmc_run._traces(data):
        posterior.append(trace.nodes['p_latent']['value'])
```

 - The aim is to hold a collection of data, and assess the statistical properties of the sample, e.g. mean, median, variance, etc. Scoring additional samples via `log_prob` is not supported due to computational reasons (while we can do some binning and log prob computation, for most real (even toy sized) problems, this will be very expensive. e.g. `torch.histc` only supports histograms for scalars).
 - Samples are all tensor values.

fritzo:
@eb8680 Would this help at all with our factor graph work #915 ? E.g. we could construct a set of Histograms, each marginalized down to a single site?

eb8680:
> Would this help at all with our factor graph work

Yeah, I was just talking to @neerajprad about this since I was working on something similar this afternoon.  We should add a section to the design doc for it.

neerajprad:
> We should add a section to the design doc for it.

Sounds great! Lets sync up on this.

PR #943, opened 2018-03-28
 Adding EmpiricalDistribution and testing on MCMC

neerajprad:
This adds the draft for the `EmpiricalDistribution` as discussed in #942, for discussion. Broadly the design is as follows:
 - `EmpiricalDistribution` holds a collection of data points with the same type and shape. These data-points are collected along the leftmost dimension. It provides a method called `add` to add a single datum to the existing collection. These are lazily collected into a tensor, so as to avoid repeated tensor copies per iteration. We can compute `mean`, `variance` and generate samples from `EmpiricalDistribution` for any arbitrary shape. 
 - `MCMC` inherits off of `TraceJoint` that stores a bag of traces corresponding to the joint distribution of latents and observed. We can call `.marginal(site=x)` on this, to return the marginal for a given sample site (we can add other methods that condition, etc.). The return type is the `EmpiricalDistribution`.

What needs to be fixed:
 - There is redundancy between `TracePosterior` and `TraceJoint`. We need to merge the two, retaining performance when we do not need `log_prob`.
 - Using `EmpiricalDistribution` in `Histogram`, or generalizing it to support `Histogram`.
 - Other improvements.

cc. @eb8680, @fritzo - Let us discuss any requirements to generalize this for #915. 

Tests: Tests added in `test_empirical.py`, and HMC/NUTS tests are modified.

PS: Let us restrict the review to discussing design requirements, if possible. I will change the label once this is ready for a full review.

### DIFF
--- a/pyro/distributions/empirical.py
+++ b/pyro/distributions/empirical.py
@@ -1,0 +1,75 @@
+from __future__ import absolute_import, division, print_function
+
+import numpy as np
+import torch
+
+from pyro.distributions.distribution import Distribution
+
+
+class Empirical(Distribution):
+    r"""
+    Empirical distribution associated with the sampled data. This is
+    used for assessing the statistical properties of the sample.
+    In particular, scoring of additional data points from the sample
+    via ``log_prob`` is not currently supported.
+    """
+
+    def __init__(self):
+        self.samples = []
+        self.tensorized_samples = None
+        super(Distribution, self).__init__()
+
+    @property
+    def finalized_tensor(self):
+        """
+        Collect samples along the left-most dimension and return the
+        tensor.
+
+        :return: torch.Tensor
+        """
+        if not self.samples or self.tensorized_samples is None:
+            return self.tensorized_samples
+        new_samples = torch.stack(self.samples, dim=0)
+        self.tensorized_samples = torch.cat([self.tensorized_samples, new_samples], dim=0)
+        self.samples = []
+        return self.tensorized_samples
+
+    def add(self, value):
+        """
+        Adds the data point to the sample. The values in successive
+        calls to ``add`` must have the same tensor shape and size.
+
+        :param torch.Tensor value: tensor to add to the sample.
+        """
+        if self.tensorized_samples is None:
+            self.tensorized_samples = value.new(0)
+        self.samples.append(value)
+
+    def sample(self, sample_shape=torch.Size()):
+        if not sample_shape:
+            rand_idx = np.random.randint(0, high=self.finalized_tensor.size(0), dtype=np.int64)
+        else:
+            rand_idx = np.random.randint(0, high=self.finalized_tensor.size(0), dtype=np.int64,
+                                         size=np.prod(sample_shape).astype(np.int64))
+        return self.finalized_tensor[rand_idx].view(sample_shape + self.batch_shape)
+
+    def log_prob(self, value):
+        raise NotImplemented("Empirical distribution does not implement `log_prob`")
+
+    @property
+    def batch_shape(self):
+        if self.finalized_tensor is None:
+            return None
+        return self.finalized_tensor.size()[1:]
+
+    @property
+    def mean(self):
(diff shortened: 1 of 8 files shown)


fritzo:
Design looks reasonable so far.

For discrete enumeration it would be nice to support `.log_prob()`. I think it's fine to treat even continuous-valued `EmpiricalDistribution`s as discrete sums of deltas; this gives us room to define a similar `SmoothedEmpiricalDistribution` with `.log_prob()` using kernel density estimation and with `.sample()` adding a bit of noise. Did you have specific reasons for not implementing `.log_prob()`?

neerajprad:
>  Did you have specific reasons for not implementing .log_prob()?

I should have explained this in more detail. I discussed the main reason with @eb8680 in #845. The issue is with doing equality comparison on large tensors to compute their weights, even if most of these weights are in practice `1`, specially with continuous samples (see the Histogram implementation). 

The other reason is that most of the use cases (that I am familiar with and see in tests/examples) are around using `EmpiricalDistribution` to compute statistical properties of the existing sample rather than score new samples. My guess is that doing this for any non toy sized problem with continuous values will be fraught with difficulties until we have sophisticated KDE algorithms built in. Until we do so, I wanted to target the remaining 95% (currently) use cases which do not need to incur a performance penalty when they are not using `log_prob`. I would like to understand some of the use cases you might have in mind though!

> For discrete enumeration it would be nice to support .log_prob()

I was thinking of `EmpiricalDistribution` in general terms as a collection of tensors with consistent shape/type. When you say discrete enumeration, do you mean any discrete sample site? Say we have a discrete sample site that generates a sample of shape `[p, q, r]` each of which is drawn from a categorical distribution with `K` categories. Would we then track the number of samples over the entire `p x q x r x K` space?

> EDIT Note that it should be cheap-ish to compute .log_prob() from tensors via something like

Is this for discrete samples? Sorry, I didn't fully understand the example, but looking forward to chatting more on that.

neerajprad:
@eb8680, @fritzo - With some minor tweaks (and @fritzo's suggestion above), I think we can extend this to support both importance weights, as well as not having to pay any book-keeping penalty for using `.log_prob`. Then we can just plug this into Histogram and TracePosterior.

fritzo:
@neerajprad One use case I'm interested in is in computing marginals of a discrete state HMM, say with one `EmpiricalDistribution` per time step. In this case we would have tensors of empty shape `()` and only a handful of total states (say 10) so `.log_prob()` should be cheap. For efficiency we may want to batch these over multiple time series so that `.weights.dim() == 2` or more; that is,
```py
weights.shape() == my_empirical_dist.batch_shape + (num_classes,)
```
@eb8680 has probably thought about this in more detail...

eb8680:
> I think we can extend this to support both importance weights, as well as not having to pay any book-keeping penalty for using .log_prob

Yep, we should be able to use the same code for both.  As in the current version of `TracePosterior`, the sample weights in the MCMC case are all 1.

> computing marginals of a discrete state HMM

Yeah, I was working on that yesterday and will try to open a PR today.  Getting batching over a global independence dimension working seamlessly has been a little more involved than I expected.

fritzo:
> Getting batching over a global independence dimension working

Can you simply use `iarange` dimensions, similar to how we parallelized the models in test_enum.py, e.g. [here](https://github.com/uber/pyro/blob/8ac0f11/tests/infer/test_enum.py#L197)?

eb8680:
@neerajprad let's split off a PR for `Empirical` and get it merged separately so we can get a version with weights working and clean up `Histogram` as well

neerajprad:
> @neerajprad let's split off a PR for Empirical and get it merged separately so we can get a version with weights working and clean up Histogram as well

Yup, I'll get on to it next.

neerajprad:
>  and clean up Histogram as well

@eb8680 - Are you doing this as part of the refactor? Feel free to assign any related sub-tasks to me.

eb8680:
Addressed by #1019

PR #945, opened 2018-03-28
Misc fixes to ADVI

fritzo:
- Support scalars in `log_abs_det_jacobian`
- Directly use `scale_tril` argument to `MultivariateNormal`
- Fix shaping issues with `StickBreakingTransform`
- Return a dict mapping sample site name -> value from the `advi.guide()`. This is free to compute and often useful.

## Tested
- unmarked lots of tests that were previously xfailing
- added a test with Dirichlet; it passes the shape part but xfails for another reason (possible bug in stick breaking jacobian)

### DIFF
--- a/pyro/infer/advi.py
+++ b/pyro/infer/advi.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import torch
-from torch.distributions import constraints, transform_to
+from torch.distributions import biject_to, constraints
 
 import pyro
 import pyro.distributions as dist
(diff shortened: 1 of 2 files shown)


fritzo:
I'm hoping the ADVI interface will settle down before 0.2.0 release. If it is still in flux we may move it temporarily to pyro.contrib and move it back to core in the 0.2.1 release.

pyro/infer/advi.py
```diff
@@ -43,9 +43,10 @@ def _setup_prototype(self, *args, **kwargs):
         # run the model so we can inspect its structure
         self.prototype_trace = poutine.block(poutine.trace(self.base_model).get_trace)(*args, **kwargs)
         self.prototype_trace = prune_subsample_sites(self.prototype_trace)
-        self.latent_dim = sum(site["value"].view(-1).size(0)
-                              for site in self.prototype_trace.nodes.values()
-                              if site["type"] == "sample" and not site["is_observed"])
+        self.latent_shapes = {name: biject_to(site["fn"].support).inv(site["value"]).shape
```
martinjankowiak: why is this necessary? can we have a comment?

fritzo: comment added.

PR #947, opened 2018-03-28
remove pyro.observe

jpchen:
#827 inching closer to 0 .2

### DIFF
--- a/pyro/__init__.py
+++ b/pyro/__init__.py
@@ -98,19 +98,6 @@ def sample(name, fn, *args, **kwargs):
         return msg["value"]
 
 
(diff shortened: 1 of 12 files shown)


neerajprad:
Thanks @jpchen! You may need to remove this from some of the notebooks too.

fritzo:
I assume that all refactoring PRs have run `grep -R` or `ag` or similar on the entire codebase. @jpchen can you confirm?

jpchen:
> I assume that all refactoring PRs have run grep -R
only did on `pyro` and `tests` :( rookie mistake

PR #848, opened 2018-03-04
Split TraceEnum_ELBO out of Trace_ELBO

fritzo:
Addresses #847 

This PR:
1. renames `Trace_ELBO` to `TraceEnum_ELBO` (since it supports enumeration by default)
2. creates a new simple `Trace_ELBO` replicating old non-enumerating behavior
3. adds the `enum_discrete` flag back to `SVI`

PR #948, opened 2018-03-28
Revise Delta distribution to support ADVI + Transforms

fritzo:
This revises our `Delta` distribution to work with transforms in ADVI.

## The Problem

Before this PR, ADVI used `TransformeDistribution(Delta, ...)`, however this combination fails due to finite precision in the `x -> t(x) -> t.inv(t(x))` round trip, leading to sampled values accidentally missing the infinitesimally narrow delta from whence they came. The reason we are using `Deltas` at all is to allow the ADVI guide to keep all randomness in a single packed `Normal` or `MultivariateNormal` latent, and to unpack this into multiple `Delta` sample statements (one per sample site in the model).

## Proposed Solution

This PR adds a bit of functionality to our `Delta` distribution so that it can function as a proper `TransformedDistribution(Delta, ...)`. This is mathematically appealing, since after this PR, the class of `Delta` distributions is closed under transformations by differentiable functions. Specifically, this PR:
1. Adds a `log_density` parameter to `Delta`, so we can replace
   ```diff
   - TransformedDistribution(Delta(v), t)
   + Delta(t(v), log_density=-t.log_abs_det_jacobian(v, t(v)))
   ```
2. Adds an `event_dim` parameter to `Delta`, so we can support multivariate transforms (and their Jacobians)
3. Removes the [unused] `.enumerate_support()` method since the `log_density` scales `log_prob` such that it is no longer normalized when considered as a discrete distribution. If we have use for discrete delta distributions in the future, we should consider creating a new class without `.rsample()`.

Alternate solutions include:
- Creating a separate `TransformedDelta` distribution
- Restricting our use of `TransformedDistribution(Delta(v), transform)` to cached transforms so they can avoid the round-trip.
- Using fuzzy comparison in `Delta.log_prob(-)`. This is a bad idea.

## Tested

- [x] added shape tests for the new `Delta` parameters
- [x] added loss assertions to existing ADVI tests
- [x] added a test with Dirichlet distribution

### DIFF
--- a/pyro/infer/trace_elbo.py
+++ b/pyro/infer/trace_elbo.py
@@ -112,7 +112,7 @@ class Trace_ELBO(ELBO):
                         guide_site = guide_trace.nodes[name]
                         guide_log_prob, score_function_term, entropy_term = guide_site["score_parts"]
 
-                        elbo_particle = elbo_particle + model_log_prob_sum - guide_log_prob.sum()
+                        elbo_particle = elbo_particle + (model_log_prob_sum.item() - guide_log_prob.sum().item())
                         surrogate_elbo_particle = surrogate_elbo_particle + model_log_prob_sum
 
                         if not is_identically_zero(entropy_term):
--- a/pyro/distributions/delta.py
+++ b/pyro/distributions/delta.py
@@ -1,9 +1,12 @@
 from __future__ import absolute_import, division, print_function
 
+import numbers
+
 import torch
 from torch.distributions import constraints
 
 from pyro.distributions.torch_distribution import TorchDistribution
+from pyro.distributions.util import sum_rightmost
 
 
 class Delta(TorchDistribution):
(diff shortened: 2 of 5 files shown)


pyro/infer/trace_elbo.py
```diff
@@ -112,7 +112,7 @@ def loss_and_grads(self, model, guide, *args, **kwargs):
                         guide_site = guide_trace.nodes[name]
                         guide_log_prob, score_function_term, entropy_term = guide_site["score_parts"]
 
-                        elbo_particle = elbo_particle + model_log_prob_sum - guide_log_prob.sum()
+                        elbo_particle = elbo_particle + (model_log_prob_sum.item() - guide_log_prob.sum().item())
```
martinjankowiak: why was this needed?

fritzo: This fixes a tensor->float conversion that I missed back in #848. Everywhere else we assume that `SVI.step()` returns a Python float as loss. This fix also increases the precision of the accumulator from (typically) 32bits to 64bit floats. I've added it to this PR because I'm testing the loss values and realized they weren't `float`s as expected.

fritzo:
Ok, I'm happy with this PR now that the tests are all passing (and not xfailing).

PR #949, opened 2018-03-29
Parallel Search inference prototyping

eb8680:
One test passes, but still exponentially expensive...

### DIFF
--- a/pyro/infer/search.py
+++ b/pyro/infer/search.py
@@ -4,6 +4,8 @@ import pyro.poutine as poutine
 from pyro.infer import TracePosterior
 from six.moves.queue import Queue
 
+from .enum import config_enumerate
+
 
 class Search(TracePosterior):
     """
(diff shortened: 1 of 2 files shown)


pyro/infer/search.py
```diff
@@ -40,3 +42,34 @@ def _traces(self, *args, **kwargs):
...
+        p = poutine.trace(
+            poutine.enum(
+                config_enumerate(self.model, default="parallel"),
+                first_available_dim=self.first_available_dim))
+        tr = p.get_trace(*args, **kwargs)
+        # now compute joint probabilities:
+        # TODO identify all global independence dimensions,
+        # and aggregate over all non-global ones
+        tr.compute_log_prob()
+        log_joints = sum([tr.nodes[name]["log_prob"] for name in tr.nodes
```
fritzo: Maybe the basic datatype should be a dict of factors?
```py
log_factors = {name: site["log_prob"]
               for name, site in tr.nodes.items()
               if site["type"] == "sample"}
return tr, log_factors
```
That way downstream consumers can use it as efficiently as they like, while still being able to resort to
```py
log_joints = sum(log_factors.values())
```

PR #950, opened 2018-03-29
Consolidate escape and enumerate into continuation, centralize _get_traces/_traces logic

eb8680:
Part of refactoring mentioned in design doc... Will update this description with more details today


fritzo:
Looks very clean so far.

fritzo:
Any progress on this? It would be nice to be able to use this, and I'll end up pulling pieces out of this PR until it merges.

eb8680:
> Any progress on this? It would be nice to be able to use this, and I'll end up pulling pieces out of this PR until it merges.

I guess this PR can be merged mostly as-is after the merge conflicts are fixed.

pyro/infer/enum.py
```diff
@@ -48,6 +51,50 @@ def iter_discrete_traces(graph_type, fn, *args, **kwargs):
...
+                           max_iarange_nesting=float('inf')):
+    """
+    Iterate over global non-parallel importance samples of a model-guide pair,
+    including both parallel and non-parallel enumeration
+    if the guide is annotated for enumeration.
+
+    :param num_particles: number of non-parallel importance samples to take
+    :param graph_type: the type of the graph, e.g. "flat" or "dense".
+    :param max_iarange_nesting: maximum depth of iaranges
+    :returns: An iterator over model/guide trace pairs
```
fritzo: I guess this returns a callable that returns a generator?

pyro/infer/trace_elbo.py
```diff
@@ -49,25 +47,8 @@ def _get_traces(self, model, guide, *args, **kwargs):
...
-            if infer.is_validation_enabled():
-                for site in model_trace.nodes.values():
-                    if site["type"] == "sample":
-                        check_site_shape(site, self.max_iarange_nesting)
-                for site in guide_trace.nodes.values():
-                    if site["type"] == "sample":
-                        check_site_shape(site, self.max_iarange_nesting)
-
-            yield model_trace, guide_trace
+        return iter_importance_traces(num_particles=self.num_particles,
```
fritzo: nit: it might clean up style throughout uses of `iter_importance_trace` if we consistently introduced a temporary `get_traces`:
```py
get_traces = iter_importance_traces(num_particles=self.num_particles, graph_type="flat")
return get_traces(model, guide, *args, **kwargs)
```

pyro/poutine/__init__.py
```diff
@@ -99,14 +98,32 @@ def escape(fn, escape_fn=None):
...
     Given a callable that contains Pyro primitive calls,
     evaluate escape_fn on each site, and if the result is True,
     raise a NonlocalExit exception that stops execution
     and returns the offending site.
     """
-    return EscapePoutine(fn, escape_fn)
+    return ContinuationPoutine(fn, escape_fn, util.escape_cont_fn)
+
+
+def enum(fn, first_available_dim):
```
fritzo: Do we really want to expose this? I'm tempted to move all the enumeration-related helpers to pyro.infer.enum. Since that library is undocumented (i.e. absent from sphinx), we can reduce guarantees of backwards compatibility.

pyro/poutine/__init__.py
```diff
@@ -126,6 +143,25 @@ def condition(fn, data):
     return ConditionPoutine(fn, data=data)
 
 
+def continuation(fn, escape_fn, continuation_fn, first_available_dim=None):
```
fritzo: Do we really want to expose this? It seems weird exposing a detail like `first_available_dim` in the poutine interface, and I can't think of examples beyond `EnumeratePoutine`

Also nit: argument should be named `cont_fn`

pyro/poutine/util.py
```diff
@@ -128,3 +128,40 @@ def all_escape(trace, msg):
     return (msg["type"] == "sample") and \
         (not msg["is_observed"]) and \
         (msg["name"] not in trace)
+
+
+def broadcast_enum_filter(msg):
```
fritzo: It seems like these helpers would naturally live in pyro.infer.enum, where most of the enumeration logic lives.

pyro/poutine/util.py
```diff
@@ -128,3 +128,40 @@ def all_escape(trace, msg):
...
+    if msg["done"] or msg["type"] != "sample" or msg["is_observed"]:
+        return False
+    return msg["infer"].get("enumerate") == "parallel"
+
+
+def broadcast_enum_cont(msg):
```
fritzo: ditto

pyro/poutine/continuation_poutine.py
```diff
@@ -0,0 +1,99 @@
...
+        Resets the next available expansion dimension.
+        """
+        self.next_available_dim = self.first_available_dim
+        return super(ContinuationMessenger, self).__enter__()
+
+    def _postprocess_message(self, msg):
+        """
+        Increments the next available expansion dimension.
+        """
+        if "next_available_dim" in msg["infer"] and \
```
fritzo: This seems pretty specific for default behavior.

pyro/poutine/continuation_poutine.py
```diff
@@ -0,0 +1,99 @@
...
+    and makes sure it gets called and unwound correctly.
+    Continuations give up (or simulate giving up) control of execution
+    by expanding along a dimension, yielding a coroutine,
+    or raising a special poutine.util.NonlocalExit exception.
+
+    It is currently used primarily for sequential and parallel enumeration.
+
+    Generalizes and replaces EscapeMessenger and EnumerateMessenger.
+    """
+    def __init__(self, escape_fn, cont_fn, first_available_dim):
```
fritzo: It seems like it would be cleaner to split this spit this into two classes: a base class `ContinuationMessenger` to handle `escape_fn,cont_fn` and a derived class `EnumerateMessenger` to handle `first_available_dim` and `next_available_dim`. It seems conflated to manage that dimension logic in e.g. `poutine.escape`. Also it will be easier to refactor dimension logic if we isolate that to the messenger that needs it, i.e. `EnumerateMessenger`.

fritzo:
also btw it sure seems like we're getting lots of circular dependencies in this PR, e.g. importing pyro.infer from within submodules of pyro.infer. Feel free to move things from some_module/__init__.py to some_module/util.py to break these dependencies.

pyro/poutine/__init__.py
eb8680: This is just for API compatibility during the refactor, will remove later

pyro/poutine/continuation_poutine.py
eb8680: I kept it this way because this is mid-refactor; in the final version (if I ever figure out dependency tracking...) `ContinuationMessenger` will maintain an internal stack of `IndepMessenger`s to mark multiple invocations of the rest of the program independent.  These will be used for both `escape` and enumeration.

Basically, there's no reason for `ContinuationMessenger` to exist if it's not handling this logic.

pyro/poutine/__init__.py
eb8680: I was just following the convention of the rest of `poutine/__init__.py` in exposing all constructor args in the alias, will remove it and rename `continuation_fn`.

pyro/poutine/continuation_poutine.py
fritzo: Ok, that's fine if it's mid-refactor. The main thing I'm trying to avoid is the leaking of specific details (like `first_available_dim`) into general interfaces. I found this to be awkward in e.g. `poutine.queue` which takes args `num_samples` and `max_tries`. I'd like to make a general effort to keep specific details of derived classes from leaking into base classes, so as to avoid all derived classes having to know about details of their siblings (and substitute oo with fp or your favorite abstraction mechanism).

PR #951, opened 2018-03-29
Support simple iarange in ADVI; fix a shape bug

fritzo:
Based on #956 

This fixes some site shapes and adds support for (non-subsampled) `iarange` in ADVI. This also adds a dependency on `contextlib` for `ExitStack` (backported from Python3's `contextlib`).

Even after this PR our ADVI implementation handles only a small class of models:
- no handling of subsampling
- no handling of `irange`
- no handling of discrete latent variables

## Tested

- added some cheap tests in tests/infer/test_advi.py
- ran a time series model with various shapes (works after this PR, failed before)


fritzo:
Ready to merge as soon as tests pass.

PR #952, opened 2018-03-29
Updating to PyTorch master with no tensor type casting

### LINKED ISSUE #944: Update to latest PyTorch which does type inference in torch.tensor()
As of https://github.com/pytorch/pytorch/pull/5997, `torch.tensor` cares about whether its args are floats or ints:
```py
>>> torch.tensor(0).type()
'torch.LongTensor'
>>> torch.tensor(0.).type()
'torch.FloatTensor'
```
This violates our assumptions in both testing and initialization.

neerajprad:
Fixes #944.

Issues were mostly related to:
 - Inferring types as `torch.LongTensor` when lists have integer entries. These are fixed.
 - `numpy.array([1.2])` is cast to torch.double when converted to a `torch.Tensor`, instead of using the default tensor type. I had to explicitly cast these to use the default tensor type. See below.
 - A number of tests are additionally failing which seem to not be due to this tensor type inference change. These are marked as xfail and mentioned in #953. I would suggest that we merge this PR asap to update to the latest master and then prioritize fixing the tests in #953. 

Change with respect to wrapping numpy arrays:

```py
>>> import torch
>>> import numpy as np
>>> torch.set_default_tensor_type(torch.float)
>>> torch.tensor(np.array([0.]))
 0
[torch.DoubleTensor of size (1,)]

>>> torch.tensor(np.array([0.])).type(torch.empty(1).type())
 0
[torch.FloatTensor of size (1,)]
```


tutorial/source/bayesian_regression.ipynb
```diff
@@ -70,7 +70,7 @@
     "    # b = 1\n",
     "    y = np.matmul(X, w) + np.repeat(1, N) + np.random.normal(0, noise_std, size=N)\n",
     "    y = y.reshape(N, 1)\n",
-    "    X, y = torch.tensor(X), torch.tensor(y)\n",
+    "    X, y = torch.tensor(X).type(torch.empty(1).type()), torch.tensor(y).type(torch.empty(1).type())\n",
```
fritzo: What does this weird trick do? Would it be possible to instead `.type(torch.Tensor)`?

examples/dmm/polyphonic_data_loader.py
```diff
@@ -117,10 +117,10 @@ def get_mini_batch(mini_batch_indices, sequences, seq_lengths, cuda=False):
     # get mask for mini-batch
     mini_batch_mask = get_mini_batch_mask(mini_batch, sorted_seq_lengths)
 
-    # wrap in PyTorch Tensors
-    mini_batch = torch.tensor(mini_batch)
-    mini_batch_reversed = torch.tensor(mini_batch_reversed)
-    mini_batch_mask = torch.tensor(mini_batch_mask)
+    # wrap in PyTorch Tensors, using default tensor type
+    mini_batch = torch.tensor(mini_batch).type(torch.empty(1).type())
```
martinjankowiak: why is this needed this seems crazy?

neerajprad: For the reason, mentioned above. Any numpy array when wrapped into a torch tensor is converted to a double tensor by default. This recasts it based on the default tensor type. Let me know if there is a less uglyish way to do this.

tutorial/source/bayesian_regression.ipynb
neerajprad: Yup, that's much cleaner :) I forgot that `torch.Tensor` just aliases the default tensor type. The reason we had to do this is because `torch.tensor(np.array([0.]))` gives a double tensor by default now.

examples/dmm/polyphonic_data_loader.py
jpchen: type_as would marginally decrease verbosity/ugliness

fehiepsi:
@neerajprad I am tracking down recent PyTorch changes. Will leave a message soon.

fehiepsi:
```
     def __init__(self, loc, covariance_matrix=None, scale_tril=None):
        super(MultivariateNormal, self).__init__(loc, covariance_matrix, scale_tril)
```
@neerajprad mvn in Pytorch support precision matrix, I think we have to change __init__ here with additional arg. And to prevent new fail like that. I propose to explicitly init `loc=loc`, `cov=cov`, `scale=scale`

neerajprad:
>  mvn in Pytorch recently support precision matrix, I think we have to change init here with additional arg. And to prevent new fail like that. I propose to explicitly init loc=loc, cov=cov, scale=scale

@fehiepsi : Thanks for digging in! I'll make the changes.

fritzo:
@neerajprad Actually we can simply remove our `MultivariateNormal` wrapper, since `.mean` and `.variance` have been moved upstream.

neerajprad:
> @neerajprad Actually we can simply remove our MultivariateNormal wrapper, since .mean and .variance have been moved upstream.

I am doing exactly that. I had removed the `__init__` method earlier, but it seems like we reintroduced it at some point. Will remove the complete wrapper now.

fehiepsi:
@neerajprad My bad, I added it again >"<

neerajprad:
> @neerajprad My bad, I added it again >"<

Np at all. Good reason to remove it now for good! :)

neerajprad:
For some reason, travis is acting up and not starting the build. Should be good to merge once tests pass. All xfails are removed.

neerajprad:
Ready to merge.

PR #954, opened 2018-03-29
refresh dmm ipython notebook

martinjankowiak:
this constitutes (hopefully most but probably not all) the changes needed to refresh the dmm notebook.

i suggest we merge these sorts of things without excessively detailed reviewing. then when a number of these changes have been merged, we can do a round robin review of some sort that ensures each tutorial was carefully read by at least one person other than the author.

### DIFF
--- a/tutorial/source/dmm.ipynb
+++ b/tutorial/source/dmm.ipynb
@@ -64,15 +64,9 @@
     "\n",
     "### The Gated Transition and the Emitter\n",
     "\n",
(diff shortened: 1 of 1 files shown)


martinjankowiak:
- did you test that the html renders correctly when built locally?
no i've never been able to get that to work locally (tried again; failed)
- are these code changes reflected in the examples source code?
i'm still working in the PR to sync things. there's no need to do that right away since no one will be pushing the html anyway

Issue #953, opened 2018-03-29
Tests failing on PyTorch master 


neerajprad:
The following tests (list may increase) fail on PyTorch master, which seem unrelated to tensor casting issue in #944. I encountered these while fixing the tensor casting issues in #952. Easy way to replicate this would be to run `bash scripts/install_pytorch.sh` with commit `2df578a` --> `1807bac` in README.md\*. I might skip these for the time being.

 - [x] `bayesian_regression.py` - `RuntimeError: you can only change requires_grad flags of leaf variables.` cc. @jpchen 
 - [x] Tests using `ADVIMultivariateNormal` in `test_advi.py` - `advi covariance off`. cc. @fritzo
 - [x] Tests in `gp/test_models.py` - `test_inference`, `test_inference_svgp`, `test_inference_deepGP`. `Lapack Error in potrf : the leading minor of order 2 is not positive definite`, and assertion failure on variance. cc. @fehiepsi 
 - [x] Tests in `gp/test_likelihoods` - same as above.
 - [x] `gp.ipynb` - Non-deterministic travis failures due to constraint checking in multivariate normal. See discussion below.

EDIT - After removing the mvn wrapper, all but `bayesian_regression.py` and `gp.ipynb` need to be fixed.

\* Separately I will modify the script to take in an optional arg for build commit.

colesbury:
I think these lines trigger the "requires_grad flags" issue:

https://github.com/uber/pyro/blob/de50a8ee5854c4f00f37cbd851b536b8fbfe2151/examples/bayesian_regression.py#L76

https://github.com/uber/pyro/blob/de50a8ee5854c4f00f37cbd851b536b8fbfe2151/examples/bayesian_regression.py#L79

The issue is that w_mu is that the `type_as` comes after the requires_grad assignment, so w_mu is no longer a leaf variable. The existing code is equivalent to:

```python
w_mu_hidden = torch.randn(1, p, requires_grad=True)  # leaf variable (optimizable)
w_mu = w_mu_hidden.type_as(data)  # backprops to of w_mu_hidden via backward type conversion
```

A fix is:

```python
w_mu = torch.randn(1, p).type_as(data)
w_mu.requires_grad = True
```

(We need a better way to avoid or detect these sorts of issues in PyTorch)

fritzo:
I've been seeing spurrious failures in gp.ipynb. Are the random seeds being fixed? @fehiepsi any ideas? https://travis-ci.org/uber/pyro/jobs/360155682

[...]

PR #956, opened 2018-03-30
Temporary fix build by disabling smoke test on gp tutorial

neerajprad:
Discussion in #953. This seems to strangely and non-deterministically trigger on travis-ci. I am not able to replicate it locally yet.

### DIFF
--- a/tutorial/source/gp.ipynb
+++ b/tutorial/source/gp.ipynb
@@ -58,7 +58,7 @@
     "pyro.set_rng_seed(0)\n",
     "\n",
     "import os\n",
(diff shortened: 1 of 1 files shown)


fritzo:
Thanks for diagnosing this, Neeraj

neerajprad:
@fritzo - I canceled the integration tests, and since the examples stage passes, we can merge this.

jpchen:
i recently learned cancelling unnecessary tests is frowned upon now, so i enabled the merging restrictions to require tests pass. i disabled it to merge this, but if they dont cost us anything, we should maybe always allow tests to run to completion. cc: @fritzo

PR #967, opened 2018-04-01
Refactor Dice towards support for message passing

fritzo:
This is a pure-refactoring PR: behavior should be unchanged.

- Generalizes `Dice` to arbitrary dags and moves the `iarange`-based specialization to `TraceEnum_ELBO`. This is really cool since it allows us to encode graph structure in an `ordinal` type and automatically allows `Dice` to be used outside of `TraceEnum_ELBO`.
- Refactors to collect log factors and identifies where a smarter sum-product algorithm could be placed.

## Performance impact

Running `tests/infer/test_enum.py`:

| Before | 48.9 sec |
| -- | -- |
| After | 49.8 sec |

### DIFF
--- a/pyro/infer/traceenum_elbo.py
+++ b/pyro/infer/traceenum_elbo.py
@@ -8,23 +8,28 @@ import pyro.poutine as poutine
 from pyro.distributions.util import is_identically_zero
 from pyro.infer.elbo import ELBO
 from pyro.infer.enum import iter_discrete_traces
-from pyro.infer.util import MultiFrameDice
+from pyro.infer.util import Dice
 from pyro.poutine.enumerate_poutine import EnumeratePoutine
 from pyro.poutine.util import prune_subsample_sites
 from pyro.util import check_model_guide_match, check_site_shape, check_traceenum_requirements, is_nan
 
 
 def _compute_dice_elbo(model_trace, guide_trace):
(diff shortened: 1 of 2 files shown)


pyro/infer/traceenum_elbo.py
```diff
@@ -8,23 +8,28 @@
...
+    elbo = 0.0
     for name, model_site in model_trace.nodes.items():
         if model_site["type"] == "sample":
             cost = model_site["log_prob"]
             if not model_site["is_observed"]:
                 cost = cost - guide_trace.nodes[name]["log_prob"]
-            dice_prob = dice.in_context(model_site["cond_indep_stack"])
             # TODO use score_parts.entropy_term to "stick the landing"
-            elbo = elbo + (dice_prob * cost).sum()
+            elbo = elbo + dice.expectation(cost, ordering[name])
```
eb8680: Why wrap this in `expectation`?  Is the final `(dice_prob * cost).sum()` in `dice.expectation` going to change?

fritzo: Good point. I was thinking that `.expectation()` would be the public interface, but I guess `.get_dice_prob()` is a littler simpler as a public interface. I'll change the interface.

fritzo:
ready to merge as soon as tests pass. thanks for reviewing!

PR #969, opened 2018-04-02
Adding implementation for EmpiricalDistribution

neerajprad:
Implements EmpiricalDistribution based on comments and suggestions from @fritzo and @eb8680.

The goal was the following:
 - Have a generic `EmpiricalDistribution` implementation that can accept importance weights.
 - Supporting `log_prob`. However, supporting this should not result in a performance penalty for cases where we do not need to compute `log_prob`. Hence any heavy pre-processing (like tensor comparisons) required when adding samples to support `log_prob` is bypassed.
 - Supporting fast `append` operation, without too much overhead - we append to a buffer which is lazily concatenated to an internal tensor when needed. 
 - Should be possible to re-use the same `EmpiricalDistribution` implementation. e.g. appending some value, computing the mean, appending some more values, re-computing the mean etc. There should be no assumption that the add operation precedes all statistical computation on the samples.

What is currently missing?
 - I have not figured out a way to support vectorized `log_prob`, i.e. the ability to score multiple samples at the same time. Doing so will likely require a different implementation that needs to pre-process samples as they are added, and will be slow. Currently, `log_prob` only accepts single values.

Addresses #942. 

Related: #943 - Alternate sketch that does not support `log_prob` or `importance_weights`.

### DIFF
--- a/pyro/distributions/empirical.py
+++ b/pyro/distributions/empirical.py
@@ -29,7 +29,8 @@ class Empirical(TorchDistribution):
         self._weights_buffer = []
         super(TorchDistribution, self).__init__(batch_shape=torch.Size(), validate_args=validate_args)
 
-    def _append_from_buffer(self, tensor, buffer):
+    @staticmethod
+    def _append_from_buffer(tensor, buffer):
         """
         Append values from the buffer to the finalized tensor, along the
         leftmost dimension.
@@ -157,3 +158,7 @@ class Empirical(TorchDistribution):
         for _ in range(self._samples.dim() - 1):
             weights = weights.unsqueeze(-1)
         return (log_sum_exp(weights, scale=deviation_squared, dim=0) - log_sum_exp(weights, dim=0)).exp()
+
+    def enumerate_support(self):
+        self._finalize()
+        return self._samples
--- a/pyro/distributions/empirical.py
+++ b/pyro/distributions/empirical.py
@@ -4,10 +4,11 @@ import numbers
 
 import math
 import torch
+from torch.distributions import constraints
 
 from pyro.distributions.torch_distribution import TorchDistribution
 from pyro.distributions.torch import Categorical
-from pyro.distributions.util import copy_docs_from, is_validation_enabled
+from pyro.distributions.util import copy_docs_from
 from pyro.distributions.util import log_sum_exp
 
 
@@ -18,14 +19,15 @@ class Empirical(TorchDistribution):
     """
 
     arg_constraints = {}
+    support = constraints.real
 
-    def __init__(self):
+    def __init__(self, validate_args=None):
         self._samples = None
         self._log_weights = None
         self._categorical = None
         self._samples_buffer = []
         self._weights_buffer = []
-        super(TorchDistribution, self).__init__()
+        super(TorchDistribution, self).__init__(batch_shape=torch.Size(), validate_args=validate_args)
 
     def _append_from_buffer(self, tensor, buffer):
         """
@@ -64,7 +66,7 @@ class Empirical(TorchDistribution):
             return 0
         return self._samples.size(0)
 
-    def add(self, value, log_weight=None, weight=None):
+    def add(self, value, weight=None, log_weight=None):
         """
         Adds a new data point to the sample. The values in successive calls to
         ``add`` must have the same tensor shape and size. Optionally, an
@@ -77,13 +79,14 @@ class Empirical(TorchDistribution):
         :param torch.Tensor log_weight: weight (optional) corresponding
             to the sample.
         """
-        if is_validation_enabled():
+        if self._validate_args:
             if weight is not None and log_weight is not None:
                 raise ValueError("Only one of ```weight`` or ``log_weight`` should be specified.")
             if torch.is_tensor(weight) and weight.dim() > 0:
                 raise ValueError("``weight.dim() > 0``, but weight should be a scalar.")
 
-        weight_type = torch.Tensor if value.dtype in (torch.int32, torch.int64) else value.type()
+        weight_type = value.new_empty(1).float().type() if value.dtype in (torch.int32, torch.int64) \
+            else value.type()
         # Apply default weight of 1.0.
         if log_weight is None and weight is None:
             log_weight = torch.tensor(1.0).type(weight_type).log()
@@ -113,7 +116,7 @@ class Empirical(TorchDistribution):
 
         :param torch.Tensor value: scalar or tensor value to be scored.
         """
-        if is_validation_enabled():
+        if self._validate_args:
             if value.size() != self.event_shape:
                 raise ValueError("``value.size()`` must be {}".format(self.event_shape))
         self._finalize()
@@ -124,10 +127,6 @@ class Empirical(TorchDistribution):
         idxs = torch.arange(self.sample_size)[selection_mask.min(dim=-1)[0]]
         log_probs = self._categorical.log_prob(idxs)
         return log_sum_exp(log_probs)
-
-    @property
-    def batch_shape(self):
-        return torch.Size()
 
     @property
     def event_shape(self):
--- a/pyro/distributions/empirical.py
+++ b/pyro/distributions/empirical.py
@@ -1,0 +1,158 @@
+from __future__ import absolute_import, division, print_function
+
+import numbers
+
+import math
+import torch
+
+from pyro.distributions.torch_distribution import TorchDistribution
+from pyro.distributions.torch import Categorical
+from pyro.distributions.util import copy_docs_from, is_validation_enabled
+from pyro.distributions.util import log_sum_exp
+
+
+@copy_docs_from(TorchDistribution)
+class Empirical(TorchDistribution):
+    r"""
+    Empirical distribution associated with the sampled data.
+    """
+
+    arg_constraints = {}
+
+    def __init__(self):
+        self._samples = None
+        self._log_weights = None
+        self._categorical = None
+        self._samples_buffer = []
+        self._weights_buffer = []
+        super(TorchDistribution, self).__init__()
+
+    def _append_from_buffer(self, tensor, buffer):
+        """
+        Append values from the buffer to the finalized tensor, along the
+        leftmost dimension.
+
+        :param torch.Tensor tensor: tensor containing existing values.
+        :param list buffer: list of new values.
+        :return: tensor with new values appended at the bottom.
+        """
+        buffer_tensor = torch.stack(buffer, dim=0)
+        return torch.cat([tensor, buffer_tensor], dim=0)
+
+    def _finalize(self):
+        """
+        Appends values collected in the samples/weights buffers to their
+        corresponding tensors.
+        """
+        if not self._samples_buffer:
+            return
+        self._samples = self._append_from_buffer(self._samples, self._samples_buffer)
+        self._log_weights = self._append_from_buffer(self._log_weights, self._weights_buffer)
+        self._categorical = Categorical(logits=self._log_weights)
+        # Reset buffers.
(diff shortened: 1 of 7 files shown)


pyro/distributions/empirical.py
```diff
@@ -0,0 +1,160 @@
...
+        if is_validation_enabled():
+            if weight is not None and log_weight is not None:
+                raise ValueError("Only one of ```weight`` or ``log_weight`` should be specified.")
+            if torch.is_tensor(weight) and weight.dim() > 0:
+                raise ValueError("``weight.dim() > 0``, but weight should be a scalar.")
+
+        weight_type = torch.Tensor if value.dtype in (torch.int32, torch.int64) else value.type()
+        # Apply default weight of 1.0.
+        if log_weight is None and weight is None:
+            log_weight = torch.tensor(1.0).type(weight_type).log()
```
martinjankowiak: why not just do zero directly without the expensive log?

neerajprad: :) Was thinking that it might be self documenting, but you are right. Will change.

pyro/distributions/util.py
```diff
@@ -321,6 +321,20 @@ def matrix_triangular_solve_compat(b, A, upper=True):
         return b.trtrs(A, upper=upper)[0].view(b.size())
 
 
+def log_sum_exp(tensor, scale=1.0, dim=-1):
```
fritzo: The scale arg here seems a little weird. I'd at least move it to the right of `dim`, but consider removing it.

pyro/distributions/empirical.py
```diff
@@ -0,0 +1,160 @@
...
+    @property
+    def mean(self):
+        self._finalize()
+        if self._samples.dtype in (torch.int32, torch.int64):
+            raise ValueError("Mean for discrete empirical distribution undefined. Consider converting samples " +
+                             "to ``torch.float32`` or ``torch.float64``.")
+        weights = self._log_weights
+        for _ in range(self._samples.dim() - 1):
+            weights = weights.unsqueeze(-1)
+        return (log_sum_exp(weights, scale=self._samples, dim=0) - log_sum_exp(weights, dim=0)).exp()
```
fritzo: It seems like it would be clearer to
```py
weights = (self._log_weights - log_sum_exp(self._log_weights)).exp()
weights = weights.reshape(weights.shape + (1,) * self.event_dim)
return (weights * self._samples).sum(0)
```
Then you could also remove the `scale` arg to `log_sum_exp()`

pyro/distributions/empirical.py
```diff
@@ -0,0 +1,160 @@
...
+        Number of samples that constitute the empirical distribution.
+
+        :return int: number of samples collected.
+        """
+        self._finalize()
+        if self._samples is None:
+            return 0
+        return self._samples.size(0)
+
+    def add(self, value, log_weight=None, weight=None):
```
fritzo: nit: `weight` seems more standard than `log_weight`, so I'd be less surprised if if was to the left of `log_weight`

pyro/distributions/empirical.py
```diff
@@ -0,0 +1,160 @@
...
+        :param torch.Tensor log_weight: weight (optional) corresponding
+            to the sample.
+        """
+        if is_validation_enabled():
+            if weight is not None and log_weight is not None:
+                raise ValueError("Only one of ```weight`` or ``log_weight`` should be specified.")
+            if torch.is_tensor(weight) and weight.dim() > 0:
+                raise ValueError("``weight.dim() > 0``, but weight should be a scalar.")
+
+        weight_type = torch.Tensor if value.dtype in (torch.int32, torch.int64) else value.type()
```
fritzo: This doesn't work with cuda int tensors. Consider using something that preserves CUDA and device, e.g.
```py
weight_type = value.new_empty(1).float().type() if ... else value.type()
```

pyro/distributions/empirical.py
```diff
@@ -0,0 +1,160 @@
...
+    def variance(self):
+        self._finalize()
+        if self._samples.dtype in (torch.int32, torch.int64):
+            raise ValueError("Variance for discrete empirical distribution undefined. Consider converting samples " +
+                             "to ``torch.float32`` or ``torch.float64``.")
+        deviation_squared = torch.pow(self._samples - self.mean, 2)
+        weights = self._log_weights
+        for _ in range(self._samples.dim() - 1):
+            weights = weights.unsqueeze(-1)
+        return (log_sum_exp(weights, scale=deviation_squared, dim=0) - log_sum_exp(weights, dim=0)).exp()
```
fritzo: ditto: it seems like it would be clearer to
```py
weights = (self._log_weights - log_sum_exp(self._log_weights)).exp()
weights = weights.reshape(weights.shape + (1,) * self.event_dim)
mean = (weights * self._samples).sum(0)
return (weights * (self._samples - mean).pow(2)).sum(0)
```

pyro/distributions/empirical.py
```diff
@@ -0,0 +1,160 @@
...
+
+@copy_docs_from(TorchDistribution)
+class Empirical(TorchDistribution):
+    r"""
+    Empirical distribution associated with the sampled data.
+    """
+
+    arg_constraints = {}
+
+    def __init__(self):
```
fritzo: Can  you pass through the `validate_args` arg as other `TorchDistributions` do? Then you can use the idomatic `if self._validate_args` rather than `if is_validation_enabled()`?

pyro/distributions/empirical.py
```diff
@@ -0,0 +1,160 @@
...
+from pyro.distributions.util import log_sum_exp
+
+
+@copy_docs_from(TorchDistribution)
+class Empirical(TorchDistribution):
+    r"""
+    Empirical distribution associated with the sampled data.
+    """
+
+    arg_constraints = {}
```
fritzo: Is there a sensible `.support`? Does it make most sense to set `support = constraints.real`?

pyro/distributions/empirical.py
```diff
@@ -0,0 +1,160 @@
...
+
+        # Seed the container tensors with the correct tensor types
+        if self._samples is None:
+            self._samples = value.new()
+            self._log_weights = log_weight.new()
+        # Append to the buffer list
+        self._samples_buffer.append(value)
+        self._weights_buffer.append(log_weight)
+
+    def sample(self, sample_shape=torch.Size()):
```
fritzo: (low priority comment:)

Is it sensible to define `rsample()`? If samples are differentiable, then the results should be differentiable wrt samples. The result is not differentiable wrt `weights` though. @martinjankowiak wdyt?

pyro/distributions/empirical.py
```diff
@@ -0,0 +1,160 @@
...
+        selection_mask = self._samples.eq(value).contiguous().view(self.sample_size, -1)
+        # Return -Inf if value is outside the support.
+        if not selection_mask.any():
+            return self._log_weights.new_zeros(torch.Size()).log()
+        idxs = torch.arange(self.sample_size)[selection_mask.min(dim=-1)[0]]
+        log_probs = self._categorical.log_prob(idxs)
+        return log_sum_exp(log_probs)
+
+    @property
+    def batch_shape(self):
```
fritzo: I think you can omit this property this if you pass the `batch_shape` arg to `super(..., self).__init__()`

pyro/distributions/util.py
neerajprad: The scale argument is adopted from scipy's [log_sum_exp](https://docs.scipy.org/doc/scipy/reference/generated/scipy.misc.logsumexp.html), though using it with tensorized scale was my idea. :) I suppose this will be more numerically stable than using the normalized weights and multiplying will have underflow issues - consider a low weight that is multiplied with a large scale factor.

We also have a `log_sum_exp` in `pyro.util` but using that leads to circular import issues.

fritzo: Ok. Can we move the `pyro.util` version to `pyro.distributions.util`?

pyro/distributions/empirical.py
neerajprad: See comment above. I think this will have underflow issues which can be avoided by including the scale factor. I can use the computation here directly rather than creating a generic utility, if you think a generic function is not warranted.

pyro/distributions/empirical.py
neerajprad: Will change.

pyro/distributions/empirical.py
neerajprad: Good idea, will change.

pyro/distributions/empirical.py
neerajprad: Will change.

pyro/distributions/empirical.py
neerajprad: Yeah, I suppose so.

pyro/distributions/empirical.py
neerajprad: Yes, let me change that.

pyro/distributions/empirical.py
neerajprad: I had it that way to begin with. Should I just move over the `log_sum_exp` logic here if the scale factor seems confusing?

pyro/distributions/util.py
neerajprad: Sure. I think we can also remove that in a later cleanup if we start using this in `AbstractInfer`, since that is the only place it is used.

pyro/distributions/empirical.py
eb8680: Are `enumerate_support` and `support` expected to satisfy any invariants in `TorchDistributions`?

fritzo: Some informal invariants are:
- `.enumerate_support()` should return a set of samples whose probabilities sum to 1
- arguments to `.log_prob()` should return `-inf` if outside of `.support`
- return values of `.sample()` and `.enumerate_support()` shoule lie in the `.support`

I don't know whether this is important; it just surprised me to see `.arg_constraints` without `.support`.

pyro/distributions/empirical.py
fritzo: I guess this is fine as is. It might be nice to avoid two different implementations of `log_sum_exp()` in the codebase, but we can always clean up in a follow-up PR.

pyro/distributions/empirical.py
fritzo: It's up to you. I didn't realize the scale arg was provided by scipy.

pyro/distributions/empirical.py
neerajprad: >  It might be nice to avoid two different implementations of log_sum_exp() in the codebase, but we can always clean up in a follow-up PR.

Already deleted, we are using the one from distributions now.

neerajprad:
Will update with `.enumerate_support`.

neerajprad:
@fritzo, @eb8680, @martinjankowiak - Thanks for reviewing this! There are a few things that I would like to change, and add another test, but that can wait for a subsequent PR. This should be good to merge otherwise.

pyro/distributions/empirical.py
```diff
@@ -0,0 +1,164 @@
...
+        :param torch.Tensor log_weight: weight (optional) corresponding
+            to the sample.
+        """
+        if self._validate_args:
+            if weight is not None and log_weight is not None:
+                raise ValueError("Only one of ```weight`` or ``log_weight`` should be specified.")
+            if torch.is_tensor(weight) and weight.dim() > 0:
+                raise ValueError("``weight.dim() > 0``, but weight should be a scalar.")
+
+        weight_type = value.new_empty(1).float().type() if value.dtype in (torch.int32, torch.int64) \
```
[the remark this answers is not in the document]
fritzo: good point, I don't know if there is a clean alternative other than lots of conditional logic.

Issue #930, opened 2018-03-26
Update tutorials/examples


martinjankowiak:
we're collecting some best practices for tutorial/example rewrites in the main body of this issue. stay tuned for refinements! feel free to add stuff and refine.

## Pyro stuff
- make sure you have `pyro.enable_validation(True)` and `smoke_test = ('CI' in os.environ)` somewhere at the top
- no need to specify `requires_grad=True` in `pyro.param`
- use Pyro constraint constructs, e.g. `pyro.param(..., constraint=constraints.positive)`
   see here: http://pytorch.org/docs/master/distributions.html?highlight=constraints#module-torch.distributions.constraints
- ensure proper shaping using `iarange` and `.reshape`
- for expensive iterative computations, set `num_iters = 2 if smoke_test else 1000` or similar

## PyTorch stuff (a number of these borrowed from @fehiepsi !)
@neerajprad @fehiepsi and others in the know, please check!
- everything should use `torch.tensor`; no `Variable` or `variable`
- initialize `FloatTensor`s with floats and `LongTensor`s with ints, e.g. `torch.tensor(0.)` versus `torch.tensor(0)`
- replace `.data[0]` with `.item()` to get float value of a scalar
- use `.tolist()` to convert a tensor to Python list.
- use `.detach()` to make `requires_grad=False`.
- use `.numpy()` on a non-grad non-cuda tensor to convert it to numpy array. they use the same memory. A good practice is to use `.cpu().data.numpy()` or `.cpu().detach().numpy()`.
- no need to use Pyro's utils `ng_zeros`, `ng_ones` (they no longer exist). use `torch.ones`, `torch.zeros` instead.
- use `x.new_zeros(...)` rather than `torch.zeros(...).type_as(x)` etc. Similar for `x.new_ones(...)`.
- use `x.new_empty(...)` to create a new empty tensor (same behavior with the legacy `x.new(...)`)
- use `torch.tensor(..., dtype=..., requires_grad=...)` to create a tensor with the corresponding type and `requires_grad` attribute. the previous ways `torch.FloatTensor(...)`, `torch.cuda.DoubleTensor(...)` also work.
- `b = torch.tensor(...).type_as(a)` is the same as `b = torch.tensor(..., dtype=a.dtype)`. However, it is better to use `a.new_tensor(...)` which also corrects gpu device.
- `.sum()` always return a scalar tensor.

 ## Jupyter stuff
- runnable code should be in a code box
[...]

PR #970, opened 2018-04-03
Fix bayesian regression example and re-enable testing

neerajprad:
Fixes the bayesian regression example from #953 using @colesbury's suggestion.

Pasting @colesbury's comment which explains the issue:

> The issue is that w_mu is that the `type_as` comes after the requires_grad assignment, so w_mu is no longer a leaf variable. The existing code is equivalent to:

```python
w_mu_hidden = torch.randn(1, p, requires_grad=True)  # leaf variable (optimizable)
w_mu = w_mu_hidden.type_as(data)  # backprops to of w_mu_hidden via backward type conversion
```

### DIFF
--- a/examples/bayesian_regression.py
+++ b/examples/bayesian_regression.py
@@ -73,10 +73,12 @@ def model(data):
 
 
 def guide(data):
-    w_loc = torch.randn(1, p, requires_grad=True).type_as(data)
+    w_loc = torch.randn(1, p).type_as(data)
+    w_loc.requires_grad = True
     w_log_sig = torch.tensor((-3.0 * torch.ones(1, p) + 0.05 * torch.randn(1, p)).type_as(data),
                              requires_grad=True)
-    b_loc = torch.randn(1, requires_grad=True).type_as(data)
+    b_loc = torch.randn(1).type_as(data)
+    b_loc.requires_grad = True
     b_log_sig = torch.tensor((-3.0 * torch.ones(1) + 0.05 * torch.randn(1)).type_as(data.data), requires_grad=True)
     # register learnable params in the param store
     mw_param = pyro.param("guide_mean_weight", w_loc)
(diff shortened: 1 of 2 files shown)


examples/bayesian_regression.py
```diff
@@ -73,10 +73,12 @@ def model(data):
 
 
 def guide(data):
-    w_loc = torch.randn(1, p, requires_grad=True).type_as(data)
+    w_loc = torch.randn(1, p).type_as(data)
+    w_loc.requires_grad = True
```
fritzo: Looks good, but could you explain the issue so I can avoid it in the future? I thought Adam suggested using the previous form in distributions tests.

neerajprad: I just added the comment from @colesbury to the top which best explains this! The issue is that the `.type_as` replaces the earlier `w_mu` which is no longer a leaf node in the computation graph. This behavior has only recently changed, so I'm guessing that Adam's suggestion is from an earlier time.

jpchen: i dont think we need `requires_grad` because it is wrapped with `pyro.param`

examples/bayesian_regression.py
```diff
@@ -73,10 +73,12 @@ def model(data):
...
     w_log_sig = torch.tensor((-3.0 * torch.ones(1, p) + 0.05 * torch.randn(1, p)).type_as(data),
                              requires_grad=True)
-    b_loc = torch.randn(1, requires_grad=True).type_as(data)
+    b_loc = torch.randn(1).type_as(data)
+    b_loc.requires_grad = True
```
jpchen: ditto above

examples/bayesian_regression.py
neerajprad: That'll make it cleaner. Great, I didn't realize the param store would handle that.

neerajprad:
@jpchen - Removed the `requires_grad` arg, should be good to merge.

neerajprad:
This doesn't do the updates from #930 which should be a separate PR, I guess. I just wanted to re-enable testing on this with minimal changes.

PR #971, opened 2018-04-03
Revise util / update code

fehiepsi:
This pull request makes following changes:

+ Update code as suggested in #930: use `.shape` (for `.size()`, `reshape` (for `.contiguous().view(...)`), `new_empty`, `new_tensor`, `new_zeros` (instead of `.new`)
+ Remove `torch_eye`, `torch_multinomial`, `get_probs_and_logits`, `get_clamped_probs` in distribution.util
+ Rename `is_nan`, `is_inf` to `torch_isnan`, `torch_isinf` which is more friendly to other pyro's function names such as `torch_data_sum`, `torch_backward`,...
+ Remove `torch_exp`, `reduce_to_target`, `reduce_to_shape` methods from `pyro.infer.util`
+ Move `zero_grads`, `detach_iterable` from `pyro.util` to `pyro.infer.util`
+ Run `make scrub` for tutorials to make their kernelspecs consistence.

### DIFF
--- a/pyro/distributions/distribution.py
+++ b/pyro/distributions/distribution.py
@@ -37,7 +37,7 @@ class Distribution(object):
         """
         Samples a random value (just an alias for ``.sample(*args, **kwargs)``).
 
-        For tensor distributions, the returned tensor should have the same ``.size()`` as the
+        For tensor distributions, the returned tensor should have the same ``.shape`` as the
         parameters.
 
         :return: A random value.
@@ -50,14 +50,14 @@ class Distribution(object):
         """
         Samples a random value.
 
-        For tensor distributions, the returned tensor should have the same ``.size()`` as the
+        For tensor distributions, the returned tensor should have the same ``.shape`` as the
         parameters, unless otherwise noted.
 
         :param sample_shape: the size of the iid batch to be drawn from the
             distribution.
         :type sample_shape: torch.Size
         :return: A random value or batch of random values (if parameters are
-            batched). The shape of the result should be ``self.size()``.
+            batched). The shape of the result should be ``self.shape()``.
         :rtype: torch.Tensor
         """
         raise NotImplementedError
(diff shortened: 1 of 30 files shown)


pyro/distributions/util.py
```diff
@@ -144,12 +143,12 @@ def sum_leftmost(value, dim):
         return value
     if dim >= value.dim():
         return value.sum()
-    return value.contiguous().view(-1, *value.shape[dim:]).sum(0)
+    return value.reshape(-1, *value.shape[dim:]).sum(0)
```
neerajprad: `.reshape` copies the tensor and may be more expensive than `.view`, so it may be advisable to retain the `.view` method.

pyro/distributions/distribution.py
```diff
@@ -50,14 +50,14 @@ def sample(self, *args, **kwargs):
...
-        For tensor distributions, the returned tensor should have the same ``.size()`` as the
+        For tensor distributions, the returned tensor should have the same ``.shape`` as the
         parameters, unless otherwise noted.
 
         :param sample_shape: the size of the iid batch to be drawn from the
             distribution.
         :type sample_shape: torch.Size
         :return: A random value or batch of random values (if parameters are
-            batched). The shape of the result should be ``self.size()``.
+            batched). The shape of the result should be ``self.shape()``.
```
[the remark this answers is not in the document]
fehiepsi: I guess the original author wanted to mention about this method: https://github.com/uber/pyro/blob/dev/pyro/distributions/torch_distribution.py#L49

pyro/distributions/util.py
fehiepsi: according to http://pytorch.org/docs/master/torch.html#torch.reshape, reshape will not copy when possible
I test it locally and it returns a view for a tensor

pyro/distributions/distribution.py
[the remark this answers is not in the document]
neerajprad: I missed that we have a method by this name too.

pyro/distributions/util.py
neerajprad: That's great! Verified that it is fast too. :)

PR #974, opened 2018-04-04
Add .median(), .quantiles() to ADVI implementations

### LINKED ISSUE #973: Feature request: display ADVI params
Hi,
When using ADVI guide, it is possible to print nicely a sample of the model parameters, with e.g.
print(advi.guide(y))
Pyro's param store contains advi_loc and advi_scale/advi_scale_tril, but it is not clear which parameter refers to which index of these vectors/arrays and, from what I understand, these are unrestricted params, not my models restricted params I care about.
So, I would like to request adding a functionality for a nice display of the means of location and scale of the model parameters.

Slawek

fritzo:
Resolves #973

This adds methods `.median()` and `.quantiles()` to our mean-field and multivariate-normal ADVI implementations. These are useful for diagnosing latent variable distributions. It is difficult to expose a `.mean()` or `.variance()` method because we parameterize in the unconstrained space, whereas moments make sense in the constrained space.

Note this refactors to reuse a nice `._unpack_latent()` method, which simplifies the `.guide()` method.

## Tested

- added unit tests of three different constraint types

### DIFF
--- a/pyro/infer/advi.py
+++ b/pyro/infer/advi.py
@@ -189,7 +189,7 @@ class ADVIMultivariateNormal(ADVI):
 
         :param quantiles: A list of requested quantiles between 0 and 1.
         :type quantiles: torch.Tensor or list
-        :return: A dictionary mapping sample site name to median tensor.
+        :return: A dictionary mapping sample site name to a list of quantile values.
         :rtype: dict
         """
         loc = pyro.param("advi_loc")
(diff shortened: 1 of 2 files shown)


fritzo:
At request of @slaweku, I'll also implement `.quantiles()` in this PR...

fritzo:
> do there need to be 2 implementations of median and quantiles?

While we could combine these, I thought it would be easier to maintain with less abstraction. The median methods are actually identical, but they are a tiny three lines of code. Also [rule-of-three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) suggests waiting until we have a third use case before building abstraction.

fritzo:
ready to merge

PR #976, opened 2018-04-05
Update some tutorials to use new PyTorch semantics

neerajprad:
As per #930, updates the following examples:
 - vae/vae_comparison
 - bayesian_regression


fehiepsi:
@neerajprad Related to this pull request, I observed that there are still some `type_as` statements in distributions module. They are minor so we can incorporate the change in any new pull request. :D

PR #978, opened 2018-04-06
Add BO tutorial

fehiepsi:
Resubmit to clean #883 (after merge dev and add a note about "disable validation").

Issue: To have consistency with Pyro, I install Pytorch version #1807bac (previously I used f8270c0 which was newer) but it gives different results. Investigating...


fehiepsi:
@eb8680 Yes!

It is strange to me that two pytorch commits give different results:
[#1807bac](http://nbviewer.jupyter.org/github/fehiepsi/pyro/blob/a0c833ed34abdb5aad263fc2a89902a46edad7a6/tutorial/source/bo.ipynb)
and
[#f8270c0 ](http://nbviewer.jupyter.org/github/fehiepsi/pyro/blob/e69cde879438ad223ea7e9bac9cd3b149e0fa832/tutorial/source/bo.ipynb)

Things are separated at the point x6, where optimizers found different local minima. Let me find down the reason to prevent some problems like this happen in the future.

fehiepsi:
@eb8680 I checked out some pytorch commits and observed that the results changed after speed-up commits from pytorch such as https://github.com/pytorch/pytorch/pull/6078 and https://github.com/pytorch/pytorch/pull/6026. I think that this situation (pytorch changed how it computed fundamental functions such as sum, exp,... which in turns made computational precision using these functions a little bit different) is unavoidable so I removed the line mentioned about x7/x6. If you think that it is fine then this pull request is ready to merge. Definitely, we should rerun the notebook when python 0.4 is released.

PR #986, opened 2018-04-07
refresh svi part ii tutorial

martinjankowiak:
i kept the changes pretty minimal. in particular i added a note at the end about the need to be careful about tensor shapes. @fritzo can you please take a look at this at some point and let me know if you think the changes are sufficient (and that everything is correct)? in particular do we need to readjust the balance of (or just repeat) some of the material between this tutorial and the tensor shapes tutorial?

### DIFF
--- a/tutorial/source/svi_part_ii.ipynb
+++ b/tutorial/source/svi_part_ii.ipynb
@@ -21,50 +21,30 @@
     "\n",
     "### `irange`\n",
     "\n",
-    "Let's return to the example we used in the [previous tutorial](svi_part_i.html). For convenience let's replicate the main logic of `model` here:"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
+    "Let's return to the example we used in the [previous tutorial](svi_part_i.html). For convenience let's replicate the main logic of `model` here:\n",
+    "\n",
+    "```python\n",
     "def model(data):\n",
     "    # sample f from the beta prior\n",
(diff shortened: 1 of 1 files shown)


fritzo:
@martinjankowiak I'm happy with how things are currently separated:
- introduce and motivate `irange` and `iarange` with subsampling in svi_part_ii
- provide reference material on complex uses of `iarange` in tensor_shapes

We could use some more material early on to alert users to Pyro 0.2's stricter shaping semantics. I think even in svi_part_i or svi_part_ii we could use `.reshape([n])` and `.reshape(extra_event_dims=1)` once or twice and mention that reshaping is important. Then we can point users to the tensor shape tutorial. I feel like the tensor shape stuff is less important than subsampling and should therefore come later.

fritzo:
> is there anything we're missing entirely

Do we have an explanation of `iarange` without subsampling? E.g. "`iarange` for discrete random choices can reduce variance inside SVI". In modern Pyro we could show the difference between independent mixture model observation
```py
# Rao-Blackwellize for low-variance.
with pyro.iarange("foo", len(data)):
    components = pyro.sample("x", dist.Categorical(probs))
    pyro.sample("obs", dist.Normal(locs[components], scale),
                obs=data)
```
and the higher-variance version
```py
# Naive high-variance score function estimator.
components = pyro.sample("x", dist.Categorical(probs).reshape(extra_event_dims=1))
pyro.sample("obs", dist.Normal(locs[components], scale).reshape(extra_event_dims=1),
            obs=data)
```
This example could live in either the svi tutorials or at the beginning of the tensor_shape tutorial.

martinjankowiak:
ready to merge?

PR #987, opened 2018-04-08
Use GP to classify MNIST dataset

fehiepsi:
This pull request is my attempt to use GP module for classifying MNIST. The accuracy is quite low (50%) so it must be improved (need ideas). Along with the notebook, I
+ Use whitening, which helps optimization ([reference](https://papers.nips.cc/paper/5875-mcmc-for-variationally-sparse-gaussian-processes))
+ Use minibatch for svgp (by modifying and using `poutine.scale`, though I am not sure if I use it correctly)
+ Use CNN to make deep kernel ([deep kernel learning](https://papers.nips.cc/paper/6426-stochastic-variational-deep-kernel-learning))

Blocked by https://github.com/pytorch/pytorch/pull/6035.


fehiepsi:
@ysaatchi I have tried various learning rate, num_inducing, batch_size, and various deep structures but cannot improve the score. Would you mind giving me some ideas so I can try to improve the result?

eb8680:
@fehiepsi have you tried a smaller dataset, like maybe something from UCI?

fehiepsi:
@eb8680 Ah yes, I havent tried UCI datasets. I agree that we should try the easy case first. By the way, this pull request is mainly for discussion because I observed that some persons at Uber worked on this direction (deep kernel learning) so I hope to learn from them to improve the GP module.

eb8680:
Since this is meant to be a tutorial introducing GP classification, I'd recommend sticking with a simpler dataset and a correspondingly simpler model.  Kernel learning is an advanced feature that you could write a whole separate tutorial about.

fehiepsi:
@eb8680 Yes, I will make a tutorial on classification with a simpler dataset on another pull request.

ysaatchi:
@fehiepsi I agree with @eb8680 for now! Getting GPs to work well with MNIST is probably worth an entirely separate pull request -- my guess is the optimization of the model is not working well at the moment. You need to iterate between optimizing final GP layer hyperparameters and then freeze them and run SGD for the CNN. So it is quite involved.

fehiepsi:
@ysaatchi It seems reasonable, the freeze technique also appears in the sv-dkl paper. Do you know any attempt which trained deep kernels end-to-end?

ysaatchi:
@fehiepsi it still is end-to-end, you can just view it as a block (stochastic) gradient descent instead of regular SGD. I am unaware of anything which works that uses just regular SGD.

fehiepsi:
Just a note that the test failed due to #999.

fehiepsi:
I have reverted changes in poutine.scale and param_store. Things should be good to review now.

eb8680:
@fehiepsi there are a bunch of other changes in this PR, can you add explanations for them?

fehiepsi:
Sure. All the changes except examples are:
+ Support whiten for variational models and add tests for it
+ Allow y=None in likelihood (because when testing, I want to use likelihood to give prediction)
+ Allow input data X be an arbitrary dimension tensor. The fact that X just interact with kernels make it unnecessary to check its validation in model (instead, we still check its validation in kernels). This is helpful when making deep kernel (e.g. images  are 4D tensors)
+ Some nits with `type_as` at some files in `distributions`

examples/contrib/gp/sv-dkl.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def test(args, test_loader, gpmodel):
+    correct = 0
+    for data, target in test_loader:
+        if args.cuda:
+            data, target = data.cuda(), target.cuda()
+        f_loc, f_var = gpmodel(data)
+        pred = gpmodel.likelihood(f_loc, f_var)
+        correct += pred.eq(target).long().cpu().sum()
```
eb8680: Add some comments here explaining how/why you're making predictions this way

examples/contrib/gp/sv-dkl.py
```diff
@@ -0,0 +1,137 @@
...
+                                       batch_size=args.batch_size,
+                                       dataset_transforms=[transforms.Normalize((0.1307,), (0.3081,))],
+                                       is_training_set=False,
+                                       shuffle=True)
+
+    cnn = CNN().cuda() if args.cuda else CNN()
+
+    # helper to inject CNN's parameters to gpmodel
+    def cnn_fn(x):
+        return pyro.module("CNN", cnn)(x)
```
eb8680: Add comment explaining why you wrapped `pyro.module` this way

examples/contrib/gp/sv-dkl.py
```diff
@@ -0,0 +1,137 @@
...
+    # create deep kernel by warping RBF with cnn
+    kernel = gp.kernels.RBF(input_dim=10, lengthscale=torch.ones(10)).warp(iwarping_fn=cnn_fn)
```
eb8680: Add comment explaining how this kernel works and what it's doing

examples/contrib/gp/sv-dkl.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # init inducing points (taken randomly from dataset)
+    Xu = next(iter(train_loader))[0][:args.num_inducing]
+    likelihood = gp.likelihoods.MultiClass(num_classes=10)
+    gpmodel = gp.models.SparseVariationalGP(X=Xu, y=None, kernel=kernel, Xu=Xu,
+                                            likelihood=likelihood, latent_shape=torch.Size([10]),
```
eb8680: Add comment explaining why you're using this likelihood and this model, how `latent_shape` is related to `likelihood`, and what whitening is and why you're using it here

eb8680:
LGTM when tests pass

fehiepsi:
@fritzo I have revised the order of likelihood if/else as you suggested and added some small tests to check shapes of outputs of `forward`. This should be good after Travis tests pass.

PR #989, opened 2018-04-08
refresh svi iii tutorial

martinjankowiak:
- added a simple rao-blackwellization example based on fritz's little snippet
- update semantics of and made sure final code snippet runs as expected
- added smoke test logic to final snippet
- note that i now have to use fakes in final snippet
- if we're happy with this refresh i should go remove ` num_particles=7 ` from svi part i (since that is not using reparameterized gradients)


tutorial/source/svi_part_iii.ipynb
```diff
@@ -343,23 +383,21 @@
...
    "metadata": {},
    "source": [
-    "```\n",
     "Doing inference with use_decaying_avg_baseline=True\n",
-    "...........\n",
-    "Did 2070 steps of inference.\n",
-    "Final absolute errors for the two variational parameters (in log space) were 0.0500 & 0.0443 \n",
+    "....................\n",
+    "Did 1932 steps of inference.\n",
+    "Final absolute errors for the two variational parameters were 0.7997 & 0.0800\n",
```
fritzo: Can you confirm that this 16x increase in absolute error is ok?

tutorial/source/svi_part_iii.ipynb
```diff
@@ -282,52 +323,51 @@
...
+    "        for k in range(self.max_steps):\n",
     "            svi.step(use_decaying_avg_baseline)\n",
     "            if k % 100 == 0:\n",
     "                print('.', end='')\n",
     "                sys.stdout.flush()\n",
     "\n",
     "            # compute the distance to the parameters of the true posterior\n",
-    "            alpha_error = param_abs_error(\"log_alpha_q\", self.log_alpha_n)\n",
-    "            beta_error = param_abs_error(\"log_beta_q\", self.log_beta_n)\n",
+    "            alpha_error = param_abs_error(\"alpha_q\", self.alpha_n)\n",
```
fritzo: Can you confirm that this stopping criterion is still ok? I had to update some of the stopping criteria in tests after we switched from log-coordinates to linear coordinates.

tutorial/source/svi_part_iii.ipynb
martinjankowiak: it's ok. it only appears like that because it's no longer in log space

tutorial/source/svi_part_iii.ipynb
martinjankowiak: yes it's ok

Issue #981, opened 2018-04-07
Remove circular dependencies before 0.2 release


fritzo:
Proposed module organization to avoid circular dependencies:
1. avoid importing `a/b/__init__.py` from any sibling module `a/b/c.py`
2. try to maintain the following low-to-high level import order:
  `distributions < param < { poutine, ops, nn } < pyro < infer < contrib`
3. avoid explicit mutual imports between pairs of modules.

eb8680:
Is there any way to check for circular dependencies automatically?  Can we add that to our CI suite?

fritzo:
Pycycle seems to say we're clean:
```sh
$ pycycle --source pyro/
Target source provided: /Users/fritzobermeyer/github/uber/pyro/pyro
Project successfully transformed to AST, checking imports for cycles..
No worries, no cycles here!
If you think some cycle was missed, please open an Issue on Github.
Finished.
```

eb8680:
Does that mean this can be closed?

PR #991, opened 2018-04-09
Refactoring imports for a few Pyro modules

neerajprad:
Addresses #981 for `nn`, `distributions` and `optim` modules.

Additionally, it does the following.
 - Changes `__init__.py` for modules so that:
   - The classes/functions that form the public API are explicitly declared in an `__all__` variable.
   - All remaining code is migrated to other classes. This is to prevent namespace pollution (and unwitting circular imports). e.g. The following (earlier working) code will throw an error now: `from pyro.optim import torch`. This separation makes it easier to generate docs too.
 - Since we are using `__all__`, the linter is happy. As such `noqa` is removed from `__init__.py`, and flake8 checks these files for unused imports.
 - Uses absolute imports in all files.

### DIFF
--- a/docs/source/pyro.optim.txt
+++ b/docs/source/pyro.optim.txt
@@ -1,23 +1,15 @@
-PyroOptim
----------
+Pyro Optimizers
+---------------
 
 .. automodule:: pyro.optim.optim
     :members:
     :undoc-members:
     :show-inheritance:
 
-ClippedAdam
-------------
+Pytorch Optimizers
+------------------
 
-.. automodule:: pyro.optim.clipped_adam
-    :members:
-    :undoc-members:
-    :show-inheritance:
-
-Pyro Optimizers
----------------
-
-.. automodule:: pyro.optim
+.. automodule:: pyro.optim.pytorch_optimizers
     :members:
     :undoc-members:
     :show-inheritance:
--- a/pyro/contrib/gp/__init__.py
+++ b/pyro/contrib/gp/__init__.py
@@ -1,9 +1,14 @@
 from __future__ import absolute_import, division, print_function
 
 
-from . import kernels
-from . import likelihoods
-from . import models
-from . import util
+from pyro.contrib.gp import kernels
+from pyro.contrib.gp import likelihoods
+from pyro.contrib.gp import models
+from pyro.contrib.gp import util
 
-# flake8: noqa
+__all__ = [
+    "kernels",
+    "likelihoods",
+    "models",
+    "util"
+]
--- a/pyro/contrib/gp/__init__.py
+++ b/pyro/contrib/gp/__init__.py
@@ -10,5 +10,5 @@ __all__ = [
     "kernels",
     "likelihoods",
     "models",
-    "util"
+    "util",
 ]
--- a/pyro/contrib/gp/kernels/__init__.py
+++ b/pyro/contrib/gp/kernels/__init__.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import, division, print_function
 
-
 from pyro.contrib.gp.kernels.brownian import Brownian
 from pyro.contrib.gp.kernels.dot_product import DotProduct, Linear, Polynomial
 from pyro.contrib.gp.kernels.isotropic import (Exponential, Isotropy, Matern32, Matern52,
@@ -12,25 +11,25 @@ from pyro.contrib.gp.kernels.static import Constant, WhiteNoise
 
 __all__ = [
     "Brownian",
+    "Cosine",
+    "Combination",
+    "Constant",
     "DotProduct",
-    "Linear",
-    "Polynomial",
+    "Exponent",
     "Exponential",
     "Isotropy",
+    "Kernel",
+    "Linear",
     "Matern32",
     "Matern52",
+    "Periodic",
+    "Polynomial",
+    "Product",
     "RationalQuadratic",
     "RBF",
-    "Combination",
-    "Exponent",
-    "Kernel",
-    "Product",
     "Sum",
     "Transforming",
     "VerticalScaling",
     "Warping",
-    "Cosine",
-    "Periodic",
-    "Constant",
-    "WhiteNoise"
+    "WhiteNoise",
 ]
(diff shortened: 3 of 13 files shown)


docs/source/pyro.optim.txt
```diff
@@ -1,23 +1,15 @@
...
----------
+Pyro Optimizers
+---------------
 
 .. automodule:: pyro.optim.optim
     :members:
     :undoc-members:
     :show-inheritance:
 
-ClippedAdam
```
jpchen: is there a reason not to combine these all under pyro optimizers? all the pytorch ones are wrapped like the custom ones.

pyro/distributions/__init__.py
```diff
@@ -12,7 +8,25 @@
 from pyro.distributions.omt_mvn import OMTMultivariateNormal
 from pyro.distributions.rejector import Rejector
 from pyro.distributions.sparse_mvn import SparseMultivariateNormal
+from pyro.distributions.torch import *  # noqa F403
+from pyro.distributions.torch import __all__ as torch_dists
```
jpchen: isnt `__all__` more inclusive than `*`? im a bit confused what this and the previous line is doing

pyro/optim/pytorch_optimizers.py
```diff
@@ -0,0 +1,21 @@
+import torch
```
jpchen: same comment as above. for the user, is there a purpose for distinguishing between pyro wrapped pt optimizers and custom ones

docs/source/pyro.optim.txt
fritzo: @neerajprad if you refactor optimizers as @jpchen  suggests, could you please check that docs links still work. it's a very delicate setup to link all the docs together

pyro/contrib/gp/__init__.py
```diff
@@ -1,9 +1,14 @@
 from __future__ import absolute_import, division, print_function
 
 
-from . import kernels
-from . import likelihoods
-from . import models
-from . import util
+from pyro.contrib.gp import kernels
```
fritzo: :+1: hooray for fully qualified names!

pyro/contrib/gp/kernels/__init__.py
```diff
@@ -1,12 +1,36 @@
...
+    "Kernel",
+    "Product",
+    "Sum",
+    "Transforming",
+    "VerticalScaling",
+    "Warping",
+    "Cosine",
+    "Periodic",
+    "Constant",
+    "WhiteNoise"
```
fritzo: nit: add trailing comma to all lines, then sort this list.

pyro/optim/pytorch_optimizers.py
neerajprad: Not really. I suppose both of these should be available anyhow through the wrapping below, so we can just remove these.

pyro/optim/__init__.py
```diff
@@ -1,40 +1,13 @@
 from __future__ import absolute_import, division, print_function
 
-import torch
+from pyro.optim.optim import PyroOptim, AdagradRMSProp, ClippedAdam
+from pyro.optim.pytorch_optimizers import *  # noqa F403
+from pyro.optim.pytorch_optimizers import __all__ as pytorch_optims
```
fritzo: If we're going to use this pattern often, it might be clearer to
```py
import pyro.optim.pytorch_optimizers

...
__all__.extend(pyro.optim.pytorch_optimizers.__all__)
```

pyro/distributions/__init__.py
neerajprad: The reason why you need the `import *` is to have the torch distributions in the local namespace. Just adding to the `__all__` isn't sufficient- it basically will add a string, but it has no information where to look for the class/function corresponding to the name.

docs/source/pyro.optim.txt
neerajprad: >  if you refactor optimizers as @jpchen suggests, could you please check that docs links still work

Will do. I checked it with this PR, and will verify after removing the two classes again.

pyro/optim/pytorch_optimizers.py
fritzo: One reason is that the below snippet correctly generates links to PyTorch docs and will not correctly generate links to Pyro docs. It would take additional logic to be able to generate links for both libraries, including work-arounds for PyTorch's convention of locating docs elsewhere than `type(x).__module__`.

pyro/distributions/__init__.py
jpchen: maybe i wasnt clear, i dont mean the `__all__ = ...` below, i mean between `from class import *` and `from class import __all__`, the latter should import everything the former does?

pyro/optim/pytorch_optimizers.py
neerajprad: @fritzo - Our wrappers for `ClippedAdam` and `AdagradRMSProp` are exactly the same as the other wrappers below, and the docstring isn't really doing much in the way of explanation. Just removing these and linking to the original PyTorch implementations like below should be self-explanatory, I think. Let me know.

fritzo: That's fine, just please preserve the links to upstream docs, as I find them useful

martinjankowiak:
maybe im missing something they don't have pytorch versions?

pyro/optim/__init__.py
neerajprad: That doesn't work unfortunately:
`AttributeError: 'module' object has no attribute 'optim'`

The problem is that `optim` hasn't finished loading in the `pyro` namespace. When we use `from ..`, we only load the end file in `sys.modules`. I don't know if there is a better way of doing this though.

pyro/optim/pytorch_optimizers.py
neerajprad: @jpchen - Sorry, I was mistaken to think that the two custom Pyro optimizers were basically wrapping over PyTorch. I think a lot of our code in general is organized in a way to facilitate doc generation; so changing that seems like a bigger hack than it is worth at this point. It is also the case with distributions where the PyTorch distributions are distinguished from Pyro ones like `Delta`. We can revisit this organization if it doesn't make too much sense for documentation purposes, but we should do that in a separate issue and PR.

fritzo:
ready to merge?

neerajprad:
Thanks for the review, @jpchen, @fritzo. Yes, this is good to merge. Will handle other modules shortly.

PR #992, opened 2018-04-10
Update contrib.named examples

fritzo:
Addresses #930

### DIFF
--- a/examples/contrib/named/tree_data.py
+++ b/examples/contrib/named/tree_data.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function
 import argparse
 
 import torch
+from torch.distributions import constraints
 
 import pyro
 import pyro.distributions as dist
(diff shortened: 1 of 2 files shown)


examples/contrib/named/tree_data.py
```diff
@@ -87,10 +91,12 @@ def main(args):
...
 
     print('Step\tLoss')
+    loss = 0.0
     for step in range(args.num_epochs):
-        if step % 100 == 0:
-            loss = inference.step(data)
+        loss += inference.step(data)
+        if step and step % 10 == 0:
             print('{}\t{:0.5g}'.format(step, loss))
+            loss = 0.0
```
jpchen: i havent looked at this example with much detail, but why is the loss being accumulated rather than set (as it was previously)?

fritzo: I'm accumulating it until I print it, then reset the accumulator. This is to overcome noise in the loss. Previous to this PR the loss looked non-monotonic.

fritzo: (also the old logic is totally wrong: most iterations didn't even .step() 😕 )

jpchen:
@fritzo good to merge?

fritzo:
ready to merge

PR #993, opened 2018-04-10
Move example utils into Pyro pip package

### LINKED ISSUE #621: Include example utils in pip-installed package
**Goal:** All examples should be runnable as stand alone python files after `pip install pyro-ppl`.

## Why?

Beginning users often start using Pyro by forking existing examples and modifying them. Currently Pyro's examples include helpers `examples/util.py` and `examples/utils/*`. Since these helpers are not part of the pip-installed package, users are forced to either (1) fork examples in Pyro's `examples/` directory and use Pyro's version control (not ideal for e.g. class projects), or (2) copy all of `pyro/examples/utils/` into their own projects.

## How?

This issue proposes to move helper code (for plotting and data loading) into the main Pyro package. This could be a single package `pyro.contrib.examples` or possibly multiple packages `pyro.contrib.visualization` + `pyro.contrib.datasets`. These new packages would have weaker guarantees for backwards compatibility, so `pyro.contrib` seems like a reasonable home.

fritzo:
Resolves #621 

This moves all the examples/util* into pyro.contrib.examples.

## Why?

The first thing new Pyro users do is to try to replicate our examples. Before this PR users need to clone the Pyro repo and replicate those examples in the examples/directory, because that's where our example utils live. After this PR users will be able to copy most of our examples and import directly from pyro.contrib.examples. (Note that some complex examples like air still require forking an entire directory tree).


fritzo:
> theres an image committed

Thanks, I force-pushed a fix.

fritzo:
ready to merge as soon as tests pass. it should be easy to move this around contrib in a follow-up PR if anyone has good ideas about organization

jpchen:
> Some of these functions, e.g. those in mnist_cached (e.g. get_ss_indices_per_class), are too specific

yeah ss-vae needs to be rewritten.. now that you point it out, most of those functions should live directly in the script kind of like how dmm and air have their own helper functions. a lot of the plotting code looks like it can be shared with vae as well. the documentation there is not standardized or formatted, and there are debugger methods in there (`print_distribution_labels`) that shouldnt be. we need to rewrite this example altogether

neerajprad:
@jpchen, @fritzo - If the goal is to make the examples runnable with only a pip install, we could just use absolute imports in the examples `from examples.utils.mnist_cached import MNISTCached as MNIST`. That said, I agree that there is value to having some generic data science utilities inside core. I am fine with merging this, but I think we may be conflating a few issues here, so lets discuss in person tomorrow.

neerajprad:
Discussed with @fritzo. We plan to have some common generic utils in contrib, and put the examples specific utils into their own folder in `examples`. I will update the PR soon.

setup.py
```diff
@@ -65,7 +65,7 @@
     version=version,
     description='A Python library for probabilistic modeling and inference',
     long_description=long_description,
-    packages=find_packages(exclude=('tests*',)),
+    packages=['pyro'],
```
neerajprad: Removing `profiler` and `examples` from the pip package.

neerajprad:
@fritzo - I pushed the following changes:
 - Moved all vae examples into their own folder, along with VAE specific utils. Some of these utilities use sklearn for tsne and would anyways not work in core.
 - Removed profiler and examples from the setup script.
 - Remove `__init__.py` from the top examples folder. Sub-folders, i.e. specific examples can have inits like `vae/utils`.
 - Kept the data loader in the `contrib.examples` directory.

jpchen:
this looks fine to me, is it ready for review?

neerajprad:
@jpchen - yes, sorry forgot to remove the WIP flag.

examples/vae/utils/mnist_cached.py
```diff
@@ -218,3 +223,18 @@ def setup_data_loaders(dataset, use_cuda, batch_size, sup_num=None, root='./data
...
+    try:
+        os.makedirs(path)
+    except OSError as exc:  # Python >2.5
+        if exc.errno == errno.EEXIST and os.path.isdir(path):
+            pass
+        else:
+            raise
+
+
+EXAMPLE_DIR = os.path.dirname(os.path.abspath(os.path.join(__file__, os.pardir)))
```
fritzo: Is there an advantage to `os.pardir` over repeated `os.path.dirname(os.path.dirname(-))`?

PR #994, opened 2018-04-10
Enable HMC tests on constrained distributions with step size adaptation

neerajprad:
Small bug-fix so that when `delta_energy` is `NaN` acceptance probability is `0.0` (clamping gives `1.0`) which results in huge step sizes during adaptation, and invalid results. I had to disable validation in the tests because `NaN` is an expected result in this case (see https://github.com/probtorch/pytorch/issues/145 for some discussion on this), so it is okay to disable warnings and run these tests without validation.

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -119,7 +119,11 @@ class HMC(TraceKernel):
             z, r, self._potential_energy, step_size)
         energy_new = potential_energy + self._kinetic_energy(r_new)
         delta_energy = energy_new - energy_current
-        # direction=1 means keep increasing step_size, otherwise decreasing step_size
+        # direction=1 means keep increasing step_size, otherwise decreasing step_size.
+        # Note that the direction is -1 if delta_energy is `NaN` which may be the
+        # case for a diverging trajectory (e.g. in the case of evaluating log prob
+        # of a value simulated using a large step size for a constrained sample
+        # site).
         direction = 1 if target_accept_logprob < -delta_energy else -1
 
         # define scale for step_size: 2 for increasing, 1/2 for decreasing
@@ -210,6 +214,10 @@ class HMC(TraceKernel):
             z = z_new
 
         if self.adapt_step_size:
+            # Set accept prob to 0.0 if delta_energy is `NaN` which could be the case
+            # which may be the case for a diverging trajectory (e.g. in the case of
+            # evaluating log prob of a value simulated using a large step size for
+            # a constrained sample site).
             accept_prob = (-delta_energy).exp().clamp(max=1).item() if not torch_isnan(delta_energy) \
                 else torch.tensor(0.0)
             self._adapt_step_size(accept_prob)
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -218,8 +218,10 @@ class HMC(TraceKernel):
             # which may be the case for a diverging trajectory (e.g. in the case of
             # evaluating log prob of a value simulated using a large step size for
             # a constrained sample site).
-            accept_prob = (-delta_energy).exp().clamp(max=1).item() if not torch_isnan(delta_energy) \
-                else torch.tensor(0.0)
+            if torch_isnan(delta_energy):
+                accept_prob = delta_energy.new_tensor(0.0)
+            else:
+                accept_prob = (-delta_energy).exp().clamp(max=1).item()
             self._adapt_step_size(accept_prob)
 
         self._t += 1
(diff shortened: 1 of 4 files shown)


pyro/infer/mcmc/nuts.py
```diff
@@ -102,7 +103,8 @@ def _build_basetree(self, z, r, z_grads, log_slice, direction, energy_current):
         tree_size = 1 if sliced_energy <= 0 else 0
         diverging = sliced_energy >= self._max_sliced_energy
         delta_energy = energy_new - energy_current
-        accept_prob = (-delta_energy).exp().clamp(max=1)
+        accept_prob = (-delta_energy).exp().clamp(max=1) if not torch_isnan(delta_energy) \
+            else torch.tensor(0.0)
```
eb8680: Add a comment explaining your reasoning?

pyro/infer/mcmc/hmc.py
```diff
@@ -210,7 +210,8 @@ def sample(self, trace):
             z = z_new
 
         if self.adapt_step_size:
-            accept_prob = (-delta_energy).exp().clamp(max=1).item()
+            accept_prob = (-delta_energy).exp().clamp(max=1).item() if not torch_isnan(delta_energy) \
+                else torch.tensor(0.0)
```
fehiepsi: Just a note if we put comments related to `nan`: I think that we should also check the logic of `direction` in `_find_reasonable_step_size`. When encounter `nan`, we should reduce step_size (which is how the current behavior works). So it would be good if we put a comment there too.

pyro/infer/mcmc/nuts.py
```diff
@@ -102,7 +103,8 @@ def _build_basetree(self, z, r, z_grads, log_slice, direction, energy_current):
         tree_size = 1 if sliced_energy <= 0 else 0
         diverging = sliced_energy >= self._max_sliced_energy
```
fehiepsi: I think that we should change this to `diverging = not (sliced_energy < self._max_sliced_energy)`

pyro/infer/mcmc/hmc.py
neerajprad: I think that logic should be fine - it will set direction = -1 if we encounter `NaN` which is what we want:
```
direction = 1 if target_accept_logprob < -delta_energy else -1
```

Will add a comment there.

fehiepsi: Yes, that logic is fine. I just wanted to mention about some relevant issues, which adding comments is especially helpful.

pyro/infer/mcmc/hmc.py
```diff
@@ -210,7 +214,12 @@ def sample(self, trace):
             z = z_new
 
         if self.adapt_step_size:
-            accept_prob = (-delta_energy).exp().clamp(max=1).item()
+            # Set accept prob to 0.0 if delta_energy is `NaN` which could be the case
+            # which may be the case for a diverging trajectory (e.g. in the case of
+            # evaluating log prob of a value simulated using a large step size for
+            # a constrained sample site).
+            accept_prob = (-delta_energy).exp().clamp(max=1).item() if not torch_isnan(delta_energy) \
+                else torch.tensor(0.0)
```
fritzo: This will not work with GPUs. Instead use `.new_zeros()` or `.new_tensor()`
```py
if torch_isnan(...):
    accept_prob = delta_energy.new_tensor(0.0)
else:
    accept_prob  = ...
```

pyro/infer/mcmc/nuts.py
```diff
@@ -100,9 +101,17 @@ def _build_basetree(self, z, r, z_grads, log_slice, direction, energy_current):
...
-        diverging = sliced_energy >= self._max_sliced_energy
-        delta_energy = energy_new - energy_current
-        accept_prob = (-delta_energy).exp().clamp(max=1)
+        # Special case: Set diverging to True and accept prob to 0 if the
+        # diverging trajectory returns `NaN` energy (e.g. in the case of
+        # evaluating log prob of a value simulated using a large step size
+        # for a constrained sample site).
+        if torch_isnan(energy_new):
+            diverging = True
+            accept_prob = torch.tensor(0.0)
```
fritzo: ditto: prefer `.new_tensor()` or `.new_zeros()`

pyro/infer/mcmc/nuts.py
```diff
@@ -100,9 +101,17 @@ def _build_basetree(self, z, r, z_grads, log_slice, direction, energy_current):
...
+        else:
+            diverging = sliced_energy >= self._max_sliced_energy
```
fritzo: nit: I find it slightly easier to read with parantheses:
```py
diverging = (sliced_energy >= self._max_sliced_energy)
```

Issue #995, opened 2018-04-10
Learn pyro.module's parameters outside model/guide


fehiepsi:
At the current state, `pyro.optim` just learns parameters of an nn.Module if it is registered by `pyro.module` inside model/guide pair. It would be nice if we can do something like this:

```
def Net(nn.Module):
    ...

net = pyro.module("net", Net())
def model(x, y):
    o = net(x)
    pyro.sample("y", dist.Normal(o, 1), obs=y)

def guide(x, y):
    pass

optimizer = optim.Adam({})
svi = SVI(model, guide, optimizer, "ELBO")
```

At the current state, we have to do as follows
```
net = Net()
def model(x, y):
    pyro_net = pyro.module("net", net)
    o = pyro_net(x)
    ...
```

eb8680:
Hmm, I'm not sure about having an inference algorithm update parameters that never appear in `param` statements in its model.  In the example snippet, why not just lift the `Module` with `lambda`?: `lambda x: pyro.module("net", net)(x)`

fehiepsi:
This is useful when we build a model which does not know ahead its name and its type. Of course, we can use `if else` to handle function/nn.Module but don't you think that it is too verbose?

eb8680:
> This is useful when we build a model which does not know ahead its name and its type

But in that case you should be able to use the `lambda` pattern I suggested, no `if`/`else` required

fehiepsi:
Yeah, you save my day! :D I spent a lot of time to debug codes before recognizing that svi does not learn params outside of its scope and ended with that pull request... Thanks!

PineappleLee:
If you lift the Module with lambda?: lambda x: pyro.module("net", net)(x), the program would thought it a expression, it is not a Module anymore, so the optimimizer couldn't learns parameters.

PineappleLee:
If you lift the Module with lambda, like lambda x: pyro.module("net", net)(x), the program would thought it a expression, it is not a Module anymore, so the optimimizer couldn't learns parameters.

PR #996, opened 2018-04-10
support train parameters outside model/guide pair

fehiepsi:
I made a small change to address https://github.com/uber/pyro/issues/995. After this, we can register an nn.Module outside of model/guide pair and still be able to learn its parameters by using `pyro.get_params_store().mark_params_active(params)`. It also fixes a bug when `params` is a generator.

### DIFF
--- a/pyro/infer/trace_elbo.py
+++ b/pyro/infer/trace_elbo.py
@@ -95,6 +95,7 @@ class Trace_ELBO(ELBO):
         Performs backward on the latter. Num_particle many samples are used to form the estimators.
         """
         elbo = 0.0
+        current_active_params = pyro.get_param_store().get_active_params()
         # grab a trace from the generator
         for model_trace, guide_trace in self._get_traces(model, guide, *args, **kwargs):
             elbo_particle = 0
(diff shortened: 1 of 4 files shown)


martinjankowiak:
i'm not sure if this is a great idea. for one thing it's not very idiomatic. it also has the consequence that the params will no longer appear in traces. do you have a use case where it's somehow clunky/inconvenient to have a `pyro.module` statement inside the model/guide?

fehiepsi:
@martinjankowiak I think that it is not necessary when we build a model from scratch. However, if we build a model which requires a function (any callable object) as an input, then it is necessary. That function can be an nn.Module or not. In case it is an nn.Module, we would like to learn its parameters so there should be a mechanism to register its parameters to pyro.optim.

An example of it is at https://github.com/fehiepsi/pyro/blob/825794e20ba8017db2b281911e013eb00e2162e4/examples/contrib/gp/sv-dkl.py#L92.

eb8680:
@fehiepsi I'm not sure I understand the problem in your example.  Shouldn't the warping kernel expose the parameters of the neural network automatically, along with the other kernel hyperparameters?

fehiepsi:
@eb8680 It won't expose unless we make that neural network a subclass of Parameterize (this class handles Parameters by using pyro.param, pyro.sample calls).

fehiepsi:
Closed thanks to @eb8680 elegant solution at #995.

Issue #999, opened 2018-04-10
Consistence pyro.poutine.scale


fehiepsi:
It would be nice if we support the old version of `poutine.scale` which takes arguments: a stochastic function and a scale factor. The current behavior returns a Messenger, which is
+ different to other `poutine.foo` functions
+ inconsistent with `poutine.scale` docs

I propose that we use `poutine.scale` for the old behavior and `poutine.scale_msg` (e.g.) for the messenger.

fritzo:
Do you have a use case for the `poutine.scale(fn, scale)` form? Could you give some example usage?

fehiepsi:
@fritzo An example for it is at https://github.com/fehiepsi/pyro/blob/825794e20ba8017db2b281911e013eb00e2162e4/examples/contrib/gp/sv-dkl.py#L105.
Of course, we can wrap gpmodel.model, gpmodel.guide with a model function and a guide function, which we can call `with messenger`... but it is a bit verbose to me ^^!

In case we keep `poutine.scale` as a messenger, then we should update its docs. Even with that, I still feel that it is quite different to other poutine calls.

eb8680:
I don't have a problem with reinstating `poutine.scale(fn, scale)` per se, but I'm not sure about the usage in that example (which I haven't looked at in detail).  Is the scaling necessary for training the neural network?

fritzo:
Hmm my original motivation for making `poutine.scale` a messenger is that I found the syntax
```py
with poutine.scale(batch_size / data_size):
    ...
```
more readable for beginning users than the older
```py
with poutine.scale(None, batch_size / data_size):
    ...
```
Is there any other way we can avoid the stray `None`s in 90% of use cases? E.g. maybe we can implement a decorator interface for messengers?
```py
class Messenger(object):
    def __call__(self, fn):
        """
        Apply this messenger as a Poutine. Works as a decorator.
        """

        @functools.wraps(fn)
        def decorated(*args, **kwargs):
            with self:
                return fn(*args, **kwargs)

        return decorated
```
Then we write `poutine.scale(1/1000)(model)` and still have nice `with poutine.scale(1/1000)` syntax.

eb8680:
> Then we write `poutine.scale(1/1000)(model)` and still have nice `with poutine.scale(1/1000)` syntax.

I've been contemplating making this interface change for all of `poutine`, but it would break a lot of internal and external code.  Thoughts?

fritzo:
> it would break a lot of internal and external code

[...]

PR #1000, opened 2018-04-10
Add a decorator interface to poutine

eb8680:
This PR adds a decorator interface to all `poutine.*Messenger` classes, so that we can do things like
```python
@poutine.ScaleMessenger(1/1000)
def model():
    ...
```
as requested by @martinjankowiak and @fritzo in #999.  It also removes all `poutine.*Poutine` classes, as they are redundant.

This PR does not break backwards compatibility, but I would like to remove the old interface entirely in a followup PR.  Because of that plan, I have also not updated `poutine`'s documentation yet.

(description edited by @fritzo)

### DIFF
--- a/docs/source/poutine.rst
+++ b/docs/source/poutine.rst
@@ -1,4 +1,4 @@
-Poutines (Pyro Coroutines)
+Poutine (Effect handlers)
 ==========================
 
 .. include:: pyro.poutine.txt
--- a/pyro/poutine/block_poutine.py
+++ b/pyro/poutine/block_poutine.py
@@ -12,8 +12,8 @@ class BlockMessenger(Messenger):
     Then any poutine outside of BlockMessenger(fn, hide=["a"])
     will not be applied to site "a" and will only see site "b":
 
-    >>> fn_inner = TraceMessenger(fn)
-    >>> fn_outer = TraceMessenger(BlockMessenger(TraceMessenger(fn), hide=["a"]))
+    >>> fn_inner = TraceMessenger()(fn)
+    >>> fn_outer = TraceMessenger()(BlockMessenger(hide=["a"])(TraceMessenger()(fn)))
     >>> trace_inner = fn_inner.get_trace()
     >>> trace_outer  = fn_outer.get_trace()
     >>> "a" in trace_inner
--- a/pyro/poutine/block_poutine.py
+++ b/pyro/poutine/block_poutine.py
@@ -1,9 +1,35 @@
 from __future__ import absolute_import, division, print_function
 
-from .poutine import Messenger, Poutine
+from .poutine import Messenger
 
 
 class BlockMessenger(Messenger):
+    # TODO fix this docstring
+    """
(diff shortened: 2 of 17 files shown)


pyro/poutine/block_poutine.py
```diff
@@ -1,9 +1,35 @@
 from __future__ import absolute_import, division, print_function
 
-from .poutine import Messenger, Poutine
+from .poutine import Messenger
 
 
 class BlockMessenger(Messenger):
+    # TODO fix this docstring
```
fritzo: Is this TODO meant for your follow-up PR? If so then LGTM.

pyro/poutine/poutine.py
```diff
@@ -13,6 +13,9 @@ class Messenger(object):
     def __init__(self):
         pass
 
+    def __call__(self, fn):
```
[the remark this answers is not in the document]
fritzo: We should add a docstring and expose this before 0.2 release. To expose this special method in sphinx, you can add a `:special-members: __call__` as in the pyro.distribution.distribution documentation https://raw.githubusercontent.com/uber/pyro/dev/docs/source/distributions.rst

fritzo:
From the [failed test logs](https://travis-ci.org/uber/pyro/jobs/364847727#L5476):
```py
with poutine.scale("advi_scope", 0.0):
    ...
```
What the heck? I think `"advi_scope"` was being silently treated as `fn` and ignored. Definitely not an error in this PR.

pyro/poutine/block_poutine.py
```diff
@@ -1,9 +1,35 @@
...
+    """
+    This Messenger selectively hides pyro primitive sites from the outside world.
+
+    For example, suppose the stochastic function fn has two sample sites "a" and "b".
+    Then any poutine outside of BlockMessenger(fn, hide=["a"])
+    will not be applied to site "a" and will only see site "b":
+
+    >>> fn_inner = TraceMessenger(fn)
```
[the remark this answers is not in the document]
eb8680: > maybe poutine.trace should special case

I don't think so, if we do that then `trace` is ambiguously typed.

pyro/poutine/block_poutine.py
eb8680: Yes

docs/source/poutine.rst
```diff
@@ -1,4 +1,4 @@
-Poutines (Pyro Coroutines)
+Poutine (Effect handlers)
```
jpchen: when you remove the old interface are you going to remove poutines entirely and expose them as pyro primitives ie `with pyro.scale(...)` or keep the existing namespace? 

>Neat, PR/issue number 1000!

"24 more until we reach a nice round number" :P

eb8680: I'm going to keep the `poutine` namespace, but the behavior of the aliases (`trace`, `replay`, etc.) will now be like the behavior of their respective `Messenger`s as of this PR.

PR #1001, opened 2018-04-11
Move all poutine usage to decorator interface

eb8680:
This PR follows up to the changes in #1000 by switching the `poutine` aliases to behave like `Messenger`s and moves all internal `poutine` usage over to that interface.  This enables use cases like
```python
import pyro.poutine.decorators as poutine

@poutine.scale(0.1)
def model(...)
    ...

with poutine.scale(0.1):
   ...

x = poutine.scale(0.1)(model)(...)
```

This PR does not change any behavior outside of `pyro.poutine`, just changes call signatures to be compatible with the new interface.  In followup PR(s), I'll update all the documentation, rename files in `poutine`, and remove the old interface entirely.

Currently blocked by #1000.  Along with #1000, addresses #999.


eb8680:
Hmm, after discussing with @fritzo it seems like there's a less invasive/awkward way to do this.

PR #1002, opened 2018-04-11
Implement discrete enumeration strategy for ADVI

fritzo:
Blocked by global state issues with `_DIM_ALLOCATOR`

1. refactors the ADVI library to make strategies more composable
2. implements a new DiscreteParallel strategy (that is only useful when combined with another strategy)

## Tested

- added a new smoke test

pair coded with @jpchen


tests/infer/test_advi.py
```diff
@@ -99,3 +99,27 @@ def model():
...
+def test_discrete_parallel(continuous_class):
+    K = 2
+    data = torch.tensor([0., 1., 10., 11., 12.])
+
+    def model(data):
+        weights = pyro.sample('weights', dist.Dirichlet(0.5 * torch.ones(K)))
+        locs = pyro.sample('locs', dist.Normal(0, 10).reshape([K]))
+        scale = pyro.sample('scale', dist.LogNormal(0, 1))
+
+        with pyro.iarange('data'):
```
fritzo: This iarange ends up allocating multiple dimensions in the global `_DIM_ALLOCATOR` when the `ADVIMaster` and `ADVIDiscreteParallel` both create [differently-blocked] `prototype_trace`s.

fritzo:
@jpchen I'm ready to merge as soon as tests pass. I'd like to move pyro.infer.advi to pyro.contrib.advi so we have freer license to refactor and make breaking changes.

martinjankowiak:
hmm, i believe in 2018 master/slave is frowned upon.... i know @jpchen is sensitive about this

PR #1004, opened 2018-04-11
Update poutine aliases to be usable as decorators and context managers

eb8680:
This PR follows up on #1000 and replaces #1001 to make the `poutine` aliases more flexible and ergonomic.  Now all `poutine` aliases take only keyword arguments and return a `Messenger` if they aren't passed a callable as their first argument.  This enables use cases like
```python
import pyro.poutine.decorators as poutine

@poutine.scale(scale=0.1)
def model(...)
    ...

with poutine.scale(scale=0.1):
   ...

x = poutine.scale(model, scale=0.1)(...)
```

This PR does not change any behavior outside of `pyro.poutine`, just changes call signatures to be compatible with the new interface.  In followup PR(s), I'll update all the documentation, rename files in `poutine`, and remove the old interface entirely.

Along with #1000, addresses #999.


eb8680:
Needs a few smoke tests

examples/air/main.py
```diff
@@ -222,7 +222,7 @@ def per_param_optim_args(module_name, param_name):
 
         if args.viz and i % args.viz_every == 0:
             trace = poutine.trace(air.guide).get_trace(examples_to_viz, None)
-            z, recons = poutine.replay(air.prior, trace)(examples_to_viz.size(0))
+            z, recons = poutine.replay(air.prior, trace=trace)(examples_to_viz.size(0))
```
fritzo: Can we use positional args when the function is specified? If so the you could shrink this PR by avoiding `trace=` in many places.

fritzo:
> Needs a few smoke tests

I think it's fine to sprinkle some decorator syntax in a few tests, just to ensure we're using all the different syntax variations.

examples/air/main.py
eb8680: I guess, but I kind of prefer the explicit keyword arguments as a convention

martinjankowiak:
is something blocking this?

fritzo:
@eb8680 we're planning to release mid next week. let's try to merge this before release.

eb8680:
Should be good to merge after review

pyro/poutine/__init__.py
```diff
@@ -150,16 +157,44 @@ def scale(fn, scale):
...
+    This messenger keeps track of stack of independence information declared by
+    nested ``irange`` and ``iarange`` contexts. This information is stored in
+    a ``cond_indep_stack`` at each sample/observe site for consumption by
+    ``TraceMessenger``.
+    """
+    msngr = IndepMessenger(name=name, size=size, dim=dim)
+    return msngr(fn) if fn is not None else msngr
+
+
+def enum(fn=None, first_available_dim=None):
```
fritzo: I thought we deleted this since it was not user facing. Is there a reason to expose it?

pyro/poutine/__init__.py
```diff
@@ -150,16 +157,44 @@ def scale(fn, scale):
...
     """
-    m = ScaleMessenger(scale=scale)
-    # temporary change to preserve API
-    return m(fn) if callable(fn) else m
+    msngr = ScaleMessenger(scale=scale)
+    # XXX temporary compatibility fix
+    return msngr(fn) if callable(fn) else msngr
+
+
+def indep(fn=None, name=None, size=None, dim=None):
```
fritzo: I thought we deleted this since it was not user facing. Is there a reason to expose it?

pyro/poutine/__init__.py
```diff
@@ -205,23 +242,26 @@ def queue(fn, queue, max_tries=None,
...
+                except util.NonlocalExit as site_container:
+                    site_container.reset_stack()
+                    for tr in extend_fn(ftr.trace.copy(), site_container.site,
+                                        num_samples=num_samples):
+                        queue.put(tr)
+
+            raise ValueError("max tries ({}) exceeded".format(str(max_tries)))
+        return _fn
+
+    return wrapper(fn) if fn is not None else wrapper
```
fritzo: I find the early-return pattern a little less invasive and easier to reason about:
```py
def queue(fn=None, queue=None, ...):
    # support use as a decorator
    if fn is None:
        return lambda fn: queue(fn=fn, queue=queue, ...)
    # ...remainder of queue() unmodified...
```
Do you have a preference either way?

eb8680: It's consistent with the primitive aliases, where it's important to return the `Messenger` object rather than a `lambda` wrapper

pyro/poutine/__init__.py
eb8680: It makes it possible to replace the implementation of `enum` in future without changing the interface

pyro/poutine/__init__.py
eb8680: It makes it possible to replace the implementation in future

PR #1007, opened 2018-04-11
Add GPLVM for dimensional reduction

fehiepsi:
This pull request adds GPLVM model following http://proceedings.mlr.press/v9/titsias10a/titsias10a.pdf

The model is just a wrap for any gpmodel which has input data X is a latent variable.

I made an anchor example for iris dataset here:
https://gist.github.com/fehiepsi/81277c39918ad4cd054157faed8b3de2


pyro/contrib/gp/util.py
```diff
@@ -66,24 +66,27 @@ def fix_param(self, param, value=None):
             value = getattr(self, param).detach()
         self._fixed_params[param] = value
 
-    def set_mode(self, mode):
+    def set_mode(self, mode, only_this_module=False):
```
fritzo: Maybe rename this to `recursive=True`?

pyro/contrib/gp/models/gplvm.py
```diff
@@ -0,0 +1,114 @@
...
+from pyro.contrib.gp.util import Parameterized
+import pyro.distributions as dist
+import pyro.infer as infer
+import pyro.optim as optim
+from pyro.params import param_with_module_name
+
+
+class GPLVM(Parameterized):
+    """
+    Gaussian Process Latent Variable Model (GPLVM) model.
```
fritzo: Could you add an `Example::` section? I'm not sure how I'd use this class.

pyro/contrib/gp/util.py
fehiepsi: Yes, `recursive` seems to be a better name.

pyro/contrib/gp/models/gplvm.py
fehiepsi: You can find an example here: https://gist.github.com/fehiepsi/81277c39918ad4cd054157faed8b3de2

I will add an example section soon.

pyro/contrib/gp/models/gplvm.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def model(self):
+        self.set_mode("model", only_this_module=True)
+
+        zero_loc = self.X_loc.new_zeros(self.X_loc.shape)
+        C = self.X_loc.shape[1]
+        Id = torch.eye(C, out=self.X_loc.new_empty(C, C))
+        X_name = param_with_module_name(self.name, "X")
+        X = pyro.sample(X_name, dist.MultivariateNormal(zero_loc, scale_tril=Id)
+                                    .independent(zero_loc.dim()-1))
```
fehiepsi: Here, we sample X from Normal(0, Id).

pyro/contrib/gp/models/gplvm.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def guide(self):
+        self.set_mode("guide", only_this_module=True)
+
+        X_loc = self.get_param("X_loc")
+        X_scale_tril = self.get_param("X_scale_tril")
+        X_name = param_with_module_name(self.name, "X")
+        X = pyro.sample(X_name,
+                        dist.MultivariateNormal(X_loc, scale_tril=X_scale_tril)
+                            .independent(X_loc.dim()-1))
```
fehiepsi: Here we define `guide` for X. The only difference to other GP models is we allow X to be a parameter (with unit normal prior and variational guide (not the MAP guide because according to literature, MAP guide produces poor results)).

fehiepsi:
@fritzo I guess that after 0.2, we will have Pyro versions released more often. So we don't need to have this merged before 0.2 release.

eb8680:
@martinjankowiak @ysaatchi can one of you check the math here?

martinjankowiak:
i can check it at some point but probably not before release

fehiepsi:
@martinjankowiak I made comments on the code but they might be overlooked so I wrap them up here. The only difference of GPLVM to the usual GP model is that its input X is a parameter which will be learned (hence the name GP latent variable model). My implementation follows [Titsias](http://proceedings.mlr.press/v9/titsias10a/titsias10a.pdf) which requires us to set a unit multivariate normal prior to X and also a variational (multivariate) normal distribution for its guide (formula (4) and (7)). Hence, to me, the math is pretty simple.

About the API, compared to [gpflow](https://github.com/GPflow/GPflow/blob/master/gpflow/models/gplvm.py#L52), this implementation is pretty simple. GPflow's GPLVM (with SGPR) likelihood is implemented from scratch, which is redundant to me.

What I didn't implement is the vanilla GPLVM, where X is a Parameter with no prior (hence no variational parameters). This model was reported to perform poorly in the literature (due to overfitting). I have thought that we can just let X is a Parameter and let users define a prior to it (so they can freely choose a DiagonalNormal distribution or MultivariateNormal distribution). This way, the implementation is even much simpler. However, currently, our Parameterize class only supports MAP guide. In the future, I intend to let users define arbitrary guides for their model's parameters. Then we can simplify the GPLVM implementation like this:

```
    def __init__(self, base_model, name="GPLVM"):
        super(GPLVM, self).__init__(name)
        self.base_model = base_model
        self.y = self.base_model.y
        self.X = Parameter(self.base_model.X)

    def model(self):
        self.set_mode("model", recursive=False)
        X = self.get_param("X")
        self.base_model.set_data(X, self.y)
        self.base_model.model()

    def guide(self):
        self.set_mode("guide", recursive=False)
        X = self.get_param("X")
        self.base_model.set_data(X, self.y)
        self.base_model.guide()
```

Please let me know if there is something else which needs to clarify. I just concerned about which features else to support for this model. We might let users decide it. :D

pyro/contrib/gp/models/gplvm.py
ysaatchi: This is great @fehiepsi ! Do add some of the math to the jupyter notebook if you can. This will clarify how the GPLVM is uniquely suited to a "pyronic" implementation like this one :)

ysaatchi: `p(X|y) \propto \prod_{d} p(y_{d} | X) p(X)` where `p(X) = N(0, Id) and p(y_{d} | X) = GP(\mu_{d}, K_{d})` (if I'm not mistaken!)

pyro/contrib/gp/models/gplvm.py
```diff
@@ -0,0 +1,131 @@
...
+
+        # sample X from variational multivariate normal distribution
+        X_loc = self.get_param("X_loc")
+        X_scale_tril = self.get_param("X_scale_tril")
+        X_name = param_with_module_name(self.name, "X")
+        X = pyro.sample(X_name,
+                        dist.MultivariateNormal(X_loc, scale_tril=X_scale_tril)
+                            .independent(X_loc.dim()-1))
+
+        self.base_model.set_data(X, self.y)
```
ysaatchi: Is this where you handle the D independent GPs which model the D-dimensions in `y` ?

fehiepsi: Yes, it depends on which `base_model` we use. For VariationalGP and VariationalSparseGP models, we have an arbitrary independent shape for the GP latent f (as long as its shape can be broadcasted to y.shape). For SparseGPRegression and GPRegression, currently, we just have 1 GP governs all D-dimensions of `y`..

pyro/contrib/gp/models/gplvm.py
fehiepsi: @ysaatchi Sadly I don't have time to write a tutorial for this model. I just added an example section for this model. The gist link just showed how to use this model.

pyro/contrib/gp/models/gplvm.py
ysaatchi: Very cool, so a lot of the machinery you need for GPLVM is already in there. This may be the shortest GPLVM implementation ever :)

eb8680:
Why does GitHub keep making me manually update this to `dev` and restart Travis?  Is that a new thing?  I tried to merge it a couple times today and couldn't because of Travis.

fritzo:
> Why does GitHub keep making me manually update

I've noticed the same thing in the last few days. I think Github must have started to be more conservative about testing unmerged branches...

PR #1008, opened 2018-04-11
Fix setup.py packages list

fritzo:
As of #993 Pyro dev cannot be pip installed except in editable mode, due to an incorrect specification of packages in setup.py


neerajprad:
Sorry about introducing this bug, and thanks for diagnosing and fixing this! :) Will merge when unit tests pass.

PR #1009, opened 2018-04-11
add docs badge to blacklist when deploying to pypi 

jpchen:
no tests


fritzo:
Nice. We should try to test the PyPi page generation on the PyPi test server before our 0.2 release: https://packaging.python.org/guides/using-testpypi/

PR #1011, opened 2018-04-11
Updating docker file to build latest PyTorch master

### LINKED ISSUE #1010: Docker build with Pyro dev, PyTorch master, and Python 2.7 fails
I was trying to build a Docker image from our dockerfile using `make build pyro_branch=dev pytorch_branch=master python_version=2.7` and got the following error:
```
Step 22/25 : RUN conda install -y python=${python_version} &&     conda install -y jupyter matplotlib
 ---> Running in d334a6741e9d
Solving environment: ...working... failed

UnsatisfiableError: The following specifications were found to be in conflict:
  - chardet -> python[version='>=3.6,<3.7.0a0']
  - python=2.7
Use "conda info <package>" to see the dependencies for each package.

The command '/bin/sh -c conda install -y python=${python_version} &&     conda install -y jupyter matplotlib' returned a non-zero code: 1
Makefile:40: recipe for target 'build' failed
```

neerajprad:
Fixes #1010. Due to some recent changes, installing Miniconda2 and later changing to python3 via `conda install python=3.6` or vice-versa isn't possible due to some incompatible libraries that are installed in the conda environment. To keep the complexity low, I am justing installing the correct Miniconda version to begin with (rather than introducing a new environment). The user is of course free to later create a new environment if needed.

Additionally :
 - Fixes the build script used by the Dockerfile to work with PyTorch master.
 - Makes the build faster - reorganizes the build so that we are executing most installation instructions under `pyromancer`. I was seeing that `chowning` the conda directory later took a really long time as the indexer would have to go through a lot of files. 

Checked that it works on python 2.7 / 3.6.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -13,10 +13,8 @@ ARG ostype=Linux
 # Configurable settings
 ENV USER_NAME pyromancer
 ENV CONDA_DIR /opt/conda
-ENV WORK_DIR /workspace
+ENV WORK_DIR /home/${USER_NAME}/workspace
 ENV PATH ${CONDA_DIR}/bin:${PATH}
-
-COPY install.sh /install.sh
 
 # Install linux utils
 RUN apt-get update && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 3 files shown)


neerajprad:
> Do we have any way to test this in travis? Or would it take too long?

Testing this on travis will still be too slow. :( We can add this to our travis cron job, but I'll have to investigate what we need to set up docker on travis.

neerajprad:
Let us not merge this yet. I am finding some CUDA issues with the gpu docker image (the CPU one works fine) which I am debugging, will update once that is fixed.

neerajprad:
@fritzo, @jpchen - This was working fine, I just realized I was not using the nvidia runtime when using the image. I have still updated the cuda to the latest version by default. Should be good to merge.

PR #1012, opened 2018-04-11
Add Poisson likelihood to GP module

fritzo:
I've forked the `Binary` likelihood to create a `Poisson` likelihood.

## Motivation

I'd like to model count-valued time series using Gaussian processes. Do I understand correctly that all I need is to implement a Poisson likelihood? Do I additionally need to implement a sampler so that I can turn GP samples into count-valued samples (to eventually compute quantiles)?

## Tested

- added test cases for Poisson likelihood


tests/contrib/gp/test_likelihoods.py
```diff
@@ -17,6 +17,9 @@
 y_binary1D = torch.tensor([0.0, 1.0, 0.0])
 y_binary2D = torch.tensor([[0.0, 1.0, 1.0], [1.0, 0.0, 1.0]])
 binary_likelihood = Binary()
+y_count1D = torch.tensor([0.0, 1.0, 4.0])
+y_count2D = torch.tensor([[5.0, 9.0, 3.0], [4.0, 0.0, 1.0]])
```
fehiepsi: I have thought that these should be LongTensor, but after checking pytorch's implementation, I agree that we should make them float.

fehiepsi:
LGTM. Thank @fritzo for implementing this! Regarding `y=None`, it depends on we merge this or #987 first.

eb8680:
Let's merge #987 first

pyro/contrib/gp/likelihoods/poisson.py
```diff
@@ -0,0 +1,49 @@
...
+        :rtype: torch.Tensor
+        """
+        # calculates Monte Carlo estimate for E_q(f) [logp(y | f)]
+        f = dist.Normal(f_loc, f_var)()
+        f_res = self.response_function(f)
+        y_dist = dist.Poisson(f_res)
+        if y is not None:
+            y_dist = y_dist.reshape(sample_shape=y.shape[:-f_res.dim()],
+                                    extra_event_dims=y.dim())
+        return pyro.sample(self.y_name, y_dist, obs=y)
```
fehiepsi: +1 for this, I will update other likelihoods in #1005.

fehiepsi:
To avoid circular merge conflict with other GP pull requests, I think it is better to merge this first. :)

examples/air/air.py
```diff
@@ -126,7 +126,7 @@ def prior_step(self, t, n, prev, z_pres_prior_p=default_z_pres_prior_p):
         # Sample presence indicators.
         z_pres = pyro.sample('z_pres_{}'.format(t),
                              dist.Bernoulli(z_pres_prior_p(t) * prev.z_pres)
-                                 .reshape(extra_event_dims=1))
+                                 .independent(1))
```
neerajprad: Do we need to merge with dev? Not sure why these changes are showing.

fritzo:
ok, should be ready to merge when tests pass

PR #1016, opened 2018-04-12
Rename .reshape(s,n) -> .expand_by(s).independent(n)

fritzo:
Addresses #827 

### Rename `.reshape(sample_shape=shape)` -> `.expand_by(shape)`

This gets closer to Pytorch `.expand()` nomenclature. Note that this behaves differently than `torch.Tensor.expand()`: `.expand_by()` adds `sample_shape` on the left, whereas `.expand()` broadcasts internal dimensions; hence the slight difference in name.

### Rename `.reshape(extra_event_dims=n)` -> `.independent(n)`

This is similar to [Tensorflow nomenclature](https://www.tensorflow.org/api_docs/python/tf/contrib/distributions/Independent) of `Independent(d, reinterpreted_batch_ndims=n)`, and anticipates an `Independent` distribution being added to PyTorch.

## Plans

In a follow-up PR I'd like to simplify `.expand_by()` further by allowing numeric `*args` rather than merely a `torch.Size`, e.g. `.expand_by(K, 1)`. This follow's PyTorch's convention for `.expand()`.

We may also split `ReshapedDistribution` into `BatchedDistribution` + `Independent`.


fritzo:
cc @katherine-h

neerajprad:
Thanks for this big renaming PR! Can I get this merged?

fritzo:
Ready to merge!

PR #1017, opened 2018-04-13
Add mean function for GP models

fehiepsi:
This is the last piece I want to add to GP module. Although mean function is not an interesting topic in Gaussian Process literature, it is useful in practice, e.g.:
+ to play the "trend" role in time series modeling (as in this forum [topic](https://forum.pyro.ai/t/how-to-model-parametric-trend-gp-residuals/146) by @fritzo )
+ to play a role of skip connection in residual network: x -> x + f(x) (as mentioned this [paper](http://papers.nips.cc/paper/7045-doubly-stochastic-variational-inference-for-deep-gaussian-processes))

I personally did not use this feature so just made a pre-support for it (e.g. no checking of the invalid shape of the output of a mean function).

- [x] There is still needed to add some tests.


fritzo:
Nice. Will I be able to train the `mean_function` during SVI, as [we've discussed](https://forum.pyro.ai/t/how-to-model-parametric-trend-gp-residuals/146/4)?

fehiepsi:
Sure, you can just define trend like this
'''
def trend(x):
    a = pyro.param("a", torch.tensor(0.0))
    b = pyro.param("b", torch.tensor(1.0))
    trend = a * x + b
    return trend
'''
then define a gp model with mean_function=trend.

This can work with trend is a complicated function such as a neural network, however, doing so might make gp model put less weight on its kernel; hence less useful to cover uncertainty.

I need to add some tests to make sure if I have made things right.

fehiepsi:
It does not work for arbitrary stochastic function with pyro.sample statements because we have to put some guides for latent variables. We can define it as a Parameterize module with its parameters having priors, so we can use the default MAP guide.

pyro/contrib/gp/models/svgp.py
```diff
@@ -146,7 +151,7 @@ def guide(self):
             pyro.sample(u_name,
                         dist.MultivariateNormal(u_loc, scale_tril=u_scale_tril)
                             .independent(u_loc.dim()-1))
-        return Xu, self.kernel, u_loc, u_scale_tril
+        return self.kernel, Xu, self.mean_function, u_loc, u_scale_tril
```
fritzo: It is generally error-prone to pack and unpack long tuples. Is there a good reason to return `self.kernel` and `self.mean_function` here? The only use I see is in `self.forward()`, where we could instead use `self.*` directly.

fehiepsi: Yes, we can use `self.kernel`, `self.mean_function` in forward. I made a return for them from `guide` to illustrate that they are ingredients to compute prediction. I agree that they are unnecessary so will remove soon. The reason to call `guide` in `forward` is to set `guide` mode for model/kernel and get parameters for prediction.

jpchen:
>  But I did import pyro at the beginning section

based on your diff, i think youre missing an `import pyro` for `test_models`

fehiepsi:
@jpchen You are right! I remembered that I looked at the imports and saw `import pyro` there before commenting. @@

PR #1019, opened 2018-04-13
Use Empirical Distribution in TracePosterior

neerajprad:
Use EmpiricalDistribution inside TracePosterior. This was mostly to make it easier to generate marginal distribution over sample sites in MCMC, but based on @eb8680 suggestions, I tried to do some refactoring to support Importance and Search (only existing tests, I am not sure if this will work with parallel search).

Changes:
 - TracePosterior has a `.marginal` method that returns a marginal over one or more sample sites.
 - There is a `Marginal` wrapper separately that calls `.marginal` under the hood but defers populating traces until we call `.sample` or `.log_prob`. This was needed to support Search, e.g. Schelling example. @eb8680 - Let me know your thoughts. 

Other utilities that could be built over TracePosterior:
 - Summary stats for sites in TracePosterior - mean, median, p10, p90.
 - Other queries like conditioning on certain sample sites.
 - Predictions: Run the model on the replayed traces and new data, to generate predictions.

### DIFF
--- a/examples/inclined_plane.py
+++ b/examples/inclined_plane.py
@@ -87,11 +87,11 @@ def model(observed_data):
 
 def main(args):
     # create an importance sampler (the prior is used as the proposal distribution)
-    posterior = Importance(model, guide=None, num_samples=args.num_samples)
+    importance = Importance(model, guide=None, num_samples=args.num_samples)
     # get posterior samples of mu (which is the return value of model)
     # from the raw execution traces provided by the importance sampler.
     print("doing importance sampling...")
-    emp_marginal = EmpiricalMarginal(posterior(observed_data))
+    emp_marginal = EmpiricalMarginal(importance.run(observed_data))
 
     # calculate statistics over posterior samples
     posterior_mean = emp_marginal.mean


examples/schelling_false.py
```diff
@@ -91,7 +88,6 @@ def main(args):
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description="parse args")
-    parser.add_argument('-n', '--num-samples', default=10, type=int)
```
neerajprad: Removed this argument since we can get the mean over the EmpiricalDistribution directly, rather than generating samples and computing the mean.

pyro/distributions/util.py
```diff
@@ -199,18 +199,16 @@ def matrix_triangular_solve_compat(b, A, upper=True):
         return b.trtrs(A, upper=upper)[0].view(b.shape)
 
 
-def log_sum_exp(tensor, dim=-1, scale=1.0):
+def log_sum_exp(tensor, dim=-1):
```
neerajprad: Removing the scale argument, which does not work when scale has negative values (which will be the case since these are just samples in our use case) and the resulting log will give NaNs. The `sum_exp` computation is moved to `Empirical._weighted_mean()`.

tests/infer/test_sampling.py
```diff
@@ -87,38 +87,25 @@ def test_complete(self):
...
-        posterior_stddev = torch.std(torch.cat(posterior_samples), 0)
-        assert_equal(0, torch.norm(posterior_mean - self.loc_mean).item(), prec=0.01)
-        assert_equal(0, torch.norm(posterior_stddev - self.loc_stddev).item(), prec=0.1)
+        posterior = pyro.infer.Importance(self.model, guide=self.guide, num_samples=5000)()
+        marginal = posterior.marginal()
+        assert_equal(0, torch.norm(marginal.mean - self.loc_mean).item(), prec=0.01)
+        assert_equal(0, torch.norm(marginal.variance.sqrt() - self.loc_stddev).item(), prec=0.1)
 
     @pytest.mark.init(rng_seed=0)
-    @pytest.mark.skip(reason='Slow test - use only for debugging')
```
neerajprad: Uncommenting as this finishes within 10s.

eb8680:
Looks great so far!  Let's handle `Search` in a separate PR and do `MCMC` and `Importance` here.

pyro/infer/abstract_infer.py
```diff
@@ -2,133 +2,63 @@
...
 
-    def _gen_weighted_samples(self, *args, **kwargs):
-        for tr, log_w in poutine.block(self.trace_dist._traces)(*args, **kwargs):
-            if self.sites == "_RETURN":
-                val = tr.nodes["_RETURN"]["value"]
-            else:
-                val = {name: tr.nodes[name]["value"]
-                       for name in self.sites}
-            yield (val, log_w)
+    @util.memoize
```
eb8680: We should probably remove this for non-`Search` marginals

examples/schelling_false.py
fritzo: Beware that this might make the test invisible to our example test discovery kludge. Could you confirm that `schelling_false.py` is still tested, or muck with test_examples.py to ensure it is?

neerajprad: I only ran this locally, and forgot about the `test_examples` kludge. Will check to see if that is running as expected.

neerajprad:
> Let's handle Search in a separate PR and do MCMC and Importance here

The only additional change for search is the `Marginal` class. It might be a bit tricky to separate that as the Search examples (particularly the Schelling examples) will not work with the TracePosterior class otherwise. Will discuss with you in detail.

pyro/infer/abstract_infer.py
neerajprad: Currently, this code path is only called by `Search` in existing tests/examples. Non Search code can call it too, but they should ideally just call `TracePosterior.marginal()` directly, which will store the results internally. In any case, disregarding any memory issues related to memoizing, I think the logic should be fine but will work in a restricted way (i.e. only when args, kwargs are the same instances).

pyro/infer/abstract_infer.py
```diff
@@ -2,133 +2,63 @@
...
 
-import pyro.distributions as dist
 import pyro.poutine as poutine
-from pyro.distributions.util import log_sum_exp
+from pyro.distributions import Distribution, Empirical
 import pyro.util as util
 
 
-def _eq(x, y):
+class Marginal(Distribution):
```
eb8680: Per discussion/design doc, rename to `EmpiricalMarginal` or `TraceMarginal`

pyro/infer/abstract_infer.py
```diff
@@ -137,12 +67,43 @@ def _traces(self, *args, **kwargs):
...
+
+        :param sites: sample site whose marginal to return.
+        :return: ``dist.Empirical``
+        """
+        if sites is None:
+            sites = "_RETURN"
+        assert isinstance(sites, (list, str))
+        sites_key = tuple(sites) if isinstance(sites, list) else sites
+        if sites_key in self._marginals:
+            return self._marginals[sites_key]
```
eb8680: What's the use case motivating caching marginals internally rather than just using explicit references?

pyro/infer/abstract_infer.py
```diff
@@ -2,133 +2,63 @@
...
     def __init__(self):
-        pass
+        self._init()
+
+    def _init(self):
+        self.log_weights = []
+        self.exec_traces = []
+        self._marginals = {}
 
     def _traces(self, *args, **kwargs):
```
eb8680: If this is still getting used externally we should remove `_`

pyro/infer/abstract_infer.py
```diff
@@ -137,12 +67,43 @@ def _traces(self, *args, **kwargs):
...
+        sites_key = frozenset([site])
+        if sites_key in self._marginals:
+            return self._marginals[sites_key]
+        marginal = Empirical()
+        for tr, log_weight in zip(self.exec_traces, self.log_weights):
+            marginal.add(tr.nodes[site]["value"], log_weight=log_weight)
+        self._marginals[sites_key] = marginal
+        return marginal
+
+    def marginal(self, sites=None):
```
eb8680: Per design doc/discussion, what do you think of having this be a separate function/class instead of a method?

pyro/infer/search.py
```diff
@@ -39,4 +42,6 @@ def _traces(self, *args, **kwargs):
             poutine.queue(self.model, queue=self.queue, max_tries=self.max_tries))
         while not self.queue.empty():
             tr = p.get_trace(*args, **kwargs)
+            if torch.isnan(tr.log_prob_sum()):
+                tr.nodes["bob_choice"]["fn"].log_prob(tr.nodes["bob_choice"]["value"])
```
eb8680: Remove example-specific code

tests/infer/mcmc/test_hmc.py
```diff
@@ -128,18 +128,14 @@ def test_hmc_conjugate_gaussian(fixture,
...
-            post_trace[param_name].append(t.nodes[param_name]['value'])
+    hmc_kernel = HMC(fixture.model, **hmc_params)
+    mcmc_run = MCMC(hmc_kernel, num_samples, warmup_steps)(fixture.data)
     for i in range(1, fixture.chain_len + 1):
         param_name = 'loc_' + str(i)
-        latent_loc = torch.mean(torch.stack(post_trace[param_name]), 0)
-        latent_std = torch.std(torch.stack(post_trace[param_name]), 0)
+        marginal = mcmc_run.marginal(param_name)
+        latent_loc = marginal.mean
+        latent_std = marginal.variance.sqrt()
```
eb8680: This API is much more user-friendly now!

pyro/distributions/empirical.py
```diff
@@ -140,24 +147,20 @@ def event_shape(self):
...
         if self._samples.dtype in (torch.int32, torch.int64):
-            raise ValueError("Mean for discrete empirical distribution undefined. Consider converting samples " +
-                             "to ``torch.float32`` or ``torch.float64``.")
-        weights = self._log_weights
-        for _ in range(self._samples.dim() - 1):
-            weights = weights.unsqueeze(-1)
-        return (log_sum_exp(weights, scale=self._samples, dim=0) - log_sum_exp(weights, dim=0)).exp()
+            raise ValueError("Mean for discrete empirical distribution undefined. " +
+                             "Consider converting samples to ``torch.float32`` " +
+                             "or ``torch.float64``.")
```
eb8680: This semantic difference seems large enough that maybe we should consider discrete empirical distributions a separate class?  Thoughts?

pyro/infer/abstract_infer.py
```diff
@@ -137,12 +67,43 @@ def _traces(self, *args, **kwargs):
...
         for tr, logit in poutine.block(self._traces)(*args, **kwargs):
-            traces.append(tr)
-            logits.append(logit)
-        logits = torch.stack(logits).squeeze()
-        logits -= log_sum_exp(logits)
-        ix = dist.Categorical(logits=logits).sample()
-        return traces[ix]
+            self.exec_traces.append(tr)
+            self.log_weights.append(logit)
+        return self
```
eb8680: Per design doc/discussion, should this return a sample trace with only latent variables?  That way you could use it for prediction as well via `replay`.  Thoughts?

pyro/infer/abstract_infer.py
neerajprad: The base class raises the exception so that the inheriting class overrides it. This isn't part of the API, and with EmpiricalMarginal, we shouldn't have to ever call `_traces`. If we have any such use cases, we will incorporate them into helper methods/functions.

pyro/infer/search.py
neerajprad: Oops..will fix.

pyro/infer/abstract_infer.py
neerajprad: There is a separate `Marginal` function, which is closer to the design doc. I think this relates to the caching issue that you mentioned above. 
 - The motivation is to populate traces once, and pass around this `TracePosterior` class to different methods, e.g. for visualization, summary stats etc. In that case, we will be doing repeated calls to `trace_posterior.marginal("site")` or other utility methods, which can be avoided. 
 - The above can be avoided by passing around these marginals to helper utilities, but that would get clunky with a lot of latent sites. What we can do is to memoize those calls, but it seems cleaner to have a functional, lazy `Marginal` and an imperative `.marginal` method.
 - The `Marginal` implementation above is lazy - it doesn't populate traces until we call `sample` or `log_prob`. While it memoizes the trace collection part, it is limited because the memoization is only based on `*args, **kwargs` ids, and if we use a different tensor instance for the same value, we will end up running the model and collecting traces again. To not have to run trace collection for each marginal `sample` or `log_prob` call (for non-Search algos), we will need a non-lazy version of Marginal which doesn't mesh well with Search.

This is a solution that I converged to while wrestling with some HMC stuff and our tests/examples (which we probably need more of), so I could easily be missing a lot of future use cases. Would love to hear suggestions from you and @fritzo to ensure that we have a generic reusable implementation, or at least one that can support those use cases in the future.

pyro/distributions/empirical.py
neerajprad: Are there any methods that you think will be specific to the discrete case? Except for Categorical, I think most other discrete distributions actually return a float tensor, so it will be a limited use case anyways.

pyro/infer/abstract_infer.py
neerajprad: This will populate `exec_traces`, which can be accessed directly as `mcmc(data).exec_traces`. Returning `self` has the advantage that we can use it to marginalize too - like `mcmc(data).marginal("site")`. I am fine with returning exec_traces directly from this method itself, and don't have a strong opinion on this otherwise.

fritzo:
> Summary stats for sites in TracePosterior - mean, median, p10, p90.

We currently have `.median()` and `.quantiles(...)` methods in the ADVI classes `AutoDiagonalNormal` and `AutoMultivariateNormal` ([code](https://github.com/uber/pyro/blob/c3b14dd/pyro/contrib/autoguide/advi.py#L269)). These are useful since they can be computed analytically and allow inspection of the under-the-hood latent. Maybe we can converge on a common interface?

neerajprad:
> Maybe we can converge on a common interface?

Agreed, this will benefit from a common interface that is reusable (at least consistent) across different inference algorithms, in the spirit of R's `summary(model_fit)`. Will discuss with you more.

eb8680:
> Agreed, this will benefit from a common interface that is reusable (at least consistent) across different inference algorithms

As discussed in the design doc, I think representations of marginal distributions are the right intermediate layer for this; they have consistent APIs but may be different for different algorithms - for example, ADVI admits a more specialized representation because the marginals can be computed analytically.  Instead of trying to squeeze everything into `TracePosterior`, we can just compute a dictionary of marginals with e.g. `{name: Marginal(posterior, site=name) for name in site_names}` and pass that to a summary function.

pyro/infer/abstract_infer.py
eb8680: As discussed in the design doc, I think the thing to give up on for now is the laziness in `Marginal`.  Instead we can bind a fixed set of input arguments to a posterior and have `Marginal` call `traces` in `Marginal.__init__`.  That way we can separate `marginal` and `TracePosterior` and directly pass around marginal distributions (or dictionaries of marginal distributions as mentioned below) to summarization helpers without having to maintain any extra mutable internal state.

neerajprad: I think I will need your help to work it out. 

If we change to a non-lazy marginal like follows:
```
class EmpiricalMarginal(Distribution):
    """
    :param trace_dist: a TracePosterior instance representing a Monte Carlo posterior.

    Marginal distribution, that wraps over a TracePosterior object to provide a
    a Distribution over the return values of the TracePosterior's model. Calls
    ``TracePosterior.marginal()`` under the hood, but defers populating traces
    until ``.sample`` or ``.log_prob`` methods are called.
    """
    has_enumerate_support = True

    def __init__(self, trace_dist, *args, **kwargs):
        assert isinstance(trace_dist, TracePosterior), \
            "trace_dist must be trace posterior distribution object"

        self.sites = kwargs.pop("sites", "_RETURN")
        self.empirical_marginal = trace_dist(*args, **kwargs).marginal(self.sites)
        super(Distribution, self).__init__()

        return self.empirical_marginal.sample(sample_shape)

    def log_prob(self, value):
        return self.empirical_marginal.log_prob(value)

    def enumerate_support(self):
        return self.empirical_marginal.enumerate_support()
```

and modify the schelling example, it throws an error about multiple sites named `loc`. I tried it earlier as well, and wasn't able to get it to work. Does it need changes in `Search.py`?

```
def alice(preference, depth):
    """
    Alice decides where to go by reasoning about Bob's choice
    """
    alice_prior = location(preference)
    return pyro.sample("bob_choice", EmpiricalMarginal(Search(bob), preference, depth - 1),
                       obs=alice_prior)


def bob(preference, depth):
    """
    Bob decides where to go by reasoning about Alice's choice
    """
    bob_prior = location(preference)
    if depth > 0:
        return pyro.sample("alice_choice", EmpiricalMarginal(Search(alice), preference, depth),
                           obs=bob_prior)
    else:
        return bob_prior
```

> without having to maintain any extra mutable internal state

Conspicuous mutability is the design choice of Python, unfortunately. Without immutable data structures and other supporting language features, many functional constructs in Python would be woefully inefficient or gnarly. :( 

Anyways, will sync with you in person, and try to work this out for Search.

eb8680: Maybe we should just delete `Search` and the Schelling examples, unless @ngoodman objects...

neerajprad: Or we can have different versions of `Marginal` as you had in the design doc, and it can be lazy for Search? I think I was hoping that we can re-use the same, but it might just be easier to have different implementations instead.

neerajprad:
@eb8680 - The last commit removes the `.marginal` in favor of an `EmpiricalMarginal` class that wraps `TracePosterior`. This simplifies the code, but this eager version of `marginal` doesn't work with `search`. :( We can temporarily disable these examples, and reinstate them later with a lazy marginal implementation.

eb8680:
I'd like to see `TracePosterior.__call__` renamed to `TracePosterior.run`, but otherwise this should be good to merge after fixing merge conflicts introduced by #1044.

fritzo:
Before release let's make sure that `EmpiricalMarginal` shows up in our sphinx docs.

neerajprad:
> I'd like to see TracePosterior.__call__ renamed to TracePosterior.run

Done.

PR #1023, opened 2018-04-13
Traceenum warnings

### LINKED ISSUE #946: Add error for incomplete enumeration specification
To enumerate over discrete variables, it's necessary both to supply the argument `enum_discrete=True` to `SVI`, and to mark discrete samples using `infer = {"enumerate": "sequential"}` (or, equivalently, wrap the guide in `config_enumerate`). However, I noticed that the model runs (albeit without enumeration) if one marks `enum_discrete=True` but neglects to mark the samples. I believe this should throw an error: with the current setup, it's easy to think you're running a model with discrete enumeration when you're actually not.

fritzo:
Resolves #946

This adds two warnings:
1. if `loss=TraceEnum_ELBO()` but no sample sites are configured for enumeration, and
2. if `loss` is not `TraceEnum_ELBO` and any sample sites are configured for enumeration.

## Tested

- added regression tests for both cases

### DIFF
--- a/tests/infer/test_inference.py
+++ b/tests/infer/test_inference.py
@@ -235,8 +235,8 @@ def test_exponential_gamma(gamma_dist, n_steps, elbo_impl):
         pyro.sample("lambda_latent", gamma_dist(alpha_q, beta_q))
 
     adam = optim.Adam({"lr": .0003, "betas": (0.97, 0.999)})
-    svi = SVI(model, guide, adam, loss=elbo_impl(),
-              max_iarange_nesting=1)
+    elbo = elbo_impl(strict_enumeration_warning=False)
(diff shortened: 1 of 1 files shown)


tests/infer/test_valid_models.py
```diff
@@ -493,6 +537,24 @@ def guide():
     assert_error(model, guide, TraceGraph_ELBO())
 
 
+@pytest.mark.parametrize("enumerate_", [None, "sequential", "parallel"])
```
fritzo: This is the main relevant test.

fritzo:
> these issue warnings regardless of pyro validation level, correct?

This PR introduces warnings that are issued iff `pyro.infer.is_validation_enabled()`.

PR #1024, opened 2018-04-14
update ss-vae

jpchen:
this PR does three things:

1. clean up existing code
2. support parallel enumeration (with fritzo's help)
3. use new ELBO semantics


jpchen:
ahhh shoot this was cloned off the advi branch.. let me clean up and resolve conflicts...

examples/vae/ss_vae_M2.py
```diff
@@ -41,6 +44,7 @@ def __init__(self, output_size=10, input_size=784, z_dim=50, hidden_layers=(500,
         self.z_dim = z_dim
         self.hidden_layers = hidden_layers
         self.epsilon_scale = epsilon_scale
+        self.config_enum = config_enum == 'parallel'
```
martinjankowiak: isn't this trail of args confusingly named? this becomes allow_broadcast?

examples/vae/ss_vae_M2.py
```diff
@@ -197,6 +200,10 @@ def guide_classify(self, xs, ys=None):
         pass
 
     def model_sample(self, ys, batch_size=1):
+        """
```
martinjankowiak: model_sample should be replaced with poutine magic

examples/vae/ss_vae_M2.py
```diff
@@ -388,29 +395,27 @@ def run_inference_ss_vae(args):
...
+    parser.add_argument('-ne', '--num-epochs', default=100, type=int,
                         help="number of epochs to run")
     parser.add_argument('--aux-loss', action="store_true",
                         help="whether to use the auxiliary loss from NIPS 14 paper (Kingma et al)")
     parser.add_argument('-alm', '--aux-loss-multiplier', default=300, type=float,
                         help="the multiplier to use with the auxiliary loss")
     parser.add_argument('-enum', '--enum-discrete', default=None,
-                        help="whether to enumerate the discrete support of the categorical distribution"
-                             "while computing the ELBO loss")
+                        help="parallel or sequential")
```
martinjankowiak: does this mean sampling the discrete is not an option?

martinjankowiak:
has someone confirmed that the parallel actually works? 

what are the plans for further clean-up/refactoring?

examples/vae/ss_vae_M2.py
fritzo: I agree this is confusing. Maybe instead set this to:
```
self.config_enum = config_enum if config_enum else None
```

examples/vae/ss_vae_M2.py
fritzo: If you leave it blank or set it to the empty string, then discretes should be sampled. @jpchen can you indicate this in the help message?

examples/vae/ss_vae_M2.py
fritzo: unrelated to this PR

fritzo: this nit is unrelated to the change; no need to address it in this PR

examples/vae/ss_vae_M2.py
jpchen: i renamed `self.config_enum` to `allow_broadcast` which is a flag based on if `enum` is parallel, which comes from the cla. `config_enum` is a string used in `config_enumerate`, and we only want to allow broadcasting for parallel enum, not sequential.  let me know if this is less confusing.

jpchen:
> has someone confirmed that the parallel actually works?

yes, see description above. originally parallel was not being tested.

> what are the plans for further clean-up/refactoring?

i agree this example should probably be rewritten, but i only addressed the main issues here.  we can make a list and throw them in the cleanup issue? a few that come to mind is refactoring utils and rewriting `model_sample()` as you stated above

jpchen:
>  could you enable --enum-discrete=parallel tests?

done, i actually noticed that but incorrectly assumed there was a parallel enum to be fixed in air as well.

PR #1027, opened 2018-04-14
Allow auxiliary sites in guide but not model; simplify ADVI

fritzo:
This relaxes our model-guide consistency constraint to allow auxiliary sample sites to appear in the guide but not the model. The motivation is to simplify pyro.contrib.autoguide's ADVI implementations (as suggested by @eb8680).

## The Problem

Pyro's current ADVI implementations need to add to the model fake `scale=0.0` sample sites for each auxiliary latent variable in the ADVI guide. This is cumbersome and requires us to wrap `model` with these extra `scale=0.0` sites.

## Proposed Solution

@eb8680 suggested trying to relax our requirement that all sample sites appear in the model. I've attempted to make this as safe as possible by requiring these extra sites to be marked `infer={'is_auxiliary': True}`.

**Pros:** This allows us to simplify ADVI implementations and to directly use the original `model` in advi, reducing our ADVI implementation code by about 50%, and making ADVI easier to use and compose. This also allows us to use auxiliary sites in user-defined guides, which is useful e.g. to jointly sample a site in the guide that corresponds to multiple sites in the model (and then extract those multiple site via `Delta` sample statements).

**Cons:** This requires minor changes to our `Trace*_ELBO` implementations.

## Tasks

- [x] Allow sites marked `infer={'is_auxiliary': True}` to appear in the guide but not model
- [x] Simplify ADVI to use the original model
- [x] Update `Trace_ELBO` to handle auxiliary sample sites
- [x] Update `TraceGraph_ELBO` to handle auxiliary sample sites
- [x] Update `TraceEnum_ELBO` to handle auxiliary sample sites

## Tested

- Existing ADVI tests cover the refactoring in this PR


fritzo:
@martinjankowiak could you take a look at the `TraceGraph_ELBO` changes? In particular I'm not sure whether `non_reparam_nodes` is being computed correctly (currently it considers only nodes in the guide).

fritzo:
> Do you think we have the right default value of is_auxiliary?

Yes, it currently defaults to `False` which matches existing Pyro behavior.

> Should we allow `is_auxiliary = True` in the model as well?

It is easy to add that in the future if we ever find a use case, but I have no present need.

pyro/infer/trace_elbo.py
```diff
@@ -77,7 +77,7 @@ def loss(self, model, guide, *args, **kwargs):
         """
         elbo = 0.0
         for model_trace, guide_trace in self._get_traces(model, guide, *args, **kwargs):
-            elbo_particle = (model_trace.log_prob_sum() - guide_trace.log_prob_sum()).item()
+            elbo_particle = torch_item(model_trace.log_prob_sum()) - torch_item(guide_trace.log_prob_sum())
```
martinjankowiak: why did torch_item become necessary?

pyro/infer/tracegraph_elbo.py
```diff
@@ -216,19 +213,8 @@ def loss(self, model, guide, *args, **kwargs):
...
-            for name in model_trace.nodes.keys():
-                if model_trace.nodes[name]["type"] == "sample":
-                    if model_trace.nodes[name]["is_observed"]:
-                        elbo_particle += model_trace.nodes[name]["log_prob_sum"]
-                    else:
-                        elbo_particle += model_trace.nodes[name]["log_prob_sum"]
-                        elbo_particle -= guide_trace.nodes[name]["log_prob_sum"]
-
-            elbo += torch_data_sum(weight * elbo_particle)
+            elbo_particle = model_trace.log_prob_sum().item() - guide_trace.log_prob_sum().item()
```
martinjankowiak: but isn't needed eg here?

pyro/util.py
```diff
@@ -266,10 +266,15 @@ def check_model_guide_match(model_trace, guide_trace, max_iarange_nesting=float(
     :param pyro.poutine.Trace guide_trace: Trace object of the guide
     :raises: RuntimeWarning, ValueError
 
-    Checks that (1) there is a bijection between the samples in the guide
-    and the samples in the model, (2) each `iarange` statement in the guide
-    also appears in the model, (3) at each sample site that appears in both
-    the model and guide, the model and guide agree on sample shape.
+    Checks the following assumptions:
+    1. Each sample site in the model also appears in the guide and is not
+        marke auxiliary.
```
martinjankowiak: marke => marked

pyro/infer/traceenum_elbo.py
```diff
@@ -13,23 +13,35 @@
 from pyro.util import check_model_guide_match, check_site_shape, check_traceenum_requirements, torch_isnan
 
 
+def _dict_iadd(dict_, key, value):
+    if key in dict_:
+        dict_[key] = dict_[key] + value
+    else:
+        dict_[key] = value
```
jpchen: nit: this can be better written as: `dict_[key] = dict_.get(key, 0) + value`

pyro/contrib/autoguide/advi.py
```diff
@@ -126,24 +116,9 @@ class ADVISlave(object):
     """
     def __init__(self, model):
         self.master = None
-        self.base_model = model
+        self.model = model
         self.prototype_trace = None
 
-    def model(self, *args, **kwargs):
```
jpchen: 👍

pyro/infer/traceenum_elbo.py
fritzo: That would add an extra tensor op `0 + tensor`.

pyro/infer/trace_elbo.py
fritzo: To increase precision in light of the reordering of operations in this PR, I've switched from accumulating `elbo_particle` as a (typically 32bit) `FloatTensor` to accumulating it as a 64-bit Python float. To achieve this, we could thus use `.item()` except that in cases of empty models or guides, `.log_prob_sum()` is a float rather than a tensor. Hence the special handling `x if isinstance(x, Number) else x.item()`.

pyro/infer/tracegraph_elbo.py
fritzo: :+1: good point, I'll add `torch_item()` here too

fritzo:
> we will also likely need to revisit this stuff someday

yeah let's brainstorm and determine under what circumstances `is_auxiliary` is valid

fritzo:
Thanks for reviewing! I'll rebase #1036 as soon as this is merged.

PR #1028, opened 2018-04-16
Expand cuda tests

jpchen:
`make test-cuda` now tests the same tests that are run on cpu


jpchen:
> I feel like there should be a more factorized way to do this

i originally abstracted out cuda logic but then remembered a wise sage warning me about the perils of over-engineering to save a few lines of code :P

PR #1031, opened 2018-04-16
Avoid namespace pollution in top level module

neerajprad:
This fixes the `__init__.py` files for the remaining modules as part of #981 - `infer`, `params` and our top level `__init__.py` file. 

Only a few primitives are imported in the top level module to prevent namespace pollution. Most of these imports were not being used outside of supporting the primitive functions themselves (except for `param_with_module_name` in GP which I have removed).

Due to undergoing poutine refactoring, I have stayed away from making changes in `poutine` module.


pyro/primitives.py
```diff
@@ -0,0 +1,439 @@
+from __future__ import absolute_import, division, print_function
```
jpchen: this file was moved verbatim from `pyro/__init__.py` right?

pyro/params/__init__.py
```diff
@@ -1,23 +1,12 @@
...
-def user_param_name(param_name):
-    if _MODULE_NAMESPACE_DIVIDER in param_name:
-        return param_name.split(_MODULE_NAMESPACE_DIVIDER)[1]
-    return param_name
+from pyro.params.param_store import _MODULE_NAMESPACE_DIVIDER, _PYRO_PARAM_STORE, \
+    module_from_param_with_module_name, param_with_module_name, user_param_name
+
+__all__ = [
+    "_MODULE_NAMESPACE_DIVIDER",
+    "_PYRO_PARAM_STORE",
```
jpchen: nit: this isnt relevant to this PR per se, but if these are being exposed or used outside the module then they shouldnt have `_` prepended?

neerajprad: +1, agreed. Let me see if we can remove this, and just use the get functions everywhere.

pyro/primitives.py
neerajprad: That's correct, this should be a straight migration.

pyro/params/__init__.py
fritzo: Hmm I consider `_` to denote pravacy wrt the repo, rather than privacy wrt a module. E.g. PyTorch uses `_` to denote internal methods with no backwards-compatibility guarantees. Therefore,
1. it should be fine to import underscored names across modules;
2. we don't need to add undescored names to `__all__`, since `__all__` only determins `*` import, and we avoid `*` import internally.

neerajprad: That makes sense, will clean that up shortly.

PR #1032, opened 2018-04-16
Upgrade to latest PyTorch master

neerajprad:
Updating to latest PyTorch master. Based on suggestions from #1031, cleaning up the `__all__` to remove global variables private to Pyro.

### DIFF
--- a/pyro/primitives.py
+++ b/pyro/primitives.py
@@ -112,7 +112,7 @@ class _Subsample(Distribution):
         """
         self.size = size
         self.subsample_size = subsample_size
-        self.use_cuda = torch.Tensor.is_cuda if use_cuda is None else use_cuda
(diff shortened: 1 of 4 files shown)


pyro/primitives.py
```diff
@@ -112,7 +112,7 @@ def __init__(self, size, subsample_size, use_cuda=None):
         """
         self.size = size
         self.subsample_size = subsample_size
-        self.use_cuda = torch.Tensor.is_cuda if use_cuda is None else use_cuda
+        self.use_cuda = torch.Tensor().is_cuda if use_cuda is None else use_cuda
```
neerajprad: This doesn't seem to be a static method anymore, and does not return a boolean unless instantiated.

PR #1033, opened 2018-04-17
Turn off distributions arg check during HMC step size adaptation

neerajprad:
- Turn off distributions arg checking during warmup phase. While writing the tutorial, I realized that the warnings during the adaptation phase are spurious, and it is clunky to keep toggling validation check when we are in warmup phase vs. otherwise.
 - Introduces a context manager for toggling off distributions validation.
 - Minor logging related changes - specify whether we are in WARMUP/SAMPLE stage.

### DIFF
--- a/pyro/distributions/util.py
+++ b/pyro/distributions/util.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import numbers
+from contextlib import contextmanager
 
 import torch
 import torch.distributions as torch_dist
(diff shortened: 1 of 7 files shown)


pyro/distributions/util.py
```diff
@@ -221,3 +222,13 @@ def enable_validation(is_validate):
 
 def is_validation_enabled():
     return _VALIDATION_ENABLED
+
+
+@contextmanager
+def validation_enabled(is_validate=True):
```
fritzo: Do you have any plans to document these? It seems excessive to document all of our utilities, but these seem like they will be useful for users.

neerajprad: I think so too..I also wanted to fix a few docs related things, and will address it in that PR.

neerajprad:
Ready to merge.

PR #1034, opened 2018-04-17
Refactor examples testing

neerajprad:
From our discussion #1019, I thought it might be a good time to refactor `test_examples.py`. It seems that an explicit enumeration is likely to be the most flexible (we can constrain the example as much as we like), and opaque way for us to test examples. 

Also added a coverage test so that we do not miss out on adding examples to `test_examples`. In the process, I realized that we were not testing the `gp` example and cuda versions of `ss-vae`, which is added.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -8,11 +8,49 @@ import pytest
 
 from tests.common import EXAMPLES_DIR, requires_cuda
 
-CPU_EXAMPLES = []
-CUDA_EXAMPLES = []
+CPU_EXAMPLES = [
+    ('air/main.py', ['--num-steps=1']),
+    ('bayesian_regression.py', ['--num-epochs=1']),
+    ('contrib/gp/sv-dkl.py', ['--epochs=1', '--num-inducing=4', '--no-cuda']),
+    ('contrib/named/mixture.py', ['--num-epochs=1']),
+    ('contrib/named/tree_data.py', ['--num-epochs=1']),
+    ('dmm/dmm.py', ['--num-epochs=1']),
+    ('dmm/dmm.py', ['--num-epochs=1', '--num-iafs=1']),
+    ('inclined_plane.py', ['--num-samples=1']),
+    ('schelling.py', ['--num-samples=1']),
+    ('schelling_false.py', ['--num-samples=1']),
+    ('vae/ss_vae_M2.py', ['--num-epochs=1']),
+    ('vae/ss_vae_M2.py', ['--num-epochs=1', '--aux-loss']),
+    ('vae/ss_vae_M2.py', ['--num-epochs=1', '--enum-discrete=parallel']),
+    ('vae/ss_vae_M2.py', ['--num-epochs=1', '--enum-discrete=sequential']),
+    ('vae/vae.py', ['--num-epochs=1']),
+    ('vae/vae_comparison.py', ['--num-epochs=1']),
+]
+
+CUDA_EXAMPLES = [
+    ('air/main.py', ['--num-steps=1', '--cuda']),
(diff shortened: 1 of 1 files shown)


tests/test_examples.py
```diff
@@ -8,57 +8,62 @@
...
+    ('dmm/dmm.py', ['--num-epochs=1', '--num-iafs=1']),
+    ('inclined_plane.py', ['--num-samples=1']),
+    ('schelling.py', ['--num-samples=1']),
+    ('schelling_false.py', ['--num-samples=1']),
+    ('vae/ss_vae_M2.py', ['--num-epochs=1']),
+    ('vae/ss_vae_M2.py', ['--num-epochs=1', '--aux-loss']),
+    ('vae/ss_vae_M2.py', ['--num-epochs=1', '--enum-discrete=parallel']),
+    ('vae/ss_vae_M2.py', ['--num-epochs=1', '--enum-discrete=sequential']),
+    ('vae/vae.py', ['--num-epochs=1']),
+    ('vae/vae_comparison.py', ['--num-epochs=1']),
```
jpchen: this is definitely more readable but seems less scalable? most examples will only need `num-epochs` set, but will require a manual modification of this file each time

neerajprad: There are examples that need multiple parameters like `ss_vae` and require us to specify circuitous logic which seems very error prone. e.g. `if 'aux_loss' in text`. 

We just need one entry here at the time when you write an example, with no future modification needed. And if you forget it, the coverage test will fail. So w.r.t. scalability, I think that our bottleneck to writing more examples will just be the effort required in writing one, rather than adding one entry here! 😄

fritzo:
ready to merge?

tests/test_examples.py
```diff
@@ -8,57 +8,62 @@
...
-
-discover_examples()
-
-
-def make_ids(examples):
-    return ['{} {}'.format(example, ' '.join(args)) for example, args in examples]
+            if '__main__' in text:
+                if example not in cpu_tests:
+                    pytest.fail('Example: {} not covered in CPU_TESTS.'.format(example))
+                if 'cuda' in text and example not in cuda_tests:
```
jpchen: nit: it's safer to check for `'--cuda' in text` because there might be mention of cuda in the comment section of non-cuda examples, but `--cuda` is universally used to turn on cuda support in the examples

fritzo: yeah fishery models are pretty common in the hierarchical bayes literature, and a barracuda might sneak in to a cpu-only example.

neerajprad: Haha..actually we have examples that use `--use-cuda` or even `--no-cuda` which we were missing out on. Hence this check in coverage. We can standardize these examples, but there is no guarantee someone might not end up sneaking in a new variant like `--go-cuda`.

neerajprad:
> ['vae/ss_vae_M2.py', '--num-epochs=1', '--enum-discrete=parallel']

I wanted to do minimal modifications to `@pytest.mark.parametrize('example,args', CPU_EXAMPLES, ids=make_ids(CPU_EXAMPLES))`; hence retained the list types to what was being populated automatically. Let me see if there's a simple way to support these parameterizations with a flat array.

tests/test_examples.py
jpchen: >actually we have examples that use --use-cuda

oof thats terrible we should standardize them

neerajprad: Yeah, we should. This coverage is designed to catch all potential examples that are cuda compatible, but I suppose we should insist on uniformity of arguments during code reviews.

jpchen:
can we merge #1037 first and then check for `--cuda` in test_examples?

PR #1036, opened 2018-04-17
Simplify ADVI interface (part 2)

fritzo:
This is step 2 in refactoring our ADVI interface before 0.2 release (after #1027), as suggested by @eb8680.

After this PR we can simply

```py
def model():
    ...

guide = AutoMultivariateNormal(model)
svi = SVI(model, guide, Adam({}), Trace_ELBO())
```

### DIFF
--- a/pyro/contrib/autoguide/advi.py
+++ b/pyro/contrib/autoguide/advi.py
@@ -28,23 +28,59 @@ def _product(shape):
     return result
 
 
-class ADVIMaster(object):
+class AutoGuide(object):
     """
-    Container class to combine multiple ADVI strategies.
+    Base class for automatic guides.
+
+    Derived classes must implement the :meth:`__call__` method.
+
+    Auto guides can be used individually or combined in an
+    :class:`AutoGuideList` object.
+    """
+    def __init__(self, model):
+        self.master = None
+        self.model = model
+        self.prototype_trace = None
+
+    def __call__(self, *args, **kwargs):
+        """
+        A guide with the same ``*args, **kwargs`` as the base ``model``.
+
+        :return: A dict mapping sample site name to sampled value.
+        :rtype: dict
+        """
+        raise NotImplementedError
+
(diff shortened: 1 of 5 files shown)


pyro/contrib/autoguide/advi.py
```diff
@@ -28,23 +28,59 @@ def _product(shape):
     return result
 
 
-class ADVIMaster(object):
+class AutoGuide(object):
```
fritzo: Formerly called `ADVISlave` (code was moved from below with little change)

pyro/contrib/autoguide/advi.py
```diff
@@ -28,23 +28,59 @@ def _product(shape):
...
+        pass
+
+    def _create_iaranges(self):
+        if self.master is not None:
+            return self.master().iaranges
+        return {frame.name: pyro.iarange(frame.name, frame.size, dim=frame.dim)
+                for frame in sorted(self._iaranges.values())}
+
+
+class AutoGuideList(AutoGuide):
```
fritzo: formerly called `ADVIMaster`

pyro/contrib/autoguide/advi.py
```diff
@@ -28,23 +28,59 @@ def _product(shape):
...
+        self.master = None
+        self.model = model
+        self.prototype_trace = None
+
+    def __call__(self, *args, **kwargs):
+        """
+        A guide with the same ``*args, **kwargs`` as the base ``model``.
+
+        :return: A dict mapping sample site name to sampled value.
+        :rtype: dict
```
eb8680: Are you using this return value anywhere?

fritzo: Yes, Slawek uses it to inspect latent variables. I guess we could force users to trace the guide instead and eliminate this logic in a future PR.

fritzo:
ready to merge.

PR #1037, opened 2018-04-17
standardize cuda semantics across examples

jpchen:
See discussion #1034 [comment](https://github.com/uber/pyro/pull/1034#discussion_r182177751)


examples/contrib/gp/sv-dkl.py
```diff
@@ -138,14 +138,13 @@ def cnn_fn(x):
                         help='number of epochs to train (default: 10)')
     parser.add_argument('--lr', type=float, default=0.01, metavar='LR',
                         help='learning rate (default: 0.01)')
-    parser.add_argument('--no-cuda', action='store_true', default=False,
-                        help='disables CUDA training')
+    parser.add_argument('--cuda', action='store_true', default=False,
```
fritzo: Did you mean change the behavior, or do you want to preserve behavior and change the default value to `True`? (I'm fine changing behavior, I just want to confirm your intent)

jpchen: i meant to change it, since for all examples, default is cpu unless specified otherwise. sorry should have called that out.

neerajprad:
There's an image that is committed with this, that we should remove.

jpchen:
you mean theres an image already committed and i removed it in this PR..?

neerajprad:
> you mean theres an image already committed and i removed it in this PR..?

Oh..didn't realize that this was removing the committed image.

PR #1038, opened 2018-04-17
Add circular von Mises distribution

fritzo:
This is useful for observation of circular data.

 This PR implements only `.log_prob()`. Even without `.sample()` this is a useful distribution to have.

## Tested

- added to existing tests against scipy version
- added a test of normalization

### DIFF
--- a/pyro/distributions/von_mises.py
+++ b/pyro/distributions/von_mises.py
@@ -1,0 +1,65 @@
+from __future__ import absolute_import, division, print_function
+
+import math
+
+import torch
+from torch.distributions import constraints
+from torch.distributions.utils import broadcast_all
+
+from pyro.distributions import TorchDistribution
+
+
+def _eval_poly(y, coef):
+    coef = list(coef)
+    result = coef.pop()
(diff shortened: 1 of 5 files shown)


pyro/distributions/von_mises.py
```diff
@@ -0,0 +1,65 @@
...
+    arg_constraints = {'loc': constraints.real, 'concentration': constraints.positive}
+    support = constraints.real
+
+    def __init__(self, loc, concentration, validate_args=None):
+        self.loc, self.concentration = broadcast_all(loc, concentration)
+        batch_shape = self.loc.shape
+        event_shape = torch.Size()
+        super(VonMises, self).__init__(batch_shape, event_shape, validate_args)
+
+    def log_prob(self, value):
```
jpchen: you removed the `normalize` flag?

fritzo: Yes, it turns out we always needed to use the normalized version to get correct derivatives.

fritzo:
Thanks for reviewing!

PR #1039, opened 2018-04-17
update gp tutorial

martinjankowiak:
@fehiepsi i've made a first pass at trying to improve the readability of this tutorial.

i'm more or less satisfied with the first part (before the sparse stuff). the second part (sparse stuff) still needs some work. i'd like to get some feedback from you before i finish editing the second part. i also wonder if some of the naming conventions in the GP code could be improved (see below)?

# for the first part (non-sparse)

note that i've refactored the plot function. in some cases i'm a little confused about when the observation noise is/isn't included in the kernel. please make sure i haven't introduced anything wrong here. also let me know if i've removed  or distorted any explanatory points that you thought were worth making.

- is there any reason we'd still need `pyro.enable_validation(False)` ?
- if i understand it right, `GPRegression` can only do gaussian likelihoods. this seems like a strange naming convention: regression doesn't mean gaussian likelihoods it means continuous outputs. or am i missing something?
- why did you want to use `gpr.fix_param("noise", torch.ones(1) * 0.01)` in the MAP section?

# for the second part (sparse)

- it seems like `RBF$$$variance_MAP` is constrained (and thus positive) while `RBF$$$variance` is not. isn't this confusing?
- similar to the point above it seems that `SparseGPRegression` implies gaussian likelihood?
- why fix the noise to 0.01 in the last model?


fehiepsi:
@martinjankowiak Thank you very much for revising gp tutorial! Here are some of my comments:
+ I think we can set_validation now because we have updated pyro to a pytorch commit which solved the instability of precision matrix computation.
+ Both [GPy implementation](https://github.com/SheffieldML/GPy/blob/devel/GPy/models/gp_regression.py#L34) and [GPflow's one](https://github.com/GPflow/GPflow/blob/master/gpflow/models/gpr.py#L49) use Gaussian likelihood for regression. In GP literature (to my knowledge), GP regression also mean Gaussian likelihood. So I think that it is a good name here.
+ About naming, I am more concerned about the name: SparseVariationalGP and SparseMVN. To me (and from literature), it is better to call it VariationalSparseGP. I used SVGP because GPflow authors used that name for their implementation. About SparseMVN, I think it is better to call it LowRankMVN (to compatible with literature about low-rank approximation of a covariance matrix, e.g. [Stein's](https://www.sciencedirect.com/science/article/pii/S2211675313000390)).
+ About constraint, if we use `pyro.get_param_store().named_parameters()` then due to [this](https://github.com/uber/pyro/blob/dev/pyro/params/param_store.py#L54), it will return unconstrained value. For all parameters of GP module, to me it is better to call `model.get_param(param_name)` or `kernel.get_param(param_name)`. We can call `pyro.param(param_name_with_module_name)` but it will take extra effort to define `param_name_with_module_name`.
+ We can set a prior to noise with a small value. I remember that it is fixed there just to illustrate `fix_param` functionality, which is unnecessary now because we have a better doc now.
+ To include/eliminate noise in the prediction, you can control it through `noiseless` parameter (in the plot function). I have included noise in all plots but you can change that behavior to what is better for you.
+ For GPR and SGPR, kernel is still the original kernel. Noise is added into the kernel (to create new kernel) just to help computation.
+ In a recent commit, you can set name to kernels, models to avoid duplicated params (if you don't want to use `pyro.clear_param_store()`). E.g.
`kernel2 = RBF(input_dim=1, variance=torch.tensor(0.1), lengthscale=torch.tensor(10.)), name="RBF2"`
+ It is good to expose SVI steps in the first inference. For the later ones, you can simplify it by
`gpmodel.optimize(optimizer=optim.Adam({"lr": 0.01}), num_steps=100)` (0.001 and 1000 are default values).

martinjankowiak:
re: regression, i'm afraid that's not the convention. for example see here:

https://papers.nips.cc/paper/3806-gaussian-process-regression-with-student-t-likelihood

fehiepsi:
@martinjankowiak Personally, I think of "GPR" as a combined name, rather than a GP model to do regression problem. The paper you mentioned said that "In the Gaussian process regression the observation model is commonly assumed to be Gaussian, which is convenient in computational perspective." in its abstract. The standard reference: [GPML book](http://www.gaussianprocess.org/gpml/chapters/) assumes Gaussian and to my knowledge, other GP libraries also follow this convention. So to not make users confused, we should find an appropriate name. Do you have a better name for it? Maybe GaussianGP?

martinjankowiak:
@fehiepsi thanks for your quick response!

- i don't love the name GPRegression but i guess it's fine. still, the docs should be updated
https://github.com/uber/pyro/blob/dev/pyro/contrib/gp/models/gpr.py
to make it clear that the 'noise' is referring to the use of a gaussian likelihood with a particular variance. it should be very clear that it doesn't support other likelihoods.

- i've made another pass. this is now ready for review. it would be great if you could take a close look!

fehiepsi:
@martinjankowiak I can't make a review directly from web due to large diff so I make comments here. Mostly they are just suggestions so you can select which ones are appropriate:
+ At this point
> Here  x,x′∈X  are points in the input space and  y∈Y  is a point in the output space.  f  is a draw from the GP prior specified by the kernel  Kf  and represents a function from  X  to Y .

Because we use y = f + eps, I think y should be an output vector rather than a point here.

+ I guess no need to call "pyro.clear_param_store()" at the beginning

+ In the plot function, I think it is better to have comment: "# (iii) plots samples from the prior (with no conditioning on observed data, **inducing points, and variational parameters**)" and update the code to cover variational model (though we don't use it)
```
noise = model.noise if type(model) != SparseVariationalGP else model.likelihood.variance
cov = kernel.forward(Xtest) + noise.expand(n_test).diag()
```
+ In the later part of **Inference** section, do you feel that there are many "Note that" phrase? ^^
+ In **Fit the model using MAP** section, to avoid the unconstrained param problem, we can use:
```
for param_name in pyro.get_param_store().get_all_param_names():
    print(param_name, pyro.param(param_name).item())
```
+ In **More Sparse GPs**, I hope you can introduce the `sv-dkl` [example](https://github.com/uber/pyro/blob/dev/examples/contrib/gp/sv-dkl.py) when mentioned about other likelihoods.

+ "So we can compare to what we've done above, we're still going to use a gaussian likelihood." -> "To compare to..."
+ Across the tutorial, do you think that it is better to use "Gaussian" instead of "gaussian"? 

About docs of GPRegression, I think we already have made it clear at [this section](https://github.com/uber/pyro/blob/dev/pyro/contrib/gp/models/gpr.py#L35).

martinjankowiak:
@fehiepsi great, thanks, i'll incorporate your feedback and make another pass.

re: the docstring of `GPRegression`, nowhere does it say that the noise is gaussian. it just says 'noise'

fehiepsi:
Ah, got it. I will update its docs in a future pull request related to GP.

martinjankowiak:
@fehiepsi  i've incorporated your feedback. thanks! this should now be good to go. 

open to comments/suggestions from anyone else!

jpchen:
@martinjankowiak is this good to merge? if @fehiepsi is fine with it im happy to merge it and add additional adjustments in future PRs

fehiepsi:
Yes, I like it. :D

PR #1043, opened 2018-04-18
Removing global default cuda tensor specification in SS VAE example

rohitsingh0812:
removed global default tensor when using cuda and started using data.new_zeros and data.new_ones inside model/guide. Without this change - we would fail with an error 'RuntimeError: randperm is only implemented for CPU' with the latest well-tested pytorch commit

```
$ python ss_vae_M2.py --cuda
Downloading http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz
Downloading http://yann.lecun.com/exdb/mnist/train-labels-idx1-ubyte.gz
Downloading http://yann.lecun.com/exdb/mnist/t10k-images-idx3-ubyte.gz
Downloading http://yann.lecun.com/exdb/mnist/t10k-labels-idx1-ubyte.gz
Processing...
Done!
  File "ss_vae_M2.py", line 453, in <module>
    main(args)
  File "ssand started using data.…_vae_M2.py", line 362, in main
    run_inference_for_epoch(data_loaders, losses, periodic_interval_batches)
  File "ss_vae_M2.py", line 250, in run_inference_for_epoch
    (xs, ys) = next(unsup_iter)
  File "/home/env/lib/python3.6/site-packages/torch/utils/data/dataloader.py", line 262, in __next__
    indices = next(self.sample_iter)  # may raise StopIteration
  File "/home/env/lib/python3.6/site-packages/torch/utils/data/sampler.py", line 138, in __iter__
    for idx in self.sampler:
  File "/home/env/lib/python3.6/site-packages/torch/utils/data/sampler.py", line 51, in __iter__
    return iter(torch.randperm(len(self.data_source)).tolist())
RuntimeError: randperm is only implemented for CPU

```

### DIFF
--- a/examples/vae/ss_vae_M2.py
+++ b/examples/vae/ss_vae_M2.py
@@ -113,13 +113,13 @@ class SSVAE(nn.Module):
         with pyro.iarange("data"):
 
             # sample the handwriting style from the constant prior distribution
-            prior_loc = torch.zeros([batch_size, self.z_dim])
-            prior_scale = torch.ones([batch_size, self.z_dim])
+            prior_loc = xs.new_zeros([batch_size, self.z_dim])
+            prior_scale = xs.new_ones([batch_size, self.z_dim])
             zs = pyro.sample("z", dist.Normal(prior_loc, prior_scale).independent(1))
 
             # if the label y (which digit to write) is supervised, sample from the
             # constant prior, otherwise, observe the value (i.e. score it against the constant prior)
-            alpha_prior = torch.ones([batch_size, self.output_size]) / (1.0 * self.output_size)
+            alpha_prior = xs.new_ones([batch_size, self.output_size]) / (1.0 * self.output_size)
             ys = pyro.sample("y", dist.OneHotCategorical(alpha_prior), obs=ys)
 
             # finally, score the image (x) using the handwriting style (z) and
@@ -173,7 +173,7 @@ class SSVAE(nn.Module):
         res, ind = torch.topk(alpha, 1)
 
         # convert the digit(s) to one-hot tensor(s)
-        ys = torch.zeros(alpha.size())
+        ys = xs.new_zeros(alpha.size())
         ys = ys.scatter_(1, ind, 1.0)
         return ys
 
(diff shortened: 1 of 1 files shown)


rohitsingh0812:
@jpchen  -- if you set a global default tensor type to be cuda.FloatTensor (as we were doing earlier), torch.randperm will complain because it makes sure that the result type requested from the operation is on the CPU (that's my understanding from the code below). torch.randperm is used by default in the random sampler that shuffles the training set every epoch. In general, globally setting the tensor type can be an issue for other operations that may only be supported on the CPU. 

https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/TensorFactories.cpp#L269

```
Tensor& randperm_out(Tensor& result, int64_t n, Generator* generator) {
  if (n < 0) {
    std::ostringstream oss;
    oss << "n must be non-negative, got " << n;
    throw std::runtime_error(oss.str());
  }
  if (result.type().backend() != at::kCPU) {
    throw std::runtime_error("randperm is only implemented for CPU");
  }

  result.resize_({n});
  auto gen = get_generator(generator);
  AT_DISPATCH_ALL_TYPES(result.type(), "randperm", [&]() -> void {
    randperm_cpu<scalar_t>(result, n, gen);
  });

  return result;
}
```

jpchen:
> globally setting the tensor type can be an issue for other operations that may only be supported on the CPU.

i understand this but i dont understand how this change would fix that. unless there're cpu tensors allocated while running on gpu (if so, where?), `new_zeros/new_ones` should just create other cuda tensors which would error when you call `randperm` for the reason you stated above. we can discuss this tomorrow, it's not pressing, i was just curious.

examples/vae/ss_vae_M2.py
```diff
@@ -299,9 +299,6 @@ def main(args):
     :param args: arguments for SS-VAE
     :return: None
     """
-    if args.cuda:
```
rohitsingh0812: @jpchen - I removed the global setting here by removing these lines - so, this should prevent pytorch from asking torch.randperm to return a cuda tensor (which was the main error that showed up). but because of this change, we can't use torch.zeros and torch.ones in the model when we want the loss to be computed on GPUs -- hence the other fixes.

PR #1044, opened 2018-04-18
remove both schelling game examples and infer.Search

eb8680:
These don't seem to be instructive or useful in practice, and they're slowing down development, so unless @ngoodman objects we should just remove them.


ngoodman:
Removing the schelling examples seems ok, but removing inference by search seems like a major step back. How is the latter getting in the way of development?

eb8680:
> How is the latter getting in the way of development?

We have to maintain it as we update `Marginal` and `TracePosterior` e.g. in #1019 and it's not being used in our codebase except in those examples.

fritzo:
> removing inference by search seems like a major step back

It sounds like @eb8680 has major refactoring plans for `Marginal` and `TracePosterior`. @eb8680 also mentioned parallelizing `Search`, which I'm guessing will be nearly a complete rewrite.

Does it make sense to remove our current implementation of `Search` and plan on a future parallel implementation that integrates more closely with our refactored `Marginal` and `TracePosterior`? @ngoodman are there any users who currently rely on `Search` and would miss it if we temporarily remove it?

PR #1047, opened 2018-04-19
torch.nograd() for evaluate_loss

martinjankowiak:
there used to be 'volatile' but no more...

(i meant to do this a long time ago but kept forgetting... was reminded when i ran into memory blow-ups during evaluation)

### DIFF
--- a/pyro/infer/svi.py
+++ b/pyro/infer/svi.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
+import torch
+
 import pyro
 from pyro.infer.elbo import ELBO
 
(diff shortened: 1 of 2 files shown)


pyro/infer/svi.py
```diff
@@ -61,7 +61,8 @@ def evaluate_loss(self, *args, **kwargs):
 
         Evaluate the loss function. Any args or kwargs are passed to the model and guide.
         """
-        return self.loss(self.model, self.guide, *args, **kwargs)
+        with torch.nograd():
```
fritzo: Typo, should be `torch.no_grad()`

martinjankowiak: whooops

PR #1049, opened 2018-04-19
fix ctx usage in omt mvn

martinjankowiak:
never write `ctx.blah = blah`. memory leaks may arise....

### DIFF
--- a/pyro/distributions/omt_mvn.py
+++ b/pyro/distributions/omt_mvn.py
@@ -36,18 +36,16 @@ class OMTMultivariateNormal(MultivariateNormal):
 class _OMTMVNSample(Function):
     @staticmethod
     def forward(ctx, loc, scale_tril, shape):
-        ctx.white = loc.new_empty(shape).normal_()
-        ctx.z = torch.matmul(ctx.white, scale_tril.t())
-        ctx.save_for_backward(scale_tril)
-        return loc + ctx.z
+        white = loc.new_empty(shape).normal_()
+        z = torch.matmul(white, scale_tril.t())
+        ctx.save_for_backward(z, white, scale_tril)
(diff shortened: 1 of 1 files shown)


pyro/distributions/omt_mvn.py
```diff
@@ -36,18 +36,16 @@ def rsample(self, sample_shape=torch.Size()):
 class _OMTMVNSample(Function):
     @staticmethod
     def forward(ctx, loc, scale_tril, shape):
-        ctx.white = loc.new_empty(shape).normal_()
-        ctx.z = torch.matmul(ctx.white, scale_tril.t())
-        ctx.save_for_backward(scale_tril)
-        return loc + ctx.z
+        white = loc.new_empty(shape).normal_()
+        z = torch.matmul(white, scale_tril.t())
+        ctx.save_for_backward(z, white, scale_tril)
```
fritzo: Does this work? I thought you could only save inputs or outputs for backward. (That was certainly the case in PyTorch 0.3)

martinjankowiak: i wouldn't push a proposed fix if i hadn't tested it

fritzo: Can we confirm that PyTorch changed behavior before merging this? I.e. locate the PyTorch PR where this behavior changed, or locate discussion about it?

martinjankowiak: is adnotation #3 helpful?

https://github.com/pytorch/pytorch/pull/1016

fritzo: It looks like the behavior may have changed in https://github.com/pytorch/pytorch/pull/4786/files so I guess this is fine.

Issue #988, opened 2018-04-08
Support Distribution-valued observations?


fritzo:
What would it take to support `Distribution`-valued observations in `pyro.sample()` statements? The idea is to Rao-Blackwellize by interpreting observe statements with distribution-valued observations
```py
pyro.sample("x", Normal(loc,scale), obs=Normal(0,1))
```
as inducing a `log_prob` term equal to `-KL(obs, dist)`
```py
site["log_prob"]  = -kl_divergence(Normal(0, 1), Normal(loc, scale))
```
This should be equivalent to but lower-variance than the current entirely-stochastic version
```py
pyro.sample("y", Normal(0, 1))
pyro.sample("x", Normal(loc, scale), obs=y)
```
In this interpretation, we can view our usual tensor-valued observations as observations from `Delta` distributions:
```py
pyro.sample("x", Normal(loc, scale), obs=Delta(y))
assert -kl_divergence(Delta(y), Normal(loc, scale)) == Normal(loc, scale).log_prob(y)
```
In case the distribution pair is missing from the analytic `kl_divergence()` table, we can always revert to a single Monte-Carlo sample as in the current implementation.

Note that ideas like this have been discussed before in the context of KL divergence #91 #688 ; this issue aims to support a limited form of `kl_divergence()` by extending the `pyro.sample()` primitive.

martinjankowiak:
hmm, this is interesting... though seems a bit strange/unidiomatic to me?
how does this interact with trace/replay?

in any case, one still has the fundamental issue that the existence of a particular analytic kl isn't a sufficient condition to integrate out a given random variable. of course one could decide that that issue is of lesser importance

eb8680:
This is a little strange to me, especially outside the context of SVI, although that doesn't mean there's not a sound probabilistic interpretation - I guess it's like saying inference is computing a conditional expectation with respect to those random variables, rather than just a posterior distribution?

> analytic KLs in SVI

Isn't the simplest/most obvious SVI use case (analytic KL between prior and guide) easier to expose by just allowing users to mark sites for analytic KLs in `site["infer"]`, like we do for enumeration?

eb8680:
> I'll leave this issue open for a week in case @karalets wants to comment, and then close it.

I think the use case you proposed is different and interesting enough that we can keep it open for more discussion.  We can create a different issue for the simpler use case if you want.

karalets:
Yeah I like this issue actually.

Generally, this is a different way of looking at loss functions.

[...]

PR #1050, opened 2018-04-20
Updating the VAE tutorial

neerajprad:
Changes:
 - I have updated the code in the code-blocks and added it as runnable code cells. Also added a smoke test for the tutorial.
 - Updated the test elbo loss figure which was not correct earlier.
 - Minor cosmetic changes.

PS: I strongly suggest that we convert our other tutorials like ss-vae to have runnable code cells. It is a bit of a time investment, but will make it easier to maintain these tutorials over time.


neerajprad:
> are the examples up to date?

The example is other wise up to date.

> do we need all that cuda logic in the tutorials? since presumably users will use the actual example if running on gpu

I ran the tutorial on the gpu. It takes quite a long time otherwise to even run 10 epochs. So I suppose we should have the cuda flag so that people can toy around with it.

martinjankowiak:
@karalets looks like this vanilla VAE is getting a test elbo of about -102 with a 50 dimensional latent space and 400 hidden units in the nn... how does that seem?

tutorial/source/vae.ipynb
```diff
@@ -85,13 +85,47 @@
...
+    "from pyro.optim import Adam"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "pyro.enable_validation(False)\n",
```
fritzo: Should we set `pyro.enable_validation(smoke_test)` to be safe?

fritzo: Ok. This would be a nice use case for #988 
```py
pyro.sample("obs", dist.Bernoulli(prediction), obs=dist.Bernoulli(ground_truth))
```
or even for parallel enumeration (if it doesn't add too much overhead):
```py
truth = pyro.sample("truth", dist.Bernoulli(ground_truth),
pyro.sample("obs", dist.Bernoulli(prediction), obs=truth)
```
But we can address this later...

tutorial/source/vae.ipynb
```diff
@@ -106,13 +140,24 @@
...
-    "```python\n",
+    "Next we define a PyTorch module that encapsulates our decoder network:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
```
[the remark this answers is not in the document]
martinjankowiak: setup the non-linearity => setup the non-linearities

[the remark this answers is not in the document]
martinjankowiak: if the fudge disappeared kill this comment: fixing numerical instabilities of sigmoid with a fudge

tutorial/source/vae.ipynb
```diff
@@ -161,60 +215,91 @@
...
-    "    # (value will be sampled by guide when computing the ELBO)\n",
-    "    z = pyro.sample(\"latent\", dist.Normal(z_loc, z_scale))\n",
-    "\n",
-    "    # decode the latent code z\n",
-    "    mu_img = self.decoder(z)\n",
-    "    # score against actual images\n",
-    "    pyro.sample(\"obs\", dist.Bernoulli(x.reshape(-1, 784)), obs=mu_img)\n",
-    "```\n",
-    "\n",
+    "    with pyro.iarange(\"data\", x.size(0)):\n",
```
martinjankowiak: this iarange needs explaining in the text

tutorial/source/vae.ipynb
```diff
@@ -161,60 +215,91 @@
...
+    "        # setup hyperparameters for prior p(z)\n",
+    "        z_loc = x.new_zeros(torch.Size((x.size(0), self.z_dim)))\n",
+    "        z_scale = x.new_ones(torch.Size((x.size(0), self.z_dim)))\n",
+    "        # sample from prior (value will be sampled by guide when computing the ELBO)\n",
+    "        z = pyro.sample(\"latent\", dist.Normal(z_loc, z_scale).independent(1))\n",
+    "        # decode the latent code z\n",
+    "        loc_img = self.decoder.forward(z)\n",
+    "        # score against actual images\n",
+    "        pyro.sample(\"obs\", dist.Bernoulli(loc_img).independent(1), obs=x.reshape(-1, 784))"
```
martinjankowiak: same with `.independent(1)`

tutorial/source/vae.ipynb
```diff
@@ -161,60 +215,91 @@
...
+    "def guide(self, x):\n",
     "    # register PyTorch module `encoder` with Pyro\n",
     "    pyro.module(\"encoder\", self.encoder)\n",
-    "    # use the encoder to get the parameters used to define q(z|x)\n",
-    "    z_loc, z_scale = self.encoder(x)\n",
-    "    # sample the latent code z\n",
-    "    pyro.sample(\"latent\", dist.Normal(z_loc, z_scale))\n",
-    "```\n",
-    "\n",
+    "    with pyro.iarange(\"data\", x.size(0)):\n",
```
martinjankowiak: worth pointing to iarange/ind in guide too

tutorial/source/vae.ipynb
```diff
@@ -247,81 +365,182 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Run options\n",
+    "LEARNING_RATE = 1.0e-3\n",
+    "USE_CUDA = False\n",
+    "NUM_EPOCHS = 1 if smoke_test else 100\n",
```
martinjankowiak: explain smoke_test with short comment

neerajprad:
Thanks for reviewing. I will address all of the comments shortly.

>  looks like this vanilla VAE is getting a test elbo of about -102 with a 50 dimensional latent space and 400 hidden units in the nn... how does that seem?

What would have been your expectation otherwise? For context, we had it go up to `-80` earlier with the same model, because Bernoulli was not using binary cross entropy for scoring.

neerajprad:
@martinjankowiak - Let me know if you have any more comments.

examples/vae/vae.py
```diff
@@ -46,10 +46,10 @@ def forward(self, x):
 class Decoder(nn.Module):
     def __init__(self, z_dim, hidden_dim):
         super(Decoder, self).__init__()
-        # setup the three linear transformations used
+        # setup the two linear transformations used
         self.fc1 = nn.Linear(z_dim, hidden_dim)
         self.fc21 = nn.Linear(hidden_dim, 784)
-        # setup the non-linearity
+        # setup the non-linearities
```
martinjankowiak: looks good but you have `fudge` in the decoder in vae.py but not in the notebook?

neerajprad: Good point; it's not needed. Removed.

martinjankowiak: but can you confirm you can get good numbers (no crashing etc.) without the fudge?

neerajprad: The test graph was generated from the notebook itself, so I can confirm that it doesn't crash, and gives reasonable results. Let me generate some sample images too, and validate.

PR #1052, opened 2018-04-20
Testing for CUDA compatibility

neerajprad:
Some changes to our tests/code to ensure that unit tests and examples can run on CUDA. This does not tackle the `_standard_gamma` patching (for running on CUDA) which is needed for tests that use gamma, dirichlet, chi2 etc.
 - Some functions like `torch.linspace` and `torch.randperm` throw an error when we call them with a default cuda tensor type. I am either fixing the device type to CPU, or using `torch.arange` instead of `torch.linspace`.
 - Adjusting the precision threshold for CUDA, and fixing the cuda rng seed.

### DIFF
--- a/pyro/primitives.py
+++ b/pyro/primitives.py
@@ -127,7 +127,8 @@ class _Subsample(Distribution):
         if subsample_size == self.size:
             result = torch.LongTensor(list(range(self.size)))
         else:
-            result = torch.randperm(self.size)[:self.subsample_size]
+            # torch.randperm does not have a CUDA implementation
+            result = torch.randperm(self.size, device=torch.device('cpu'))[:self.subsample_size]
         return result.cuda() if self.use_cuda else result
 
     def log_prob(self, x):
--- a/pyro/util.py
+++ b/pyro/util.py
@@ -162,12 +162,9 @@ def set_rng_seed(rng_seed):
     :param int rng_seed: The seed value.
     """
     torch.manual_seed(rng_seed)
-    if torch.cuda.is_available():
-        torch.cuda.manual_seed(rng_seed)
     random.seed(rng_seed)
     try:
         import numpy as np
-
         np.random.seed(rng_seed)
     except ImportError:
         pass
--- a/pyro/util.py
+++ b/pyro/util.py
@@ -162,6 +162,8 @@ def set_rng_seed(rng_seed):
     :param int rng_seed: The seed value.
     """
     torch.manual_seed(rng_seed)
+    if torch.cuda.is_available():
+        torch.cuda.manual_seed(rng_seed)
     random.seed(rng_seed)
     try:
         import numpy as np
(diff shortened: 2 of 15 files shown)


pyro/util.py
```diff
@@ -162,6 +162,8 @@ def set_rng_seed(rng_seed):
     :param int rng_seed: The seed value.
     """
     torch.manual_seed(rng_seed)
+    if torch.cuda.is_available():
```
fritzo: As of PyTorch 0.3 this was not necessary: `torch.manual_seed()` also calls `torch.cuda.manual_seed()`. Adam recommended we remove this. Has anything changed?

tests/contrib/gp/test_models.py
```diff
@@ -209,16 +212,16 @@ def test_inference_svgp():
...
     kernel = RBF(input_dim=1)
-    Xu = torch.linspace(0, 5, 10)
+    Xu = torch.arange(0, 5.5, 0.5)
 
     svgp = SparseVariationalGP(X, y, kernel, Xu, Gaussian())
-    svgp.optimize(optim.Adam({"lr": 0.01}), num_steps=1000)
+    svgp.optimize(optim.Adam({"lr": 0.01}), num_steps=2000)
 
-    Xnew = torch.linspace(0, 5, 100)
+    Xnew = torch.arange(0, 5.05, 0.05)
```
fritzo: I find `linspace` easier to reason about than `arange`. Is `linspace` not supported in CUDA?

neerajprad: Yes, this will throw an error to that effect.

pyro/util.py
neerajprad: Hmm..I probably read some outdated documentation in that case. Let me check and remove it if not needed.

pyro/primitives.py
```diff
@@ -127,7 +127,8 @@ def sample(self, sample_shape=torch.Size()):
         if subsample_size == self.size:
             result = torch.LongTensor(list(range(self.size)))
         else:
-            result = torch.randperm(self.size)[:self.subsample_size]
+            # torch.randperm does not have a CUDA implementation
+            result = torch.randperm(self.size, device=torch.device('cpu'))[:self.subsample_size]
```
jpchen: if `randperm` does not have cuda implementation, thsi will throw an error either way on cuda right

neerajprad: This is basically generating a random permutation on the CPU, via the `device` argument and then sending it to the GPU on the next line. The device argument is needed now, as otherwise `randperm` internally checks the device type and throws an error if it is cuda (which it will be for instance if we set the default tensor type as CUDA).

neerajprad:
@fritzo - Lets merge #1053 first, and then ensure that `test-cuda` has no failing tests after this PR.

fritzo:
Should I merge this?

neerajprad:
@fritzo - Just fixing a few more failures. Will update shortly.

neerajprad:
Should be ready to merge once tests pass.

PR #1054, opened 2018-04-20
Drop support for string value of loss arg to SVI

### LINKED ISSUE #1014: Deprecate "ELBO" arg to SVI; require object instead
(pulled out of #827)

@eb8680 suggested removing the string arg `loss` `SVI.__init__` and instead requiring an `ELBO` object: 
```diff
- svi = SVI(model, guide, optim, 'ELBO', ...)
+ elbo = Trace_ELBO(...)
+ SVI(model, guide, optim, elbo)
```

This is advantageous in that it makes it easier for users to implement custom losses, as e.g. requested in #958. This is a huge breaking change that will touch lots of files. It would be nice to finish this before 0.2 release.

## Tasks (in multiple small PRs)

- [x] #1018 add support for `loss` `ELBO` object to `SVI.__init__()`
- [x] #1018 remove support for `loss` function arg from `SVI.__init__()`
- [x] #1020 update all tests
- [x] #1021 update all examples
- [x] #1022 update all tutorials
- [ ] #1054 remove support for `loss` string arg from `SVI.__init__()`
- [ ] #1054 remove `ELBO.make()` factory

fritzo:
Resolves #1014

### DIFF
--- a/pyro/contrib/autoguide/advi.py
+++ b/pyro/contrib/autoguide/advi.py
@@ -74,7 +74,7 @@ class AutoGuideList(AutoGuide):
         guide = AutoGuideList(my_model)
         guide.add(AutoDiagonalNormal(poutine.block(model, hide=["assignment"])))
         guide.add(AutoDiscreteParallel(poutine.block(model, expose=["assignment"])))
-        svi = SVI(model, guide, optim, 'ELBO')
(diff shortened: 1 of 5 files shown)


fritzo:
should be ready to merge

PR #961, opened 2018-03-30
Remove torch distribution wrappers and rename parameters

neerajprad:
Addresses distribution renaming from #827:
 - Removes all distribution wrappers.
 - Renames distribution parameters in tests, examples and tutorials.
 - Also re-enables smoke test on `gp.ipynb` but disables `pyro.validation` (see #953).

This was done using a combination of automated replacements, and manual vetting. #961

PR #1055, opened 2018-04-21
Binomial distribution with vectorized total_count

neerajprad:
This moves the changes proposed in https://github.com/pytorch/pytorch/pull/6720 to Pyro, as they will not make it to the `0.4` release. In the interim, we can have a Binomial distribution with vectorized `total_count` inside of Pyro. 

**Note:** I could not extend `torch.distributions.Binomial` because the semantics are quite different - the constructor (`__init__`), `sample`, `log_prob`, `batch_shape`, `enumerate_support` and `arg_constraints` are different. Hence calling the `super.__init__` fails validation checks.

cc. @martinjankowiak


pyro/distributions/torch.py
```diff
@@ -2,6 +2,7 @@
 
 import torch
 
+from pyro.distributions.binomial import Binomial  # noqa: F403
```
fritzo: It seems weird to double inherit from `TorchDistributionMixin`. I think cleaner approaches would be to either
1. move `from pyro.distributions.torch import *` higher up in pyro/distributions/__init__.py so that `pyro.distributions.Binomial` refers to your new `Binomial`, or
2. blacklist `Binomial` from the loop in `pyro.distributions.torch`

pyro/distributions/torch.py
```diff
@@ -16,6 +17,9 @@
 
     try:
         _PyroDist = locals()[_name]
+        # Do not override existing __doc__
```
fritzo: This seems difficult to maintain. I think it would be simpler to
`if _name == 'Binomial': continue`.

neerajprad: I am fine with that. However, if the class has an existing `__doc__`, it does not seem right to override that. Maybe we should append to it?

neerajprad: Actually, lets get to it when we have more distributions that are overriding `__doc__`. Will change.

pyro/distributions/torch.py
neerajprad: I think 2. might better. Will fix.

pyro/distributions/torch.py
fritzo: The only distributions that should be in `locals()` should be distributions defined in this file, and those should be mere thin wrappers. As of your #961 there are no remaining wrappers, so this logic is unused. If we add wrappers in the future, I think we would not add a docstring. But I defer to your taste :smile:

neerajprad:
Thanks for reviewing this, @fritzo!

Issue #1041, opened 2018-04-18
GP related checklist


fehiepsi:
From some discussions in #1039, I would like to make the following (possible) changes before Pyro 0.2 if there is no objection:
- [ ] Rename SparseVariationalGP to VariationalSparseGP
- [ ] Rename SparseMultivariateNormal to LowRankMultivariateNormal
- [ ] Rename GPRegression, SparseGPRegression (cc @martinjankowiak )
- [ ] Rename `approx` argument of SparseGPRegression to `method`.
- [ ] Rename the flag `noiseless` in regression models to (might be) `with_noise`
- [ ] Update triangular solver because we have gpu `torch.trtrs` now in pytorch master (might remove the confusing `pack` and `unpack` method for triangular solver as previously used in gp module and use trtrs directly).
- [ ] Enable validation check and rerun bo tutorial.
- [ ] Add timer for sv-dkl example to record running time (to observe performance changes after some commits from pyro and pytorch).

martinjankowiak:
i don't have strong feelings about most of these but you can probably leave `SparseVariationalGP` as is. 
`SparseMultivariateNormal` seems like the one that needs changing most

fritzo:
@fehiepsi We're aiming to release mid-week next week.
- :+1: on any renaming you'd like to do
- if you have time to implement `torch.trtrs` by Monday then fine, otherwise it's fine to delay till 0.4.1
- :+1: on ebling validation in bo tutorial
- no need to add a timer before release, i think it's more important to focus on interface

PR #1057, opened 2018-04-21
Update GP related for 0.2

### LINKED ISSUE #1051: Tests giving incorrect results on CUDA
I am trying to fix tests that don't work with CUDA..e.g. by adjusting precision, changing functions that do not have CUDA implementations, etc. 

Following is a working list of tests that are giving incorrect results, and which seem to have deeper issues:
 - [ ] #1057 `tests/contrib/gp/test_models.py::test_inference` - Results are too inaccurate. cc.@fehiepsi

fehiepsi:
This pull request addresses #1041 with the following changes:

+ Rename SparseVariationalGP to VariationalSparseGP (I did a search for "Variational Sparse Gaussian Process" and "Sparse Variational Gaussian Process" and found that the latter term is not used in literature.)
+ Rename SparseMultivariateNormal to LowRankMultivariateNormal
+ Add timer for sv-dkl.py example
+ Update regression docs to be more explicit about Gaussian noise
+ Use `torch.trtrs` for gpu triangle_solve. I didn't delete `matrix_triangular_solve_compat` because I saw that it is still useful to solve A X = b, where b can be a matrix or a vector. In addition, removing it requires changing much more code, which is unnecessary right now.
+ Fix #1051. cc @neerajprad 
+ Enable validation for bo tutorial. When rerun the tutorial, I found that LBFGS is not stable so I filed an [issue](https://github.com/pytorch/pytorch/issues/6838) and reduced the num_steps from 10 to 8. In addition, I also remove smoke_test env because this tutorial is quite fast to run.
+ Use whiten flag for vsgp model in gp tutorial to avoid noisy loss. cc @martinjankowiak 
+ Run "make scrub" for svi tutorials.

- [x] ~~Detect reason for performance loss in sv-dkl~~ my bad, the data is overlapped by another data.


tests/contrib/gp/test_models.py
```diff
@@ -180,12 +178,9 @@ def test_inference(model_class, X, y, kernel, likelihood):
...
 
-    variance = gp.kernel.get_param("variance")
-    lengthscale = gp.kernel.get_param("lengthscale")
-    target_variance = kernel.get_param("variance")
-    target_lengthscale = kernel.get_param("lengthscale")
-    assert_equal(variance, target_variance, prec=0.2)
-    assert_equal(lengthscale, target_lengthscale, prec=0.2)
+    y_cov = gp.kernel(X)
+    target_y_cov = kernel(X)
+    assert_equal(y_cov, target_y_cov, prec=0.1)
```
fehiepsi: @neerajprad This is a more reasonable test due to two different kernels can have the same output.

pyro/contrib/gp/models/gpr.py
```diff
@@ -39,7 +39,7 @@ class GPRegression(GPModel):
         f &\sim \mathcal{GP}(0, k(X, X)),\\
         y & \sim f + \epsilon,
 
-    where :math:`\epsilon` is noise.
+    where :math:`\epsilon` is a Gaussian noise.
```
martinjankowiak: it needs to be clear (line 59) whether `noise` is variance or square root variance

fehiepsi: Yes, I changed it too "Variance of Gaussian noise of this model."

fehiepsi:
@martinjankowiak I just run `make scrub` to make all tutorial consistency. Someone forgot to run it along the way, so we get some notebooks used python 2.7.13, while the others used 2.7.14.

pyro/contrib/gp/kernels/kernel.py
```diff
@@ -316,10 +316,11 @@ class Warping(Transforming):
     learning architecture. For example:
 
         >>> linear = torch.nn.Linear(10, 3)
-        # register its parameters to Pyro's ParamStore
-        >>> pyro_linear = pyro.module("linear", linear)
+        # register its parameters to Pyro's ParamStore and wrap it by lambda
+        # to call the primitive pyro.module each time we use the linear function
+        >>> pyro_linear_fn = lambda x: pyro.module("linear", linear)(x)
```
fritzo: Interesting. I guess this is only needed if `pyro_linear_fn` is defined outside of `model`?

pyro/contrib/gp/models/vgp.py
```diff
@@ -149,10 +149,10 @@ def forward(self, Xnew, full_cov=False):
         :rtype: tuple(torch.Tensor, torch.Tensor)
         """
         self._check_Xnew_shape(Xnew)
-        tmp_sample_latent = self._sample_latent
+        # avoid sampling the unnecessary latent f
         self._sample_latent = False
         f_loc, f_scale_tril = self.guide()
-        self._sample_latent = tmp_sample_latent
+        self._sample_latent = True
```
fritzo: Can you confirm that `self._sample_latent` is always `True` when entering this function? If that is not the case, consider using the save-and-restore pattern
```py
saved_sample_latent = self._sample_latent
self._sample_latent = False
f_loc, f_scale_tril = self.guide()
self._sample_latent = saved_sample_latent
```

pyro/contrib/gp/models/vsgp.py
```diff
@@ -173,10 +173,10 @@ def forward(self, Xnew, full_cov=False):
         :rtype: tuple(torch.Tensor, torch.Tensor)
         """
         self._check_Xnew_shape(Xnew)
-        tmp_sample_latent = self._sample_latent
+        # avoid sampling the unnecessary latent u
         self._sample_latent = False
         Xu, u_loc, u_scale_tril = self.guide()
-        self._sample_latent = tmp_sample_latent
+        self._sample_latent = True
```
fritzo: ditto: can you confirm that `self._sample_latent` is always `True`, or otherwise use save-and-restore?

pyro/contrib/gp/kernels/kernel.py
fehiepsi: Yes. It is required when arguments are functions and those functions are `nn.Module`s. In case there is no parameter in that function or that function is a Parameterized class, we don't have to use lambda wrapper. :)

pyro/contrib/gp/models/vgp.py
fehiepsi: `self._sample_latent` should always be `True` unless we want to set it `False` to achieve something. This is the only place we change its value so `self._sample_latent` will always be True when entering this function (unless users do something like `gpmodel._sample_latent = False` (I don't know why they want to do it and even if they do, this place will set `self._sample_latent` back to its `True` value).

PR #1060, opened 2018-04-23
Improve support for custom losses in SVI

eb8680:
This PR removes a bit of cruft in SVI so that it's easier to work with custom losses.  In particular, it:
1. Removes `ParamStoreDict.mark_param_active()` and related machinery
2. Replaces that machinery with a `poutine.trace` in `SVI.step` so that users can simply define a loss and not worry about moving parameters around
3. Adds a `param_only` flag to `TraceMessenger` that allows `poutine.trace` to record only parameter sites
4. Removes the restriction in `SVI` requiring `loss` to be an `ELBO` object
5. Restores a default value for `SVI.loss_and_grads` (calling backward on the result of `SVI.loss`)

The upshot is that now, to create a custom loss, all that's needed is to define it as a function that takes in a model, guide, and any other arguments necessary to compute it and returns a differentiable scalar loss value.  For example:
```python
def simple_elbo(model, guide, *args, **kwargs):
    guide_trace = poutine.trace(guide).get_trace(*args, **kwargs)
    model_trace = poutine.trace(
        poutine.replay(model, trace=guide_trace)).get_trace(*args, **kwargs)
    return -1*(model_trace.log_prob_sum() - guide_trace.log_prob_sum())

svi = SVI(model, guide, optim, loss=simple_elbo)
losses = []
for i in range(N):
    losses.append(svi.step())
```

### DIFF
--- a/pyro/infer/svi.py
+++ b/pyro/infer/svi.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function
 import torch
 
 import pyro
+import pyro.poutine as poutine
 from pyro.infer.elbo import ELBO
 
 
@@ -72,7 +73,15 @@ class SVI(object):
         Any args or kwargs are passed to the model and guide
         """
         # get loss and compute gradients
-        loss = self.loss_and_grads(self.model, self.guide, *args, **kwargs)
+        param_captured_loss = poutine.trace(poutine.block(self.loss_and_grads,
+                                                          expose_types=["param"]))
+        loss = param_captured_loss(self.model, self.guide, *args, **kwargs)
+
+        trainable_params = set(site["value"].unconstrained()
+                               for site in param_captured_loss.trace.nodes.values()
+                               if site["type"] == "param")
+
+        pyro.get_param_store().mark_params_active(trainable_params)
 
         # get active params
         params = pyro.get_param_store().get_active_params()
--- a/pyro/infer/svi.py
+++ b/pyro/infer/svi.py
@@ -5,6 +5,7 @@ import torch
 import pyro
 import pyro.poutine as poutine
 from pyro.infer.elbo import ELBO
+from pyro.infer.util import torch_item
 
 
 class SVI(object):
@@ -67,7 +68,7 @@ class SVI(object):
         Evaluate the loss function. Any args or kwargs are passed to the model and guide.
         """
         with torch.no_grad():
-            return self.loss(self.model, self.guide, *args, **kwargs)
+            return torch_item(self.loss(self.model, self.guide, *args, **kwargs))
 
     def step(self, *args, **kwargs):
         """
@@ -92,4 +93,4 @@ class SVI(object):
         # zero gradients
         pyro.infer.util.zero_grads(params)
 
-        return loss
+        return torch_item(loss)
--- a/pyro/infer/svi.py
+++ b/pyro/infer/svi.py
@@ -73,18 +73,11 @@ class SVI(object):
         Any args or kwargs are passed to the model and guide
         """
         # get loss and compute gradients
-        param_captured_loss = poutine.trace(poutine.block(self.loss_and_grads,
-                                                          expose_types=["param"]))
-        loss = param_captured_loss(self.model, self.guide, *args, **kwargs)
+        with poutine.trace(param_only=True) as param_capture:
+            loss = self.loss_and_grads(self.model, self.guide, *args, **kwargs)
 
-        trainable_params = set(site["value"].unconstrained()
-                               for site in param_captured_loss.trace.nodes.values()
-                               if site["type"] == "param")
-
-        pyro.get_param_store().mark_params_active(trainable_params)
-
-        # get active params
-        params = pyro.get_param_store().get_active_params()
+        params = set(site["value"].unconstrained()
+                     for site in param_capture.trace.nodes.values())
 
         # actually perform gradient steps
         # torch.optim objects gets instantiated for any params that haven't been seen yet
@@ -92,8 +85,5 @@ class SVI(object):
 
         # zero gradients
         pyro.infer.util.zero_grads(params)
-
-        # mark parameters in the param store as inactive
-        pyro.get_param_store().mark_params_inactive(params)
 
         return loss
--- a/pyro/infer/svi.py
+++ b/pyro/infer/svi.py
@@ -38,11 +38,17 @@ class SVI(object):
         self.optim = optim
 
         if isinstance(loss, ELBO):
-            self.ELBO = loss
-            self.loss = self.ELBO.loss
-            self.loss_and_grads = self.ELBO.loss_and_grads
+            self.loss = loss.loss
+            self.loss_and_grads = loss.loss_and_grads
         else:
-            raise TypeError("Unsupported loss type. Expected an ELBO instance, got a {}".format(type(loss)))
+            if loss_and_grads is None:
+                def _loss_and_grads(*args, **kwargs):
+                    loss_val = loss(*args, **kwargs)
+                    loss_val.backward()
+                    return loss_val
+                loss_and_grads = _loss_and_grads
+            self.loss = loss
+            self.loss_and_grads = loss_and_grads
 
     def __call__(self, *args, **kwargs):
         """
@@ -61,7 +67,7 @@ class SVI(object):
         Evaluate the loss function. Any args or kwargs are passed to the model and guide.
         """
         with torch.no_grad():
-            return self.loss(self.model, self.guide, *args, **kwargs)
+            return self.loss.loss(self.model, self.guide, *args, **kwargs)
 
     def step(self, *args, **kwargs):
         """
@@ -74,7 +80,7 @@ class SVI(object):
         """
         # get loss and compute gradients
         with poutine.trace(param_only=True) as param_capture:
-            loss = self.loss_and_grads(self.model, self.guide, *args, **kwargs)
+            loss = self.loss.loss_and_grads(self.model, self.guide, *args, **kwargs)
 
         params = set(site["value"].unconstrained()
                      for site in param_capture.trace.nodes.values())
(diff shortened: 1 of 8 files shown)


pyro/infer/svi.py
```diff
@@ -37,11 +38,17 @@ def __init__(self,
...
-            self.loss_and_grads = self.ELBO.loss_and_grads
+            self.loss = loss.loss
+            self.loss_and_grads = loss.loss_and_grads
         else:
-            raise TypeError("Unsupported loss type. Expected an ELBO instance, got a {}".format(type(loss)))
+            if loss_and_grads is None:
+                def _loss_and_grads(*args, **kwargs):
+                    loss_val = loss(*args, **kwargs)
+                    loss_val.backward()
+                    return loss_val
```
fritzo: In `ELBO` implementations, `.loss()` returns a float rather than a tensor. This makes it difficult to support automatic definition of `.loss_and_grads()`.

What do you think of defining a `Loss` base class from which `ELBO` inherits, and requiring that base class to implement either `.differentiable_loss` or both of `.loss()` and `.loss_and_grads()`? This would allow us to refactor `TraceEnum_ELBO` to expose our Dice loss:
```py
class Loss(object):
    def differentiable_loss(self, *args, **kwargs):
        raise NotImplementedError
    def loss(self, *args, **kwargs):
        return self.diffentiable_loss(*args, **kwargs).item()
    def loss_and_grads(self, *args, **kwargs):
        loss = self.differentiable_loss
        loss.backward()  # TODO support create_graph=True
        return loss.item()

class ELBO(Loss):
    # ...unchanged...

class TraceEnum_ELBO(ELBO):
    def differentiable_loss(self, *args, **kwargs):
        # ...simplified...
```
This seems like it also paves the way to supporting higher-order optimizers. Thoughts?

pyro/infer/trace_elbo.py
```diff
@@ -124,15 +123,14 @@ def loss_and_grads(self, model, guide, *args, **kwargs):
             elbo += elbo_particle / self.num_particles
 
             # collect parameters to train from model and guide
-            trainable_params = set(site["value"].unconstrained()
-                                   for trace in (model_trace, guide_trace)
-                                   for site in trace.nodes.values()
-                                   if site["type"] == "param")
+            trainable_params = len(set(site["value"].unconstrained()
```
fritzo: Let's simplify to
```py
trainable_params = any(site["type"] == param
                       for trace in (model_trace, guide_trace)
                       for site in trace.nodes.values())
```

pyro/infer/traceenum_elbo.py
```diff
@@ -128,15 +127,14 @@ def loss_and_grads(self, model, guide, *args, **kwargs):
             elbo += elbo_particle.item() / self.num_particles
 
             # collect parameters to train from model and guide
-            trainable_params = set(site["value"].unconstrained()
-                                   for trace in (model_trace, guide_trace)
-                                   for site in trace.nodes.values()
-                                   if site["type"] == "param")
+            trainable_params = len(set(site["value"].unconstrained()
```
fritzo: ditto, simplify

pyro/infer/tracegraph_elbo.py
```diff
@@ -261,15 +261,14 @@ def _loss_and_grads_particle(self, weight, model_trace, guide_trace):
             surrogate_elbo += surrogate_elbo_term
 
         # collect parameters to train from model and guide
-        trainable_params = set(site["value"].unconstrained()
-                               for trace in (model_trace, guide_trace)
-                               for site in trace.nodes.values()
-                               if site["type"] == "param")
+        trainable_params = len(set(site["value"].unconstrained()
```
fritzo: ditto: simplify

pyro/poutine/__init__.py
```diff
@@ -23,10 +23,11 @@
 ############################################
 
 
-def trace(fn=None, graph_type=None):
+def trace(fn=None, graph_type=None, param_only=None):
```
fritzo: This arg is fine for now but would be awkward if we ever generalize. Do you think it would be reasonable to replace it with `sample_sites=True, param_sites=True`?

Also, do I understand correctly that even with `param_only=True`, the `_INPUT` and `_RETURN` sites will be recorded?

Also, why not simply `block` or filter to `{name: site for site in trace.nodes.items() if site['type'] == 'param'}`?

pyro/poutine/trace_poutine.py
```diff
@@ -159,11 +170,11 @@ def __call__(self, *args, **kwargs):
         stores the arguments and return value of the function in special sites,
         and returns self.fn's return value
         """
-        self.msngr.trace = Trace(graph_type=self.msngr.graph_type)
-        self.msngr.trace.add_node("_INPUT",
-                                  name="_INPUT", type="args",
-                                  args=args, kwargs=kwargs)
-        ret = super(TraceHandler, self).__call__(*args, **kwargs)
+        with self.msngr:
```
fritzo: Interesting, does this fix a bug related to exception handling? I don't understand the difference between old and new versions.

pyro/poutine/trace_poutine.py
```diff
@@ -41,16 +41,24 @@ class TraceMessenger(Messenger):
...
         if graph_type is None:
             graph_type = "flat"
+        if param_only is None:
+            param_only = False
         assert graph_type in ("flat", "dense")
         self.graph_type = graph_type
+        self.param_only = param_only
+        self.trace = Trace(graph_type=self.graph_type)
+
+    def __enter__(self):
```
fritzo: Does this require any new usage patterns, other than the one use in `poutine.trace` below?

pyro/infer/trace_elbo.py
```diff
@@ -124,15 +123,14 @@ def loss_and_grads(self, model, guide, *args, **kwargs):
...
+                                       for trace in (model_trace, guide_trace)
+                                       for site in trace.nodes.values()
+                                       if site["type"] == "param")) > 0
```
jpchen: nit: i think` > 0` is redundant here (and ones below)

pyro/poutine/trace_poutine.py
```diff
@@ -41,16 +41,24 @@ class TraceMessenger(Messenger):
     We can also use this for visualization.
     """
 
-    def __init__(self, graph_type=None):
+    def __init__(self, graph_type=None, param_only=None):
         """
         :param string graph_type: string that specifies the type of graph
```
jpchen: add `param_only` to docstring

pyro/infer/svi.py
```diff
@@ -72,10 +79,11 @@ def step(self, *args, **kwargs):
         Any args or kwargs are passed to the model and guide
         """
         # get loss and compute gradients
-        loss = self.loss_and_grads(self.model, self.guide, *args, **kwargs)
+        with poutine.trace(param_only=True) as param_capture:
+            loss = self.loss_and_grads(self.model, self.guide, *args, **kwargs)
 
-        # get active params
-        params = pyro.get_param_store().get_active_params()
+        params = set(site["value"].unconstrained()
```
fritzo: :+1:

pyro/infer/svi.py
eb8680: > In ELBO implementations, .loss() returns a float rather than a tensor.

Why do we need to do this instead of just using `with torch.no_grad()`?

pyro/poutine/__init__.py
eb8680: > This arg is fine for now but would be awkward if we ever generalize. Do you think it would be reasonable to replace it with sample_sites=True, param_sites=True?

I don't see a use case for `sample_sites=True, param_sites=False`, so I don't know why we'd need to support that.

> Also, why not simply block 

This version has no effect on any enclosing code

> or filter

This way it's fine to wrap around something where a sample site would otherwise appear multiple times, e.g. a multi-particle ELBO where the model and guide are invoked repeatedly.

pyro/infer/svi.py
fritzo: WDYT of the `Loss` base class that would help simplify `TraceEnum_ELBO`?

There are a couple reasons we're currently returning float-valued loss:
1. floats are easier to deal with when e.g. printing, and that's all that can be done with them by the time they're returned from a `.loss()` function (`.loss()` need not guarantee differentiability, and `.loss_and_grads()` need not create an autograd graph)
2. python floats have higher precision (64 bits) than typical tensors (32bits), so we can simplify some of the aggregation in `Trace_ELBO`

pyro/poutine/trace_poutine.py
eb8680: This uses `self.msngr.__enter__` to reset `self.msngr.trace`, then adds the `_INPUT` site after the reset happens, allowing `poutine.trace` to be used correctly as a context manager.

pyro/poutine/trace_poutine.py
eb8680: This is necessary to reset `self.trace` so that `poutine.trace` can be used as a context manager

pyro/poutine/trace_poutine.py
fritzo: Also, `__init__` methods typically do not have docstrings, instead this information should be at the end of the class docstring (unrelated to this PR, but might as well change since you're editing the docstring).

jpchen: oh right i didnt even notice that

eb8680:
> if i want to only optimize a certain set of parameters, i would have to block out those sites with poutine.block instead of marking them as inactive right

Yeah, that should just work now as well.

pyro/infer/svi.py
fritzo: 3. users would often forget to call `.detach()`. They then run out of memory in the training loop via `losses.append(svi.step())`, as autograd graphs are never released. If we support custom loss functions w/o a `Loss` base class, we should at least make `SVI.step()` safer by
    ```py
    return loss.item() if torch.is_tensor(loss) else loss
    ```

eb8680: > WDYT of the Loss base class that would help simplify TraceEnum_ELBO?

Just wrapped the return values of `SVI.step` and `SVI.evaluate_loss` with `torch_item`, should that accomplish the same thing?

PR #1061, opened 2018-04-23
Update API usage in language intro tutorials

eb8680:
Addresses the language intro portion of #930 

This PR updates the language intro tutorials to be compatible with the API introduced in #1019 by using `TracePosterior.run` and `EmpiricalMarginal` instead of `Marginal`.


tutorial/source/intro_part_i.ipynb
```diff
@@ -3,7 +3,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "collapsed": true
+   },
```
fehiepsi: Nit: you can remove these metadata by using `make scrub` :)

pyro/infer/svi.py
```diff
@@ -51,15 +51,6 @@ def _loss_and_grads(*args, **kwargs):
             self.loss = loss
             self.loss_and_grads = loss_and_grads
 
-    def __call__(self, *args, **kwargs):
```
martinjankowiak: why is this being removed here?

eb8680: We don't want to support this API, and I remembered it while going over the tutorial

tutorial/source/intro_part_ii.ipynb
```diff
@@ -346,7 +360,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "collapsed": true
+   },
```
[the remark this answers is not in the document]
eb8680: It's all runnable, but none of these cells take more than a couple seconds so I think it's fine

tutorial/source/intro_part_i.ipynb
```diff
@@ -163,12 +161,12 @@
     "    if t is None:\n",
     "        t = 0\n",
     "    x = pyro.sample(\"x_{}\".format(t), dist.Bernoulli(p))\n",
-    "    if torch.equal(x.data, torch.zeros(1)):\n",
+    "    if x.item() == 0:\n",
```
fritzo: nit: actually `if x == 0:` now works even without `.item()`

PR #1062, opened 2018-04-23
add clipped nn to docs

jpchen:
this was omitted from the docs. it's still used in examples like ss-vae

@fritzo are these clipped non-linearities still needed or can this be handled by pytorch 0.4?

### DIFF
--- a/docs/source/pyro.nn.txt
+++ b/docs/source/pyro.nn.txt
@@ -6,3 +6,10 @@ AutoRegressiveNN
     :undoc-members:
     :show-inheritance:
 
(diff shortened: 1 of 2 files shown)


fritzo:
> can this be handled by pytorch 0.4?

As far as I know PyTorch 0.4 doesn't solve all NAN issues, so these are still needed.

Would you mind also updating the docstrings to link to the PyTorch docs:
```py
"""
... :class:`torch.nn.Softmax` ...
"""
```

jpchen:
done

PR #1064, opened 2018-04-23
Rename all remaining poutine files and modules to messenger

eb8680:
This PR essentially renames all files and modules named `*_poutine` to `*_messenger`.  It doesn't change any code except imports and filenames, and a couple of internal variable names.


martinjankowiak:
don't worry the effect handler library is still called poutine ; )

Issue #980, opened 2018-04-07
Better code organization and removing circular imports


neerajprad:
- [ ] Remove circular imports from modules.
 - [ ] Better organization of utilities - move common utilities used across modules into its own module.

Other related tasks - #827, #930

fehiepsi:
Some of my thoughts on this :)
+ move ops to mcmc.util
+ move nn to a better place
+ move memoize from pyro.util to pyro.infer.util
+ move pyro.shim back to pyro.util

PR #1066, opened 2018-04-23
Move some pyro internals into poutine.runtime

eb8680:
This PR cleans up the global namespace and utilities by moving a bunch of global variables and core handler stack processing that are not user-facing into `pyro/poutine/runtime.py`.

Part of #980


eb8680:
Will push documentation in a followup PR after #1070 is merged

PR #1067, opened 2018-04-23
Small tweaks to sampling from model in vae

jpchen:
Uses the poutine mechanism to extract the model latent, instead of actually running the decoder

it's certainly fewer lines of code, though more machinery to work with to extract something you can do with a helper function. not sure what the 'right' approach is


examples/vae/utils/vae_plots.py
```diff
@@ -5,16 +5,19 @@ def plot_conditional_samples_ssvae(ssvae, visdom_session):
     """
     This is a method to do conditional sampling in visdom
     """
+    from pyro import poutine
```
fritzo: Why not move these imports to the top?

jpchen:
> return the bernoulli probs as one of its return values 

yeah i like that better, though it sort of violates the idea of model taking parameters and returning data

martinjankowiak:
there is no such contract

jpchen:
it's the right way to think about generative models imo. in any case, i updated the PR to reflect that

examples/vae/utils/vae_plots.py
```diff
@@ -1,3 +1,10 @@
+import matplotlib.pyplot as plt
```
fritzo: This breaks the `.use('Agg')` behavior (as discussed offline)

PR #1073, opened 2018-04-23
fix tutorials so they render properly in html

jpchen:
misc changes here and there. verified by inspection. if anyone has time i would recommend walking through and clicking on all the links, etc.


eb8680:
Links seem fine to me

PR #1074, opened 2018-04-24
Sort imports via isort

fritzo:
This PR
1. configures `make format` to run isort (but not yapf)
2. runs `make format`, which then sorts a bunch of headers


Makefile
```diff
@@ -22,8 +22,7 @@ scrub: FORCE
 	find tutorial -name "*.ipynb" | xargs python tutorial/source/cleannb.py
 
 format: FORCE
-	yapf -i *.py pyro/distributions/*.py profiler/*.py docs/source/conf.py
-	isort --recursive *.py pyro/ tests/ profiler/*.py docs/source/conf.py
+	isort --recursive *.py pyro/ examples/ tests/ profiler/*.py docs/source/conf.py
```
fritzo: These are the only two lines I modified by hand. The rest of this diff was automatic.

jpchen: luckilyy for me those are the only two lines i read

examples/contrib/gp/sv-dkl.py
```diff
@@ -21,13 +21,13 @@
...
 import torch.nn as nn
 import torch.nn.functional as F
-from torchvision import transforms
 
 import pyro
 import pyro.contrib.gp as gp
 import pyro.infer as infer
 import pyro.optim as optim
 from pyro.contrib.examples.util import get_data_loader
+from torchvision import transforms
```
fehiepsi: This place for `torchvision` is strange to me.

examples/vae/utils/mnist_cached.py
```diff
@@ -5,8 +5,8 @@
 import numpy as np
 import torch
 from torch.utils.data import DataLoader
-from torchvision.datasets import MNIST
 
+from torchvision.datasets import MNIST
```
fehiepsi: similar to above

examples/vae/vae_comparison.py
```diff
@@ -7,14 +7,14 @@
...
 from torch.nn import functional
-from torchvision.utils import save_image
 
 import pyro
-from utils.mnist_cached import RESULTS_DIR, DATA_DIR
 from pyro.contrib.examples import util
 from pyro.distributions import Bernoulli, Normal
 from pyro.infer import SVI, Trace_ELBO
 from pyro.optim import Adam
+from torchvision.utils import save_image
```
fehiepsi: same here

fritzo:
should be ready to merge

PR #1077, opened 2018-04-24
Minor cleanups for HMC

neerajprad:
Some minor changes for HMC/NUTS as I was working on the example:
 - Making them available for imports directly from `pyro.infer.mcmc`.
 - Some changes to make kernels reusable for more than 1 run - e.g. not mutating `self.adapt_step_size` but instead using another attribute to designate adaptation phase.

### DIFF
--- a/pyro/infer/mcmc/__init__.py
+++ b/pyro/infer/mcmc/__init__.py
@@ -1,0 +1,9 @@
+from pyro.infer.mcmc.hmc import HMC
+from pyro.infer.mcmc.mcmc import MCMC
+from pyro.infer.mcmc.nuts import NUTS
+
+__all__ = [
+    "HMC",
+    "MCMC",
+    "NUTS",
+]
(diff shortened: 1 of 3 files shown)


pyro/infer/mcmc/hmc.py
```diff
@@ -100,6 +100,7 @@ def _reset(self):
         self._args = None
         self._kwargs = None
         self._prototype_trace = None
+        self._adapt_phase = False
```
fehiepsi: +1 for this flag

pyro/infer/mcmc/__init__.py
```diff
@@ -0,0 +1,9 @@
+from pyro.infer.mcmc.hmc import HMC
+from pyro.infer.mcmc.mcmc import MCMC
+from pyro.infer.mcmc.nuts import NUTS
+
+__all__ = [
+    "HMC",
+    "MCMC",
+    "NUTS",
+]
```
fehiepsi: yeah, I wanted to do this at some moment ago but forgot to do.

PR #1079, opened 2018-04-24
Add posterior predictive class to AbstractInfer

neerajprad:
Adds a posterior predictive class that allows us to generate posterior predictive samples given model execution traces and new/old data. These can be wrapped inside marginal to provide an empirical distribution over samples from the posterior predictive distribution.

### DIFF
--- a/pyro/infer/abstract_infer.py
+++ b/pyro/infer/abstract_infer.py
@@ -6,17 +6,21 @@ from six import add_metaclass
 import torch
 
 import pyro.poutine as poutine
-from pyro.distributions import Empirical
+from pyro.distributions import Categorical, Empirical
 
 
 class EmpiricalMarginal(Empirical):
     """
-    :param trace_dist: a TracePosterior instance representing a Monte Carlo posterior.
-
     Marginal distribution, that wraps over a TracePosterior object to provide a
     a marginal over one or more latent sites or the return values of the
     TracePosterior's model. If multiple sites are specified, they must have the
     same tensor shape.
+
+    :param TracePosterior trace_posterior: a TracePosterior instance representing
+        a Monte Carlo posterior.
+    :param list sites: optional list of sites for which we need to generate
+        the marginal distribution for. Note that for multiple sites, the shape
+        for the site values must match.
     """
     def __init__(self, trace_posterior, sites=None, validate_args=None):
         assert isinstance(trace_posterior, TracePosterior), \
@@ -71,3 +75,43 @@ class TracePosterior(object):
             self.exec_traces.append(tr)
             self.log_weights.append(logit)
         return self
+
+
+class PosteriorPredictive(TracePosterior):
+    """
+    Generates and holds traces from the posterior predictive distribution,
+    given model execution traces from the approximate posterior.
+
+    :param model: arbitrary Python callable containing Pyro primitives.
+    :param model_traces: execution traces from the model.
+    :param int num_samples: number of samples to generate.
+    :param list hide_nodes: list of nodes that should be hidden when
+        replaying the model against ``model_traces``. Usually, this
+        corresponds to observed nodes.
+    :param list log_weights: optional importance weights of execution
+        traces to be used during sampling.
+    """
+    def __init__(self, model, model_traces, num_samples, hide_nodes=[], log_weights=None):
+        self.model = model
+        self.model_traces = model_traces
+        self.hide_nodes = hide_nodes
+        self.num_samples = num_samples
+        if not log_weights:
+            log_weights = torch.zeros(len(model_traces,))
+        else:
+            log_weights = torch.tensor(log_weights)
+        self._categorical = Categorical(logits=log_weights)
+        super(PosteriorPredictive, self).__init__()
+
+    def _get_random_pruned_trace(self):
+        random_idx = self._categorical.sample()
+        trace = self.model_traces[random_idx].copy()
+        for node in self.hide_nodes:
+            trace.remove_node(node)
+        return trace
+
+    def _traces(self, *args, **kwargs):
+        for _ in range(self.num_samples):
+            model_trace = self._get_random_pruned_trace()
+            replayed_trace = poutine.trace(poutine.replay(self.model, model_trace)).get_trace(*args, **kwargs)
+            yield (replayed_trace, 0)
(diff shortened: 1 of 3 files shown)


pyro/infer/abstract_infer.py
```diff
@@ -71,3 +76,46 @@ def run(self, *args, **kwargs):
...
+        random_idx = self._categorical.sample()
+        trace = self.model_traces[random_idx].copy()
+        for node in self.hide_nodes:
+            trace.remove_node(node)
+        return trace
+
+    def _traces(self, *args, **kwargs):
+        for _ in range(self.num_samples):
+            model_trace = self._get_random_pruned_trace()
+            replayed_trace = poutine.trace(poutine.replay(self.model, model_trace)).get_trace(*args, **kwargs)
```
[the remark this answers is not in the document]
neerajprad: Passing this explicitly makes it easier to generate predictions on new data though. Otherwise, we would have to rewrite the same machinery separately.

pyro/infer/abstract_infer.py
```diff
@@ -71,3 +76,46 @@ def run(self, *args, **kwargs):
...
+    :param int num_samples: number of samples to generate.
+    :param list hide_nodes: list of nodes that should be hidden when
+        replaying the model against ``model_traces``. Usually, this
+        corresponds to observed nodes.
+    :param list log_weights: optional importance weights of execution
+        traces to be used during sampling.
+    """
+    def __init__(self, model, model_traces, num_samples, hide_nodes=[], log_weights=None):
+        self.model = model
+        self.model_traces = model_traces
```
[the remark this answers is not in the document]
neerajprad: I suppose the issue there (at least for HMC) is that you would not want to run your model again which will be way more expensive than sampling existing traces. I wonder if we can bypass `self.posterior.run` if execution traces are already populated.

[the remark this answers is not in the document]
neerajprad: Let me try something and update the PR. Thanks for the suggestion!

pyro/infer/abstract_infer.py
```diff
@@ -71,3 +76,46 @@ def run(self, *args, **kwargs):
...
+class PosteriorPredictive(TracePosterior):
+    """
+    Generates and holds traces from the posterior predictive distribution,
+    given model execution traces from the approximate posterior. This is
+    achieved by constraining latent sites to randomly sampled parameter
+    values from the model execution traces and running the model forward
+    to generate traces with new response ("_RETURN") sites.
+
+    :param model: arbitrary Python callable containing Pyro primitives.
+    :param model_traces: execution traces from the model.
```
fritzo: An iterable? A list? An unbounded generator?

pyro/infer/abstract_infer.py
```diff
@@ -71,3 +76,46 @@ def run(self, *args, **kwargs):
...
+    :param list log_weights: optional importance weights of execution
+        traces to be used during sampling.
+    """
+    def __init__(self, model, model_traces, num_samples, hide_nodes=[], log_weights=None):
+        self.model = model
+        self.model_traces = model_traces
+        self.hide_nodes = hide_nodes
+        self.num_samples = num_samples
+        if not log_weights:
+            log_weights = torch.zeros(len(model_traces,))
```
fritzo: `len(model_traces)` (remove stray comma?)

pyro/infer/abstract_infer.py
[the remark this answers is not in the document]
fritzo: I agree it would be nice to eliminate `log_weights`.

pyro/infer/abstract_infer.py
```diff
@@ -71,3 +76,46 @@ def run(self, *args, **kwargs):
...
+    :param list hide_nodes: list of nodes that should be hidden when
+        replaying the model against ``model_traces``. Usually, this
+        corresponds to observed nodes.
+    :param list log_weights: optional importance weights of execution
+        traces to be used during sampling.
+    """
+    def __init__(self, model, model_traces, num_samples, hide_nodes=[], log_weights=None):
+        self.model = model
+        self.model_traces = model_traces
+        self.hide_nodes = hide_nodes
```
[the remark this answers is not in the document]
eb8680: Can we just use `condition` or `obs` on `self.model` before it's passed in?

pyro/infer/abstract_infer.py
neerajprad: Removed.

pyro/infer/abstract_infer.py
neerajprad: Removed.

pyro/infer/abstract_infer.py
```diff
@@ -58,6 +64,10 @@ def _traces(self, *args, **kwargs):
         """
         raise NotImplementedError("inference algorithm must implement _traces")
 
+    def __call__(self, *args, **kwargs):
+        random_idx = self._categorical.sample()
+        return self.exec_traces[random_idx].copy()
```
eb8680: It should be fine to prune observe sites here rather than in `TracePredictive`

pyro/infer/abstract_infer.py
```diff
@@ -70,4 +80,43 @@ def run(self, *args, **kwargs):
...
+        self.model = model
+        self.posterior = posterior
+        self.hide_nodes = hide_nodes
+        self.num_samples = num_samples
+        super(TracePredictive, self).__init__()
+
+    def _get_random_pruned_trace(self):
+        trace = self.posterior()
+        for node in self.hide_nodes:
+            trace.remove_node(node)
```
eb8680: I think you can just remove this method.  You can prune observed sites in `TracePosterior.__call__` and include observed sites in `TracePredictive` by adding observations to `self.model` outside of `TracePredictive`.

pyro/infer/abstract_infer.py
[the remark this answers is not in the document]
neerajprad: Sorry, missed this comment earlier. That's a good point. Let me check.

pyro/infer/abstract_infer.py
neerajprad: Yup, makes sense. Let me modify this and get the examples working.

neerajprad: Removed obs sites by default. This looks much nicer.

Issue #826, opened 2018-02-27
Improve documentation of some Pyro internals


eb8680:
This issue tracks some documentation improvements that should be made before the 0.2 release.
- [x] Improve documentation of `Marginal` or remove it (#543)
- [x] Improve documentation of `TracePosterior` or remove it
- [x] Improve documentation of `Trace` 
- [ ] Correct and improve introductory documentation for Poutine
- [x] Format everything correctly for sphinx: handlers and `Trace` at top level, `Messenger`s and `runtime` collapsed

PR #1087, opened 2018-04-24
Move handlers into their own file and clean up poutine namespace

eb8680:
Part of #826 

This PR creates a file `pyro/poutine/handlers.py` with the definitions that used to be in `pyro/poutine/__init__.py` to clean up the `pyro.poutine` namespace.  It also removes a couple of `Messenger`s in `pyro/primitives.py` that should have been handlers.


eb8680:
> assuming this is purely refactoring

Yep, apart from the sphinx changes

eb8680:
Should be good to review/merge, I'll change docstring contents in a followup PR.

docs/source/poutine.rst
```diff
@@ -8,13 +8,28 @@ programs. Pyro's inference algorithms are all built by applying these handlers t
...
+
+.. automodule:: pyro.poutine.trace
+    :members:
+    :undoc-members:
+    :show-inheritance:
 
 Messengers
 -----------
 
 .. include:: pyro.poutine.txt
```
jpchen: not part of this PR, but this file should technically be changed to `pyro.poutine.messenger`

fritzo:
This definitely improves our code organization, but I think we should update our sphinx docs. I'd like to be able to refer to e.g. `poutine.trace` as `pyro.poutine.trace`. To do this I believe we'll need to replace the `automodule:: pyro.poutine.trace` with
```
.. autoclass pyro.poutine.trace
    :members:
    :undoc-members:
    :show-inheritance:
```
and similar with a number of our other imported handlers and messengers like `pyro.poutine.do` (or if you prefer, simply `autoclass:: pyro.do`. This is fine to address in a follow-up PR.

eb8680:
> To do this I believe we'll need to replace the `automodule:: pyro.poutine.trace`

I moved `trace.py` to `trace_struct.py`, does that help with this?

fritzo:
> I moved trace.py to trace_struct.py

That helps :smile: but it is orthogonal to the issue I described with sphinx. The issue is that when we use `automodule:: pyro.foo` sphinx determines the fully qualified name of each thing `x` in `foo` by examining `x.__module__`. Since we've moved `trace` into `pyro.poutine.handlers`, sphinx determines its fully qualified name to be `pyro.poutine.handlers.trace`. While this is technically correct, we prefer to  call it `pyro.poutine.trace`. To resolve this issue we can instruct sphinx of our desired name by replacing `automodule` with one entry per thing in that module, `autofunction` or `autoclass` depending on each thing. Sphinx will use whatever name we provide, so e.g. we can achieve `pyro.poutine.trace` by a one-liner
```
.. autofunction pyro.poutine.trace
```
It's a little less automatic, but we should be able to replace
```diff
- .. automodule pyro.poutine
+ .. autofunction pyro.poutine.trace
+ .. autofunction pyro.poutine.replay
+ .. autofunction pyro.poutine.lift
+ .. autofunction pyro.poutine.block
+ .. autofunction pyro.poutine.escape
+ .. autofunction pyro.poutine.condition
+ .. autofunction pyro.poutine.infer_config
+ .. autofunction pyro.poutine.scale
+ .. autofunction pyro.poutine.indep
+ .. autofunction pyro.poutine.enum
+ .. autofunction pyro.poutine.do
+ .. autofunction pyro.poutine.queue
```
Let me know offline if this seems wrong or confusing.

PR #1090, opened 2018-04-24
Point to PyTorch official release in CI

neerajprad:
- Use latest release for CI.
 - Wheels are updated (same as `v0.4.0`), in case anyone runs `bash scripts/install_pytorch.sh`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,9 @@ env:
 
 install:
     - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then
-          pip install http://download.pytorch.org/whl/cpu/torch-0.4.0-cp27-cp27mu-linux_x86_64.whl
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -7,8 +7,7 @@ env:
         - PYTHONPATH=$PWD:$PYTHONPATH
 
 install:
-    # Install PyTorch wheel using the install script
-    - bash scripts/install_pytorch.sh
+    - pip install torch
```
jpchen: this will only work for OSX right? otherwise you need the URL of the whl `http://download.pytorch.org/whl/cpu/torch-0.4.0-cp35-cp35m-linux_x86_64.whl `

neerajprad: I think you are right, from what I remember of yesteryears. Will change.

neerajprad:
I hope I don't need to make it any more right.

PR #1092, opened 2018-04-24
Update Poutine docs

eb8680:
Resolves #826 

This PR updates the docs of the `pyro.poutine` module to be more descriptive and to reflect all the changes that have happened.

PR #1094, opened 2018-04-24
bump version to 0.2.0

jpchen:
Blocked by #1092

0.2!


pyro/primitives.py
```diff
@@ -18,7 +18,7 @@
 from pyro.poutine.runtime import _DIM_ALLOCATOR, _MODULE_NAMESPACE_DIVIDER, _PYRO_PARAM_STORE, am_i_wrapped, apply_stack
 from pyro.util import deep_getattr, set_rng_seed  # noqa: F401
 
-version_prefix = '0.2.0-a0'
+version_prefix = '0.2.0'
```
neerajprad: Not part of this PR, but we should remove this from `primitives.py`. I don't think its needed.

fritzo: I believe this should be in `__init__.py`

fritzo: Actually, can we remove it from primitives.py since it's already in `__init__.py`?

PR #1098, opened 2018-04-25
Fix gp/bo tutorials

fehiepsi:
This fixes item list rendering of gp/bo tutorials.


martinjankowiak:
@fehiepsi is it OK if i rework the bo tutorial similar to the way i reworked the gp tutorial?

fehiepsi:
Sure @martinjankowiak, It would be great to me!

Issue #1099, opened 2018-04-25
Error while installing


Jakedismo:
```
(GPUEnv) C:\Users\Joken Surfacebook 2>pip install pyro-ppl
Collecting pyro-ppl
Collecting graphviz>=0.8 (from pyro-ppl)
  Cache entry deserialization failed, entry ignored
  Cache entry deserialization failed, entry ignored
  Cache entry deserialization failed, entry ignored
  Downloading https://files.pythonhosted.org/packages/05/e4/8fcc76823534d47f079c0ff1b3d8b57784e8fba63ceb1ded32c9f4dd993c/graphviz-0.8.2-py2.py3-none-any.whl
Requirement already satisfied: six>=1.10.0 in c:\anaconda\envs\gpuenv\lib\site-packages (from pyro-ppl)
Requirement already satisfied: torch in c:\anaconda\envs\gpuenv\lib\site-packages (from pyro-ppl)
Requirement already satisfied: contextlib2 in c:\anaconda\envs\gpuenv\lib\site-packages (from pyro-ppl)
Exception:
Traceback (most recent call last):
  File "C:\Anaconda\envs\GPUEnv\lib\site-packages\pip\basecommand.py", line 215, in main
    status = self.run(options, args)
  File "C:\Anaconda\envs\GPUEnv\lib\site-packages\pip\commands\install.py", line 335, in run
    wb.build(autobuilding=True)
  File "C:\Anaconda\envs\GPUEnv\lib\site-packages\pip\wheel.py", line 756, in build
    self.requirement_set.prepare_files(self.finder)
  File "C:\Anaconda\envs\GPUEnv\lib\site-packages\pip\req\req_set.py", line 380, in prepare_files
    ignore_dependencies=self.ignore_dependencies))
  File "C:\Anaconda\envs\GPUEnv\lib\site-packages\pip\req\req_set.py", line 666, in _prepare_file
    check_dist_requires_python(dist)
  File "C:\Anaconda\envs\GPUEnv\lib\site-packages\pip\utils\packaging.py", line 48, in check_dist_requires_python
    feed_parser.feed(metadata)
  File "C:\Anaconda\envs\GPUEnv\lib\email\feedparser.py", line 175, in feed
    self._input.push(data)
  File "C:\Anaconda\envs\GPUEnv\lib\email\feedparser.py", line 103, in push
    self._partial.write(data)
TypeError: string argument expected, got 'NoneType'

```
Is this a bug? 
Python 3.6.5
[...]

PR #1103, opened 2018-04-25
Pin PyTorch version in setup.py

fritzo:
Addresses #1099

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,7 +78,7 @@ setup(
         'graphviz>=0.8',
         'networkx>=2.0.0',
         'six>=1.10.0',
-        'torch',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -78,7 +78,7 @@
         'graphviz>=0.8',
         'networkx>=2.0.0',
         'six>=1.10.0',
-        'torch',
+        'torch>=0.4.0',
```
neerajprad: If we also add `numpy` to `install_requires`, then we can just do a `pip install pyro-ppl` out of the box.

Also, this one liner will only work on mac I think, as the binaries for linux/windows aren't hosted on pypi.

PR #1105, opened 2018-04-25
Use internally hosted baseball dataset for example

neerajprad:
I noticed that a recent build failed while fetching the `baseball` dataset. Moving this to our S3 folder `datasets`, which we can also use to host other datasets in tutorials/examples.


neerajprad:
> nice - is this open to the world or do you some creds stored somewhere in the env?

It should be open to all core devs; will ping on slack.

PR #1108, opened 2018-04-26
fix broken rtd

jpchen:
docs were broken from #1107. this fixes that.

TODO(jpchen) a more verbose PR description would help reviewers

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -3,7 +3,7 @@ import sys
 
 import sphinx_rtd_theme
 
-if 'READTHEDOCS' not in os.environ:
+if 'READTHEDOCS' in os.environ:
     # install pyro on rtd so it can be imported
     os.system('pip install pyro-ppl')
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -3,11 +3,6 @@ import sys
 
 import sphinx_rtd_theme
 
-if 'READTHEDOCS' in os.environ:
-    # install pyro on rtd so it can be imported
-    os.system('pip install pyro-ppl')
-
-from pyro import __version__  # noqaE402
 
 # import pkg_resources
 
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -1,9 +1,14 @@
 import os
 import sys
-from pyro import __version__
 
 import sphinx_rtd_theme
 
+if 'READTHEDOCS' not in os.environ:
+    # install pyro on rtd so it can be imported
+    os.system('pip install pyro-ppl')
```
fritzo: This will set `pyro.__version__` to whatever is on `PyPi`, not what is in the local file system. Perhaps you should instead `pip install .` or similar?

fritzo: sorry I still don't understand. Is this fix intended for local development in environments where pyro has not been installed? Or is it for travis to check that docs can be built?

docs/source/conf.py
```diff
@@ -1,9 +1,14 @@
 import os
 import sys
-from pyro import __version__
 
 import sphinx_rtd_theme
 
+if 'READTHEDOCS' not in os.environ:
+    # install pyro on rtd so it can be imported
```
fritzo: Why does this say "on rtd" when the above conditional says "if not on rtd"?

jpchen: shoot thats a bug, good catch. double checked it against the tested branch

docs/source/conf.py
fritzo: Don't query pypi for the Pyro version when building docs on rtd. This would break every time we update the version, since rtd docs are created on git push, whereas pypi version is updated on `python setup.py sdist upload`.

Just hard-code the version.

jpchen: > Just hard-code the version.

as in the snippet above?

and just to be clear, rtd doesnt actually *use* the version, so even if the pypi is out of date, it wouldnt matter

fritzo: If rtd doesn't use the version then set `__version__ = None` or something?

fritzo: Or set `__version__ = 'ignored'`

PR #1111, opened 2018-05-01
Implement coregionalization kernel for contrib.gp

fritzo:
This implements a `Coregionalize` kernel to learn low-rank covariance structure among features.

The typical use case is for modeling correlations among outputs of a multi-output GP, where outputs are coded as distinct data points with on-hot coded features denoting which output each datapoint represents.

## Tested

- added examples to test_kernels.py
- trained an example model with weekly data coded as on-hot vector, and verified that `Coregionalize` performs similarly to `Periodic`.

### DIFF
--- a/pyro/contrib/gp/kernels/coregionalize.py
+++ b/pyro/contrib/gp/kernels/coregionalize.py
@@ -1,0 +1,47 @@
+from __future__ import absolute_import, division, print_function
+
+import torch
+from torch.nn import Parameter
+
+from .kernel import Kernel
+
+
+class Coregionalize(Kernel):
+    r"""
+    A kernel for the linear model of coregionalization :math:`x R R^T z^T`
+    where :math:`R` is an ``input_dim``-by-``rank`` regionalization matrix
+    where typically ``rank < input_dim``.
+
+    :param int input_dim: Number of feature dimensions of inputs.
+    :param str name: Name of the kernel.
+    :param torch.Tensor regions: An optional ``(input_dim, rank)`` shaped
+        matrix that maps features to ``rank``-many regions. If unspecified,
+        this will be randomly initialized.
+    :param int rank: Optional rank. This is only used if ``regions`` is
+        unspecified. If unspecified, ``rank`` defaults to ``input_dim``.
+    :param list active_dims: List of feature dimensions of the input which the
+        kernel acts on.
+    """
+
+    def __init__(self, input_dim, regions=None, rank=None, active_dims=None, name="coregionalize"):
+        super(Coregionalize, self).__init__(input_dim, active_dims, name)
+        if regions is None:
+            if rank is None:
+                rank = input_dim
+            regions = torch.randn(input_dim, rank) / rank ** 0.5
+        if regions.dim() != 2 or regions.shape[0] != input_dim:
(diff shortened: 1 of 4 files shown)


pyro/contrib/gp/kernels/coregionalize.py
```diff
@@ -0,0 +1,57 @@
...
+
+    References:
+
+    [1] Mauricio A. Alvarez, Lorenzo Rosasco, Neil D. Lawrence (2012)
+        Kernels for Vector-Valued Functions: a Review
+
+    :param int input_dim: Number of feature dimensions of inputs.
+    :param torch.Tensor regions: An optional ``(input_dim, rank)`` shaped
+        matrix that maps features to ``rank``-many regions. If unspecified,
+        this will be randomly initialized.
```
fehiepsi: Should we check the positive semidefinite of this kernel? To guarantee it, we can add an additional diagonal term (the form is similar to LowRankMultivariateNormal).

pyro/contrib/gp/kernels/coregionalize.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+
+    def __init__(self, input_dim, regions=None, rank=None, active_dims=None, name="coregionalize"):
+        super(Coregionalize, self).__init__(input_dim, active_dims, name)
+        if regions is None:
+            if rank is None:
+                rank = input_dim
+            regions = torch.randn(input_dim, rank) / rank ** 0.5
+        if regions.dim() != 2 or regions.shape[0] != input_dim:
+            raise ValueError("Expected region.shape == ({},r), actual {}".format(regions.shape))
```
fehiepsi: Nit: missing an argument `input_dim` here for the format.

pyro/contrib/gp/kernels/coregionalize.py
fritzo: Sure, I can add a vector of independent variances to make this numerically stable (note that mathematically `R R^t` will always be positive semidefinite). In an effort to avoid greek letters, could rename `regions` to `components` (in analogy with principle components analysis), and denote this vector of independent noises as `bias` (in analogy with the `Polynomial` kernel)?

fritzo: Actually I don't understand the motivation for the `kappa` part of the kernel and it seems to complicate the implementation. Do you understand the motivation, or why we should bake a vector of variances into this implementation rather than separate that out as a distinct `Kernel` class?

fehiepsi: Yeah, agree! The `components` name seems better to me. I also think that `kappa` is unnecessary and only used to guarantee the positiveness of coregionalization matrix (in case of R has rank less than `r`). I guess ignoring it will not be an issue (hopefully) because we already have `jitter` term in GP models to deal with the singularity of covariance matrix.

fritzo: @fehiepsi I've noticed that this kernel is sensitive to initialization. I've implemented the `kappa` thing locally to experiment. If it appears to make things more robust (more convex), I'll send a follow-up PR. Thanks for suggesting this!

fritzo: Here's the follow-up PR: https://github.com/uber/pyro/pull/1116

PR #1112, opened 2018-05-02
Fix a tiny bug when rendering gp tutorial

fehiepsi:
This fixes a rendering bug in Data section of gp tutorial: http://pyro.ai/examples/gp.html


fehiepsi:
Oppps, I will not make a pull request on the top of an old one in the future.

PR #1114, opened 2018-05-02
Update to NUTS/HMC docstrings

neerajprad:
One of the users on our Slack distribution channel mentioned a small issue with the example in the NUTS docstring, so taking the opportunity to clean up the docstrings for HMC and NUTS.

 - Added the same example to HMC, and ensured that both are working. Updated to use EmpiricalMarginal.
 - Also pointing to the `baseball.py` example which is important as it is not something that users would discover easily otherwise.

### DIFF
--- a/docs/source/mcmc.rst
+++ b/docs/source/mcmc.rst
@@ -1,9 +1,4 @@
 MCMC
 ====
 
-.. automodule:: pyro.infer.mcmc
-    :members:
-    :undoc-members:
-    :show-inheritance:
-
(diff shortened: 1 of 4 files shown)


eb8680:
Maybe we should aim to start using [doctest](https://docs.python.org/3.6/library/doctest.html) to make sure our docstring examples are up-to-date?

neerajprad:
> Maybe we should aim to start using doctest to make sure our docstring examples are up-to-date?

Yeah, I think that's a good idea. Will create a task for that.

PR #1116, opened 2018-05-02
Add diagonal component to coregionalization kernel

fritzo:
## Motivation

The idea is to change our `Coregionalization` kernel to allow "low-rank + diagonal" matrices, which anecdotally are easier to optimize than pure low-rank matrices.

## Design choices

The most common use case is to provide only the `rank` arg to `Coregionalization()`. In this case I've chosen to randomly initialize the `components` matrix and set `diagonal` to a constant, and scale both so that the expected value of the kernel is the identity matrix.

If `rank == num_inputs` I avoid the diagonal matrix entirely. This is sensible since `C C^T` is no longer sparse, and adding the `diagonal` component is redundant.

### DIFF
--- a/pyro/contrib/gp/kernels/coregionalize.py
+++ b/pyro/contrib/gp/kernels/coregionalize.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import torch
+from torch.distributions import constraints
 from torch.nn import Parameter
 
 from .kernel import Kernel
@@ -9,14 +10,19 @@ from .kernel import Kernel
 class Coregionalize(Kernel):
     r"""
     A kernel for the linear model of coregionalization
-    :math:`k(x,z) = x^T R R^T z` where :math:`R` is an
-    ``input_dim``-by-``rank`` matrix and typically ``rank < input_dim``.
+    :math:`k(x,z) = x^T (C C^T + D) z` where :math:`C` is an
+    ``input_dim``-by-``rank`` matrix and typically ``rank < input_dim``,
+    and ``D`` is a diagonal matrix.
 
     This generalizes the
     :class:`~pyro.contrib.gp.kernels.dot_product.Linear` kernel to multiple
-    features. The typical use case is for modeling correlations among outputs
-    of a multi-output GP, where outputs are coded as distinct data points with
-    one-hot coded features denoting which output each datapoint represents.
+    features with a low-rank-plus-diagonal weight matrix. The typical use case
+    is for modeling correlations among outputs of a multi-output GP, where
+    outputs are coded as distinct data points with one-hot coded features
+    denoting which output each datapoint represents.
+
+    If only ``rank`` is specified, the kernel ``(C C^T + D)`` will be
+    randomly initialized to a matrix with expected value the identity matrix.
 
     References:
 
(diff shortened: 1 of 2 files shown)


pyro/contrib/gp/kernels/coregionalize.py
```diff
@@ -9,50 +10,86 @@
...
+
+        # Add a low-rank kernel with expected value torch.eye(input_dim, input_dim) * rank / input_dim.
         if components is None:
-            if rank is None:
-                rank = input_dim
-            components = torch.randn(input_dim, rank) / rank ** 0.5
-        if components.dim() != 2 or components.shape[0] != input_dim:
-            raise ValueError("Expected region.shape == ({},rank), actual {}".format(input_dim, components.shape))
+            rank = input_dim if rank is None else rank
+            components = torch.randn(input_dim, rank) / input_dim ** 0.5
```
fehiepsi: I think that the expectation of coregionalization matrix B is Identity. The `components` will transforms (one-hot version of) the set {1, ..., D} to the row vectors {w1, ..., wD). Hence, diag(B) = [w1.w1, w2.w2, ..., wD.wD]. So we expect E(wi.wi) = 1. This is achieved if we initialize `components = randn(...) / rank ** 0.5`.

fritzo: I'd like the expectation of the entire kernel `C C^T + D` to be the identity. That's why I've split the identity in to the `rank / input_dim` + `(1 - rank / input_dim)` parts corresponding to `E[ C C^T ]` and `D`, respectively.

fritzo:
>  in case input_dim < rank, W.Wt is not a semi-positive matrix

Hmm as I understand,
- If `input_dim <= rank` then almost surely `W.Wt` is positive definite (but we don't really care about the overparametrized case `input_dim < rank`).
- If `0 < rank < input_dim` then almost surely `W.Wt` is positive semidefinite aka nonnegative definite.

Are you concerned with algebra or with finite-precision numerics?

pyro/contrib/gp/kernels/coregionalize.py
fehiepsi: Got it! If you want to keep the diagonal part, then how about making them `rank / (input_dim + rank)`, `(1 - rank / input_dim + rank)`? When using diagonal, then this diagonal term will give the independent info for outputs, hence it should have a bigger weight. This way, we will have `diagonal` for case `input_dim <= rank`, so the implementation will be simplified.

fehiepsi:
@fritzo My bad! I have overlooked the word "semi". So assume B is diagonal, `input_dim = rank`, we will allow entries `0` in its diagonal, and rely on `jitter` term of models for stability?

fritzo:
> we will allow entries 0 in [B's] diagonal

Correct, the optimizer could send some of the diagonal entries of `C C^T` to zero in which case we'd rely on the model's jitter for numerical stability.

pyro/contrib/gp/kernels/coregionalize.py
fritzo: > so the implementation will be simplified

Hmm I'm not seeing how the implementation would simplify the code. IIUC your suggestion would involve replacing the `torch.randn(input_dim, rank) / rank` with `torch.randn(input_dim, rank) / (input_dim + rank)` and replacing `1 - rank / input_dim` with `1 - rank / (input_dim + rank)`. Could you explain?

fritzo:
cc @yajiayang This seems to make our experiments more stable.

pyro/contrib/gp/kernels/coregionalize.py
fehiepsi: I think that in that case, we can remove the situation `diagonal = None` in the implementation because `diagon > 0` regardless rank >= input_dim.

fehiepsi:
@fritzo It seems that coregionalization matrix is closely related to the covariance matrix of outputs in [factor analysis](https://en.wikipedia.org/wiki/Factor_analysis) where `components` is `loading matrix` (`factor loadings` in Bishop book) and `diagonal` is `unobserved stochastic error terms`. It differs a little bit to Probabilistic PCA that in PPCA we require entries in `diagonal` equal. In [scikit-learn](http://scikit-learn.org/stable/modules/generated/sklearn.decomposition.FactorAnalysis.html#sklearn.decomposition.FactorAnalysis)'s implementation, they call them `components` and `noise_variance` (estimated noise variance for each feature, is this a better name?). These references explain the meaning of C (it seems to me that W is a better notion here, as we always use `eps`-not n-for noise) and D terms in the kernel formula.

So I propose to keep the old scale for `components` and just initialize `diagonal` a value-1 vector without scaling (if not specified). Or we can reduce them by a factor `0.5` as in [GPy](http://gpy.readthedocs.io/en/deploy/_modules/GPy/kern/src/coregionalize.html)'s implementation. The drawback of current scaling is when input_dim >> rank (which is typical when we have a large number of outputs to predict), the weight we give to W is so small, hence the error term will dominate the correlation info.

fritzo:
Ok, I've renamed `C` to `W` and set the weights to 0.5 + 0.5, so that `E[W.Wt] = I/2 = D`, which is the initialization used by GPy.

pyro/contrib/gp/kernels/coregionalize.py
```diff
@@ -9,50 +10,86 @@
...
+        else:
+            rank = components.shape[-1]
+        if components.shape != (input_dim, rank):
+            raise ValueError("Expected components.shape == ({},rank), actual {}".format(input_dim, components.shape))
         self.components = Parameter(components)
 
+        # Add a diagonal component only if rank < input_dim.
+        # The expected value should be torch.eye(input_dim, input_dim) * (1 - rank / input_dim),
+        # such that the result has expected value the identity matrix.
+        if diagonal is None and rank < input_dim:
```
fehiepsi: I think that we can safely remove this condition `rank < input_dim` now?

fritzo:
Thanks for reviewing!

fritzo:
ready to merge as soon as tests pass

PR #1117, opened 2018-05-03
Relax check for gp kernel active_dims to allow overlap

fritzo:
This removes the check that, when combining kernels `k1` and `k2`, their `.active_dims` are either identical or disjoint. After this PR, arbitrary overlapping is allowed.

I'm unsure of the original intent of this check. The check appears to be no longer needed since all `._slice_input()` calls are made at the leaf kernels of `Combination` kernels and hence leaf kernels can safely grab their components.

Why do I want overlapping kernels? One example is in a dataset with three groups of features, say X0, X1, X2. I want to use X0 to modulate X1 and X2 via
```py
modulated1 = Coregionalize(active_dims=X0, ...).mul(RBF(active_dims=X1, ...))
modulated2 = Coregionalize(active_dims=X0, ...).mul(RBF(active_dims=X2, ...))
kernel = modulated1.add(modulated2)  # <---- error here
```
Of course I could work around this by combining with a `Constant`ly zero kernel
```py
workaround1 = Constant(input_dim=len(X2), active_dims=X2, variance=torch.zeros(1),
                       name="workaround1")
workaround1.fix_param("variance")
modulated1 = modulated1.add(workaround1)

workaround2 = Constant(input_dim=len(X1), active_dims=X1, variance=torch.zeros(1),
                       name="workaround2")
workaround2.fix_param("variance")
modulated2 = modulated2.add(workaround2)
```
but that feels very kludgey.

@fehiepsi Does this seem reasonable?

### DIFF
--- a/pyro/contrib/gp/kernels/kernel.py
+++ b/pyro/contrib/gp/kernels/kernel.py
@@ -175,16 +175,7 @@ class Combination(Kernel):
 
         active_dims = set(kern0.active_dims)
         if isinstance(kern1, Kernel):
-            active_dims1 = set(kern1.active_dims)
-            if active_dims == active_dims1:  # on the same active_dims
-                pass
-            elif len(active_dims & active_dims1) == 0:  # on disjoint active_dims
-                active_dims = active_dims | active_dims1
-            else:
-                raise ValueError("Sub-kernels must act on the same active dimensions "
-                                 "or disjoint active dimensions (to create direct sum "
-                                 "or tensor product of kernels).")
-
+            active_dims  |= set(kern1.active_dims)
         active_dims = sorted(active_dims)
         input_dim = len(active_dims)
         super(Combination, self).__init__(input_dim, active_dims, name)
--- a/pyro/contrib/gp/kernels/kernel.py
+++ b/pyro/contrib/gp/kernels/kernel.py
@@ -175,7 +175,7 @@ class Combination(Kernel):
 
         active_dims = set(kern0.active_dims)
         if isinstance(kern1, Kernel):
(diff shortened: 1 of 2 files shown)


fritzo:
@fehiepsi Here's a proof attempt :smile:

## Lemma

Let `k` be a positive semidefinite kernel on dims `d1`. Let
```py
k2 = k.add(Constant(variance=1, active_dims=d2))
```
be the kernel `k` lifted to nuisance features `d2`. Then `k2` is positive semidefinite.

### Proof

Apply tensor product lemma, since `Constant(...)` is also positive semidefinite. []

## Theorem

Let `k1` be a positive definite kernel on dims `d1 u d2` and `k2` be a positive  kernel on dims `d2 u d3`. Then the product kernel
```
k12(x, y) = k1(x[d1+d2], y[d1,d2]) * k2(x[d2,d3], y[d2,d3])
```
is also positive semidefinite.

### Proof

By the lemma the following lifted kernels are also positive definite.
```py
k1c(x,y) = k1(x[d1,d2], y[d1,d2]) * Constant(1)(x[d3], y[d3])
k2c(x,y) = k2(x[d1,d2], y[d1,d2]) * Constant(1)(x[d1], y[d1])
```
Then `k12 = k1 * k2 = k1c * k2c` which is positive definite from the identical-spaces lemma. []

fehiepsi:
Thank you a lot @fritzo !

fritzo:
@fehiepsi can you hit the merge button? We have a policy "authors do not merge their own PRs". If permissions disallow you I'll ask Martin.

jpchen:
@fritzo only repo owners can merge PRs. ping one of us if you need a PR merged

PR #1120, opened 2018-05-03
Implement TorchDistribution.expand()

### LINKED ISSUE #1119: Implement TorchDistribution.expand()
## Why?

Pyro currently provides an `.expand_by()` method to aid/burden users when reshaping distributions. However `.expand_by()` is limited to expanding only by adding new dimensions on the left. It appears that to automate some of this expansion (#238, #1115), we'll need to additionally support broadcasting at internal dimension, as is done by `torch.Tensor.expand()`.

Here is an example demonstrating the necessity of `.expand()` (regardless of dim ordering):
```py
X, Y = 320, 200
x_axis = pyro.iarange("x_axis", X, dim=-2)
y_axis = pyro.iarange("y_axis", Y, dim=-1)
with x_axis:
    x = pyro.sample("x", dist.Normal(0, 1).expand_by([X, 1])      # .expand_by() suffices
with y_axis:
    y = pyro.sample("y", dist.Normal(0, 1).expand_by([Y])         # .expand_by() suffices
with x_axis, y_axis:
    yx = pyro.sample("yx", dist.Normal(y_noise, 1).expand_by([X]) # .expand_by() suffices
    xy = pyro.sample("xy", dist.Normal(x_noise, 1).expand([X, Y]) # .expand() is needed
    ...
```

## How?

First `.expand()` will operate only on `batch_shape`. Note that `.expand_by()` is agnostic about which shape is being expanded (`batch_shape` of the `.log_prob()` results -versus- `batch_shape+event_shape` of `.sample()` results). In contrast, `.expand()` will need to decide which of the two shapes is being expanded. Here are three reasons why `batch_shape` is more sensible:
1. It is unclear how to expand `event_shape` in general (e.g. how do you expand a normalized set of probabilities?)
2. `.expand()` can only change dimension of size 1 to dimensions of size > 1. Yet none of our event shapes have meaningful dimensions of size 1. (e.g. a `OneHotCategorical` of size 1 is degenerate; `MultivariateNormal` of size 1 is just a `Normal`).
3. The current use cases only need to expand `batch_shape`.

Mechanically, `.expand()` can attempt to apply `.expand_by()` under the hood. As a last resort, it can combine `.expand_by()` with a `PermutedDistribution` that continues to use `.sample(sample_shape)` to add batching on the left, but then permutes that batching into internal dims. Since `PermutedDistribution` may be expensive and lead to non-contiguous tensors, we should implement overrides in wrappers for the most commonly used distributions, e.g.
```py
# in pyro/distributions/torch.py

class Normal(torch.distributions.Normal, TorchDistributionMixin):
    def expand(self, batch_shape):
        loc = self.loc.expand(batch_shape)
        scale = self.scale.expand(batch_shape)
        return Normal(loc, scale)

    def expand(self, batch_shape):
        loc = self.loc.expand(batch_shape + self.event_shape)
        # we'll want to check for covariance_matrix and precision_matrix, but roughly:
        scale_tril = self.scale_tril.expand(batch_shape + 2 * self.event_shape)
        return MultivariateNormal(loc, scale_tril=scale_tril)
```

fritzo:
Resolves #1119 

This PR combines a partial implementation of `TorchDistribution.expand()` with a number of custom implementations for most distribution classes. A general implementation would be tricky to implement and slow in practice, since permuting dimensions would make the resulting tensors non-contiguous.

## Tested

- added a number of new tests of `.expand()` and `.expand_by()`
- fixed some existing distributions (e.g. `ReshapedDistribution`) to pass new tests
- ignored errors due to https://github.com/pytorch/pytorch/pull/7269

### DIFF
--- a/pyro/distributions/torch_distribution.py
+++ b/pyro/distributions/torch_distribution.py
@@ -271,15 +271,17 @@ class ReshapedDistribution(TorchDistribution):
         return self.base_dist.rsample(sample_shape + self.sample_shape)
 
     def log_prob(self, value):
-        return sum_rightmost(self.base_dist.log_prob(value), self.reinterpreted_batch_ndims).expand(self.batch_shape)
+        shape = broadcast_shape(self.batch_shape, value.shape[:value.dim() - self.event_dim])
+        return sum_rightmost(self.base_dist.log_prob(value), self.reinterpreted_batch_ndims).expand(shape)
 
     def score_parts(self, value):
+        shape = broadcast_shape(self.batch_shape, value.shape[:value.dim() - self.event_dim])
         log_prob, score_function, entropy_term = self.base_dist.score_parts(value)
-        log_prob = sum_rightmost(log_prob, self.reinterpreted_batch_ndims).expand(self.batch_shape)
+        log_prob = sum_rightmost(log_prob, self.reinterpreted_batch_ndims).expand(shape)
         if not isinstance(score_function, numbers.Number):
-            score_function = sum_rightmost(score_function, self.reinterpreted_batch_ndims).expand(self.batch_shape)
+            score_function = sum_rightmost(score_function, self.reinterpreted_batch_ndims).expand(shape)
         if not isinstance(entropy_term, numbers.Number):
-            entropy_term = sum_rightmost(entropy_term, self.reinterpreted_batch_ndims).expand(self.batch_shape)
+            entropy_term = sum_rightmost(entropy_term, self.reinterpreted_batch_ndims).expand(shape)
         return ScoreParts(log_prob, score_function, entropy_term)
 
     def enumerate_support(self):
--- a/pyro/distributions/torch_distribution.py
+++ b/pyro/distributions/torch_distribution.py
@@ -69,7 +69,16 @@ class TorchDistributionMixin(Distribution):
         Expands a distribution to a desired
         :attr:`~torch.distributions.distribution.Distribution.batch_shape`.
 
-        :param torch.Size batch_shape: The target ``batch_shape``.
+        Note that this is more general than :meth:`expand_by` because
+        ``d.expand_by(sample_shape)`` can be reduced to
+        ``d.expand(sample_shape + d.batch_shape)``.
+
+        :param torch.Size batch_shape: The target ``batch_shape``. This must
+            compatible with ``self.batch_shape`` similar to the requirements
+            of :func:`torch.Tensor.expand`: the target ``batch_shape`` must
+            be at least as long as ``self.batch_shape``, and for each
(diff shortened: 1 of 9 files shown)


fritzo:
@alicanb Do you think the `.expand()` method is worth pushing upstream to PyTorch?

eb8680:
> Do you think the .expand() method is worth pushing upstream to PyTorch?

+1 to this in the interest of not having Pyro's broadcasting usage/semantics depart from that of PyTorch distributions unnecessarily

tests/distributions/test_distributions.py
```diff
@@ -96,3 +96,62 @@ def test_distribution_validate_args(dist_class, args, validate_args):
...
+    for idx in range(dist.get_num_test_data()):
+        small = dist.pyro_dist(**dist.get_dist_params(idx))
+        large = small.expand(sample_shape + small.batch_shape)
+        assert large.batch_shape == sample_shape + small.batch_shape
+        check_sample_shapes(small, large)
+
+
+@pytest.mark.parametrize('sample_shape', [(), (2,), (2, 3)])
+def test_expand_new_dim(dist, sample_shape):
+    for idx in range(dist.get_num_test_data()):
```
neerajprad: This looks to be the same as the test above?

pyro/distributions/torch_distribution.py
```diff
@@ -62,6 +64,32 @@ def shape(self, sample_shape=torch.Size()):
         """
         return sample_shape + self.batch_shape + self.event_shape
 
+    def expand(self, batch_shape):
+        """
+        Expands a distribution to a desired
```
neerajprad: Could you elaborate on the description here - what are the constraints on `batch_shape` provided to `.expand` and how is this different from `.expand_by`?

neerajprad:
> +1 to this in the interest of not having Pyro's broadcasting usage/semantics depart from that of PyTorch distributions unnecessarily

While I would like it to be upstream, it would be tricky to justify as a general enough use case, given that `torch.distributions` already distinguishes between `sample_shape` and `batch_shape`, whereas `expand` just makes the `sample` broadcasting semantics lax by allowing for sample_shape to modify dimensions to the right of batch shape. e.g. allowing not just drawing samples of shape `(4, 3, 1)` from a distribution that has batch shape `(3, 1)`, but also samples of shape `(3, 5)`. Maybe we could relax the sample shape restrictions.

neerajprad:
> A general implementation would be tricky to implement and slow in practice.

By controlling our iarange dims explicitly, I think we may not even need this.

alicanb:
>Do you think the .expand() method is worth pushing upstream to PyTorch?

I'm not sure it's use case is general enough to justify writing separate implementations for most distributions. Let me ask a few people from our side if they can see themselves using this in the future.

PR #1122, opened 2018-05-04
Expose contrib.gp classes in sphinx toctree

fritzo:
I think pyro.contrib.gp would be easier to navigate if all kernels and likelihoods were discoverable in the sphinx sidebar toc.

This PR:
1. rearranges our sphinx toc so that pyro.contrib modules can show deeper nesting
2. autogenerates documentation for `gp.kernels` and `gp.likelihoods` so that all kernels and likelihoods show up in the toc.

I've used alphabetic ordering to make it easier to find documentation if you need to. I think it would be overly complex to retain by-source ordering because that would involve an additional layer of nesting (e.g. gp -> kernels -> dot_product -> linear), and I haven't been able to get this deep nesting to show up in the sphinx toc. I'm open to suggestions if anyone knows how to fix this.

Here's what the generated docs look like:
http://fritzo.org/temp/pyro/html/contrib.gp.html#periodic


pyro/contrib/gp/kernels/__init__.py
```diff
@@ -34,3 +34,19 @@
...
     "WhiteNoise",
 ]
+
+# Create sphinx documentation.
+__doc__ = '\n\n'.join([
+
+    '''
+    {0}
+    ----------------------------------------------------------------
+    .. autoclass:: pyro.contrib.gp.kernels.{0}
```
jpchen: 👍

docs/source/index.rst
```diff
@@ -22,7 +22,15 @@ Pyro Documentation
    nn
    optimization
    poutine
-   contrib
+
+.. toctree::
+   :glob:
+   :maxdepth: 2
```
jpchen: > I haven't been able to get this deep nesting to show up in the sphinx toc

does bumping `maxdepth` help

fritzo: no, that was the first thing i tried

fritzo:
ready to merge as soon as tests pass

PR #1124, opened 2018-05-04
adding a straight through estimator for relaxed categorical

karalets:
(builds on code from @karalets and @fritzo)

    An implementation of RelaxedOneHotCategorical with a straight-through gradient estimator.

    This distribution has the following properties:
    -- the samples returned by the `rsample` method are discrete/quantized
    -- the `log_prob` method returns the log probability of the relaxed/unquantized sample
       using the GumbelSoftmax distribution
    -- in the backward pass the gradient of the sample with respect to the parameters of the
       distribution uses the relaxed/unquantized sample

now has more legit tests. bernoulli can be done in another pr.


pyro/distributions/relaxed_straight_through.py
```diff
@@ -0,0 +1,61 @@
...
+    In detail, it uses discrete samples for the forward step but the relaxed sample for the backwards gradient.
+    """
+    def __init__(self, temperature, probs=None, logits=None, validate_args=None):
+        super(RelaxedCategoricalStraightThrough, self).__init__(temperature=temperature, probs=probs, logits=logits, validate_args=validate_args)
+        self._unquantize = {}
+
+    def rsample(self, sample_shape=torch.Size()):
+        soft_sample = super(RelaxedCategoricalStraightThrough, self).rsample(sample_shape)
+        hard_sample = QuantizeCategorical.apply(soft_sample)
+        self._unquantize[id(hard_sample)] = soft_sample
```
martinjankowiak: why not use the tensor itself as key?

karalets: ok, the hard_sample is a onehot vector mapping of a dense vector.
However, multiple dense vectors can map to the same onehot vector.

As such, in order to make the mapping unique, we decided to use the id instead of the value.

Makes sense?

pyro/distributions/relaxed_straight_through.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    def __init__(self, temperature, probs=None, logits=None, validate_args=None):
+        super(RelaxedOneHotCategoricalStraightThrough, self).__init__(temperature=temperature, probs=probs,
+                                                                      logits=logits, validate_args=validate_args)
+        self._unquantize = {}
+
+    def rsample(self, sample_shape=torch.Size()):
+        soft_sample = super(RelaxedOneHotCategoricalStraightThrough, self).rsample(sample_shape)
+        hard_sample = QuantizeCategorical.apply(soft_sample)
+        self._unquantize[id(hard_sample)] = soft_sample
```
martinjankowiak: would it be better to use `hard_sample.data_ptr()` instead of `id()`

pyro/distributions/relaxed_straight_through.py
```diff
@@ -0,0 +1,56 @@
...
+        argmax = input.max(-1)[1]
+        result = input.new_zeros(input.shape)
+        # result[argmax] = 1
+        if argmax.dim() < result.dim():
+            argmax = argmax.unsqueeze(-1)
+        return result.scatter_(-1, argmax, 1)
+
+    @staticmethod
+    def backward(ctx, grad):
+        return grad
```
martinjankowiak: there was a `grad.clone()` here. i got rid of it. shouldn't be needed?

tests/distributions/test_relaxed_straight_through.py
```diff
@@ -0,0 +1,71 @@
...
+        temperature = torch.tensor(0.5)
+        probs = torch.tensor(probs, requires_grad=True)
+        d = RelaxedOneHotCategoricalStraightThrough(temperature, probs=probs)
+        sample = d.rsample()
+        log_prob = d.log_prob(sample)
+        grad_probs = grad(log_prob.sum(), [probs])[0]
+        assert grad_probs.shape == probs.shape
+
+
+@pytest.mark.parametrize('temp', [0.2, 0.5, 1.0])
```
martinjankowiak: this test basically just tests plumbing

tests/distributions/test_relaxed_straight_through.py
```diff
@@ -0,0 +1,71 @@
...
+    dist_q = RelaxedOneHotCategoricalStraightThrough(temperature=temp, probs=q)
+    z = dist_q.rsample(sample_shape=(num_samples,))
+    actual = grad(dist_q.log_prob(z).sum(), [q])[0] / num_samples
+
+    assert_equal(expected, actual, prec=0.1,
+                 msg='bad grad for RelaxedOneHotCategoricalStraightThrough (expected {}, got {})'.
+                 format(expected, actual))
+
+
+def test_svi_usage():
```
martinjankowiak: this test actually shows how the (biased) gradients can be used in svi

martinjankowiak:
cc @karalets take a look should be ready to go

pyro/distributions/relaxed_straight_through.py
fritzo: I'd omit the `id()` entirely since `Tensor` already hashes on id. `.data_ptr()` is too clever imo.

pyro/distributions/relaxed_straight_through.py
fritzo: sounds reasonable, i do not understand the purpose of the `.clone()`

pyro/distributions/__init__.py
```diff
@@ -33,6 +34,7 @@
     "Rejector",
     "TorchDistribution",
     "VonMises",
+    "RelaxedOneHotCategoricalStraightThrough"
```
fritzo: nit: make these alphabetical and add a trailing comma.

pyro/distributions/relaxed_straight_through.py
```diff
@@ -0,0 +1,56 @@
...
+
+import torch
+
+from pyro.distributions.torch import RelaxedOneHotCategorical
+from pyro.distributions.util import copy_docs_from
+
+
+@copy_docs_from(RelaxedOneHotCategorical)
+class RelaxedOneHotCategoricalStraightThrough(RelaxedOneHotCategorical):
+    """
```
martinjankowiak: hope these docs come out looking the way they ought?

pyro/distributions/relaxed_straight_through.py
fritzo: Actually this is just plain buggy. I believe a correct implementation could instead use a [WeakKeyDictionary](https://docs.python.org/2/library/weakref.html#weakref.WeakKeyDictionary).

tests/distributions/test_relaxed_straight_through.py
```diff
@@ -0,0 +1,71 @@
...
+        d = RelaxedOneHotCategoricalStraightThrough(temperature, probs=probs)
+        sample = d.rsample()
+        log_prob = d.log_prob(sample)
+        grad_probs = grad(log_prob.sum(), [probs])[0]
+        assert grad_probs.shape == probs.shape
+
+
+@pytest.mark.parametrize('temp', [0.2, 0.5, 1.0])
+def test_entropy_grad(temp):
+    num_samples = 1000000
```
fritzo: Why does 1e6 samples only lead to `prec=0.1`? Is this due to finite temperature effects?

martinjankowiak: it's partially that. it's also that the scale isn't 1.0. some of the elements of `expected` are of size 7.6 or what not

pyro/distributions/relaxed_straight_through.py
martinjankowiak: ok well it would be great if, once everything else in this PR is in order, you could push the needed changes. because i don't know what you have in mind

pyro/distributions/relaxed_straight_through.py
jpchen: if only there existed something like `make docs` for you to see them...

pyro/distributions/relaxed_straight_through.py
fritzo: done.

martinjankowiak: great thanks!

pyro/distributions/__init__.py
martinjankowiak: done

karalets:
lgtm after my questions in person, please merge

Issue #1115, opened 2018-05-02
Adjusting the iarange dim allocation scheme


neerajprad:
Currently, our `iarange` dimension allocator assigns dimensions to `iarange` starting from the right - i.e. the outermost `iarange` starts at -1, the next one at -2 and so on. Consider the following code:

```python
with pyro.iarange("outer", 5):
    outer_sample = pyro.sample("outer_sample", dist.Categorical(torch.ones(5, 4) / 4))
    assert outer_sample.shape == (5,)
    with pyro.iarange("inner", 3):
        inner_sample = pyro.sample("inner_sample", dist.Categorical(torch.ones(3, 5, 4) / 4))
        # Note how the shape is inverted from the order in which the `iarange`s appear.
        assert inner_sample.shape == (3, 5)  
```

After discussing with @fritzo, @eb8680 and @martinjankowiak, I would like to propose the following change to the `iarange` dimension allocation: when the user specifies `max_iarange_nesting`, we allocate dims starting from `-max_iarange_nesting` and moving to the right, i.e. iarange dims follow their order of appearance in the model and guide. Notice how the code sample above would look like under this scheme:

```python
# set_max_iarange_nesting = 2
with pyro.iarange("outer", 5):
    outer_sample = pyro.sample("outer_sample", 
# Note the additional 1 in the batch_shape below, which corresponds to the additional 
# batch dim specified by `max_iarange_nesting`. 
# With implicit broadcasting, this won't be neeeded.
    dist.Categorical(torch.ones(5, 1, 4) / 4))  
    assert outer_sample.shape == (5, 1)
    with pyro.iarange("inner", 3):
        inner_sample = pyro.sample("inner_sample", dist.Categorical(torch.ones(5, 3, 4) / 4))
        assert inner_sample.shape == (5, 3)
```

What we would finally like to have with implicit broadcasting (no need to adjust the batch sizes manually):
```python
# set_max_iarange_nesting = 2
with pyro.iarange("outer", 5):
    outer_sample = pyro.sample("outer_sample", dist.Categorical(torch.ones(4) / 4))  
    assert outer_sample.shape == (5, 1)
    with pyro.iarange("inner", 3):
        inner_sample = pyro.sample("inner_sample", dist.Categorical(torch.ones(4) / 4))
        assert inner_sample.shape == (5, 3)
```

### What does this give us?
[...]

PR #1125, opened 2018-05-04
Implicitly broadcast sample sites using iarange dim and size information

neerajprad:
This makes a small change to use the `expand` logic from #1119 to implement implicit broadcasting via a `broadcast` poutine.

This should be safe to merge as this logic will not be exercised until the user wraps their model/guides inside `poutine.broadcast`. We could later decide to reuse this to do the broadcasting behind the scenes for parallelizing over `num_particles`, for instance. Or use our learnings to build more composable broadcasting effects as @eb8680 mentioned in #1115. 

e.g. Using @fritzo's example from #1119
```py
X, Y = 320, 200
x_axis = pyro.iarange("x_axis", X, dim=-2)
y_axis = pyro.iarange("y_axis", Y, dim=-1)
with x_axis:
    x_noise = pyro.sample("x", dist.Normal(0, 1).expand_by([X, 1])) # .expand_by() suffices
with y_axis:
    y_noise = pyro.sample("y", dist.Normal(0, 1).expand_by([Y]))    # .expand_by() suffices
with x_axis, y_axis:
    yx = pyro.sample("yx", dist.Normal(y_noise, 1).expand_by([X])) # .expand_by() suffices
    xy = pyro.sample("xy", dist.Normal(x_noise, 1).expand([X, Y])) # .expand() is needed
    ...
```

we do not need to `expand` or `expand_by` if the model is wrapped inside `poutine.broadcast`:

```py
x_axis = pyro.iarange("x_axis", X, dim=-2)
y_axis = pyro.iarange("y_axis", Y, dim=-1)
with x_axis:
    x_noise = pyro.sample("x", dist.Normal(0, 1))
    assert x_noise.shape == torch.Size((320, 1))
with y_axis:
    y_noise = pyro.sample("y", dist.Normal(0, 1))
    assert y_noise.shape == torch.Size((200,))
with x_axis, y_axis:
    yx = pyro.sample("yx", dist.Normal(y_noise, 1))
    assert yx.shape == torch.Size((320, 200))
    xy = pyro.sample("xy", dist.Normal(x_noise, 1))
    assert xy.shape == torch.Size((320, 200))
```

 - Note that this requires the user to line up the different `iarange`s correctly via the `dim` arg.
 - **Tests**: Added a couple of tests to `test_valid_models`. Also modified one of the gradient tests to use the broadcasting logic.

### DIFF
--- a/pyro/poutine/broadcast_messenger.py
+++ b/pyro/poutine/broadcast_messenger.py
@@ -21,20 +21,19 @@ class BroadcastMessenger(Messenger):
         dist = msg["fn"]
         actual_batch_shape = getattr(dist, "batch_shape", None)
         if actual_batch_shape is not None:
-            target_batch_shape = []
+            target_batch_shape = [None if size == 1 else size for size in actual_batch_shape]
             for f in msg["cond_indep_stack"]:
-                if f.dim is not None:
-                    assert f.dim < 0
-                    target_batch_shape = [None] * (-f.dim - len(target_batch_shape)) + target_batch_shape
-                    if target_batch_shape[f.dim] is not None:
-                        raise ValueError('\n  '.join([
-                            'at site "{}" within iarange("", dim={}), dim collision'
-                            .format(msg["name"], f.name, f.dim),
-                            'Try setting dim arg in other iaranges.']))
-                    target_batch_shape[f.dim] = f.size
+                if f.dim is None:
+                    continue
+                assert f.dim < 0
+                target_batch_shape = [None] * (-f.dim - len(target_batch_shape)) + target_batch_shape
+                if target_batch_shape[f.dim] not in (None, f.size):
+                    raise ValueError("Shape mismatch inside iarange('{}') at site {} dim {}, {} vs {}".format(
+                        f.name, msg['name'], f.dim, f.size, target_batch_shape[f.dim]))
+                target_batch_shape[f.dim] = f.size
             # Starting from the right, if expected size is None at an index,
             # set it to the actual size if it exists, else 1.
             for i in range(-len(target_batch_shape) + 1, 1):
                 if target_batch_shape[i] is None:
-                    target_batch_shape[i] = actual_batch_shape[i] if len(actual_batch_shape) > -i else 1
+                    target_batch_shape[i] = actual_batch_shape[i] if len(actual_batch_shape) >= -i else 1
             msg["fn"] = msg["fn"].expand(target_batch_shape)
--- a/pyro/poutine/broadcast_messenger.py
+++ b/pyro/poutine/broadcast_messenger.py
@@ -34,7 +34,7 @@ class BroadcastMessenger(Messenger):
                     target_batch_shape[f.dim] = f.size
             # Starting from the right, if expected size is None at an index,
             # set it to the actual size if it exists, else 1.
-            for i in range(-len(target_batch_shape)+1, 1):
+            for i in range(-len(target_batch_shape) + 1, 1):
                 if target_batch_shape[i] is None:
                     target_batch_shape[i] = actual_batch_shape[i] if len(actual_batch_shape) > -i else 1
             msg["fn"] = msg["fn"].expand(target_batch_shape)
--- a/pyro/poutine/broadcast_messenger.py
+++ b/pyro/poutine/broadcast_messenger.py
@@ -32,8 +32,8 @@ class BroadcastMessenger(Messenger):
                             .format(msg["name"], f.name, f.dim),
                             'Try setting dim arg in other iaranges.']))
                     target_batch_shape[f.dim] = f.size
-            # If expected shape is None at an index, infer as either 1,
-            # or the actual shape starting from the right.
+            # Starting from the right, if expected size is None at an index,
+            # set it to the actual size if it exists, else 1.
             for i in range(-len(target_batch_shape)+1, 1):
                 if target_batch_shape[i] is None:
                     target_batch_shape[i] = actual_batch_shape[i] if len(actual_batch_shape) > -i else 1
--- a/pyro/poutine/broadcast_messenger.py
+++ b/pyro/poutine/broadcast_messenger.py
@@ -1,0 +1,40 @@
+from __future__ import absolute_import, division, print_function
+
+from .messenger import Messenger
+
+
+class BroadcastMessenger(Messenger):
+    """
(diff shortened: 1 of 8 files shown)


pyro/util.py
```diff
@@ -218,6 +218,7 @@ def check_site_shape(site, max_iarange_nesting):
     if max_iarange_nesting < len(actual_shape):
         actual_shape = actual_shape[len(actual_shape) - max_iarange_nesting:]
 
+    expected_shape = broadcast_shape(expected_shape, actual_shape)
```
jpchen: we should do strict checking here iiuc that the flag toggles allowing reshaping

neerajprad:
> is this idempotent? in the sense that if i wrap my models in poutine.broadcast but then dont nest/use broadcasting in my iaranges should everything still work?

It should work; but this is worth adding as a test.

neerajprad:
Removed generic broadcasting, and added test for idempotence.

pyro/poutine/broadcast_messenger.py
```diff
@@ -0,0 +1,40 @@
...
+        """
+        :param msg: current message at a trace site.
+        """
+        if msg["done"] or msg["type"] != "sample":
+            return
+
+        dist = msg["fn"]
+        actual_batch_shape = getattr(dist, "batch_shape", None)
+        if actual_batch_shape is not None:
+            target_batch_shape = []
```
fritzo: Nice! I think this could be a little stricter and a little simpler if you used `-1` sizes, something like
```py
target_batch_shape = [-1 if size == 1 else size for size in actual_batch_shape]
for f in msg["conde_indep_stack"]:
    if f.dim is None:
        continue
    assert f.dim < 0
    if -f.dim > len(target_batch_shape):
        target_batch_shape = [-1] * (-f.dim - len(target_batch_shape)) + target_batch_shape
    elif target_batch_shape[f.dim] not in (-1, f.size):
       raise ValueError("... dim collision ...")
    target_batch_shape[f.dim] = f.size
    msg["fn"] = msg["fn"].expand(target_batch_shape)
```

tests/infer/test_gradient.py
```diff
@@ -73,31 +74,30 @@ def test_iarange(Elbo, reparameterized):
...
-            with pyro.iarange("data", len(data)):
-                pyro.sample("z", Normal(mus, scale))
+        with pyro.iarange("particles", num_particles, dim=-2):
+            with pyro.iarange("data", len(data), dim=-1):
+                pyro.sample("z", Normal(loc, scale))
         pyro.sample("nuisance_b", Normal(2, 3))
         pyro.sample("nuisance_a", Normal(0, 1))
 
     optim = Adam({"lr": 0.1})
+    model, guide = poutine.broadcast(model), poutine.broadcast(guide)
```
fritzo: nit: I think it's a little clearer to decorate
```py
@poutine.broadcast
def model():
    ...

@poutine.broadcast
def guide():
    ...
```
That way readers know as soon as they start reading the model that it should be read with broadcast semantics.

pyro/poutine/broadcast_messenger.py
neerajprad: The problem that I was facing was with expanding smaller-sized latent sites when the iarange dims appear in a staggered fashion:

```python
with pyro.iarange("num_particles", 10, dim=-3):
    with pyro.iarange("components", 2, dim=-1):
         # with .expand([10, -1, 2]), we get s.shape == (10,)
         # with .expand([10, 1, 2]), we get s.shape == (10, 1, 2)
         s = pyro.sample("sample", dist.Bernoulli(0.5))    
```

tests/infer/test_valid_models.py
```diff
@@ -826,3 +827,55 @@ def model():
...
+            with pyro.iarange("components", 2, dim=-1):
+                p = pyro.sample("p", dist.Beta(torch.tensor(1.1), torch.tensor(1.1)))
+                assert p.shape == torch.Size((10, 1, 2))
+            with pyro.iarange("data", data.shape[0], dim=-2):
+                pyro.sample("obs", dist.Bernoulli(p), obs=data)
+
+    def guide():
+        with pyro.iarange("num_particles", 10, dim=-3):
+            with pyro.iarange("components", 2, dim=-1):
+                pyro.sample("p", dist.Beta(torch.tensor(1.1), torch.tensor(1.1)))
```
eb8680: Maybe put this or something similar in the `handlers.broadcast` docstring as a usage example?

pyro/poutine/broadcast_messenger.py
fritzo: The snippet I suggested is order invariant. Note the final `.expand()` statement is outside of the loop.

neerajprad: Just updated the snippet to clarify. The problem comes in the sample site in the second `iarange` where the outermost iarange is at -3 and the inner `components` one is at -1 (the `data` iarange is yet to come). If we expand `s` as `dist.Bernoulli(0.5).expand([10, -1, 2])` (note the default `-1`) it will give us `s.shape == torch.Size((10,))` whereas we want it to be of `torch.Size((10, 1, 2))` so that it can be correctly broadcasted by the `data` iarange.

tests/infer/test_valid_models.py
neerajprad: Will update.

pyro/poutine/broadcast_messenger.py
fritzo: Thanks for explaining, I didn't know that -1 cannot be used for new dimensions in `torch.expand()`. Could we still try to catch expand errors early, while the `iarange.name` is still around? Maybe
```py
target_batch_shape = [None if size == 1 else size for size in actual_batch_shape]
for f in msg["cond_indep_stack"]:
    if f.dim is None:
        continue
    assert f.dim < 0
    if -f.dim > len(target_batch_shape):
        target_batch_shape = [None] * (-f.dim - len(target_batch_shape)) + target_batch_shape
    elif target_batch_shape[f.dim] not in (None, f.size):
        raise ValueError("Shape mismatch inside iarange('{}') at site {} dim {}, {} vs {}".format(
            f.name, msg['name'], f.dim, f.size, target_batch_shape[dim]))
    target_batch_shape[f.dim] = f.size
# ... remainder of your code ...
```

neerajprad: Good point; will update.

Issue #1063, opened 2018-04-23
Support PyTorch JIT compilation


fritzo:
What would it take to support JIT-compilation of an entire `SVI.step()`?

For motivation, consider [a beta bernoulli model](https://forum.pyro.ai/t/comparison-of-beta-bernoulli-process-in-edward-and-pyro/175) which is 10x slower in Pyro than Edward.

## Feasibility

PyTorch `@jit.compile` appears to be functional input->output. Therefore we would need to at least
1. run a warm-up iteration to populate the param store
2. expose the Pyro param store to the jit compiler
3. expose a helper `.differentiable_loss()` inside `loss_and_grads` (everything up to `.backward()`)

## Tasks

- [x] assess feasibility
- [x] #1109 refactor ELBO to expose `.differentiable_loss()` (see also #628)
- [x] #1126  add an abstract jit decorator to make it easy to write new jitted functions
- [x] #1276 Change to using `torch.jit.trace` instead of the deprecated `torch.jit.compile`
- [x] https://github.com/pytorch/pytorch/issues/10041 Support `broadcast_all()` in the jit
- [x] https://github.com/pytorch/pytorch/pull/10321 Support `Categorical.log_prob()` in the jit
- [x] #1296 add `--jit` option to SVI examples
- [x] #1299 support jit in HMC
- ~~add tutorial~~

## Remaining PyTorch issues

To see remaining PyTorch jit issues:
1. Install PyTorch master
    ```sh
    cd ~/github/pytorch/pytorch  # or wherever your clone lives
    git checkout master
    git pull
    python setup.py build develop
    ```
2. Run Pyro's jit tests with the `pytorch-0.4.1` branch of Pyro
    ```sh
    cd ~/github/uber/pyro  # or wherever your clone lives
    git checkout pytorch-0.4.1
    git pull
    make install
    make test-jit  # saves results to jit.log
    ```
Here is an example [jit.log](http://fritzo.org/temp/jit.log).

eb8680:
Are any `torch.distributions` samplers or log-probabilities supported by the JIT at all?  Does the JIT have the same broadcasting semantics?

fritzo:
I believe this is closed by #1431

PR #1126, opened 2018-05-05
Add helper decorator for JITing parametrized functions

eb8680:
Addresses #1063 

This PR adds a helpful utility `pyro.ops.jit.compile` that serves as a drop-in replacement for `torch.jit.compile` that is compatible with functions that call `pyro.param` internally.  Every such function is implicitly a closure over the Pyro parameter store, but `torch.jit.compile` does not allow closures over tensors.  `pyro.ops.jit.compile` enables this use case without passing the entire parameter store as an argument and also ensures that parameters are always exposed, even after compilation.

This makes the JIT easier to use for common Pyro use cases like computing loss functions in SVI or the energy in HMC that depend on parameters:
```python
def model(x):
    scale = pyro.param("scale", torch.tensor(0.5), constraint=constraints.positive)
    return pyro.sample("y", dist.Normal(x, scale))

@pyro.ops.jit.compile(nderivs=1)
def model_log_prob_fn(x, y):
    cond_model = pyro.condition(model, data={"y": y})
    tr = pyro.poutine.trace(cond_model).get_trace(x)
    return tr.log_prob_sum()

...

# even after compilation, model_log_prob_fn still exposes params
with poutine.trace(param_only=True) as param_capture:
    model_log_prob_fn(torch.tensor(1.0), torch.tensor(2.0))

assert "p" in param_capture.trace
```

The PR also slightly refactors `JitTrace*_ELBO` to use `pyro.ops.jit.compile` instead of `torch.jit.compile` and modifies `ReplayMessenger` to support replaying of parameters.


pyro/infer/trace_elbo.py
```diff
@@ -166,18 +165,11 @@ class JitTrace_ELBO(Trace_ELBO):
...
-                        pass
-            self._param_names = list(param_capture.trace.nodes.keys())
-
             # build a closure for loss_and_surrogate_loss
             weakself = weakref.ref(self)
 
-            @torch.jit.compile(nderivs=1)
-            def loss_and_surrogate_loss(args_list, param_list):
+            @pyro.ops.jit.compile(nderivs=1)
+            def loss_and_surrogate_loss(*args_list):
```
fritzo: nit: I guess we can just call this `*args` now to match `Trace_ELBO.loss_and_grads()` idioms?

eb8680:
GitHub's weird new merging behavior cluttered up the commit history...

pyro/infer/traceenum_elbo.py
```diff
@@ -171,30 +170,22 @@ class JitTraceEnum_ELBO(TraceEnum_ELBO):
...
             weakself = weakref.ref(self)
 
-            @torch.jit.compile(nderivs=1)
-            def differentiable_loss(args_list, param_list):
+            @pyro.ops.jit.compile(nderivs=1)
+            def differentiable_loss(*args_list):
                 self = weakself()
                 elbo = 0.0
-                for model_trace, guide_trace in self._get_traces(model, guide, *args_list, **kwargs):
+                for model_trace, guide_trace in self._get_traces(model, guide,
```
fritzo: nit: Can we replace this with a single line to match other uses in the file:
```py
```

pyro/infer/tracegraph_elbo.py
```diff
@@ -303,18 +304,11 @@ class JitTraceGraph_ELBO(TraceGraph_ELBO):
...
-                        pass
-            self._param_names = list(param_capture.trace.nodes.keys())
-
             # build a closure for loss_and_surrogate_loss
             weakself = weakref.ref(self)
 
-            @torch.jit.compile(nderivs=1)
-            def loss_and_surrogate_loss(args_list, param_list):
+            @pyro.ops.jit.compile(nderivs=1)
+            def loss_and_surrogate_loss(*args_list):
```
fritzo: nit: ditto `*args`

pyro/ops/jit.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Example::
+
+        def model(x):
+            scale = pyro.param("scale", torch.tensor(0.5), constraint=constraints.positive)
+            return pyro.sample("y", dist.Normal(x, scale))
+
+        @pyro.ops.jit.compile(nderivs=1)
+        def model_log_prob_fn(x, y):
+            cond_model = pyro.condition(model, data={"y": y})
```
fritzo: :+1: that's awesome we support fancy poutine usage!

pyro/ops/jit.py
```diff
@@ -0,0 +1,80 @@
...
+                        fn(*args, **kwargs)
+
+                self._param_names = list(set(first_param_capture.trace.nodes.keys()))
+
+                weakself = weakref.ref(self)
+
+                @torch.jit.compile(**self._kwargs)
+                def compiled(unconstrained_params, *args):
+                    self = weakself()
+                    constrained_params = {name: pyro.param(name, unconstrained_param)
```
fritzo: This isn't doing what it looks like it's doing, and it would be wrong if the unconstrained param were ever used to initialized the param. I think it would be more transparent to
```py
constrained_params = {}
for name, unconstrained_param in zip(self._param_names, unconstrained_params):
    constrained_param = pyro.param(name)  # assume param has been initialized
    assert constrained_param.unconstrained() is unconstrained_param
    constrained_params[name] = constrained_param
```

pyro/ops/jit.py
```diff
@@ -0,0 +1,80 @@
...
+
+                    return poutine.replay(
+                        fn, params=constrained_params)(*args, **kwargs)
+
+                self.compiled = compiled
+
+            param_list = [pyro.param(name).unconstrained()
+                          for name in self._param_names]
+
+            with poutine.block(hide=self._param_names):
```
fritzo: Can you help me undersand why you're blocking and tracing here?

pyro/ops/jit.py
```diff
@@ -0,0 +1,80 @@
...
+                with poutine.trace(param_only=True) as param_capture:
+                    ret = self.compiled(param_list, *args, **kwargs)
+
+            new_params = filter(lambda name: name not in self._param_names,
```
fritzo: In what circumstances would we see new params? Should we handle them gracefully or error?

eb8680: If a new input argument shape configuration triggers recompilation and the resulting new trace has different parameters, we'll need to capture those.

Since `torch.jit.compile` handles this, we should too.

pyro/ops/jit.py
eb8680: I guess I wanted to avoid having `pyro.param` be called twice for params in `self._param_names`, but I'm not sure that matters?

pyro/ops/jit.py
fritzo: Does that mean we should also reset the compiled function, since its signature has changed?
```py
if any(new_params):
    self.compiled = None
```
or will the original version automatically work when different params are passed in?

pyro/ops/jit.py
fritzo: Ok, I see that it makes sense to capture new params, and if you're capturing new params, it makes sense to block. Thanks for explaining.

pyro/poutine/replay_messenger.py
```diff
@@ -8,42 +8,17 @@ class ReplayMessenger(Messenger):
...
+    def __init__(self, trace=None, params=None):
         """
         :param trace: a trace whose values should be reused
 
         Constructor.
         Stores trace in an attribute.
         """
         super(ReplayMessenger, self).__init__()
-        assert trace is not None, "must provide trace"
+        assert trace is not None or params is not None, "must provide trace"
```
fritzo: nit: "must provide either trace or params"

pyro/ops/jit.py
eb8680: The function's signature shouldn't change, since the parameters are always passed in through a list.  I'd prefer to delegate all bookkeeping to `torch.jit.compile`.

pyro/poutine/replay_messenger.py
```diff
@@ -57,19 +32,23 @@ def _pyro_sample(self, msg):
...
-                    self.trace.nodes[g_name]["is_observed"]:
-                raise RuntimeError("site {} must be sample in trace".format(g_name))
+                return None
+            if guide_msg["type"] != "sample" or \
+                    guide_msg["is_observed"]:
+                raise RuntimeError("site {} must be sample in trace".format(name))
             msg["done"] = True
-            msg["value"] = self.trace.nodes[g_name]["value"]
+            msg["value"] = guide_msg["value"]
+            msg["infer"] = guide_msg["infer"]
```
fritzo: Do I understand correctly that we never needed this before since we only replayed models against guides, but now we're replaying guides against guides as well? Or are you merely cleaning up?

tests/poutine/test_poutines.py
```diff
@@ -98,19 +98,6 @@ def test_replay_full(self):
             assert_equal(model_trace.nodes[name]["value"],
                          guide_trace.nodes[name]["value"])
 
-    def test_replay_partial(self):
```
fritzo: Did this break?

pyro/poutine/replay_messenger.py
fritzo: Actually since this is user-facing we should
```py
if trace is None and params is None:
    raise ValueError("must provide either trace or params")
```

tests/poutine/test_poutines.py
eb8680: No, it was a test for a dead code path

pyro/poutine/replay_messenger.py
eb8680: Just cleanup, the sites logic never got used

pyro/ops/jit.py
```diff
@@ -0,0 +1,80 @@
...
+            return pyro.sample("y", dist.Normal(x, scale))
+
+        @pyro.ops.jit.compile(nderivs=1)
+        def model_log_prob_fn(x, y):
+            cond_model = pyro.condition(model, data={"y": y})
+            tr = pyro.poutine.trace(cond_model).get_trace(x)
+            return tr.log_prob_sum()
+    """
+    def __init__(self, **kwargs):
+        self._kwargs = kwargs
```
fritzo: nit: Can we call this `self._jit_options` or something more evocative?

pyro/ops/jit.py
```diff
@@ -0,0 +1,80 @@
...
+        self.compiled = None
+        self._param_names = None
+
+    def __call__(self, fn):
+
+        assert self.compiled is None, \
+            "Cannot use a compile instance on more than one function"
+
+        def wrapper(*args, **kwargs):
```
fritzo: What do you think about adding a check like
```py
if self._fixed_kwargs is None:
    self._fixed_kwargs = kwargs
elif kwargs != self._fixed_kwargs:
    raise NotImplementedError('pyro.jit.compile does not support dynamic **kwargs')
```

eb8680: I don't know if that's true about `torch.jit.compile` though?

fritzo: ok

pyro/ops/jit.py
```diff
@@ -0,0 +1,79 @@
...
+            scale = pyro.param("scale", torch.tensor(0.5), constraint=constraints.positive)
+            return pyro.sample("y", dist.Normal(x, scale))
+
+        @pyro.ops.jit.compile(nderivs=1)
+        def model_log_prob_fn(x, y):
+            cond_model = pyro.condition(model, data={"y": y})
+            tr = pyro.poutine.trace(cond_model).get_trace(x)
+            return tr.log_prob_sum()
+    """
+    class wrapper(object):
```
fritzo: I would recommend defining this big class outside of `compile`:
```py
class CompiledFunction(object):  # formerly class wrapper
    def __init__(self, fn, jit_options):
        ...
```
and then adding decorator magic in the `compile` helper, something like
```
def compile(fn=None, **jit_options):
    if fn is None:
        return lambda fn: compile(fn, **jit_options)
    return CompiledFunction(fn, jit_options) 
```

eb8680:
> Do you want to add any tests in this PR, or should we wait till next week to add tests of e.g. dynamic len(args) and new params?

We can add tests separately

PR #1132, opened 2018-05-08
Allow use_decaying_avg_baseline for TraceGraph_ELBO to be used on GPU

talesa:
Previously attempting `use_decaying_avg_baseline=True` for `TraceGraph_ELBO` on GPU resulted in an error because `torch.tensor(0.0)` was placed on the wrong device.
This change fixes it although it might not be the most optimal way to do it.
Passes tests.


pyro/infer/tracegraph_elbo.py
```diff
@@ -137,7 +137,7 @@ def _compute_elbo_non_reparam(guide_trace, non_reparam_nodes, downstream_costs):
             param_name = "__baseline_avg_downstream_cost_" + node
             with torch.no_grad():
                 avg_downstream_cost_old = pyro.param(param_name,
-                                                     torch.tensor(0.0).expand(dc_shape).clone())
+                                                     guide_site['value'].new_zeros(dc_shape).clone())
```
fehiepsi: you can also remove `.clone()` here

talesa: done!

talesa:
Do you think it would be useful to have a test for `use_decaying_avg_baseline=True` for `TraceGraph_ELBO`?

fritzo:
> Do you think it would be useful to have a test

Most of our existing cuda tests call `torch.set_default_tensor_type('torch.cuda.DoubleTensor')`. It will be a lot of work to test cuda code more thoroughly, and I think that is best done in a consistent way for all our inference algorithms. No need to add partial coverage in this PR.

Issue #1113, opened 2018-05-02
Support dynamic learning rate


slaweku:
Hi,
Pytorch appear to have a capable module for change learning rates while training: 
https://pytorch.org/docs/stable/optim.html?highlight=scheduler#how-to-adjust-learning-rate
But it embraces the optimizer and when I try to do e.g.
scheduler = MultiStepLR(optim, milestones=[5,10], gamma=0.1)
I am getting
TypeError: PyroOptim is not an Optimizer

So, I was trying to use something like this:

LR_dict={5:1e-3, 10:3e-4}
def adjust_learning_rate(optimizer, epoch, lr_dict):
    if epoch in lr_dict.keys():
        new_lr=lr_dict[epoch]
        print("changing learning rate to:",new_lr)
        for param_group in optimizer.param_groups:
            param_group['lr'] = new_lr

but unfortunately 
AttributeError: 'PyroOptim' object has no attribute 'param_groups'

Regards,
Slawek

martinjankowiak:
thanks, i've been meaning to do this. supporting this will take a little bit of extra engineering in pyro. not very much though so we can try to make the necessary changes within a day or two

slaweku:
Thank you!

martinjankowiak:
you can already do this:

```
from pyro.optim import Adam

def per_param_callable(module_name, param_name):
    if param_name == 'my_special_parameter':
        return {"lr": 0.010}
    else:
        return {"lr": 0.001}

optimizer = Adam(per_param_callable)
```
this should give you the control you want although it doesn't actually use the pytorch 'group' construct

slaweku:
Lovely. Thank you!

PR #1133, opened 2018-05-09
Support dynamic learning rate

jpchen:
Addresses #1113

### Usage
```python
pyro_scheduler = ExponentialLR({'optimizer': torch.optim.SGD, 'optim_args': {'lr': 0.01}, 'gamma': 0.1})
SVI(model, guide, pyro_scheduler, Trace_Elbo())
```
`optimizer` is a mandatory kwarg. i went back and forth between a couple possible designs and i think this is the least icky and most consistent with how one would use existing pyro optimizers


pyro/optim/lr_scheduler.py
```diff
@@ -0,0 +1,127 @@
...
+
+import torch
+
+import pyro
+from pyro.params import module_from_param_with_module_name, user_param_name
+
+
+class PyroLRScheduler(object):
+    """
+    A wrapper for torch.optim.lr_scheduler objects that helps with managing dynamically generated parameters.
```
martinjankowiak: that helps schedule learning rates for dynamically generated parameters

pyro/optim/lr_scheduler.py
```diff
@@ -0,0 +1,127 @@
+from __future__ import absolute_import, division, print_function
+
+import torch
+
+import pyro
+from pyro.params import module_from_param_with_module_name, user_param_name
+
+
+class PyroLRScheduler(object):
```
martinjankowiak: given the design you've chosen why is this a separate object instead of subclass of `PyroOptim` or the like?

jpchen: solely to follow the namespace convention of [pytorch's scheduler](https://pytorch.org/docs/stable/optim.html?#torch.optim.lr_scheduler.LambdaLR) (note the LR classes extend `_LRScheduler` not `Optimizer` ). i could make it a `PyroOptim` but then that blurs the distinction between schedulers and optimizers in our interface (which may be fine? not sure)

martinjankowiak:
yeah that's what i suggested to jp originally

pyro/optim/lr_scheduler.py
jpchen: @neerajprad that would require changing the existing interface, as currently our Pyro Optims automatically wrap the torch class, so users use `pyro.optim.Adam` instead of `pyro.optim.PyroOptim(torch.optim.Adam, ...)` which is what i believe you're suggesting, which is also required by the scheduler. what martin is suggesting is to just make the scheduler `pyro.optim.LambdaLR` instead of `pyro.optim.lr_scheduler.LambdaLR`. i think we can do what youre suggesting but that would require a backwards change to the existing optimizers

eb8680: One question is whether we want the scheduler to be automatically called in `SVI.step`.  The PyTorch idiom is to call them per epoch rather than per update, suggesting we should perhaps keep it separate from `PyroOptim` as in the current design.

jpchen:
ready to merge

martinjankowiak:
surely by "ready to merge" you mean "ready for more CR"

jpchen:
yes pending additional comments, its ready to merge

tests/optim/test_optim.py
```diff
@@ -75,6 +75,47 @@ def optim_params(module_name, param_name):
...
+                                       optim.StepLR,
+                                       optim.ExponentialLR,
+                                       ])
+def test_dynamic_lr(scheduler):
+    pyro.clear_param_store()
+    gamma = 2
+    step_size = 1
+    optimizer = torch.optim.SGD
+    if scheduler is optim.LambdaLR:
+        pyro_scheduler = scheduler({'optimizer': optimizer, 'optim_args': {'lr': 0.01},
```
neerajprad: nit: why not just pass into the parameters themselves:

```
@pytest.mark.parametrize('scheduler', [optim.LambdaLR(..scheduler args..),)
```

or separately as `scheduler_args` if you prefer. That makes the test free of any parametrization logic.

pyro/optim/lr_scheduler.py
```diff
@@ -0,0 +1,68 @@
+from __future__ import absolute_import, division, print_function
+
+import pyro
+from pyro.optim.optim import PyroOptim
+
+
+class PyroLRScheduler(PyroOptim):
```
neerajprad: Could you add this to the docs as well?

tests/optim/test_optim.py
jpchen: i had this originally but it looks really gross to make it pep8 compatible (a lot of line breaks to make it < 120 char). i can do it if you think its cleaner?

jpchen:
> a callable passed to PyroOptim can return pytorch schedulers
so are you imagining something like this:
```python
def scheduler(args):
    return torch.optim.lr_scheduler.LambdaLR(args)
opt = PyroOptim(scheduler, args)
SVI(..., opt, ...)
```
? there are args for the optimizer and args for the scheduler, how are they distinguished?
> why do you think this is better?
i found this to be the best middle ground between how pyro optimizers are currently used and how a `torch.optim.lr_scheduler` is used

tests/optim/test_optim.py
neerajprad: You could make more space by breaking after the `[`:

```
@pytest.mark.parametrize('scheduler', [
    (param1),
    (param2),
    ..]
```

and each param could take multiple lines, if needed. That'll probably be cleaner, but I'll leave it to your judgement.

neerajprad:
Won't you also need to pass `epoch` or any other auxiliary info from `SVI` for some of these schedulers?

eb8680:
> Won't you also need to pass epoch or any other auxiliary info from SVI for some of these schedulers?

I think so, which is why I suggested keeping schedulers separate from `SVI.step`

jpchen:
if i understand your question correctly, svi is just like any other loss; because of how pyro optimizers work instead of 1 optimizer, we have `n` where n is the number of parameters, so they still act as youd expect an optimizer to work in a normal training loop. the same for schedulers. all the other hyperparameters are passed explicitly to the scheduler beforehand.

pyro/optim/lr_scheduler.py
```diff
@@ -0,0 +1,68 @@
...
+                self.optim_objs[p] = scheduler
+
+                # set state from _state_waiting_to_be_consumed if present
+                param_name = pyro.get_param_store().param_name(p)
+                if param_name in self._state_waiting_to_be_consumed:
+                    state = self._state_waiting_to_be_consumed.pop(param_name)
+                    self.optim_objs[p].load_state_dict(state)
+
+            # actually perform the step for the optim object
+            self.optim_objs[p].step(*args, **kwargs)
```
neerajprad: My point of confusion is that we are not passing in auxiliary information in `*args, **kwargs` here as this is just called from `self.optim(params)` in `SVI`. For instance, for the `optim.LambdaLR` scheduler, when do we pass in the `epoch` to adjust the learning rate? Maybe I'm missing something.

jpchen: it is stateful - it [keeps track of the epoch in the scheduler object](https://github.com/pytorch/pytorch/blob/master/torch/optim/lr_scheduler.py#L53)

tests/optim/test_optim.py
```diff
@@ -75,6 +75,38 @@ def optim_params(module_name, param_name):
...
+        return pyro.sample('obs', Normal(sample, torch.tensor(0.2)), obs=torch.tensor(0.1))
+
+    def guide():
+        loc = pyro.param('loc', torch.tensor(0.))
+        scale = pyro.param('scale', torch.tensor(0.5))
+        pyro.sample('latent', Normal(loc, scale))
+
+    svi = SVI(model, guide, scheduler, loss=TraceGraph_ELBO())
+    for epoch in range(2):
+        svi.step()
```
fritzo: Can you add a test with multiple steps per epoch?

pyro/optim/lr_scheduler.py
fritzo: @jpchen Maybe you could add a `.set_epoch()` method to the scheduler interface, since we're already eating all the `*arg,**kwarg`s in `SVI.step()`.

(and remember to increment your counter `num_times_args_kwargs_led_to_complexity_later_on += 1`)

neerajprad: >  it keeps track of the epoch in the scheduler object

I think that will end up incrementing the epoch per mini-batch by default.

tests/optim/test_optim.py
fritzo: The PyTorch lr_scheduler interface has an optional `epoch` argument that would be used like
```py
for epoch in range(10):
    for minibatch in batcher(data):
        compute_loss(minibatch).backward()
        optim.step(epoch=epoch)
```
You'll need more plumbing to use the `epoch` kwarg in Pyro, something like
```py
optim = LambdaLR(...)
svi = SVI(model, guide, optim, Trace_ELBO())
for epoch in range(10):
    for minibatch in batcher(data):
        optim.set_epoch(epoch)  # inform the scheduler of the latest epoch
        svi.step()
```

jpchen: on `step` - yeah i plumbed that through but `last_epoch` behaves weirdly. it should use the index of the epoch you specify. let me investigate it a bit i may be using it wrong

tests/optim/test_optim.py
```diff
@@ -75,6 +75,54 @@ def optim_params(module_name, param_name):
...
+            assert opt.state_dict()['param_groups'][0]['lr'] == 0.02
+            assert opt.state_dict()['param_groups'][0]['initial_lr'] == 0.01
+            opt = scheduler.optim_objs[scale].optimizer
+            assert opt.state_dict()['param_groups'][0]['lr'] == 0.02
+            assert opt.state_dict()['param_groups'][0]['initial_lr'] == 0.01
+
+    # test multi-step
+    pyro.clear_param_store()
+    for epoch in range(2):
+        scheduler.set_epoch(epoch)
```
jpchen: @fritzo could also use the attribute directly `scheduler.epoch = epoch`

tests/optim/test_optim.py
```diff
@@ -75,6 +75,54 @@ def optim_params(module_name, param_name):
...
+            loc = pyro.param('loc')
+            scale = pyro.param('scale')
+            opt = scheduler.optim_objs[loc].optimizer
+            assert opt.state_dict()['param_groups'][0]['lr'] == 0.02
+            assert opt.state_dict()['param_groups'][0]['initial_lr'] == 0.01
+            opt = scheduler.optim_objs[scale].optimizer
+            assert opt.state_dict()['param_groups'][0]['lr'] == 0.02
+            assert opt.state_dict()['param_groups'][0]['initial_lr'] == 0.01
+
+    # test multi-step
```
fritzo: nit: consider a parametrized test?
```py
@pytest.mark.parametrize('steps_per_epoch', [1, 2])
def test_dynamic_lr(scheduler, steps_per_epoch):
    ...
    for epoch in range(2):
        scheduler.set_epoch(epoch)
        for _ in range(steps_per_epoch):
            svi.step()
```

pyro/optim/lr_scheduler.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def __call__(self, params,  *args, **kwargs):
+        """
+        :param params: a list of parameters
+        :type params: an iterable of strings
+
+        Do an optimization step for each param in params. If a given param has never been seen before,
+        initialize an optimizer for it.
+        """
+        kwargs['epoch'] = self.epoch
```
fritzo: Nice, this is simpler than I expected!

jpchen:
> Why not instead implement a single class

yes lets sync up offline. i did something similar originally (wrapping pyro optim), but i thought that was more cognitive overhead, to use a combination of pyro and torch optimizers so i tried to make it as analogous as possible to how one would use it in torch (`scheduler(optimizer, args)`) in this version

>(not to be subclassed)

isnt the snippet you gave also subclassed?

pyro/optim/optim.py
```diff
@@ -40,14 +43,19 @@ def __call__(self, params,  *args, **kwargs):
         Do an optimization step for each param in params. If a given param has never been seen before,
         initialize an optimizer for it.
         """
-
+        if hasattr(self, 'pt_scheduler_constructor'):
```
fritzo: nit: `hasattr` smells bad; does it make sense to remove these two lines and instead
```py
class PyroLRScheduler(PyroOptim):
    def __call__(self, params,  *args, **kwargs):
        kwargs['epoch'] = self.epoch
        super(PyroLRScheduler, self).__call__(params, *args, **kwargs)
```

pyro/optim/optim.py
```diff
@@ -40,14 +43,19 @@ def __call__(self, params,  *args, **kwargs):
...
         for p in params:
             # if we have not seen this param before, we instantiate and optim object to deal with it
             if p not in self.optim_objs:
                 # get our constructor arguments
                 def_optim_dict = self._get_optim_args(p)
                 # create a single optim object for that param
-                self.optim_objs[p] = self.pt_optim_constructor([p], **def_optim_dict)
+                optim = self.pt_optim_constructor([p], **def_optim_dict)
+                # PyroOptim is a scheduler
+                if hasattr(self, 'pt_scheduler_constructor'):
```
fritzo: nit: `hasattr` smells bad; does it make sense to define two methods
```py
class PyroOptim(object):
    def _get_optim(self, p):
        return self.pt_optim_constructor([p], **self._get_optim_args(p))

class PyroLRScheduler(PyroOptim):
    def _get_optim(self, p):
        optim = super(PyroLRScheduler, self)._get_optim(p)
        return self.pt_scheduler_constructor(optim, **self.kwargs)
```
and then replace this code with
```py
self.optim_objs[p] = self._get_optim(p)
```

PR #1139, opened 2018-05-15
Add pyro.ops.newton.newton_step_2d

fritzo:
This implements a `newton_step_2d()` op for batched 2-dimensional tensors with twice-differentiable loss.

This also exposes the `pyro.ops` module in docs as "Miscellaneous ops". I'm open to moving this around, e.g. moving the docs for `pyro.ops.newton` into the "Optimization" section. However I foresee more ops being added in the future. In the future we may add `newton_step_3d()`, `newton_step_scalar()`, and possibly some approximate newton methods for larger matrices.

- [x] add docs
- [x] add tests of update
- [x] add tests of gradient
- [x] add tests of covariance
- [x] add tests for trust region bound
- [x] fix xfailing tests


pyro/ops/newton.py
```diff
@@ -0,0 +1,79 @@
...
+        raise ValueError('Expected loss to be a scalar, actual shape {}'.format(loss.shape))
+    if x.dim() < 1 or x.shape[-1] != 2:
+        raise ValueError('Expected x to have rightmost size 2, actual shape {}'.format(x.shape))
+
+    # compute derivatives
+    g = grad(loss, [x], create_graph=True)[0]
+    H = torch.stack([grad(g[..., 0].sum(), [x], create_graph=True)[0],
+                     grad(g[..., 1].sum(), [x], create_graph=True)[0]], -1)
+    assert g.shape[-1:] == (2,)
+    assert H.shape[-2:] == (2, 2)
```
jpchen: should we add some validation checks/nan warnings for the grad?

pyro/ops/newton.py
```diff
@@ -0,0 +1,90 @@
...
+    ``loss`` must be twice-differentiable as a function of ``x``. If ``loss``
+    is ``2+d``-times differentiable, then the return value of this function is
+    ``d``-times differentiable.
+
+    When ``loss`` is interpreted as a negative log probability density, then
+    the return value of this function can be used to construct a Laplace
+    approximation ``MultivariateNormal(mode,cov)``.
+
+    .. warning:: Take care to detach the result of this function when used in
+        an optimization loop. If you forget to detach the result of this
```
eb8680: Is there any extra protection we could/should add to guard against this?

fritzo: The only way I see to add protection is to pass in a `loss_fn` callback and perform those ops for the user, but that feels quite inflexible. I prefer the unix style simple-function-with-provisos.

fritzo:
OK, I've added stronger tests and some NAN warnings.

Issue #628, opened 2017-12-14
Migrate from .backward() to torch.autograd.grad()


fritzo:
## Why?

A number of past algorithm development efforts have been made overly complex due to the side-effectful nature of `.backward()` (e.g. MAP #573 , HMC #611 , LAX in `TraceGraph_ELBO` https://github.com/uber/pyro/tree/lax, early attempts to support `enum_discrete` in `TraceGraph_ELBO`, Laplace approximation #590 ). These efforts would likely be easier if we switched from `.backward()` to `torch.autograd.grad(...)` in a few places in Pyro (maybe everywhere).

## How?

PyTorch has two autograd interfaces: `.backward()` is more side-effectful and `torch.autograd.grad(...)` is more functional. Pyro currently uses `.backward()` almost everywhere ([verlet integrator](https://github.com/uber/pyro/blob/dev/pyro/ops/integrator.py) is an exception). By replacing `.backward()` with more functional handling of gradients returned by `grad(...)` in Pyro's ELBO implementations, we could improve modularity.

Currently PyTorch optimizers do not support use of external gradients. @soumith [is open to the possibility](https://discuss.pytorch.org/t/using-optimizers-with-torch-autograd-grad/11078) of adding this support in 0.4. To ease migration Pyro could provide a temporary workaround in the `pyro.optim` layer to manually set `.grad` attributes of optimized tensors.

## Tasks

- [ ] Work with PyTorch devs to support `grad(...)` in torch optimizers https://github.com/pytorch/pytorch/issues/4179
- [ ] Replace `.backward()` with `grad(...)` in `Trace_ELBO`.
- [ ] Replace `.backward()` with `grad(...)` in `TraceGraph_ELBO`.

martinjankowiak:
yeah let's do this. let me finish my v0 lax implementation first : )

ngoodman:
does this have implications for performance? in particular, are there in-place gradients that we do using `.backward()` that we wouldn't be able to do in place with `.grad()`?

martinjankowiak:
@ngoodman it may. in many cases it probably won't but we'll definitely have to do some tests and think things through carefully before we commit to anything. one place i could imagine a performance hit is elbo with `num_particles>1`.

fritzo:
The motivating problems appear solved now that `Trace_ELBO` and `TraceEnum_ELBO` have `.differentiable_loss()` methods. Closing this issue for now; we can re-open if other motivating problems arise.

PR #1140, opened 2018-05-16
Implement interface for higher-order optimizers

fritzo:
This implements an interface compatible with higher-order optimizers, those that use not only gradient information but also some number of Hessian-vector products. The motivating optimizer is `Newton2d`. Additionally we've implemented wrappers for Pyro optimizers and PyTorch optimizers, and added a container class to apply different optimizers to different parameters.

Docs: http://fritzo.org/temp/pyro/html/optimization.html#module-pyro.optim.multi

## Remaining work

- [x] support differentiation through the optimization process
- [ ] expose `cov` return value from `newton_step_2d()`

## Tested

- added convergence tests for all new classes. tests run in under 2 seconds.

Pair coded with @eb8680

### DIFF
--- a/pyro/optim/multi.py
+++ b/pyro/optim/multi.py
@@ -73,11 +73,7 @@ class PyroMultiOptimizer(MultiOptimizer):
         self.optim = optim
 
     def step(self, loss, params):
-        names = []
-        values = []
-        for name, value in params.items():
-            names.append(name)
-            values.append(value)
+        values = params.values()
         grads = torch.autograd.grad(loss, values, create_graph=True)
         for x, g in zip(values, grads):
             x.grad = g
--- a/pyro/optim/multi.py
+++ b/pyro/optim/multi.py
@@ -1,0 +1,114 @@
+from __future__ import absolute_import, division, print_function
+
+import torch
+
+from pyro.ops.newton import newton_step_2d
+from pyro.optim.optim import PyroOptim
+
+
+class MultiOptimizer(object):
+    """
+    Base class of optimizers that make use of higher-order derivatives.
+
+    Higher-order optimizers generally use :func:`torch.autograd.grad` rather
+    than :meth:`torch.Tensor.backward`, and therefore require a different
+    interface from usual Pyro and PyTorch optimizers. In this interface,
+    the :meth:`step` method inputs a ``loss`` tensor to be differentiated,
+    and backpropagation is triggered one or more times inside the optimizer.
+
+    Derived classes must implement :meth:`step` to compute derivatives and
+    update parameters in-place.
+
+    Example::
+
+        tr = poutine.trace(model).get_trace(*args, **kwargs)
+        loss = -tr.log_prob()
+        params = {name: pyro.param(name).unconstrained()
+                  for name in pyro.get_param_store().get_all_param_names()}
+        optim.step(loss, params)
+    """
+    def step(self, loss, params):
+        """
+        Performs an optimization step on parameters given a differentiable
+        ``loss`` tensor.
+
+        :param torch.Tensor loss: A differentiable tensor to be minimized.
+            Some optimizers require this to be differentiable multiple
+            times.
+        :param dict params: A dictionary mapping param name to unconstrained
+            value as stored in the param store.
+        """
+        raise NotImplementedError
+
+
+class PyroMultiOptimizer(MultiOptimizer):
+    """
+    Facade to wrap :class:`~pyro.optim.optim.PyroOptim` objects
+    in a :class:`MultiOptimizer` interface.
+    """
+    def __init__(self, optim):
+        self.optim = optim
+
+    def step(self, loss, params):
+        names = []
+        values = []
+        for name, value in params.items():
+            names.append(name)
+            values.append(value)
+        grads = torch.autograd.grad(loss, values, create_graph=True)
+        for x, g in zip(values, grads):
+            x.grad = g
+        self.optim(values)
+
+
+class TorchMultiOptimizer(PyroMultiOptimizer):
+    """
+    Facade to wrap :class:`~torch.optim.Optimizer` objects
+    in a :class:`MultiOptimizer` interface.
+    """
+    def __init__(self, optim_constructor, optim_args):
+        optim = PyroOptim(optim_constructor, optim_args)
+        super(TorchMultiOptimizer, self).__init__(optim)
+
+
+class MixedMultiOptimizer(MultiOptimizer):
+    """
+    Container class to combine different :class:`MultiOptimizer` instances for
+    different parameters.
+
+    :param list parts: A list of ``(names, optim)`` pairs, where each
+        ``names`` is a list of parameter names, and each ``optim`` is a
+        :class:`MultiOptimizer` object to be used for the named parameters.
+        Together the ``names`` should partition up all desired parameters to
+        optimize.
+    """
+    def __init__(self, parts):
+        self.parts = list(parts)
+
+    def step(self, loss, params):
+        for names_part, optim in self.parts:
+            optim.step(loss, {name: params[name] for name in names_part})
+
+
+class Newton2d(MultiOptimizer):
+    """
+    Implementation of :class:`MultiOptimizer` that performs a Newton update
+    on batched 2d variables, optionally regularizing via a per-parameter
+    ``trust_radius``. See :func:`~pyro.ops.newton.newton_step_2d` for details.
+
+    :param dict trust_radii: a dict mapping parameter name to radius of trust
+        region. Missing names will use unregularized Newton update, equivalent
+        to infinite trust radius.
+    """
+    def __init__(self, trust_radii={}):
+        self.trust_radii = trust_radii
+
+    def step(self, loss, params):
+        updated_values = {}
+        for name, value in params.items():
+            trust_radius = self.trust_radii.get(name)
+            updated_value, cov = newton_step_2d(loss, value, trust_radius)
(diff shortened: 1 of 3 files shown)


pyro/optim/multi.py
```diff
@@ -0,0 +1,114 @@
...
+        self.trust_radii = trust_radii
+
+    def step(self, loss, params):
+        updated_values = {}
+        for name, value in params.items():
+            trust_radius = self.trust_radii.get(name)
+            updated_value, cov = newton_step_2d(loss, value, trust_radius)
+            updated_values[name] = updated_value.detach()
+        for name, value in params.items():
+            with torch.no_grad():
```
fritzo: We'll need to avoid the `no_grad` to support differentiating through the optimization process, but I believe that will require avoiding in-place tensor update and instead using `ParamStore.replace_param()`. I'm not sure the cleanest way to do this.

pyro/optim/multi.py
```diff
@@ -0,0 +1,114 @@
...
+        to infinite trust radius.
+    """
+    def __init__(self, trust_radii={}):
+        self.trust_radii = trust_radii
+
+    def step(self, loss, params):
+        updated_values = {}
+        for name, value in params.items():
+            trust_radius = self.trust_radii.get(name)
+            updated_value, cov = newton_step_2d(loss, value, trust_radius)
```
fritzo: We'll need to return this, say through a metadata dict return value from `MultiOptimizer.step()`

pyro/optim/multi.py
```diff
@@ -0,0 +1,114 @@
...
+        :class:`MultiOptimizer` object to be used for the named parameters.
+        Together the ``names`` should partition up all desired parameters to
+        optimize.
+    """
+    def __init__(self, parts):
+        self.parts = list(parts)
+
+    def step(self, loss, params):
+        for names_part, optim in self.parts:
+            optim.step(loss, {name: params[name] for name in names_part})
```
eb8680: Does this behave correctly when each optimizer takes multiple derivatives of `loss`?  Does it require special care from the user or optimizer, e.g. using `retain_graph=True` in `grad` or not calling `backward`?

pyro/optim/multi.py
eb8680: Is `cov` meant to be used for a Laplace approximation?

pyro/optim/multi.py
eb8680: Seems like we should just make it easier to update the param store as discussed yesterday, perhaps by adding a `force` flag or `set_tensor` kwarg to `param(...)`.

tests/optim/test_multi.py
```diff
@@ -0,0 +1,49 @@
...
+import pyro
+import pyro.distributions as dist
+import pyro.optim
+import pyro.poutine as poutine
+from pyro.optim.multi import MixedMultiOptimizer, Newton2d, PyroMultiOptimizer, TorchMultiOptimizer
+from tests.common import assert_equal
+
+FACTORIES = [
+    lambda: PyroMultiOptimizer(pyro.optim.Adam({'lr': 0.05})),
+    lambda: TorchMultiOptimizer(torch.optim.Adam, {'lr': 0.05}),
```
eb8680: Should we test more Pyro and Torch optimizers?  Are there any we'd expect to fail (LBFGS?)?

pyro/optim/multi.py
fritzo: Yes this should behave correctly since we've taken care to set `create_graph=True` when calling `grad()` in `PyroMultiOptimizer`. All derived classes of `MultiOptimizer` will set `create_graph=True` so they are composable.

pyro/optim/multi.py
fritzo: Yes `cov` can be interpreted as a Laplace approximation. In one application we use `cov` to filter out bad parameter values.

pyro/optim/multi.py
eb8680: This PR is fine as-is but we should do that next

pyro/optim/multi.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def step(self, loss, params):
+        names = []
+        values = []
+        for name, value in params.items():
+            names.append(name)
+            values.append(value)
+        grads = torch.autograd.grad(loss, values, create_graph=True)
+        for x, g in zip(values, grads):
+            x.grad = g
```
fehiepsi: I think that the usual behavior of a Pytorch `optimizer` accumulates gradients, not set `.grad` like this. How about use [torch.autograd.backward](https://pytorch.org/docs/master/autograd.html#torch.autograd.backward) here instead? Something like this:
```
grads = torch.autograd.grad(loss, values, create_graph=True)
torch.autograd.backward(values, grads)
```

pyro/optim/multi.py
```diff
@@ -0,0 +1,114 @@
...
+class PyroMultiOptimizer(MultiOptimizer):
+    """
+    Facade to wrap :class:`~pyro.optim.optim.PyroOptim` objects
+    in a :class:`MultiOptimizer` interface.
+    """
+    def __init__(self, optim):
+        self.optim = optim
+
+    def step(self, loss, params):
+        names = []
```
fehiepsi: I guess we don't need to store this list in this method.

fehiepsi: So I propose we use `params` as a list (not a dict) of parameters. :)

pyro/optim/multi.py
fritzo: I think we cannot use `.backward()` because backward operates on the entire compute graph reachable (upstream from) the `loss` tensor, whereas we want to restrict to computing gradients only wrt the given variables. This is important for interleaving multiple optimizers via `MixedMultiOptimizer`. Does that seem correct to you?

pyro/optim/multi.py
fritzo: After discussion with @eb8680 we think it would be cleaner to start using dictionaries rather than lists in `PyroOptim` in Pyro 0.3. I also find it cleaner to use dicts rather than lists when combining optimizers in `MixedMultiOptimizer`. Does this sound reasonable?

pyro/optim/multi.py
fehiepsi: I think that `torch.autograd.backward(values, grads)` just operates on the param values, not the entire graph. According to Adam in Pytorch's slack autograd channel (you can search about `torch.autograd.backward` there), this method is slightly faster.

tests/optim/test_multi.py
```diff
@@ -0,0 +1,49 @@
...
+    lambda: Newton2d(trust_radii={'z': 0.2}),
+    lambda: MixedMultiOptimizer([(['y'], PyroMultiOptimizer(pyro.optim.Adam({'lr': 0.05}))),
+                                 (['x', 'z'], Newton2d())]),
+]
+
+
+@pytest.mark.parametrize('factory', FACTORIES)
+def test_optimizers(factory):
+    optim = factory()
```
jpchen: nit: why are you calling lambdas instead of just passing in the object?

pyro/optim/multi.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def step(self, loss, params):
+        values = params.values()
+        grads = torch.autograd.grad(loss, values, create_graph=True)
+        for x, g in zip(values, grads):
+            x.grad = g
+        self.optim(values)
+
+
+class TorchMultiOptimizer(PyroMultiOptimizer):
```
jpchen: what is the purpose of torch multi optimizer since it is essentially a pyro multi optimizer (`optim` is wrapped in `PyroOptim` during initialization anyway)?  is there a use case for this?

fritzo: This is intended to help us explore the possibility of moving towards `grad`-based optimizers rather than `backward`-based `PyroOptim` #628 . @eb8680 and I want to use this assess whether we can get by with the `TorchMultiOptimizer` interface and possibly deprecate the `PyroOptim` interface, moving `PyroOptim`'s logic into `TorchMultiOptimizer`. We do not intend to address whether we should make this refactoring, only whether it is possible.

pyro/optim/multi.py
fritzo: `torch.autograd.backward(values, grads)` "Computes the sum of gradients of given tensors w.r.t. graph leaves.", whereas we want to compute gradients only w.r.t. a subset of the leaves. Thus `backward` is not compatible with `MixedMultiOptimizer`. Adam has commented that `grad` is negligibly faster than `backward`; it performs the same tensor ops and has to walk the autograd graph twice, however this walking is cheap.

fehiepsi: @fritzo I think that in this case, these `values` are leaves, so this method just accumulates the `grads` into their `.grad` attributes (no real `backward` is processed). However, I agree that if your purpose is to not accumulate (`x.grad = g` instead of `x.grad = x.grad + g`), then we don't need to use that hacking statement.

fritzo: I believe the `values` are not leaves but rather roots. By default `values = [loss]` and `grads = [torch.tensor(1.)]`

fehiepsi: I mean `values` and `grads` are `values` and `grads` in this statement `grads = torch.autograd.grad(loss, values, create_graph=True)` and these `values` are params. The code I have mentioned is
```
grads = torch.autograd.grad(loss, values, create_graph=True)
torch.autograd.backward(values, grads)  # not torch.autograd.backward([loss], [torch.tensor(1.)])
```
Sorry if I still miss something.

PR #1142, opened 2018-05-16
Enable doctests to test code snippets in docstrings

### LINKED ISSUE #1130: Enable doctest to keep Pyro docstrings up to date
We should add it to our doc testing stage, and figure out a way to ignore the output that would differ for stochastic functions when tested on different systems.

neerajprad:
Fixes #1130.
 - fixes a number of outdated/buggy code in the docstrings.
 - refactors some code snippets to be independently runnable.
 - `+SKIPS` output of stochastic functions, or certain snippets that are hard to test.

There are a few design considerations with regard to handling imports. 
 1. Import the modules required by the docstrings in the file itself, to avoid imports (often repetitive) within each of the docstrings. 
    - Pros: Relatively straightforward, prevents repetitive imports within the same file. 
    - Cons: Unnecessary imports that may not be needed by the functions themselves - as such these will need to be specially marked as `#noqa` and could lead to circular import issues.
 2. Use [extraglobs](https://docs.python.org/3/library/doctest.html#basic-api), or as suggested by @fehiepsi, the doctest namespace fixture by pytest. 
     - Pros - Cleaner than the previous approach. 
     - Cons - Will be coupled to pytest and won't run via a simple doctest invocation that is supported by most IDEs.
 3. Explicitly define all imports for each code snippet. 
     - Pros - Each code snippet can be copy pasted and is guaranteed to run without figuring out where the global dictionary is defined. It will also be decoupled from pytest and be runnable by different IDEs.
     - Cons - Repetitive import invocations, often within the same file.

We have (3) currently, and I am leaning towards (2) to at least put the most common imports in a global namespace like torch, pyro, pyro.distributions. Let me know your thoughts.

### DIFF
--- a/pyro/contrib/gp/models/model.py
+++ b/pyro/contrib/gp/models/model.py
@@ -36,9 +36,17 @@ class GPModel(Parameterized):
     corresponding methods from :class:`~pyro.contrib.gp.util.Parameterized`. A typical
     way to define a Gaussian Process model is
 
+        >>> import pyro
+        >>> import pyro.contrib.gp as gp
+        >>> from pyro.infer.mcmc import HMC, MCMC
+        >>> import pyro.distributions as dist
+        >>> from pyro.params import param_with_module_name
+
         >>> X = torch.tensor([[1., 5, 3], [4, 3, 7]])
         >>> y = torch.tensor([2., 1])
         >>> kernel = gp.kernels.RBF(input_dim=3)
+        >>> kernel.set_prior("variance", dist.Uniform(torch.tensor(0.5), torch.tensor(1.5)))
+        >>> kernel.set_prior("lengthscale", dist.Uniform(torch.tensor(1.0), torch.tensor(3.0)))
         >>> gpr = gp.models.GPRegression(X, y, kernel)
 
     There are two ways to train a Gaussian Process model:
@@ -60,7 +68,7 @@ class GPModel(Parameterized):
         >>> optimizer = pyro.optim.Adam({"lr": 0.01})
         >>> svi = SVI(gpr.model, gpr.guide, optimizer, loss=Trace_ELBO())
         >>> for i in range(1000):
-        ...     svi.step()
+        ...     svi.step()  # doctest: +SKIP
 
     To give a prediction on new dataset, simply use :meth:`forward` like any PyTorch
     :class:`torch.nn.Module`:
@@ -136,20 +144,31 @@ class GPModel(Parameterized):
 
         + Batch training on a sparse variational model:
 
+            >>> import pyro
+            >>> import pyro.contrib.gp as gp
+            >>> import pyro.distributions as dist
+
+            >>> X = torch.tensor([[1., 5, 3], [4, 3, 7]])
+            >>> y = torch.tensor([2., 1])
+            >>> kernel = gp.kernels.RBF(input_dim=3)
+            >>> kernel.set_prior("variance", dist.Uniform(torch.tensor(0.5), torch.tensor(1.5)))
+            >>> kernel.set_prior("lengthscale", dist.Uniform(torch.tensor(1.0), torch.tensor(3.0)))
+            >>> optimizer = pyro.optim.Adam({"lr": 0.01})
+
             >>> Xu = torch.tensor([[1., 0, 2]])  # inducing input
             >>> likelihood = gp.likelihoods.Gaussian()
-            >>> vsgp = gp.models.SparseVariationalGP(X, y, kernel, Xu, likelihood)
+            >>> vsgp = gp.models.VariationalSparseGP(X, y, kernel, Xu, likelihood)
             >>> svi = SVI(vsgp.model, vsgp.guide, optimizer, Trace_ELBO())
             >>> batched_X, batched_y = X.split(split_size=10), y.split(split_size=10)
             >>> for Xi, yi in zip(batched_X, batched_y):
             ...     vsgp.set_data(Xi, yi)
-            ...     svi.step()
+            ...     svi.step()  # doctest: +SKIP
 
         + Making a two-layer Gaussian Process stochastic function:
 
-            >>> gpr1 = gp.models.GPRegression(X, None, kernel1, name="GPR1")
-            >>> Z, _ = gp_layer1.model()
-            >>> gpr2 = gp.models.GPRegression(Z, y, kernel2, name="GPR2")
+            >>> gpr1 = gp.models.GPRegression(X, None, kernel1, name="GPR1")  # doctest: +SKIP
+            >>> Z, _ = gp_layer1.model()                                      # doctest: +SKIP
+            >>> gpr2 = gp.models.GPRegression(Z, y, kernel2, name="GPR2")     # doctest: +SKIP
             >>> def two_layer_model():
             ...     Z, _ = gpr1.model()
             ...     gpr2.set_data(Z, y)
--- a/pyro/contrib/gp/models/model.py
+++ b/pyro/contrib/gp/models/model.py
@@ -41,6 +41,7 @@ class GPModel(Parameterized):
         >>> from pyro.infer.mcmc import HMC, MCMC
         >>> import pyro.distributions as dist
         >>> from pyro.params import param_with_module_name
+        >>> import torch
 
         >>> X = torch.tensor([[1., 5, 3], [4, 3, 7]])
         >>> y = torch.tensor([2., 1])
@@ -147,6 +148,7 @@ class GPModel(Parameterized):
             >>> import pyro
             >>> import pyro.contrib.gp as gp
             >>> import pyro.distributions as dist
+            >>> import torch
 
             >>> X = torch.tensor([[1., 5, 3], [4, 3, 7]])
             >>> y = torch.tensor([2., 1])
--- a/pyro/infer/mcmc/nuts.py
+++ b/pyro/infer/mcmc/nuts.py
@@ -9,7 +9,7 @@ import pyro.distributions as dist
 from pyro.ops.integrator import single_step_velocity_verlet
 from pyro.util import torch_isnan
 
-from .hmc import HMC
+from pyro.infer.mcmc.hmc import HMC
 
 # sum_accept_probs and num_proposals are used to calculate
 # the statistic accept_prob for Dual Averaging scheme;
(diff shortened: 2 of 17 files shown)


fehiepsi:
I think you can define some default namespace (for example: imports) using https://docs.pytest.org/en/latest/doctest.html#the-doctest-namespace-fixture

neerajprad:
> I think you can define some default namespace (for example: imports) using https://docs.pytest.org/en/latest/doctest.html#the-doctest-namespace-fixture

Good suggestion - I added that above.

jpchen:
i havent thought this through much but why is this needed? arent code snippets used to demo a certain component of the code that may or may not be explicitly defined? why do docstrings need to be executable? even if you deal with the import issues as you address above, a lot of upstream dependent code would be verbose to specify, and wouldnt allow syntax like:
```
SVI(model, guide, ...)
```
or the [code snippet here](https://pytorch.org/docs/stable/optim.html?highlight=optimizer#torch.optim.lr_scheduler.ReduceLROnPlateau)
without my defining model, guide, and all the other args and functions irrelevant to the specific function being documented. imo it's much more verbose, less readable, so unless there's a compelling reason we need this, i dont think this is the best plan of action

pyro/contrib/gp/models/model.py
```diff
@@ -136,20 +145,32 @@ def set_data(self, X, y=None):
...
+            ...     svi.step()  # doctest: +SKIP
 
         + Making a two-layer Gaussian Process stochastic function:
 
-            >>> gpr1 = gp.models.GPRegression(X, None, kernel1, name="GPR1")
-            >>> Z, _ = gp_layer1.model()
-            >>> gpr2 = gp.models.GPRegression(Z, y, kernel2, name="GPR2")
+            >>> gpr1 = gp.models.GPRegression(X, None, kernel1, name="GPR1")  # doctest: +SKIP
+            >>> Z, _ = gp_layer1.model()                                      # doctest: +SKIP
+            >>> gpr2 = gp.models.GPRegression(Z, y, kernel2, name="GPR2")     # doctest: +SKIP
```
jpchen: would these skip comments show up throughout the docs? is there a way to ignore them?

neerajprad:
> arent code snippets used to demo a certain component of the code that may or may not be explicitly defined?

I don't think the general template for writing docstring snippets should be to make them not runnable. If we can put in a small amount of effort and make them testable, we probably should. If you don't need the code to be testable, you can just not include the `>>>` marker and put them inside sphinx code block (look at how the HMC docstring tests were previously), and doctest wouldn't test it. 

>  imo it's much more verbose, less readable, so unless there's a compelling reason we need this

The only reason for this or for the existence of doctest in general is to keep your docstring examples up to date. e.g. #1114 or this has many examples of bugs or docstrings getting out of date. You can take a look at it and see if it is compelling.

pyro/contrib/gp/models/model.py
neerajprad: > would these skip comments show up throughout the docs?

They wouldn't show up in the docs, don't worry about it.

> Ah, just wonder why we skip them?

`gplayer1` and a couple of other variables aren't defined, but let me know if there's an easy fix and I'll put it in.

neerajprad:
>  most of it should come for free as a docstring is associated with each function/class, so if you modify that function, you just need to remember to modify its docstring. 

I think that would be ideal but it is very easy to miss updating a docstring, and not catching all the places where you needed to make a change. e.g. we still had some `observe` statements in the code snippet, we had forgotten to rename a class, and in #1114 I had to actually copy paste the code to see which part wasn't working. Even then, I missed out a colon when I updated the docstring.

> the other question is even if it is testable, if we want the overhead of making it so - ie the imports and defining all the upstream variables and dependencies. i think doing so makes the code more convoluted and the purpose of docs isnt to be runnable, but to convey usage, as directly as possible. 

That is a good reason against doctest, but I believe that all the negative stuff you think doctest would make us inherit is cosmetic, and there are ways around that. `import` statements can be centralized as mentioned above, and the few places where you need to define your data can be hidden away or not highlighted by sphinx. There are just a handful of places where we needed to define the data. So my thought is that this is a net gain - the test is quite fast, and we are flexible to not have to use it. Also, there are users who just copy paste from the docs to get the first example going, so they definitely will not be complaining about more detail to examples.

Those are just my thoughts - I will be happy to remove the doctest and only keep the corrections to the docstring, if we feel that it is cumbersome.

neerajprad:
> would it make sense to have a few auto imports?

Yup, that would be option (2) above. I'll make that change, so as to remove the imports. Another option would be have sphinx hide them or collapse them.

fritzo:
> ...discussion @jpchen @neerajprad 

I'd like to use doctests in a few select places, and I agree with @jpchen that we should use these with discretion and avoid overwhelming succinct docstrings with boilerplate test code. It seems reasonable to me to enable doctests in a few docstrings for now and gradually test more docstrings as we learn tricks to balance readability with testability.

pyro/poutine/block_messenger.py
```diff
@@ -23,6 +23,14 @@ class BlockMessenger(Messenger):
     Then any poutine outside of BlockMessenger(fn, hide=["a"])
     will not be applied to site "a" and will only see site "b":
 
+    >>> import pyro
+    >>> import pyro.distributions as dist
+    >>> from pyro.poutine.trace_messenger import TraceMessenger
```
fritzo: How is it that you don't need to import `BlockMessenger`? Does doctest automatically import the class/function being documented?

pyro/infer/mcmc/nuts.py
```diff
@@ -53,22 +53,25 @@ class NUTS(HMC):
...
+
+        >>> def model(data):
+        ...     coefs_mean = torch.zeros(dim)
+        ...     coefs = pyro.sample('beta', dist.Normal(coefs_mean, torch.ones(3)))
+        ...     y = pyro.sample('y', dist.Bernoulli(logits=(coefs * data).sum(-1)), obs=labels)
+        ...     return y
+
+        >>> nuts_kernel = NUTS(model, adapt_step_size=True)
+        >>> mcmc_run = MCMC(nuts_kernel, num_samples=500, warmup_steps=300).run(data)
+        >>> posterior = EmpiricalMarginal(mcmc_run, 'beta')  # doctest: +SKIP
```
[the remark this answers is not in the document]
neerajprad: This can be removed. I wanted to test if its possible to skip some steps from execution itself so that we can save time by skipping expensive steps, but doctest runs it even though it doesn't complain if any errors arise.

pyro/poutine/block_messenger.py
neerajprad: Yes, that would be within the scope. Also any file imports will be in the docstring's global scope, and won't need to be imported.

neerajprad:
> I'd like to use doctests in a few select places, and I agree with @jpchen that we should use these with discretion and avoid polluting docstrings with boilerplate testing code. 

I think the boilerplate code will be minimal once we remove the imports. I will close this PR and open a new one with only a few doctests enabled e.g. HMC, and handlers, and incorporating all review comments from here. If it looks good, I'll leave it at the discretion of the docstring authors whether they would like to use doctest. Does that sound reasonable?

pyro/infer/mcmc/nuts.py
```diff
@@ -53,22 +53,25 @@ class NUTS(HMC):
...
-            coefs = pyro.sample('beta', dist.Normal(coefs_mean, torch.ones(3)))
-            y = pyro.sample('y', dist.Bernoulli(logits=(coefs * data).sum(-1)), obs=labels)
-            return y
-
-        nuts_kernel = NUTS(model, adapt_step_size=True)
-        mcmc_run = MCMC(nuts_kernel, num_samples=500, warmup_steps=300).run(data)
-        posterior = EmpiricalMarginal(mcmc_run, 'beta')
-        print(posterior.mean)
+        >>> from pyro.infer import EmpiricalMarginal
+        >>> from pyro.infer.mcmc import MCMC
```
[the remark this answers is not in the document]
neerajprad: I will expose them anyhow in the imports namespace so that they don't need to be imported here.

fritzo:
> open a new one with only a few doctests enabled

This PR looks reasonable to me as-is, as long as we can reduce the `import` boilerplate. I'd just like to avoid enabling doctests for every code snippet everywhere. @jpchen WDYT?

neerajprad:
> I'd just like to avoid enabling doctests for every code snippet everywhere.

This PR would make it the default. Let me do it for a few files/modules, and you guys can take a look at the generated docs to see if that looks reasonable. We can include more files to the whitelist on a need-be basis.

jpchen:
im outvoted so proceed as you see fit :D , though yes eliminating boilerplate, and even commonly used variables would help readability a lot (i wonder if its possible to hide them, or predefine things like `loc`, `scale` so we can use them without having `torch.tensor(1)` everywhere. it wouldnt allow copy pasting snippets out of the box, but it would improve readability, and virtually all of our uesrs will be modifying those variables to input their own anyway)

fritzo:
> This PR would make it the default

Could we make it the default (in this or in a follow-up PR) and use `>>>` to toggle whether doctests are run? This seems nice and convenient to me (though the first diff will be gross).

neerajprad:
Thanks for your comments, @jpchen and @fritzo. 

> eliminating even commonly used variables would help readability a lot 

I'm sure there is a way to hide this boilerplate in sphinx (like the details tag in Github) if needed. Anyways to begin with I'll only incorporate docstrings where this doesn't require a lot of state to set up so that readability isn't compromised, and put all imports into a separate namespace.

> Could we make it the default (in this or in a follow-up PR) and use >>> to toggle whether doctests are run? 

I'm fine with this not being the default as well, i.e. snippets that need testing need to be specially declared or files whitelisted. If we make it the default, I suppose we need a way to blacklist code snippets and changing `>>>` to `>>` seems like the least invasive way to do it out of the box. Will look into it more.

neerajprad:
> If you can confirm that this suffices, we could add it to CONTRIBUTING.md

That should suffice, and doctest will ignore all those code snippets.

pyro/contrib/gp/models/model.py
fehiepsi: @neerajprad My bad, it is a mismatch of names. `gpr1` should be `gp_layer1` and `gpr2` should be `gp_layer2`. Would you help me fix it in your next pull request? Thanks in advance!

PR #1143, opened 2018-05-18
autoguide error on model with no latents

jpchen:
exactly what it sounds like.


pyro/contrib/autoguide/__init__.py
```diff
@@ -262,6 +262,8 @@ def _setup_prototype(self, *args, **kwargs):
             self._cond_indep_stacks[name] = site["cond_indep_stack"]
 
         self.latent_dim = sum(_product(shape) for shape in self._unconstrained_shapes.values())
+        if self.latent_dim == 0:
+            raise ValueError('Latent dims was 0; Use an empty guide instead.')
```
fritzo: Can you add the class name, e.g.
```py
'{} found latent_dim == 0; Use an empty guide instead'.format(type(self).__name__)
```
I ask because I just encountered this error half an hour ago, and it would have helped me to know which autoguide class had encountered the error.

jpchen: ok. also it probably should be a runtime error ?

jpchen:
i tested this manually for a different project, but added unit test

PR #1144, opened 2018-05-18
Add custom autoguide prefixes to facilitate composition

fritzo:
This adds a `prefix="auto"` arg to autoguides so that e.g. multiple `AutoMultivariateNormal` guides can be combined on different parts of a single model. This pattern follows @fehiepsi's pattern in pyro.contrib.gp, where kernels can take an optional name argument.

This PR also adds a check to the `ParamStore` to ensure that existing param shapes match init_tensor arguments. This issue came up in autoguides when I was using two autoguides with conflicting param names, but one was broadcastable to the other and the error was silently ignored.

This PR is backwards-compatible since `prefix` defaults to the current `"auto"`.

## Tested

- Added a test wher two guides of the same type are combined by giving them different names; this was impossible to do prior to this PR.

### DIFF
--- a/pyro/contrib/autoguide/__init__.py
+++ b/pyro/contrib/autoguide/__init__.py
@@ -63,10 +63,14 @@ class AutoGuide(object):
 
     Auto guides can be used individually or combined in an
     :class:`AutoGuideList` object.
+
+    :param callable model: A pyro model.
+    :param str name: A name that will be prefixed to all param internal sites.
     """
-    def __init__(self, model):
+    def __init__(self, model, name="auto"):
         self.master = None
         self.model = model
+        self.name = name
         self.prototype_trace = None
         self._iaranges = {}
 
@@ -123,8 +127,8 @@ class AutoGuideList(AutoGuide):
 
     :param callable model: a Pyro model
     """
-    def __init__(self, model):
-        super(AutoGuideList, self).__init__(model)
+    def __init__(self, model, name="auto"):
+        super(AutoGuideList, self).__init__(model, name)
         self.parts = []
         self.iaranges = {}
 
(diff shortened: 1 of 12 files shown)


pyro/contrib/autoguide/__init__.py
```diff
@@ -63,10 +63,14 @@ class AutoGuide(object):
...
     :class:`AutoGuideList` object.
+
+    :param callable model: A pyro model.
+    :param str name: A name that will be prefixed to all param internal sites.
     """
-    def __init__(self, model):
+    def __init__(self, model, name="auto"):
         self.master = None
         self.model = model
+        self.name = name
```
jpchen: nit: i prefer a name like `prefix` instead of `name` since this is prepended to the param names

pyro/contrib/autoguide/__init__.py
```diff
@@ -455,28 +465,31 @@ class AutoLowRankMultivariateNormal(AutoContinuous):
     :param callable model: a generative model
     :param int rank: the rank of the low-rank part of the covariance matrix
     """
-    def __init__(self, model, rank=1):
+    def __init__(self, model, name="auto", rank=1):
```
jpchen: add this to the docstring above?

pyro/contrib/autoguide/__init__.py
```diff
@@ -539,7 +552,7 @@ def __call__(self, *args, **kwargs):
         for site, Dist, param_spec in self._discrete_sites:
             name = site["name"]
             dist_params = {
-                param_name: pyro.param("auto_discrete_{}_{}".format(name, param_name), param_init,
+                param_name: pyro.param("{}_{}_{}".format(self.name, name, param_name), param_init,
```
jpchen: `auto_discrete` will now default to `auto`; is that intended

fritzo: Yes, that is intended.

PR #1145, opened 2018-05-18
add sparse gamma deep exponential family example

martinjankowiak:
no longer blocked, thanks @eb8680


examples/sparse_gamma_def.py
```diff
@@ -0,0 +1,157 @@
...
+        self.sigma_init = 0.1
+        self.softplus = nn.Softplus()
+
+    # define the model
+    def model(self, x):
+        x_size = x.size(0)
+        with pyro.iarange('data'):
+            with pyro.iarange('locals'):
+                z_top = pyro.sample("z_top", Gamma(self.alpha_z.expand(x_size, self.top_width),
+                                                   self.alpha_z.expand(x_size, self.top_width)).independent(1))
```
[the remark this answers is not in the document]
fritzo: As pair coded with @martinjankowiak offline:
```py
@poutine.broadcast
def model(self, x):
    x_size = x.size(0)
    with pyro.iarange("w_top_iarange", self.top_width * self.mid_width):
        w_top = pyro.sample("w_top", Gamma(self.alpha_w, self.beta_w))
        assert w_top.shape == (self.top_width * self.mid_width,)
    ...similar for w_mid, w_bot...
    with pyro.iarange("data", x_size):
        z_top = pyro.sample("z_top", Gamma(self.alpha_z, self.alpha_z).expand([self.top_width]).independent(1))
        assert z_top.shape == (x_size, self.top_width)
        mean_mid = torch.mm(z_top, w_top.reshape(self.top_width, self.mid_width))
        z_mid = pyro.sample("z_mid", Gamma(self.alpha_z, self.alpha_z / mean_mid).independent(1))
        assert z_mid.shape == (x_size, self.mid_width)
        ...similar for z_bot...
```

examples/sparse_gamma_def.py
```diff
@@ -0,0 +1,153 @@
...
+import os
+
+import numpy as np
+import torch
+from torch import nn as nn
+
+import pyro
+import pyro.optim as optim
+import pyro.poutine as poutine
+import wget
```
jpchen: can you add this to the extra_requires in setup.py

examples/sparse_gamma_def.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+from __future__ import absolute_import, division, print_function
+
+import argparse
+import os
+
+import numpy as np
+import torch
+from torch import nn as nn
```
jpchen: why do you need `as nn`?

examples/sparse_gamma_def.py
```diff
@@ -0,0 +1,153 @@
...
+        # define hyparameters that control the prior
+        self.alpha_z = torch.tensor(0.1)
+        self.beta_z = torch.tensor(0.1)
+        self.alpha_w = torch.tensor(0.1)
+        self.beta_w = torch.tensor(0.3)
+        # define parameters used to initialize variational parameters
+        self.alpha_init = 0.5
+        self.mean_init = 0.0
+        self.sigma_init = 0.1
+        self.softplus = nn.Softplus()
```
jpchen: nit: if youre only using nn for softplus, theres a `torch.softplus` now which (i believe) is the same thing

martinjankowiak: uh thanks

examples/sparse_gamma_def.py
martinjankowiak: ok

examples/sparse_gamma_def.py
martinjankowiak: ok

examples/sparse_gamma_def.py
```diff
@@ -0,0 +1,153 @@
...
+    # define the guide a.k.a. variational distribution.
+    # (note the guide is mean field)
+    @poutine.broadcast
+    def guide(self, x):
+        x_size = x.size(0)
+
+        # define a helper function to sample z's for a single layer
+        def sample_zs(name, width):
+            alpha_z_q = pyro.param("log_alpha_z_q_%s" % name,
+                                   torch.tensor(self.alpha_init * torch.ones(x_size, width) +
```
fritzo: nit: What is the purpose of the `torch.tensor(...)` wrapper here and in three places below? I'm surprised this is needed.

Also I would recommend wrapping the tensor init arg in a `lambda: ...` to avoid calling `torch.randn` on later iterations.

martinjankowiak: well don't i need it so that the param is a proper leaf node?

fritzo: I believe as of #908 `pyro.param` automatically creates a leaf.

martinjankowiak: ok, done

examples/sparse_gamma_def.py
```diff
@@ -0,0 +1,152 @@
...
+            # observe the data using a poisson likelihood
+            pyro.sample('obs', Poisson(mean_obs).independent(1), obs=x)
+
+    # define the guide a.k.a. variational distribution.
+    # (note the guide is mean field)
+    @poutine.broadcast
+    def guide(self, x):
+        x_size = x.size(0)
+
+        # helper for initiqlizing variational parameters
```
fritzo: nit: initializing

Issue #1146, opened 2018-05-18
Set up Pyro test crons


neerajprad:
The perf cron job is currently disabled due to the high amount of noise in the VMs. Let us re-enable a cron on a workstation for perf testing. We should run some small tests in travis to make sure that perf testing suite does not get outdated.
 - [ ] Update the perf test suite.
 - [ ] Run tests on PRs so that the suite does not go outdated.
 - [ ] Add crons to run this test daily, and send email alerts. 

Regular cuda testing. 
 - [ ] Change to using FloatTensor for cuda testing, so that we do not get OOM errors (also reduce parallel threads if needed).
 - [ ] Set up a daily cron for cuda testing.

PR #1147, opened 2018-05-18
Minor updates to perf tests

neerajprad:
Our test cron on travis is currently disabled, due to the high amount of noise and failed builds that we were getting. I am in the process of setting up test crons (#1146), and noticed that our perf tests were outdated. 

This includes some minor fixes, and also runs the suite on PRs by disabling the benchmark so that the tests are not replicated, and should be quite fast. This is to prevent the perf test suite from getting outdated.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -8,6 +8,9 @@ import pytest
 
 from tests.common import EXAMPLES_DIR, requires_cuda
 
+pytestmark = pytest.mark.stage('test_examples')
+
+
 CPU_EXAMPLES = [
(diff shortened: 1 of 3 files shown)


tests/test_examples.py
```diff
@@ -8,6 +8,9 @@
 
 from tests.common import EXAMPLES_DIR, requires_cuda
 
+pytestmark = pytest.mark.stage('test_examples')
```
neerajprad: Adding a common marker so that `test_coverage` is also included; otherwise it would run for other test stages too.

PR #1148, opened 2018-05-18
Support decorating methods with handlers

eb8680:
Currently, decorating methods with handlers like `poutine.broadcast` fails because `Messenger.__call__` returns a callable `Handler` object, not a function.  This PR removes `Handler` objects in favor of returning a function to support that use case.

cc @martinjankowiak

### DIFF
--- a/tests/poutine/test_poutines.py
+++ b/tests/poutine/test_poutines.py
@@ -753,3 +753,22 @@ def test_decorator_interface_do():
         else:
             assert name not in tr
             assert_equal(tr.nodes["_RETURN"]["value"][name], data[name])
+
+
+def test_method_decorator_interface_condition():
+
+    class cls_model(object):
+
+        @poutine.condition(data={"b": torch.tensor(1.)})
+        def model(self, p):
+            pyro.sample("a", Bernoulli(p))
+            pyro.sample("b", Bernoulli(torch.tensor([0.5])))
+
+        def _model(self, p):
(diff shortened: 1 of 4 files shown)


tests/poutine/test_poutines.py
```diff
@@ -753,3 +753,22 @@ def model():
...
+def test_method_decorator_interface_condition():
+
+    class cls_model(object):
+
+        @poutine.condition(data={"b": torch.tensor(1.)})
+        def model(self, p):
+            pyro.sample("a", Bernoulli(p))
+            pyro.sample("b", Bernoulli(torch.tensor([0.5])))
+
+        def _model(self, p):
```
fritzo: What is `._model()` used for?

fritzo:
Consider using `ContextDecorator`, available in contextlib 3.5 and contextlib2 2.7
- https://docs.python.org/3/library/contextlib.html#contextlib.ContextDecorator
- https://contextlib2.readthedocs.io/en/stable/#contextlib2.ContextDecorator

eb8680:
> Consider using ContextDecorator, available in contextlib 3.5 and contextlib2 2.7

I don't like that part of the standard library, I think it's too opaque for the tiny amount of functionality it provides.  Also, I'm setting an attribute on the return value of `Messenger.__call__`, which means I'd have to redefine `Messenger.__call__` anyway, and `TraceMessenger.__call__` has to be defined from scratch regardless.

PR #1149, opened 2018-05-18
Adding doctest for testing Pyro docstring code

neerajprad:
This resurrects #1142, with the following changes:

 - Moves all imports to a separate `doctest_namespace` which is imported by pytest before running doctest.
 - Hides all data / setting up code in a `doctest :hide:` directive, so that this is only used by doctest and does not appear in the sphinx generated documentation.

I was initially planning to only do this for a couple of docstrings, but given that we can hide all setup code, I didn't feel that any of the changes introduced are compromising readability. Let me know if we should remove doctest from any of these, and I'll just move them to a literal block instead.

### DIFF
--- a/pyro/contrib/gp/models/gplvm.py
+++ b/pyro/contrib/gp/models/gplvm.py
@@ -27,19 +27,29 @@ class GPLVM(Parameterized):
     its posterior by a multivariate normal distribution with two variational
     parameters: ``X_loc`` and ``X_scale_tril``.
 
-    For example, we can do dimensional reduction on Iris dataset as follows::
+    For example, we can do dimensional reduction on Iris dataset as follows:
 
-        # With y is the 2D Iris data of shape 150x4 and we want to reduce its dimension
-        # to a tensor X of shape 150x2, we will use GPLVM.
-        # First, define the initial values for X_loc parameter:
+        >>> # With y as the 2D Iris data of shape 150x4 and we want to reduce its dimension
+        >>> # to a tensor X of shape 150x2, we will use GPLVM.
+
+        .. doctest::
+           :hide:
+
+            >>> # Simulating iris data.
+            >>> y = torch.stack([dist.Normal(4.8, 0.1).sample((150,)),
+            ...                 dist.Normal(3.2, 0.3).sample((150,)),
+            ...                 dist.Normal(1.5, 0.4).sample((150,)),
+            ...                 dist.Exponential(0.5).sample((150,))])
+
+        >>> # First, define the initial values for X_loc parameter:
         >>> X_loc = torch.zeros(150, 2)
-        # Then, define a Gaussian Process model with input X_loc and output y:
+        >>> # Then, define a Gaussian Process model with input X_loc and output y:
         >>> kernel = gp.kernels.RBF(input_dim=2, lengthscale=torch.ones(2))
         >>> Xu = torch.zeros(20, 2)  # initial inducing inputs of sparse model
         >>> gpmodel = gp.models.SparseGPRegression(X_loc, y, kernel, Xu)
-        # Finally, wrap gpmodel by GPLVM, optimize, and get the "learned" mean of X:
+        >>> # Finally, wrap gpmodel by GPLVM, optimize, and get the "learned" mean of X:
         >>> gplvm = gp.models.GPLVM(gpmodel)
-        >>> gplvm.optimize()
+        >>> gplvm.optimize()  # doctest: +SKIP
         >>> X = gplvm.get_param("X_loc")
 
     Reference:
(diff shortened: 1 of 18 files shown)


neerajprad:
@jpchen, @fritzo - does this seem reasonable? the change to the sphinx generated documentation is minimal (in fact its mostly just bug fixes).

pyro/contrib/gp/models/gplvm.py
```diff
@@ -27,19 +27,29 @@ class GPLVM(Parameterized):
...
-        # With y is the 2D Iris data of shape 150x4 and we want to reduce its dimension
-        # to a tensor X of shape 150x2, we will use GPLVM.
-        # First, define the initial values for X_loc parameter:
+        >>> # With y as the 2D Iris data of shape 150x4 and we want to reduce its dimension
+        >>> # to a tensor X of shape 150x2, we will use GPLVM.
+
+        .. doctest::
+           :hide:
+
+            >>> # Simulating iris data.
```
neerajprad: This won't be rendered by the sphinx docs.

pyro/contrib/gp/models/model.py
```diff
@@ -136,20 +138,30 @@ def set_data(self, X, y=None):
...
+            >>> vsgp = gp.models.VariationalSparseGP(X, y, kernel, Xu, likelihood)
             >>> svi = SVI(vsgp.model, vsgp.guide, optimizer, Trace_ELBO())
             >>> batched_X, batched_y = X.split(split_size=10), y.split(split_size=10)
             >>> for Xi, yi in zip(batched_X, batched_y):
             ...     vsgp.set_data(Xi, yi)
-            ...     svi.step()
+            ...     svi.step()  # doctest: +SKIP
 
-        + Making a two-layer Gaussian Process stochastic function:
+        + Making a two-layer Gaussian Process stochastic function::
```
fritzo: Can you explain the `:` vs `::` rule you're following? This has always confused me, and I'd like to learn the right way.

pyro/contrib/named.py
```diff
@@ -38,8 +38,8 @@
     >>> state = named.Object("state")
     >>> loc = state.loc.param_(torch.zeros(1, requires_grad=True))
     >>> scale = state.scale.param_(torch.ones(1, requires_grad=True))
-    >>> z = state.z.sample_(dist.normal, loc, scale)
-    >>> state.x.observe_(dist.normal, z, loc, scale)
+    >>> z = state.z.sample_(dist.Normal(loc, scale))
```
fritzo: Thanks for fixing this, and glad we have doctests to find this kind of error!

pyro/poutine/handlers.py
```diff
@@ -39,11 +39,11 @@
     >>> traced_model = poutine.trace(conditioned_model)
 
 Many inference algorithms or algorithmic components can be implemented
-in just a few lines of code:
+in just a few lines of code::
 
-    >>> guide_tr = poutine.trace(guide).get_trace(...)
-    >>> model_tr = poutine.trace(poutine.replay(conditioned_model, trace=tr)).get_trace(...)
-    >>> monte_carlo_elbo = model_tr.log_prob_sum() - guide_tr.log_prob_sum()
+    guide_tr = poutine.trace(guide).get_trace(...)
```
neerajprad: Putting this into a literal block without `>>>` so as to avoid running doctest on this snippet.

neerajprad:
Looks good so far.

Do I understand correctly that doctests are now enabled by default on all >>> blocks, and there is no whitelist or blacklist?

Yes, it is enabled by default on all >>> blocks.

pyro/contrib/gp/models/model.py
neerajprad: I think `:` will have no effect, and will just appear as is, so the question is whether to use `::` (which also appears as a regular `:`) or not.

 - If we have an interactive (>>>) block, we do not need `::`. If we put `::`, it will treat the `.. doctest` directive as part of the literal block, which we do not want. So with the interactive prompt, its always safer to use `:`.
 - If we don't have an interactive code block (no `>>>`), we need to use `::`, which will treat the code block as a block of python code, by default.

neerajprad: Let me enforce some consistency in these code blocks actually.

pyro/contrib/gp/models/model.py
```diff
@@ -134,22 +136,33 @@ def set_data(self, X, y=None):
 
         Some examples to utilize this method are:
 
+        .. doctest::
+           :hide:
```
jpchen: does this mean everything in this indentation block is collapsible or not rendered?

neerajprad: @jpchen - This won't be rendered by sphinx.

neerajprad:
> edit: how come import torch is needed in gp kernel?

I probably missed that; will remove.

neerajprad:
This should be good to merge, pending any further comments.

PR #1150, opened 2018-05-18
fix requires_grad error

yangky11:
Running the following code leads to a runtime error (pytorch 0.4.0, pyro 0.2.0), which is probably related to https://github.com/pytorch/pytorch/issues/3890. 
This pull request fixes the error.

```
import torch
import pyro
a = torch.rand(3, 4, requires_grad=True)
b = a + 1
pyro.param('b', b)
```

```
<ipython-input-5-270f067febca> in <module>()
----> 1 pyro.param('b', b)

/Users/yangky/anaconda3/lib/python3.5/site-packages/pyro/primitives.py in param(name, *args, **kwargs)
    314     """
    315     if not am_i_wrapped():
--> 316         return _PYRO_PARAM_STORE.get_param(name, *args, **kwargs)
    317     else:
    318         msg = {

/Users/yangky/anaconda3/lib/python3.5/site-packages/pyro/params/param_store.py in get_param(self, name, init_tensor, constraint)
    100             with torch.no_grad():
    101                 unconstrained_param = transform_to(constraint).inv(init_tensor)
--> 102             unconstrained_param.requires_grad = True
    103             self._params[name] = unconstrained_param
    104             self._constraints[name] = constraint

RuntimeError: you can only change requires_grad flags of leaf variables.
```


yangky11:
@eb8680 Thanks. I was trying to share a variable between the `model()` and the `guide()` and I later figured out `pyro.param` is not the correct way to use. The variable I want to share is a hidden state of LSTM, which is not a random variable. Currently, I just make this variable an attribute of the `nn.Module` on which `model()` and `guide()` are defined, so that it can be accessed inside both functions. Is this correct?

eb8680:
> Currently, I just make this variable an attribute of the nn.Module on which model() and guide() are defined, so that it can be accessed inside both functions. Is this correct?

Yes, that should be fine.

martinjankowiak:
do we need any parallel changes in any of these places?
```
pyro/ops/integrator.py:        node.requires_grad = True
pyro/ops/integrator.py:        node.requires_grad = False
pyro/ops/newton.py:            x.requires_grad = True  # ensure loss is differentiable wrt x
tests/distributions/test_cuda.py:                val.requires_grad = True
tests/distributions/test_cuda.py:                val.requires_grad = True
tests/ops/test_newton.py:        x.requires_grad = requires_grad
tests/ops/test_newton.py:        x.requires_grad = True
```

PR #1159, opened 2018-05-24
Add a crude trace-based scoping mechanism for sample names

eb8680:
~~Blocked by #1161~~ 

This PR adds a new handler `contrib.autoname.scope` that adds a prefix to a sample site name (I needed something like this for a different project):
```python
@scope(prefix="a")
def model():
    return pyro.sample("x", Normal(0., 1.))

assert "a/x" in poutine.trace(model).get_trace()
```

If no prefix is provided but a function is, `poutine.scope` will try to use the function's name as a prefix:
```python
@scope
def model():
    return pyro.sample("x", Normal(0., 1.))

assert "model/x" in poutine.trace(model).get_trace()
```

Beyond naive prepending, it supports recursion:
```python
@scope
def model(r=True):
    return model(r=False) if r else pyro.sample("x", Normal(0., 1.))

assert "model/model/x" in poutine.trace(model).get_trace()
```

Additionally, the PR includes a modification to `TraceMessenger` that allows using the trace to add a counter suffix to names that have already appeared:
```python
@scope
def model():
    pyro.sample("x", Normal(0., 1.))  # x
    pyro.sample("x", Normal(0., 1.))  # x_0
    pyro.sample("x", Normal(0., 1.))  # x_1

# the PR currently makes strict=False the new default
# set strict=True for the old behavior (raising an error on duplicate sites)
tr = poutine.trace(model, strict=False).get_trace()
assert "model/x" in tr
assert "model/x_0" in tr
assert "model/x_1" in tr
```


eb8680:
Hmm, I think I want to move this out to `contrib.named` and modify the semantics so that scopes rather than entire names get suffixes.  Will reopen or submit a clean PR.

eb8680:
Design issues open for discussion:
1. Should suffixes be added to entire names or to individual scopes?
2. Should this mechanism apply to parameters as well?

fritzo:
> Should this mechanism apply to parameters as well?

Sounds good to me. I'd like to be able to create modular "learnable functions" that automatically create param names. We're already doing this manually in `pyro.module`, `pyro.contrib.gp`, and `pyro.contrib.autoguide`. Maybe we could unify the approaches?

eb8680:
> I'd like to be able to create modular "learnable functions" that automatically create param names. We're already doing this manually in pyro.module, pyro.contrib.gp, and pyro.contrib.autoguide. Maybe we could unify the approaches?

Yeah, and that way we could probably simplify `contrib.gp` by just using `poutine.lift` to set the various priors instead of having the special `Parametrized` machinery.

eb8680:
> Should suffixes be added to entire names or to individual scopes?

Also, some additional context for this: consider a model that calls a submodel multiple times, like the following:
```python
@scope
def submodel(p):
    return pyro.sample("y", Bernoulli(p))

@scope
def model():
    p = pyro.sample("p", Beta(...))
    return submodel(p) and submodel(p)
```
Suppose I wanted to constrain both `submodel` calls with different data.  Under the current scheme, the different `y` sites have globally determined suffixes:
```python
data = {"model/submodel/y": 1, "model/submodel/y_0": 0}
constrained_model = poutine.condition(model, data=data)
```
However, we might instead prefer the `"submodel"` scope to have a suffix, so that all variables inside `submodel` have the same names in different calls modulo the enclosing scope:
```python
data = {"model/submodel/y": 1, "model/submodel_0/y": 0}
constrained_model = poutine.condition(model, data=data)
```
This example is sort of trivial, but if `submodel` and the constraint were more complicated or there were more nested scopes (both of which are true in the other thing I was doing that motivated this PR), the scope-level suffixes would be much easier to work with.  On the other hand, it's trickier to implement and also not clear where to place the suffix in recursion.

eb8680:
> Maybe you could use reusable context managers for scope-level suffixes, similar to our reusable context managers for iarange and irange. That is, a suffix could be incremented on context creation, rather than on context entry.

The problem with this is that, unlike for `iarange` composition, order matters for nested scopes.  Consider this example:
```python
@scope
def submodel():
    pyro.sample("y", ...)

def model():
    submodel()  # "submodel/y"
    submodel()  # "submodel_0/y"
    with scope(prefix="inter"):
        submodel()  # "inter/submodel/y"
    submodel()  # "submodel_1/y"
```
In particular, node that the third site is `"inter/submodel/y"`, not `"inter/submodel_1/y"`.  I added a different hack that uses the same trace-based tracking mechanism to ensure scopes get suffixes; the example above now works if you replace `@scope` with `@outer_scope`.

pyro/contrib/autoname/scoping.py
```diff
@@ -0,0 +1,82 @@
...
+            "infer": {},
+            "scale": 1.0,
+            "cond_indep_stack": (),
+            "done": False,
+            "stop": False,
+            "continuation": None,
+            "PRUNE": True
+        }
+        apply_stack(msg)
+        self.prefix = msg["name"].split("/")[-1]
```
eb8680: Here's my hack: we're basically pretending that entering this scope context is the same as a sample site, so that the same mechanism used to add suffixes to sample names can be co-opted and used to add suffixes to scopes.

fritzo:
@eb8680 let me know when you would like another review.

eb8680:
I removed scoping from parameter names, added docstrings, and added a minimal example based on `examples/contrib/autoname/mixture.py`.  It should be ready for a detailed review.

fritzo:
Hmm, I was hoping to use scoping also for parameter names, so we could have independennt differentiable units. Can you explain the design choice of treating the two names differently (happy to discuss off-line).

fritzo:
Dow we have a good reason to "makes strict=False the new default"? I often rely on the old strict behavior to catch bugs in my models, so I'd be inclined to preserve `strict=True`.

eb8680:
> Hmm, I was hoping to use scoping also for parameter names, so we could have independennt differentiable units. Can you explain the design choice of treating the two names differently

I think scoped parameter names should behave a little differently to make lookup and reuse less confusing and follow the design of Python closures.  In particular, if the parameter store can't find a parameter in a scope, it should recursively check if the name appears in enclosing scopes before creating and using a new parameter value.  This seemed like a big change for an already large PR, so I figured I could address it in a follow-up and leave parameter names global here.

> Do we have a good reason to "makes strict=False the new default"? I often rely on the old strict behavior to catch bugs in my models, so I'd be inclined to preserve strict=True.

I would argue that those shouldn't be considered bugs, and that you should use `scope` if you want to avoid collisions ;).  However, this change did raise the question of providing a mechanism for changing the backend `Messenger`s of handlers, since in many cases `trace` is being called inside an `ELBO` and we don't have direct access to its arguments.  Since we don't have such a mechanism at the moment, I can revert the change if you want.

fritzo:
> I can revert the change if you want.

How about we split the behavior change into a follow-up PR so we can discuss and clarify implications. It's always nice to factor PRs into (1) enable optional new behavior (2) toggle the behavior to be default.

eb8680:
> How about we split the behavior change into a follow-up PR so we can discuss and clarify implications?

Ok, I've reverted the change.

pyro/poutine/trace_messenger.py
```diff
@@ -41,7 +41,7 @@ class TraceMessenger(Messenger):
     We can also use this for visualization.
     """
 
-    def __init__(self, graph_type=None, param_only=None):
+    def __init__(self, graph_type=None, param_only=None, strict=None):
```
fritzo: Do you foresee any future notions of strictness you'd like to add? If so, consider renaming this kwarg to `strict_names`.

eb8680:
I'm not seeing the DMM failures?

pyro/poutine/trace_messenger.py
eb8680: done

PR #1161, opened 2018-05-24
Create contrib.autoname and move contrib.named to contrib.autoname.named

eb8680:
This PR adds a `contrib.autoname` module as a centralized location for experimental tools for generating and aligning names in Pyro.  It moves `contrib.named` to `contrib.autoname.named` and fixes all associated import and sphinx issues.  I'd like to reopen #1159 with `poutine.scope` moved to `contrib.autoname.scope`, since that code is still experimental.


pyro/contrib/autoname/__init__.py
```diff
@@ -0,0 +1,12 @@
+"""
+The :mod:`pyro.contrib.autoname` module provides tools for automatically
+generating unique, semantically meaningful names for sample sites.
+"""
+from __future__ import absolute_import, division, print_function
+
+from pyro.contrib.autoname import named
```
fritzo: This is fine, but I believe you can omit `named` from this file entirely and `from pyro.contrib.autoname import named` will still work.

eb8680:
Should be good to merge

Issue #1160, opened 2018-05-24
Support regex or filter arg in poutine.block


fritzo:
Related to #73 #1159 

It would be nice to have args `expose_regex`, `hide_regex`, `expose_fn`, and `hide_fn` in `@poutine.block`. This will be even more important once we have name scopes.

## Why?

Often site names are automatically generated e.g.
```py
x = 0.
for t in range(1, len(data)):
    x = pyro.sample('x_{}'.format(t), dist.Normal(x, 1.))
```
In these cases, blocking those sites is cumbersome
```py
poutine.block(model, hide=['x_{}'.format(t) for t in range(1, len(data))])
```
It would be nice to instead write one of
```py
poutine.block(model, hide_regex='^x_.*')
poutine.block(model, hide_fn=lambda n: n.startswith('x_'))
```

## How?

Note that `expose_regex` is trivially implemented in terms of `expose_fn`:
```py
expose_fn = re.compile(expose_regex).search  # or .match ?
```

fritzo:
Note a third option is to use unix-style filename matching using [fnmatch](https://docs.python.org/2/library/fnmatch.html#fnmatch.fnmatch). This might make more sense if we use `tf.scope`-style naming conventions.

PR #1162, opened 2018-05-24
Add hide_fn parameter to poutine.block

eb8680:
Addresses #1160 

Example:
```python
def model():
    pyro.sample("x1", dist.Bernoulli(0.5))
    pyro.sample("x2", dist.Bernoulli(0.5))
    pyro.sample("y", dist.Bernoulli(0.5))

blocked = poutine.block(model, hide_fn=lambda msg: "x" in msg["name"])
assert "x1" not in poutine.trace(blocked).get_trace(...)
assert "y" in poutine.trace(blocked).get_trace(...)
```

### DIFF
--- a/pyro/poutine/block_messenger.py
+++ b/pyro/poutine/block_messenger.py
@@ -3,6 +3,62 @@ from __future__ import absolute_import, division, print_function
 from pyro.poutine.messenger import Messenger
 
 
+def _make_default_hide_fn(hide_all, expose_all, hide, expose, hide_types, expose_types):
+    # first, some sanity checks:
+    # hide_all and expose_all intersect?
+    assert (hide_all is False and expose_all is False) or \
+        (hide_all != expose_all), "cannot hide and expose a site"
+
+    # hide and expose intersect?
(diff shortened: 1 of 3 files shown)


eb8680:
@fritzo is this what you had in mind?

eb8680:
> What do you think of adding an expose_fn for symmetry

Done

pyro/poutine/block_messenger.py
```diff
@@ -55,78 +118,22 @@ class BlockMessenger(Messenger):
...
-        # otherwise expose
-        else:
-            return False
+            self.hide_fn = _make_default_hide_fn(hide_all, expose_all,
+                                                 hide, expose,
+                                                 hide_types, expose_types)
 
     def _process_message(self, msg):
-        msg["stop"] = self._block_up(msg)
+        msg["stop"] = self.hide_fn(msg)
```
fritzo: Maybe `bool(self.hide_fn)` to be safe?

eb8680: Done

PR #1163, opened 2018-05-24
Relax broadcasting constraint on scale_tensor

fritzo:
This is needed for parallel enumeration, where e.g. the a sample site's values are the same in all enumeration branches (leading to a log_prob that is not broadcast), but a `poutine.scale` mask does depend on the enumeration values. In this case we need to broadcast up the site log_prob to handle the `scale` of larger shape.

## Tested

- added exhaustive tests
- tested on an internal application

### DIFF
--- a/pyro/distributions/util.py
+++ b/pyro/distributions/util.py
@@ -148,30 +148,34 @@ def sum_leftmost(value, dim):
 
 def scale_tensor(tensor, scale):
     """
-    Safely scale a tensor without increasing its ``.shape``.
-    This avoids NANs by assuming ``inf * 0 = 0 * inf = 0``.
+    Safely scale a tensor, avoiding NANs by assuming::
+
+        inf * 0 = 0 * inf = 0
     """
+    # Check for trivial cases.
     if isinstance(tensor, numbers.Number):
-        if isinstance(scale, numbers.Number):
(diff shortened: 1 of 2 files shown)


pyro/distributions/util.py
```diff
@@ -148,30 +148,34 @@ def sum_leftmost(value, dim):
...
-        else:
-            return tensor * scale
+
     result = tensor * scale
-    result[(scale == 0).expand_as(result)] = 0  # avoid NANs
-    if result.shape != tensor.shape:
-        raise ValueError("Broadcasting error: scale is incompatible with tensor: "
-                         "{} vs {}".format(scale.shape, tensor.shape))
+
+    # avoid NANs
```
martinjankowiak: why do you want these?

fritzo: The main reason is to allow `poutine.scale` to completely omit some senseless log prob terms by enforcing hard constraints.

fritzo: The main reason is to allow `poutine.scale` to completely omit some senseless log prob terms by enforcing hard constraints.

jpchen: to safely mask out any zero scale values when `poutine.scale`ing

PR #1165, opened 2018-05-30
[Do not merge] Infer local names of non-anonymous sample statements

eb8680:
This PR is experimenting with using [glom](http://glom.readthedocs.io/en/latest/api.html#object-oriented-access-and-method-calls-with-t) and AST rewriting to infer semantically meaningful names automatically for `pyro.sample` and `pyro.param` calls that appear on the right-hand side of an assignment statement.  I'm currently using some boilerplate AST parsing/compilation code adapted from Tangent (hence the license in the PR), but would probably remove it before actually merging this to `contrib.autoname`.

`glom` is a library for querying nested data structures.  `glom.T` is sort of like a souped-up version of the data structures in `contrib.autoname.named`.  As in `contrib.autoname.named`, I've just used the string representations of those data structures for a proof of concept, but it's possible that in the future we could replace the string-based query language for sample sites with `glom` or something like it.

Mostly posted for @fritzo to look at the test in `tests/contrib/autoname/test_glom_named.py`.

### DIFF
--- a/LICENSE.Apache-2.txt
+++ b/LICENSE.Apache-2.txt
@@ -1,0 +1,182 @@
+Applies to pyro/contrib/autoname/compilation.py, which is modified from 
+https://github.com/google/tangent/blob/master/tangent/compile.py and
+https://github.com/google/tangent/blob/master/tangent/quoting.py.
+
+-------------------------------------------------------------------------
+                              Apache License
+                        Version 2.0, January 2004
+                     http://www.apache.org/licenses/
+
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+1. Definitions.
+
+   "License" shall mean the terms and conditions for use, reproduction,
+   and distribution as defined by Sections 1 through 9 of this document.
+
+   "Licensor" shall mean the copyright owner or entity authorized by
+   the copyright owner that is granting the License.
+
+   "Legal Entity" shall mean the union of the acting entity and all
+   other entities that control, are controlled by, or are under common
+   control with that entity. For the purposes of this definition,
+   "control" means (i) the power, direct or indirect, to cause the
+   direction or management of such entity, whether by contract or
+   otherwise, or (ii) ownership of fifty percent (50%) or more of the
+   outstanding shares, or (iii) beneficial ownership of such entity.
+
+   "You" (or "Your") shall mean an individual or Legal Entity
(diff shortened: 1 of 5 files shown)


eb8680:
Will reopen later this week after getting #1159 merged and significantly updating/simplifying

tests/contrib/autoname/test_glom_named.py
```diff
@@ -0,0 +1,49 @@
...
+        i = 2
+        z[i] = pyro.sample(dist.Bernoulli(0.5))
+        zz = {}
+        zz["a"] = pyro.sample(dist.Bernoulli(0.5))
+        ab = B()
+        ab.b = pyro.sample(dist.Bernoulli(0.5))
+        zz["c"] = [B(), None, None, None]
+        zz["c"][0].b = pyro.sample(dist.Bernoulli(0.5))
+        for j in range(1, 3):
+            zz["c"][j] = pyro.sample(dist.Bernoulli(0.5))
```
fritzo: This is really nice. How smart is it? Can we
```py
for key, value in my_dict.items():
    zz["{}_{}".format(self.prefix, key)] = pyro.sample(dist.Normal(value, 1.))
```

tests/contrib/autoname/test_glom_named.py
```diff
@@ -0,0 +1,49 @@
...
+import pyro
+import pyro.distributions as dist
+import pyro.poutine as poutine
+
+from pyro.contrib.autoname import glom_name
+
+
+class B(object):
+    def __init__(self):
+        b = 3
```
fritzo: `self.b = 3`?

LICENSE.Apache-2.txt
```diff
@@ -0,0 +1,182 @@
+Applies to pyro/contrib/autoname/compilation.py, which is modified from
```
fritzo: It's common to create a `vendor/` subdirectory and embed other code there, including license.

tests/contrib/autoname/test_glom_named.py
eb8680: Yes, that would work exactly as expected.  The main limitation of this autonaming scheme is that, as currently written, it doesn't support closures due to the way the recompilation works.  For example, note that in this test, the class `B` has to be defined at the module level rather than inside the test function.  I'm guessing there's a workaround for this shortcoming, though.

PR #1166, opened 2018-05-30
Add AutoGuideList.median() method

fritzo:
This adds a `.median()` method to `AutoGuideList`, simply aggregating the `.median()`s of the components, e.g. `AutoDelta.median()` and `AutoDiagonalNormal.median()`. This makes it much easier to experiment by swapping in an `AutoGuideList` for an `AutoDiagonalNormal` while keeping the same code for latent variable inspection.

## Tested

- added an example to test_median
- tested in an internal application


fritzo:
Thanks, @jpchen !

PR #1167, opened 2018-05-31
Fix CUDA failure in advi test

neerajprad:
Fixes CUDA cron job.


fritzo:
Thanks for fixing this!

PR #1172, opened 2018-06-03
Accept callables to AutoGuideList.add(-)

### LINKED ISSUE #1168: composable guides/autoguides
Hey team,

I was looking at the autoguides a bit and have the following request.

It would be cool to be able to define guides over some parts of the model (imagine I have custom inference strategies for particular components int he model) and let autoguides take care of the rest.

This would correspond to composing guides to explain a model and would require some mechanism that would test which variables are taken care of explicitly in the custom guide and which other ones need to be automagically guided.

Currently it feels as if there is some component missing to do this, as we are not attaching scopes to autoguides, but I may also have missed some discussions of course.

One might also consider having two different autoguides scoped over different parts of the model (i.e. a neural autoguide for local variables and an advi autoguide over globals), although I have not thought this particular case through and it may have issues with building proper inverse models and conditioning on previous samples and so on.

Any thoughts on this?

With Regards,

Theo

fritzo:
Resolves #1168 

This creates a simple wrapper `AutoCallable` to allow callables to be `.add()`ed to an `AutoGuideList`.

## Tested

- smoke test with no return value
- smoke test with dict return value
- value test with optional `median` callable

### DIFF
--- a/docs/source/contrib.autoguide.rst
+++ b/docs/source/contrib.autoguide.rst
@@ -19,6 +19,14 @@ AutoGuideList
     :special-members: __call__
     :show-inheritance:
 
+AutoCallable
+------------
+.. autoclass:: pyro.contrib.autoguide.AutoCallable
+    :members:
+    :undoc-members:
+    :special-members: __call__
+    :show-inheritance:
+
 AutoDelta
 ---------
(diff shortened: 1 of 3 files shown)


fritzo:
> the class is redundant

Users need never interact with the `AutoCallable`; callable guides will be wrapped automatically.

PR #1173, opened 2018-06-04
add transformed mvn autoguide

jpchen:
the simplest of fancier autoguides:
 - fixes bugs in auto reg nn
 - implements a simple autoguide learns gaussian noise generated via iaf
 - increases num iters to pass cuda tests (#1177)

### DIFF
--- a/pyro/contrib/autoguide/__init__.py
+++ b/pyro/contrib/autoguide/__init__.py
@@ -113,6 +113,15 @@ class AutoGuide(object):
                 else:
                     raise NotImplementedError("AutoGuideList does not support pyro.irange")
 
+    def median(self, *args, **kwargs):
+        """
+        Returns the posterior median value of each latent variable.
+
+        :return: A dict mapping sample site name to median tensor.
+        :rtype: dict
+        """
+        raise NotImplementedError
+
 
 class AutoGuideList(AutoGuide):
     """
@@ -174,6 +183,18 @@ class AutoGuideList(AutoGuide):
         result = {}
         for part in self.parts:
             result.update(part(*args, **kwargs))
+        return result
+
+    def median(self, *args, **kwargs):
+        """
+        Returns the posterior median value of each latent variable.
+
+        :return: A dict mapping sample site name to median tensor.
+        :rtype: dict
+        """
+        result = {}
+        for part in self.parts:
+            result.update(part.median(*args, **kwargs))
         return result
 
 
--- a/pyro/contrib/autoguide/__init__.py
+++ b/pyro/contrib/autoguide/__init__.py
@@ -565,7 +565,7 @@ class AutoLowRankMultivariateNormal(AutoContinuous):
 class AutoTransformedNormal(AutoContinuous):
     """
     This implementation of :class:`AutoContinuous` uses a Transformed
-    Diagonal Normal distribution parameterized via an IAF to construct a guide
+    Diagonal Normal distribution via an IAF to construct a guide
     over the entire latent space. The guide does not depend on the model's
     ``*args, **kwargs``.
 
@@ -585,27 +585,24 @@ class AutoTransformedNormal(AutoContinuous):
         super(AutoTransformedNormal, self).__init__(model, prefix)
 
     def sample_latent(self, *args, **kwargs):
-        """
-        Samples the (single) multivariate normal latent used in the auto guide.
-        """
         if self.hidden_dim is None:
             self.hidden_dim = self.latent_dim
-        self.iaf = dist.InverseAutoregressiveFlow(self.latent_dim, self.hidden_dim, sigmoid_bias=self.sigmoid_bias)
-        pyro.module("{}_iaf".format(self.prefix), self.iaf.module)
-        iaf_dist = dist.TransformedDistribution(dist.Normal(0., 1.).expand([self.hidden_dim]), self.iaf)
-        loc = pyro.sample("{}_loc".format(self.prefix), iaf_dist.independent(1), infer={"is_auxiliary": True})
-        scale = pyro.param("{}_scale".format(self.prefix), lambda: torch.ones(self.latent_dim))
-        return pyro.sample("_{}_latent".format(self.prefix), dist.Normal(loc, scale).independent(1),
+        loc = pyro.param("{}_loc".format(self.prefix),
+                         lambda: torch.zeros(self.latent_dim))
+        scale = pyro.param("{}_scale".format(self.prefix),
+                           lambda: torch.ones(self.latent_dim),
+                           constraint=constraints.positive)
+        iaf = dist.InverseAutoregressiveFlow(self.latent_dim, self.hidden_dim,
+                                             sigmoid_bias=self.sigmoid_bias)
+        pyro.module("{}_iaf".format(self.prefix), iaf.module)
+        self.iaf_dist = dist.TransformedDistribution(dist.Normal(loc, scale), [iaf])
+        return pyro.sample("_{}_latent".format(self.prefix), self.iaf_dist.independent(1),
                            infer={"is_auxiliary": True})
 
     def _loc_scale(self, *args, **kwargs):
-        prototype = pyro.param("{}_scale".format(self.prefix))
-        loc = self.iaf(prototype.new_zeros(self.hidden_dim))
-        scale = None  # no analytical solution
+        loc = pyro.param("{}_loc".format(self.prefix))
+        scale = pyro.param("{}_scale".format(self.prefix))
         return loc, scale
-
-    def quantiles(self, quantiles, *args, **kwargs):
-        raise NotImplementedError('scale cannot be computed analytically')
 
 
 class AutoDiscreteParallel(AutoGuide):
--- a/pyro/contrib/autoguide/__init__.py
+++ b/pyro/contrib/autoguide/__init__.py
@@ -585,6 +585,8 @@ class AutoIAFNormal(AutoContinuous):
         super(AutoIAFNormal, self).__init__(model, prefix)
 
     def sample_latent(self, *args, **kwargs):
+        if self.latent_dim == 1:
+            raise ValueError('latent dim = 1. Consider using AutoDiagonalNormal instead')
         if self.hidden_dim is None:
             self.hidden_dim = self.latent_dim
         iaf = dist.InverseAutoregressiveFlow(self.latent_dim, self.hidden_dim,
--- a/pyro/contrib/autoguide/__init__.py
+++ b/pyro/contrib/autoguide/__init__.py
@@ -519,6 +519,53 @@ class AutoLowRankMultivariateNormal(AutoContinuous):
         return loc, scale
 
 
+class AutoTransformedNormal(AutoContinuous):
+    """
+    This implementation of :class:`AutoContinuous` uses a Transformed
+    Normal distribution via a IAF to construct a guide
+    over the entire latent space. The guide does not depend on the model's
+    ``*args, **kwargs``.
+
+    Usage::
+
+        guide = AutoTransformed(model, rank=10)
+        svi = SVI(model, guide, ...)
+
+    :param callable model: a generative model
+    :param int rank: the rank of the low-rank part of the covariance matrix
+    :param str prefix: a prefix that will be prefixed to all param internal sites
+    """
+    def __init__(self, model, prefix="auto", rank=1):
+        if not isinstance(rank, numbers.Number) or not rank > 0:
+            raise ValueError("Expected rank >= 0 but got {}".format(rank))
+        self.rank = rank
+        super(AutoLowRankMultivariateNormal, self).__init__(model, prefix)
+
+    def sample_latent(self, *args, **kwargs):
+        """
+        Samples the (single) multivariate normal latent used in the auto guide.
+        """
+        loc = pyro.param("{}_loc".format(self.prefix),
+                         lambda: torch.zeros(self.latent_dim))
+        W_term = pyro.param("{}_W_term".format(self.prefix),
+                            lambda: torch.randn(self.rank, self.latent_dim) * (0.5 / self.rank) ** 0.5)
+        D_term = pyro.param("{}_D_term".format(self.prefix),
+                            lambda: torch.ones(self.latent_dim) * 0.5,
+                            constraint=constraints.positive)
+        iaf = dist.InverseAutoregressiveFlow(10, 40)
+        pyro.module("iaf", iaf.module)
+        iaf_dist = dist.TransformedDistribution(dist.LowRankMultivariateNormal(loc, W_term, D_term), [iaf])
+        return pyro.sample("_{}_latent".format(self.prefix),
+                           af_dist,
+                           infer={"is_auxiliary": True})
+
+    def _loc_scale(self, *args, **kwargs):
+        loc = pyro.param("{}_loc".format(self.prefix))
+        W_term = pyro.param("{}_W_term".format(self.prefix))
+        D_term = pyro.param("{}_D_term".format(self.prefix))
+        return loc, scale
+
+
 class AutoDiscreteParallel(AutoGuide):
     """
     A discrete mean-field guide that learns a latent discrete distribution for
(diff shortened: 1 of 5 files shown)


pyro/contrib/autoguide/__init__.py
```diff
@@ -519,6 +520,52 @@ def _loc_scale(self, *args, **kwargs):
         return loc, scale
 
 
+class AutoTransformedNormal(AutoContinuous):
```
jpchen: open to more descriptive names that arent egregiously long

pyro/contrib/autoguide/__init__.py
```diff
@@ -519,6 +520,52 @@ def _loc_scale(self, *args, **kwargs):
...
+        Samples the (single) multivariate normal latent used in the auto guide.
+        """
+        loc = pyro.param("{}_loc".format(self.prefix),
+                         lambda: torch.zeros(self.latent_dim))
+        W_term = pyro.param("{}_W_term".format(self.prefix),
+                            lambda: torch.randn(self.rank, self.latent_dim) * (0.5 / self.rank) ** 0.5)
+        D_term = pyro.param("{}_D_term".format(self.prefix),
+                            lambda: torch.ones(self.latent_dim) * 0.5,
+                            constraint=constraints.positive)
+        iaf = dist.InverseAutoregressiveFlow(self.latent_dim, 10*self.latent_dim)
```
jpchen: `10*self.latent_dim` is a bit ad-hoc... suggestions?

pyro/contrib/autoguide/__init__.py
```diff
@@ -519,6 +520,42 @@ def _loc_scale(self, *args, **kwargs):
         return loc, scale
 
 
+class AutoTransformedNormal(AutoLowRankMultivariateNormal):
```
jpchen: open to more descriptive names that arent egregiously long

fritzo: You should not inherit from `AutoLowRankMultivariateNormal`, because that class assumes the `._loc_scale()` method produces quantiles of the models variables (not quantiles of an arbitrary latent). Instead I would recommend forking that class's code and defining custom versions of its `.sample_latent()` and `._loc_scale()` methods so that `.median()` and `.quantiles()` will be defined correctly.

tests/contrib/autoguide/test_advi.py
```diff
@@ -113,15 +118,25 @@ def model():
         infer.step()
 
     median = guide.median()
-    assert_equal(median["x"], torch.tensor(0.0), prec=0.1)
-    if auto_class is AutoDelta:
-        assert_equal(median["y"], torch.tensor(-1.0).exp(), prec=0.1)
+    if auto_class is AutoTransformedNormal:
```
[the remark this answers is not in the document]
jpchen: yes for 1e3 steps and 1e-3 lr, ill modify the hyperparams instead since it's only for this one class

pyro/contrib/autoguide/__init__.py
```diff
@@ -561,6 +562,57 @@ def _loc_scale(self, *args, **kwargs):
...
+        self.iaf_dist = dist.TransformedDistribution(dist.LowRankMultivariateNormal(loc, W_term, D_term), [iaf])
+        return pyro.sample("_{}_latent".format(self.prefix), self.iaf_dist, infer={"is_auxiliary": True})
+
+    def _loc_scale(self, *args, **kwargs):
+        loc = pyro.param("{}_loc".format(self.prefix))
+        W_term = pyro.param("{}_W_term".format(self.prefix))
+        D_term = pyro.param("{}_D_term".format(self.prefix))
+        scale = (W_term.pow(2).sum(0) + D_term).sqrt()
+        transform = biject_to(self.iaf_dist.support)
+        loc, scale = transform(loc), transform(scale)
```
fritzo: nonsense

pyro/contrib/autoguide/__init__.py
```diff
@@ -561,6 +562,57 @@ def _loc_scale(self, *args, **kwargs):
...
+                         lambda: torch.zeros(self.latent_dim))
+        W_term = pyro.param("{}_W_term".format(self.prefix),
+                            lambda: torch.randn(self.rank, self.latent_dim) * (0.5 / self.rank) ** 0.5)
+        D_term = pyro.param("{}_D_term".format(self.prefix),
+                            lambda: torch.ones(self.latent_dim) * 0.5,
+                            constraint=constraints.positive)
+        hidden_dim = self.hidden_dim if self.hidden_dim is not None else self.latent_dim
+        iaf = dist.InverseAutoregressiveFlow(self.latent_dim, hidden_dim, sigmoid_bias=self.sigmoid_bias)
+        pyro.module("{}_iaf".format(self.prefix), iaf.module)
+        self.iaf_dist = dist.TransformedDistribution(dist.LowRankMultivariateNormal(loc, W_term, D_term), [iaf])
```
fritzo: I think we instead want to transform white noise via the iaf, then apply learnable diagonal noise:
```py
def sample_latent(self, *args, **kwargs):
    iaf = self._get_iaf()  # calls pyro.module internally
    iaf_dist = dist.TransformedDistribution(dist.Normal(0., 1.).expand([self.hidden_dim]), iaf)
    loc = pyro.sample("_{}_loc".format(self.prefix), iaf_dist.independent(1))
    scale = pyro.param("{}_scale".format(self.prefix), lambda: torch.ones(self.latent_dim))
    return pyro.sample("_{}_latent".format(self.prefix), dist.Normal(loc, scale).independent(1))

    iaf = self._get_iaf()  # calls pyro.module internally
    prototype = pyro.param("{}_scale".format(self.prefix))
    loc = iaf(prototype.new_zeros(self.hidden_dim))
    scale = None  # no analytical solution
    return loc, scale
```

pyro/contrib/autoguide/__init__.py
```diff
@@ -417,6 +418,8 @@ def quantiles(self, quantiles, *args, **kwargs):
         :rtype: dict
         """
         loc, scale = self._loc_scale(*args, **kwargs)
+        if loc is None or scale is None:
+            raise ValueError('loc and scale cannot be None')
```
fritzo: Thanks for being safe. How about:
```py
if scale is None:
    raise NotImplementedError('scale cannot be computed analytically')
```

pyro/contrib/autoguide/__init__.py
```diff
@@ -561,6 +564,51 @@ def _loc_scale(self, *args, **kwargs):
...
+    over the entire latent space. The guide does not depend on the model's
+    ``*args, **kwargs``.
+
+    Usage::
+
+        guide = AutoTransformedNormal(model, rank=10)
+        svi = SVI(model, guide, ...)
+
+    :param callable model: a generative model
+    :param int rank: the rank of the low-rank part of the covariance matrix
```
fritzo: You can now omit `rank` throughout.

pyro/contrib/autoguide/__init__.py
```diff
@@ -561,6 +564,51 @@ def _loc_scale(self, *args, **kwargs):
         return loc, scale
 
 
+class AutoTransformedNormal(AutoContinuous):
+    """
+    This implementation of :class:`AutoContinuous` uses a Transformed
+    Low Rank MV Normal distribution via a IAF to construct a guide
```
fritzo: Low Rank MV Normal -> Diagonal Normal

pyro/nn/auto_reg_nn.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, input_dim, hidden_dim, output_dim_multiplier=1,
 
         if mask_encoding is None:
             # the dependency structure is chosen at random
-            self.mask_encoding = 1 + torch.multinomial(torch.ones(input_dim - 1) / (input_dim - 1),
+            self.mask_encoding = 1 + torch.multinomial(torch.tensor(input_dim - 1).fill_(1.).float(),
```
fritzo: The input probs are no longer normalized. Can you confirm that `torch.multinomial` normalizes internally, so that your change is valid?

pyro/contrib/autoguide/__init__.py
```diff
@@ -561,6 +564,51 @@ def _loc_scale(self, *args, **kwargs):
...
+        self.sigmoid_bias = sigmoid_bias
+        self.hidden_dim = hidden_dim
+        self.rank = rank
+        super(AutoTransformedNormal, self).__init__(model, prefix)
+
+    def sample_latent(self, *args, **kwargs):
+        """
+        Samples the (single) multivariate normal latent used in the auto guide.
+        """
+        hidden_dim = self.hidden_dim if self.hidden_dim is not None else self.latent_dim
```
fritzo: In might be simpler to
```py
if self.hidden_dim is None:
    self.hidden_dim = self.latent_dim
```
and thereafter refer only to `self.hidden_dim`.

pyro/contrib/autoguide/__init__.py
jpchen: this is in the `AutoContinuous` class so should error if either `loc` or `scale` isnt computed with a general error message. do you want an overridden check in `AutoTransformedNormal`?

pyro/nn/auto_reg_nn.py
jpchen: see [note](https://pytorch.org/docs/stable/torch.html?highlight=multinomial#torch.multinomial)

pyro/contrib/autoguide/__init__.py
jpchen: yeah i did that at first then didnt want masked errors if `_loc_scale` is called before `sample_latent` since `self.latent_dim` is only constructed during `__call__`

fritzo: I think we should just error if `_loc_scale()` is called before `__call__()`

pyro/contrib/autoguide/__init__.py
fritzo: I just think we should raise a `NotImplementedError` rather than a `ValueError`. Feel free to `raise NotImplementedError` with an empty error message.

martinjankowiak:
let me take a look at this please

jpchen:
@fritzo im happy to merge as is (after martin takes a look), per our discussions offline, i may want to add some additional bells and whistles after running them on the example suite

jpchen:
@martinjankowiak @fritzo please take another look.  i think this is what we converged to, per our discussion offline. adding multiple iafs/nns can be done in a follow up

pyro/contrib/autoguide/__init__.py
```diff
@@ -561,6 +562,49 @@ def _loc_scale(self, *args, **kwargs):
         return loc, scale
 
 
+class AutoTransformedNormal(AutoContinuous):
+    """
+    This implementation of :class:`AutoContinuous` uses a Transformed
+    Diagonal Normal distribution via an IAF to construct a guide
```
martinjankowiak: uses a Diagonal Normal distribution transformed by an Inverse Autoregressive Flow

martinjankowiak: also it should probably just becalled something like AutoIAFNormal

pyro/contrib/autoguide/__init__.py
```diff
@@ -561,6 +562,49 @@ def _loc_scale(self, *args, **kwargs):
...
+    """
+    def __init__(self, model, hidden_dim=None, sigmoid_bias=2.0, prefix="auto"):
+        self.sigmoid_bias = sigmoid_bias
+        self.hidden_dim = hidden_dim
+        super(AutoTransformedNormal, self).__init__(model, prefix)
+
+    def sample_latent(self, *args, **kwargs):
+        if self.hidden_dim is None:
+            self.hidden_dim = self.latent_dim
+        loc = pyro.param("{}_loc".format(self.prefix),
```
martinjankowiak: loc/param shouldn't be trainable

pyro/contrib/autoguide/__init__.py
```diff
@@ -561,6 +562,49 @@ def _loc_scale(self, *args, **kwargs):
...
+                           constraint=constraints.positive)
+        iaf = dist.InverseAutoregressiveFlow(self.latent_dim, self.hidden_dim,
+                                             sigmoid_bias=self.sigmoid_bias)
+        pyro.module("{}_iaf".format(self.prefix), iaf.module)
+        self.iaf_dist = dist.TransformedDistribution(dist.Normal(loc, scale), [iaf])
+        return pyro.sample("_{}_latent".format(self.prefix), self.iaf_dist.independent(1),
+                           infer={"is_auxiliary": True})
+
+    def _loc_scale(self, *args, **kwargs):
+        loc = pyro.param("{}_loc".format(self.prefix))
```
martinjankowiak: if _loc_scale is used for median this doesn't make sense

pyro/nn/auto_reg_nn.py
```diff
@@ -83,8 +83,11 @@ def __init__(self, input_dim, hidden_dim, output_dim_multiplier=1,
         self.mask2 = torch.zeros(input_dim * self.output_dim_multiplier, hidden_dim)
 
         for k in range(hidden_dim):
-            # fill in mask1
-            m_k = self.mask_encoding[k].item()
+            if self.mask_encoding.dim():
```
martinjankowiak: why is this needed?

pyro/contrib/autoguide/__init__.py
fritzo: Right, you should simply omit this method

pyro/nn/auto_reg_nn.py
jpchen: you cant index a scalar.

pyro/contrib/autoguide/__init__.py
```diff
@@ -561,6 +562,39 @@ def _loc_scale(self, *args, **kwargs):
         return loc, scale
 
 
+class AutoIAFNormal(AutoContinuous):
+    """
+    This implementation of :class:`AutoContinuous` uses a Transformed
```
martinjankowiak: uses a Diagonal Normal distribution transformed via an IAF

pyro/contrib/autoguide/__init__.py
```diff
@@ -561,6 +562,39 @@ def _loc_scale(self, *args, **kwargs):
...
+    Diagonal Normal distribution via an IAF to construct a guide
+    over the entire latent space. The guide does not depend on the model's
+    ``*args, **kwargs``.
+
+    Usage::
+
+        guide = AutoIAFNormal(model, rank=10)
```
martinjankowiak: rank is gone

tests/contrib/autoguide/test_advi.py
```diff
@@ -179,6 +186,7 @@ def model():
     AutoDiagonalNormal,
     AutoMultivariateNormal,
     AutoLowRankMultivariateNormal,
+    AutoIAFNormal,
```
martinjankowiak: wasn't this to be removed?

are there any "actual" tests of autoiafnormal here?

pyro/contrib/autoguide/__init__.py
martinjankowiak: link to IAF docs

pyro/nn/auto_reg_nn.py
martinjankowiak: i'd remove this. cuz of an IAF is 1D it'll just be an affine transformation of the base distribution. raise an error if user wants 1d autoiaf?

jpchen: we can catch this for auto iaf, but like you said a 1d iaf is a valid thing, and even if it werent that error should be caught, not throw a pytorch tensor error.

tests/contrib/autoguide/test_advi.py
jpchen: why would this be removed?

jpchen: why is that needed for testing parallel enumeration?

pyro/nn/auto_reg_nn.py
```diff
@@ -60,12 +62,14 @@ def __init__(self, input_dim, hidden_dim, output_dim_multiplier=1,
                  mask_encoding=None, permutation=None):
         super(AutoRegressiveNN, self).__init__()
         self.input_dim = input_dim
+        if hidden_dim == 1:
+            warnings.warn('AutoRegressiveNN hidden_dim = 1. Consider using an affine transformation instead.')
```
martinjankowiak: this should be input_dim

eb8680:
@martinjankowiak good to merge?

PR #1176, opened 2018-06-05
Vectorize ELBO computation over num particles

### LINKED ISSUE #791: Parallelize ELBO computation over num_particles
ELBO implementations currently iterate over `num_particles`. Let's also do this in parallel.

## Why?

This is most useful for speeding up our gradient tests. If we had cheap parallelized estimates of the ELBO, then we could replace many of Pyro's expensive integration tests with cheaper gradient tests. This in turn would allow us to test over much larger grids of parameter settings rather than over a handful of carefully crafted models.

## How?

This can follow logic similar to parallel discrete enumeration in #776 . We might set `num_particles=1000`, `particles="parallel"` as arguments to `ELBO`.

Note that a klugey way to do this that already works in `Trace_ELBO` is to add a fake `Categorical` sample site to the beginning of the guide and model, and then use that to broadcast all remaining samples.
```py
def guide():
    pyro.sample("parallelize",
                dist.Categorical(Variable(torch.ones(num_particles))),
    # ...rest of guide code here...
```

neerajprad:
Resolves #791.

My initial plan was to write a separate SVI class for this; but it turned out that modifying the base `ELBO` class was the simplest way to percolate this change to all the other loss classes. 

I have modified only a few tests that were hand-parallelizing over num particles, but we can do a broader sweep later after this PR is merged. It is not strictly necessary though since those tests are fast anyways. Also modified the `conjugate_gradient` test that was not parallelized.

### DIFF
--- a/pyro/infer/elbo.py
+++ b/pyro/infer/elbo.py
@@ -24,7 +24,7 @@ class ELBO(object):
         sample sites in parallel, e.g. if a site sets
         ``infer={"enumerate": "parallel"}``.
     :param bool vectorize_particles: Whether to vectorize the ELBO computation
-        over num_particles. Defaults to False. This requires static structure
+        over `num_particles`. Defaults to False. This requires static structure
         in model and guide. In addition, this wraps the model and guide inside a
         :class:`~pyro.poutine.broadcast` poutine for automatic broadcasting of
         sample site batch shapes, and requires specifying a finite value for
--- a/pyro/infer/elbo.py
+++ b/pyro/infer/elbo.py
@@ -23,7 +23,7 @@ class ELBO(object):
         :func:`pyro.iarange` contexts. This is only required to enumerate over
         sample sites in parallel, e.g. if a site sets
         ``infer={"enumerate": "parallel"}``.
-    :param bool auto_vectorize: Vectorize ELBO computation over `num_particles`.
+    :param bool vectorize_particles: Vectorize ELBO computation over `num_particles`.
         In addition, this wraps the model and guide inside a
         :class:`~pyro.poutine.broadcast` poutine for automatic broadcasting of
         sample site batch shapes. This requires specifying a finite value for
@@ -45,17 +45,16 @@ class ELBO(object):
     def __init__(self,
                  num_particles=1,
                  max_iarange_nesting=float('inf'),
-                 auto_vectorize=False,
+                 vectorize_particles=False,
                  strict_enumeration_warning=True):
         self.num_particles = num_particles
         self.max_iarange_nesting = max_iarange_nesting
-        self.auto_vectorize = auto_vectorize
-        if self.auto_vectorize:
-            if self.num_particles > 1:
-                if self.max_iarange_nesting == float('inf'):
-                    raise ValueError("Automatic vectorization over num_particles requires " +
-                                     "a finite value for `max_iarange_nesting` arg.")
-                self.max_iarange_nesting += 1
+        self.vectorize_particles = vectorize_particles
+        if self.vectorize_particles:
+            if self.num_particles > 1 and self.max_iarange_nesting == float('inf'):
+                raise ValueError("Automatic vectorization over num_particles requires " +
+                                 "a finite value for `max_iarange_nesting` arg.")
+            self.max_iarange_nesting += 1
         self.strict_enumeration_warning = strict_enumeration_warning
 
     def _vectorized_num_particles(self, fn):
@@ -100,7 +99,7 @@ class ELBO(object):
         Runs the guide and runs the model against the guide with
         the result packaged as a trace generator.
         """
-        if self.auto_vectorize:
+        if self.vectorize_particles:
             yield self._get_vectorized_trace(model, guide, *args, **kwargs)
         else:
             for i in range(self.num_particles):
--- a/pyro/infer/elbo.py
+++ b/pyro/infer/elbo.py
@@ -51,10 +51,11 @@ class ELBO(object):
         self.max_iarange_nesting = max_iarange_nesting
         self.vectorize_particles = vectorize_particles
         if self.vectorize_particles:
-            if self.num_particles > 1 and self.max_iarange_nesting == float('inf'):
-                raise ValueError("Automatic vectorization over num_particles requires " +
-                                 "a finite value for `max_iarange_nesting` arg.")
-            self.max_iarange_nesting += 1
+            if self.num_particles > 1:
+                if self.max_iarange_nesting == float('inf'):
+                    raise ValueError("Automatic vectorization over num_particles requires " +
+                                     "a finite value for `max_iarange_nesting` arg.")
+                self.max_iarange_nesting += 1
         self.strict_enumeration_warning = strict_enumeration_warning
 
     def _vectorized_num_particles(self, fn):
--- a/pyro/infer/elbo.py
+++ b/pyro/infer/elbo.py
@@ -23,10 +23,11 @@ class ELBO(object):
         :func:`pyro.iarange` contexts. This is only required to enumerate over
         sample sites in parallel, e.g. if a site sets
         ``infer={"enumerate": "parallel"}``.
-    :param bool vectorize_particles: Vectorize ELBO computation over `num_particles`.
-        In addition, this wraps the model and guide inside a
+    :param bool vectorize_particles: Whether to vectorize the ELBO computation
+        over num_particles. Defaults to False. This requires static structure
+        in model and guide. In addition, this wraps the model and guide inside a
         :class:`~pyro.poutine.broadcast` poutine for automatic broadcasting of
-        sample site batch shapes. This requires specifying a finite value for
+        sample site batch shapes, and requires specifying a finite value for
         `max_iarange_nesting`.
     :param bool strict_enumeration_warning: Whether to warn about possible
         misuse of enumeration, i.e. that
--- a/pyro/infer/elbo.py
+++ b/pyro/infer/elbo.py
@@ -1,6 +1,14 @@
 from __future__ import absolute_import, division, print_function
 
+from abc import abstractmethod, ABCMeta
 
+from six import add_metaclass
+
+import pyro
+import pyro.poutine as poutine
+
+
+@add_metaclass(ABCMeta)
 class ELBO(object):
     """
     :class:`ELBO` is the top-level interface for stochastic variational
@@ -15,6 +23,11 @@ class ELBO(object):
         :func:`pyro.iarange` contexts. This is only required to enumerate over
         sample sites in parallel, e.g. if a site sets
         ``infer={"enumerate": "parallel"}``.
+    :param bool auto_vectorize: Vectorize ELBO computation over `num_particles`.
+        In addition, this wraps the model and guide inside a
+        :class:`~pyro.poutine.broadcast` poutine for automatic broadcasting of
+        sample site batch shapes. This requires specifying a finite value for
+        `max_iarange_nesting`.
     :param bool strict_enumeration_warning: Whether to warn about possible
         misuse of enumeration, i.e. that
         :class:`pyro.infer.traceenum_elbo.TraceEnum_ELBO` is used iff there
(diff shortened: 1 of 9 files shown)


pyro/infer/elbo.py
```diff
@@ -32,7 +45,63 @@ class ELBO(object):
...
+        :return: wrapped callable.
+        """
+
+        def wrapped_fn(*args, **kwargs):
+            if self.num_particles == 1:
+                return fn(*args, **kwargs)
+            with pyro.iarange("num_particles_vectorized", self.num_particles, dim=-self.max_iarange_nesting):
+                return fn(*args, **kwargs)
+
+        return poutine.broadcast(wrapped_fn)
```
fritzo: Is it worth checking whether `fn` is already wrapped in `poutine.broadcast` and avoid wrapping a second time?

pyro/infer/elbo.py
```diff
@@ -32,7 +45,63 @@ class ELBO(object):
     def __init__(self,
                  num_particles=1,
                  max_iarange_nesting=float('inf'),
+                 auto_vectorize=False,
                  strict_enumeration_warning=True):
         self.num_particles = num_particles
         self.max_iarange_nesting = max_iarange_nesting
+        self.auto_vectorize = auto_vectorize
+        if self.auto_vectorize:
```
fritzo: nit: `if self.vectorize_particles and self.num_particles > 1: ...`

pyro/infer/elbo.py
```diff
@@ -15,6 +23,11 @@ class ELBO(object):
         :func:`pyro.iarange` contexts. This is only required to enumerate over
         sample sites in parallel, e.g. if a site sets
         ``infer={"enumerate": "parallel"}``.
+    :param bool auto_vectorize: Vectorize ELBO computation over `num_particles`.
```
jpchen: can you add what it defaults to

pyro/infer/elbo.py
```diff
@@ -32,7 +45,63 @@ class ELBO(object):
...
+        against it.
+        """
+        raise NotImplementedError
+
+    def _get_traces(self, model, guide, *args, **kwargs):
+        """
+        Runs the guide and runs the model against the guide with
+        the result packaged as a trace generator.
+        """
+        if self.auto_vectorize:
```
fritzo: This may error if `self.num_particles == 1` since `max_iarange_nesting` will not have been incremented. Instead gate on `if self.vectorize_particles and self.num_particles > 1`, as used elsewhere, or simplify the gate above.

pyro/infer/trace_elbo.py
```diff
@@ -45,37 +45,36 @@ class Trace_ELBO(ELBO):
...
-            if is_validation_enabled():
-                for site in model_trace.nodes.values():
-                    if site["type"] == "sample":
-                        check_site_shape(site, self.max_iarange_nesting)
-                for site in guide_trace.nodes.values():
-                    if site["type"] == "sample":
-                        check_site_shape(site, self.max_iarange_nesting)
-
-            yield model_trace, guide_trace
+        guide_trace = poutine.trace(guide).get_trace(*args, **kwargs)
```
jpchen: is this method identical sans the for loop over num_particles? the diff shows the whole method has changed but from a cursory glance it looks the same

fritzo: they are identical. have you tried the old "cross your eyes at the side-by-side diff" trick?

jpchen: i do the "cmd-f a randomly sampled substring and see if it appears on both sides". 90% of the time, it works every time.

pyro/infer/elbo.py
neerajprad: The `_vectorized_num_particles` method does a special check for `num_particles=1`, so it should work fine. Note that `vectorize_particles` also calls `poutine.broadcast` (the two effect are hard to decouple), so even when the user has `num_particles=1` with `parallelize_particles=True`, we would like to wrap the model and guide inside a `poutine.broadcast` so that the behavior is consistent.

pyro/infer/trace_elbo.py
neerajprad: Yup, that's right! Mostly `Trace_ELBO`, `TraceGraph_ELBO` and `TraceEnum_ELBO` will only differ in how they generate a single trace. This is reused when we want to generate traces vectorized over num_particles.

pyro/infer/elbo.py
neerajprad: Good idea. I'll check it in `poutine.broadcast` itself.

neerajprad:
> looks good! do you have any basic speed comparisons, at least on the test examples?

@jpchen - This will be many orders of magnitude faster than sequentially sampling traces num particles times. 

e.g. for `test_conjugate_gradients.py` 103 m.s. for 10k particles when they are parallelized vs. more than 2 mins otherwise.

I intend to do some comparisons w.r.t. hand parallelization which is what we should be comparing against. I did not observe any significant difference for most models, but there are cases when we end up adding up to 20% more time as compared to hand parallelization. It is still very fast though (all in milliseconds vs seconds or mins).

pyro/infer/elbo.py
neerajprad: I had to add `fn.__broadcasted__` as an attribute to mark a function that has already been broadcasted.

eb8680: Shouldn't `broadcast` be idempotent?

fritzo: Yes `broadcast` should be idempotent, I'm asking only whether it is worth avoiding double wrapping for efficiency purposes. If it costs more than 5 lines of code, then it's probably not worth avoiding the double wrap.

pyro/poutine/handlers.py
```diff
@@ -248,7 +248,14 @@ def broadcast(fn=None):
     ...     return sample
     """
     msngr = BroadcastMessenger()
-    return msngr(fn) if fn is not None else msngr
+    if fn is not None:
```
fritzo: I guess we can omit this logic until we see a real performance cost of double wrapping.

pyro/infer/elbo.py
neerajprad: Agreed, I'll remove this.

eb8680:
> question: i guess as the code stands we expect the user to know when auto-vectorization will break (e.g. in cases with stochastic control flow etc.)?

Yeah, we'd have to do a bit of static analysis for automatic validation

neerajprad:
> question: i guess as the code stands we expect the user to know when auto-vectorization will break (e.g. in cases with stochastic control flow etc.)?

That's true. I think if an iarange does not specify a `size` arg or if the `size` arg is not consistent with the sample site's observed batch shape, we should throw an error in broadcast poutine itself (by having some kind of a strict mode under which this operates). That will ensure that we are are not doing something odd or silently failing in those situations.

tests/infer/test_enum.py
```diff
@@ -288,26 +288,26 @@ def test_elbo_berns(enumerate1, enumerate2, enumerate3):
...
-            pyro.sample("x3", dist.Bernoulli(q).expand_by([num_particles]), infer={"enumerate": enumerate3})
+        pyro.sample("x1", dist.Bernoulli(q), infer={"enumerate": enumerate1})
+        pyro.sample("x2", dist.Bernoulli(q), infer={"enumerate": enumerate2})
+        pyro.sample("x3", dist.Bernoulli(q), infer={"enumerate": enumerate3})
 
     kl = sum(kl_divergence(dist.Bernoulli(q), dist.Bernoulli(p)) for p in [0.1, 0.2, 0.3])
     expected_loss = kl.item()
     expected_grad = grad(kl, [q])[0]
 
     elbo = TraceEnum_ELBO(max_iarange_nesting=1,
```
fritzo: You can now set `max_iarange_nesting=0` since it is incremented internally.

tests/infer/test_enum.py
```diff
@@ -288,26 +288,26 @@ def test_elbo_berns(enumerate1, enumerate2, enumerate3):
     q = pyro.param("q", torch.tensor(0.75, requires_grad=True))
 
     def model():
-        with pyro.iarange("particles", num_particles):
-            pyro.sample("x1", dist.Bernoulli(0.1).expand_by([num_particles]))
-            pyro.sample("x2", dist.Bernoulli(0.2).expand_by([num_particles]))
-            pyro.sample("x3", dist.Bernoulli(0.3).expand_by([num_particles]))
+            pyro.sample("x1", dist.Bernoulli(0.1))
```
fritzo: nit: outdent

tests/infer/test_enum.py
```diff
@@ -464,39 +464,39 @@ def test_elbo_iarange_iarange(outer_dim, inner_dim, enumerate1, enumerate2, enum
...
-            context1 = pyro.iarange("outer", outer_dim, dim=-2)
-            context2 = pyro.iarange("inner", inner_dim, dim=-3)
-            pyro.sample("w", d.expand_by([num_particles]))
-            with context1:
-                pyro.sample("x", d.expand_by([outer_dim, num_particles]))
-            with context2:
-                pyro.sample("y", d.expand_by([inner_dim, 1, num_particles]))
-            with context1, context2:
-                pyro.sample("z", d.expand_by([inner_dim, outer_dim, num_particles]))
+        context1 = pyro.iarange("outer", outer_dim, dim=-2)
```
fritzo: Can we change these dims to `-1,-2` now?

tests/infer/test_gradient.py
```diff
@@ -26,35 +26,35 @@ def test_subsample_gradient(Elbo, reparameterized, subsample):
     pyro.clear_param_store()
     data = torch.tensor([-0.5, 2.0])
     subsample_size = 1 if subsample else len(data)
-    num_particles = 50000
     precision = 0.06
     Normal = dist.Normal if reparameterized else fakes.NonreparameterizedNormal
 
     def model(subsample):
```
fritzo: I think you should wrap these model and guide with `@poutine.broadcast` and avoid relying on the implicit `@poutine.broadcat` due to `num_particles>1` and `vectorize_particles=True`. While it is correct without the explicit decorator, it is better style to be explicit, and users may look to our tests for idiomatic usage.

tests/infer/test_valid_models.py
```diff
@@ -879,3 +879,103 @@ def model():
...
+    guide = config_enumerate(guide) if Elbo is TraceEnum_ELBO else guide
+    assert_ok(model, guide, Elbo(num_particles=10,
+                                 vectorize_particles=True,
+                                 max_iarange_nesting=2,
+                                 strict_enumeration_warning=False))
+
+
+@pytest.mark.parametrize('enumerate_', [None, "sequential", "parallel"])
+def test_enum_discrete_vectorized_num_particles(enumerate_):
+    num_particles = 50
```
fritzo: Can you `@pytest.mark.parametrize('num_particles', [1, 10])` to test the single-particle edge case?

tests/infer/test_enum.py
neerajprad: My bad, fixed.

tests/infer/test_enum.py
neerajprad: Done.

tests/infer/test_enum.py
neerajprad: Done.

tests/infer/test_gradient.py
neerajprad: `vectorize_particles` with `num_particles=1` would still use `@poutine.broadcast`, so as not to have some special surprising behavior for `num_particles=1`. That was part of the reason why I thought of calling it `auto_vectorize` to emphasize that its a bit more general than adding an outer iarange, and could be used a replacement for `@poutine.broadcast` with or without `num_particles`. That said, `vectorize_particles` is a clearer arg name, but it makes the broadcasting side-effect seem unexpected. Will change to using `@poutine.broadcast`, unless there is a better name that we can use for this arg.

tests/infer/test_valid_models.py
neerajprad: Good idea, will do.

tests/infer/test_conjugate_gradients.py
```diff
@@ -17,15 +16,15 @@ def do_test_gradients(self, N, reparameterized):
...
+        elbo = TraceGraph_ELBO(num_particles=10000, vectorize_particles=True, max_iarange_nesting=1)
         elbo.loss_and_grads(self.model, self.guide, reparameterized=reparameterized)
 
         for i in range(1, N + 1):
             for param_prefix in ["loc_q_%d", "log_sig_q_%d", "kappa_q_%d"]:
                 if i == N and param_prefix == 'kappa_q_%d':
                     continue
                 actual_grad = pyro.param(param_prefix % i).grad
-                assert_equal(actual_grad, 0.0 * actual_grad, prec=0.20, msg="".join([
+                assert_equal(actual_grad, 0.0 * actual_grad, prec=0.10, msg="".join([
```
fritzo: Nice!

pyro/infer/elbo.py
```diff
@@ -15,6 +23,11 @@ class ELBO(object):
         :func:`pyro.iarange` contexts. This is only required to enumerate over
         sample sites in parallel, e.g. if a site sets
         ``infer={"enumerate": "parallel"}``.
+    :param bool vectorize_particles: Vectorize ELBO computation over `num_particles`.
```
fritzo: consider adding more info:
"Whether to vectorize the ELBO computation over `num_particles`. Defaults to False. This requires static structure in model and guide."

neerajprad: Thanks, will update.

neerajprad:
The failure is due to the `observations` library unable to download MNIST, and is unrelated. I am fine with merging this PR as is, once tests pass. Will address the issue of slowness in `.expand`, and adding more warnings separately.

PR #1177, opened 2018-06-05
Increase test iterations for CUDA testing

neerajprad:
The `test_advi.py::test_median` test seems to require significantly more (3x-5x) iterations to converge to the expected results on CUDA, with `auto_guide_callable` needing the most number of iterations. I am not sure why that is the case.

 @fritzo - I am just increasing the number of iterations for now, but let me know if this needs to be further investigated.


jpchen:
i upped the num iterations in #1173 but only for the transformed, i can do it for everything in that PR, to save conflicts ?

neerajprad:
> i upped the num iterations in #1173 but only for the transformed, i can do it for everything in that PR, to save conflicts ?

@jpchen - Sure. Happy to close this, and you can increase the iterations in #1173 if that's ready to merge soon.

Issue #1137, opened 2018-05-11
Mass Matrix Adaptation for HMC and NUTS


LoganWalls:
I'd like to contribute the mass matrix adaptation mentioned in #1093 . I actually have some preliminary work on it already, but I've run into an issue and I'm not sure how you'd like to proceed: 

Currently the MCMC kernels have no access to the number of warmup iterations (since transition between warmup and sampling is handled by the MCMC wrapper), but if we want to adapt the mass matrix during warmup using a block scheme similar to stan (see section 34.2 in the [stan reference manual](http://mc-stan.org/users/documentation/)) then the kernel needs access to the number of warmup iterations to divide the warmup into blocks.

I thought about putting this scheduling logic into the MCMC wrapper, but since adapting a mass matrix in this way is specific to certain kernels (HMC and NUTS) it feels like it belongs in the kernels themselves.

However, giving the kernels access to the number of warmup iterations is not quite straightforward: since they are initialized by the user (not the MCMC wrapper) we can't pass it to the constructor, and currently all of the arguments to `HMC.setup()` are passed directly into `get_trace()`.

Some options that come to mind are:

1. Add the number of warmup iterations as a parameter in `HMC.setup()` (Seems reasonably clean to me, but maybe I'm missing something?)
2. Implementing something like `kernel.configure_adaptation(warmup_duration=200)` which would be called by the wrapper. (This approach feels less clean because maybe not every kernel will need parameter adaptation, but it won't modify the existing `kernel.setup()` signature).
3. In the MCMC wrapper directly set a property on the kernel: `self.kernel.warmup_duration = 200`. (Simple, but feels more like a hack).

Any thoughts on which of these options (or others?) would be the best approach?

eb8680:
Option 2 seems fine for now; you can just add a base implementation of `configure_adaptation` to `TraceKernel`.

@neerajprad we should probably clean up the MCMC code a little bit to make it easier to customize this sort of thing in the future.

neerajprad:
I would suggest going with option (1), so that we do not have two places that specify the `warmup_steps` - the kernel themselves and the MCMC constructor (since burnin or warmup is a more general concept). 

So you can just modify the `MCMC` class as follows and the corresponding `setup` methods in HMC and NUTS. I think that should cover your use case?

```
self.kernel.setup(self.warmup_steps, *args, **kwargs)
```

neerajprad:
[...]

PR #1178, opened 2018-06-05
HMC Mass Matrix Adaptation

LoganWalls:
This implements mass-matrix adaptation for HMC / NUTS kernels as discussed in #1137 .

### Usage
Two new parameters have been introduced for HMC/NUTS kernels:
- `adapt_mass (bool)`: controls whether or not to use mass adaptation
- `mass_structure (str)`: the (assumed) structure of the mass matrix (one of: `identity` (default), `diagonal`, or `dense`)

(The decision to make these two separate parameters was based on consistency with the step-size adaptation API, and to make supporting user-supplied mass matrices easier in the future).

If mass is set to `identity` or `adapt_mass` is false then no mass adaptation will take place. Otherwise, :
```python
# Both use no mass adaptation (same behavior as HMC before this PR)
nuts_kernel = NUTS(model, adapt_mass=True, mass_structure='identity')
nuts_kernel = NUTS(model, adapt_mass=False, mass_structure='dense')

# Estimates a diagonal mass matrix
nuts_kernel = NUTS(model, adapt_mass=True, mass_structure='diagonal')
```

### Changes
- `warmup_steps` is now part of the `TraceKernel.setup()` API  and the transition between warmup and sampling (eg. `HMC.end_warmup()`) is now handled by the kernel itself instead of the `MCMC` wrapper (see #1137 ).
- Adaptation (both for step-size and mass) follows a windowed schedule similar to the one used in Stan (see [Stan Reference](https://github.com/stan-dev/stan/releases/download/v2.17.0/stan-reference-2.17.0.pdf) - section 34.2).
- To support consistent ordering when unrolling the stochastic nodes `z` in `HMC.sample()` and `NUTS.sample()` have been changed from `dict` to `OrderedDict`.
- Adds a new module `pyro.ops.welford` containing an online (co)variance estimator based on [Welford's method](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Online_algorithm).

### Tests Implemented
- Tests for Welford estimator to ensure it properly calculates sample (co)variance.
- Tests for HMC windowed parameter adaptation (ensures the correct parameters are being adapted at each step of the warm-up period).

### Remaining Issues
- As mentioned in #1137 , it appears the mass adaptation and dual-averaging don't play nice together. If you run the sampler with both step-size and mass adaptation enabled, as soon as a mass matrix is applied to the kinetic energy calculation the step-size plummets and sampling takes forever. Does anyone have thought on this?
_(If I enforce a minimum step-size while using the mass-adaptation the acceptance-rate does improve compared to using no mass-adaptation so I think it's being applied correctly, but I'd appreciate a second opinion!)_


neerajprad:
Thanks for implementing this! Can we test this on some simple models? An example would be this model from the [forum](https://forum.pyro.ai/t/simple-mcmc-on-gamma-beta-providing-wrong-estimates/259/3), and it will be interesting to see if mass matrix adaptation helps NUTS in this case. You can also just add a mass matrix adaptation as a parameter to existing HMC/NUTS tests to validate that we get the same answers  as we do without mass adaptation. I will be reviewing this in more detail.

pyro/ops/welford.py
```diff
@@ -0,0 +1,70 @@
...
+            delta_pre = site_value - mean
+            updated_mean = mean + delta_pre / self.n_samples
+            delta_post = site_value - updated_mean
+            deltas[site_name] = (updated_mean, delta_pre, delta_post)
+        return deltas
+
+    def _unroll(self, z):
+        unrolled = OrderedDict()
+        for site_name, rolled in z.items():
+            for i, x in enumerate(rolled.view(-1)):
```
fritzo: Use `.reshape(-1)` rather than `.view(-1)` to avoid issues with non-contiguous tensors.

pyro/ops/welford.py
```diff
@@ -0,0 +1,70 @@
...
+    def update(self, z):
+        self.n_samples += 1
+        deltas = self._deltas(self._unroll(z))
+        for i, (site_x, (mean_x, delta_x_pre, delta_x_post)) in enumerate(deltas.items()):
+            self.means[site_x] = mean_x
+            for j, (site_y, (mean_y, delta_y_pre, delta_y_post)) in enumerate(deltas.items()):
+                # Only compute the upper triangular covariance.
+                if i == j or (i < j and not self.diagonal):
+                    self.variances[(site_x, site_y)] = self.variances.get((site_x, site_y), 0.) + \
+                                                       delta_x_pre * delta_y_post
```
fritzo: I think this will work for scalars but not for tensors of different shape. I think you'll need to keep track of outer products via something like
```py
self.covariances[site_x, site_y] += delta_x_pre.unsqueeze(-1) * delta_y_post
```
Alternatively, you could `torch.cat` everything into a single tensor, as we do in pyro.contrib.autoguide.

pyro/infer/mcmc/hmc.py
```diff
@@ -78,6 +87,9 @@ def __init__(self, model, step_size=None, trajectory_length=None,
             self.trajectory_length = 2 * math.pi  # from Stan
         self.num_steps = max(1, int(self.trajectory_length / self.step_size))
         self.adapt_step_size = adapt_step_size
+        self.mass_structure = mass_structure
+        # Don't need to adapt mass if the mass structure is identity.
+        self.adapt_mass = adapt_mass and self.mass_structure != 'identity'
```
neerajprad: Can we just use the `mass_structure` param with `identity` as the default, since the `adapt_mass` is redundant? We already have a few parameters that are dependent, and we would like to reduce dependent parameters as much as possible.

pyro/infer/mcmc/hmc.py
```diff
@@ -160,6 +186,62 @@ def _find_reasonable_step_size(self, z):
...
+                i: adaptation_window(adapt_step_size=self.adapt_step_size, adapt_mass=True)
+                for i in mass_adaptation_windows})
+            self._adaptation_schedule[start_buffer + mass_adaptation_period] = adaptation_window(
+                    adapt_step_size=self.adapt_step_size, adapt_mass=False)
+
+    def _adapt_parameters(self, accept_prob, z):
+        if self._t in self._adaptation_schedule:
+            self._adaptation_window = self._adaptation_schedule[self._t]
+            if self._adaptation_window.adapt_mass and self._mass_estimator:
+                # Update mass and reset online estimator's memory.
```
neerajprad: Will this mass update get applied for the last schedule where `adapt_mass=True`?

pyro/ops/welford.py
```diff
@@ -0,0 +1,70 @@
...
+
+    [1] `The Art of Computer Programming`,
+    Donald E. Knuth
+    """
+    def __init__(self, diagonal=True):
+        self.diagonal = diagonal
+        self.reset()
+
+    def _deltas(self, z):
+        deltas = OrderedDict()
```
neerajprad: Since we are using `OrderedDict` for `z` and `r` in HMC and NUTS, can we simplify the interface of `WelfordCovariance` to use lists instead, and just pass in `z.values()` here? It will be nice to keep utilities in `ops` independent of constructs like `site_name` which are still specific to Pyro inference.

pyro/infer/mcmc/hmc.py
```diff
@@ -160,6 +186,62 @@ def _find_reasonable_step_size(self, z):
...
+                raise ValueError(
+                        "mass_type must be one of: 'identity', 'diagonal', 'dense' (got '{}')".format(
+                                self.mass_structure))
+            # Buffer and window sizes from stan
+            start_buffer = 75
+            end_buffer = 25
+            initial_window = 50
+            total = start_buffer + end_buffer + initial_window
+            if self._warmup_steps < total:
+                raise ValueError(
```
neerajprad: How about we just use these number of warmup steps by default if step size or mass adaptation is enabled?

fehiepsi:
@LoganWalls Thanks for implementing this! I have just finished my defense. When you separate this, I will give detailed reviews on them. :)

pyro/ops/welford.py
LoganWalls:
```python
self.variances[site_x, site_y] += delta_x_pre * delta_y_post
```
I think the above is actually ok in this case because we're iterating over the deltas (which have already been unrolled) so both x and y are always scalars in this case. Unless I've misunderstood?

fehiepsi:
@LoganWalls I have revised your pull request at #1442. Thanks a lot for your contribution and nice discussion along the way!! :)

Issue #1175, opened 2018-06-04
[Feature request] Improving output messages and warning handling for MCMC


arranger1044:
### Issue Description
As for version `0.2.1`, logging capabilities of MCMC inference routines (`HMC` and `NUTS`) could be be improved in warning/error handling, when critical states are happening.

For instance, acceptance probability is everywhere 0 and/or no actual samples from the posterior after burn in have been collected.

One possibility would to put such warnings into the `INFO` level of logging in python.

Generally, PyMC3 does a remarkable job in highlighting many diagnostics during sampling by default and could be taken as one example.

### Code Snippet
For one sample use case, see this other issue: https://github.com/uber/pyro/issues/1174

PR #1179, opened 2018-06-06
Improve default handling of logger messages

neerajprad:
We had our logging set up in a way that the default handler was a `NullHandler`, and the onus of setting up the logging was on the user. While this works great for most libraries used in production applications, most users do not use Pyro inside an application where the root logger is already configured. e.g. #1175, and many other forum postings where some issues with HMC/NUTS sampling were not apparent simply because the user did not configure a logger.

This is a minor change to enforce the following:
 - If no root logger is configured, the default handler will just output to stdout, as most users would expect (and not swallow `INFO`, `DEBUG` level messages).
 - If a root logger is already configured (say, via `logging.basicConfig(...)`) before "pyro" is imported, then all log messages will be directed to the root logger.  In such cases, the output should not be duplicated. e.g. in certain tests, the output from the logger was getting duplicated. 
    ```
    tests/infer/test_enum.py::test_gmm_iter_discrete_traces[gmm_model-dense-3] D 
    tests.infer.test_enum 	 M z_0 = 1
    DEBUG M z_0 = 1
    D tests.infer.test_enum 	 M z_0 = 1
    DEBUG M z_0 = 1 
    D tests.infer.test_enum 	 M   z_1 = 1
    ...
    ```
 - If a root logger is configured after "pyro" is imported, we will use the default handler to output messages from inside "pyro" and the application specific logging will go to the root logger. e.g. `baseball.py` uses a different format for its own output.

Partly addresses #1175; we need better warnings and diagnostics separately.

### DIFF
--- a/pyro/logger.py
+++ b/pyro/logger.py
@@ -1,0 +1,16 @@
+from __future__ import absolute_import, division, print_function
+
+import logging
+
+
+default_format = '%(levelname)s \t %(message)s'
+log = logging.getLogger("pyro")
+log.setLevel(logging.INFO)
+
+
+if not logging.root.handlers:
+    default_handler = logging.StreamHandler()
+    default_handler.setLevel(logging.INFO)
+    default_handler.setFormatter(logging.Formatter(default_format))
+    log.addHandler(default_handler)
+    log.propagate = False
--- a/pyro/__init__.py
+++ b/pyro/__init__.py
@@ -1,7 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
-import logging
-
+from pyro.logger import log
 import pyro.poutine as poutine
 from pyro.poutine import condition, do
 from pyro.primitives import (clear_param_store, enable_validation, get_param_store, iarange, irange, module, param,
@@ -16,10 +15,6 @@ try:
 except ImportError:
     __version__ = version_prefix
 
-# Default logger to prevent 'No handler found' warning.
-logging.getLogger(__name__).addHandler(logging.NullHandler())
-
-
 __all__ = [
     "__version__",
     "clear_param_store",
@@ -29,6 +24,7 @@ __all__ = [
     "get_param_store",
     "iarange",
     "irange",
+    "log",
     "module",
     "param",
     "poutine",
--- a/pyro/infer/mcmc/mcmc.py
+++ b/pyro/infer/mcmc/mcmc.py
@@ -1,8 +1,8 @@
 from __future__ import absolute_import, division, print_function
 
-import logging
 import math
 
+import pyro
 from pyro.infer import TracePosterior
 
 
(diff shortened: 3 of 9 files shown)


pyro/logger.py
```diff
@@ -0,0 +1,16 @@
...
+log = logging.getLogger("pyro")
+log.setLevel(logging.INFO)
+
+
+if not logging.root.handlers:
+    default_handler = logging.StreamHandler()
+    default_handler.setLevel(logging.INFO)
+    default_handler.setFormatter(logging.Formatter(default_format))
+    log.addHandler(default_handler)
+    log.propagate = False
```
neerajprad: This is to ensure that we are not duplicating log messages by passing the same upwards to the root logger.

pyro/logger.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import absolute_import, division, print_function
+
+import logging
+
+
+default_format = '%(levelname)s \t %(message)s'
+log = logging.getLogger("pyro")
+log.setLevel(logging.INFO)
```
jpchen: should this be configurable? at least between debug and info

fritzo:
I guess an alternative would be to start using `logging.basicConfig(level=logging.INFO)` in more of our examples and tutorials. This seems to work for [gensim](https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/Corpora_and_Vector_Spaces.ipynb).

pyro/logger.py
neerajprad: We can configure it as `pyro.log.setLevel(logging.DEBUG)`. I'm keeping `INFO` as default as a lot of diagnostic messages are passed in `INFO`.

neerajprad:
> LGTM, although I don't know how idiomatic this use of logging is.

I think our older approach of binding the default handler to a `NullHandler` was probably more idiomatic for libraries. However, I think that our typical use isn't that of most libraries which are a small part of the stack used to run a bigger application, and it is okay to apply a global config to suppress all low level messages from the stack. In our case, the use is much more exploratory, so those `INFO` messages provide valuable diagnostics, and it is okay to have a more visible default logger, like Stan or PyMC3 (a global config can still be applied, just that we default to more visibility). I adapted this from how PyMC3 initializes logging. TensorFlow has a much more involved [logging](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/platform/tf_logging.py) setup, but it will be an overkill until we better understand more logging requirements. 

> I guess an alternative would be to start using logging.basicConfig(level=logging.INFO) in more of our examples and tutorials. 

That might help. We also had it in all of HMC/NUTS tests and examples, but I still got quite a few issues on the forum and on our distributions slack where the users did not specify an initial config, and would have benefitted from some diagnostics information to debug the issue.

fritzo:
Thanks for explaining your design choices!

PR #1181, opened 2018-06-07
add .gitattributes so that can be correctly labelled as python project by github linguistic

overshiki:
currently the project is labelled as jupyter notebook, which is not proper


neerajprad:
It seems that the issue is that a lot of the auto generated lines of code in notebooks begins to weigh over the python lines of code - which is why the stats say 63% ipython notebook and 35% python. The classification of files seems alright. See - https://github.com/github/linguist/issues/3282, https://github.com/github/linguist/issues/3316. So rather than overriding this to classify all files as python, an alternative as suggested in one of the issues would be to omit notebook files.

```
.ipynb linguist-documentation
```

jpchen:
completely off tangent, but i still wish github would handle notebooks better in general (eg diffs and integration)

.gitattributes
```diff
@@ -0,0 +1,2 @@
+*.py linguist-language=python
+*.ipynb linguist-documentation
```
neerajprad: nit: newline.

neerajprad:
I think all our PRs are failing trying to download the MNIST dataset on `observations`.

jpchen:
yeah i restarted that failing test, but ive been encountering that myself. maybe we should move off that library

neerajprad:
> yeah i restarted that failing test, but ive been encountering that myself. maybe we should move off that library

I am seeing some network issues with travis in the last couple of days, so it could just be that. Will cut out a task if this issue with `observations` persists.

jpchen:
thanks for doing this @overshiki !

missikram:
_Updating_ **.gitattributes** _with below line of code will set the language as python._
- [ ] *.python linguist-language=python

Issue #1185, opened 2018-06-13
Master issue for pyro.contrib.tracking


fritzo:
## Components

Data association in pyro.contrib.tracking.assignment (@fritzo):
- [x] #1186 multi-frame enumeration solver
- [x] #1190 flat assignment solvers (for data with one observation per frame)
- [x] #1194 dense multi-frame bp solver
- [x] #1217 tests of algebraic properties
- ~~sparse multi-frame bp solver~~

Motion models and measurement models:
- [x] ND gaussian/kalman filters (@jpchen)

Track maintenance utilities:
- [x] #1211 `ApproxSet` for cheap track initialization via space carving
- [x] #1211 Euclidean `LSH` for cheap track merging
- [x] #1219 A generic `merge_points()` function using `heapq` and `LSH`

Lots of examples
- [x] #1186 simple 1D example
- [x] #1394 tutorial on differentiable EKF
- ~~example of neural detector~~

## Design Questions:

- Will the generative model work at all, or will we need a factor-graph model?
- How should we encode exact inference via dynamic programming? E.g. unscented filtering via TraceEnum message passing, or Laplace approximation message passing (a la EKF).

alicanb:
Re. metric evaluation, I've been going through some implementations online. It looks like pymot hasn't been maintained for a while... The evaluation suite that's used in [MOT Challenge](https://bitbucket.org/amilan/motchallenge-devkit) in Matlab, but there's a python version of it [here](https://github.com/cheind/py-motmetrics/tree/develop) but people have been complaining results are different from the original.

PR #1186, opened 2018-06-13
Create minimal pyro.contrib.tracking module

fritzo:
Addresses #1185

This implements a first draft of a minimal tracking module in Pyro. The core parts of this PR are:
- A marginal assignment solver for multi-frame multi-object data association. (this assumes object existence does not change in time; we will need to validate that assumption)
- An example `model` and `guide` that make use of the assignment solver. The interesting part is the fully generative model that is compatible with the assignment solver. While the guide does use the solver it currently uses only very naive state estimation; smarter filtering will be moved to a follow-up PR.

## Tasks for this first PR

- [x] add enumeration implementation 
- [ ] add bp implementation
- [ ] add tests for `MarginalAssignmentPersistent`
- [ ] add tests for bp implementation
- [ ] fix naive svi example


fritzo:
OK I think this is now ready for detailed review.

pyro/contrib/tracking/assignment.py
```diff
@@ -0,0 +1,152 @@
...
+import math
+import numbers
+import warnings
+
+import torch
+
+import pyro.distributions as dist
+
+
+def _warn_if_nan(tensor, name):
```
jpchen: I've wanted to put this in a `pyro.debug` module so it can be exposed as a tool for users but we can do that before release

fritzo:
ready to merge.

pyro/contrib/tracking/assignment.py
fritzo: agreed. we're also using it in [pyro.ops.newton](https://github.com/uber/pyro/blob/dev/pyro/ops/newton.py#L9)

Issue #1180, opened 2018-06-07
Make poutine.broadcast as fast as hand parallelization


neerajprad:
Currently, `poutine.broadcast` will be slower than hand-parallelizing the same example using `.expand_by`. This is because `poutine.broadcast` uses `.expand` which constructs new distribution instances under the hood, whereas `.expand_by` reuses distribution instances. 

Example:
```python
In [10]: import torch
In [11]: prof_expand = lambda d: d.expand((1000, 1000, 10)).log_prob(torch.ones(1000, 1000, 10))
In [11]: prof_expand_by = lambda d: d.expand_by((1000, 1000, 10)).log_prob(torch.ones(1000, 1000, 10))

In [26]: %timeit prof_expand(dist.Normal(0., 1.))
226 ms ± 4.37 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [27]: %timeit prof_expand_by(dist.Normal(0., 1.))
121 ms ± 1.32 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

This is understandable as constructing distribution instances isn't cheap. For many small models, this difference might not be perceptible, but with lots of sample sites and iaranges, the difference could add up for more complicated models. 

With more usage of `poutine.broadcast`, it will be good to start thinking about how we can make `.expand` more performant. A couple of options are:
 - defer to `.expand_by` whenever possible.
 - implement `.expand` in `torch.distributions` in a way that reuses those instances or at least bypasses some of the expensive steps in `__init__` (I haven't thought through this fully, and it might not be possible).

fritzo:
It sounds reasonable to defer to `.expand_by()` whenever possible, since this is not very intrusive.

neerajprad:
Deferring to `.expand_by` mostly fixes the issue, but I discovered that there is another source of slowness in the `Dice` class when the `num_particles` dimension is anything but the rightmost:

```python
class Dice(object):
    ...
    def in_context(self, shape, ordinal):
        ...
        # TODO replace this naive sum-product computation with message passing.
        log_prob = sum(self._get_log_factors(ordinal))
```

This arises due to summing a tensor with a large batch dimension (due to num particles for instance) that is not the right-most dimension.

```python
In [13]: %timeit sum([np.random.randn(2, 1000000, 1, 1), np.random.randn(2, 1, 1000000, 2, 1)])
258 ms ± 10.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

[...]

PR #1187, opened 2018-06-14
Defer to .expand_by wherever possible in distribution.expand

neerajprad:
Mostly addresses #1180, except for another perf issue (relatively minor) that I discovered and mentioned in #1180.

### Changes:
 - `.expand` calls `.expand_by` if we are merely appending batch dims to the left.
 - Implements `ReshapedDistribution.expand`. Multiple calls to this should be much faster especially if the base distribution has scalar parameters as this merely readjusts the sample shape. Without this implementation, multiple expands would otherwise call the more restrictive `TorchDistributionMixin.expand` and fail. e.g. 

    ```python
    d = dist.Normal(0., 1.)
    d.expand([2, 1, 1])  # Returns ReshapedDistribution
    d.expand([2, 1, 2])  # Calls TorchDistributionMixin.expand which 
                         # is more restrictive and fails.
     ```
 - Our non reparameterized distributions were accidentally getting their `has_rsample` attribute reset when expanded. This adds a fix for the issue, and increases the number of particles for the failing tests.

### Perf impact (pasted from #1180):
`test_enum.py::test_elbo_iarange_iarange` (set of 81 total tests) on my local mac notebook:

 - Without any fixes - 46 seconds.
 - Deferring to .expand - 35 seconds.
 - Using iarange as rightmost - 30 seconds.

### DIFF
--- a/pyro/distributions/torch.py
+++ b/pyro/distributions/torch.py
@@ -3,197 +3,304 @@ from __future__ import absolute_import, division, print_function
 import torch
 
 from pyro.distributions.torch_distribution import TorchDistributionMixin
+from pyro.distributions.util import ShapeMismatchError
 
 
 class Bernoulli(torch.distributions.Bernoulli, TorchDistributionMixin):
     def expand(self, batch_shape):
-        validate_args = self.__dict__.get('validate_args')
-        if 'probs' in self.__dict__:
-            probs = self.probs.expand(batch_shape)
-            return Bernoulli(probs=probs, validate_args=validate_args)
-        else:
-            logits = self.logits.expand(batch_shape)
-            return Bernoulli(logits=logits, validate_args=validate_args)
+        try:
+            sample_shape = self._sample_shape(batch_shape)
+            return self.expand_by(sample_shape)
+        except ShapeMismatchError:
+            validate_args = self.__dict__.get('validate_args')
+            if 'probs' in self.__dict__:
+                probs = self.probs.expand(batch_shape)
+                return Bernoulli(probs=probs, validate_args=validate_args)
+            else:
+                logits = self.logits.expand(batch_shape)
+                return Bernoulli(logits=logits, validate_args=validate_args)
 
 
 class Beta(torch.distributions.Beta, TorchDistributionMixin):
     def expand(self, batch_shape):
-        validate_args = self.__dict__.get('validate_args')
-        concentration1 = self.concentration1.expand(batch_shape)
-        concentration0 = self.concentration0.expand(batch_shape)
-        return Beta(concentration1, concentration0, validate_args=validate_args)
+        try:
+            sample_shape = self._sample_shape(batch_shape)
+            return self.expand_by(sample_shape)
+        except ShapeMismatchError:
+            validate_args = self.__dict__.get('validate_args')
+            concentration1 = self.concentration1.expand(batch_shape)
+            concentration0 = self.concentration0.expand(batch_shape)
+            beta = Beta(concentration1, concentration0, validate_args=validate_args)
+            beta.has_rsample = getattr(self, 'has_rsample')
+            return beta
 
 
 class Categorical(torch.distributions.Categorical, TorchDistributionMixin):
     def expand(self, batch_shape):
-        batch_shape = torch.Size(batch_shape)
-        validate_args = self.__dict__.get('validate_args')
-        if 'probs' in self.__dict__:
-            probs = self.probs.expand(batch_shape + self.probs.shape[-1:])
-            return Categorical(probs=probs, validate_args=validate_args)
-        else:
-            logits = self.logits.expand(batch_shape + self.logits.shape[-1:])
-            return Categorical(logits=logits, validate_args=validate_args)
+        try:
+            sample_shape = self._sample_shape(batch_shape)
+            return self.expand_by(sample_shape)
+        except ShapeMismatchError:
+            batch_shape = torch.Size(batch_shape)
+            validate_args = self.__dict__.get('validate_args')
+            if 'probs' in self.__dict__:
+                probs = self.probs.expand(batch_shape + self.probs.shape[-1:])
+                return Categorical(probs=probs, validate_args=validate_args)
+            else:
+                logits = self.logits.expand(batch_shape + self.logits.shape[-1:])
+                return Categorical(logits=logits, validate_args=validate_args)
 
 
 class Cauchy(torch.distributions.Cauchy, TorchDistributionMixin):
     def expand(self, batch_shape):
-        validate_args = self.__dict__.get('validate_args')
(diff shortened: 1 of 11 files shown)


pyro/distributions/torch.py
```diff
@@ -3,197 +3,304 @@
...
-        return Cauchy(loc, scale, validate_args=validate_args)
+        try:
+            sample_shape = self._sample_shape(batch_shape)
+            return self.expand_by(sample_shape)
+        except ShapeMismatchError:
+            validate_args = self.__dict__.get('validate_args')
+            loc = self.loc.expand(batch_shape)
+            scale = self.scale.expand(batch_shape)
+            cauchy = Cauchy(loc, scale, validate_args=validate_args)
+            cauchy.has_rsample = getattr(self, 'has_rsample')
```
fritzo: I suspect this is an incorrect fix for the `.has_rsample` issue, and that what we really want is
```py
return type(self)(loc, scale, validate_args=validate_args)
```
This way individual methods can be correctly overridden, but we must assume consistency of `__init__` args. WDYT?

pyro/distributions/torch_distribution.py
```diff
@@ -64,6 +64,35 @@ def shape(self, sample_shape=torch.Size()):
...
+        # check sizes of existing dims
+        for dim in range(-1, -1 - len(self.batch_shape), -1):
+            if batch_shape[dim] == -1:
+                batch_shape[dim] = self.batch_shape[dim]
+            elif batch_shape[dim] != self.batch_shape[dim]:
+                if self.batch_shape[dim] != 1:
+                    raise ValueError("Cannot broadcast dim {} of size {} to size {}".format(
+                        dim, self.batch_shape[dim], batch_shape[dim]))
+                else:
+                    raise ShapeMismatchError("The rightmost dimensions of the existing batch " +
```
fritzo: Hmm it seems like we're doing a lot of work before raising this `ShapeMismatchError`. I would have guessed that a simple check would have worked, something like
```py
def expand(self, shape):
    shape = torch.Size(shape)
    cut = len(shape) - len(self.batch_shape)
    left, right = shape[:cut], shape[cut:]
    if right == self.batch_shape:
        return self.expand_by(left)
    # ... continue with existing expand() implementation...
```
Can you help me se what subtlety this simple solution misses?

pyro/distributions/torch_distribution.py
```diff
@@ -82,22 +111,14 @@ def expand(self, batch_shape):
...
-                        dim, self.batch_shape[dim], batch_shape[dim]))
-                else:
-                    raise NotImplementedError("https://github.com/uber/pyro/issues/1119")
-        sample_shape = batch_shape[:len(batch_shape) - len(self.batch_shape)]
-        return self.expand_by(sample_shape)
+        try:
+            sample_shape = self._sample_shape(batch_shape)
+            return self.expand_by(sample_shape)
+        except ShapeMismatchError:
+            raise NotImplementedError("`TorchDistributionMixin.expand()` cannot expand " +
```
fritzo: nit: no need for plus signs here, it's cheaper to concatenate the strings in parsing than in execution.

pyro/distributions/torch_distribution.py
```diff
@@ -258,6 +279,24 @@ def __init__(self, base_dist, sample_shape=torch.Size(), reinterpreted_batch_ndi
         batch_shape, event_shape = shape[:batch_dim], shape[batch_dim:]
         super(ReshapedDistribution, self).__init__(batch_shape, event_shape)
 
+    def expand(self, batch_shape):
+        batch_shape = torch.Size(batch_shape)
+        proposed_shape = broadcast_shape(self.batch_shape, batch_shape)
+        if tuple(reversed(proposed_shape)) > tuple(reversed(batch_shape)):
+            raise ValueError("Existing batch shape {} cannot be expanded" +
```
fritzo: nit: ditto, no need for plus sign

pyro/distributions/torch.py
neerajprad: +1, I overlooked that the fakes implementation is just subclassing off of the distribution classes. This should be the correct fix. Will change!

pyro/distributions/torch_distribution.py
neerajprad: This was mostly just migrated as is from the existing `expand` implementation, but I think your simpler solution will work great and we don't need the more involved implementation (now)! Will change.

pyro/distributions/testing/naive_dirichlet.py
```diff
@@ -14,9 +14,9 @@ class NaiveDirichlet(Dirichlet):
     This naive implementation has stochastic reparameterized gradients, which
     have higher variance than PyTorch's ``Dirichlet`` implementation.
     """
-    def __init__(self, concentration):
+    def __init__(self, concentration, validate_args=False):
```
neerajprad: I needed to add `validate_args` since the `.expand` in the parent method ends up could end up calling this constructor with this kwarg via `type(self)(*args, validate_params=..)`

neerajprad:
This is ready for any further review.

pyro/distributions/torch_distribution.py
```diff
@@ -258,6 +252,30 @@ def __init__(self, base_dist, sample_shape=torch.Size(), reinterpreted_batch_ndi
...
+        # e.g. (2, 4) -> (2, 1)
+        proposed_shape = broadcast_shape(self.batch_shape, batch_shape)
+        if tuple(reversed(proposed_shape)) > tuple(reversed(batch_shape)):
+            raise ValueError("Existing batch shape {} cannot be expanded "
+                             "to the new batch shape {}."
+                             .format(self.batch_shape, batch_shape))
+        # Adjust existing sample shape if possible.
+        base_dist = self.base_dist
+        base_batch_shape = batch_shape + self.event_shape[:self.reinterpreted_batch_ndims]
+        cut = len(base_batch_shape) - len(base_dist.batch_shape)
```
fritzo: Is this safe in cases when `cut < 0`?

fritzo:
Do you think anyone's code will break, i.e. is anyone using `.expand(N)` for integer `N`?

pyro/distributions/testing/naive_dirichlet.py
neerajprad: Makes sense, will change.

pyro/distributions/torch_distribution.py
neerajprad: `cut` should not be less than 1. If it is, we will end up throwing a ValueError above because in that case the proposed batch shape will be less than the original batch shape. Let me modify the tests to assert that.

fritzo:
:+1: for making `.expand()` more restrictive, requiring an iterable of sizes.

neerajprad:
Thanks for reviewing, @fritzo!

PR #1188, opened 2018-06-14
add relaxed bernoulli with straight through gradient estimator

martinjankowiak:
the bernoulli version of #1124

currently  tests/distributions/test_relaxed_straight_through.py::test_bernoulli_svi_usage() is failing.
there is some sort of strange log_prob nan. i'm not sure what's happening. @fritzo can you please take
a look when you have a chance?


fritzo:
@martinjankowiak The NAN was caused by an edge sample `soft_sample=1.0`. I was able to fix this by applying `clamp_probs` to the `soft_sample`:
```diff
diff --git a/pyro/distributions/relaxed_straight_through.py b/pyro/distributions/relaxed_straight_through.py
index 853d4934..b5ff58ef 100644
--- a/pyro/distributions/relaxed_straight_through.py
+++ b/pyro/distributions/relaxed_straight_through.py
@@ -4,6 +4,7 @@ import torch
 
 from pyro.distributions.torch import RelaxedOneHotCategorical, RelaxedBernoulli
 from pyro.distributions.util import copy_docs_from
+from torch.distributions.utils import clamp_probs
 
 @copy_docs_from(RelaxedOneHotCategorical)
@@ -31,6 +32,7 @@ class RelaxedOneHotCategoricalStraightThrough(RelaxedOneHotCategorical):
 
     def rsample(self, sample_shape=torch.Size()):
         soft_sample = super(RelaxedOneHotCategoricalStraightThrough, self).rsample(sample_shape)
+        soft_sample = clamp_probs(soft_sample)
         hard_sample = QuantizeCategorical.apply(soft_sample)
         return hard_sample
 
@@ -79,6 +81,7 @@ class RelaxedBernoulliStraightThrough(RelaxedBernoulli):
 
     def rsample(self, sample_shape=torch.Size()):
         soft_sample = super(RelaxedBernoulliStraightThrough, self).rsample(sample_shape)
+        soft_sample = clamp_probs(soft_sample)
         hard_sample = QuantizeBernoulli.apply(soft_sample)
         return hard_sample
```
but tests then fail due to numerical inaccuracy.

tests/distributions/test_relaxed_straight_through.py
```diff
@@ -102,27 +102,6 @@ def test_bernoulli_entropy_grad(temp):
     z = dist_q.rsample(sample_shape=(num_samples,))
     actual = grad(dist_q.log_prob(z).sum(), [q])[0] / num_samples
 
-    assert_equal(expected, actual, prec=0.03,
+    assert_equal(expected, actual, prec=0.04,
                  msg='bad grad for RelaxedBernoulliStraightThrough (expected {}, got {})'.
                  format(expected, actual))
-
-
-def test_bernoulli_svi_usage():
```
fritzo: Why does this test fail?

martinjankowiak: it's a biased gradient estimator so there's no guarantee that it'll work properly. iguess in the categorical case biased is good enough but it can cause problems in this bernoulli setting?

fritzo: OK, I guess we'll see how it works in practice.

martinjankowiak:
ok, added to .rst

fritzo:
@martinjankowiak push if you have time, otherwise I'll tidy up docs and merge this tomorrow.

PR #1189, opened 2018-06-15
add MixtureOfDiagNormalsSharedCovariance to pyro

martinjankowiak:
Mixture of normal distributions with diagonal covariance matrices.
    This distribution supports pathwise derivatives.

    :param torch.Tensor locs: K x D mean matrix
    :param torch.Tensor scale: shared D-dimensional scale vector
    :param torch.Tensor logits: K-dimensional vector of softmax logits


pyro/distributions/diag_normal_mixture_shared_cov.py
```diff
@@ -0,0 +1,139 @@
...
+from torch.autograd import Function
+from torch.autograd.function import once_differentiable
+from torch.distributions import Categorical, constraints
+
+from pyro.distributions.torch_distribution import TorchDistribution
+from pyro.distributions.util import sum_leftmost
+
+
+class MixtureOfDiagNormalsSharedCovariance(TorchDistribution):
+    """Mixture of Normal distributions with diagonal covariance matrices. That is,
```
fritzo: nit: could you follow our standard docstring format:
```py
"""
First a one line description.

Then a paragraph of longer description that is wrapped at 80 characters
so that if fits into help() messages in Jupyter. This is easy to do in vim:
select the lines to wrap and type gq<ENTER>.

...
"""
```

pyro/distributions/diag_normal_mixture_shared_cov.py
```diff
@@ -0,0 +1,139 @@
...
+        self.dim = locs.size(-1)
+        self.categorical = Categorical(logits=logits)
+        self.probs = self.categorical.probs
+        super(MixtureOfDiagNormalsSharedCovariance, self).__init__(batch_shape=batch_shape, event_shape=(self.dim,))
+
+    def log_prob(self, value):
+        scale = self.scale.unsqueeze(-2) if self.batch_mode else self.scale
+        epsilon = (value.unsqueeze(-2) - self.locs) / scale  # L B K D
+        eps_sqr = 0.5 * torch.pow(epsilon, 2.0).sum(-1)  # L B K
+        eps_sqr_min = torch.min(eps_sqr, -1)[0]  # L B K
```
fritzo: `eps_sqr_min` is shape `L B`

tests/distributions/test_gaussian_mixtures.py
```diff
@@ -0,0 +1,108 @@
...
+
+import pytest
+from pyro.distributions import MixtureOfDiagNormalsSharedCovariance
+from tests.common import assert_equal
+
+
+@pytest.mark.parametrize('mix_dist', [MixtureOfDiagNormalsSharedCovariance])
+@pytest.mark.parametrize('K', [3])
+@pytest.mark.parametrize('D', [2, 4])
+@pytest.mark.parametrize('batch_mode', [True, False])
```
fritzo: Add a test for `len(sample_shape) > 1`, e.g. via
```py
@pytest.mark.parametrize('sample_shape', [(), (n_samples,), (2, n_samples)])
```

tests/distributions/test_gaussian_mixtures.py
```diff
@@ -0,0 +1,108 @@
...
+    if not flat_logits:
+        logits = torch.tensor(1.5 * torch.rand(K), requires_grad=True)
+    else:
+        logits = torch.tensor(0.1 * torch.rand(K), requires_grad=True)
+    omega = torch.tensor(0.2 * torch.ones(D) + 0.1 * torch.rand(D), requires_grad=False)
+
+    _pis = torch.exp(logits)
+    pis = _pis / _pis.sum()
+
+    if cost_function == 'cosine':
```
fritzo: Either remove this if statement or add `cosine` to the `@pytest.mark.parametrize` decorator

tests/distributions/test_gaussian_mixtures.py
```diff
@@ -0,0 +1,108 @@
...
+                     mix_dist.__name__, analytic.item(), cost.item()))
+    print("analytic_grads_logit", analytic_grads['logits'].detach().numpy())
+
+    for param_name, param in params.items():
+        assert_equal(param.grad, analytic_grads[param_name], prec=0.06,
+                     msg='bad {} grad for {} (expected {}, got {})'.format(
+                         param_name, mix_dist.__name__, analytic_grads[param_name], param.grad))
+
+
+@pytest.mark.parametrize('batch_size', [1, 3])
```
fritzo: ditto: strengthen this via something like
```py
@pytest.mark.parametrize('sample_shape', [(), (4,), (3,2)])
```

tests/distributions/test_gaussian_mixtures.py
```diff
@@ -0,0 +1,108 @@
...
+    assert logits.grad.shape == logits.shape
+
+    sigmas.grad.zero_()
+    logits.grad.zero_()
+    locs.grad.zero_()
+
+    if mix_dist == MixtureOfDiagNormalsSharedCovariance:
+        params = {'locs': locs, 'scale': sigmas, 'logits': logits}
+        if batch_mode:
+            locs = locs.unsqueeze(0).expand(n_samples, K, D)
```
fritzo: I don't think you need these `.unsqeeze(0)`s. If you generalize from `batch_mode` to `sample_shape` you should be able to unconditionally
```py
locs = locs.expand(sample_shape + (K, D))
sigmas = sigmas.expand(sample_shape + (D,))
logits = logits.expand(sample_shape + (K,))
```

pyro/distributions/diag_normal_mixture_shared_cov.py
```diff
@@ -0,0 +1,139 @@
...
+        return _MixDiagNormalSharedCovarianceSample.apply(self.locs, self.scale, self.logits, self.probs,
+                                                          which, sample_shape + self.scale.shape)
+
+
+class _MixDiagNormalSharedCovarianceSample(Function):
+    @staticmethod
+    def forward(ctx, locs, scale, logits, pis, which, noise_shape):
+        dim = scale.size(-1)
+        white = locs.new(noise_shape).normal_()
+        n_unsqueezes = locs.dim() - which.dim()
```
fritzo: Can you explain `n_unsqueezes` with a comment? I'm having trouble seeing why it should ever be anything but 1. In particular, if `len(sample_shape) > 0`, I still think you should `.unsqueeze(-1)` exactly once, and additionally `locs.expand(which.shape + (dim,))`.

tests/distributions/test_gaussian_mixtures.py
martinjankowiak: these tests are expensiveish can only run so many variants in CI

pyro/distributions/diag_normal_mixture_shared_cov.py
martinjankowiak: done

pyro/distributions/diag_normal_mixture_shared_cov.py
martinjankowiak: done

tests/distributions/test_gaussian_mixtures.py
martinjankowiak: i've run these tests before but think it's too much for travis

tests/distributions/test_gaussian_mixtures.py
martinjankowiak: why can't i keep an unused test around so that it's available during future refactoring? it's only a few lines of code

fritzo: If it's not tested it will break.

tests/distributions/test_gaussian_mixtures.py
martinjankowiak: i don't think i have time to reason carefully about these kinds of proposed changes before i leave

pyro/distributions/diag_normal_mixture_shared_cov.py
martinjankowiak: i can't recall exactly but i needed this to get gather to work properly and produce correct samples. 

if you run the test you will find that `n_squeezes` takes values 2/1 for `batch_mode` = True/False

tests/distributions/test_gaussian_mixtures.py
martinjankowiak: it is tested. with another cost function.

tests/distributions/test_gaussian_mixtures.py
fritzo: Ok, I'll take that to me that you volunteer to maintain this complex test as we refactor and update due to PyTorch changes.

tests/distributions/test_gaussian_mixtures.py
fritzo: Ok, I you volunteer to be assigned to any bugs due to broadcasting errors.

martinjankowiak: yes i volunteer

tests/distributions/test_gaussian_mixtures.py
martinjankowiak: yes i volunteer

PR #1192, opened 2018-06-18
Add eight schools example

irustandi:
Add eight schools example from the book Bayesian Data Analysis, using MCMC and SVI.


examples/eight_schools/main_mcmc.py
```diff
@@ -0,0 +1,58 @@
...
+import pyro
+import pyro.distributions as dist
+import pyro.poutine as poutine
+from pyro.infer import EmpiricalMarginal
+from pyro.infer.mcmc import MCMC, NUTS
+
+
+def model(sigma):
+    eta_dist = dist.Normal(torch.zeros(data.J), torch.ones(data.J))
+    mu_dist = dist.Normal(loc=torch.zeros(1), scale=10 * torch.ones(1))
```
jpchen: nit: you should use a consistent format for parameterizing these distributions. i recommend using positional args to be consistent with the other examples

examples/eight_schools/main_mcmc.py
```diff
@@ -0,0 +1,58 @@
...
+def conditioned_model(model, sigma, y):
+    return poutine.condition(model, data={"obs": y})(sigma)
+
+
+logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
+
+parser = argparse.ArgumentParser(description='Eight Schools MCMC')
+parser.add_argument('--num-samples', type=int, default=1000, metavar='NUMSAMPLES',
+                    help='number of MCMC samples (default: 1000)')
+parser.add_argument('--warmup-steps', type=int, default=1000, metavar='WARMUPSTEPS',
```
jpchen: i recommend putting all of this in a `main()` method, see [vae](https://github.com/uber/pyro/blob/dev/examples/vae/vae.py#L197) for example

examples/eight_schools/main_svi.py
```diff
@@ -0,0 +1,84 @@
...
+data = torch.stack([y_tensor, sigma_tensor], dim=1)
+
+
+def model(data):
+    y = data[:, 0]
+    sigma = data[:, 1]
+
+    eta_dist = dist.Normal(torch.zeros(J), torch.ones(J))
+    mu_dist = dist.Normal(loc=torch.zeros(1), scale=10 * torch.ones(1))
+    tau_dist = dist.HalfCauchy(loc=torch.zeros(1), scale=25 * torch.ones(1))
```
jpchen: same comment as above

examples/eight_schools/main_svi.py
```diff
@@ -0,0 +1,84 @@
...
+
+    theta = mu + tau * eta
+
+    pyro.sample("obs", dist.Normal(theta, sigma), obs=y)
+
+
+def guide(data):
+    eta_loc = torch.randn(J, 1)
+    # note that we initialize our scales to be pretty narrow
+    eta_log_sig = torch.tensor(-3.0 * torch.ones(J, 1) + 0.05 * torch.randn(J, 1))
```
jpchen: redundant `torch.tensor()`

examples/eight_schools/main_svi.py
```diff
@@ -0,0 +1,84 @@
...
+    mu_loc = torch.randn(1)
+    mu_log_sig = torch.tensor(-3.0 * torch.ones(1) + 0.05 * torch.randn(1))
+    logtau_loc = torch.randn(1)
+    logtau_log_sig = torch.tensor(-3.0 * torch.ones(1) + 0.05 * torch.randn(1))
```
jpchen: same comment as above

examples/eight_schools/main_svi.py
```diff
@@ -0,0 +1,84 @@
...
+    pyro.sample('mu', mu_dist)
+    pyro.sample('tau', tau_dist)
+
+
+logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
+
+parser = argparse.ArgumentParser(description='Eight Schools MCMC')
+parser.add_argument('--lr', type=float, default=0.01, metavar='LR',
+                    help='learning rate (default: 0.01)')
+parser.add_argument('--num-epochs', type=int, default=1000, metavar='NUMEPOCHS',
```
jpchen: same comment as above, can you wrap this in a `main()` method

examples/eight_schools/main_mcmc.py
```diff
@@ -0,0 +1,58 @@
...
+marginal_mu_tau = EmpiricalMarginal(posterior, sites=["mu", "tau"])\
+    .get_samples_and_weights()[0].squeeze().numpy()
+marginal_eta = EmpiricalMarginal(posterior, sites=["eta"])\
+    .get_samples_and_weights()[0].squeeze().numpy()
+marginal = np.concatenate([marginal_mu_tau, marginal_eta], axis=1)
+params = ['mu', 'tau', 'eta[0]', 'eta[1]', 'eta[2]', 'eta[3]', 'eta[4]', 'eta[5]', 'eta[6]', 'eta[7]']
+df = pd.DataFrame(marginal, columns=params).transpose()
+df_summary = df.apply(pd.Series.describe, axis=1)[["mean", "std", "25%", "50%", "75%"]]
+
+print(df_summary)
```
neerajprad: Could you change this to use the `logging` module instead, so that this plays nicely with the logger stdout/stderr from the `HMC` module?

examples/eight_schools/main_mcmc.py
```diff
@@ -0,0 +1,58 @@
...
+
+nuts_kernel = NUTS(conditioned_model, adapt_step_size=True)
+posterior = MCMC(nuts_kernel, num_samples=args.num_samples, warmup_steps=args.warmup_steps)\
+    .run(model, data.sigma_tensor, data.y_tensor)
+marginal_mu_tau = EmpiricalMarginal(posterior, sites=["mu", "tau"])\
+    .get_samples_and_weights()[0].squeeze().numpy()
+marginal_eta = EmpiricalMarginal(posterior, sites=["eta"])\
+    .get_samples_and_weights()[0].squeeze().numpy()
+marginal = np.concatenate([marginal_mu_tau, marginal_eta], axis=1)
+params = ['mu', 'tau', 'eta[0]', 'eta[1]', 'eta[2]', 'eta[3]', 'eta[4]', 'eta[5]', 'eta[6]', 'eta[7]']
```
neerajprad: It will be more interesting to see the predictive distribution of the treatment effects for the 8 schools using `TracePredictive` - check out `sample_posterior_predictive` function in `baseball.py`.

examples/eight_schools/main_mcmc.py
```diff
@@ -0,0 +1,58 @@
...
+import pandas as pd
+import torch
+
+import data
+import pyro
+import pyro.distributions as dist
+import pyro.poutine as poutine
+from pyro.infer import EmpiricalMarginal
+from pyro.infer.mcmc import MCMC, NUTS
+
```
neerajprad: Please add the following lines so as to enable validation and fix the RNG seed for deterministic results.

```
pyro.enable_validation(True)
pyro.set_rng_seed(0)
```

examples/eight_schools/main_mcmc.py
```diff
@@ -0,0 +1,58 @@
...
+    theta = mu + tau * eta
+
+    return pyro.sample("obs", dist.Normal(theta, sigma))
+
+
+def conditioned_model(model, sigma, y):
+    return poutine.condition(model, data={"obs": y})(sigma)
+
+
+logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
```
neerajprad: You could just modify the format to `%(message)s` and use this instead of the `print` function. You shouldn't have to force the stream to `sys.stdout` which should be the case by default.

examples/eight_schools/main_mcmc.py
```diff
@@ -0,0 +1,58 @@
...
+
+parser = argparse.ArgumentParser(description='Eight Schools MCMC')
+parser.add_argument('--num-samples', type=int, default=1000, metavar='NUMSAMPLES',
```
neerajprad: nit: remove `metavar`, I think the default `NUM_SAMPLES` or `WARMUP_STEPS` is sensible.

examples/eight_schools/main_svi.py
neerajprad: I would also suggest moving these into the `sample` statements to make it consistent with the `mcmc` model.

examples/eight_schools/data.py
```diff
@@ -0,0 +1,11 @@
+from __future__ import absolute_import, division, print_function
+
+import numpy as np
+import torch
+
+J = 8
+y = np.array([28,  8, -3,  7, -1,  1, 18, 12])
```
neerajprad: Why not just torch.tensor([28., ...])?

examples/eight_schools/main_svi.py
```diff
@@ -0,0 +1,84 @@
...
+    mu_loc = torch.randn(1)
+    mu_log_sig = torch.tensor(-3.0 * torch.ones(1) + 0.05 * torch.randn(1))
+    logtau_loc = torch.randn(1)
+    logtau_log_sig = torch.tensor(-3.0 * torch.ones(1) + 0.05 * torch.randn(1))
+
+    # register learnable params in the param store
+    m_eta_param = pyro.param("guide_mean_eta", eta_loc)
+    s_eta_param = softplus(pyro.param("guide_log_scale_eta", eta_log_sig))
+    m_mu_param = pyro.param("guide_mean_mu", mu_loc)
+    s_mu_param = softplus(pyro.param("guide_log_scale_mu", mu_log_sig))
```
neerajprad: nit: remove redundant `guide` from the site names.

examples/eight_schools/main_svi.py
```diff
@@ -0,0 +1,84 @@
...
+    eta_loc = torch.randn(J, 1)
+    # note that we initialize our scales to be pretty narrow
+    eta_log_sig = torch.tensor(-3.0 * torch.ones(J, 1) + 0.05 * torch.randn(J, 1))
+    mu_loc = torch.randn(1)
+    mu_log_sig = torch.tensor(-3.0 * torch.ones(1) + 0.05 * torch.randn(1))
+    logtau_loc = torch.randn(1)
+    logtau_log_sig = torch.tensor(-3.0 * torch.ones(1) + 0.05 * torch.randn(1))
+
+    # register learnable params in the param store
+    m_eta_param = pyro.param("guide_mean_eta", eta_loc)
```
neerajprad: nit: would suggest changing the names to `loc_eta`, `scale_eta`, etc. for readability.

examples/eight_schools/main_svi.py
```diff
@@ -0,0 +1,84 @@
+from __future__ import absolute_import, division, print_function
```
neerajprad: nit: remove `main` from the filenames.

examples/eight_schools/main_svi.py
```diff
@@ -0,0 +1,84 @@
...
+    # note that we initialize our scales to be pretty narrow
+    eta_log_sig = torch.tensor(-3.0 * torch.ones(J, 1) + 0.05 * torch.randn(J, 1))
+    mu_loc = torch.randn(1)
+    mu_log_sig = torch.tensor(-3.0 * torch.ones(1) + 0.05 * torch.randn(1))
+    logtau_loc = torch.randn(1)
+    logtau_log_sig = torch.tensor(-3.0 * torch.ones(1) + 0.05 * torch.randn(1))
+
+    # register learnable params in the param store
+    m_eta_param = pyro.param("guide_mean_eta", eta_loc)
+    s_eta_param = softplus(pyro.param("guide_log_scale_eta", eta_log_sig))
```
neerajprad: As an alternative to `softmax`, a more idiomatic way of constraining the param to positive would be to use the constraint keyword (you will need to change the initialization accordingly).

```
pyro.param("guide_log_scale_logtau", logtau_log_sig, constraint=constraints.positive)
```

examples/eight_schools/main_svi.py
irustandi: @neerajprad Could you clarify what you meant by moving these into the `sample` statements? Thanks!

jpchen: @irustandi 
```diff
- eta_dist = dist.Normal(...)
- eta = pyro.sample('eta', eta_dist)
+ pyro.sample('eta', dist.Normal(...)
```

examples/eight_schools/svi.py
```diff
@@ -0,0 +1,84 @@
...
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Eight Schools SVI')
+    parser.add_argument('--lr', type=float, default=0.01,
+                        help='learning rate (default: 0.01)')
+    parser.add_argument('--num-epochs', type=int, default=1000,
+                        help='number of epochs (default: 1000)')
+    args = parser.parse_args()
+
+    main(args)
```
eb8680: Add newline here

irustandi:
Incorporated most of the reviews. I'll leave the posterior predictive suggestions for later. Hope that's OK.

PR #1194, opened 2018-06-19
Implement compute_marginals_persistent_bp()

fritzo:
Addresses #1185 

This implements loopy belief propagation for multi-frame multi-object data association.

## Tests

The following tests pass:

- [x] smoke tests
- [x] tested two different BP algorithms against each other in `test_flat_vs_persistent`. they should agree up to numerical precision.
- [x] tested exact vs BP. they agree approximately, to within 1%


fritzo:
Thanks @jpchen and @alicanb for your help with the math!

> a bit worried about the logsumexps

I agree. I expect we'll need to add stabilizing tricks as this code matures.

PR #1197, opened 2018-06-20
Fix CUDA tests on Pyro dev

neerajprad:


tests/contrib/tracking/test_assignment.py
```diff
@@ -66,10 +67,10 @@ def test_sparse_smoke():
     num_detections = 2
     pyro.set_rng_seed(0)
     exists_logits = torch.zeros(num_objects)
-    edges = torch.LongTensor([
+    edges = exists_logits.new_tensor([
         [0, 0, 1, 0, 1, 0],
         [0, 1, 1, 2, 2, 3],
-    ])
+    ], dtype=torch.long)
```
jpchen: (just curious) is there a difference between this and `.long()` for cpu/gpu compatability?

fritzo: I believe they result in the same behavior, but `dtype=torch.long` avoids creating an intermediate float tensor.

PR #1199, opened 2018-06-21
Add momentum to compute_marginals_persistent_bp()

fritzo:
This also sets momentum = 0.5 by default.

## Tested

- strengthened existing test parameter ranges
- added an exact test on a tree-shaped graph
- tested inside tracking_1d.ipynb; momentum successfully avoided oscillation

### DIFF
--- a/pyro/contrib/tracking/assignment.py
+++ b/pyro/contrib/tracking/assignment.py
@@ -156,6 +156,8 @@ class MarginalAssignmentPersistent(object):
         given time frame a given detection associates with a single object.
     :param int bp_iters: optional number of belief propagation iterations. If
         unspecified or ``None`` an expensive exact algorithm will be used.
+    :param float bp_momentum: optional momentum to use for belief propagation.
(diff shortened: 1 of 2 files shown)


pyro/contrib/tracking/assignment.py
```diff
@@ -168,7 +170,7 @@ class MarginalAssignmentPersistent(object):
         final element denotes spurious detection, and
         ``.batch_shape == (num_frames, num_detections)``.
     """
-    def __init__(self, exists_logits, assign_logits, bp_iters=None):
+    def __init__(self, exists_logits, assign_logits, bp_iters=None, bp_momentum=0.5):
         assert exists_logits.dim() == 1, exists_logits.shape
         assert assign_logits.dim() == 3, assign_logits.shape
         assert assign_logits.shape[-1] == exists_logits.shape[-1]
```
alicanb: maybe an assert for bp_momentum here?

fritzo:
Thanks for reviewing!

PR #1202, opened 2018-06-23
Refactor `warn_if_nan`

alicanb:
`warn_if_nan` was multiply defined across the codebase. This PR refactors them in single function in `pyro.util`. One significant functional change is `torch_isinf` and `torch.isnan` returns true even if a single element in tensor is inf or nan.

### DIFF
--- a/pyro/util.py
+++ b/pyro/util.py
@@ -30,20 +30,34 @@ def set_rng_seed(rng_seed):
 
 def torch_isnan(x):
     """
-    A convenient function to check if a Tensor contains all nan; also works with numbers
+    A convenient function to check if a Tensor contains any nan; also works with numbers
     """
     if isinstance(x, numbers.Number):
         return x != x
-    return torch.isnan(x).all()
+    return torch.isnan(x).any()
 
 
 def torch_isinf(x):
     """
-    A convenient function to check if a Tensor contains all inf; also works with numbers
+    A convenient function to check if a Tensor contains any inf; also works with numbers
     """
     if isinstance(x, numbers.Number):
         return x == float('inf')
-    return (x == float('inf')).all()
+    return (x == float('inf')).any()
+
+
+def warn_if_nan(name, value):
+    """
+    A convenient function to warn if a Tensor or its grad contains any nan or inf,
+    also works with numbers.
+    """
+    if torch.is_tensor(value) and value.requires_grad:
+            value.register_hook(lambda x: warn_if_nan(name, x))
+    if torch_isnan(value):
+        warnings.warn("Encountered NAN in {}".format(name))
+    if torch_isinf(value) and value > 0:
+        warnings.warn("Encountered +inf in {}".format(name))
+    # Note that -inf log_prob_sum is fine: it is merely a zero-probability event.
 
 
 def save_visualization(trace, graph_output):
--- a/pyro/util.py
+++ b/pyro/util.py
@@ -39,25 +39,37 @@ def torch_isnan(x):
 
 def torch_isinf(x):
     """
-    A convenient function to check if a Tensor contains any inf; also works with numbers
+    A convenient function to check if a Tensor contains any +inf; also works with numbers
     """
     if isinstance(x, numbers.Number):
-        return x == float('inf')
-    return (x == float('inf')).any()
+        return x == float('inf') or x == -float('inf')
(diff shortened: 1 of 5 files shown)


pyro/util.py
```diff
@@ -30,20 +30,34 @@ def set_rng_seed(rng_seed):
...
+    return (x == float('inf')).any()
+
+
+def warn_if_nan(name, value):
+    """
+    A convenient function to warn if a Tensor or its grad contains any nan or inf,
+    also works with numbers.
+    """
+    if torch.is_tensor(value) and value.requires_grad:
+            value.register_hook(lambda x: warn_if_nan(name, x))
```
neerajprad: Nice to have this warning on NaN gradients. Should have 4 spaces here; otherwise this will fail the lint check.

pyro/util.py
```diff
@@ -30,20 +30,34 @@ def set_rng_seed(rng_seed):
...
-    A convenient function to check if a Tensor contains all inf; also works with numbers
+    A convenient function to check if a Tensor contains any inf; also works with numbers
     """
     if isinstance(x, numbers.Number):
         return x == float('inf')
-    return (x == float('inf')).all()
+    return (x == float('inf')).any()
+
+
+def warn_if_nan(name, value):
```
neerajprad: If we have use cases where it is fine to have `+inf` values but not `NaN`s, then it would make sense to have an optional kwarg `warn_if_posinf=True`, so that optionally turn off the check on `+inf` values. That said, if you haven't found any code / failing tests yet, we can address this later.

neerajprad: What do you think about changing `name` to an optional but more general `msg=""` kwarg, which could be specified as `msg="log_prob_sum at site {}".format(name)`, or `msg="site: {}".format(name)`, and used as follows below:

```
warnings.warn("Encountered inf. " + msg)
```

pyro/util.py
```diff
@@ -30,20 +30,34 @@ def set_rng_seed(rng_seed):
...
+    """
+    A convenient function to warn if a Tensor or its grad contains any nan or inf,
+    also works with numbers.
+    """
+    if torch.is_tensor(value) and value.requires_grad:
+            value.register_hook(lambda x: warn_if_nan(name, x))
+    if torch_isnan(value):
+        warnings.warn("Encountered NAN in {}".format(name))
+    if torch_isinf(value) and value > 0:
```
fritzo: `warn_if_nan()` should not warn if inf. Instead, create a separate `warn_if_inf()` method. In particular, most users of `warn_if_nan()` will want to allow `-inf` and many will want to allow `inf`.

pyro/util.py
```diff
@@ -30,20 +30,34 @@ def set_rng_seed(rng_seed):
...
+
+def warn_if_nan(name, value):
+    """
+    A convenient function to warn if a Tensor or its grad contains any nan or inf,
+    also works with numbers.
+    """
+    if torch.is_tensor(value) and value.requires_grad:
+            value.register_hook(lambda x: warn_if_nan(name, x))
+    if torch_isnan(value):
+        warnings.warn("Encountered NAN in {}".format(name))
```
jpchen: is this actually being tested? where is `warnings` being imported?

alicanb:
Hey guys, I addressed most of the reviews:
- new method: `warn_if_inf(value, msg, allow_neg_inf,allow_posinf)`
- `torch_isinf` checks for both +inf and -inf
- added `test_util.py` to test suite (only checks `warn_if_inf` and `warn_if_nan` for now)

pyro/util.py
```diff
@@ -30,20 +30,46 @@ def set_rng_seed(rng_seed):
...
+    if torch_isnan(value):
+        warnings.warn("Encountered NaN{}".format((': ' if msg else '.') + msg))
+
+
+def warn_if_inf(value, msg="", allow_posinf=False, allow_neginf=False):
+    """
+    A convenient function to warn if a Tensor or its grad contains any inf,
+    also works with numbers.
+    """
+    if torch.is_tensor(value) and value.requires_grad:
```
jpchen: nit: to prevent needless memory allocation you can `return` or throw an error if both `allow_posinf` and `allow_neginf` are true

fritzo:
Thanks for cleaning this up, @alicanb!

PR #1203, opened 2018-06-23
Welford's Scheme Implementation

LoganWalls:
This implements [Welford's online scheme](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Online_algorithm) for calculating (co)variance.

_Note: This is one component of mass matrix adaptation for the HMC and NUTS samplers (The original PR: #1178 is being broken into multiple smaller PRs)._

### Usage
```python
from pyro.ops.welford import WelfordCovariance

# Diagonal can be set to True to save resources if the
# user is only interested in the diagonal.
w = WelfordCovariance(diagonal=False)

# Each sample should be a list of numbers or tensors,
# where each index in the list corresponds to a variable.
samples = [
    [var1_0, var2_0, var3_0],
    [var1_1, var2_1, var3_1],
    [var1_1, var2_2, var3_2]]
for s in samples:
    w.update(s)

# The co-variance matrix is regularized towards the identity matrix by default
# but this can be disabled by passing regualrize=False
covariance_matrix = w.get_estimates(regularize=True)
```

### Tests
 - A couple tests using random dummy data and comparing the answer against `pytorch` / `numpy` co-variance calculations.
 - A test to ensure an error is raised if `get_estimates()` is called when fewer than 2 samples have been observed.

(Sorry this took so long everyone! I meant to get it done sooner but alas, life gets in the way sometimes. I think I've incorporated all of the changes suggested in #1178 but please let me know if I missed any.)

### DIFF
--- a/pyro/ops/welford.py
+++ b/pyro/ops/welford.py
@@ -17,9 +17,9 @@ class WelfordCovariance(object):
         self.diagonal = diagonal
         self.reset()
 
-    def _deltas(self, z):
+    def _deltas(self, sample):
         deltas = []
-        for i, x in enumerate(z):
+        for i, x in enumerate(sample):
             if i > len(self.means) - 1:
                 self.means.append(0.)
             mean = self.means[i]
@@ -29,9 +29,9 @@ class WelfordCovariance(object):
             deltas.append((updated_mean, delta_pre, delta_post))
         return deltas
 
-    def _unroll(self, z):
+    def _unroll(self, sample):
         unrolled = []
-        for rolled in z:
+        for rolled in sample:
             unrolled.extend(rolled.reshape(-1).data.tolist())
         return unrolled
 
@@ -40,9 +40,9 @@ class WelfordCovariance(object):
         self.variances = defaultdict(float)
         self.n_samples = 0
 
-    def update(self, z):
+    def update(self, sample):
         self.n_samples += 1
-        deltas = self._deltas(self._unroll(z))
+        deltas = self._deltas(self._unroll(sample))
         for i, (mean_x, delta_x_pre, delta_x_post) in enumerate(deltas):
             self.means[i] = mean_x
             for j, (mean_y, delta_y_pre, delta_y_post) in enumerate(deltas):
--- a/pyro/ops/welford.py
+++ b/pyro/ops/welford.py
@@ -22,25 +22,25 @@ class WelfordCovariance(object):
         return unrolled
 
     def reset(self):
-        self.means = 0.
-        self.variances = 0.
+        self._mean = 0.
+        self._m2 = 0.
         self.n_samples = 0
 
     def update(self, sample):
         self.n_samples += 1
-        delta_pre = sample - self.means
-        self.means = self.means + delta_pre / self.n_samples
-        delta_post = sample - self.means
+        delta_pre = sample - self._mean
+        self._mean = self._mean + delta_pre / self.n_samples
+        delta_post = sample - self._mean
 
         if self.diagonal:
-            self.variances += delta_pre * delta_post
+            self._m2 += delta_pre * delta_post
         else:
-            self.variances += delta_pre * delta_post.reshape(-1, 1)
+            self._m2 += delta_pre * delta_post.reshape(-1, 1)
 
-    def get_estimates(self, regularize=True):
+    def get_covariance(self, regularize=True):
         if self.n_samples < 2:
             raise RuntimeError('Insufficient samples to estimate covariance')
-        cov = self.variances / (self.n_samples - 1)
+        cov = self._m2 / (self.n_samples - 1)
         if regularize:
             # Regularization from stan
             scaled_cov = (self.n_samples / (self.n_samples + 5.)) * cov
--- a/pyro/ops/welford.py
+++ b/pyro/ops/welford.py
@@ -1,5 +1,3 @@
-from collections import defaultdict
-
 import torch
 
 
@@ -17,18 +15,6 @@ class WelfordCovariance(object):
         self.diagonal = diagonal
         self.reset()
 
-    def _deltas(self, sample):
-        deltas = []
-        for i, x in enumerate(sample):
-            if i > len(self.means) - 1:
-                self.means.append(0.)
-            mean = self.means[i]
-            delta_pre = x - mean
-            updated_mean = mean + delta_pre / self.n_samples
-            delta_post = x - updated_mean
-            deltas.append((updated_mean, delta_pre, delta_post))
-        return deltas
-
     def _unroll(self, sample):
         unrolled = []
         for rolled in sample:
@@ -36,35 +22,31 @@ class WelfordCovariance(object):
         return unrolled
 
     def reset(self):
-        self.means = []
-        self.variances = defaultdict(float)
+        self.means = 0.
+        self.variances = 0.
         self.n_samples = 0
 
     def update(self, sample):
         self.n_samples += 1
-        deltas = self._deltas(self._unroll(sample))
-        for i, (mean_x, delta_x_pre, delta_x_post) in enumerate(deltas):
-            self.means[i] = mean_x
-            for j, (mean_y, delta_y_pre, delta_y_post) in enumerate(deltas):
-                # Only compute the upper triangular covariance.
-                if i == j or (i < j and not self.diagonal):
-                    self.variances[(i, j)] += delta_x_pre * delta_y_post
+        delta_pre = sample - self.means
+        self.means = self.means + delta_pre / self.n_samples
+        delta_post = sample - self.means
+
+        if self.diagonal:
+            self.variances += delta_pre * delta_post
+        else:
+            self.variances += delta_pre * delta_post.reshape(-1, 1)
 
     def get_estimates(self, regularize=True):
         if self.n_samples < 2:
             raise RuntimeError('Insufficient samples to estimate covariance')
-        rows = []
-        for i in range(len(self.means)):
-            row = []
-            for j in range(len(self.means)):
-                if i == j or not self.diagonal:
-                    key = (i, j) if i <= j else (j, i)
-                    estimate = self.variances[key] / (self.n_samples - 1)
-                    if regularize:
-                        # Regularization from stan
-                        scaled_estimate = (self.n_samples / (self.n_samples + 5.)) * estimate
-                        shrinkage = 1e-3 * (5. / (self.n_samples + 5.0)) if i == j else 0.
-                        estimate = scaled_estimate + shrinkage
-                    row.append(estimate)
-            rows.append(row)
-        return torch.tensor(rows)
+        cov = self.variances / (self.n_samples - 1)
+        if regularize:
+            # Regularization from stan
+            scaled_cov = (self.n_samples / (self.n_samples + 5.)) * cov
+            shrinkage = 1e-3 * (5. / (self.n_samples + 5.0))
+            if self.diagonal:
+                cov = scaled_cov + shrinkage
+            else:
+                cov = scaled_cov + torch.diagflat(shrinkage)
+        return cov
--- a/pyro/ops/welford.py
+++ b/pyro/ops/welford.py
@@ -15,12 +15,6 @@ class WelfordCovariance(object):
         self.diagonal = diagonal
         self.reset()
 
-    def _unroll(self, sample):
-        unrolled = []
-        for rolled in sample:
-            unrolled.extend(rolled.reshape(-1).data.tolist())
-        return unrolled
-
     def reset(self):
         self._mean = 0.
         self._m2 = 0.
--- a/pyro/ops/welford.py
+++ b/pyro/ops/welford.py
@@ -1,0 +1,70 @@
+from collections import defaultdict
+
+import torch
+
+
+class WelfordCovariance(object):
+    """
+    Implements Welford's online scheme for estimating (co)variance (see :math:`[1]`).
+    Useful for adapting diagonal and dense mass structures for HMC.
+
+    **References**
+
+    [1] `The Art of Computer Programming`,
+    Donald E. Knuth
+    """
+    def __init__(self, diagonal=True):
+        self.diagonal = diagonal
+        self.reset()
+
+    def _deltas(self, z):
+        deltas = []
+        for i, x in enumerate(z):
+            if i > len(self.means) - 1:
+                self.means.append(0.)
+            mean = self.means[i]
+            delta_pre = x - mean
+            updated_mean = mean + delta_pre / self.n_samples
+            delta_post = x - updated_mean
+            deltas.append((updated_mean, delta_pre, delta_post))
+        return deltas
+
+    def _unroll(self, z):
+        unrolled = []
+        for rolled in z:
+            unrolled.extend(rolled.reshape(-1).data.tolist())
+        return unrolled
+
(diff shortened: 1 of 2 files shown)


fehiepsi:
Thanks a lot @LoganWalls ! I have a question:
Did you save all the past calculations? I thought that Welford is an online scheme, so we just need to update `n_samples`, `_mean` (we don't need to expose `mean` and `variances` I guess), `_M2` in the method `update()`. The way we update covariance is quite similar to the way wikipedia updated `variance`, just replace `M2 = M2 + delta * delta2` by `M2 = M2 + delta.unsqueeze(0).matmul(delta2.unsqueeze(1))`.

LoganWalls:
@fehiepsi 
> Did you save all the past calculations? I thought that Welford is an online scheme, so we just need to update `n_samples`,` _mean` (we don't need to expose `mean` and `variances` I guess), `_M2` in the method `update()`. 

- No, the only thing saved are the means and the M2s (which are stored in the `variances` dictionary).
- You're probably right, we can make `mean` and  `variances` private.

> The way we update covariance is quite similar to the way wikipedia updated `variance`, just replace `M2 = M2 + delta * delta2` by `M2 = M2 + delta.unsqueeze(-1).matmul(delta2.unsqueeze(-2))`.

I'm sorry, I'm not sure I understand. The reason the current version of `update()` is working with scalars is that `_unroll()` unrolls all of the variables in `sample` into one long concatenated tensor, so `delta_x` and `delta_y` are always scalars.

**Current Behavior Ex.**
```
[ [1,2,3],  # Variable A, shape = (3,)
  [[4,5,6], # Variable B, shape= (3,2)
   [7,8,9]] ]
```
becomes
```
[1,2,3,4,5,6,7,8,9] # Unrolled, shape=(9,)
```
---
Are you suggesting I modify the unrolling so that is unrolls each variable in the sample separately?

**Ex.**
```
[ [1,2,3],  # Variable A, shape = (3,)
  [[4,5,6], # Variable B, shape= (3,2)
   [7,8,9]] ]
```
becomes
```
# Unrolled, (list)
[ [1,2,3],  # Variable A, shape = (3,)
  [4,5,6.7,8,9] ] # Variable B, shape= (6,)
```

fehiepsi:
@LoganWalls I think it is better to unroll each variable separately because we don't calculate their interactions. Am I right? In addition, I don't quite understand what `variances[(i, j)]` means. Could you elaborate on it a little bit more, please?

To me, it is better to have `sample` is a dictionary with variable names as keys (though `list` is also fine to me). The update method in my mind is something like this:
```
def update(self, sample):
    self.n_samples += 1
    if self.n_samples == 1:
         # initiate self._mean, self._M2
    delta_pre = {name: sample[name] - self._mean[name] for name in sample}
    for name in sample:
        self._mean[name] = self._mean[name] + delta_pre / self.n_samples()
    delta_pos = {name: sample[name] - self._mean[name] for name in sample}
    for name in sample:
        self._M2[name] = self._M2[name] + torch.matmul(delta_pos[name].reshape(-1, 1), delta_pre[name].reshape(1, -1))
```

pyro/ops/welford.py
```diff
@@ -0,0 +1,70 @@
...
+        self.n_samples += 1
+        deltas = self._deltas(self._unroll(sample))
+        for i, (mean_x, delta_x_pre, delta_x_post) in enumerate(deltas):
+            self.means[i] = mean_x
+            for j, (mean_y, delta_y_pre, delta_y_post) in enumerate(deltas):
+                # Only compute the upper triangular covariance.
+                if i == j or (i < j and not self.diagonal):
+                    self.variances[(i, j)] += delta_x_pre * delta_y_post
+
+    def get_estimates(self, regularize=True):
```
neerajprad: Can we rename this to something more explicit, like `get_covariance`? We can also similarly expose a `get_mean` functionality later.

fehiepsi:
Ah, so I guess what I have understood was wrong. Let's me review things before giving a comment. :ear:

fehiepsi:
@LoganWalls @neerajprad I need some discussions on this. Please let me know what you think.

I guess you all want to get a big "mass matrix" for unrolled+joined latent variables. Then I wonder how we can do integrator steps with that "mass matrix"? Will we unroll+joined "z" and "r" to get the updated "z_next" and "r_next"? If that is the case then I propose to unroll all at the starting point and applied Welford in a vectorized way, rather than concatenating things in the current implementation. For example, we should use `torch.cat([param.view(-1) for param in sample])` to get unrolled vector instead of storing things in a list, then calculating each entry, then concatenate.

However, I think that using big "mass matrix" just makes things complicated. Currently, we do integrator step for each latent parameter separately. That suggests that we didn't care about their covariance. We can still use dense scale mass matrix for each parameter, which means that we only care about the covariance of dimensions of each parameter.

In Neal's paper, the author only discussed using "diagonal" matrix for hierarchy models (and using "dense" for a parameter). For Stan and PyMC, I don't know what they did but I guess they unroll all at first place. Anyway, "diagonal" is the default ones in both PyMC and Stan.

neerajprad:
You bring up some good points, @fehiepsi, that are worth discussing in detail.

> Then I wonder how we can do integrator steps with that "mass matrix"? Will we unroll+joined "z" and "r" to get the updated "z_next" and "r_next"? If that is the case then I propose to unroll all at the starting point and applied Welford in a vectorized way, rather than concatenating things in the current implementation. 

In terms of implementation, (I'll think a bit more on this), we'll need to unroll the variables during the computation of the kinetic energy and incorporate the mass matrix. Beyond that, I don't think we should need any changes to the integrator. That said, there are some other advantages to using a list of vectors for `q` and `r` instead of a dict of tensors, particularly when using JIT (which doesn't support dicts). The disadvantage is reduced visibility that makes it difficult to see what's happening behind the scenes. We should discuss that separately though. 

> I think that using big "mass matrix" just makes things complicated. Currently, we do integrator step for each latent parameter separately. That suggests that we didn't care about their covariance. We can still use dense scale mass matrix for each parameter, which means that we only care about the covariance of dimensions of each parameter.

The biggest drawback will be in having to do big dense matrix multiplies every time we compute the kinetic energy. But as you rightly mention, this should not be the default mode. I think having a "diagonal" sparse mass matrix should be the default. We can go a step further and have a sparse "block diagonal" mass matrix which takes into account the covariance of the dimensions within each sample site, and finally a "fully dense" matrix which may not be suitable for bigger problems. Anyways, I think the important thing is that we should be able to support all of these modes without necessarily paying a penalty for dense mass matrix multiplications when we don't need to, e.g. when we have a diagonal mass matrix. 

So the important question is - what are the changes we need to make to this PR so that we take a minimal runtime hit when we use a "diagonal" mass matrix? I think the important thing there is to have a fast Welford "update" step, and to not do full matrix multiplies during kinetic energy computation. Let me know what you guys think. The remaining optimizations (e.g. using unrolled tensors for `q`, `p` for the "full dense" mode, or having an intermediate more efficient "block diagonal" implementation) can be handled separately.

pyro/ops/welford.py
```diff
@@ -0,0 +1,70 @@
...
+            delta_pre = x - mean
+            updated_mean = mean + delta_pre / self.n_samples
+            delta_post = x - updated_mean
+            deltas.append((updated_mean, delta_pre, delta_post))
+        return deltas
+
+    def _unroll(self, sample):
+        unrolled = []
+        for rolled in sample:
+            unrolled.extend(rolled.reshape(-1).data.tolist())
```
fehiepsi: make `unrolled` a list of "unrolled" tensors or a unrolled+joined tensor, rather than "unroll+joined" list of floats.

pyro/ops/welford.py
```diff
@@ -0,0 +1,70 @@
...
+    [1] `The Art of Computer Programming`,
+    Donald E. Knuth
+    """
+    def __init__(self, diagonal=True):
+        self.diagonal = diagonal
+        self.reset()
+
+    def _deltas(self, sample):
+        deltas = []
+        for i, x in enumerate(sample):
```
fehiepsi: calculate deltas in a vectorized way, rather than compute each entry

pyro/ops/welford.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def update(self, sample):
+        self.n_samples += 1
+        deltas = self._deltas(self._unroll(sample))
+        for i, (mean_x, delta_x_pre, delta_x_post) in enumerate(deltas):
+            self.means[i] = mean_x
+            for j, (mean_y, delta_y_pre, delta_y_post) in enumerate(deltas):
+                # Only compute the upper triangular covariance.
+                if i == j or (i < j and not self.diagonal):
+                    self.variances[(i, j)] += delta_x_pre * delta_y_post
```
fehiepsi: similar to delta, let's calculate variances in a vectorized way

pyro/ops/welford.py
```diff
@@ -0,0 +1,70 @@
...
+                if i == j or not self.diagonal:
+                    key = (i, j) if i <= j else (j, i)
+                    estimate = self.variances[key] / (self.n_samples - 1)
+                    if regularize:
+                        # Regularization from stan
+                        scaled_estimate = (self.n_samples / (self.n_samples + 5.)) * estimate
+                        shrinkage = 1e-3 * (5. / (self.n_samples + 5.0)) if i == j else 0.
+                        estimate = scaled_estimate + shrinkage
+                    row.append(estimate)
+            rows.append(row)
```
fehiepsi: with "vectorized" ones, you don't have to do these kinds of `for` and `append` job.

fehiepsi:
> Beyond that, I don't think we should need any changes to the integrator.
@neerajprad Will you do other transformations on `z` and `r`? I follow the formula (2.7), (2.8) of Neal's paper and saw that we need to update (joined) `z` according to `dz/dt = M\r`. If we transform `r`, then we can keep `dz/dt = r` but we have to incorporate it in `dr/dt`. Either way, we have to unroll, join, and multiply with big matrix. Or you mean that we don't use an explicit form of `dz/dt = M\r` but use `dz/dt = d(kinetic_energy)/dr`? But this will affect the "diagonal" version. Please correct me if I am wrong.

About `dict` or `list` of tensors, they are both fine to me. And about a fast Welford, let's do it in "vectorized" way.

neerajprad:
@LoganWalls - Incorporating @fehiepsi's suggestions on vectorization will simplify the code quite a bit, and will be much faster! Also, feel free to add me as a contributor to your branch if you need help in incorporating those changes.

Or you mean that we don't use an explicit form of dz/dt = M\r but use dz/dt = d(kinetic_energy)/dr? But this will affect the "diagonal" version because we don't use an explicit formula for d(kinetic)/dt. 

@fehiepsi - Sorry, I thought that we are differentiating the KE term, but I see that we are actually using the explicit gradient instead (as we should). It doesn't make sense to use autograd here which will be needlessly expensive. We can make the change to incorporate `M^{-1}` in the integrator when computing `dz/dt` and have a more optimized version that avoids the `.inverse` and matrix multiply for the diagonal mass matrix. It would be ideal if operations like `inverse` and `torch.mm` were optimized for diagonal matrices,  but I doubt if that's the case. We can make these changes in a separate PR. Let me know your thoughts.

fehiepsi:
@neerajprad Diagonal matrix can be stored as a vector. Multiplication is pretty simple: `Dx = D * x`.

In case we stick with this big "mass matrix", then I propose that Welford scheme just takes input `sample` is a vector (not a list of vectors). We will unroll or join things in `hmc` implementation instead. This makes things more transparent (so we don't have to explain what does it means by this big mass matrix: covariance of unrolled and joined variables).

For `torch.inverse` or `torch.trtrs`, I think that we just need to use it when we sample `r`. It might be expensive but just be executed at the beginning of an MCMC step, so I guess it does not cause much trouble.

neerajprad:
> In case we stick with this big "mass matrix"

I think we should have a mode where full mass matrix estimation is possible, even if this is not the default, just like Stan or PyMC3. We can shoot for a sparse block diagonal estimate in the future. It is okay if the dense matrix estimation is slow in practice to begin with, for as long as we aren't compromising the efficiency of sampling with diagonal mass matrix. 

> Diagonal matrix can be stored as a vector. Multiplication is pretty simple: Dx = D * x. In case we stick with this big "mass matrix", then I propose that Welford scheme just takes input sample is a vector (not a list of vectors). We will unroll or join things in hmc implementation instead. This makes things more transparent (so we don't have to explain what does it means by this big mass matrix: covariance of unrolled and joined variables).

Yes, this sounds reasonable. Just to confirm - you are proposing that the Welford scheme return a vector when we are estimating a diagonal mass matrix, and a dense matrix when we are estimating the full mass matrix, and we handle operations on them appropriately in HMC? 

> For torch.inverse or torch.trtrs, I think that we just need to use it when we sample r. It might be expensive but just be executed at the beginning of an MCMC step (and can be stored after adaptation phase), so I guess it does not cause much trouble.

I think we will need it to sample `r` from `dist.Normal` (for diagonal mass matrix) and `dist.MultivariateNormal` (for the dense mass matrix), but we will also need to multiply it during the update for `z` in each Leapfrog step. I don't think that can be avoided. Let me know.

fehiepsi:
> you are proposing that the Welford scheme return a vector when we are estimating a diagonal mass matrix, and a dense matrix when we are estimating the full mass matrix, and we handle operations on them appropriately in HMC?

Yup, that's what I have in mind.

LoganWalls:
Based on @fehiepsi 's previous comments I've gone ahead and changed `WelfordCovariance` to accept samples as a single vector rather than a list of vectors, since it sounds like we'll need to unroll elsewhere in `HMC` anyway. I hope this is alright?

pyro/ops/welford.py
```diff
@@ -0,0 +1,46 @@
...
+    def update(self, sample):
+        self.n_samples += 1
+        delta_pre = sample - self._mean
+        self._mean = self._mean + delta_pre / self.n_samples
+        delta_post = sample - self._mean
+
+        if self.diagonal:
+            self._m2 += delta_pre * delta_post
+        else:
+            self._m2 += delta_pre * delta_post.reshape(-1, 1)
```
fehiepsi: Use `torch.matmul` is faster (I made a benchmark [here](https://gist.github.com/fehiepsi/6c5c769af2b2ca9e29c301ba30b98f6d)). In addition, to add an additional dimension, I think that using `.unsqueeze()` is better (faster) than `.reshape()`.

pyro/ops/welford.py
```diff
@@ -0,0 +1,46 @@
...
+            raise RuntimeError('Insufficient samples to estimate covariance')
+        cov = self._m2 / (self.n_samples - 1)
+        if regularize:
+            # Regularization from stan
+            scaled_cov = (self.n_samples / (self.n_samples + 5.)) * cov
+            shrinkage = 1e-3 * (5. / (self.n_samples + 5.0))
+            if self.diagonal:
+                cov = scaled_cov + shrinkage
+            else:
+                cov = scaled_cov + torch.diagflat(shrinkage)
```
fehiepsi: Will this give a correct behavior? We want to add `shrinkage` to diagonal part, and I guess this line will add `shrinkage` to all entries. The fastest way to do is
`scale_cov.view(-1)[::scaled_cov.shape[0]+1] += shrinkage`! (I learned this trick from Adam in pytorch slack.)

fehiepsi: Yes, it is fine for diagonal version.

PR #1206, opened 2018-06-26
clarify param_store.load() docs

### LINKED ISSUE #1204: Update paramstore documentation
Params are not properly reused from the paramstore via `.load()` since the default flag `update_module_params` is false (iirc this is because of `random_module`).  the interface/defaults should be revisited and at the very least, documentation updated.

[source](https://forum.pyro.ai/t/pyro-get-param-store-load-documentation/299)

jpchen:
Resolves #1204

### DIFF
--- a/pyro/params/param_store.py
+++ b/pyro/params/param_store.py
@@ -174,6 +174,14 @@ class ParamStoreDict(object):
         """
         Loads parameters from disk
 
+        .. note::
+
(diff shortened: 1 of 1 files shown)


jpchen:
i dont know why this has been running for a full day and the travis link is broken so i cant rebuild... i think this is merge-able

PR #1211, opened 2018-06-28
add LSH to contrib.tracking

alicanb:
Addresses #1185 

This PR adds LSH from pyro-iei to contrib.tracking. One design choice I made is to allow overwriting in `lsh.add`. We can easily change that to not allow overwriting in `lsh.add` and add `lsh.modify` to facilitate that need.

### DIFF
--- a/pyro/contrib/tracking/hashing.py
+++ b/pyro/contrib/tracking/hashing.py
@@ -33,7 +33,7 @@ class LSH(object):
         set(['a', 'c'])
         >>> lsh.remove('b')
         >>> lsh.nearby('a')
-        set()
+        set([])
 
 
     :param float radius: Scaling parameter used in hash function. Determines the size of the neighbourhood.
--- a/pyro/contrib/tracking/hashing.py
+++ b/pyro/contrib/tracking/hashing.py
@@ -7,28 +7,34 @@ from numbers import Number
 
 class LSH(object):
     """
-    Implements locality-sensitive hashing. Allows to efficiently find neighbours
-    of a point. Provides 2 guarantees:
-    - Difference between coordinates of points not returned by :meth:`nearby`
-    and input point is larger than `radius`.
-    - Difference between coordinates of points returned by :meth:`nearby` and
-    input point is smaller than 2`radius`.
-    :param float radius: scaling parameter used in hash function. Determines
-    the size of the neighbourhood.
+    Implements locality-sensitive hashing.
 
-    Example::
-        radius = 1
-        lsh = LSH(radius)
-        a = torch.tensor([-0.49, -0.49]) # hash(a)=(0,0)
-        b = torch.tensor([1.0, 1.0]) # hash(b)=(1,1)
-        c = torch.tensor([2.49, 2.49]) # hash(c)=(2,2)
-        lsh.add('a', a)
-        lsh.add('b', b)
-        lsh.add('c', c)
-        lsh.nearby('a') # {'b'}
-        lsh.nearby('b') # {'a','c'}
-        lsh.remove('b')
-        lsh.nearby('a') # {}
+
+    Allows to efficiently find neighbours of a point. Provides 2 guarantees:
+
+    -   Difference between coordinates of points not returned by :meth:`nearby`
+        and input point is larger than ``radius``.
+    -   Difference between coordinates of points returned by :meth:`nearby` and
+        input point is smaller than 2 ``radius``.
+
+    Example:
+
+        >>> radius = 1
+        >>> lsh = LSH(radius)
+        >>> a = torch.tensor([-0.51, -0.51]) # hash(a)=(-1,-1)
+        >>> b = torch.tensor([-0.49, -0.49]) # hash(a)=(0,0)
+        >>> c = torch.tensor([1.0, 1.0]) # hash(b)=(1,1)
+        >>> lsh.add('a', a)
+        >>> lsh.add('b', b)
+        >>> lsh.add('c', c)
+        >>> lsh.nearby('a') # {'b'}, even though c is within 2radius of a
+        >>> lsh.nearby('b') # {'a','c'}
+        >>> lsh.remove('b')
+        >>> lsh.nearby('a') # {}
+
+
+    :param float radius: Scaling parameter used in hash function. Determines the size of the neighbourhood.
+
     """
     def __init__(self, radius):
         assert radius > 0 if isinstance(radius, Number) else (radius > 0).all(), \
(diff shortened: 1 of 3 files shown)


alicanb:
BTW these tests take much longer time than I expected, probably mainly in checking set equivalence. Is there a faster way to do that? I don't want to add 6 secs to test time just for this.

tests/contrib/tracking/test_hashing.py
```diff
@@ -0,0 +1,131 @@
...
+def test_lsh_add(scale):
+    lsh = LSH(scale)
+    a = torch.rand(10)
+    lsh.add('a', a)
+    assert lsh._hash_to_key[lsh._key_to_hash['a']] == {'a'}
+
+
+@pytest.mark.parametrize('scale', [0.1, 1, 10, 100])
+def test_lsh_hash_nearby(scale):
+    k = 10
```
fritzo: Lookup has complexity `O(3^k)`. If these tests are slow, try reducing k to say 7.

fritzo: BTW we could easily replace this implementation with an `O(2^k)` version which is much cheaper for higher dimensions.

pyro/contrib/tracking/hashing.py
```diff
@@ -0,0 +1,66 @@
...
+import itertools
+from collections import defaultdict
+from numbers import Number
+
+
+class LSH(object):
+    """
+    Implements locality-sensitive hash for low-dimensional euclidean space.
+    """
+    def __init__(self, scale):
```
fritzo: Can we rename `scale` to `radius` and describe the radius guarantee in the docstring?

pyro/contrib/tracking/hashing.py
```diff
@@ -0,0 +1,66 @@
+from __future__ import absolute_import, division, print_function
+
+import itertools
+from collections import defaultdict
+from numbers import Number
+
+
+class LSH(object):
+    """
+    Implements locality-sensitive hash for low-dimensional euclidean space.
```
fritzo: Can you describe the `:param float radius: ...` in this docstring, and maybe add an `Example::` section with usage?

pyro/contrib/tracking/hashing.py
```diff
@@ -0,0 +1,66 @@
...
+            "scale must be greater than 0, given: {}".format(scale)
+        self._scale = scale
+        self._hash_to_key = defaultdict(set)
+        self._key_to_hash = {}
+
+    def _hash(self, point):
+        coords = (point.detach().cpu() / self._scale).round()
+        return tuple(map(int, coords))
+
+    def add(self, key, point):
```
fritzo: add docstring for public methods

pyro/contrib/tracking/hashing.py
```diff
@@ -0,0 +1,66 @@
...
+        _hash = self._hash(point)
+        if key in self._key_to_hash.keys():
+            self.remove(key)
+        self._key_to_hash[key] = _hash
+        self._hash_to_key[_hash].add(key)
+
+    def remove(self, key):
```
fritzo: add docstring for public methods

pyro/contrib/tracking/hashing.py
```diff
@@ -0,0 +1,66 @@
...
+        _hash = self._key_to_hash.pop(key)
+        self._hash_to_key[_hash].remove(key)
+
+    def nearby(self, key):
```
fritzo: add docstring for public methods

pyro/contrib/tracking/hashing.py
```diff
@@ -0,0 +1,66 @@
...
+        self._bins = set()
+
+    def _hash(self, point):
+        coords = (point.detach().cpu() / self._scale).round()
+        return tuple(map(int, coords))
+
+    def try_add(self, point):
+        """
+        If bin is unoccupied, adds to bin and returns True.
+        If bin is already occupied, returns False.
```
fritzo: improve docstring: add a `:param torch.Tensor point: ...` section and a `:return: ...` section

pyro/contrib/tracking/hashing.py
```diff
@@ -0,0 +1,66 @@
...
+            result |= self._hash_to_key[nearby_hash]
+        result.remove(key)
+        return result
+
+
+class ApproxSet(object):
+    """
+    Queries low-dimensional euclidean space for approximate occupancy.
+    """
+    def __init__(self, scale):
```
fritzo: Can we rename `scale` to `radius` and describe the guarantee?

pyro/contrib/tracking/hashing.py
```diff
@@ -0,0 +1,111 @@
...
+        return tuple(map(int, coords))
+
+    def add(self, key, point):
+        """
+        Adds (`key`, `hash(point)`) pair to the hash
+        :param key: key used identify point
+        :param point: data, should be hashable
+        """
+        _hash = self._hash(point)
+        if key in self._key_to_hash.keys():
```
fritzo: Remove the unnecessary `.keys()`. In Python2 this creates an entire copy of keys list each call.

fritzo:
Can you ensure that `make docs` works for you locally and produces desired results? On my mac I usually `make docs && open docs/build/html/index.html` and snoop around. A couple style notes on documentation:
- double ticks trigger code font (single ticks are merely italic)
- It's nice to follow a standard pattern separated by double newlines:
    - a single sentence description
    - paragraphs explaining details
    - a :param: section

pyro/contrib/tracking/hashing.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def _hash(self, point):
+        coords = (point.detach().cpu() / self._radius).round()
+        return tuple(map(int, coords))
+
+    def add(self, key, point):
+        """
+        Adds (`key`, `hash(point)`) pair to the hash
+        :param key: key used identify point
+        :param point: data, should be hashable
```
fritzo: Isn't it `key` rather than `point` that should be hashable?
Also: `:param torch.Tensor point`

pyro/contrib/tracking/hashing.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    def __init__(self, radius):
+        assert radius > 0 if isinstance(radius, Number) else (radius > 0).all(), \
+            "radius must be greater than 0, given: {}".format(radius)
+        self._radius = radius
+        self._hash_to_key = defaultdict(set)
+        self._key_to_hash = {}
+
+    def _hash(self, point):
+        coords = (point.detach().cpu() / self._radius).round()
```
fritzo: I think it's reasonable to expect the caller to have called `.detach().cpu()` on the large tensor of points from which this point is sliced out.

alicanb:
Thanks @fritzo, I fixed the docs. something's wrong with the CI build though...

pyro/contrib/tracking/hashing.py
```diff
@@ -0,0 +1,130 @@
...
+        >>> a = torch.tensor([-0.51, -0.51]) # hash(a)=(-1,-1)
+        >>> b = torch.tensor([-0.49, -0.49]) # hash(a)=(0,0)
+        >>> c = torch.tensor([1.0, 1.0]) # hash(b)=(1,1)
+        >>> lsh.add('a', a)
+        >>> lsh.add('b', b)
+        >>> lsh.add('c', c)
+        >>> print(lsh.nearby('a')) # {'b'}, even though c is within 2radius of a
+        >>> print(lsh.nearby('b')) # {'a','c'}
+        >>> lsh.remove('b')
+        >>> print(lsh.nearby('a')) # {}
```
neerajprad: @alicanb - This is failing the doctest. You can run `make docs` to replicate the failure. You can just remove the prints and append the output below for the tests to pass. Let me know if you need any help fixing this. e.g.

```
>>> lsh.nearby('a')
set(['b'])
```

alicanb: Thanks @neerajprad , weird- I can build successfully on my machine... Anyways I pushed a fix, I'll ask you if this doesn't work.

fritzo:
@ae-foster This is a simple low-dimensional euclidean binning algorithm rather than a high-dimensional random hyperplane cut algorithm. Do you think 'LSH' is an appropriate name, or should we rename this to make room for / avoid confusion with a random hyperplane implementation?

ae-foster:
Though I'm no expert, I say LSH goes with any hashing algorithm. E.g. in this package https://github.com/ekzhu/datasketch we called it MinHashLSH- it's LSH using MinHash. You could call it <HashAlgorithm>LSH?

neerajprad:
+1. Maybe `EuclideanLSH` or `EuclideanBinningLSH`?  (leaving room for RandomHyperplanesLSH for cosine distance)?

tests/contrib/tracking/test_hashing.py
```diff
@@ -0,0 +1,131 @@
...
+import pytest
+import torch
+
+from pyro.contrib.tracking.hashing import LSH, ApproxSet
+from tests.common import assert_equal
+
+
+@pytest.mark.parametrize('scale', [-1., 0., -1 * torch.ones(2, 2)])
+def test_lsh_init(scale):
+    num_error = 0
```
fritzo: A cleaner way to write death tests is to use pytest:
```py
@pytest.mark.parametrize('scale', [-1., 0., -1 * torch.ones(2, 2)])
def test_lsh_init(scale):
    with pytest.raises(ValueError):
        LSH(scale)    
```

pyro/contrib/tracking/hashing.py
```diff
@@ -0,0 +1,133 @@
...
+class ApproxSet(object):
+    """
+    Queries low-dimensional euclidean space for approximate occupancy.
+
+
+    :param float radius: scaling parameter used in hash function. Determines the size of the bin.
+                         See :class:`LSH` for details.
+    """
+    def __init__(self, radius):
+        assert radius > 0 if isinstance(radius, Number) else (radius > 0).all(), \
```
fritzo: int: user-facing errors should be `ValueError`s rather than assertions:
```py
if not (radius > 0):
    raise ValueError("radius must be greater than 0, given: {}".format(radius))
```
Also I think it's fine to require float rather than vector radius. If a user wants a diagonal scale, they can map the `point`s by that scale.

pyro/contrib/tracking/hashing.py
```diff
@@ -0,0 +1,133 @@
+from __future__ import absolute_import, division, print_function
+
+import itertools
+from collections import defaultdict
+from numbers import Number
+
+
+class LSH(object):
+    """
+    Implements locality-sensitive hashing.
```
fritzo: How about we keep the class name `LSH` and describe a little more specifically:
> Implements locality-sensitive hashing for low-dimensional euclidean space.

alicanb: I kept the name `LSH` for now, will we have additional hashing schemes in the future?

fritzo: It's conceivable that we might track in embedding space rather than physical space, and in that case a cosine-distance random-hyperplane LSH would work better than this version. But let's address that when it arises.

PR #1214, opened 2018-07-04
Implement renyi elbo

fehiepsi:
This pull request adds Renyi Elbo. In the case `alpha = 0`, we have IWELBO.

I wrote the derivation [here](https://gist.github.com/fehiepsi/7916400d8380cb05d21e859681b7a694) for reference. Note that this works with non-parameterizable variables too.

Todo list:
- [x] Implement
- [x] Add docs, comments
- [x] Add tests

### DIFF
--- a/pyro/distributions/util.py
+++ b/pyro/distributions/util.py
@@ -209,7 +209,7 @@ def log_sum_exp(tensor, dim=-1):
     :param dim: Dimension to be summed out.
     """
     max_val = tensor.max(dim)[0]
-    return max_val + (tensor - max_val.unsqueeze(-1)).exp().sum(dim=dim).log()
+    return max_val + (tensor - max_val.unsqueeze(dim)).exp().sum(dim=dim).log()
 
 
 def enable_validation(is_validate):
--- a/pyro/infer/renyi_elbo.py
+++ b/pyro/infer/renyi_elbo.py
@@ -112,7 +112,7 @@ class RenyiELBO(ELBO):
             for name, site in model_trace.nodes.items():
                 if site["type"] == "sample":
                     if is_vectorized:
-                        log_prob_sum = site["log_prob"].reshape(self.num_particles, -1).sum(-1).detach()
+                        log_prob_sum = site["log_prob"].detach().reshape(self.num_particles, -1).sum(-1)
                     else:
                         log_prob_sum = torch_item(site["log_prob_sum"])
 
@@ -122,7 +122,7 @@ class RenyiELBO(ELBO):
                 if site["type"] == "sample":
                     log_prob, score_function_term, entropy_term = site["score_parts"]
                     if is_vectorized:
-                        log_prob_sum = log_prob.reshape(self.num_particles, -1).sum(-1).detach()
+                        log_prob_sum = log_prob.detach().reshape(self.num_particles, -1).sum(-1)
                     else:
                         log_prob_sum = torch_item(site["log_prob_sum"])
 
@@ -140,7 +140,7 @@ class RenyiELBO(ELBO):
 
             elbo_particles_scaled = (1. - self.alpha) * elbo_particles
             elbo_scaled = log_sum_exp(elbo_particles_scaled, dim=0) - math.log(self.num_particles)
-            elbo = elbo_scaled.sum().item() / (1 - self.alpha)
+            elbo = elbo_scaled.sum().item() / (1. - self.alpha)
 
         loss = -elbo
         warn_if_nan(loss, "loss")
@@ -211,7 +211,7 @@ class RenyiELBO(ELBO):
 
         if self.num_particles == 1:
             elbo = elbo_trace[0]
-            surrogate_elbo = surrogate_elbo_trace[0]
+            surrogate_elbo_particles = surrogate_elbo_trace[0]
         else:
             if is_vectorized:
                 elbo_particles = elbo_trace[0]
@@ -222,7 +222,7 @@ class RenyiELBO(ELBO):
 
             elbo_particles_scaled = (1. - self.alpha) * elbo_particles
             elbo_scaled = log_sum_exp(elbo_particles_scaled, dim=0) - math.log(self.num_particles)
-            elbo = elbo_scaled.sum().item() / (1 - self.alpha)
+            elbo = elbo_scaled.sum().item() / (1. - self.alpha)
 
         # collect parameters to train from model and guide
         trainable_params = any(site["type"] == "param"
--- a/pyro/infer/renyi_elbo.py
+++ b/pyro/infer/renyi_elbo.py
@@ -151,7 +151,7 @@ class RenyiELBO(ELBO):
         if self.num_particles == 1:
             elbo_scaled = elbo_particles_scaled
         else:
-            elbo_scaled = log_sum_exp(elbo_particle_scaled, dim=0) - math.log(self.num_particles)
+            elbo_scaled = log_sum_exp(elbo_particles_scaled, dim=0) - math.log(self.num_particles)
 
         loss = elbo_scaled.sum().item() / (self.alpha - 1.)
         warn_if_nan(loss, "loss")
--- a/pyro/infer/renyi_elbo.py
+++ b/pyro/infer/renyi_elbo.py
@@ -3,12 +3,14 @@ from __future__ import absolute_import, division, print_function
 import math
 import warnings
 
+import torch
+
 import pyro.poutine as poutine
 from pyro.distributions.util import is_identically_zero, log_sum_exp
 from pyro.infer.elbo import ELBO
 from pyro.infer.util import is_validation_enabled
 from pyro.poutine.util import prune_subsample_sites
-from pyro.util import check_model_guide_match, check_site_shape, torch_isnan
+from pyro.util import check_model_guide_match, check_site_shape, warn_if_nan
 
 
 class RenyiELBO(ELBO):
@@ -23,7 +25,7 @@ class RenyiELBO(ELBO):
 
     .. note:: Setting :math:`\alpha < 1` gives a better bound than the usual ELBO.
         For :math:`\alpha = 1`, it is better to use
-        :class:`~pyro.infer.trace_elbo.Trace_ELBO` class because it helps reducing
+        :class:`~pyro.infer.trace_elbo.Trace_ELBO` class because it helps reduce
         variances of gradient estimations.
 
     .. warning:: Mini-batch training is not supported yet.
@@ -99,29 +101,60 @@ class RenyiELBO(ELBO):
 
         Evaluates the ELBO with an estimator that uses num_particles many samples/particles.
         """
-        model_trace, guide_trace = next(self._get_traces(model, guide, *args, **kwargs))
-        elbo_particle = 0
-        for name, site in model_trace.nodes.items():
-            if site["type"] == "sample":
-                elbo_particle = elbo_particle + site["log_prob"].detach()
+        elbo_list = []
+        tensor_holder = None
 
-        for name, site in guide_trace.nodes.items():
-            if site["type"] == "sample":
-                log_prob, score_function_term, entropy_term = site["score_parts"]
-                elbo_particle = elbo_particle - log_prob.detach()
+        # grab a vectorized trace from the generator
+        for model_trace, guide_trace in self._get_traces(model, guide, *args, **kwargs):
+            elbo_particle = 0
 
-        if is_identically_zero(elbo_particle):
+            # compute elbo
+            for name, site in model_trace.nodes.items():
+                if site["type"] == "sample":
+                    if self.vectorize_particles and self.num_particles > 1:
+                        log_prob_sum = site["log_prob"].reshape(self.num_particles, -1).sum(-1)
+                    else:
+                        log_prob_sum = site["log_prob_sum"]
+                    elbo_particle = elbo_particle + log_prob_sum.detach()
+
+            for name, site in guide_trace.nodes.items():
+                if site["type"] == "sample":
+                    log_prob, score_function_term, entropy_term = site["score_parts"]
+                    if self.vectorize_particles and self.num_particles > 1:
+                        log_prob_sum = log_prob.reshape(self.num_particles, -1).sum(-1)
+                    else:
+                        log_prob_sum = site["log_prob_sum"]
+
+                    elbo_particle = elbo_particle - log_prob_sum.detach()
+
+            if is_identically_zero(elbo_particle):
+                if tensor_holder is not None:
+                    elbo_particle = tensor_holder.new_zeros(tensor_holder.shape)
+            else:  # elbo_particle is not None
+                if tensor_holder is None:
+                    tensor_holder = elbo_particle.new_empty(elbo_particle.shape)
+                    # change types of previous `elbo_particle`s
+                    for i in range(len(elbo_list)):
+                        elbo_list[i] = tensor_holder.new_zeros(tensor_holder.shape)
+
+            elbo_list.append(elbo_particle)
+
+        if tensor_holder is None:
             return 0.
 
-        elbo_particle_scaled = (1. - self.alpha) * elbo_particle
+        if (not self.vectorize_particles) and self.num_particles > 1:
+            elbo_particles = torch.stack(elbo_list)
+        else:
+            elbo_particles = elbo_list[0]
+
+        elbo_particles_scaled = (1. - self.alpha) * elbo_particles
         if self.num_particles == 1:
-            elbo_scaled = elbo_particle_scaled
+            elbo_scaled = elbo_particles_scaled
         else:
             elbo_scaled = log_sum_exp(elbo_particle_scaled, dim=0) - math.log(self.num_particles)
 
         loss = elbo_scaled.sum().item() / (self.alpha - 1.)
-        if torch_isnan(loss):
-            warnings.warn('Encountered NAN loss')
+        warn_if_nan(loss, "loss")
         return loss
 
     def loss_and_grads(self, model, guide, *args, **kwargs):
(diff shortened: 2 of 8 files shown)


pyro/distributions/util.py
```diff
@@ -209,7 +209,7 @@ def log_sum_exp(tensor, dim=-1):
     :param dim: Dimension to be summed out.
     """
     max_val = tensor.max(dim)[0]
-    return max_val + (tensor - max_val.unsqueeze(-1)).exp().sum(dim=dim).log()
+    return max_val + (tensor - max_val.unsqueeze(dim)).exp().sum(dim=dim).log()
```
fehiepsi: I guess this is a bug?

neerajprad: Thanks for fixing, this is indeed a bug.

alicanb: btw apparently there is logsumexp in aten [now](https://github.com/t-vi/pytorch/commit/dec9932b114ef5351cc7e5ce9ce33c5a8a32fc34), you might want to switch to that altogether.

fehiepsi: Yes, we'll switch when updating Pyro to match Pytorch master. I listed it here: https://github.com/uber/pyro/issues/1212

fehiepsi:
To reviewers: I cannot find parameters to pass BernoulliBeta test so I leave an xfail mark there. To be honest, I am not so sure about the correctness of my implementation w.r.t. `iarange` context in Pyro, so it is hard for me to detect bugs in my code. Please help me debug the code if possible. Thanks in advance!

pyro/infer/renyi_elbo.py
```diff
@@ -0,0 +1,179 @@
...
+            return 0.
+
+        elbo_particle_scaled = (1. - self.alpha) * elbo_particle
+        if self.num_particles == 1:
+            elbo_scaled = elbo_particle_scaled
+        else:
+            elbo_scaled = log_sum_exp(elbo_particle_scaled, dim=0) - math.log(self.num_particles)
+
+        loss = elbo_scaled.sum().item() / (self.alpha - 1.)
+        if torch_isnan(loss):
```
jpchen: ditto

pyro/infer/renyi_elbo.py
```diff
@@ -0,0 +1,179 @@
...
+    follows reference [1].
+
+    To have a lower bound, we require :math:`\alpha \ge 0`. However, according to
+    reference [1], depending on the dataset, :math:`\alpha < 0` might give better
+    results. In the special case :math:`\alpha = 0`, we have important weighted
+    lower bound derived in reference [2].
+
+    .. note:: Setting :math:`\alpha < 1` gives a better bound than the usual ELBO.
+        For :math:`\alpha = 1`, it is better to use
+        :class:`~pyro.infer.trace_elbo.Trace_ELBO` class because it helps reducing
```
jpchen: nit: "helps reduce"

pyro/infer/renyi_elbo.py
```diff
@@ -0,0 +1,179 @@
...
+                    'If you want to enumerate sites, you need to use TraceEnum_ELBO instead.']))
+        guide_trace = prune_subsample_sites(guide_trace)
+        model_trace = prune_subsample_sites(model_trace)
+
+        model_trace.compute_log_prob()
+        guide_trace.compute_score_parts()
+        if is_validation_enabled():
+            for site in model_trace.nodes.values():
+                if site["type"] == "sample":
+                    check_site_shape(site, self.max_iarange_nesting)
```
jpchen: this is used verbatim in almost every *elbo implementation, we should eventually move this out to a helper

tests/infer/test_inference.py
```diff
@@ -213,6 +225,7 @@ def guide():
     Trace_ELBO,
     TraceGraph_ELBO,
     TraceEnum_ELBO,
+    xfail_param(RenyiELBO, reason="Assertation errors at alpha_q, beta_q", run=False),
```
jpchen: hm this is worrisome, we can leave this as is until i have time to dive deeper into this unless @neerajprad wants to take a look

neerajprad: @jpchen - I don't see anything obviously wrong atm, but I do want to dig deeper.

neerajprad:
@fehiepsi: The gradient computation seems to be off by a huge magnitude in `test_gradient.py::test_iarange_elbo_vectorized_particles`. You can add `RenyiELBO` to the test parameters to replicate. When you set the weights to 1., you do retrieve the correct gradient (as you should given that it is no different from `Trace_ELBO`). However, with the importance weights (which seem to be correctly scaled), the gradients are hugely off (maybe there is some unaccounted gradient flowing back?).

pyro/infer/renyi_elbo.py
```diff
@@ -0,0 +1,179 @@
...
+        # collect parameters to train from model and guide
+        trainable_params = any(site["type"] == "param"
+                               for trace in (model_trace, guide_trace)
+                               for site in trace.nodes.values())
+
+        if trainable_params and getattr(surrogate_elbo_particle, 'requires_grad', False):
+            if self.num_particles == 1:
+                weights = 1.
+            else:
+                weights = (elbo_particle_scaled - elbo_scaled).exp()
```
neerajprad: do we need to `detach` the weights?

pyro/infer/renyi_elbo.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+
+    def __init__(self, alpha=0., num_particles=2, max_iarange_nesting=2,
+                 strict_enumeration_warning=True):
+        if alpha == 1:
+            raise ValueError("The order alpha should not be equal to 1. Please use Trace_ELBO class"
+                             "for the case alpha = 1.")
+        self.alpha = alpha
+        super(RenyiELBO, self).__init__(num_particles, max_iarange_nesting, vectorize_particles=True,
+                                        strict_enumeration_warning=True)
```
neerajprad: Replace `True` with value from constructor.

pyro/infer/renyi_elbo.py
fehiepsi: these tensors have no gradients I guess

pyro/infer/renyi_elbo.py
fehiepsi: hah, thx :D

fehiepsi:
Thanks @neerajprad! This gradient off might be relevant to https://github.com/pytorch/pytorch/pull/8721: the `log_prob` in guide_trace will be expanded to be added into `surrogate_loss`. I still can not figure out what is the reason now but will go deeper into it in the next couple of days.

neerajprad:
> I still can not figure out what is the reason now but will go deeper into it in the next couple of days.

Sure, let me know if you face any issues. I will also try debugging this at my end.

tests/infer/test_inference.py
```diff
@@ -47,12 +47,18 @@ def setUp(self):
         self.batch_size = 4
 
     def test_elbo_reparameterized(self):
-        self.do_elbo_test(True, 5000)
+        self.do_elbo_test(True, 5000, Trace_ELBO())
```
neerajprad: For these tests, the value of the loss seems to be somewhat different (almost 3X or so) the value from `Trace_ELBO` with `alpha=0.99` (or any other value for that matter). Is that expected?

fehiepsi:
@neerajprad I think that this article points out the reason for grad off https://arxiv.org/abs/1802.04537. They showed that "the probability that the gradient is positive or negative becomes roughly equal for larger values of K, meaning the optimizer is equally likely to increase as decrease the inference network parameters at the next iteration." (K is num_particles).

tests/infer/test_inference.py
fehiepsi: If num_particles = 1, the loss would be the same.

fehiepsi:
@neerajprad I think that there is something wrong related to `log_prob_sum` and `log_prob.expand(shape).sum()`. I will check this.

fehiepsi:
Yes, in ELBO, we take the log_prob_sum without expanding to independent dimensions. In Renyi, I expand the log_prob for independent dimensions, then take log_prob_sum.

tests/infer/test_inference.py
neerajprad: Sorry, I should have mentioned - with `num_particles=1`, in `NormalNormalTest`, the elbo loss returned by `svi.step` is around 6 for `test_elbo_reparameterized` and around 20 for `test_renyi_reparameterized` (and doesn't seem to change too much with `num_particles`).

pyro/infer/renyi_elbo.py
```diff
@@ -0,0 +1,179 @@
...
+        Yuri Burda, Roger Grosse, Ruslan Salakhutdinov
+    """
+
+    def __init__(self, alpha=0., num_particles=2, max_iarange_nesting=2,
+                 strict_enumeration_warning=True):
+        if alpha == 1:
+            raise ValueError("The order alpha should not be equal to 1. Please use Trace_ELBO class"
+                             "for the case alpha = 1.")
+        self.alpha = alpha
+        super(RenyiELBO, self).__init__(num_particles, max_iarange_nesting, vectorize_particles=True,
```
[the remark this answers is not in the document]
neerajprad: I realized that it might be better to keep `vectorize_particles=False`, at least for now, because otherwise we will have to assume some value for `max_iarange_nesting` which is set to 2 here. This will cause dim clashes for models with more than 2 iaranges. We revert to True later, and set a higher default value for `max_iarange_nesting`, and throwing a sensible error if the model has more iaranges.

neerajprad:
> In Renyi, I expand the log_prob for independent dimensions, then take log_prob_sum.

Could you point me to the code where we are expanding `site['log_prob']`? I think `site['log_prob']` should already reflect all the existing batch dimensions, right? With `auto_vectorize=True` the outermost batch dimension should already be included in `log_prob.shape`.

pyro/infer/renyi_elbo.py
```diff
@@ -0,0 +1,251 @@
...
+
+        # grab a vectorized trace from the generator
+        for model_trace, guide_trace in self._get_traces(model, guide, *args, **kwargs):
+            elbo_particle = 0
+
+            # compute elbo
+            for name, site in model_trace.nodes.items():
+                if site["type"] == "sample":
+                    if self.vectorize_particles and self.num_particles > 1:
+                        log_prob_sum = site["log_prob"].reshape(self.num_particles, -1).sum(-1)
```
fehiepsi: @neerajprad last time I used `log_prob`, which leads to the expanding issue.

tests/infer/test_inference.py
```diff
@@ -243,7 +256,12 @@ def guide(alpha0, beta0, alpha_n, beta_n):
         pyro.sample("lambda_latent", gamma_dist(alpha_q, beta_q))
 
     adam = optim.Adam({"lr": .0003, "betas": (0.97, 0.999)})
-    elbo = elbo_impl(strict_enumeration_warning=False)
+    if elbo_impl is RenyiELBO:
+        elbo = elbo_impl(alpha=0.2, num_particles=3, strict_enumeration_warning=False)
+        if gamma_dist is ShapeAugmentedGamma:
+            pytest.skip("Assertation errors at beta_q!")
```
fehiepsi: @neerajprad  @jpchen  only this test is failed now. I guess it is fine?

fehiepsi:
I have refactored the code. It is good to review if tests pass.

tests/infer/test_inference.py
```diff
@@ -243,7 +256,12 @@ def guide(alpha0, beta0, alpha_n, beta_n):
         pyro.sample("lambda_latent", gamma_dist(alpha_q, beta_q))
 
     adam = optim.Adam({"lr": .0003, "betas": (0.97, 0.999)})
-    elbo = elbo_impl(strict_enumeration_warning=False)
+    if elbo_impl is RenyiELBO:
+        elbo = elbo_impl(alpha=0.2, num_particles=3, strict_enumeration_warning=False)
+        if gamma_dist is ShapeAugmentedGamma:
+            pytest.skip("RenyiELBO currently does not work with Rejector!")
```
fehiepsi: I made the issue #1222 for this skip.

jpchen: itd be great if you could link the issue in the comment

fehiepsi: I'll add it after @neerajprad 's review.

pyro/infer/renyi_elbo.py
```diff
@@ -0,0 +1,244 @@
...
+        # grab a vectorized trace from the generator
+        for model_trace, guide_trace in self._get_traces(model, guide, *args, **kwargs):
+            elbo_particle = 0
+            surrogate_elbo_particle = 0
+
+            # compute elbo and surrogate elbo
+            for name, site in model_trace.nodes.items():
+                if site["type"] == "sample":
+                    if is_vectorized:
+                        log_prob_sum = site["log_prob"].reshape(self.num_particles, -1).sum(-1)
```
fehiepsi: @neerajprad Things got hidden after some of my revisions. This is the point which bypasses the expanding issue. If I don't do this, then (assume num_particles=2) the sum of log_prob of a site x with shape `2 x 1 x 1` and log_prob of another site y with shape `2 x 1 x 2` would be different to the sum of log_prob_sum of x and log_prob_sum of y.

neerajprad: Yes, this seems like the correct behavior!

pyro/infer/renyi_elbo.py
```diff
@@ -0,0 +1,244 @@
...
+                    else:
+                        log_prob_sum = site["log_prob_sum"]
+
+                    elbo_particle = elbo_particle - log_prob_sum.detach()
+
+                    if not is_identically_zero(entropy_term):
+                        surrogate_elbo_particle = surrogate_elbo_particle - log_prob_sum
+
+                    if not is_identically_zero(score_function_term):
+                        surrogate_elbo_particle = (surrogate_elbo_particle +
```
neerajprad: I'm not familiar with the score parts computation, but it does seem like we are not making use of it in the non reparameterized case. Is this expected?

fehiepsi: Yes, we don't use it in the case "alpha = 0". The REINFORCE way (for non-reparameterized distribution) will not update the guide when "alpha = 0". I searched for literature and found that [Wake-Sleep](https://arxiv.org/abs/1406.2751) algorithm updates the guide **separately** (by maximizing another likelihood function). Their "wake phase to update guide" is equivalent to add this score_function_term to surrogate_elbo_particle (with some magnitude).

How to solve this problem and how to reduce variance for non-reparameterized distributions are good topics for a research paper in my opinion (because I couldn't find any article discussing them).

tests/infer/test_inference.py
```diff
@@ -243,7 +256,13 @@ def guide(alpha0, beta0, alpha_n, beta_n):
         pyro.sample("lambda_latent", gamma_dist(alpha_q, beta_q))
 
     adam = optim.Adam({"lr": .0003, "betas": (0.97, 0.999)})
-    elbo = elbo_impl(strict_enumeration_warning=False)
+    if elbo_impl is RenyiELBO:
+        elbo = elbo_impl(alpha=0.2, num_particles=3, strict_enumeration_warning=False)
+        if gamma_dist is ShapeAugmentedGamma:
```
fritzo: I would instead use `with xfail_if_not_implemented():` here and raise `NotImolementedError` inside `Reny_ELBO` in multiple score_parts terms are nonzero.

pyro/infer/renyi_elbo.py
```diff
@@ -0,0 +1,244 @@
...
+                        elbo_trace[i] = tensor_holder.new_zeros(tensor_holder.shape)
+                        surrogate_elbo_trace[i] = tensor_holder.new_zeros(tensor_holder.shape)
+
+            elbo_trace.append(elbo_particle)
+            surrogate_elbo_trace.append(surrogate_elbo_particle)
+
+        if tensor_holder is None:
+            return 0.
+
+        if self.num_particles == 1:
```
neerajprad: Related to above comment - can we simplify this and only distinguish between vectorized and non-vectorized case. From what I can see, with `n=1` the weight would be immaterial.

pyro/infer/renyi_elbo.py
```diff
@@ -0,0 +1,244 @@
...
+            surrogate_elbo_particles = surrogate_elbo_trace[0]
+        else:
+            if is_vectorized:
+                elbo_particles = elbo_trace[0]
+                surrogate_elbo_particles = surrogate_elbo_trace[0]
+            else:
+                elbo_particles = torch.stack(elbo_trace)
+                surrogate_elbo_particles = torch.stack(surrogate_elbo_trace)
+
+            elbo_particles_scaled = (1. - self.alpha) * elbo_particles
```
neerajprad: Related to above, can we simplify this to only compute weights at the very end:
```
   scaled_weights = (1 - self.alpha) * elbo_particles
   weights = (scaled_weights - log_sum_exp(scaled_weights) + math.log(self.num_particles)).exp()
```

pyro/infer/renyi_elbo.py
```diff
@@ -0,0 +1,244 @@
...
+                if site["type"] == "sample":
+                    log_prob, score_function_term, entropy_term = site["score_parts"]
+                    if is_vectorized:
+                        log_prob_sum = log_prob.reshape(self.num_particles, -1).sum(-1)
+                    else:
+                        log_prob_sum = site["log_prob_sum"]
+
+                    elbo_particle = elbo_particle - log_prob_sum.detach()
+
+                    if not is_identically_zero(entropy_term):
```
neerajprad: Can we just simply remove this, and throw `NotImplementedError` when both the entropy and score function terms are present?

pyro/infer/renyi_elbo.py
neerajprad: Thanks for explaining, @fehiepsi. Makes sense!

pyro/infer/renyi_elbo.py
```diff
@@ -0,0 +1,244 @@
...
+                        surrogate_elbo_particle = surrogate_elbo_particle - log_prob_sum
+
+                    if not is_identically_zero(score_function_term):
+                        surrogate_elbo_particle = (surrogate_elbo_particle +
+                                                   (self.alpha / (1. - self.alpha)) * log_prob_sum)
+
+            if is_identically_zero(elbo_particle):
+                if tensor_holder is not None:
```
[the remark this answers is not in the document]
eb8680: I wonder if we could use `util.Dice` to simplify this instead as in `TraceEnum_ELBO`?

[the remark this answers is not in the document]
neerajprad: hmm..interesting. I'll have to read more on the dice operator to comment.

tests/infer/test_inference.py
fehiepsi: I'll do it.

pyro/infer/renyi_elbo.py
```diff
@@ -0,0 +1,244 @@
...
+                        log_prob_sum = site["log_prob_sum"]
+
+                    elbo_particle = elbo_particle - log_prob_sum.detach()
+
+                    if not is_identically_zero(entropy_term):
+                        surrogate_elbo_particle = surrogate_elbo_particle - log_prob_sum
+
+                    if not is_identically_zero(score_function_term):
+                        surrogate_elbo_particle = (surrogate_elbo_particle +
+                                                   (self.alpha / (1. - self.alpha)) * log_prob_sum)
```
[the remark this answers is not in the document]
fehiepsi: We subtract this `1 / (1 - alpha)` to `1.` (from `f` in my gist) then get `alpha / (1 - alpha)`.

pyro/infer/renyi_elbo.py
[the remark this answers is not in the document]
fehiepsi: @neerajprad I use `tensor_holder` because I worry that for some particle, elbo_particle might be 0 (for any reason), and for other particles, elbo_particle are tensors. This is pretty cheap I guess.

I did have two containers: elbo_trace and surrogate_trace which stores what you suggested.

pyro/infer/renyi_elbo.py
fehiepsi: But I guess we don't want to operate `log_sum_exp` for the case `n=1`.

pyro/infer/renyi_elbo.py
fehiepsi: This computation is for `elbo` also. I don't think your suggestion will simplify things. Please correct me if I am wrong.

pyro/infer/renyi_elbo.py
fehiepsi: yes, I'll do that.

fehiepsi:
@neerajprad Did you mean that we will change
```
        if self.num_particles == 1:
            elbo = elbo_trace[0]
            surrogate_elbo_particles = surrogate_elbo_trace[0]
        else:
            if is_vectorized:
                elbo_particles = elbo_trace[0]
                surrogate_elbo_particles = surrogate_elbo_trace[0]
            else:
                elbo_particles = torch.stack(elbo_trace)
                surrogate_elbo_particles = torch.stack(surrogate_elbo_trace)

            elbo_particles_scaled = (1. - self.alpha) * elbo_particles
            elbo_scaled = log_sum_exp(elbo_particles_scaled, dim=0) - math.log(self.num_particles)
            elbo = elbo_scaled.sum().item() / (1. - self.alpha)
```
into this
```
        if not is_vectorized:
            elbo_particles = torch.stack(elbo_trace)
            surrogate_elbo_particles = torch.stack(surrogate_elbo_trace)
        else:
            elbo_particles = elbo_trace[0]
            surrogate_elbo_particles = surrogate_elbo_trace[0]
        
        if self.num_particles == 1:
            elbo = elbo_particles
        else:
            elbo_particles_scaled = (1. - self.alpha) * elbo_particles
            elbo_scaled = log_sum_exp(elbo_particles_scaled, dim=0) - math.log(self.num_particles)
            elbo = elbo_scaled.sum().item() / (1. - self.alpha)
```

pyro/infer/renyi_elbo.py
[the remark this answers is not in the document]
neerajprad: From what I can see, it should be safe to just consider the vectorized and non vectorized cases separately. In the vectorized case, we'll just have one element in the `elbo_particles` list and in the non-vectorized case each element should be a scalar (so it shouldn't matter if something is a 0).

pyro/infer/renyi_elbo.py
neerajprad: Once we have the weights, they can be used to return the elbo loss,  say something like this? Not sure about how much simpler it is, but seems more intuitive. :)

```
        scaled_weights = (1 - self.alpha) * elbo_particles
        log_mean_weight = log_sum_exp(scaled_weights) - math.log(self.num_particles)
        weights = (scaled_weights - log_mean_weight).exp()

        if trainable_params and getattr(surrogate_particles, 'requires_grad', False):
            surrogate_elbo = -(weights * surrogate_particles).sum() / self.num_particles
            surrogate_elbo.backward()

        loss = - (1. / (1. - self.alpha) * log_mean_weight)

        warn_if_nan(loss, "loss")
        return loss
```

neerajprad:
> @neerajprad Did you mean that we will change

I meant something like this (I could totally be oversimplifying! :) ) :

```
    def loss_and_grads(self, model, guide, *args, **kwargs):
        elbo_particles = []
        surrogate_particles = []
        is_vectorized = self.vectorize_particles and self.num_particles > 1

            ...
            # Append to containers
            surrogate_particles.append(surrogate_elbo_particle)
            elbo_particles.append(surrogate_elbo_particle)

        # handle vectorized / non vectorized case
        if is_vectorized:
            elbo_particles = elbo_particles[0]
            surrogate_particles = surrogate_particles[0]
        else:
            elbo_particles = torch.stack(elbo_particles)
            surrogate_particles = torch.stack(surrogate_particles)
        ...
        # compute weights
        scaled_weights = (1 - self.alpha) * elbo_particles
        log_mean_weight = log_sum_exp(scaled_weights) - math.log(self.num_particles)
        weights = (scaled_weights - log_mean_weight).exp()

        if trainable_params and getattr(surrogate_particles, 'requires_grad', False):
            surrogate_elbo = -(weights * surrogate_particles).sum() / self.num_particles
            surrogate_elbo.backward()
       
        # return losses
        loss = - (1. / (1. - self.alpha) * log_mean_weight)
        warn_if_nan(loss, "loss")
        return loss
```

pyro/infer/renyi_elbo.py
fehiepsi: now, I get what you mean. Will change things accordingly soon. :)

pyro/infer/renyi_elbo.py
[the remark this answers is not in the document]
fehiepsi: So you mean that we initiate `elbo_particle` with `torch.tensor(0.)` instead of `0.`? How could we know the data type and device then?

fehiepsi:
Thanks, @neerajprad! I will change it accordingly.

pyro/infer/renyi_elbo.py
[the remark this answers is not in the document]
neerajprad: I see the issue. For the non-vectorized case, you are right that there is a bit of an issue, since we may (rarely, but perhaps due to conditional execution), have a case where we have `elbo_particles` like `[0, torch.tensor(-1.3), torch.tensor(-2.2)]`.

neerajprad:
Thanks for all the work you put into this, @fehiepsi! Let me know if this is good to merge.

fehiepsi:
Yes, it is ready to me. :)

Issue #1212, opened 2018-07-03
List of updates from pytorch master


fehiepsi:
There are several updates from Pytorch master which are relevant to Pyro. So I put them here to have references for future pull requests.

- [x] pytorch has vectorized count Binomial, HalfCauchy, HalfNormal
- [x] pytorch LowRankMultivariateNormal does not have trace_term
- [x] logsumexp is available in Pytorch master
- [x] bo tutorial's code can be more stable after pytorch's lbfgs supporting line search (under reviewing)
- [x] we can remove `matrix_triangular_solve_compat` in Pyro and update SparseGPRegression to deal with trace_term because pytorch's lowrank_mvn does not support it
- [x] `tensor.new()` is depricated, use `.new_empty` instead

Some performance tips:
- [x] use `K.view(-1)[::n+1] += 1` to add identity matrix to a contiguous matrix K
- [x] ~use `x.size(0)` (faster) instead of `x.shape[0]` for a single element retrive from shape~ not important

eb8680:
Also, `torch.jit` has changed significantly, so `pyro.ops.jit` is basically obsolete.

fehiepsi:
@eb8680 Long time ago, you mentioned about a trick to add a term to log_prob from a model. Could you be more explicit about that trick? I am trying to move the `trace_term` [here](https://github.com/uber/pyro/blob/dev/pyro/contrib/gp/models/sgpr.py#L155) outside of `LowRankMultivariateNormal` distribution (to compatible to pytorch master). That `trace_term` is added directly to `log_prob` of lowrank_mvn.

Thanks!

eb8680:
> a trick to add a term to log_prob from a model

You can add an arbitrary factor to a model's trace by observing a Bernoulli distribution:
```python
pyro.sample("factor", dist.Bernoulli(logits=arbitrary_value), obs=1.)
```

fehiepsi:
Thank you @eb8680 ! I still can not write down the math here. I see that `log(sigmoid(logit))` is different to `logit`. I'll write things down again after lunch. Maybe I miss something.

eb8680:
I guess there's a bit of extra plumbing in `dist.Bernoulli`, but there's some manipulation such that `Bernoulli(...).log_prob(1.) == arbitrary_value`, which is what you're trying to achieve.   You could do something similar with `Categorical`.

fehiepsi:
Yeah, got it! Thanks a lot! :)

eb8680:
[...]

PR #1215, opened 2018-07-04
Revise pyro to match pytorch master

fehiepsi:
This pull request revises the following points
+ Set default values for `loc` and `scale` of HalfCauchy, so when updating, it is enough to delete the implementation of HalfCauchy.
+ Add `dim` argument for `log_sum_exp` operator, so when updating, it is enough to change its name to `torch.logsumexp`
+ Make `matrix_triagular_solve_compat` function hidden inside `lowrank_mvn`. So when updating, it is enough to delete the implementation of lowrank_mvn. 
+ Use `torch.trtrs` directly in GP module
+ Use `.new_tensor`, `.new_empty` instead of `.new` at various places
+ Make the sum of a matrix with a diagonal matrix faster by using `view` trick.

There is still one thing needed to be updated relevant to `trace_term` of lowrank_mvn. I will address it in the other pull request.

### DIFF
--- a/tests/contrib/gp/test_models.py
+++ b/tests/contrib/gp/test_models.py
@@ -330,9 +330,10 @@ def _pre_test_mean_function():
     def f(x):
         return 2 * x + 3 + 5 * torch.sin(7 * x)
 
-    X = torch.arange(100)
+    tensor_holder = torch.tensor([])
+    X = tensor_holder.new_tensor(torch.arange(100))
     y = f(X)
-    Xnew = torch.arange(100, 150)
+    Xnew = tensor_holder.new_tensor(torch.arange(100, 150))
     ynew = f(Xnew)
 
     kernel = Cosine(input_dim=1)
--- a/examples/baseball.py
+++ b/examples/baseball.py
@@ -102,7 +102,7 @@ def partially_pooled(at_bats):
     """
     num_players = at_bats.shape[0]
     loc = pyro.sample("loc", Normal(at_bats.new_tensor(-1), at_bats.new_tensor(1)))
-    scale = pyro.sample("scale", HalfCauchy(at_bats.new_tensor(0), at_bats.new_tensor(1)))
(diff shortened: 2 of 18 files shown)


tests/contrib/gp/test_models.py
```diff
@@ -330,9 +330,10 @@ def _pre_test_mean_function():
     def f(x):
         return 2 * x + 3 + 5 * torch.sin(7 * x)
 
-    X = torch.arange(100)
+    tensor_holder = torch.tensor([])
```
fritzo: Can you explain the change in this file? I don't understand why it's needed.

fehiepsi: I used pytorch master, and `torch.arange` returns long tensor instead of float as in pytorch 0.4.

eb8680:
@fehiepsi did you actually run any Pyro tests against PyTorch master?  I'm seeing a bunch of our tests in `tests/infer/test_valid_models.py` and `tests/infer/test_enum.py` failing due to deprecation warnings.

fehiepsi:
@eb8680 I just run the gp model tests and observe some deprecation warning about bce_with_logits. So I made a pull request in pytorch regarding that warning. I didn't run other Pyro tests.

I will run the tests in those two files to update more.

eb8680:
I'm not sure the failures have anything to do with this PR, but we probably don't want Pyro dev to be tracking PyTorch master too closely because of the additional maintenance burden.  I guess it's OK for `contrib` modules, though, and the changes in this PR and #1212 seem reasonable.

Also, FYI, those two test modules I mentioned are very useful for development - they have a lot of small, reliable tests that run very quickly and cover a large subset of core SVI and broadcasting functionality.  I run them all the time when I'm making changes to core Pyro to catch bugs I've introduced.

fehiepsi:
@eb8680 I run the test with pytorch master and do not observe the deprecate warnings for those two files. Maybe the warnings you got were related to Bernoulli distribution, which I have made a fix (which has been merged) in pytorch master.

PR #1220, opened 2018-07-06
Implement VonMises3D distribution

fritzo:
This implements the `.log_prob()` part of a `VonMises3D` distribution, similar to the 2D `VonMises` distribution.

This also fixes an attribute name bug (my fault) introduced in #1187.

## Tested

- [x] simple normalization test
- [x] tested on internal applications
- [ ] any other tests?

### DIFF
--- a/pyro/distributions/von_mises.py
+++ b/pyro/distributions/von_mises.py
@@ -45,6 +45,13 @@ class VonMises(TorchDistribution):
     """
     A circular von Mises distribution.
 
+    This implementation uses polar coordinates. The ``loc`` and ``value`` args
+    can be any real number (to facilitate unconstrained optimization), but are
+    interpreted as angles modulo 2 pi.
+
+    See :class:`~pyro.distributions.VonMises3D` for a 3D cartesian coordinate
+    cousin of this distribution.
+
     Currently only :meth:`log_prob` is implemented.
 
     :param torch.Tensor loc: an angle in radians.
--- a/pyro/distributions/von_mises.py
+++ b/pyro/distributions/von_mises.py
@@ -68,7 +68,7 @@ class VonMises(TorchDistribution):
         try:
             return super(VonMises, self).expand(batch_shape)
         except NotImplementedError:
-            validate_args = self.__dict__.get('validate_args')
+            validate_args = self.__dict__.get('_validate_args')
             loc = self.loc.expand(batch_shape)
             concentration = self.concentration.expand(batch_shape)
             return type(self)(loc, concentration, validate_args=validate_args)
--- a/pyro/distributions/von_mises_3d.py
+++ b/pyro/distributions/von_mises_3d.py
@@ -12,6 +12,12 @@ class VonMises3D(TorchDistribution):
     """
     Spherical von Mises distribution.
     https://en.wikipedia.org/wiki/Von_Mises%E2%80%93Fisher_distribution
+
+    Currently only :meth:`log_prob` is implemented.
+
+    :param torch.Tensor concentration: A combined location-and-concentration
+        vector. The direction of this vector is the location, and its
+        magnitude is the contration.
     """
     arg_constraints = {'concentration': constraints.real}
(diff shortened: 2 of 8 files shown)


alicanb:
Do you think it's worthwhile to merge `VonMises` and `VonMises3D`?

fritzo:
Do you mean "put them in the same file"? I think they're different enough to warrant separate classes, since we use polar coords for `VonMises` and cartesian coords for `VonMises3D`.

pyro/distributions/von_mises_3d.py
```diff
@@ -0,0 +1,50 @@
...
+class VonMises3D(TorchDistribution):
+    """
+    Spherical von Mises distribution.
+    https://en.wikipedia.org/wiki/Von_Mises%E2%80%93Fisher_distribution
+
+    Currently only :meth:`log_prob` is implemented.
+
+    :param torch.Tensor concentration: A combined location-and-concentration
+        vector. The direction of this vector is the location, and its
+        magnitude is the contration.
```
alicanb: contration -> concentration

pyro/distributions/von_mises.py
```diff
@@ -68,7 +68,7 @@ def expand(self, batch_shape):
         try:
             return super(VonMises, self).expand(batch_shape)
         except NotImplementedError:
-            validate_args = self.__dict__.get('validate_args')
+            validate_args = self.__dict__.get('_validate_args')
```
neerajprad: Thanks for fixing this, and sorry I should have caught this earlier!

tests/distributions/test_von_mises.py
```diff
@@ -14,3 +14,19 @@ def test_log_prob_normalized(concentration):
...
+@pytest.mark.parametrize('scale', [0.1, 0.5, 0.9, 1.0, 1.1, 2.0, 10.0])
+def test_von_mises_3d(scale):
+    concentration = torch.randn(3)
+    concentration = concentration * (scale / concentration.norm(2))
+
+    num_samples = 100000
+    samples = torch.randn(num_samples, 3)
+    samples = samples / samples.norm(2, dim=-1, keepdim=True)
+
+    d = VonMises3D(concentration, validate_args=True)
```
neerajprad: If we do not explicitly turn this flag to `True`, is the distributions arg checking disabled? We should have it enabled by default, unless there has been some regression lately.

pyro/distributions/von_mises_3d.py
```diff
@@ -0,0 +1,50 @@
...
+import torch
+from torch.distributions import constraints
+
+from pyro.distributions import TorchDistribution
+
+
+class VonMises3D(TorchDistribution):
+    """
+    Spherical von Mises distribution.
+    https://en.wikipedia.org/wiki/Von_Mises%E2%80%93Fisher_distribution
```
jpchen: nit: this would be better as a hyperlink. also we should have this for all distributions or none imo

tests/distributions/test_von_mises.py
fritzo: Good point, I'll update this.

fritzo:
> we could probably bake some logic over which Implementation to use

No, `VonMises` is in polar coords whereas `VonMises3D` is in Cartesian submanifold coords. If we ever generalize, it would be from `VonMises3D` to `VonMisesFisher` or something that would use Cartesian coords for all dimensions.

I'll add comments explaining this in both distributions, since it has already caused confusion twice.

pyro/distributions/von_mises_3d.py
fritzo: Sure, I'll add explanation and remove the link.

jpchen:
> VonMises is in polar coords whereas VonMises3D is in Cartesian submanifold coords.

ah i missed the discussion above, yeah i guess in that case it's better to fail than silently do the wrong thing, though i usually assume the user is cognizant of the nuances of args s/he passes to distributions

PR #1225, opened 2018-07-11
Separate trace from low rank mvn

fehiepsi:
The lowrank_mvn version in PyTorch will not have trace_term. To remedy it, I use the trick suggested by @eb8680 to add trace_term to the log_prob.

In addition, in PyTorch, covariance_matrix = W @ W.t() + D. In our lowrank_mvn, covariance_matrix = W.t() @ W + D. So I also make a small fix for it.

### DIFF
--- a/pyro/contrib/gp/models/sgpr.py
+++ b/pyro/contrib/gp/models/sgpr.py
@@ -135,14 +135,13 @@ class SparseGPRegression(GPModel):
         W = Kuf.trtrs(Luu, upper=False)[0]
 
         D = noise.expand(W.shape[1])
-        trace_term = 0
         if self.approx == "FITC" or self.approx == "VFE":
             Kffdiag = self.kernel(self.X, diag=True)
             Qffdiag = W.pow(2).sum(dim=0)
             if self.approx == "FITC":
                 D = D + Kffdiag - Qffdiag
             else:  # approx = "VFE"
-                trace_term += (Kffdiag - Qffdiag).sum() / noise
+                trace_term = (Kffdiag - Qffdiag).sum() / noise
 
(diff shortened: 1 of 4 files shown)


pyro/distributions/lowrank_mvn.py
```diff
@@ -49,6 +49,7 @@ class LowRankMultivariateNormal(TorchDistribution):
     has_rsample = True
 
     def __init__(self, loc, W_term, D_term, trace_term=None):
+        W_term = W_term.t()
```
fehiepsi: This is a simple fix. Otherwise, we have to change the computation of log_prob, covariance_matrix,... which is unnecessary because we will delete this implementation at the end.

fritzo: Please also update the docstring above and error message below, so as to inform existing users of the breaking change you are making.

pyro/contrib/gp/models/sgpr.py
```diff
@@ -135,24 +135,28 @@ def model(self):
         W = Kuf.trtrs(Luu, upper=False)[0]
```
fritzo: Would be cleaner to make this `W` agree with the eventual upstream `LowRankMultivariateNormal` and define
```py
W = Kuf.trtrs(Luu, upper=False[0].t()

D = noise.expand(W.shape[0])
...
Wffdiag = W.pow(2).sum(dim=1)
...
f_var = D + W.pow(2).sum(dim=1)
...
dist.LowRankMultivariateNormal(f_loc, W, D)
```
?

fehiepsi: Yup! I'll change things in autoguide too. :)

PR #1228, opened 2018-07-13
Numerically stable log slice computation in NUTS

### LINKED ISSUE #1174: [Bug?] HMC and NUTS are too sensitive to step-size for simplle Gamma/Beta model
### Issue Description

 A very simple generative model---a Beta likelihood whose two parameters have a Gamma prior on them:

$$\alpha\sim Gamma(1, 1)$$
$$\beta\sim Gamma(1, 1)$$
$$x\sim Beta(\alpha, \beta)$$

and then running `NUTS` to perform MCMC inference on it, leads to
the acceptance probability to be 0 poor and hence garbage estimates for $\alpha$ and $\beta$.

Changing the stepsize in `HMC`fixes the issue, but `NUTS` should be more robust to this scenario.

Plus, without enabling debugging log messages, the fact that 0 uncorrelated samples were actually provided was totally hidden. I guess I should open a new issue for a feature request for better output?

This issue has been referenced also in this forum post https://forum.pyro.ai/t/simple-mcmc-on-gamma-beta-providing-wrong-estimates/259

### Environment
 - Ubuntu `17.04`, python `3.6`
 - PyTorch `0.4.0`
 - Pyro `0.2.1`

### Code Snippet
```
import numpy as np
import scipy.stats as ss
import torch
import pyro
import pyro.infer.mcmc as mcmc
from pyro.infer.abstract_infer import EmpiricalMarginal


a = 5
b = 1
n_samples = 1000
beta_data = ss.beta(a=a, b=b).rvs(n_samples).astype(np.float32)
print(beta_data.shape)
plt.hist(beta_data, bins=100)
plt.show()

def beta_model(data):
    alpha_0 = torch.ones(1)
    beta_0 = torch.ones(1)
    alpha_prior = pyro.sample('alpha', dist.Gamma(concentration=alpha_0, rate=beta_0))
    beta_prior = pyro.sample('beta', dist.Gamma(concentration=alpha_0, rate=beta_0))
    x = pyro.sample('x', dist.Beta(concentration1=alpha_prior, concentration0=beta_prior), obs=data)
    return x

# hmc_kernel = mcmc.HMC(beta_model, step_size=0.0855, num_steps=4)
nuts_kernel = mcmc.NUTS(beta_model, adapt_step_size=True)
mcmc_run = mcmc.MCMC(nuts_kernel, num_samples=1000, warmup_steps=10000).run(torch.from_numpy(beta_data))
beta_posterior = EmpiricalMarginal(mcmc_run, 'beta')
print('beta', beta_posterior.mean) 
alpha_posterior = EmpiricalMarginal(mcmc_run, 'alpha')
print('alpha', alpha_posterior.mean)
```

neerajprad:
Fixes #1174. 

We were having problems with certain models (e.g. in #1174) where the acceptance probability was extremely low in NUTS (with or without step size adaptation). The issue was that if we are in a state such that the energy term is a high negative number (similar issue with high positive value), then `exp(-energy)` blows up to infinity. This results in a inf `log_slice` term that would result in rejections no matter what the new state is (the base tree itself would be marked as diverging). 

This makes a minor change to the computation of the `log_slice` variable in NUTS so that we do not need to exponentiate the energy term, but can use it directly with the existing energy term which can have very high/low values (e.g. +/- 400). The trick (as mentioned by Neal in his Slice Sampling paper) is that instead of sampling a uniform random from `U(0, e(-energy))` and taking the log, we instead sample an auxiliary z from an exponential distribution with mean 1, and use `log u = -energy - z, where z ~ Exp(1)`. Tested using the model from #1174, which works now. 

Additionally, some minor clean up: 
 - Changed the test names to correspond to the convention of specifying the latent variable first. We have a few tests that use the opposite convention in `test_inference.py` which we should change as well.
 - Another minor change is to disable arg validation during the initial step size adaptation in setup (it is already disabled during the adaptation phase).

### DIFF
--- a/tests/infer/mcmc/test_hmc.py
+++ b/tests/infer/mcmc/test_hmc.py
@@ -171,7 +171,7 @@ def test_logistic_regression():
     assert_equal(rmse(true_coefs, beta_posterior.mean).item(), 0.0, prec=0.1)
 
 
-def test_bernoulli_beta():
+def test_beta_bernoulli():
     def model(data):
         alpha = torch.tensor([1.1, 1.1])
         beta = torch.tensor([1.1, 1.1])
@@ -187,7 +187,7 @@ def test_bernoulli_beta():
     assert_equal(posterior.mean, true_probs, prec=0.05)
 
 
-def test_normal_gamma():
+def test_gamma_normal():
     def model(data):
         rate = torch.tensor([1.0, 1.0])
         concentration = torch.tensor([1.0, 1.0])
@@ -203,7 +203,7 @@ def test_normal_gamma():
     assert_equal(posterior.mean, true_std, prec=0.05)
 
 
-def test_categorical_dirichlet():
+def test_dirichlet_categorical():
     def model(data):
         concentration = torch.tensor([1.0, 1.0, 1.0])
         p_latent = pyro.sample('p_latent', dist.Dirichlet(concentration))
@@ -236,7 +236,7 @@ def test_logistic_regression_with_dual_averaging():
     assert_equal(rmse(posterior.mean, true_coefs).item(), 0.0, prec=0.1)
 
 
-def test_bernoulli_beta_with_dual_averaging():
+def test_beta_bernoulli_with_dual_averaging():
     def model(data):
         alpha = torch.tensor([1.1, 1.1])
         beta = torch.tensor([1.1, 1.1])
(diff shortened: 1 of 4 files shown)


tests/infer/mcmc/test_hmc.py
```diff
@@ -171,7 +171,7 @@ def model(data):
     assert_equal(rmse(true_coefs, beta_posterior.mean).item(), 0.0, prec=0.1)
 
 
-def test_bernoulli_beta():
+def test_beta_bernoulli():
```
fritzo: Ha ha this has always annoyed me, thanks for swapping 😆

neerajprad: Same here, it's been on my mind for some time. :)

neerajprad:
Thanks for reviewing, @fritzo, @fehipsi.

PR #1229, opened 2018-07-13
Add RSA tutorials

ngoodman:
Adding a pair of tutorials explaining the RSA implicature and hyperbole examples. Code is very similar to that in examples, and uses the Search implementation from there (which should be moved into core at some point).


neerajprad:
> @neerajprad does this need to be scrubbed of output?

The following commands from our Makefile will scrub these tutorials:
```
find tutorial -name "RSA*.ipynb" | xargs python -m nbstripout --keep-output --keep-countpython 
find tutorial -name "RSA*.ipynb" | xargs python tutorial/source/cleannb.py
```

Alternatively, we can merge this, and I can scrub all the tutorials in a single pass after that, as there are some minor changes introduced by scrubbing in all our tutorials.

eb8680:
> Alternatively, we can merge this, and I can scrub all the tutorials in a single pass after that

Let's do this

PR #1230, opened 2018-07-16
Consolidate *ELBO._get_trace

eb8680:
Part of #1234 

The ELBO implementations have some duplicated functionality in their `_get_trace`/`_get_traces` methods.  This PR collects some of that functionality (mostly `_get_trace`) into one place and reuses it across the different ELBOs.

`TraceEnum_ELBO` is still a bit awkward because I wanted to limit the scope of the changes in this PR and avoid changing tests and validation logic.  I'd like to continue reorganizing the SVI code in followup PR(s), starting with removing `pyro.infer.enum.iter_discrete_traces` which is now deprecated.

### DIFF
--- a/pyro/infer/traceenum_elbo.py
+++ b/pyro/infer/traceenum_elbo.py
@@ -12,7 +12,7 @@ import pyro.ops.jit
 import pyro.poutine as poutine
 from pyro.distributions.util import is_identically_zero
 from pyro.infer.elbo import ELBO
-from pyro.infer.enum import _get_importance_trace, _iter_discrete_escape, _iter_discrete_extend
+from pyro.infer.enum import get_importance_trace, iter_discrete_escape, iter_discrete_extend
 from pyro.infer.util import Dice, is_validation_enabled
 from pyro.util import check_traceenum_requirements, warn_if_nan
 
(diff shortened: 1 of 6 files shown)


pyro/infer/traceenum_elbo.py
```diff
@@ -11,10 +12,9 @@
 import pyro.poutine as poutine
 from pyro.distributions.util import is_identically_zero
 from pyro.infer.elbo import ELBO
-from pyro.infer.enum import iter_discrete_traces
+from pyro.infer.enum import _get_importance_trace, _iter_discrete_escape, _iter_discrete_extend
```
fritzo: nit: If we're importing these across modules, I think it makes sense to remove the leading underscore.

pyro/infer/traceenum_elbo.py
```diff
@@ -74,55 +74,43 @@ def _get_trace(self, model, guide, *args, **kwargs):
...
-                for model_trace, guide_trace in self._get_trace(model, guide, *args, **kwargs):
-                    yield model_trace, guide_trace
+
+        # enable parallel enumeration over the vectorized guide.
+        guide = poutine.enum(guide, first_available_dim=self.max_iarange_nesting)
+        q = queue.LifoQueue()
+        guide = poutine.queue(guide, q,
+                              escape_fn=_iter_discrete_escape,
+                              extend_fn=_iter_discrete_extend)
+        for i in range(1 if self.vectorize_particles else self.num_particles):
```
fritzo: nice!

fritzo:
It looks like only the `JitTrace*_ELBO` tests are failing. Do you want to simply delete our outdated JIT work? Also Soumith wants to meet to discuss Jit work in the near future, to make sure our team knows their longer term plans.

neerajprad:
Ready to merge?

eb8680:
Yep

Issue #1227, opened 2018-07-13
Cleaning up SVI internals to support nested inference


eb8680:
The internals of our SVI code need a bit of cleanup to better support nesting `SVI` inside other inference algorithms, a necessary component of several projects including `contrib.oed`.

Tasks:
- [ ] Make `infer.SVI` a `TracePosterior` instance again so that it works with `EmpiricalMarginal` and can be interchanged with other methods like `MCMC`
- [ ] Split side effect-free `differentiable_loss` methods, which would only compute surrogate losses, out of the various `*ELBO.loss_and_grads` methods, which both compute surrogate losses and call `backward`, so that gradients aren't computed twice in `SVI.step` when using `pyro.optim.multi.PyroMultiOptimizer`
- [ ] Collect all the duplicated `*ELBO._get_traces` logic into a single set of functions in `infer.enum`, as suggested in e.g. #1214.  Ideally we could also use the same code as part of #1226.

fritzo:
Note regarding #628 and https://forum.pyro.ai/t/checkpoint-in-pyro/303:
There are some disadvantages of using `grad` instead of `.backward()` (at least in `Trace_ELBO`):
- `grad` has less `jit` support at the moment
- `grad` has no checkpointing support at the moment
- `grad` is slightly more expensive for tiny tensors (since it traverses the graph twice)

However these are minor and could be outweighed by code simplification.

PR #1231, opened 2018-07-16
Make SVI a TracePosterior subclass

eb8680:
Part of #1227 

This PR makes SVI a `TracePosterior` instance again, allowing it to be used with `EmpiricalMarginal` and more easily swapped out with `MCMC` and other inference algorithms.  It also updates `contrib.oed` to reflect this change.

### DIFF
--- a/pyro/contrib/oed/eig.py
+++ b/pyro/contrib/oed/eig.py
@@ -1,0 +1,64 @@
+import pyro
+from pyro.contrib.oed.search import Search
+from pyro.infer import EmpiricalMarginal, Importance, SVI, Trace_ELBO
+from pyro.contrib.autoguide import mean_field_guide_entropy
+
+
+def vi_ape(model, design, observation_labels, vi_parameters, is_parameters):
+    """Estimates the average posterior entropy (APE) loss function using
+    variational inference (VI).
+
+    The APE loss function estimated by this method is defined as
+
+        :math:`APE(d)=E_{Y\\sim p(y|\\theta, d)}[H(p(\\theta|Y, d))]`
+
+    where :math:`H[p(x)]` is the `differential entropy
+    <https://en.wikipedia.org/wiki/Differential_entropy>`_.
+    The APE is related to expected information gain (EIG) by the equation
+
+        :math:`EIG(d)=H[p(\\theta)]-APE(d)`
+
+    in particular, minimising the APE is equivalent to maximising EIG.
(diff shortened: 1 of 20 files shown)


fritzo:
@eb8680 I think your implementation is correct. It looks like your test failures are due to new stricter plumbing that exposes latent bugs (probably my fault). Let's blame these errors on the dangerous practice of discarding unused `**kwargs`.

ae-foster:
Looks good @eb8680 - should you write a test for this new way of running SVI (with `.run(...)`) or update any old SVI tests? From the OED side, things looks much cleaner.

pyro/infer/svi.py
```diff
@@ -51,6 +59,19 @@ def _loss_and_grads(*args, **kwargs):
...
+            with poutine.block():
+                for i in range(self.num_steps):
+                    self.step(*args, **kwargs)
+        return super(SVI, self).run(*args, **kwargs)
+
+    def _traces(self, *args, **kwargs):
+        for i in range(self.num_samples):
+            guide_trace = poutine.trace(self.guide).get_trace(*args, **kwargs)
+            model_trace = poutine.trace(poutine.replay(self.model, trace=guide_trace)).get_trace(*args, **kwargs)
+            yield model_trace, (model_trace.log_prob_sum() - guide_trace.log_prob_sum())
```
neerajprad: We can probably set the importance weights to some constant here, right (seems like double adjustment otherwise)?

eb8680: Yeah, you're right

pyro/contrib/oed/eig.py
```diff
@@ -45,17 +46,12 @@ def posterior_entropy(y_dist, design):
         y = pyro.sample("conditioning_y", y_dist)
         y_dict = {label: y[i, ...] for i, label in enumerate(observation_labels)}
         conditioned_model = pyro.condition(model, data=y_dict)
-        posterior = SVI(conditioned_model, vi_parameters["guide"],
-                        vi_parameters["optim"], loss=Trace_ELBO())
-        with pyro.poutine.block():
-            for _ in range(vi_parameters["num_steps"]):
-                posterior.step(design)
+        posterior = SVI(conditioned_model, **vi_parameters).run(design)  # noqa: F841
```
fritzo: nit: Why not remove the `posterior =` as suggested by flake8?

PR #1233, opened 2018-07-16
Scrub tutorials to ensure uniform metadata

neerajprad:
Ran `make scrub` - all changes just ensure that the default kernel uses python 3.6.


eb8680:
Did this miss the RSA tutorials, or are they correct already?

neerajprad:
> Did this miss the RSA tutorials, or are they correct already?

Ha..it seems to be already correct!

PR #1237, opened 2018-07-17
Implement Laplace approximation

fehiepsi:
This pull request implements Laplace approximation, which is popular is statistics' textbook. A long time ago, I raised an [issue](https://github.com/uber/pyro/issues/583) for an attempt to implement it in Pyro. But it seems that it can be achieved without having to refer to that issue.

Here is to-do list:
- [x] Implement Laplace approximation
- [x] Test hessian method
- [x] Test inference


fehiepsi:
> it appears that your implementation does not attempt to find the mode

@fritzo I thought that using SVI with AutoDelta guide would give us the MAP point, which is also the mode?

pyro/contrib/autoguide/__init__.py
```diff
@@ -297,6 +294,59 @@ def median(self, *args, **kwargs):
         return self(*args, **kwargs)
 
 
+def _hessian(y, xs):
+    dys = torch.autograd.grad(y, xs, create_graph=True)
+    flat_dy = torch.cat([dy.reshape(-1) for dy in dys])
+    H = []
+    for dyi in flat_dy:
+        Hi = torch.cat([Hij.reshape(-1) for Hij in torch.autograd.grad(dyi, xs, retain_graph=True)])
```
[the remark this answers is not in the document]
fehiepsi: Yup, what I understand from autograd's docs is that: With `create_graph` enable, backward will build the graph to compute gradients from inputs; for second-order derivatives, we don't need to enable `create_graph` because we don't want to compute third-order derivatives here. However, we have to enable `retain_graph` because we will use the graph again and again for each `dyi`.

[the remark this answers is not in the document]
fehiepsi: It seems that things will be freed if we don't use `retain_graph`.
```
x = torch.rand(3, requires_grad=True)
y = x ** 2
z = torch.autograd.grad(y[0], x)
t = torch.autograd.grad(y[1], x)  # <- RuntimeError
```
We get RuntimeError said that "Trying to backward from the graph second time, but the buffers have already been freed. Specify `retain_graph=True` when calling backward the first time".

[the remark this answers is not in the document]
fritzo: Thanks for explaining!

fritzo:
> I thought that using SVI with AutoDelta guide would give us the MAP point

You're right that SVI with `AutoDelta` would find the mode, and that this mode could then be used in a Laplace approximation by either overriding the `.get_posterior()` method (as in this PR) or by converting to a `MultivariateNormal` distribution (as I've suggested). However your implementation overrides `.get_posterior()` *before* training, so that `AutoLaplace` will *not* find the mode. This is why I suggest splitting up training (via `AutoDelta`) followed by prediction (via the `AutoMultivariateNormal` returned by `AutoDelta.laplace_approximation()`). Does that make sense?

fehiepsi:
@fritzo The AutoLaplace is a subclass of AutoDelta, which does not have `get_posterior` method. SVI has nothing to do with this `get_posterior` method (which is different from AutoContinuous where SVI uses `get_posterior` to sample latents).

fritzo:
Oh I think I see now. But `AutoDelta.get_posterior()` is not a useful interface. Instead you should create an `AutoMultivariateNormal` instance that handles unpacking, offers moments and quantiles, and can be used directly as a guide.

fritzo:
Re: constrained vs unconstrained, I'd prefer to yield constrained samples, so that we create a Laplace approximation in the transformed space. Otherwise samples from the Laplace approximation may be invalid (may raise errors).

fehiepsi:
Yes, I think that it is the problem of Laplace approximation. That's why I put the note in docs that the support is always real. Doing MAP in unconstrained space is different to doing MAP in constrained space. Similar for Laplace. If we agree to do inference in unconstrained space, then it is nice to follow your suggestion. Let me think for a while which is better (one for literature, and one for practical).

pyro/contrib/autoguide/__init__.py
```diff
@@ -601,6 +621,64 @@ def get_posterior(self, *args, **kwargs):
...
+        pyro.param("auto_loc", torch.randn(latent_dim))
+    """
+
+    def get_posterior(self, *args, **kwargs):
+        """
+        Returns a diagonal Normal posterior distribution transformed by
+        :class:`~pyro.distributions.iaf.InverseAutoregressiveFlow`.
+        """
+        loc = pyro.param("{}_loc".format(self.prefix),
+                         lambda: torch.zeros(self.latent_dim))
```
fehiepsi: @fritzo I don't know why we use `lambda` in other autoguide classes, so I keep that pattern. Could you let me know the reason?

pyro/contrib/autoguide/__init__.py
```diff
@@ -296,6 +304,22 @@ def median(self, *args, **kwargs):
...
+        model_trace = poutine.trace(
+            poutine.replay(self.model, trace=guide_trace)).get_trace(*args, **kwargs)
+        loss = -model_trace.log_prob_sum()
+
+        latents = []
+        for _, site in guide_trace.iter_stochastic_nodes():
+            latents.append(site["value"])
+
+        H = _hessian(loss, latents)
+        return torch.inverse(H)
```
fehiepsi: @fritzo I keep this functionality in case some user (like me) wants to do Laplace approximation in constrained space. I hope that it is fine for you.

tests/contrib/autoguide/test_inference.py
```diff
@@ -90,14 +92,24 @@ def model():
         loss = svi.step()
         assert np.isfinite(loss), loss
 
-    loc, scale = guide._loc_scale()
+    if auto_class is AutoDelta:
+        latents = guide.median()
+        loc = torch.tensor([latents["x"], latents["y"]])
+        scale = guide.covariance().diag().sqrt()
```
fehiepsi: test for `AutoDelta.covariance`

tests/contrib/autoguide/test_inference.py
```diff
@@ -90,14 +92,24 @@ def model():
...
+    else:
+        if auto_class is AutoLaplace:
+            guide = guide.laplace_approximation()
+
+        loc, scale = guide._loc_scale()
+
     assert_equal(loc, torch.tensor([-0.2, 0.2]), prec=0.05,
                  msg="guide mean off")
-    assert_equal(scale, torch.tensor([1.21, 0.71]), prec=0.08,
+    assert_equal(scale, torch.tensor([1.2, 0.7]), prec=0.08,
```
fehiepsi: it seems to me that `1.21` and `0.71` are mistakes

pyro/contrib/autoguide/__init__.py
fritzo: The `lambda` is a minor optimization that [avoids construction of the tensor](https://github.com/uber/pyro/blob/93c4038/pyro/params/param_store.py#L96) except on the first call. Usually we prefer clarity over minor optimizations, but since this is an internal library I opted for the cheaper lazy initialization.

pyro/contrib/autoguide/__init__.py
```diff
@@ -296,6 +304,22 @@ def median(self, *args, **kwargs):
         """
         return self(*args, **kwargs)
 
+    def covariance(self, *args, **kwargs):
+        """
+        Returns covariance of latent variables under Laplace (quadratic) approximation.
```
fritzo: Could you mention that this is the packed, transformed variables rather than the user facing variables? One thing I've attempted to do in this library is make a clean separation between the internal transformed packed variables and external user-facing variables. This distinction is very important while getting models to work in practice, since it is helpful to inspect values and statistics of the external variables, and I don't want to confuse modelers with internal developer-facing functions. Consider even renaming this to `_covariance()`

pyro/contrib/autoguide/__init__.py
```diff
@@ -601,6 +621,64 @@ def get_posterior(self, *args, **kwargs):
...
+    the user should call :func:`pyro.param` before beginning inference, e.g.::
+
+        latent_dim = 10
+        pyro.param("auto_loc", torch.randn(latent_dim))
+    """
+
+    def get_posterior(self, *args, **kwargs):
+        """
+        Returns a diagonal Normal posterior distribution transformed by
+        :class:`~pyro.distributions.iaf.InverseAutoregressiveFlow`.
```
fritzo: Update comment

pyro/contrib/autoguide/__init__.py
fritzo: Why don't you go further and make this into a method that returns a constrained `AutoMultivariateNormal` guide?

pyro/contrib/autoguide/__init__.py
fehiepsi: oppss... thanks, @fritzo

pyro/contrib/autoguide/__init__.py
fehiepsi: I am afraid that it would be complicated. The `AutoMultivariateNormal` works on unconstrained variables. Things will be wrong if we use this `covariance` as the covariance of AutoMultivariateNormal class. In addition, as you have pointed out that this Laplace approximation can give wrong samples (not belong to the support), would it be better to avoid it? By this method, I just keep the most important statistics from Laplace approximation: the covariance. Users can use it to see the covariance, correlation, or get samples at their own risk.

pyro/contrib/autoguide/__init__.py
fehiepsi: I think the current AutoDelta works on transformed variables (no log_density), so we don't need to mention about "transformed" thing here. I agree that it is better to mention that this is the covariance of the pack of flatted variables, arranged according to their appearance in `model`. Will add it tomorrow. :)

What do you think about making this hidden `_get_full_covariance()`, saving the result into a hidden property `_full_covariance` and expose another method like
```
def covariance(self, varname1, varname2):
    # return the covariance matrix of var1 and var2, which is part of self._full_covariance
```
This way we don't have to mention about "packed" thing.

neerajprad:
> I still feel that the unconstrained Laplace approximation is un-pyronic insofar as it ignores the support of random variables and thereby produces an approximate posterior with infinite KL divergence from the true posterior. I feel that by disregarding type information we would be perpetuating a legacy of statistics-done-wrong. Can anyone else weigh in on this? @eb8680 @neerajprad @alicanb ?

If I understood correctly, doing Laplace approximation in the unconstrained space, and then using methods from `AutoContinuous` like `median` or `__call__` will give us posterior samples from the constrained space, but we are worried that the resulting Laplace approximation isn't well grounded? On the other hand, if we do Laplace approximation in the original constrained space, the posterior samples will have real support, not the original constrained support. This is apparently what Edward does, as @fehiepsi mentioned. I am not sure what would be a way to generate constrained samples using this latter strategy (as a result, the approximation may still have infinite KL) other than perhaps using a truncated multivariate normal implementation, which we don't have.

fehiepsi:
@neerajprad Almost agree with you except

> we are worried that the resulting Laplace approximation isn't well grounded

which I have no opinion on it.

fehiepsi:
@fritzo That would be nice! Indeed, I want ConstrainedLaplaceApproximation for my own use (I want to see the correlation of latent variables) but have no idea to get samples with right supports from the approximation. Waiting to see how you resolve it. :)

fehiepsi:
@fritzo It would be better to merge this and rename in a follow-up because I believe you will write a better doc than me. ^^

PR #1238, opened 2018-07-17
Tracking examples

alicanb:


examples/contrib/tracking/plot_utils.py
```diff
@@ -0,0 +1,71 @@
...
+def init_visdom(visdom_flag):
+    if visdom_flag:
+        try:
+            from visdom import Visdom
+            viz = Visdom()
+            startup_sec = 1
+            while not viz.check_connection() and startup_sec > 0:
+                time.sleep(0.1)
+                startup_sec -= 0.1
+            assert viz.check_connection(), 'No connection could be formed quickly'
```
fritzo: nit: I think this would be simpler as
```py
if viz.check_connection():
    return viz
else:
    warnings.warn('No connection could be formed quickly')
    return None
```

examples/contrib/tracking/tracking_svi_em.py
```diff
@@ -0,0 +1,286 @@
...
+    p_exists = assignment.exists_dist.probs
+    positions = get_positions(states_loc, args.num_frames)
+    plot_solution(observations, p_exists, positions, true_positions, args,
+                  'after 10 EM (with prune and merge)', viz=viz)
+    plot_exists_prob(p_exists, viz)
+
+
+def arg_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--num_frames', default=40, type=int, help='number of frames')
```
fritzo: nit: argparsae convention is to use dashes rather than underscores, and let the library automatically convert dashes to underscores. So you can write
```py
parser.add_argument('--max-num-objects', ...)
```
and still use python identifiers on the result
```py
args.max_num_objects
```

examples/contrib/tracking/tracking_svi_em.py
```diff
@@ -0,0 +1,286 @@
+from __future__ import absolute_import, division, print_function
```
fritzo: :+1: This looks much easier to collaborate on than Jupyter notebooks, thanks for figuring out how to do this!

examples/contrib/tracking/plot_utils.py
alicanb: I wanted to catch all the `*Error`s urllib throws if visdom is not running but I can't catch them no matter what I do 😄

fritzo: I'm not sure why you're catching an `AssertionError`. Maybe you could broaden to `Exeption`
```py
try:
    ...
except Exception as e:
    warnings.warn('Failed to create Visdom instance: {}'.format(e))
    return None
```

pyro/util.py
```diff
@@ -52,7 +52,7 @@ def warn_if_nan(value, msg=""):
     also works with numbers.
     """
     if torch.is_tensor(value) and value.requires_grad:
-        value.register_hook(lambda x: warn_if_nan(x, msg))
+        value.register_hook(lambda x: warn_if_nan(x, "grad " + msg))
```
alicanb: I find this to be useful when checking which grad is problematic (e.g. warnings for 2nd derivative say "grad grad myvar")

fritzo: Feel free to submit this as a 1-line PR :smile:

PR #1210, opened 2018-06-28
Implement 3d and 1d Newton solvers

jpchen:
Addresses #1207 

implements:
- 3d newton solver
- 1d newton solver
- shim to handle different `x` dims
- refactor tests

+@fritzo

Issue #1213, opened 2018-07-03
Support independence assumptions about params


fritzo:
We'd like to support batched second-order optimizers, where the loss Hessian has user-declared block diagonal structure.

One natural way to declare the param independence assumption is to define `param`s inside `iarange` contexts. Whereas the old issue #238 suggests adding broadcasting behavior to param-inside-iarange, this new issue suggests adding a semantic assumption that the resulting ELBO optimization problem should factor into blocks. Note that this assumption is a no-op when using first-order optimizers, and only arises when using higher-order optimizers like `newton_step()`.

The motivating use case is nested inference with an EM inner loop inside an SVI outer loop. In this case we've often hand-implemented the EM inner loop to solve a batch of low-dimensional problems via `newton_step()`. A more Pyronic way would be to implement the M-step using Pyro's `TraceEnum_ELBO`, something like
```py
def model_for_m_step(data, discrete_probs_from_e_step):
    with pyro.iparam("independent_subproblems", theta_dim):
        theta = pyro.param("theta", ...)
        assert theta.shape[-1] == theta_dim
        z = pyro.sample("discrete_variable",
                        dist.Categorical(discrete_probs_from_e_step),
                        infer={"enumerate": "parallel"})
        pyro.sample("obs", dist.Something(theta), obs=data[z])
        
def guide_for_m_step(data, discrete_probs_from_e_step):
    pass

elbo = TraceEnum_ELBO(max_iarange_nesting=1)
for _ in range(em_steps):
    discrete_probs_from_e_step = ...
    loss = elbo.differentiable_loss(model_for_m_step,
                                    guide_for_m_step,
                                    data, discrete_probs_from_e_step)
    theta_old = pyro.param("theta")
    theta_new = theta_old + newton_step(loss, theta_old)
    pyro.get_parm_store().replace_param("theta", theta_old, theta_new)
```

neerajprad:
@fritzo, @eb8680: I was trying to understand what changes would we need in SVI, but I'm likely missing a few things:

> this new issue suggests adding a semantic assumption that the resulting ELBO optimization problem should factor into blocks. 

If `pyro.param("theta", ...)` still returns the global `theta`, then whether the optimization problem factorizes into sub-components would be determined by what parts of `theta` the subsequent `pyro.sample(..)` statements using them index into. Correct me if I am mistaken. Since the model above isn't using `theta`, its not very clear to me.

[...]

PR #1239, opened 2018-07-17
Refactor to use generic newton solver in pyro.optim.multi

fritzo:
This is the first step in refactoring our multi-optimizers
1. Switches from `Newton2d` to a more general `Newton` optimizer, following #1210 .
2. Changes the example and test to use `trace.nodes` rather than `pyro.get_param_store()`. The class interface remains unchanged.

We may follow this PR up with a second PR to switch dicts from name-to-tensor to name-to-site, but I'm not totally sure this is necessary.

### DIFF
--- a/pyro/optim/multi.py
+++ b/pyro/optim/multi.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 import torch
 
-from pyro.ops.newton import newton_step_2d
+from pyro.ops.newton import newton_step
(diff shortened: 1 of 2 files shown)


eb8680:
> We may follow this PR up with a second PR to switch dicts from name-to-tensor to name-to-site, but I'm not totally sure this is necessary.

Does that mean you're not planning on adding a general solution for #1213 ?

Issue #1232, opened 2018-07-16
Expose distribution from `sample_latent` in `AutoContinuous`


fehiepsi:
The current logic of all autoguide's method is:
```py
def sample_latent(self, *args, **kwargs):
    param = pyro.param(...)
    dist = dist.Distribution(param)
    return pyro.sample(name, dist, infer={"is_auxiliary": True})
```
It would be nice if we return `dist` in some method (e.g. `get_dist` / `get_posterior`) and refactor `sample_latent` into
```py
def sample_latent(self, *args, **kwargs):
    dist = self.get_dist(*args, **kwargs)
    return pyro.sample("_{}_latent".format(self.prefix), dist, infer={"is_auxiliary": True})
```

With the method `get_dist`, users can draw samples from this posterior or calculate other statistics.

eb8680:
Seems like a nice enhancement, feel free to open a PR

eb8680:
Resolved by #1240

PR #1240, opened 2018-07-17
Expose distribution from `sample_latent` in autoguide

fehiepsi:
Address #1232.

Please feel free to suggest other names for `get_posterior` method.

### DIFF
--- a/pyro/contrib/autoguide/__init__.py
+++ b/pyro/contrib/autoguide/__init__.py
@@ -85,6 +85,12 @@ class AutoGuide(object):
         """
         raise NotImplementedError
 
+    def get_posterior(self, *args, **kwargs):
+        """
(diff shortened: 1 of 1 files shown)


pyro/contrib/autoguide/__init__.py
```diff
@@ -85,6 +85,12 @@ def __call__(self, *args, **kwargs):
         """
         raise NotImplementedError
 
+    def get_posterior(self, *args, **kwargs):
```
fritzo: I believe `AutoGuide` should not have this method. The method appears specific to `AutoContinuous`. And certainly the method should not return None.

fehiepsi: Yup, agree. :)

PR #1241, opened 2018-07-17
 Pyro lr scheduler working perfectly and not at all

idostyle:
Intended to be a bug report.
It perfectly schedules the learning rate, but it just doesn't do any optimization steps.


jpchen:
thanks for bringing this to our attention.  unfortunately im out of the country this week so the earliest i can take a look is next week.  ive opened an issue to track this issue in #1242.

Issue #1234, opened 2018-07-16
Cleaning up SVI-related internals


eb8680:
This issue is for discussing instances of cruft or technical debt in SVI-related code and proposed fixes.  We can create new issues for individual instances if needed.

Potential issues/fixes:
- [ ] Collect all the duplicated `*ELBO._get_traces` logic into a single set of functions in `infer.enum`, as suggested in e.g. #1214.  Ideally we could also use the same code as part of #1226. (Issue moved from #1227 and work started in #1230)
- [ ] Remove deprecated `infer.enum.iter_discrete_traces`
- [ ] Remove `SVI.evaluate_loss` in favor of calling `loss(...)` directly
- [ ] Collect duplicated `Trace*_ELBO.loss` functionality into one place, e.g. `ELBO.loss`
- [ ] Remove `ELBO` base class entirely, split functionality into reusable functions that can be used outside SVI e.g. in `Importance`
- [ ] Standardize interfaces of SVI validation logic (e.g. `check_model_guide_match`, `check_site_shape`) so that it's more composable
- [ ] Simplify `ParamStore` and `PyroOptim` by removing `param_name()` and related internals mapping from parameters to names
- [ ] Make `ParamStore` more `dict`-like by replacing current interface with `__getitem__`/`__setitem__`/`update`, or even just make it a subclass of `dict`

fritzo:
To help guide this effort, I'd like to take a look at some long-term goals:
- #238 
- making `newton_step()` look more like `SVI` or `TracePosterior`
- making `MarginalAssignment` and friends look more like `SVI` or `TracePosterior`
- using `SVI.differentiable_loss` inside `newton_step()`

PR #1243, opened 2018-07-17
Implement TraceEnum_ELBO.differentiable_loss() method

fritzo:
Addresses #1234 

This exposes the DiCE operator as a `.differentiable_loss()` method for use by e.g. the `Newton` optimizer.

Note that this keeps around the old `.loss()` method for three reasons:
- minor performance: `.loss()` frees the compute graph after each particle, whereas `.differentiable_loss().item()` requires `num_particles`-times more memory.
- minor accuracy: `.loss()` accumulates loss terms as python floats (64 bit) rather than typical PyTorch floats (32-bit)
- backwards compatbility: we currently assume `.loss()` returns a float.

## Tested
- added `.differentiable_loss` assertions or variants to a few existing tests
- tested agreement between `differentiable_loss` and `loss_and_grads`

### DIFF
--- a/pyro/infer/traceenum_elbo.py
+++ b/pyro/infer/traceenum_elbo.py
@@ -131,6 +131,33 @@ class TraceEnum_ELBO(ELBO):
         warn_if_nan(loss, "loss")
         return loss
 
+    def differentiable_loss(self, model, guide, *args, **kwargs):
+        """
+        :returns: returns an estimate of the ELBO
+        :rtype: torch.Tensor
+        :raises ValueError: if the ELBO is not differentiable (e.g. is
+            identically zero)
+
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -52,6 +52,7 @@
 
 # examples/tutorials
 EXTRAS_REQUIRE = [
+    'ipywidgets==7.2.0',  # work around https://github.com/jupyter-widgets/ipywidgets/issues/2137
```
neerajprad: We can remove this fix now, it works fine locally.

fritzo:
Actually, I'll postpone the `loss_and_grads` refactoring. I looked into it and it requires more poutine tracing stuff than I have time to thing about atm. In particular, we would need to plumb through the set of params to optimize.

PR #1246, opened 2018-07-18
Fix scheduler bug

### LINKED ISSUE #1242: lr scheduler not optimizing params
`svi.step()` is not updating parameters with each optimizer.  see #1241

jpchen:
Fixes #1242 

- fixes bug
- adds clarifying usage docs
- adds tests that would have caught this


fritzo:
Thanks for fixing this @jpchen!

PR #1247, opened 2018-07-19
Add examples of composing inference algorithms (BP, EM, SVI, Adam, Newton)

fritzo:
This sketches some Pyronic attempts at composing inference algorithms. Discussion welcome.

This adds three example inference strategies for a single object mapping model:
1.  loopy belief propagation (via `MarginalAssignment`) inside an EM algorithm (using `TraceEnum_ELBO.differentiable_loss()` and `Newton.step()`),
2. loopy BP inside EM inside vanilla SVI (using `SVI`)
3. loopy BP inside SVI (via `TraceEnum_ELBO.differentiable_loss()` and `MixedMultOptimizer.step()`) where one variable is optimized via `Adam` and another via `Newton`

Note that this avoids annotating independence of params #1213, and relies on the user to take care in mixing optimizer methods.

### DIFF
--- a/pyro/optim/multi.py
+++ b/pyro/optim/multi.py
@@ -43,7 +43,8 @@ class MultiOptimizer(object):
         updated_values = self.get_step(loss, params)
         for name, value in params.items():
             with torch.no_grad():
-                value[...] = updated_values[name]
+                # we need to detach because updated_value may depend on value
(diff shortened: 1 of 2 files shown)


fritzo:
I think this is ready to merge. I'd like to pursue the other things we discussed, and those can be left to follow-up PRs.

PR #1249, opened 2018-07-20
Marginalize out discrete latent sites in HMC via enumeration

### LINKED ISSUE #1128: Automatically enumerate discrete variables in HMC
## Why?

Lots of models have discrete latent variables, e.g. mixture models. See e.g. https://forum.pyro.ai/t/nuts-with-discrete-latent-site-very-slow/204

## How?

Following `TraceEnum_ELBO` it should be possible to marginalize out any finite discrete latent variables:
1. wrap the model in `@config_enumerate`
2. use `EnumerateMessenger` to generate a bag of enumerated model trace (multiple traces come from sequential enumeration; broadcasted tensors come from parallel enumeration; both are reasonable)
3. block the enumerated sites from HMC and instead using their `site['log_prob'].exp()` to multiply downstream `site['log_prob']`.

neerajprad:
Resolves #1128 via parallel enumeration.

 - If a model has discrete latents, they will be enumerated over in parallel. In such cases, we need to supply `max_iarange_nesting` as arg to HMC/NUTS constructor to reserve dims for parallel enumeration. Also in these cases, we do not allow for stray batch dims, i.e. batch dimensions need to be wrapped inside an `iarange`.
 - Tested on gmm and a latent bernoulli model. Also, added some cheap tests in `mcmc/test_valid_models.py` to assert error/ok behavior.

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -12,6 +12,7 @@ import pyro.poutine as poutine
 from pyro.distributions.util import log_sum_exp
 from pyro.infer import config_enumerate, is_validation_enabled
 from pyro.infer.mcmc.trace_kernel import TraceKernel
+from pyro.infer.util import Dice
 from pyro.ops.dual_averaging import DualAveraging
 from pyro.ops.integrator import single_step_velocity_verlet, velocity_verlet
 from pyro.primitives import _Subsample
@@ -121,20 +122,25 @@ class HMC(TraceKernel):
             raise ValueError("Finite value required for `max_iarange_nesting` when model "
                              "has discrete (enumerable) sites.")
         model_trace.compute_log_prob()
+        model_trace.compute_score_parts()
         ordering = {name: frozenset(site["cond_indep_stack"])
                     for name, site in model_trace.nodes.items()
                     if site["type"] == "sample"}
 
         log_probs = defaultdict(float)
         for name, site in model_trace.nodes.items():
-            if site["type"] == "sample":
+            if site["type"] == "sample" and not site["fn"].has_enumerate_support:
                 log_probs[ordering[name]] = log_probs[ordering[name]] + site["log_prob"]
 
         log_prob_sum = 0.
-        for log_prob in log_probs.values():
+        dice = Dice(model_trace, ordering)
+        for ordinal, log_prob in log_probs.items():
             enum_dim = log_prob.dim() - self.max_iarange_nesting
+            dice_weights = dice.in_context(log_prob.shape, ordinal)
+            log_prob = log_prob + dice_weights.log()
             if enum_dim > 0:
-                log_prob = log_sum_exp(log_prob.reshape(-1, *log_prob.shape[enum_dim:]), dim=0)
+                log_prob = log_sum_exp(log_prob.reshape(-1, *log_prob.shape[enum_dim:]),
+                                       dim=0)
             log_prob_sum += log_prob.sum()
         return log_prob_sum
 
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import math
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 
 import torch
 from torch.distributions import biject_to, constraints
@@ -9,10 +9,13 @@ from torch.distributions import biject_to, constraints
 import pyro
 import pyro.distributions as dist
 import pyro.poutine as poutine
+from pyro.distributions.util import log_sum_exp
+from pyro.infer import config_enumerate, is_validation_enabled
 from pyro.infer.mcmc.trace_kernel import TraceKernel
 from pyro.ops.dual_averaging import DualAveraging
 from pyro.ops.integrator import single_step_velocity_verlet, velocity_verlet
-from pyro.util import torch_isinf, torch_isnan, optional
+from pyro.primitives import _Subsample
+from pyro.util import torch_isinf, torch_isnan, optional, check_site_shape
 
 
 class HMC(TraceKernel):
@@ -44,6 +47,9 @@ class HMC(TraceKernel):
         If not specified and the model has sites with constrained support,
         automatic transformations will be applied, as specified in
         :mod:`torch.distributions.constraint_registry`.
+    :param int max_iarange_nesting: Optional bound on max number of nested
+        :func:`pyro.iarange` contexts. This is required if model contains
+        discrete sample sites that can be enumerated over in parallel.
 
     Example:
 
@@ -65,10 +71,18 @@ class HMC(TraceKernel):
         tensor([ 0.9819,  1.9258,  2.9737])
     """
 
-    def __init__(self, model, step_size=None, trajectory_length=None,
-                 num_steps=None, adapt_step_size=False, transforms=None):
-        self.model = model
-
+    def __init__(self,
+                 model,
+                 step_size=None,
+                 trajectory_length=None,
+                 num_steps=None,
+                 adapt_step_size=False,
+                 transforms=None,
+                 max_iarange_nesting=float("inf")):
+        # Wrap model in `poutine.enum` to enumerate over discrete latent sites.
+        # No-op if model does not have any discrete latents.
+        self.model = poutine.enum(config_enumerate(model, default="parallel"),
+                                  first_available_dim=max_iarange_nesting)
         self.step_size = step_size if step_size is not None else 1  # from Stan
         if trajectory_length is not None:
             self.trajectory_length = trajectory_length
@@ -81,6 +95,7 @@ class HMC(TraceKernel):
         self._target_accept_prob = 0.8  # from Stan
 
         self.transforms = {} if transforms is None else transforms
+        self.max_iarange_nesting = max_iarange_nesting
         self._automatic_transform_enabled = True if transforms is None else False
         self._reset()
         super(HMC, self).__init__()
@@ -93,6 +108,35 @@ class HMC(TraceKernel):
         trace_poutine(*self._args, **self._kwargs)
         return trace_poutine.trace
 
+    @staticmethod
+    def _iter_latent_nodes(trace):
+        for name, node in sorted(trace.iter_stochastic_nodes(), key=lambda x: x[0]):
+            if not (node["fn"].has_enumerate_support or isinstance(node["fn"], _Subsample)):
+                yield (name, node)
+
+    def _compute_trace_log_prob(self, model_trace):
+        if not self._has_enumerable_sites:
+            return model_trace.log_prob_sum()
+        if self.max_iarange_nesting == float("inf"):
+            raise ValueError("Finite value required for `max_iarange_nesting` when model "
+                             "has discrete (enumerable) sites.")
+        model_trace.compute_log_prob()
+        ordering = {name: frozenset(f for f in site["cond_indep_stack"])
+                    for name, site in model_trace.nodes.items()
+                    if site["type"] == "sample"}
+
+        log_probs = defaultdict(float)
+        for name, site in model_trace.nodes.items():
+            if site["type"] == "sample":
+                log_probs[ordering[name]] = log_probs[ordering[name]] + site["log_prob"]
+
+        log_prob_sum = 0.
+        for log_prob in log_probs.values():
+            while log_prob.dim() > self.max_iarange_nesting:
+                log_prob = log_sum_exp(log_prob, dim=0)
+            log_prob_sum += log_prob.sum()
+        return log_prob_sum
+
     def _kinetic_energy(self, r):
         return 0.5 * sum(x.pow(2).sum() for x in r.values())
 
@@ -103,7 +147,7 @@ class HMC(TraceKernel):
         for name, transform in self.transforms.items():
             z_constrained[name] = transform.inv(z_constrained[name])
         trace = self._get_trace(z_constrained)
-        potential_energy = -trace.log_prob_sum()
+        potential_energy = -self._compute_trace_log_prob(trace)
         # adjust by the jacobian for this transformation.
         for name, transform in self.transforms.items():
             potential_energy += transform.log_abs_det_jacobian(z_constrained[name], z[name]).sum()
@@ -121,6 +165,7 @@ class HMC(TraceKernel):
         self._prototype_trace = None
         self._adapt_phase = False
         self._adapted_scheme = None
+        self._has_enumerable_sites = False
 
     def _find_reasonable_step_size(self, z):
         step_size = self.step_size
@@ -169,9 +214,14 @@ class HMC(TraceKernel):
         self.num_steps = max(1, int(self.trajectory_length / self.step_size))
 
     def _validate_trace(self, trace):
-        trace_log_prob_sum = trace.log_prob_sum()
+        trace_log_prob_sum = self._compute_trace_log_prob(trace)
         if torch_isnan(trace_log_prob_sum) or torch_isinf(trace_log_prob_sum):
             raise ValueError("Model specification incorrect - trace log pdf is NaN or Inf.")
+        # If model has enumerable sites, check for correct `iarange` placement.
+        if self._has_enumerable_sites and is_validation_enabled():
+            for site in trace.nodes.values():
+                if site["type"] == "sample":
+                    check_site_shape(site, self.max_iarange_nesting)
 
     def initial_trace(self):
         return self._prototype_trace
@@ -185,7 +235,12 @@ class HMC(TraceKernel):
         self._prototype_trace = trace
         if self._automatic_transform_enabled:
             self.transforms = {}
-        for name, node in sorted(trace.iter_stochastic_nodes(), key=lambda x: x[0]):
+        for name, node in trace.iter_stochastic_nodes():
+            if isinstance(node["fn"], _Subsample):
+                continue
+            if node["fn"].has_enumerate_support:
+                self._has_enumerable_sites = True
+                continue
             site_value = node["value"]
             if node["fn"].support is not constraints.real and self._automatic_transform_enabled:
                 self.transforms[name] = biject_to(node["fn"].support).inv
@@ -197,7 +252,7 @@ class HMC(TraceKernel):
 
         if self.adapt_step_size:
             self._adapt_phase = True
-            z = {name: node["value"] for name, node in trace.iter_stochastic_nodes()}
+            z = {name: node["value"] for name, node in self._iter_latent_nodes(trace)}
             for name, transform in self.transforms.items():
                 z[name] = transform(z[name])
             with pyro.validation_enabled(False):
@@ -218,7 +273,7 @@ class HMC(TraceKernel):
         self._reset()
 
     def sample(self, trace):
-        z = {name: node["value"].detach() for name, node in trace.iter_stochastic_nodes()}
+        z = {name: node["value"].detach() for name, node in self._iter_latent_nodes(trace)}
         # automatically transform `z` to unconstrained space, if needed.
         for name, transform in self.transforms.items():
             z[name] = transform(z[name])
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import math
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict
 
 import torch
 from torch.distributions import biject_to, constraints
@@ -9,10 +9,9 @@ from torch.distributions import biject_to, constraints
 import pyro
 import pyro.distributions as dist
 import pyro.poutine as poutine
-from pyro.distributions.util import log_sum_exp
 from pyro.infer import config_enumerate, is_validation_enabled
 from pyro.infer.mcmc.trace_kernel import TraceKernel
-from pyro.infer.util import Dice
+from pyro.infer.mcmc.util import EnumTraceProbEvaluator
 from pyro.ops.dual_averaging import DualAveraging
 from pyro.ops.integrator import single_step_velocity_verlet, velocity_verlet
 from pyro.primitives import _Subsample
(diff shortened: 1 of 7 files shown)


pyro/infer/mcmc/hmc.py
```diff
@@ -93,6 +108,35 @@ def _get_trace(self, z):
...
+                    if site["type"] == "sample"}
+
+        log_probs = defaultdict(float)
+        for name, site in model_trace.nodes.items():
+            if site["type"] == "sample":
+                log_probs[ordering[name]] = log_probs[ordering[name]] + site["log_prob"]
+
+        log_prob_sum = 0.
+        for log_prob in log_probs.values():
+            while log_prob.dim() > self.max_iarange_nesting:
```
fritzo: nit: it would.be cheaper to reshape, something like:
```py
enum_dim = log_prob.dim() - self.max_iarange_nesting
if enum_dim > 0:
    log_prob = log_sum_exp(log_prob.reshape(-1, *log_prob.shape[enum_dim:]), dim=0)
```

neerajprad: +1; will make the change.

fritzo:
Do I understand correctly that:
1. This PR implements parallel enumeration but not sequential enumeration.
2. The restriction to parallel enumeration is reasonable because HMC already makes the assumption that model structure is fixed, and the main use case of sequential enumeration is for models with dynamic structure.

pyro/infer/mcmc/hmc.py
```diff
@@ -93,6 +108,36 @@ def _get_trace(self, z):
...
+        log_probs = defaultdict(float)
+        for name, site in model_trace.nodes.items():
+            if site["type"] == "sample":
+                log_probs[ordering[name]] = log_probs[ordering[name]] + site["log_prob"]
+
+        log_prob_sum = 0.
+        for log_prob in log_probs.values():
+            enum_dim = log_prob.dim() - self.max_iarange_nesting
+            if enum_dim > 0:
+                log_prob = log_sum_exp(log_prob.reshape(-1, *log_prob.shape[enum_dim:]), dim=0)
```
fritzo: I believe you'll need to do something fancier like `Dice.in_context()` here. At the moment, you're summing over only the local probabilities rather than all probabilities. This will fail for models with nested enumeration, e.g.
```py
@poutine.broadcast
def model(data):
    p = pyro.sample("p", dist.Uniform(0., 1.))
    y = pyro.sample("y", dist.Bernoulli(p))
    q = 0.5 + 0.25 * y
    with pyro.iarange("data", len(data)):
        z = pyro.sample("z", dist.Bernoulli(q))
        mean = 2 * z - 1
        pyro.sample("obs", dist.Normal(mean, 1.), obs=data)
```
In this case I believe you'll currently compute
```py
p(p) = p_unif(p)
     + sum(p_bern(y | p) for y in [0,1])
     + prod(sum(p_bern(zi | q(y)) for zi in [0,1] for y in [0,1])
            for i in range(len(data)))
```
which is not equal to the intended
```py
p(p) = p_unif(p)
     + sum(p_bern(y | p) * prod(sum(p_bern(zi | q(y)) for zi in [0,1])
                                for i in range(len(data)))
           for y in [0,1])
```
Could we add something like this as a unit test against a vetted `log_prob` computation?

fritzo:
Maybe it's worth moving `_compute_dice_elbo()` from traceenum_elbo.py into the `Dice` class as a `Dice.compute_elbo()` method, and then implementing what you want as a `Dice.compute_log_prob()` method? It seems like their code bodies are quite similar, but maybe their assumptions are sufficiently different for them to more naturally lie closer to their callers?

neerajprad:
> EDIT I've attempted to do some of this in #1251, hopefully that helps a bit.

Thanks for the PR, @fritzo! Let me go through this in more detail, and reply to your comments.

neerajprad:
> Do I understand correctly that:

> This PR implements parallel enumeration but not sequential enumeration. The restriction to parallel enumeration is reasonable because HMC already makes the assumption that model structure is fixed, and the main use case of sequential enumeration is for models with dynamic structure.

That's right. Sequential enumeration will greatly add to the already considerable computational overhead (since I think we'll need to run the integrator on each trace separately), making it not as useful practically.

pyro/infer/mcmc/hmc.py
neerajprad: Thanks for this example, really helped me understand the issue with nested enumeration. I will add a test to validate the log prob computation.

eb8680:
> Sequential enumeration will greatly add to the already considerable computational overhead (since I think we'll need to run the integrator on each trace separately), making it not as useful practically.

This could still be very useful for models whose log-probability computation graph after enumeration is static, since those can still be accelerated by the PyTorch JIT.  I don't think we need to handle the sequential case in this PR now.

neerajprad:
Updated the PR with the following:
 - The log prob computation used `Dice.in_context` to get the weights from the enumerated sites.
 - Added a test that validates the log prob computation for nested enumeration (can be computed by hand). Also modified the latent bernoulli test to catch the case explained by @fritzo.

I am seeing that using `Dice` and computing `score_parts` has increased the runtime by 40-50%, so there might be some opportunity for optimizing specifically for HMC.

neerajprad:
I could not use `Dice.compute_expectation` because the cost (probs of continuous sites) factors do not add up between different ordinals (but rather need to multiplied together). Similarly the dims within `max_iarange_nesting` need to be multiplied. If we use `log_probs` instead, that leads to something which isn't as general, and is basically what we have in `HMC._compute_trace_log_prob` currently.

pyro/infer/mcmc/hmc.py
```diff
@@ -93,6 +109,41 @@ def _get_trace(self, z):
...
+                             "has discrete (enumerable) sites.")
+        model_trace.compute_log_prob()
+        model_trace.compute_score_parts()
+        ordering = {name: frozenset(site["cond_indep_stack"])
+                    for name, site in model_trace.nodes.items()
+                    if site["type"] == "sample"}
+
+        log_probs = defaultdict(float)
+        for name, site in model_trace.nodes.items():
+            if site["type"] == "sample" and not site["fn"].has_enumerate_support:
```
fritzo: I believe this incorrectly ignores discrete observe sites?

neerajprad: Thanks for catching, will change and a test for this.

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,105 @@
+from collections import defaultdict
+
+from pyro.distributions.util import log_sum_exp
+from pyro.infer.util import is_validation_enabled
+from pyro.util import check_site_shape
+
+
+class EnumTraceProbEvaluator(object):
+    def __init__(self,
```
neerajprad: @frtizo - As discussed, I have removed the dependency on Dice, and moved the log prob computation to its own class to make it easy to test (test cases are added in `infer/mcmc/test_valid_models.py`). I think this should be self-explanatory (and very similar to the Dice logic), but let me know in case I may have overlooked some corner cases.

neerajprad: Also, this should resolve the bugs in the previous implementation. I realized that the logic of only looking at the enumerated dims when computing the upstream log probs was wrong.

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,105 @@
...
+                 model_trace,
+                 has_enumerable_sites=False,
+                 max_iarange_nesting=float("inf")):
+        """
```
fritzo: nit: could you move this comment to a class-level comment rather than an `__init__` comment?

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,105 @@
...
+        :param bool has_enumerable_sites: whether the trace contains any
+            discrete enumerable sites.
+        :param int max_iarange_nesting: Optional bound on max number of nested
+            :func:`pyro.iarange` contexts.
+        """
+        self.model_trace = model_trace
+        self.has_enumerable_sites = has_enumerable_sites
+        self.max_iarange_nesting = max_iarange_nesting
+        self.log_probs = defaultdict(list)
+        self._log_factors_cache = {}
```
fritzo: nit: It's cheaper and probably simpler to use a cache of pairs rather than a pair of caches.

tests/infer/mcmc/test_valid_models.py
```diff
@@ -0,0 +1,253 @@
...
+    @poutine.broadcast
+    def model(data):
+        p = pyro.sample("p", dist.Uniform(0., 1.))
+        y = pyro.sample("y", dist.Bernoulli(p))
+        mean = 2 * y - 1
+        n = pyro.sample("n", dist.Normal(mean, 1.))
+        with pyro.iarange("data", len(data)):
+            pyro.sample("obs", dist.Bernoulli(sigmoid(n)), obs=data)
+
+    conditioned_model = poutine.enum(
```
fritzo: nit: why not add these to decorators above?
```py
@poutine.enum(first_available_dim=1)
@config_enumerate(default="parallel")
@poutine.condition(data={"p": ..., "n": ...})
@poutine.broadcast
def model(data):
    ...
```

pyro/infer/mcmc/util.py
neerajprad: My bad, will change.

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,105 @@
...
+            return self.model_trace.log_prob_sum()
+        if self.max_iarange_nesting == float("inf"):
+            raise ValueError("Finite value required for `max_iarange_nesting` when model "
+                             "has discrete (enumerable) sites.")
+        self._compute_log_prob_terms()
+
+        # Sum up terms from predecessor, and gather leaf nodes.
+        leaves_log_probs = {}
+        for target_ordinal in sorted(self.log_probs.keys()):
+            leaves_log_probs[target_ordinal] = self.log_probs[target_ordinal]
```
fritzo: nit: this is a bit confusing. Consider simplifying to
```py
for target_ordinal, log_prob in self.log_probs.items():
    if any(target_ordinal < other for other in self.log_probs):
        continue  # not a leaf
    predecessors, log_factors = self._get_predecessors_log_factors(target_ordinal)
    leaves_log_probs[target_ordinal] = sum(log_prob + log_factors)
```

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def _get_predecessors_log_factors(self, target_ordinal):
+        """
+        Returns the list of predecessors for `target_ordinal `and
+        their log_prob factors.
+        """
+        if target_ordinal in self._predecessors_cache:
+            return self._predecessors_cache[target_ordinal]
+        log_factors = []
+        predecessors = set()
```
fritzo: Am I correct that `predecessors` is no longer used, so we can remove this data structure?

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,101 @@
...
+            predecessors, log_factors = self._get_predecessors_log_factors(target_ordinal)
+            leaves_log_probs[target_ordinal] = sum(leaves_log_probs[target_ordinal] + log_factors)
+
+        # Reduce the log prob terms for each leaf node:
+        # - taking log_sum_exp of factors in enum dims (i.e.
+        # adding up the probability terms).
+        # - summing up the dims within `max_iarange_nesting`.
+        # (i.e. multiplying probs within independent batches).
+        log_prob_sum = 0.
+        for ordinal in leaves_log_probs:
```
fritzo: nit: simplfy using `.items()`
```py
for ordinal, log_prob in leaves_log_probs.items():
    ...
```

pyro/infer/mcmc/util.py
neerajprad: Fixed.

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,101 @@
...
+        if not self.has_enumerable_sites:
+            return self.model_trace.log_prob_sum()
+        if self.max_iarange_nesting == float("inf"):
+            raise ValueError("Finite value required for `max_iarange_nesting` when model "
+                             "has discrete (enumerable) sites.")
+        self._compute_log_prob_terms()
+
+        # Sum up terms from predecessor, and gather leaf nodes.
+        leaves_log_probs = {}
+        for target_ordinal in sorted(self.log_probs.keys()):
```
fritzo: nit: simplify using `.items()`
```py
for target_ordinal, log_prob in self.log_probs.items():
    if ...
        continue:
    log_factors = self._get_log_factors(target_ordinal)
    leaves_log_probs[target_ordinal] = log_prob + sum(log_factors)
```

fritzo:
> we just need to multiply the terms per independence context

I guess you could aggregate up the tree, starting with leaves and moving outward/upward. As you leave each independence context you `prod(sum(-))`? Another approach that I've taken in sequential enumeration is to explicitly scale for multiple counting by accumulating a `denominators` dict keyed on `ordinal`. Whatever works :smile:

neerajprad:
> I guess you could aggregate up the tree, starting with leaves and moving outward/upward. As you leave each independence context you prod(sum(-))?

This is the approach I have taken now. It cleanly solves a bunch of issues we were getting earlier, but this does not work for general DAGs (i.e., indep contexts with multiple parents).

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,137 @@
+import torch
+from collections import defaultdict
+
+from pyro.distributions.util import log_sum_exp
+from pyro.infer.util import is_validation_enabled
+from pyro.poutine.indep_messenger import CondIndepStackFrame
+from pyro.util import check_site_shape
+
+
+class EnumTraceProbEvaluator(object):
```
neerajprad: @fritzo - This is ready for another round of review. 😄The important change is confined to this class. In HMC we are calling the constructor once to populate the children nodes and compute which dims should be evaluated/enumerated out. When we call `log_prob` subsequently, we reuse this computed metadata to aggregate the log prob terms from the leaf nodes upwards.

pyro/infer/mcmc/hmc.py
```diff
@@ -93,6 +108,15 @@ def _get_trace(self, z):
         trace_poutine(*self._args, **self._kwargs)
         return trace_poutine.trace
 
+    @staticmethod
+    def _iter_latent_nodes(trace):
+        for name, node in sorted(trace.iter_stochastic_nodes(), key=lambda x: x[0]):
```
fritzo: Why sort by name rather than the default order (by execution time)?

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,137 @@
...
+    :param int max_iarange_nesting: Optional bound on max number of nested
+        :func:`pyro.iarange` contexts.
+    """
+    def __init__(self,
+                 model_trace,
+                 has_enumerable_sites=False,
+                 max_iarange_nesting=float("inf")):
+        self.has_enumerable_sites = has_enumerable_sites
+        self.max_iarange_nesting = max_iarange_nesting
+        default_cond_stack = (CondIndepStackFrame(name="default", dim=0, size=0, counter=None),)
```
fritzo: Why not the empty tuple?
```py
self.root = frozenset((()))
```

pyro/infer/mcmc/hmc.py
neerajprad: Yeah, any fixed order is fine. The advantage to using names is that the dict based data structures that we are using can be sorted anytime by key names (otherwise, we'll need to store the key order somewhere and then sort using that order).

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,137 @@
...
+        self.root = frozenset(default_cond_stack)
```
fritzo: Why not the empty `frozenset()`?

neerajprad: I think this can be removed now. I was initially doing something like `if frozenset() in frozenset([1])` (which evaluates to False), so this was to not have to put in special cases to deal with the root node. Will remove this now.

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,137 @@
...
+    def _compute_log_prob_terms(self, model_trace):
+        """
+        Computes the conditional probabilities for each of the sites
+        in the model trace, and stores the result in `self._log_probs`.
+        """
+        model_trace.compute_log_prob()
+        ordering = {}
+        for name, site in model_trace.nodes.items():
+            if site["type"] == "sample":
+                if len(site["cond_indep_stack"]) == 0:
```
fritzo: If you make the `root` an empty frozenset then these four lines simplify to
```py
ordering[name] = frozenset(site["cond_indep_stack"])
```

neerajprad: Changed, having a special default key isn't needed anymore.

fritzo: or simply
```py
ordering = {name: frozenset(site["cond_indep_stack"])
            for name, site in model_trace.nodes.items()
            if site["type"] == "sample"}
```

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,137 @@
...
+
+        # Collect log prob terms per independence context.
+        log_probs = defaultdict(list)
+        for name, site in model_trace.nodes.items():
+            if site["type"] == "sample":
+                if is_validation_enabled():
+                    check_site_shape(site, self.max_iarange_nesting)
+                log_probs[ordering[name]].append(site["log_prob"])
+
+        for ordinal in log_probs:
```
fritzo: nit: If you use `.items()` you can avoid half of the hashing of frozensets.
```py
for ordinal, terms in log_probs.items():
    self._log_probs[ordinal] = sum(terms)
```
Also, cool trick to avoid `0.0 + x` tensor ops!

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,137 @@
+import torch
+from collections import defaultdict
+
+from pyro.distributions.util import log_sum_exp
```
fritzo: Could we try to use the `logsumexp` form internally, since that's what PyTorch 0.4.1 calls it?

neerajprad: Oh..sorry, I forgot to update. Will do so.

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,137 @@
...
+        for name, site in model_trace.nodes.items():
+            if site["type"] == "sample":
+                if is_validation_enabled():
+                    check_site_shape(site, self.max_iarange_nesting)
+                log_probs[ordering[name]].append(site["log_prob"])
+
+        for ordinal in log_probs:
+            self._log_probs[ordinal] = sum(log_probs[ordinal])
+
+    def _reduce(self, ordinal, agg_log_prob=torch.tensor(0.)):
```
fritzo: It's dangerous to create a `torch.tensor` at library load time, since it will not pick up `torch.set_default_tensor_type()` invocations. Instead I'd default to `None` and to an `if agg_log_prob is not None:` check.

neerajprad: Good catch! Will change.

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,137 @@
...
+            log_prob = log_sum_exp(log_prob, dim=enum_dim, keepdim=True)
+        for marginal_dim in self._iarange_dims[ordinal]:
+            log_prob = log_prob.sum(dim=marginal_dim, keepdim=True)
+        return log_prob
+
+    def _aggregate_log_probs(self, ordinal):
+        """
+        Aggregate the `log_prob` terms using depth first search.
+        """
+        if self._children[ordinal] is None:
```
fritzo: Hmm `self._children` is a `defaultdict(list)`. Do you instead mean to
```py
if not self._children[ordinal]:
    return self._reduce(ordinal)
```

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,137 @@
...
+            return self._reduce(ordinal)
+        agg_log_prob = torch.tensor(0.)
```
fritzo: You could avoid `torch.tensor` and save a tensor op by instead using python's `sum`:
```py
agg_log_prob = sum(map(self._aggregate_log_probs, self._children[ordinal]))
```

pyro/infer/mcmc/util.py
neerajprad: I think it will be safe here to just use a python scalar. If a trace has no enumerable sites, we won't be going through this code path anyways, and if it has this 0. will get added as expected on the correct device.

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        Returns the log pdf of `model_trace` by appropriately handling
+        enumerated log prob factors.
+
+        :return: log pdf of the trace.
+        """
+        if not self.has_enumerable_sites:
+            return model_trace.log_prob_sum()
+        self._compute_log_prob_terms(model_trace)
+        return self._aggregate_log_probs(self.root).sum()
```
fritzo: Do I understand correctly that:
1. all enumeration dims are to the left of `-max_iarange_nesting`, and they will have already been summed out by `self._aggregate_log_probs`
2. the result of `self._aggregate_log_probs(self.root)` will have non-reduced dims in exactly the global iaranges, i.e. the iaranges inside of which all sample statements are called, e.g. parallel particle iaranges?

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        For each ordinal, populate the `iarange` and `enum` dims to be
+        evaluated or enumerated out.
+        """
+        log_prob = self._log_probs[ordinal]
+        iarange_dims = sorted([frame.dim for frame in ordinal - parent_ordinal])
+        enum_dims = set((i for i in range(-log_prob.dim(), -self.max_iarange_nesting)
+                         if log_prob.shape[i] > 1))
+        self._iarange_dims[ordinal] = iarange_dims
+        self._enum_dims[ordinal] = sorted(list(enum_dims - parent_enum_dims))
```
fritzo: nit: no need for `list( )`

pyro/infer/mcmc/util.py
```diff
@@ -0,0 +1,137 @@
...
+    def _parse_model_structure(self, model_trace):
+        if not self.has_enumerable_sites:
+            return
+        if self.max_iarange_nesting == float("inf"):
+            raise ValueError("Finite value required for `max_iarange_nesting` when model "
+                             "has discrete (enumerable) sites.")
+        self._compute_log_prob_terms(model_trace)
+        # 1. Infer model structure - compute parent-child relationship.
+        sorted_ordinals = sorted(self._log_probs.keys())
+        for i in range(len(sorted_ordinals)):
```
fritzo: nit: consider simplifying to
```py
for i, child_node in enumerate(sorted_ordinals):
```

neerajprad:
I think your comment got collapsed:
> Do I understand correctly that: all enumeration dims are to the left of -max_iarange_nesting, and they will have already been summed out by self._aggregate_log_probs

That's correct! 

> the result of self._aggregate_log_probs(self.root) will have non-reduced dims in exactly the global iaranges, i.e. the iaranges inside of which all sample statements are called, e.g. parallel particle iaranges?

They should be reduced dims, except that they will all be singleton, i.e. like `[[[1.3161]]]`. We don't currently have any global iaranges since we are not running parallel chains.

pyro/infer/mcmc/util.py
neerajprad: done.

pyro/infer/mcmc/util.py
neerajprad: done.

neerajprad:
> Can you confirm that you have tests that would have caught the double-counting errors that plagued earlier versions of this PR?

Yes, I have added many tests that would catch such errors in `test_valid_models`. There were a couple of tests that would have caught this even earlier, had I stand guilty of having double counted myself while writing the expected values. :) 

> After this merges, I'd like to add some more interesting models as tests, e.g. HMM and GMM.

There is a small gmm test in `test_hmc.py`, but it'll be great to have HMM and other interesting models.

neerajprad:
I think there is a dim error in the `logsumexp` implementation, which I'm looking into. Will update shortly.

neerajprad:
Thanks a ton for the detailed review, @fritzo. I think I have addressed all comments, and fixed the `logsumexp` bug now.

fritzo:
Nice work, @neerajprad!

martinjankowiak:
yaaay!
this is awesome @neerajprad !
also, seems like this PR also benefited from some serious CR from @fritzo !

eb8680:
Awesome!

PR #1250, opened 2018-07-20
Add Bayesian Optimisation to AB testing example

ae-foster:
To complete our first OED example, I've added a GP-based Bayesian optimizer based on http://pyro.ai/examples/bo.html . This uses Thompson sampling to acquire new data in batches.

Summary of changes:
 - `ab_test.py` - optimizing the APE using `GPBayesOptimizer`
 - `gp_bayes_opt.py` - new code implementing the BO example with Thompson sampling: probably need to move this somewhere else or can we merge it with existing BO code?
 - `gpr.py` - added a method to lazily create a multivariate sample from a Gaussian process posterior

### DIFF
--- a/examples/contrib/oed/gp_bayes_opt.py
+++ b/examples/contrib/oed/gp_bayes_opt.py
@@ -38,7 +38,9 @@ class GPBayesOptimizer:
         """
         # transform x to an unconstrained domain
         unconstrained_x_init = transform_to(self.constraints).inv(x_init)
-        unconstrained_x = torch.tensor(unconstrained_x_init, requires_grad=True)
+        unconstrained_x = unconstrained_x_init.new_tensor(
+            unconstrained_x_init, requires_grad=True)
+        # TODO: Use LBFGS with line search by pytorch #8824 merged
         minimizer = optim.LBFGS([unconstrained_x], max_eval=20)
 
         def closure():
@@ -70,16 +72,14 @@ class GPBayesOptimizer:
         :rtype: tuple
         """
 
-        x_init = self.gpmodel.X[-1:].new_empty(1).uniform_(self.constraints.lower_bound,
-                                                           self.constraints.upper_bound)
         candidates = []
         values = []
         for j in range(num_candidates):
+            x_init = self.gpmodel.X.new_empty(1).uniform_(
+                self.constraints.lower_bound, self.constraints.upper_bound)
             x, y = self.find_a_candidate(differentiable, x_init)
             candidates.append(x)
             values.append(y)
-            x_init = x.new_empty(1).uniform_(self.constraints.lower_bound,
-                                             self.constraints.upper_bound)
 
         mvalue, argmin = torch.min(torch.cat(values), dim=0)
         return candidates[argmin.item()], mvalue
@@ -96,7 +96,7 @@ class GPBayesOptimizer:
         """
 
         # Initialize the return tensor
-        X = torch.zeros(num_acquisitions, *self.gpmodel.X.shape[1:])
+        X = self.gpmodel.X.new_empty(num_acquisitions, *self.gpmodel.X.shape[1:])
 
         for i in range(num_acquisitions):
             sampler = self.gpmodel.iter_sample(noiseless=False)
@@ -105,7 +105,7 @@ class GPBayesOptimizer:
 
         return X
 
-    def run(self, num_steps, num_acquisitions, acquisition_func=self.acquire_thompson):
+    def run(self, num_steps, num_acquisitions, acquisition_func=None):
         """
         Optimizes `self.f` in `num_steps` steps, acquiring `num_acquisitions`
         new function evaluations at each step.
@@ -117,6 +117,8 @@ class GPBayesOptimizer:
         :return: the minimiser and the minimum value
         :rtype: tuple
         """
+        if acquisition_func is None:
+            acquisition_func = self.acquire_thompson
 
         for i in range(num_steps):
             X = acquisition_func(num_acquisitions=num_acquisitions)
--- a/examples/contrib/oed/gp_bayes_opt.py
+++ b/examples/contrib/oed/gp_bayes_opt.py
@@ -84,12 +84,10 @@ class GPBayesOptimizer:
         mvalue, argmin = torch.min(torch.cat(values), dim=0)
         return candidates[argmin.item()], mvalue
 
-    def acquire(self, method="Thompson", num_acquisitions=1, **opt_params):
+    def acquire_thompson(self, num_acquisitions=1, **opt_params):
         """Selects `num_acquisitions` query points at which to query the
-        original function.
+        original function by Thompson sampling.
 
-        :param str method: the method to use for acquisition. Choose from:
-            Thompson
         :param int num_acquisitions: the number of points to generate
         :param dict opt_params: additional parameters for optimization
             routines
@@ -97,34 +95,31 @@ class GPBayesOptimizer:
         :rtype: torch.Tensor
         """
 
-        if method == "Thompson":
+        # Initialize the return tensor
+        X = torch.zeros(num_acquisitions, *self.gpmodel.X.shape[1:])
 
-            # Initialize the return tensor, add the UCB1 point
-            X = torch.zeros(num_acquisitions, *self.gpmodel.X.shape[1:])
+        for i in range(num_acquisitions):
+            sampler = self.gpmodel.iter_sample(noiseless=False)
+            x, _ = self.opt_differentiable(sampler, **opt_params)
+            X[i, ...] = x
 
-            for i in range(num_acquisitions):
-                sampler = self.gpmodel.iter_sample(noiseless=False)
-                x, _ = self.opt_differentiable(sampler, **opt_params)
-                X[i, ...] = x
+        return X
 
-            return X
-
-        else:
-            raise NotImplementedError("Only method Thompson implemented for acquisition")
-
-    def run(self, num_steps, num_acquisitions):
+    def run(self, num_steps, num_acquisitions, acquisition_func=self.acquire_thompson):
         """
         Optimizes `self.f` in `num_steps` steps, acquiring `num_acquisitions`
         new function evaluations at each step.
 
-        :param int num_steps:
-        :param int num_steps"
+        :param int num_steps: overall number of BO steps to run
+        :param int num_acquisitions: number of points to acquire at each step
+        :param function acquisition_func: a function to generate acquisitions.
+            It should return a torch.Tensor of new points to query.
         :return: the minimiser and the minimum value
         :rtype: tuple
         """
 
         for i in range(num_steps):
-            X = self.acquire(num_acquisitions=num_acquisitions)
+            X = acquisition_func(num_acquisitions=num_acquisitions)
             y = self.f(X)
             self.update_posterior(X, y)
 
--- a/examples/contrib/oed/gp_bayes_opt.py
+++ b/examples/contrib/oed/gp_bayes_opt.py
@@ -3,23 +3,30 @@ import torch.autograd as autograd
 import torch.optim as optim
 from torch.distributions import transform_to
 
+import pyro
 
-class GPBayesOptimizer:
+
+class GPBayesOptimizer(pyro.optim.multi.MultiOptimizer):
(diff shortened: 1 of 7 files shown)


fritzo:
@ae-foster to get CI tests to run, you'll need to merge dev into your branch and then push.

examples/contrib/oed/gp_bayes_opt.py
```diff
@@ -0,0 +1,164 @@
...
+
+        :param function differentiable: a function amenable to torch
+            autograd
+        :param torch.Tensor x_init: the initial point
+
+        """
+        # transform x to an unconstrained domain
+        unconstrained_x_init = transform_to(self.constraints).inv(x_init)
+        unconstrained_x = torch.tensor(unconstrained_x_init, requires_grad=True)
+        minimizer = optim.LBFGS([unconstrained_x], max_eval=20)
```
ae-foster: Cap evals at `20` to avoid some numerical instability

examples/contrib/oed/gp_bayes_opt.py
```diff
@@ -0,0 +1,164 @@
...
+
+        def closure():
+            minimizer.zero_grad()
+            if (torch.log(torch.abs(unconstrained_x)) > 25.).any():
```
ae-foster: Somewhat arbitrary cutoff- necessary to prevent overflow during LBFGS run

examples/contrib/oed/gp_bayes_opt.py
```diff
@@ -0,0 +1,164 @@
...
+    #     with torch.no_grad():
+    #         loc, var = self.gpmodel(Xnew, full_cov=False, noiseless=False)
+    #         sd = var.sqrt()
+    #         ax1.plot(Xnew.numpy(), loc.numpy(), "r", lw=2)  # plot predictive mean
+    #         ax1.fill_between(Xnew.numpy(), loc.numpy() - 2*sd.numpy(), loc.numpy() + 2*sd.numpy(),
+    #                          color="C0", alpha=0.3)  # plot uncertainty intervals
+    #     ax1.set_xlim(-1, 101)
+    #     ax1.set_title("Find {}".format(xlabel))
+    #     if with_title:
+    #         ax1.set_ylabel("Gaussian Process Regression")
```
ae-foster: Commented code is unsightly but useful for debugging. `test_examples` will not admit `matplotlib` in example code.

pyro/contrib/gp/models/gpr.py
```diff
@@ -141,3 +142,80 @@ def forward(self, Xnew, full_cov=False, noiseless=True):
...
+            Kffnew = Kff.new_empty(N+1, N+1)
+            Kffnew[:N, :N] = Kff
+            cross = self.kernel(X, xnew).squeeze()
+            end = self.kernel(xnew, xnew).squeeze()
+            Kffnew[N, :N] = cross
+            Kffnew[:N, N] = cross
+            # No noise, just jitter for numerical stability
+            Kffnew[N, N] = end + self.jitter
+            # Heuristic to avoid adding degenerate points
+            if Kffnew.logdet() > -15.:
```
ae-foster: Arbitrary cutoff: prevent `Kff` becoming singular.

pyro/contrib/gp/models/gpr.py
```diff
@@ -141,3 +142,80 @@ def forward(self, Xnew, full_cov=False, noiseless=True):
...
+            # Compute conditional mean and variance
+            loc, cov = conditional(xnew, X, self.kernel, y_residual, None, Lff,
+                                   False, jitter=self.jitter)
+            if not noiseless:
+                cov = cov + noise
+
+            # Todo use pyro.sample
+            d = normal.Normal(torch.tensor(0.), torch.tensor(1.))
+            # Reparametrize explicitly - aids autograd
+            ynew = (loc + self.mean_function(xnew)) + d.sample()*cov.sqrt()
```
ae-foster: It would be nice to make this a true pyronic sampler if possible

examples/contrib/oed/gp_bayes_opt.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        self.f = f
+        self.constraints = constraints
+        self.gpmodel = gpmodel
+
+    def update_posterior(self, X, y):
+        X = torch.cat([self.gpmodel.X, X])
+        y = torch.cat([self.gpmodel.y, y])
+        self.gpmodel.set_data(X, y)
+        self.gpmodel.optimize()
```
ae-foster: Not very pyronic

examples/contrib/oed/gp_bayes_opt.py
fritzo: Feel free to use local imports for plotting dependencies:
```py
def plot(self, gs, ...):
    from matplotlib import pyplot as plt
    ...
```
That's not perfect, but it's better than commented-out code :wink:

pyro/contrib/gp/models/gpr.py
eb8680: Why can't you write `torch.distributions.Normal(loc + self.mean_function(xnew), cov.sqrt()).rsample()`?

pyro/infer/trace_elbo.py
```diff
@@ -115,7 +115,7 @@ def loss_and_grads(self, model, guide, *args, **kwargs):
 
             if trainable_params and getattr(surrogate_elbo_particle, 'requires_grad', False):
                 surrogate_loss_particle = -surrogate_elbo_particle / self.num_particles
-                surrogate_loss_particle.backward()
+                surrogate_loss_particle.backward(retain_graph=True)
```
eb8680: This change has some performance and memory implications.  I'll push a fix today as part of #1227 that exposes a `differentiable_loss` function that you can use when you need `retain_graph=True`

pyro/contrib/gp/models/gpr.py
```diff
@@ -141,3 +142,80 @@ def forward(self, Xnew, full_cov=False, noiseless=True):
...
+            kernel's parameters have been learned from a training procedure (MCMC or
+            SVI).
+
+        :param bool noiseless: A flag to decide if we want to add sampling noise
+            to the samples beyond the noise inherent in the GP posterior.
+        :returns: sampler
+        :rtype: function
+        """
+        # Make these visible in the inner function
+        global X, y, Kff, N
```
eb8680: Instead of using `global` here and below, how about making `sample_next` take these variables as arguments and returning a curried function with `functools.partial`?

```python
def sample_next(X, y, Kff, N, xnew):
    ...

return functools.partial(sample_next, X, y, Kff, xnew)
```

examples/contrib/oed/gp_bayes_opt.py
eb8680: Can we do some GP magic to make these updates cheaper than recomputing the full posterior?

examples/contrib/oed/gp_bayes_opt.py
```diff
@@ -0,0 +1,164 @@
...
+        :param str method: the method to use for acquisition. Choose from:
+            Thompson
+        :param int num_acquisitions: the number of points to generate
+        :param dict opt_params: additional parameters for optimization
+            routines
+        :return: a tensor of points to evaluate `self.f` at
+        :rtype: torch.Tensor
+        """
+
+        if method == "Thompson":
```
eb8680: Let's split this out into an acquisition function parameter instead of implementing acquisition functions inside `acquire`, so it's easier to experiment with different Monte Carlo acquisition functions.

examples/contrib/oed/gp_bayes_opt.py
```diff
@@ -0,0 +1,164 @@
+# import matplotlib.gridspec as gridspec
+# import matplotlib.pyplot as plt
+import torch
+import torch.autograd as autograd
+import torch.optim as optim
+from torch.distributions import transform_to
+
+
+class GPBayesOptimizer:
```
eb8680: We should be able to refactor this a bit into a `pyro.optim.multi.MultiOptimizer` so that it looks more like other Pyro optimizers.  That way when we experiment with other optimizers e.g. gradient-based optimizers it'll be easy to swap them out

examples/contrib/oed/gp_bayes_opt.py
eb8680: You should probably just add a separate test for this and remove the test code here

examples/contrib/oed/gp_bayes_opt.py
```diff
@@ -0,0 +1,164 @@
...
+    """Performs Bayesian Optimization using a Gaussian Process as an
+    emulator for the unknown function.
```
eb8680: Once the code is a little more settled down, you should expand this docstring and add a couple usage examples

examples/contrib/oed/gp_bayes_opt.py
```diff
@@ -0,0 +1,164 @@
+# import matplotlib.gridspec as gridspec
+# import matplotlib.pyplot as plt
```
eb8680: Remove these as discussed below

examples/contrib/oed/gp_bayes_opt.py
ae-foster: I'll remove the code, keep it on a local branch

pyro/contrib/gp/models/gpr.py
ae-foster: Yes, had it in my mind that that didn't work- it does

examples/contrib/oed/gp_bayes_opt.py
ae-foster: Yes depends how much magic you want. Here https://github.com/uber/pyro/blob/dev/pyro/contrib/gp/models/gpr.py#L80 we could have cached `Kff` and just add the kernel computations we need. One step further is to use the magic of Schur complements https://en.wikipedia.org/wiki/Schur_complement to avoid having to invert the whole kernel matrix. I actually started implementing a Schur complement method for `iter_sample` but in the end the existing solution works and reuses more existing code. In either case, we would have to refactor internals of `GPRegression`

pyro/infer/trace_elbo.py
ae-foster: Kk I'll rebase from `dev` once that is merged

pyro/contrib/gp/models/gpr.py
ae-foster: The reason I didn't do that was that we change `X` inside the inner function (by writing `X = Xnew`). I thought scoping rules would mean that that change wasn't saved, but honestly I didn't actually try it

examples/contrib/oed/gp_bayes_opt.py
eb8680: Let's open a separate issue about this, it seems like generally useful functionality for `contrib.gp`

pyro/infer/trace_elbo.py
eb8680: Added in #1252

examples/contrib/oed/gp_bayes_opt.py
ae-foster: Sure :) I didn't realize that was in the pipeline

examples/contrib/oed/gp_bayes_opt.py
ae-foster: I think we're best to talk over this together- we need to implement a `step` function which would replace my `run` function. Would be also need to refactor the example to use `pyro.param`? Right now, it's not really true that after a `step`, the params are updated to new near-optimal values. But obviously I could add `self.opt_differentiable(lambda x: self.gpmodel(x)[0])` into each step (minimize the current GP mean function)

examples/contrib/oed/gp_bayes_opt.py
ae-foster: https://github.com/uber/pyro/issues/1258

pyro/infer/trace_elbo.py
eb8680: Let's remove this change and switch to using `TraceEnum_ELBO` below

pyro/contrib/gp/models/gpr.py
eb8680: You can also put these variables into a dictionary and mutate that instead of using `global`

ae-foster:
@eb8680 have the CI passing now :)

PR #1252, opened 2018-07-20
Add differentiable_loss to Trace_ELBO

eb8680:
This PR adds `differentiable_loss` to `infer.Trace_ELBO` as proposed in e.g. #1227 and done for `TraceEnum_ELBO` in #1243.  It is necessary for #1250

### DIFF
--- a/pyro/infer/trace_elbo.py
+++ b/pyro/infer/trace_elbo.py
@@ -103,12 +103,14 @@ class Trace_ELBO(ELBO):
         Computes the surrogate loss that can be differentiated with autograd
         to produce gradient estimates for the model and guide parameters
         """
+        loss = 0.
         surrogate_loss = 0.
         for model_trace, guide_trace in self._get_traces(model, guide, *args, **kwargs):
             loss_particle, surrogate_loss_particle = self._differentiable_loss_particle(model_trace, guide_trace)
             surrogate_loss += surrogate_loss_particle / self.num_particles
+            loss += loss_particle / self.num_particles
         warn_if_nan(surrogate_loss, "loss")
-        return surrogate_loss
+        return loss + (surrogate_loss - surrogate_loss.detach())
 
     def loss_and_grads(self, model, guide, *args, **kwargs):
         """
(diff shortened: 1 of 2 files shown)


pyro/infer/trace_elbo.py
```diff
@@ -70,6 +70,46 @@ def loss(self, model, guide, *args, **kwargs):
...
+        """
+        Computes the surrogate loss that can be differentiated with autograd
+        to produce gradient estimates for the model and guide parameters
+        """
+        surrogate_loss = 0.
+        for model_trace, guide_trace in self._get_traces(model, guide, *args, **kwargs):
+            loss_particle, surrogate_loss_particle = self._differentiable_loss_particle(model_trace, guide_trace)
+            surrogate_loss += surrogate_loss_particle / self.num_particles
+        warn_if_nan(surrogate_loss, "loss")
+        return surrogate_loss
```
fritzo: What do you think about wrapping the `(loss, surrogate_loss)` pair in a `torch.autograd.Function` with a `@once_differentiable` `.backward()` method. That would be a little safer for use by multi optimizers, since the 0th and 1st derivatives would be correct and the 2nd derivative would raise an error.

fritzo: Or maybe we should define
```py
def differentiable_loss(...):
    loss, surrogate_loss = ...
    return loss + (surrogate_loss - surrogate_loss.detach())
```

ae-foster:
LGTM @eb8680 - should we add a test comparing `Trace` and `DiffTrace` where one raises a pytorch error and the other doesn't?

pyro/infer/trace_elbo.py
eb8680: I'd forgotten about this since in OED we haven't had any score-function terms yet.  We should probably just allow the DICE estimator in `TraceEnum_ELBO` to be used even when we're not enumerating - I think it currently throws an error in that case.

fritzo: I think you can set `strict_enumeration_warning=False` to use `TraceEnum_ELBO` for non-enumerated models.

fritzo: I'm happy to merge this if you `return loss + (surrogate_loss - surrogate_loss.detach())`. Do you want to make that change or just close this PR and use `TraceEnum_ELBO`?

eb8680: Done

PR #1253, opened 2018-07-22
change the way we initiate param for GP

fehiepsi:
A small change which is consistent with how AutoDelta and HMC init param/sample.

### DIFF
--- a/pyro/contrib/gp/util.py
+++ b/pyro/contrib/gp/util.py
@@ -129,8 +129,8 @@ class Parameterized(nn.Module):
             p = pyro.sample(param_name, prior)
         else:  # prior != None and mode = "guide"
             MAP_param_name = param_name + "_MAP"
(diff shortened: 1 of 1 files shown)


fehiepsi:
The failure is due to "unable to install matplotlib".

fritzo:
> I change it to a sample from prior.

I do not see a `.sample()` method being called. Did you intend to instead
```diff
- MAP_param = pyro.param(MAP_param_name, prior.mean.detach())
+ MAP_param = pyro.param(MAP_param_name, prior.sample())
```

fehiepsi:
@fritzo I thought that we can put a `callable` to `pyro.param(...)` and `prior()` is equivalent to `prior.sample()`: https://github.com/uber/pyro/blob/dev/pyro/distributions/distribution.py#L36.

PR #1254, opened 2018-07-22
add vectorized test for BernoulliBeta

fehiepsi:
The bug https://github.com/pytorch/pytorch/issues/9521 in pytorch master makes the vectorized BernoulliBeta test fail. We want to add a test to detect that bug in future versions of both Pyro and PyTorch.

### DIFF
--- a/tests/infer/test_inference.py
+++ b/tests/infer/test_inference.py
@@ -295,6 +295,8 @@ class BernoulliBetaTests(TestCase):
     def test_elbo_nonreparameterized(self):
         self.do_elbo_test(False, 10000, Trace_ELBO())
 
+    # these "vectorized" tests are used to detect bugs similar to
+    # the one in https://github.com/pytorch/pytorch/issues/9521
     def test_elbo_reparameterized_vectorized(self):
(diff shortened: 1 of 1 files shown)


fritzo:
Could you add a comment above each function pointing to the bug?

tests/infer/test_inference.py
```diff
@@ -295,6 +295,16 @@ def test_elbo_reparameterized(self):
     def test_elbo_nonreparameterized(self):
         self.do_elbo_test(False, 10000, Trace_ELBO())
 
+    # these "vectorized" tests are used to detect bugs similar to
+    # the one in https://github.com/pytorch/pytorch/issues/9521
+    def test_elbo_reparameterized_vectorized(self):
+        self.do_elbo_test(True, 5000, Trace_ELBO(num_particles=2, vectorize_particles=True,
+                                                 max_iarange_nesting=1))
+
+    def test_elbo_nonreparameterized_vectorized(self):
```
fritzo: sorry could you add a comment to each function? "these" is ambiguous and becomes more ambiguous as more tests are added.

fehiepsi: yup, I'll make the change :)

PR #1259, opened 2018-07-24
add regex search support for param store

jpchen:
requested by @ae-foster 

**Example Usage:**
```python
for i in range(3):
  pyro.param('a_{}'.format(i), torch.ones(1))
  pyro.param('b_{}'.format(i), torch.ones(1))

pyro.get_param_store().match('a_[0-9]+')
# {'a_0': tensor([ 1.]),
#  'a_1': tensor([ 1.]),
#  'a_2': tensor([ 1.])}
```

### DIFF
--- a/pyro/params/param_store.py
+++ b/pyro/params/param_store.py
@@ -123,18 +123,16 @@ class ParamStoreDict(object):
 
         return param
 
-    def get_params(self, name, constraint=constraints.real):
+    def match(self, name):
         """
         Get all parameters that match regex. The parameter must exist.
 
         :param name: regular expression
         :type name: str
-        :param constraint: torch constraint
-        :type constraint: torch.distributions.constraints.Constraint
         :returns: dict with key param name and value torch Tensor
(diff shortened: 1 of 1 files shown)


pyro/params/param_store.py
```diff
@@ -120,6 +123,21 @@ def get_param(self, name, init_tensor=None, constraint=constraints.real):
 
         return param
 
+    def get_params(self, name, constraint=constraints.real):
```
fritzo: 1. Consider renaming this to `.match()` or `.search()` or `.params_matching()`
2. I don't think you want a `constraint` kwarg; it is only useful for creating params in the param store (when combined with an `init_tensor`). I believe it will always be ignored here, since you're only iterating over params that already exist.

jpchen: 1. sure, though i wanted the name to denote returning multiple params
2. ah right the `_constraints` registry already has all the constrained values at call time

PR #1262, opened 2018-07-24
More flexibility in MADE

stefanwebb:
The MADE NN is used for inverse autoregressive flow. I've added the following features:

- changing the number of hidden layers in the MADE
- use skip connections from the input to the output
- changing the nonlinearity
- creating MADE masks for conditional distributions (to be used in the future)

The default parameters create the same type of MADE currently used in the IAF class, so this change doesn't break IAF


pyro/nn/auto_reg_nn.py
```diff
@@ -2,11 +2,75 @@
...
+    :type input_dim: int
+    :param hidden_dim: the dimensionality of the hidden layer
+    :type hidden_dim: int
+    :param simple: True to sample indices uniformly, false to space indices evenly and round up or down randomly
+    :type simple: bool
+    """
+    start_integer = 0 if conditional else 1
+
+    if simple:
+        return np.random.randint(start_integer, input_dim, size=(hidden_dim,))
```
jpchen: why not do this in torch?

pyro/nn/auto_reg_nn.py
```diff
@@ -54,73 +119,67 @@ class AutoRegressiveNN(nn.Module):
...
+        self.M_V = torch.FloatTensor(M_V)
+
+        # Create masked layers
+        layers = [MaskedLinear(input_dim, hidden_dim, self.M_W[0])]
+        for i in range(1, num_layers):
+            layers.append(MaskedLinear(hidden_dim, hidden_dim, self.M_W[i]))
+        self.layers = nn.ModuleList(layers)
+
+        if skip_connections:
+            self.M_A = torch.FloatTensor(M_A)
```
jpchen: why `FloatTensor`? shouldnt this be the default tensor type?

pyro/nn/auto_reg_nn.py
```diff
@@ -54,73 +119,67 @@ class AutoRegressiveNN(nn.Module):
...
-        """
-        Get the mask encoding associated with the neural network: basically the quantity m(k) in the MADE paper.
-        """
-        return self.mask_encoding
+        # Create masks
+        M_W, M_V, M_A = create_mask(input_dim=input_dim, observed_dim=0, hidden_dim=hidden_dim,
+                                    num_layers=num_layers, permutation=self.permutation,
+                                    output_dim_multiplier=output_dim_multiplier)
+        self.M_W = [torch.FloatTensor(M) for M in M_W]
+        self.M_V = torch.FloatTensor(M_V)
```
jpchen: same comment

pyro/nn/auto_reg_nn.py
```diff
@@ -2,11 +2,75 @@
...
+    :type num_layers: int
+    :param permutation: the order of the input variables
+    :type permutation: np.array
+    :param output_dim_multiplier: tiles the output (e.g. for when a separate mean and scale parameter are desired)
+    :type output_dim_multiplier: int
+    """
+    # Create mask indices for input, hidden layers, and final layer
+    # We use 0 to refer to the elements of the variable being conditioned on,
+    # and range(1:(D_latent+1)) for the input variable
+    m_input = np.concatenate((np.zeros(observed_dim), 1 + permutation))
```
jpchen: same comment as above

jpchen:
(note if you run `make lint` or `flake8 [directory]` it will tell you all your pep8 errors instead of running the CI to do so)

stefanwebb:
Ah thanks, let me rewrite those NumPy parts in PyTorch...

pyro/nn/auto_reg_nn.py
```diff
@@ -2,11 +2,75 @@
 
 import warnings
 
+import numpy as np
 import torch
 import torch.nn as nn
 from torch.nn import functional as F
 
 
+def sample_mask_indices(input_dim, hidden_dim, simple=False, conditional=True):
```
martinjankowiak: conditional is missing a :param:

martinjankowiak: in general the meanings of the two flags could be more carefully explained

pyro/nn/auto_reg_nn.py
```diff
@@ -2,11 +2,75 @@
 
 import warnings
 
+import numpy as np
```
martinjankowiak: as mentioned elsewhere, please remove all numpy usage

stefanwebb: Okay, I've removed all NumPy usage and made the requested changes

pyro/nn/auto_reg_nn.py
```diff
@@ -7,6 +7,75 @@
...
+        return ints
+
+
+def create_mask(input_dim, observed_dim, hidden_dim, num_layers, permutation, output_dim_multiplier):
+    """
+    Creates MADE masks for a conditional distribution
+
+    :param input_dim: the dimensionality of the input variable
+    :type input_dim: int
+    :param observed_dim: the dimensionality of the variable that is conditioned on (for conditional densities)
```
martinjankowiak: as i understand it this `observed_dim` functionality isn't being tested anywhere. can we add a test for this?

pyro/nn/auto_reg_nn.py
```diff
@@ -7,6 +7,75 @@
...
+        # "Non-simple" procedure creates fractional indices evenly then rounds at random
+        mk = torch.linspace(start_integer, input_dim - 1, steps=hidden_dim)
+
+        # Randomly decide whether to round up or down indices
+        ints = mk.floor()
+        ints += torch.bernoulli(mk - ints)
+        return ints
+
+
+def create_mask(input_dim, observed_dim, hidden_dim, num_layers, permutation, output_dim_multiplier):
```
martinjankowiak: in general can you please add tests that test the logic of `create_mask` separately?

pyro/nn/auto_reg_nn.py
```diff
@@ -7,6 +7,75 @@
...
+    :param permutation: the order of the input variables
+    :type permutation: torch.Tensor
+    :param output_dim_multiplier: tiles the output (e.g. for when a separate mean and scale parameter are desired)
+    :type output_dim_multiplier: int
+    """
+    # Create mask indices for input, hidden layers, and final layer
+    # We use 0 to refer to the elements of the variable being conditioned on,
+    # and range(1:(D_latent+1)) for the input variable
+    var_index = permutation.clone().type(torch.get_default_dtype())
+    var_index[permutation] = torch.arange(input_dim, dtype=torch.get_default_dtype())
```
martinjankowiak: @neerajprad can you please take a look at and comment on the way typing is handled in this PR?

pyro/nn/auto_reg_nn.py
```diff
@@ -49,78 +119,67 @@ class AutoRegressiveNN(nn.Module):
...
+        for i in range(1, num_layers):
+            layers.append(MaskedLinear(hidden_dim, hidden_dim, self.M_W[i]))
+        self.layers = nn.ModuleList(layers)
+
+        if skip_connections:
+            self.skip_p = MaskedLinear(input_dim, input_dim * output_dim_multiplier, self.M_A, bias=False)
+        else:
+            self.M_A = None
+            self.skip_p = None
+        self.p = MaskedLinear(hidden_dim, input_dim * output_dim_multiplier, self.M_V)
```
martinjankowiak: can `self.p` be given a more transparent name?

pyro/nn/auto_reg_nn.py
```diff
@@ -49,78 +119,67 @@ class AutoRegressiveNN(nn.Module):
         for any i, j in range(0, output_dim_multiplier) the subset of outputs [i, :] has identical
         autoregressive structure to [j, :]. defaults to `1`
     :type output_dim_multiplier: int
-    :param mask_encoding: a torch Tensor that controls the autoregressive structure (see reference). by default
-        this is chosen at random.
-    :type mask_encoding: torch.LongTensor
     :param permutation: an optional permutation that is applied to the inputs and controls the order of the
         autoregressive factorization. in particular for the identity permutation the autoregressive structure
-        is such that the Jacobian is upper triangular. by default this is chosen at random.
+        is such that the Jacobian is upper triangular. By default this is the identity permutation.
```
martinjankowiak: "By default this is the identity permutation." 

isn't this incorrect?

pyro/nn/auto_reg_nn.py
stefanwebb: Not at all! It's my first open-source contribution and a fun learning experience. :)

I think I've run it on the GPU, but let me double check.

Yes for the observed_dim, I can add a test that checks that *all* members of the jacobian are effected when you change the conditioning variable

pyro/nn/auto_reg_nn.py
stefanwebb: I did the `torch.get_default_dtype()` because I need to convert integers to floats but it will be 32- or 64-bit depending on the system I think. Is this consistent with the rest of Pytorch?

pyro/nn/auto_reg_nn.py
stefanwebb: Yes and I think some of the other variables like the mask names can have more transparent names too

pyro/nn/auto_reg_nn.py
stefanwebb: That's correct, let me change it

pyro/nn/auto_reg_nn.py
stefanwebb: I've removed observed_dim as I realized it isn't needed

pyro/nn/auto_reg_nn.py
stefanwebb: These are all renamed now, and I made it a bit more concise by eliminating self.p

pyro/nn/auto_reg_nn.py
stefanwebb: I think the logic of this is captured in the test for AutoRegressiveNN. If the mask is created incorrectly in any way then the Jacobian will not be triangular

pyro/nn/auto_reg_nn.py
stefanwebb: I've made a subtle change in how the MADE masks are created so that it matches the method in TensorFlow rather than the exact description in the MADE paper.

stefanwebb:
I've made a subtle change in how the MADE masks are created so that it matches the method in TensorFlow rather than the exact description in the MADE paper.

pyro/nn/auto_reg_nn.py
martinjankowiak: agreed that if there are "too many" connections then the jacobian will not be triangular but what if there are "too few"? how do we catch that? (might happen in some strange edge case, hidden dimension of 1 or 2 or what not)

stefanwebb: I think this could be handled by a separate test in `AutoRegressiveNNTests(TestCase)` that checks that all of the Jacobian above the diagaonal are nonzero

stefanwebb: I'm not sure how rigorous this would be though because if the weights in the network were small then these Jacobian elements could possibly underflow to 0

martinjankowiak: well that's what it might be nice to test the masks more directly. because the jacobian being triangular means that things are 'correct' in the sense that one will get sensible log probabilities but it still might not be the expected behavior in terms of the connectivity patterns of the masks

stefanwebb: Okay, let me look into it

pyro/nn/auto_reg_nn.py
```diff
@@ -7,6 +7,70 @@
...
+    :param permutation: the order of the input variables
+    :type permutation: torch.Tensor
+    :param output_dim_multiplier: tiles the output (e.g. for when a separate mean and scale parameter are desired)
+    :type output_dim_multiplier: int
+    """
+    # Create mask indices for input, hidden layers, and final layer
+    # We use 0 to refer to the elements of the variable being conditioned on,
+    # and range(1:(D_latent+1)) for the input variable
+    var_index = permutation.clone().type(torch.get_default_dtype())
+    var_index[permutation] = torch.arange(input_dim, dtype=torch.get_default_dtype())
```
neerajprad: Seems reasonable. Consider initializing as follows for clarity: 
```
var_index = torch.empty(permutation.shape, dtype=torch.get_default_dtype())
```

pyro/nn/auto_reg_nn.py
```diff
@@ -7,6 +7,70 @@
...
+
+    # Create the indices that are assigned to the neurons
+    input_indices = torch.cat((torch.zeros(observed_dim), 1 + var_index))
+    hidden_indices = [sample_mask_indices(input_dim, hidden_dim) for i in range(num_layers)]
+    output_indices = (var_index + 1).repeat(output_dim_multiplier)
+
+    # Create mask from input to output for the skips connections
+    mask_skip = (output_indices.unsqueeze(-1) > input_indices.unsqueeze(0)).type_as(output_indices)
```
neerajprad: I suppose we can replace all the `.type_as` with `.float()` instead. That will directly pick up the default tensor type whichever is used.

pyro/nn/auto_reg_nn.py
stefanwebb: Sorry for the delay - I've been thinking about this but haven't been able to decide on the best solution.

There are many different masks that are valid, it depends on how the indices are assigned to the hidden units. They could be sampled uniformly, evenly spaced and rounded, etc. It might not be best if the test was tied to a particular method.

Perhaps the test could start at the output and trace back through the masks to see which input units it's connected to? That way you could make sure that there aren't too few connections

martinjankowiak: well, at a minimum i imagined checking that 
-- each output is connected to at least one input
-- each input is connected to at least one output
perhaps there are other reasonable checks on top of that?

stefanwebb:
Okay, I've added a test that makes sure the right outputs are hooked up to the right inputs when the MADE masks are created. Also, I made the type changes

tests/distributions/test_iaf.py
```diff
@@ -96,6 +97,64 @@ def nonzero(x):
...
     def test_jacobians(self):
         for input_dim in [2, 3, 5, 7, 9, 11]:
             self._test_jacobian(input_dim, 3 * input_dim + 1, 2)
+
+    def test_masks(self):
+        for input_dim in [1, 3, 5]:
+            for observed_dim in [0, 3]:
+                for num_layers in [1, 3]:
+                    for output_dim_multiplier in [1, 2, 3]:
+                        # NOTE: the hidden dimension must be greater than the input_dim for the
```
martinjankowiak: thanks, these mask tests look great. 

"the hidden dimension must be greater..." is this checked in autoregressivenn somewhere?

stefanwebb: No but I've just pushed another commit that checks this. Also I added a comment about the mask structure

pyro/nn/auto_reg_nn.py
```diff
@@ -130,6 +133,10 @@ def __init__(self, input_dim, hidden_dim, output_dim_multiplier=1, permutation=N
         self.output_dim_multiplier = output_dim_multiplier
         self.num_layers = num_layers
 
+        # Hidden dimension must be not less than the input otherwise it isn't
+        # possible to connect to the outputs correctly
+        assert hidden_dim >= input_dim
```
martinjankowiak: nit: since this is user-facing this should be `raise ValueError("...")`

but with that change this PR should be good to merge once tests pass. thank you for your contribution!

stefanwebb:
done!

stefanwebb:
Ah that was because I had changed my `type_as()` to `float()`, but it seems `type_as()` is the way to go so you get the right precision floating-point type

stefanwebb:
There are some other tests that are failing, but I can't quite decipher why... I don't think it has to do with my commits

martinjankowiak:
well it seems to be `_test_masks` that's failing?

stefanwebb:
I think I've fixed it. The test was running fine on my machine, but because of the setup, `torch.arange` returned a different type on Travis. Once Travis restarts I think it's good this time

martinjankowiak:
@stefanwebb great, let's merge this! if it turns out the typing wasn't quite right, we can address in future PRs.

PR #1264, opened 2018-07-26
Implement MaskedMixture distribution

fritzo:
This implements a `MaskedMixture` distribution that can replace many of our uses of pairs of `poutine.scale` in models with element-wise dependencies. See docs and `test_em.py` for example usage.

About 90% of the logic is for broadcasting and validation, including two new constraints: `IndependentConstraint` and `MaskedConstraint`.

## Questions for reviewers

1. Which orientation is more intuitive?
    - `MaskedMixture(mask, if_zero, if_one)` is currently implemented. This is nice in that it is similar to a future `IndexedMixture(index, [case_0, case_1, ..., case_n])` where index is a `LongTensor`.
    - `MaskedMixture(cond, if_true, if_false)` is closer to `tf.Cond`
2. Should this be named something else like `Interleaved` or `CondDistribution` or other?

## Tested

- unit tests
- refactored `test_em.py` to use this distribution rather than `poutine.scale`

This PR was originally suggested by @eb8680

### DIFF
--- a/pyro/distributions/mixture.py
+++ b/pyro/distributions/mixture.py
@@ -46,10 +46,6 @@ class MaskedMixture(TorchDistribution):
         with pyro.iarange("data", len(data)):
             pyro.sample("obs", MaskedMixture(mask, dist1, dist2), obs=data)
 
-    Note that :ivar:`support` is set to ``component0.support`` due to lack of
-    programmatic ability to union constraints. To correctly set support of an
-    instance, simply overwrite its :ivar:`support` attribute.
-
     :param torch.Tensor mask: A byte tensor toggling between ``component0``
         and ``component1``.
     :param pyro.distributions.TorchDistribution component0: a distribution
--- a/pyro/distributions/mixture.py
+++ b/pyro/distributions/mixture.py
@@ -1,0 +1,93 @@
+from __future__ import absolute_import, division, print_function
+
+import torch
+from torch.distributions.utils import lazy_property
+
+from pyro.distributions.torch_distribution import TorchDistribution
+
+
+class MaskedMixture(TorchDistribution):
+    """
+    A masked deterministic mixture of two distributions.
+
+    This is useful when the mask is sampled from another distribution,
+    possibly correlated across the batch. Often the mask can be
+    marginalized out via enumeration.
+
+    Example::
+
+        change_point = pyro.sample("change_point",
+                                   dist.Categorical(torch.ones(len(data) + 1)),
+                                   infer={'enumerate': 'parallel'})
+        mask = torch.arange(len(data), dtype=torch.long) >= changepoint
+        with pyro.iarange("data", len(data)):
+            pyro.sample("obs", MaskedMixture(mask, dist1, dist2), obs=data)
+
+    Note that :ivar:`support` is set to ``component0.support`` due to lack of
+    programmatic ability to union constraints. To correctly set support of an
+    instance, simply overwrite its :ivar:`support` attribute.
+
+    :param torch.Tensor mask: A byte tensor toggling between ``component0``
+        and ``component1``.
+    :param pyro.distributions.TorchDistribution component0: a distribution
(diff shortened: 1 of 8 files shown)


tests/contrib/tracking/test_em.py
```diff
@@ -54,15 +54,12 @@ def model(detections, args):
         assign_probs[-1] = p_fake
         assign = pyro.sample('assign', dist.Categorical(logits=assign_probs))
         is_fake = (assign == assign.shape[-1] - 1)
```
fritzo: This is the motivating use case

pyro/distributions/mixture.py
```diff
@@ -0,0 +1,134 @@
...
+                             .format(component0.event_shape, component1.event_shape))
+        batch_shape = broadcast_shape(mask.shape, component0.batch_shape, component1.batch_shape)
+        if mask.shape != batch_shape:
+            mask = mask.expand(batch_shape)
+        if component0.batch_shape != batch_shape:
+            component0 = component0.expand(batch_shape)
+        if component1.batch_shape != batch_shape:
+            component1 = component1.expand(batch_shape)
+
+        self.mask = mask
```
alicanb: check if `mask` is bytetensor here?

alicanb:
I think we can also expand this into list of distributions and a longtensor mask, where `sample` will call `torch.gather` or something

fritzo:
> I think we can also expand this into list of distributions and a longtensor mask

Yeah I was thinking we could keep a few implementations in `mixture.py`, starting with `MaskedMixture` and then implementing an `IndexedMixture` with a `LongTensor` mask. I think it's cleaner to keep the implementations separate, just as we do for `Bernoulli` vs `Categorical`.

pyro/distributions/mixture.py
```diff
@@ -0,0 +1,136 @@
...
+        change_point = pyro.sample("change_point",
+                                   dist.Categorical(torch.ones(len(data) + 1)),
+                                   infer={'enumerate': 'parallel'})
+        mask = torch.arange(len(data), dtype=torch.long) >= changepoint
+        with pyro.iarange("data", len(data)):
+            pyro.sample("obs", MaskedMixture(mask, dist1, dist2), obs=data)
+
+    :param torch.Tensor mask: A byte tensor toggling between ``component0``
+        and ``component1``.
+    :param pyro.distributions.TorchDistribution component0: a distribution
```
jpchen: "corresponding to 0 in the mask"

pyro/distributions/mixture.py
```diff
@@ -0,0 +1,136 @@
...
+        if component1.batch_shape != batch_shape:
+            component1 = component1.expand(batch_shape)
+
+        self.mask = mask
+        self.component0 = component0
+        self.component1 = component1
+        super(MaskedMixture, self).__init__(batch_shape, component0.event_shape, validate_args)
+
+        # We need to disable _validate_sample on each component since samples are only valid on the
+        # component from which they are drawn. Instead we perform validation using a MaskedConstraint.
```
jpchen: neat

fritzo: It seems kludgey to me, but I haven't seen a way around it. Let's keep this shortcoming in mind as we evolve the `torch.distributions` validation patterns. cc @alicanb @vishwakftw

pyro/distributions/torch_distribution.py
```diff
@@ -225,6 +225,33 @@ class TorchDistribution(torch.distributions.Distribution, TorchDistributionMixin
     pass
 
 
+# TODO move this upstream to torch.distributions
```
fritzo: @alicanb Let's try to move this constraint upstream by next Pytorch release, and use it in the `Independent` distribution and all other multivariate distributions.

fritzo:
@alicanb I think we can merge this as soon as it passes tests.

alicanb:
Thanks @fritzo !

PR #1266, opened 2018-07-26
 Added flag to allow loading of CUDA-trained parameters to CPU.

ztangent:
After training a model on a GPU cluster, I found myself needing to load the Pyro parameter store on a non-CUDA-enabled device. This quick patch adds a flag to make that possible using this method:
https://discuss.pytorch.org/t/on-a-cpu-device-how-to-load-checkpoint-saved-on-gpu-device/349/4

### DIFF
--- a/pyro/params/param_store.py
+++ b/pyro/params/param_store.py
@@ -165,23 +165,23 @@ class ParamStoreDict(object):
         Save parameters to disk
 
         :param filename: file name to save to
-        :type name: str
+        :type filename: str
         """
         with open(filename, "wb") as output_file:
             torch.save(self.get_state(), output_file)
 
-    def load(self, filename, load_to_cpu=False):
+    def load(self, filename, map_location=None):
         """
         Loads parameters from disk
 
         :param filename: file name to load from
-        :type name: str
-        :param load_to_cpu: flag to load CUDA-trained parameters to CPU
-        :type name: bool
+        :type filename: str
+        :param map_location: specifies how to remap storage locations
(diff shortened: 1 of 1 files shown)


pyro/params/param_store.py
```diff
@@ -170,15 +170,20 @@ def save(self, filename):
...
 
-    def load(self, filename):
+    def load(self, filename, load_to_cpu=False):
         """
         Loads parameters from disk
 
         :param filename: file name to load from
         :type name: str
+        :param load_to_cpu: flag to load CUDA-trained parameters to CPU
+        :type name: bool
```
jpchen: this is actually a typo above it should be  `:type filename: str` and `:type load_to_cpu: bool`

pyro/params/param_store.py
```diff
@@ -170,15 +170,20 @@ def save(self, filename):
         with open(filename, "wb") as output_file:
             torch.save(self.get_state(), output_file)
 
-    def load(self, filename):
+    def load(self, filename, load_to_cpu=False):
```
fritzo: Why not simply thread through the `map_location` arg to `torch.load`? Then you could achieve what you want with
```py
pyro.load(filename, map_location=lambda s,l: 'cpu')
```
and also load to gpu via e.g.
```py
pyro.load(filename, map_location=lambda s,l: 'cuda:0')
```

ztangent:
Made the requested changes!

ztangent:
Wow, was extremely careless and left in the 'if load_to_cpu' line, but everything is fixed now.

Issue #1267, opened 2018-07-26
Update to PyTorch 0.4.1


fritzo:
- [ ] #1294 Fix failing CPU tests (@neerajprad, @fritzo)
    - [x] Update .travis.yml
    - [x] Work around https://github.com/pytorch/pytorch/issues/9917 bug in `torch.bernoulli(p.expand(s))`
    - [x] Work around https://github.com/pytorch/pytorch/issues/9521 bug in `Dirichlet` of expanded tensors
    - [x] Work around https://github.com/pytorch/pytorch/issues/9977 `RuntimeError` in `test_em.py` (@fritzo)
    - [ ] Work around https://github.com/pytorch/pytorch/issues/10241 broken `log`, `exp`
- [x] #1271 Use `Independent` rather than `ReshapedDistribution` (@alicanb)
- [x] #1273 Clamp to nonnegative before `.sqrt()` in  `newton_step_2d()`
- [x] #1274 Use `torch.logsumexp` when available
- [x] #1276 Update `pyro.ops.jit` (@eb8680, @fritzo)
- [x] #1268 Wrap all new distributions in pyro/distributions/torch.py (@fritzo)
- [x] Fix failing GPU tests (@neerajprad , @jpchen )
- [ ] Update docker files (@neerajprad)
- [ ] Cut a Pyro release 0.2.2 (@fritzo, @jpchen)

neerajprad:
I will start with the tests / breaking code on `dev` and update `.travis.yml`. We can cover the remaining tasks in parallel.

fritzo:
Great, I'll start migrating our distributions in #1268. This will fail CI until we switch torch version on travis.

alicanb:
I just merged #1264 into dev, don't forget to pull before pushing!

eb8680:
Is `pyro.ops.jit.compile` deprecated/obsolete under PyTorch 0.4.1? Do we need to remove it?

alicanb:
https://github.com/pytorch/pytorch/pull/10269 -> for expand bugs

PR #1268, opened 2018-07-27
Move HalfCauchy,Binomial upstream

fritzo:
Addresses #1267


jpchen:
ill wait till @neerajprad's PR bumping pytorch's version in travis to rerun these tests before merging

tests/distributions/conftest.py
```diff
@@ -204,17 +204,6 @@
                  'test_data': [[0.4], [0.35]]}
             ],
             scipy_arg_fn=lambda loc, scale: ((), {"loc": np.array(loc), "scale": np.array(scale)})),
-    Fixture(pyro_dist=dist.HalfCauchy,
```
neerajprad: I suppose we can keep this fixture as is? These tests are fast, and help in catching any errors due to the wrapping layers (e.g. `.expand_by`).

fritzo:
@jpchen Yeah, I'll dump all distributions changes into this PR.

tests/distributions/conftest.py
fritzo: Sure, will do. I'll just remove the `loc` argument, which does not exist in the upstream version.

pyro/distributions/torch.py
```diff
@@ -124,6 +139,15 @@ def expand(self, batch_shape):
             return type(self)(loc, scale, validate_args=validate_args)
 
 
+class HalfCauchy(torch.distributions.HalfCauchy, TorchDistributionMixin):
+    def expand(self, batch_shape):
+        try:
+            return super(HalfCauchy, self).expand(batch_shape)
+        except NotImplementedError:
+            scale = self.scale.expand(batch_shape)
+            return type(self)(scale)
```
fehiepsi: @fritzo We also have HalfNormal :)

fritzo: done.

neerajprad:
This should be covered in our `pytorch-1.0` branch now.

PR #1271, opened 2018-07-27
make TorchDistributionsMixin.independent() return dist.Independent

alicanb:
- `TorchDistributionsMixin.independent()` now returns `dist.Independent`
- fix bug in `Independent.expand`
- add tests that would have caught the bug.

### DIFF
--- a/pyro/distributions/torch.py
+++ b/pyro/distributions/torch.py
@@ -125,12 +125,23 @@ class Gumbel(torch.distributions.Gumbel, TorchDistributionMixin):
 
 
 class Independent(torch.distributions.Independent, TorchDistributionMixin):
+    @constraints.dependent_property
+    def support(self):
+        return IndependentConstraint(self.base_dist.support, self.reinterpreted_batch_ndims)
+
+    @property
+    def _validate_args(self):
+        return self.base_dist._validate_args
+
+    @_validate_args.setter
+    def _validate_args(self, value):
+        self.base_dist._validate_args = value
+
     def expand(self, batch_shape):
         batch_shape = torch.Size(batch_shape)
         validate_args = self.__dict__.get('_validate_args')
-        extra_shape = self.base_dist.event_shape[:self.reinterpreted_batch_ndims]
-        base_dist = self.base_dist.expand(batch_shape + extra_shape)
-        return Independent(base_dist, self.reinterpreted_batch_ndims, validate_args=validate_args)
+        base_dist = self.base_dist.expand(batch_shape + self.event_shape)
+        return type(self)(base_dist, self.reinterpreted_batch_ndims, validate_args=validate_args)
 
 
 class Laplace(torch.distributions.Laplace, TorchDistributionMixin):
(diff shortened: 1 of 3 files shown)


pyro/distributions/torch.py
```diff
@@ -125,12 +125,23 @@ def expand(self, batch_shape):
...
+    def _validate_args(self, value):
+        self.base_dist._validate_args = value
+
     def expand(self, batch_shape):
         batch_shape = torch.Size(batch_shape)
         validate_args = self.__dict__.get('_validate_args')
-        extra_shape = self.base_dist.event_shape[:self.reinterpreted_batch_ndims]
-        base_dist = self.base_dist.expand(batch_shape + extra_shape)
-        return Independent(base_dist, self.reinterpreted_batch_ndims, validate_args=validate_args)
+        base_dist = self.base_dist.expand(batch_shape + self.event_shape)
```
fritzo: Should this be
```py
base_dist = self.base_dist.expand(batch_shape + self.base_dist.batch_shape)
```
to handle multivariate `base_dist`?

fritzo: Or hmm maybe
```py
reinterpreted_shape = self.base_dist.batch_shape[len(self.base_dist.batch_shape) - self.reinterpreted_batch_ndims]
base_dist = self.base_dist.expand(batch_shape + reinterpreted_shape)
return base_dist.independent(self.reinterpreted_batch_ndims)
```

pyro/distributions/torch_distribution.py
```diff
@@ -5,6 +5,7 @@
 import torch
 from torch.distributions import biject_to, constraints, transform_to
 
+import pyro.distributions
```
fritzo: Hmm, maybe it would be a bit less circular to create a smaller depencency cycle? Does this work?
```py
import pyro.distributions.torch
```

tests/distributions/test_independent.py
```diff
@@ -0,0 +1,59 @@
...
+@pytest.mark.parametrize('reinterpreted_batch_ndims', [0, 1, 2, 3])
+@pytest.mark.parametrize('base_dist',
+                         [dist.Normal(1., 2.), dist.Exponential(2.),
+                          dist.MultivariateNormal(torch.zeros(2), torch.eye(2))],
+                         ids=['normal', 'exponential', 'mvn'])
+def test_independent(base_dist, sample_shape, batch_shape, reinterpreted_batch_ndims):
+    if batch_shape:
+        base_dist = base_dist.expand_by(batch_shape)
+    if reinterpreted_batch_ndims > len(base_dist.batch_shape):
+        with pytest.raises(ValueError):
```
fritzo: nice!

alicanb:
Thanks @fritzo for the review!

pyro/distributions/torch.py
```diff
@@ -125,12 +125,25 @@ def expand(self, batch_shape):
...
+    def _validate_args(self):
+        return self.base_dist._validate_args
+
+    @_validate_args.setter
+    def _validate_args(self, value):
+        self.base_dist._validate_args = value
+
     def expand(self, batch_shape):
         batch_shape = torch.Size(batch_shape)
         validate_args = self.__dict__.get('_validate_args')
```
fritzo: Hmm I think we actually don't want to pass through `_validate_args`, since we're treating this as a special property with a setter. I think we should instead 
```py
return type(self)(base_dist, self.reinterpreted_batch_ndims)  # <-- no validate_args
```

PR #1273, opened 2018-07-28
Clamp to nonnegative before .sqrt() in newton_step_2d()

fritzo:
Addresses #1267 

This is needed due to finickier `.sqrt()` function in PyTorch 0.4.1. This removes a NAN issue in an internal application.


fritzo:
This does not fix the `test_em.py` errors; we'll need to continue debugging...

fehiepsi:
@fritzo @neerajprad I think clamping to zero is not a solution (gradient of sqrt at 0 will give you `inf`). It is better to add an epsilon term after clamping by 0 or clamping with an epsilon term. For example,

```
x = torch.zeros(1, requires_grad=True)
y = x.sqrt()
x.backward()
x.grad  # 'inf'
```

In gp module, I made [torch_sqrt](https://github.com/uber/pyro/blob/dev/pyro/contrib/gp/kernels/isotropic.py#L10) to solve this problem.

PR #1274, opened 2018-07-28
Use torch.logsumexp when available

fritzo:
Addresses #1267 

This replaces our internal `pyro.distributions.util.log_sum_exp` with `torch.logsumexp` when available, i.e. after pytorch 0.4.1. The old implementation is kept as a fallback, and I've kept a `log_sum_exp` alias to avoid breaking internal applications.

### DIFF
--- a/pyro/distributions/util.py
+++ b/pyro/distributions/util.py
@@ -188,16 +188,19 @@ def torch_sign(value):
     return torch.sign(value)
 
 
-def log_sum_exp(tensor, dim=-1):
-    """
-    Numerically stable implementation for the `LogSumExp` operation. The
-    summing is done along the dimension specified by ``dim``.
+try:
+    from torch import logsumexp  # for pytorch 0.4.1 and later
+except ImportError:
+    def logsumexp(tensor, dim=-1):
+        """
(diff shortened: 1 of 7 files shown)


pyro/distributions/util.py
```diff
@@ -188,16 +188,22 @@ def torch_sign(value):
...
+        summing is done along the dimension specified by ``dim``.
+
+        :param torch.Tensor tensor: Input tensor.
+        :param dim: Dimension to be summed out.
+        """
+        max_val = tensor.max(dim)[0]
+        return max_val + (tensor - max_val.unsqueeze(dim)).exp().sum(dim=dim).log()
+
+
+log_sum_exp = logsumexp  # DEPRECATED
```
neerajprad: Since this isn't part of the public API, I suppose we can take the liberty of deleting it with the update?

fritzo: We're actually using this internally in a customer pipeline, and I'd like to update that pipeline before deleting this alias.

neerajprad: Got it, makes sense.

fritzo:
Thanks for reviewing, @neerajprad !

PR #1276, opened 2018-07-29
Update JIT usage to PyTorch 0.4.1

fritzo:
Addresses #1267, #1063

@eb8680 I've set the base to the `fix-dist-0.4.1` branch, so feel free to merge this at any time.

Some differences include:
1. I've temporarily dropped support for dynamically created params. This is partly due to 0.4.1's new restriction that inputs should be an `*args` of tensors, not a recursive `arg ::= list[arg] | tensor`. We can add this back in a later PR.
2. `@pyro.ops.jit.trace` now differs from `torch.jit.trace` in that it is lazy and does not require specifying all args on input. This seems easier to use, and makes it easier to splice in params to the `*args` sent to the compiled function.
    ```py
    @torch.jit.trace(*args)  # <--- requires *args
    def function(*args):
        ...

    @pyro.ops.jit.trace  # <--- does not require *args
    def function(*args):
        ...
    ```

More of our tests now pass, but some `JitTrace*_ELBO` tests still fail.

### DIFF
--- a/pyro/distributions/torch_patch.py
+++ b/pyro/distributions/torch_patch.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
+from numbers import Number
+
 import torch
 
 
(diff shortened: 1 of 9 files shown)


pyro/distributions/torch_patch.py
```diff
@@ -45,4 +47,51 @@ def _torch_dirichlet_grad(x, concentration, total):
...
+        ValueError: if any of the values is not a `numbers.Number` or
+            `torch.*Tensor` instance
+    """
+    values = list(values)
+    scalar_idxs = [i for i in range(len(values)) if isinstance(values[i], Number)]
+    tensor_idxs = [i for i in range(len(values)) if values[i].__class__.__name__ == 'Tensor']
+    if len(scalar_idxs) + len(tensor_idxs) != len(values):
+        raise ValueError('Input arguments must all be instances of numbers.Number or torch.tensor.')
+    if tensor_idxs:
+        broadcast_shape = sum(values).size()  # expensive alternative to torch._C._infer_size()
```
neerajprad: Calling `.size` ends up returning a tuple of pytorch scalars when JITed. I wonder why this gets triggered only in `TraceEnum_ELBO` though.

neerajprad:
Thanks for digging into this. Good to merge?

fritzo:
Good to merge.

PR #1269, opened 2018-07-27
Fix PyTorch 0.4.1 errors

fritzo:
Addresses #1267

PR #1277, opened 2018-07-30
Fix failing CUDA tests on dev

neerajprad:
Addresses #1267.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,9 +1,14 @@
 from __future__ import absolute_import, division, print_function
 
+import os
 import warnings
+
+import torch
 
 import pyro
 
+
+torch.set_default_tensor_type(os.environ.get('PYRO_TENSOR_TYPE', 'torch.DoubleTensor'))
 
 def pytest_configure(config):
     config.addinivalue_line("markers",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,6 +10,7 @@ import pyro
 
 torch.set_default_tensor_type(os.environ.get('PYRO_TENSOR_TYPE', 'torch.DoubleTensor'))
 
+
 def pytest_configure(config):
     config.addinivalue_line("markers",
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -1,10 +1,15 @@
...
 
+import os
 import warnings
 
+import torch
+
 import pyro
 
 
+torch.set_default_tensor_type(os.environ.get('PYRO_TENSOR_TYPE', 'torch.DoubleTensor'))
```
neerajprad: Since some of our tests don't use utility functions from `common.py` (by directly using `assert`), moving this from `common.py` to `conftest.py` which is called for each test.

tests/optim/test_optim.py
```diff
@@ -100,16 +101,16 @@ def guide():
...
+            scale = pyro.param('scale').unconstrained()
             opt = scheduler.optim_objs[loc].optimizer
             assert opt.state_dict()['param_groups'][0]['lr'] == 0.02
             assert opt.state_dict()['param_groups'][0]['initial_lr'] == 0.01
             opt = scheduler.optim_objs[scale].optimizer
             assert opt.state_dict()['param_groups'][0]['lr'] == 0.02
             assert opt.state_dict()['param_groups'][0]['initial_lr'] == 0.01
             assert abs(pyro.param('loc').item()) > 1e-5
-            assert abs(pyro.param('scale').item()) - 0.5 > 1e-5
+            assert abs(pyro.param('scale').item() - 0.5) > 1e-5
```
neerajprad: @jpchen - I was getting scale has invalid values validation error so added a positive constraint.

jpchen: i noticed that error but never got around to fixing it. thanks for fixing it 😄

neerajprad:
Good to merge. I think the remaining failures are not CUDA related, but will check again once #1269 is merged.

PR #1278, opened 2018-07-30
Fix geometric() in tutorial

cfperez:
As currently implemented, it's is actually doing geometric(1-p).

Also, the print outputs from cells were not included.


alicanb:
minor nit: There are 2 ways to characterize `Geometric`, count of trials, and count of failures. Can we make the implementation here to be count of failures (i.e. `return x` -> `return 0`) so that it's consistent with `torch.distributions.Geometric`?

PR #1279, opened 2018-07-31
Add expand kwarg to .enumerate_support()

fritzo:
Addresses #915
Motivated by #1289

This adds an `expand` kwarg to the `TorchDistribution.enumerate_support()` interface. By default `expand=True` and the resulting values are of shape
```py
expanded_shape == (cardinality,) + batch_shape + event_shape
```
This PR allows a new setting of `expand=False` yielding values of shape
```py
compressed_shape == (cardinality,) + (1,)*len(batch_shape) + event_shape
```
which is useful in message passing enumeration.

Once we decide on the interface, we'll push these changes upstream to PyTorch distributions.

## Distributions

- [x] Independent
- [x] ReshapedDistribution
- [x] MaskedDistribution
- [x] Bernoulli
- [x] Categorical
- [x] OneHotCategorical
- [x] Binomial
- [x] Empirical

## Tests

- [x] tests for `expand=True` plumbing
- [x] tests for `expand=False` behavior

Joint work with @eb8680.


eb8680:
The distribution interface change seems reasonable.  What should the model-level interface be?  Should we add a keyword argument to `EnumerateMessenger` and make dependent enumeration the new default?

fritzo:
I think we can simply make dependent enumeration the new default. I haven't thought through the value shape issues though. I think this should be ok within the tensor DSL since we're now using `@poutine.broadcast` inside enumeration.

fritzo:
@eb8680 it looks like we will have to use `@poutine.broadcast` by default now inside `TraceEnum_ELBO`. I think we'll also need to do more hand-broadcasting, e.g. adding `torch.zeros(width, height)` in the tensor shapes tutorial.

Have you already dealt with these issues in your previous branch (that inspired this PR)?

fritzo:
Hmm the changed shape semantics are looking like a bigger change than I anticipated. I think it might be safer to gate this behind a flag to `TraceEnum_ELBO`  (maybe named `expand=False` or `dependent=False`). Then we can flip the flag's default value to `expand=True` in our 0.3 release.

fritzo:
OK I've moved inference changes to a follow-up PR #1289. I think this first PR should be ready to merge.

PR #1283, opened 2018-08-03
Minor fixes to the VAE tutorial

neerajprad:
Making some minor fixes that were brought to my notice - mostly clearing the param store so that we don't get unexpected results when we rerun the same cell, and removing `enumerate` when not needed. 

Made some other smaller modifications to get the tutorials to run locally:
 - Adding encoding to unpickling in DMM without which I am unable to run the example on python 3.6 locally (seems like this is a common issue with numpy in python 3.x).
 - Also realized that `CI=1` flag wasn't actually getting passed, so our `test-examples` stage was taking much longer than it should (since we were running more than 1 epoch)!


.travis.yml
```diff
@@ -65,8 +65,8 @@ jobs:
           env: STAGE=examples
           script:
               - pytest -vs --cov=pyro --stage test_examples
-              - CI=1 grep -l smoke_test tutorial/source/*.ipynb | xargs grep -L 'smoke_test = False' \
-                  | xargs pytest -vx --nbval-lax --current-env
+              - grep -l smoke_test tutorial/source/*.ipynb | xargs grep -L 'smoke_test = False' \
+                  | CI=1 xargs pytest -vx --nbval-lax --current-env
```
fritzo: :+1: good to know!

PR #1284, opened 2018-08-03
bump travis docs version

jpchen:
[sphinx doesnt support ABC in python 2](https://github.com/sphinx-doc/sphinx/issues/4059) which we need. testing to see if this fixes the broken doctests. 

**this will require everyone developing pyro to upgrade sphinx to** `1.7.6`

### DIFF
--- a/pyro/contrib/tracking/hashing.py
+++ b/pyro/contrib/tracking/hashing.py
@@ -30,10 +30,11 @@ class LSH(object):
         >>> lsh.add('a', a)
         >>> lsh.add('b', b)
         >>> lsh.add('c', c)
-        >>> lsh.nearby('a') # even though c is within 2radius of a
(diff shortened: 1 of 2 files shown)


pyro/contrib/tracking/hashing.py
```diff
@@ -31,12 +31,12 @@ class LSH(object):
         >>> lsh.add('b', b)
         >>> lsh.add('c', c)
         >>> lsh.nearby('a') # even though c is within 2radius of a
-        set(['b'])
+        {'b'}
```
neerajprad: I don't think this will work in python 2.7..maybe its best to to skip the output check for these -  # doctest: +SKIP.

neerajprad:
@jpchen - good to merge?

jpchen:
Y

PR #1285, opened 2018-08-03
Separate models from examples/tests; benchmarking script

ae-foster:
This PR does several things:
 - allow unknown covariance in linear models and introduce the normal inverse gamma guide family for inference
 - reorganize the code to separate models from examples
 - add a benchmarking script - right now comparing VI against exact answers only

### DIFF
--- a/examples/contrib/oed/eig_estimation_benchmarking.py
+++ b/examples/contrib/oed/eig_estimation_benchmarking.py
@@ -48,7 +48,8 @@ def lm_true_ape(X_lm, sqrtlambda, obs_sd):
 
 
 @pytest.mark.parametrize("arglist",
-    [[(X_lm, vi_for_lm, torch.tensor([.1, .4]), None, torch.tensor(10.), torch.tensor(10.), 5000, 10)],
+    [# Warning: do not do this, not a mean-field guide!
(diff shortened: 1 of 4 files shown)


examples/contrib/oed/eig_estimation_benchmarking.py
```diff
@@ -0,0 +1,82 @@
+import time
```
martinjankowiak: is this supposed to work? when i run this i get "UserWarning: Encountered NaN: loss"

ae-foster: Yeah I saw this once before. What happens if you run it again?

martinjankowiak: ah, seems to be a pytorch 0.4.1 versus 0.4 problem.

@neerajprad @fritzo if you're looking for 0.4.1/0.4 discrepancies try

pytest -s -v examples/contrib/oed/eig_estimation_benchmarking.py 

works fine in 0.4 but i get "UserWarning: Encountered NaN: loss" in 0.4.1

no idea why

ae-foster: Hmmm... I'll stick to 0.4 for local testing for now

PR #1288, opened 2018-08-05
Add Codecov to Travis

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

Example report https://codecov.io/gh/gliptak/pyro/branch/codecov1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ jobs:
         - python: 2.7
           env: STAGE=examples
           script:
-              - pytest -vs --cov=pyro --stage test_examples
+              - pytest -vs --stage test_examples
               - grep -l smoke_test tutorial/source/*.ipynb | xargs grep -L 'smoke_test = False' \
                   | CI=1 xargs pytest -vx --nbval-lax --current-env
         - python: 3.5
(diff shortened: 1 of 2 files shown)


neerajprad:
Thanks for adding a code coverage report to Pyro. Just to confirm - it looks like this is using all the test stages from travis? Would it be simple to remove one or more stages from the coverage report, say the examples stage for instance?

neerajprad:
> Would it be simple to remove one or more stages from the coverage report, say the examples stage for instance?

I think we can just remove the `--cov` invocation from pytest for those stages. We can do that later (it won't really affect the marginal code coverage so we could leave it in as well) and add a badge to our README too. 

LGTM. @jpchen, @fritzo - Let me know if you have any thoughts / comments on inc. a code coverage report for PRs.

jpchen:
> I think we can just remove the --cov invocation from pytest for those stages

sounds fine to me. i dont really know how accurate this is, afaik it does static analysis of coverage and cant do anything fancy like testing going down different branches in models with dynamic control flow

gliptak:
@jpchen Thank you for the pointer
@neerajprad The easiest way to exclude would be to not generate coverage for example
https://github.com/uber/pyro/blob/dev/.travis.yml#L75
I will update the PR

neerajprad:
> @neerajprad The easiest way to exclude would be to not generate coverage for example
https://github.com/uber/pyro/blob/dev/.travis.yml#L75

That's right. we can just remove it from the examples stage.

gliptak:
Updated as per above, please review

neerajprad:
Looks good! Thanks for contributing.

PR #1289, opened 2018-08-06
Support .enumerate_support(expand=False) in SVI

fritzo:
Addresses #915 "dependent enumeration"

This adds support for unexpanded discrete enumeration as required by message passing inference. Under the hood this allows calling `.enumerate_support(expand=False)`. The interface is a new `expand` kwarg accessible in both the infer dict of guide sites and the `config_enumerate()` decorator.

New usage requires changes to user handling of the values of sample sites. E.g. for a Bernoulli site for MNIST pixels, we have many possible interpretations:
```py
@poutine.broadcast
def model(enumerate_, expand):
    with pyro.iarange("x_axls", 28), pyro.iarange("y_axis", 28):
        image = pyro.sample("image", dist.Bernoulli(0.5),
                            infer={"enumerate": enumerate_, "expand": expand})

    # check shapes
    if enumerate_ == "parallel":
        if expand:
            assert image.shape == (2, 28, 28)  # expanded parallel enumeration
        else:
            assert image.shape == (2, 1, 1)  # unexpanded parallel enumeration
    elif enumerate_ == "sequential":
        if expand:
            assert image.shape == (28, 28)  # expanded sequential enumeration
        else:
            assert image.shape == (1, 1)  # unexpanded sequential enumeration
    else:
        assert image.shape == (28, 28)  # sample semantics
        
```
Since this would be a breaking change, this PR preserves the `expand=True` behavior as default. I'd like to flip the default to `expand=False` in our 0.3 release.

Note this also weakens the errors reported by `poutine.broadcast` by silently ignoring `iarange`s with unspecified size. This is convenient since `TraceEnum_ELBO` now uses `poutine.broadcast` internally, and we still want to accept user models that don't bother with size annotations.

## Tests

- [x] added new params and assertions to `test_enum.py`
- [x] added new params and assertions to `test_valid_models.py`


pyro/poutine/enumerate_messenger.py
```diff
@@ -2,6 +2,8 @@
 
 from .messenger import Messenger
 
+EXPAND_DEFAULT = True  # TODO(fritzo,eb8680) Flip this to False in Pyro 0.3.
```
fritzo: I've created a global default to make it easier for us to flip behavior in 0.3.

tests/test_examples.py
```diff
@@ -80,12 +82,14 @@ def test_coverage():
 
 @pytest.mark.parametrize('example,args', CPU_EXAMPLES, ids=make_ids(CPU_EXAMPLES))
 def test_cpu(example, args):
+    logger.info('Running:\npython examples/{} {}'.format(example, ' '.join(args)))
```
fritzo: These logging statements make it easier to reproduce failures on travis.

tests/infer/test_valid_models.py
```diff
@@ -967,16 +1022,42 @@ def model():
...
         with y_iarange:
             c = pyro.sample("c", dist.Bernoulli(0.5))
-            if enumerate_ == "parallel":
-                assert c.shape == torch.Size((2, num_particles, 6, 1) if num_particles > 1 else (2, 6, 1))
-            else:
-                assert c.shape == torch.Size((num_particles, 6, 1) if num_particles > 1 else (6, 1))
         with x_iarange, y_iarange:
             d = pyro.sample("d", dist.Bernoulli(b))
+
+        # check shapes
```
neerajprad: Thanks for reorganizing these asserts and making them so much more readable!

tests/infer/test_valid_models.py
```diff
@@ -852,79 +858,123 @@ def guide():
...
-                else:
-                    assert d.shape == torch.Size((50, 6, 5))
 
-    assert_ok(model, model, TraceEnum_ELBO(max_iarange_nesting=3,
-                                           strict_enumeration_warning=(enumerate_ == "parallel")))
+        # check shapes
+        assert b.shape == (50, 1, 5)
+        if enumerate_ == "parallel":
+            if expand:
+                assert c.shape == (2, 50, 6, 1)  # FIXME actual shape == (2, 1, 1, 1)
```
neerajprad: The results seem fine to me, could you expand on the `FIXME`s here and below?

pyro/infer/enum.py
```diff
@@ -34,6 +34,8 @@ def get_importance_trace(graph_type, max_iarange_nesting, model, guide, *args, *
     Returns a single trace from the guide, and the model that is run
     against it.
     """
+    guide = poutine.broadcast(guide)
```
neerajprad: With this change, can we remove `poutine.broadcast` annotation from our examples completely (since it will be done automatically behind the scenes)?

tests/infer/test_valid_models.py
fritzo: opps, I fixed these in 53d4b30. will remove.

pyro/infer/enum.py
fritzo: Well, we still need `poutine.broadcast` for the model, and we'll still need it to run the guide under vanilla `Trace_ELBO`. I think the clearest policy would be to use `poutine.broadcast` for any model/guide that requires broadcasting when run outside of SVI.

neerajprad:
> @neerajprad can we chat tomorrow about making a similar change to HMC?

Lets chat today, I think the changes might be minimal after this.

pyro/infer/enum.py
neerajprad: Yes, we'll still need it for HMC for instance. We are using `poutine.broadcast` on the model here too, and `Trace_ELBO` also calls `get_importance_trace`, so I think that should be covered..

PR #1292, opened 2018-08-07
ignores for code coverage

jpchen:
increasing that code coverage. we're better than 89%

optimizers that are untested are used in examples. presumably theyd be covered if we ran coverage reports in test_examples. (verifying by running the CI, cant do it locally...)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,3 @@
+ignore:
+  - "pyro/docutil.py"
+  - "pyro/logger.py"
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ jobs:
         - python: 2.7
           env: STAGE=examples
           script:
-              - pytest -vs --stage test_examples
+              - pytest -vs --cov=pyro --stage test_examples
               - grep -l smoke_test tutorial/source/*.ipynb | xargs grep -L 'smoke_test = False' \
                   | CI=1 xargs pytest -vx --nbval-lax --current-env
         - python: 3.5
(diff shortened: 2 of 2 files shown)


neerajprad:
> optimizers that are untested are used in examples

It will be nicer to have actual tests than relying on examples for coverage though. Just the examples and tutorials themselves would cover a lot of code but its without any assertions and isn't specific enough to help with fixes.

.codecov.yml
```diff
@@ -0,0 +1,3 @@
+ignore:
+  - "pyro/docutil.py"
```
neerajprad: Nice.

jpchen:
> It will be nicer to have actual tests than relying on examples for coverage though.

yeah i definitely agree with that, i just dont foresee that happening anytime soon. in any case, this is just an experiment to see how codecov works (so i can compare reports). dont need to necessarily merge this

neerajprad:
> yeah i definitely agree with that, i just dont foresee that happening anytime soon.

Good to see the results. I am fine merging this too. I think a big part of the value is in seeing marginal coverage increase/decrease on new PRs, and making sure that we are not merging PRs with large chunks of untested code paths. The overall numbers that we advertise is less of a concern, since its good either ways. 😄

neerajprad:
Apart from `clipped_adam` and `adagrad_rmsprop` in `optim`, the other two files whose coverage improves are example `utils` (not surprising or important since they are meant to be consumed by examples), and `gpr.py` in the GP module.

jpchen:
yeah other than that the largely untested modules are the omt distributions. i thought of `ignore`ing utils but those contain functions that should be tested.  and i agree, i think this at least keeps us honest with respect to testing, though id prefer not having sub-90 in our readme badge.

PR #1293, opened 2018-08-07
add tests for ClippedAdam

alicanb:


martinjankowiak:
this is all fine and good, but if engineering effort is to be spent in this direction it probably makes more sense do either of:
-- get rid of clipped adam (i.e. figure out how to support clipping within svi in general)
-- move any optimizers that have tests upstream

neerajprad:
Thanks for adding these tests! 

@jpchen - Is the coverage info wrong? Seems like it is including the files that you ignored earlier for some reason.

jpchen:
most likely because it is not using the latest `dev`. @alicanb can you merge from dev and push?

jpchen:
good to merge?

PR #1294, opened 2018-08-07
Update to PyTorch >=0.4.1

fritzo:
**EDIT:** This branch is only being used to test PyTorch JIT compatibility with Pyro. Please create separate PRs for any non JIT related updates.

-----------------------------------

Addresses #1267 
Formerly #1269 (closing due to a poorly chosen branch name of that PR)

This additionally includes JIT changes after 0.4.1 release.

This PR works around known bugs in PyTorch 0.4.1, mostly via pyro/distributions/torch_patch.py

## Remaining tasks

- [ ] Fix flaky examples/contrib/oed/ab_test.py (@ae-foster)
- [ ] Fix flaky baseball example (see [example failure](https://travis-ci.org/uber/pyro/jobs/414226990#L1082))


pyro/distributions/torch.py
```diff
@@ -39,6 +39,14 @@ def expand(self, batch_shape):
 
 
 class Categorical(torch.distributions.Categorical, TorchDistributionMixin):
+    def log_prob(self, value):
```
fritzo: This can be reverted after https://github.com/pytorch/pytorch/pull/10321

fritzo:
Yay, I think this is ready to merge as soon as tests pass. I've worked around the PyTorch issues.

neerajprad:
This is awesome, @fritzo! Taking a look.

pyro/distributions/torch_patch.py
```diff
@@ -45,4 +45,49 @@ def _torch_dirichlet_grad(x, concentration, total):
     return unpatched_fn(x, concentration, total)
 
 
+if torch.__version__.startswith('0.4.1'):
```
fritzo: ¡ssoɹƃ os sı sıɥ⊥

tests/contrib/tracking/test_em.py
```diff
@@ -63,23 +63,23 @@ def model(detections, args):
 
 
 # This should match detection_model's existence part.
-def exists_log_likelihood(objects, args):
+def compute_exists_logits(objects, args):
```
fritzo: The changes in this file work around https://github.com/pytorch/pytorch/issues/9977

pyro/distributions/util.py
```diff
@@ -172,8 +180,12 @@ def scale_tensor(tensor, scale):
 
     # avoid NANs
     if not isinstance(scale, numbers.Number):
+        if scale.dtype == torch.int64:
+            scale = scale.float()
```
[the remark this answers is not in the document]
fritzo: These lines are only needed when run under `torch.jit`. Within jitted code, `Number`s are often promoted to tensors, e.g. floats to `FloatTensor`s, ints to `LongTensor`s. Under this code, we saw some scales of 0 or 1 promoted to `LongTensor`. This wouldn't have been a problem, but additionally `LongTensor` has limited support for == comparison in PyTorch master.

I suppose an alternative would be to replace those occurrences of 0 and 1 with 0. and 1., but that still wouldn't deal with Python's builtin `sum()` operation which returns an int on empty input.

pyro/distributions/torch_patch.py
```diff
@@ -45,4 +45,49 @@ def _torch_dirichlet_grad(x, concentration, total):
...
+
+    # work around https://github.com/pytorch/pytorch/issues/9521
+    @_patch('torch.distributions.gamma._standard_gamma')  # noqa: F811
+    def _standard_gamma(concentration):
+        concentration = concentration.contiguous()
+        if concentration.is_cuda:
+            return concentration.cpu()._standard_gamma().cuda(concentration.get_device())
+        return concentration._standard_gamma()
+
+    # work around https://github.com/pytorch/pytorch/issues/9521
```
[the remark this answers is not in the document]
fritzo: We'll certainly add regression tests in PyTorch when these are fixed upstream.

pyro/distributions/util.py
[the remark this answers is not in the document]
neerajprad: Thanks for explaining, @fritzo.

tests/contrib/tracking/test_em.py
fritzo: FYI @alicanb you may need to make similar changes to get tracking models to work in 0.4.1 (i.e. avoid constructing a `torch.stack`ed or `torch.cat`'d tensor). Or you could simply avoid upgrading.

fritzo:
I'm still seeing NANs when running @ae-foster's example locally
```py
python examples/contrib/oed/ab_test.py --num-vi-steps=1000 --num-acquisitions=2
```
but it's unclear if they are due to lack of robustness or an actual bug.

- @ae-foster can you confirm that this PR does not break your experiments?
- @alicanb can you confirm that this PR does not break your experiments, or at least offers a simple workaround?

neerajprad:
> Fix flaky baseball example

Looking into this, but I am not able to reproduce this locally yet.

alicanb:
@fritzo do you have an estimate when this will be stable enough for me to pull and test tracking code?

ae-foster:
So... I found a bug in pytorch 0.4.1 and it's pretty bad.

Here are the steps to reproduce

```
>>> import torch
>>> torch.__version__
'0.4.1'
>>> from torch.distributions.utils import broadcast_all
>>> loc = torch.zeros(11, 1, 2)
>>> scale = torch.tensor([[10., 0.1]])
>>> scale.shape
torch.Size([1, 2])
>>> loc, scale = broadcast_all(loc, scale)
>>> scale.log()
tensor([[[   2.3026,   -2.3026]],

        [[ -45.5473,   22.8741]],

        [[     -inf,  -65.8490]],

        [[ -45.5490,       nan]],

        [[-103.2789,      -inf]],

        [[ -99.8132,      -inf]],

        [[ -73.1301, -103.2789]],

        [[-102.5858,      -inf]],

        [[-100.3345,      -inf]],

        [[ -74.1284,       nan]],

        [[ -89.5495,  -67.2353]]])
```
I'm guessing that broadcasting is weirdly broken so that those new log values are coming from some uninitialized memory (e.g. what you get from typing `double x; log(x);` in C)

This breaks `log_prob` for Normal distributions, hence the `nan` loss

fritzo:
> @alicanb ...when [will this] be stable enough for me to pull and test tracking code?

There is no hurry. This PR is already mostly stable; the only recent changes are for JIT support.

alicanb:
@ae-foster yeah that's known issue -> https://github.com/pytorch/pytorch/pull/10269

ae-foster:
Should I open a pytorch issue about this?

ae-foster:
Oh ok they know and fixed it

alicanb:
@ae-foster Well, no. They know alright, but fix is not there yet... ¯\_(ツ)_/¯

fritzo:
Let's just wait for the next release. PyTorch 0.4.1 is too broken for us to spend any more time on.

neerajprad:
> Let's just wait for the next release. PyTorch 0.4.1 is too broken for us to spend any more time on. We can continue to use this branch for jit research.

I think most of these changes are compatible with PyTorch 0.4.0 (except a few like `torch.logsumexp`). Is it possible to remove those few changes and merge the remaining ones to dev, keeping ourselves pinned to `torch0.4.0`? I think even the JIT tests will be skipped and anyways require PyTorch master, so we can merge those too. I can try taking a stab at absorbing these changes into a separate PR too.

neerajprad:
Hmm..well the tests pass finally.

fritzo:
> well the tests pass finally.

I'm worried that if `exp` and `log` are broken then so are `sqrt`, `pow`, `log1p` etc.

> I can try taking a stab at absorbing these changes into a separate PR too.

I like the idea of pulling out all non-breaking changes into separate PRs. e.g. `arange` changes. @neerajprad feel free to do so.

fritzo:
Closing in favor of #1376

PR #1296, opened 2018-08-07
Add xfailing jit tests to many anchor examples

fritzo:
Addresses #1063

This enables a `--jit` option in many of our examples. None of the examples currently work with `--jit`. The purpose of this PR is to help motivate and prioritize op support in upstream work on `torch.jit`.

This PR is already incorporated into the pytorch-0.4.1 branch, and I'd like to avoid that branch diverging too far from dev.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -7,7 +7,7 @@ from subprocess import check_call
 
 import pytest
 
-from tests.common import EXAMPLES_DIR, requires_cuda
+from tests.common import EXAMPLES_DIR, requires_cuda, xfail_param
 
 logger = logging.getLogger(__name__)
 pytestmark = pytest.mark.stage('test_examples')
@@ -39,6 +39,8 @@ CPU_EXAMPLES = [
     'vae/ss_vae_M2.py --num-epochs=1 --enum-discrete=sequential',
     'vae/vae.py --num-epochs=1',
     'vae/vae_comparison.py --num-epochs=1',
+    xfail_param('dmm/dmm.py --num-epochs=1 --jit', reason='not jittable'),
+    xfail_param('vae/vae.py --num-epochs=1 --jit', reason='not jittable'),
 ]
 
 CUDA_EXAMPLES = [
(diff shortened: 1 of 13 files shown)


tests/test_examples.py
```diff
@@ -90,3 +110,13 @@ def test_cuda(example):
     filename, args = example[0], example[1:]
     filename = os.path.join(EXAMPLES_DIR, filename)
     check_call([sys.executable, filename] + args)
+
+
```
neerajprad: Can we also add a `skipif` on `CI` for now? The examples stage takes quite a bit of travis time.

fritzo:
Thanks for reviewing!

tests/test_examples.py
```diff
@@ -90,3 +110,14 @@ def test_cuda(example):
...
+@pytest.mark.skipif('CI' in os.environ, reason='slow test')
+@pytest.mark.xfail(reason='not jittable')
+@pytest.mark.parametrize('example', JIT_EXAMPLES)
+def test_jit(example):
+    logger.info('Running:\npython examples/{}'.format(example))
+    example = example.split()
+    filename, args = example[0], example[1:]
+    filename = os.path.join(EXAMPLES_DIR, filename)
+    check_call([sys.executable, filename] + args)
```
jpchen: if `CUDA_EXAMPLES` and `JIT_EXAMPLES` are subsets of `CPU_EXAMPLES` what do you think of writing the cuda and jit tests as:
```
@pytest.mark.parametrize('example', CPU_EXAMPLES)
def test_jit(example):
  try:
    ...
    filename, args = example[0], example[1:] + ' --jit'
    ...
  except CalledProcessError as e:
    if 'unrecognized arguments' in e.output:
    # if the examples does not support jit, skip it
      continue
```
or would that run too many needless examples?

fritzo: That seems like too much automation. I'd worry that as we add more logic we'd accidentally disable all tests and silently stop testing (this has happened a couple times already). Therefore I'm more comfortable hand-unrolling the examples lists.

jpchen:
fine by me and fritz

PR #1297, opened 2018-08-08
add coverage config file

jpchen:
generating accurate coverage reports. all the ignores are:
- warnings
- errors (this is standard in pytest-cov config files)
- backward implementations (this is tested but the control flow goes to pytorch then back to pyro, which cant be checked by static analysis)
- main methods (we actually test this, though copied from the pytest-cov example config)

testing on CI

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,3 +1,8 @@
 ignore:
   - "pyro/docutil.py"
   - "pyro/logger.py"
+
+coverage:
+  range: 60..95
+  round: nearest
+  precision: 2
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,7 +1,7 @@
 [report]
 exclude_lines =
     pragma: no cover
-    def backward  # these are run via Pytorch
+    def backward
     raise AssertionError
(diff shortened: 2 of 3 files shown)


martinjankowiak:
"backward implementations (this is tested but the control flow goes to pytorch then back to pyro, which cant be checked by static analysis)"

wow that sounds an awful lot like something i said 🙄

.codecov.yml
```diff
@@ -1,3 +1,8 @@
 ignore:
   - "pyro/docutil.py"
   - "pyro/logger.py"
+
+coverage:
+  range: 60..95
```
neerajprad: What does this range represent?

jpchen: how green that badge is :D

neerajprad: Hahaha.. 👍

jpchen: i know what youre going to say but this is vital stuff :P

neerajprad: To be honest..that light green badge was irking me a bit too.

PR #1299, opened 2018-08-09
Use JIT traced potential energy computation in HMC

neerajprad:
Addresses #1063. 

This allows for tracing the Potential Energy computation inside HMC, and using this optimized trace for subsequent model executions in the integrator and for computing the energy outside the integrator. 

The output from this function is then differentiated to compute the gradient terms in the integrator. To JIT the integrator internals would require us to JIT the gradient computation using `@torch.jit.script` and wrapping up the integrator steps inside a `@torch.jit.trace`. However, `@torch.jit.script` has many more limitations which makes this infeasible. Just wrapping the PE computation offers a reduction of 10%-30% (more on certain examples) in the runtime.

Most tests pass on PyTorch master, except marked otherwise. Both the examples currently fail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,8 @@ test-cuda: lint FORCE
 test-jit: FORCE
 	@echo See jit.log
 	pytest -v -n auto --tb=short --runxfail tests/infer/test_jit.py tests/test_examples.py::test_jit | tee jit.log
+	pytest -v -n auto --tb=short --runxfail tests/infer/mcmc/test_hmc.py tests/infer/mcmc/test_nuts.py \
+		-k JIT=True | tee -a jit.log
 
 clean: FORCE
 	git clean -dfx -e pyro-egg.info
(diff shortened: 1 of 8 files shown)


neerajprad:
Thanks or reviewing, @fritzo!

1. Do you think it's worth adding tests/infer/mcmc/test_hmc.py and test_nuts.py to the test-jit command in the Makefile? (Note the two failing examples will already be run by test-jit). If the other tests are slow, it may be worth instead adding failing tests to tests/infer/test_jit.py.

Will add these to `test-jit` in the Makefile. Will also add the failing tests to `test_jit.py` so that we have a consolidated set of tests that need attention, that we can pass upstream.

> 2. Is it worth jitting all of ._energy() rather than merely ._potential_energy()? I'm not sure how expensive ._kinetic_energy() is, but it looks like a good candidate for op fusion.

Good point. `potential_energy` is called directly from inside the integrator and is therefore called many more times than `kinetic_energy` (which is just squaring the tensors and adding). Will try to jit `.energy` and update though, it can only be beneficial.

neerajprad:
Good to merge, pending any further comments.

 - Added to the `test-jit` command in the Makefile.
 - JITing the Kinetic Energy computation ended up making no reduction to the runtime (one reason could be that the python overhead is already minimal, and its called many fewer times - O(num_samples) rather than O(num_samples * num_steps)). As such, I have restricted to only JITing the Potential Energy computation for now. 
 - The savings on the HMM example are substantial - the longest example finishes in 2s rather than 8s without JIT!

neerajprad:
Yes, good to merge into the base pytorch-0.4.1 branch.

PR #1300, opened 2018-08-10
Update dev branch for compatibility with PyTorch release

neerajprad:
This is simply pulling in many important fixes from @fritzo in #1294 that can be merged into `dev` currently being pinned to PyTorch 0.4.1, without breaking CI.


fritzo:
@neerajprad I think we cannot revert to 0.4.0 in this branch because the available torch functions torch.distributions has changed, e.g. this branch uses upstream `torch.distributions.Binomial` and upstream `torch.logsumexp`. I think it's safer to selectively pull out files from this PR into separate PRs.

pyro/distributions/torch.py
```diff
@@ -39,6 +39,19 @@ def expand(self, batch_shape):
 
 
 class Categorical(torch.distributions.Categorical, TorchDistributionMixin):
+
+    # workaround for jit errors in broadcast_all.
+    # this can be deleted after https://github.com/pytorch/pytorch/pull/10321
+    if hasattr(torch, 'broadcast_tensors'):
```
fritzo: Would you mind either
1. cherry-picking 2a93a9a4e049 so that `OneHotCategorical` also gets a correct `log_prob`, or
2. reverting this change, since it's only available post-0.4.1?

pyro/infer/trace_elbo.py
```diff
@@ -160,7 +160,7 @@ def loss_and_grads(self, model, guide, *args, **kwargs):
             # build a closure for loss_and_surrogate_loss
             weakself = weakref.ref(self)
 
-            @pyro.ops.jit.compile(nderivs=1)
+            @pyro.ops.jit.trace
```
fritzo: I don't think we should merge the jit changes into 0.4.0. I guess nobody is using the jit, but it seems wrong to put 0.4.1 jit code in a repo pinned to 0.4.0. I think you can mostly revert individual files.

neerajprad: Makes sense. Let me revert that so that the other `0.4.1` branch only contains JIT changes and doesn't need to be merged anytime soon.

neerajprad:
Should be good to merge if tests pass.

neerajprad:
Thanks for reviewing, @fritzo.

PR #1302, opened 2018-08-10
Implement local num_particles for TraceEnum_ELBO

fritzo:
Addresses #915 

This leverages Pyro's parallel enumeration machinery to implement a local multi-particle estimator. This is mechanically similar to the local IWELBO estimator of Aitchenson (2018) ["Tensor Monte Carlo"](https://arxiv.org/abs/1806.08593), but does not do importance weighting.

The motivating use case (by @karalets) is a model `x -> y -> obs` with an expensive deterministic dependency between `x` and `y`. This PR allows drawing a single sample of `x`, performing a single expensive `x -> y` computation (e.g. a neural net), and then draw multiple samples of `y`.

The syntax is the same as parallel enumeration, with the addition of a `num_samples` option. This can be specified either through `@config_enumerate` or though per-site `infer` dict, e.g.
```py
def guide(data):
    x = pyro.sample("x", dist.Normal(0., 1.))
    y_loc, y_scale = expensive_function(x)
    y = pyro.sample("y", dist.Normal(y_loc, y_scale),
                    infer={"enumerate": "parallel", "num_samples": 10})
    pyro.sample("obs", dist.Normal(y, 1.), obs=data)
```

## Questions for reviewers

- [x] Does this look correct? It was surprisingly simple.
- [x] Can you think of other good tests?
- [x] Is this interface reasonable?

## Tasks

- [x] support enumeration of continuous sites
- [x] add simple test
- [x] add tests of nested structure
- [x] add tests mixing discrete + continuous distributions
- [x] add tests with nonreparametrized distributions
- [x] add warnings and checks
- [x] add tests with subsampling
- [x] add test of compatibility with `poutine.scale`
- [x] update docs


eb8680:
Neat!  Maybe we should go ahead and implement quadrature next if it's that easy.

> Does this look correct?

Seems right, but it doesn't admit the space savings of dependent enumeration so it may not be as widely applicable.

> Can you think of other good tests?

How about one that mixes enumeration and TMC?

fritzo:
@martinjankowiak I've added a few more tests and I now believe this is correct. I think I'm ready to merge, pending review and passing tests.

pyro/poutine/enumerate_messenger.py
```diff
@@ -32,9 +32,14 @@ def _pyro_sample(self, msg):
...
-            # Enumerate over the support of the distribution.
             dist = msg["fn"]
-            value = dist.enumerate_support(expand=msg["infer"].get("expand", EXPAND_DEFAULT))
+            num_samples = msg["infer"].get("num_samples")
+            if num_samples is None:
+                # Enumerate over the support of the distribution.
+                value = dist.enumerate_support(expand=msg["infer"].get("expand", EXPAND_DEFAULT))
+            else:
+                # Monte Carlo sample the distribution.
+                value = dist(sample_shape=(num_samples,))
```
neerajprad: Very cool! I think we can just deprecate the global `num_particles` logic now, and use `config_enumerate` instead.

neerajprad:
> Is this interface reasonable?

The interface seems reasonable. Just that we have a proliferation of ELBO classes with some redundancy in their application and ways of parallelizing over `num_particles` (At least for `TraceEnum_ELBO` it will be nice to just resort to `config_enumerate` on the backend if the user provides a global `num_samples` arg). 

Related, now  that `TraceEnum_ELBO` can be used for models with continuous sites, what do you and @eb8680  think about consolidating `Trace_ELBO` and `TraceEnum_ELBO` - is it doable / worth doing?

Looks great overall, these are all outside the scope of this PR.

pyro/poutine/enumerate_messenger.py
fritzo: I think it's still useful to have a global `num_particles` for some models. For example when there are multiple top-level variables, this PR does not give a way to sample them in a shared tensor dimension, as your parallel particles would:
```py
def model(data):
   x = pyro.sample('x', dist_x())
   y = pyro.sample('y', dist_y())  # does not depend on x
   z = pyro.sample('z', dist_z(x, y))
```

fritzo:
> @neerajprad ... consolidating `Trace_ELBO` and `TraceEnum_ELBO`

I'd like to keep around `Trace_ELBO` because it simple and auditable and serves as a good starting point to fork when creating a new type of ELBO, e.g. as was done with `RenyiElbo`. In general I don't mind maintaining multiple pieces of code when one is a strictly simpler version of another. E.g. it is nice to test `TraceEnum_ELBO` against the simpler `Trace_ELBO`. But @eb8680 what do you think?

martinjankowiak:
yes agree with @fritzo. `TraceEnum_ELBO` is complex; we don't want bugs
creeping into `Trace_ELBO` unexpectedly. also the latter is probably faster
in some cases.

2018-08-13 19:52 GMT-07:00 Fritz Obermeyer <notifications@github.com>:
>
> @neerajprad ... consolidating Trace_ELBO and TraceEnum_ELBO
>
> I'd like to keep around Trace_ELBO because it simple and auditable and
serves as a good starting point to fork when creating a new type of ELBO,
e.g. as was done with RenyiElbo. In general I don't mind maintaining
multiple pieces of code when one is a strictly simpler version of another.
E.g. it is nice to test TraceEnum_ELBO against the simpler Trace_ELBO. But
@eb8680 what do you think?

pyro/poutine/enumerate_messenger.py
neerajprad: Ahh..right. That’s an interesting distinction.

eb8680:
> I'd like to keep around Trace_ELBO because it simple and auditable and serves as a good starting point to fork when creating a new type of ELBO, e.g. as was done with RenyiElbo

I'd really like to see a lot of the complex logic in `TraceEnum_ELBO` move into the `Trace` data structure - that way `Trace` would have more of a reason to exist (as a data structure for efficient broadcasted log-density computation) and some of our `ELBO` implementations would become a little more declarative.

Issue #1235, opened 2018-07-16
Add mechanism for dynamically changing implementations of handlers


eb8680:
The side effects in `pyro.poutine` are broken up into interfaces in `poutine.handlers` and their implementations as `Messenger`s, so that it's easy to change the source code of the implementation of any particular Pyro component without affecting downstream code.  

However, there are cases where we'd like to modify the implementation of a `handler` on the fly.  For example, #1159 introduced alternate behavior for handling duplicate names in `TraceMessenger`, but we can't easily toggle that behavior in existing ELBO implementations.  This issue proposes adding a mechanism, perhaps exposed as `poutine.<some_handler>.register(<implementation>)`.

eb8680:
Closing this as outdated

PR #1306, opened 2018-08-14
Add Handler for changing poutine effect implementations

eb8680:
Addresses #1235 

This PR adds a small wrapper class `Handler` to sit between `Messenger`s and inference code.  Specifically,
1. The interfaces in `pyro/poutine/handlers.py` are automatically generated by wrapping `Messenger` classes with `Handler` instead of each defined by hand
2. The documentation for all effects now lives with their implementations instead of being split across messengers and handlers
3. `Handler.set(...)` allows swapping implementation of effects without affecting existing code
4. `TraceMessenger` has been split into two `Messenger`s that behave differently when seeing a name twice.

Most of the changes in this PR come from moving docstrings from `handlers.py` to the various `Messenger`s and deleted code in `handlers.py`.

Wanted to post this for discussion before it went stale...


eb8680:
The immediate use cases I had in mind for this are better handled by the changes in #1446.  Will revisit in future if necessary.

PR #1307, opened 2018-08-14
Make iarange and irange into Messengers

eb8680:
Part of #1234 

Pair coded with @fritzo 

This exploratory PR refactors `pyro.iarange` and `pyro.irange` into a single `Messenger`, hopefully simplifying their internals while preserving their semantics.  I've attempted to retain the current behavior and interface of `iarange` and `irange` but in the future I'd like to consider merging the two constructs into a single user-facing `plate` construct.


fritzo:
I kind of like the idea of merging `@broadcast` into `pyro.plate` so that `plate = irange+(iarange+broadcast)`. @neerajprad does that seem reasonable to you? If so @eb8680 does it seem reasonable to do that in this PR? (I'm happy to help update docs in this PR)

neerajprad:
I think that's reasonable. We are already doing automatic broadcasting in SVI and HMC, so this should be a smooth transition.

eb8680:
> I kind of like the idea of merging `@broadcast` into pyro.plate so that plate = irange+(iarange+broadcast)

The reason we didn't do that before is that it seemed to me that a model with and without `iarange`s should have exactly the same behavior.  Do we think that reasoning is less important/no longer valid now after getting more feedback?

fritzo:
> a model with and without `iaranges` should have exactly the same behavior

Hmm, good point. I believe `@broadcast` already breaks the "exact same behavior" property, since `@broadcast` effects only `iarange`-wrapped code. In non-broadcasted `plate`s, the property should still hold, so the only difference will be that there is no way to turn off broadcasting (for debugging purposes). This seems like a reasonable design choice to me, especially given the complexity of explaining what `iarange` means with- and without- broadcasting. WDYT?

eb8680:
> @broadcast already breaks the "exact same behavior" property

Yeah, I meant models should have exactly the same behavior without `broadcast` - explicitly writing `@broadcast` is a reminder to the user that the model's semantics are changing.  I still agree with that in principle, but given the difficulty of explaining `iarange` and `broadcast` it's probably best to favor any changes making them easier to understand, as you say.

neerajprad:
Actually, there might be another consideration (soon) - many of our poutine wrapped models are not picklable, and would pose problems in a multiprocessing context (e.g. in #1443 for HMC). It will be nice to refactor to allow for this, and have some tests that verify this (I'll try to add in my branch).

eb8680:
> many of our poutine wrapped models are not picklable

Is this true even with `cloudpickle`?

neerajprad:
I think `dill` or `cloudpickle` cover a much broader range of what is picklable, but the multiprocessing library use `pickle` internally, so it isn't easy to swap it out. [pathos](https://github.com/uqfoundation/pathos) provides a wrapper around multiprocessing using `dill`, but we can't use it because we are using the torch wrapper instead (I don't think it will support shared tensors or NNs easily which `torch.multiprocessing` provides).

pyro/poutine/handlers.py
```diff
@@ -354,16 +354,36 @@ def mask(fn=None, mask=None):
     return msngr(fn) if fn is not None else msngr
 
 
-def indep(fn=None, name=None, size=None, dim=None):
+def indep(fn=None, name=None, size=None, dim=None, sites=None):
```
fritzo: Should we rename this `plate` and add `indep` as an alias with a `DeprecationWarning`?

pyro/poutine/plate_messenger.py
```diff
@@ -0,0 +1,14 @@
+from __future__ import absolute_import, division, print_function
+
+from .broadcast_messenger import BroadcastMessenger
+from .subsample_messenger import SubsampleMessenger
+
+
+class PlateMessenger(SubsampleMessenger):
+    """
+    Swiss army knife of broadcasting amazingness:
+    combines shape inference, independence annotation, and subsampling
```
jpchen: 😄

pyro/poutine/subsample_messenger.py
```diff
@@ -0,0 +1,122 @@
...
+            subsample_size = -1
+        elif subsample is None:
+            msg = {
+                "type": "sample",
+                "name": name,
+                "fn": _Subsample(size, subsample_size, use_cuda, device),
+                "is_observed": False,
+                "args": (),
+                "kwargs": {},
+                "value": None,
```
jpchen: i think you need a `"mask": None` here to be consistent with the msg struct.

pyro/poutine/indep_messenger.py
```diff
@@ -53,7 +71,46 @@ def next_context(self):
         """
         self.counter += 1
 
+    def __enter__(self):
+        if self._vectorized is not False:
+            self._vectorized = True
+
+        if self._vectorized is True:
```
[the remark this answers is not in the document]
eb8680: I'd rather leave it explicit like this, since we're using three values (`None`, `False`, `True`) in the `_vectorized` state machine rather than just two.

neerajprad:
Sounds good! Looking at the changes, I think we might not even need to do any refactoring to support multiprocessing, since the handler call happens later when the model runs. LGTM if our CUDA tests pass.

eb8680:
Hmm, @neerajprad I'm seeing a CUDA test failures in `test_hmc`:
```
tests/infer/mcmc/test_hmc.py F                                                                 [100%]

============================================== FAILURES ==============================================
_________________________________________ test_gamma_normal __________________________________________

    def test_gamma_normal():
        def model(data):
            rate = torch.tensor([1.0, 1.0])
            concentration = torch.tensor([1.0, 1.0])
            p_latent = pyro.sample('p_latent', dist.Gamma(rate, concentration))
            pyro.sample("obs", dist.Normal(3, p_latent), obs=data)
            return p_latent
    
        true_std = torch.tensor([0.5, 2])
        data = dist.Normal(3, true_std).sample(sample_shape=(torch.Size((2000,))))
        hmc_kernel = HMC(model, trajectory_length=1)
        mcmc_run = MCMC(hmc_kernel, num_samples=200, warmup_steps=100).run(data)
        posterior = EmpiricalMarginal(mcmc_run, sites='p_latent')
>       assert_equal(posterior.mean, true_std, prec=0.05)

tests/infer/mcmc/test_hmc.py:219: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/common.py:179: in assert_equal
    assert_tensors_equal(x, y, prec, msg)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

a = tensor([ 0.5015,  2.0550]), b = tensor([ 0.5000,  2.0000]), prec = 0.05, msg = ''

    def assert_tensors_equal(a, b, prec=1e-5, msg=''):
        assert a.size() == b.size(), msg
        if prec == 0:
            assert (a == b).all(), msg
        elif a.numel() > 0:
            b = b.type_as(a)
            b = b.cuda(device=a.get_device()) if a.is_cuda else b.cpu()
            # check that NaNs are in the same locations
            nan_mask = a != a
            assert torch.equal(nan_mask, b != b), msg
            diff = a - b
            diff[a == b] = 0  # handle inf
            diff[nan_mask] = 0
            if diff.is_signed():
                diff = diff.abs()
            max_err = diff.max().item()
>           assert max_err < prec, msg
E           AssertionError

tests/common.py:140: AssertionError
---------------------------------------- Captured stderr call ----------------------------------------
Sample: 100%|██████████| 300/300 [00:34<00:00,  8.77it/s, Step size=0.537, Acceptance rate=0.913]
```

pyro/poutine/indep_messenger.py
```diff
@@ -108,7 +109,7 @@ def _reset(self):
     @property
     def indices(self):
         if self._indices is None:
-            self._indices = torch.arange(self.size, dtype=torch.long)
+            self._indices = torch.arange(self.size, dtype=torch.long).to(self.device)
```
fritzo: just curious, why `.to(self.device)` rather than `dtype=torch.long, device=self.device`? Is `arange` not implemented for cuda in some versions of PyTorch?

eb8680: I just copied this from `_Subsample`, no deeper reason

eb8680:
> Are CUDA failures limited to HMC?

Yep, good to merge

neerajprad:
> Are you able to do SVI stuffs using multiprocessing now?

I am not sure if `multiprocessing` will be useful in SVI, at least not in a naive way. What are you trying to do? What's the pickling error you are getting?

fehiepsi:
Ah, for each experiment, I generate a randomized dataset, then doing SVI on that dataset to get inference of a parameter. I repeatly make 1000 such experiment in a loop. Because each experiment is indepedent of the others, it is good to utilize *multiprocessing* for my purpose (collecting results from 1000 different experiments). However, when I use futures.concurrent.ProcessPoolExecutor, it said some lambda function is not pickable. Because I dont use lambda in my function, so I thought this come from poutine stuffs. Never mind, I just remember that I used a wrapper of SVI so I'll check again if there is any lambda expression in my wrapper of SVI first.

neerajprad:
> However, when I use futures.concurrent.ProcessPoolExecutor, it said some lambda function is not pickable. Because I dont use lambda in my function, so I thought this come from poutine stuffs. 

I see- I have mostly seen this in poutine wrapped models, e.g. in `poutine.enum` or `poutine.broadcast`, which are not picklable and raise an error.

fehiepsi:
Oops, sorry, I got a lambda function in my implementation. It works now.

PR #1308, opened 2018-08-14
Fix AutoRegNN when default tensor type is CUDA

neerajprad:
Minor fixes to `AutoRegNN` so that it works when the default tensor type is CUDA. The problem comes because certain functions like `randperm` and `linspace` do not have CUDA implementations. For the most part, this won't matter since we explicitly lift the module to CUDA using `module.cuda()` but these fixes are needed to make sure that this does not break our CUDA tests.


neerajprad:
> looks pretty gross

I know. :( Can't figure out a better way though.

PR #1311, opened 2018-08-16
Decoupling AutoRegressiveNN from InverseAutoregressiveFlow

stefanwebb:
I've extracted the MADE from IAF, so the IAF can take an arbitrary network as its instantiation parameter. So in the future you can mix-and-match with other components, like WaveNet and PixelRNN (other types of autoregressive neworks).

I've also change the interface to AutoRegressiveNN so it can have hidden layers with differing numbers of hidden units, and so that it returns a tuple of parameters, to make it easier to integrate with various normalizing flows.


pyro/nn/auto_reg_nn.py
```diff
@@ -121,22 +123,34 @@ class AutoRegressiveNN(nn.Module):
         autoregressive factorization. in particular for the identity permutation the autoregressive structure
         is such that the Jacobian is upper triangular. By default this is chosen at random.
     :type permutation: torch.LongTensor
+    :param skip_connections: Whether to add skip connections from the input to the output.
+    :type skip_connections: bool
+    :param nonlinearity: The nonlinearity to use in the feedforward network.
```
martinjankowiak: give an exmaple here? `nn....`?

pyro/nn/auto_reg_nn.py
```diff
@@ -110,8 +108,12 @@ class AutoRegressiveNN(nn.Module):
...
     :type input_dim: int
-    :param hidden_dim: the dimensionality of the hidden units
-    :type hidden_dim: int
+    :param hidden_dims: the dimensionality of the hidden units per layer
+    :type hidden_dims: list[int]
+    :param count_params: number of parameters to return from a forward call of the network.
+    :type count_params: int
+    :param param_dim: the dimensionality of each parameter
+    :type param_dim: int
     :param output_dim_multiplier: the dimensionality of the output is given by input_dim x output_dim_multiplier.
```
martinjankowiak: isn't `output_dim_multiplier` a function of the other parameters?

pyro/nn/auto_reg_nn.py
```diff
@@ -110,8 +108,12 @@ class AutoRegressiveNN(nn.Module):
 
     :param input_dim: the dimensionality of the input
     :type input_dim: int
-    :param hidden_dim: the dimensionality of the hidden units
-    :type hidden_dim: int
+    :param hidden_dims: the dimensionality of the hidden units per layer
+    :type hidden_dims: list[int]
+    :param count_params: number of parameters to return from a forward call of the network.
+    :type count_params: int
+    :param param_dim: the dimensionality of each parameter
```
martinjankowiak: param_dim: the dimensionality of each parameter

should this be a list for more flexibility?

also this param description could be a bbit more informative?

martinjankowiak: e.g. is this clear from the docstring?

# Shape the output into self.count_params of dimension self.param_dim*self.input_dim

pyro/nn/auto_reg_nn.py
```diff
@@ -183,4 +198,17 @@ def forward(self, x):
         if self.skip_layer is not None:
             h = h + self.skip_layer(x)
 
-        return h
+        # Shape the output into self.count_params of dimension self.param_dim*self.input_dim
+        if self.count_params == 1:
+            return h
+        else:
+            params = [h[..., (i * self.param_dim * self.input_dim):((i + 1) * self.param_dim * self.input_dim)]
```
martinjankowiak: can you use `torch.unbind` or the like here?

pyro/distributions/iaf.py
```diff
@@ -18,7 +17,7 @@ class InverseAutoregressiveFlow(Transform):
     Example usage:
 
     >>> base_dist = dist.Normal(torch.zeros(10), torch.ones(10))
-    >>> iaf = InverseAutoregressiveFlow(10, 40)
+    >>> iaf = InverseAutoregressiveFlow(AutoRegressiveNN(10, [40]))
```
martinjankowiak: iaf needs updated docstrings. in particular it should be clear it requires an autoregressive_nn of a particular sort (e.g. count_params=2)

pyro/nn/auto_reg_nn.py
stefanwebb: I've put 3 examples now that should clear up what the output is of arn(x)

pyro/nn/auto_reg_nn.py
stefanwebb: Ah sorry, I hadn't removed some defunct parameters from the docstring

pyro/nn/auto_reg_nn.py
stefanwebb: I've put something along the lines of what you suggest

pyro/nn/auto_reg_nn.py
stefanwebb: It's seems quite complicated because there are several cases - see the latest commit. `torch.unbind` can be used for the case where all the parameters have the same dimension, but otherwise I think you need to extract the parameters with slices

pyro/distributions/iaf.py
stefanwebb: Done!

pyro/nn/auto_reg_nn.py
```diff
@@ -104,39 +102,66 @@ class AutoRegressiveNN(nn.Module):
...
-    MADE: Masked Autoencoder for Distribution Estimation [arXiv:1502.03509]
-    Mathieu Germain, Karol Gregor, Iain Murray, Hugo Larochelle
+    Example usage:
+    >>> x = torch.randn(100, 10)
+    >>> arn = AutoRegressiveNN(10, [50], param_dims=[1])
+    >>> p = arn(x)  # 1 parameters of size (100, 10)
+    >>> arn = AutoRegressiveNN(10, [50], param_dims=[1, 1])
+    >>> m, s = arn(x) # 2 parameters of size (100, 10)
+    >>> arn = AutoRegressiveNN(10, [50], param_dims=[1, 5, 3])
+    >>> a, b, c = arn(x) # 3 parameters of sizes, (100, 1, 10), (100, 5, 10), (100, 3, 10)
```
martinjankowiak: one might expect sizes (100, 10) (100, 50) (100, 30) ?

martinjankowiak: why this choice?

pyro/nn/auto_reg_nn.py
martinjankowiak: sorry i meant for the param `nonlinearity`

pyro/nn/auto_reg_nn.py
stefanwebb: I thought that you would typically reshape it like that anyway before you use it in another class. Also it avoids any confusion about whether the ordering is by (param_dim, input_dim) or (input_dim, param_dim). What do you think?

pyro/nn/auto_reg_nn.py
stefanwebb: Ah okay!

pyro/nn/auto_reg_nn.py
martinjankowiak: seems reasonable

pyro/nn/auto_reg_nn.py
```diff
@@ -104,39 +102,66 @@ class AutoRegressiveNN(nn.Module):
...
+
     """
 
-    def __init__(self, input_dim, hidden_dim, output_dim_multiplier=1, permutation=None, skip_connections=False,
-                 num_layers=1, nonlinearity=nn.ReLU()):
+    def __init__(
+            self,
+            input_dim,
+            hidden_dims,
+            param_dims=[1, 1],
```
martinjankowiak: this default `param_dims=[1, 1]` should be specified in the doc string

stefanwebb: Done!

PR #1313, opened 2018-08-17
Use opt_einsum for message passing in SVI

fritzo:
Adresses #915 
Pair coded with @eb8680 

This uses the [opt_einsum](https://github.com/dgasmith/opt_einsum) package to implement a cheap batched `sumproduct()` operation which is then used in `TraceEnum_ELBO` to perform message passing. This allows us to marginalize out discrete latent variables. E.g. for HMMs, this allows linear-time inference (per SVI step).

Currently this implementation does not handle `iarange` and falls back to naive inference for models with `iarange`.

## Tested

- [x] added tests for `pyro.ops._einsum` module
- [x] added new tests for `TraceEnum_ELBO`
- [x] enabled larger HMM test cases

### DIFF
--- a/pyro/infer/util.py
+++ b/pyro/infer/util.py
@@ -8,7 +8,7 @@ import torch
 from torch.distributions.utils import broadcast_all
 
 from pyro.distributions.util import is_identically_zero
-from pyro.ops.sumproduct import sumlogsumexp
+from pyro.ops.sumproduct import sumproduct
 from pyro.poutine.util import site_is_subsample
 
 _VALIDATION_ENABLED = False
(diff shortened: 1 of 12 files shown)


pyro/infer/util.py
```diff
@@ -115,6 +128,13 @@ def __repr__(self):
             '({}, ...)'.format(frames) for frames in self]))
 
 
+def deduplicate_by_shape(tensors, combine=lambda a, b: a + b):
+    grouped = defaultdict(list)
+    for tensor in tensors:
+        grouped[getattr(tensor, 'shape', None)].append(tensor)
+    return [reduce(combine, parts) for parts in grouped.values()]
```
neerajprad: You'll need to use the one from `six.moves` for compatibility with python 3.

pyro/ops/sumproduct.py
```diff
@@ -0,0 +1,108 @@
...
+
+def _product(factors):
+    result = 1.
+    for factor in factors:
+        result = result * factor
+    return result
+
+
+def zip_align_right(xs, ys):
+    return reversed(zip(reversed(xs), reversed(ys)))
```
neerajprad: `reversed(list(zip..` for python 3.

PR #1315, opened 2018-08-18
Add support for iarange in einsum-based Dice

fritzo:
Addresses #915 
Pair coded with @eb8680 

This fixes a number of bugs in the einsum backend for `Dice`, and switches to this backend for all models.

## Tested

- enabled einsum backend on existing unit tests (some of which failed before this PR)
- added unit tests for `sumproduct`

### DIFF
--- a/pyro/ops/sumproduct.py
+++ b/pyro/ops/sumproduct.py
@@ -1,19 +1,13 @@
 from __future__ import absolute_import, division, print_function
 
+import operator
 from numbers import Number
 
-import opt_einsum
 import torch
 
+import opt_einsum
 from pyro.distributions.util import broadcast_shape
 from pyro.ops.einsum.deferred import deferred_tensor
-
-
-def _product(factors):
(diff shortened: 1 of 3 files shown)


pyro/ops/sumproduct.py
```diff
@@ -27,14 +21,12 @@ def sumproduct(factors, target_shape=(), optimize=True, backend='torch'):
     for t in factors:
         (numbers if isinstance(t, Number) else tensors).append(t)
     if not tensors:
-        return _product(numbers)
+        return reduce(operator.mul, numbers, 1.)
     shape = broadcast_shape(*(t.shape for t in tensors))
     if numbers:
-        number_part = _product(numbers)
+        number_part = reduce(operator.mul, numbers, 1.)
```
neerajprad: import `reduce` from `six.moves`.

fritzo: done

PR #1316, opened 2018-08-18
Refactor einsum sharing to be eager, composable

fritzo:
Addresses #915 

This PR refactors einsum `shared_intermediates` in two ways:
1. Following @jcmgray's [suggestion](https://github.com/dgasmith/opt_einsum/issues/9#issuecomment-414023717), switch from deferred to eager evaluation. This obviates the need for `Deferred` subclasses, greatly simplifying the new `pyro.ops.einsum.shared` backend.
2. Following @eb8680's suggestion, make `shared_intermediates` an effect, and bake a little effect-handling logic into `contract()`.

As a result:
- `shared_intermediates` is now compatible with other backends like `torch_log`
- there is no longer a need to wrap inputs in `deferred_tensor`; this is handled inside `contract()`
- `sumproduct` can be shared without specifying the `shared` backend (there is no need for a user to ever specify the `shared` backend)


pyro/ops/einsum/shared.py
```diff
@@ -0,0 +1,147 @@
...
+    Note that intermediate computations will not be garbage collected until
+    1. this context exits, and
+    2. the yielded cache is garbage collected (if it was captured).
+
+    :param dict cache: an optional dict
+    :yields: a cache dict suitable for saving and reusing as an input
+    """
+    if cache is None:
+        cache = {}
+    _SHARING_STACK.append(cache)
```
neerajprad: Will this correctly handle multiple backends - say, if we use this as a context manager inside of which we are using two different backends?

fritzo: Good point, I'll key the cache by backend as well.

neerajprad:
Looks great! The `torch_log` backend works fine (I realized though that the speedup I was seeing earlier was because of the wrong handling of python scalars :( ).

PR #1322, opened 2018-08-18
Use einsum for trace log prob evaluation in HMC

neerajprad:
This uses the einsum backend for evaluating the log pdf of a model trace, and is gated behind the `experimental_use_einsum` flag. This allows for handling the HMM example with large number of steps. This was previously intractable and led to OOM issues. 

This is however slow for many simple models (https://github.com/pytorch/pytorch/issues/10661), and is therefore gated behind a flag. Once the perf issues are fixed upstream in `torch.einsum`, we can simply remove the `experimental_use_einsum` flag and have this be a completely backend change.


neerajprad:
@fritzo, @eb8680 - Since it might take a while for https://github.com/pytorch/pytorch/issues/10661 to get resolved, can we try getting this merged into dev? We can deprecate the experimental flag once all perf issues are resolved upstream.

pyro/infer/mcmc/nuts.py
```diff
@@ -56,6 +56,13 @@ class NUTS(HMC):
     :param int max_iarange_nesting: Optional bound on max number of nested
         :func:`pyro.iarange` contexts. This is required if model contains
         discrete sample sites that can be enumerated over in parallel.
+    :param bool use_einsum: Whether to use an einsum operation to evaluate
```
fritzo: nit: Delete this param (it has been renamed to `experimental_use_einsum`

pyro/infer/mcmc/hmc.py
```diff
@@ -50,6 +50,10 @@ class HMC(TraceKernel):
     :param int max_iarange_nesting: Optional bound on max number of nested
         :func:`pyro.iarange` contexts. This is required if model contains
         discrete sample sites that can be enumerated over in parallel.
+    :param bool experimental_use_einsum: Whether to use an einsum operation
+        to evaluat log pdf for the model trace. No-op unless the trace has
```
fritzo: nit: evaluat -> evaluate

PR #1323, opened 2018-08-18
Work around performance issues in torch.einsum

fritzo:
This can be removed after https://github.com/pytorch/pytorch/issues/10661 is fixed. The motivating example is tests/infer/mcmc/test_hmc.py::test_bernoulli_latent_model

# Tested

- added a performance test


neerajprad:
Thanks for the patch! Will remove once pytorch/pytorch#10661 is fixed.

PR #1324, opened 2018-08-19
Simplify einsum shared

fritzo:
Addresses #915 
Pair coded with @eb8680 

This further simplifies the `pyro.ops.einsum.shared` backend in preparation for moving this backend upstream to `opt_einsum` https://github.com/dgasmith/opt_einsum/pull/43.

## Tested

- existing tests (this is pure refactoring)
-  updated growth tests to measure finer-grained op counts

PR #1325, opened 2018-08-20
Optimize to omit guide sites that exactly cancel model sites

fritzo:
Addresses #915 
~~Blocked by #1324~~

This implements an optimization in `TraceEnum_ELBO` that omits `log(p/q)` terms when `p` is exactly `q`. This is a common use case where part of the model is used exactly in the guide. For example consider the Baum-Welch algorithm where the hidden transition part of an HMM appears identically in the model and guide, so only the emission terms need to be considered.
```py
def hidden(num_steps):
    transition_probs = pyro.param("transition_probs", ...)
    x = 0
    for t in range(num_steps):
        x = pyro.sample("x_{}".format(t), dist.Categorical(transition_probs[x]))
        yield x

def model(data):
    emission_probs = pyro.param("emission_probs", ...)
    xs = list(hidden(len(data)))
    for t in pyro.irange("data", len(data)):
        pyro.sample("y_{}".format(t), dist.Categorical(emission_probs[xs[t]]),
                    obs=data[t])

@config_enumerate(default="parallel", expand=False)
def guide(data):
    with poutine.infer_config(config_fn=lambda msg: {'exact': True}):
        return list(hidden(len(data)))  # <--- this exactly matches the model
```
To help `TraceEnum_ELBO` omit cancelling terms, I've configured the guide's sample site infer dict with `{'exact': True}`. This results in significant savings in the number of `sumproduct` calls and `einsum` operations:
```
tests/infer/test_enum.py::test_elbo_hmm_exact[2] naive_cost = 5, optim_cost = 4
tests/infer/test_enum.py::test_elbo_hmm_exact[3] naive_cost = 11, optim_cost = 8
tests/infer/test_enum.py::test_elbo_hmm_exact[4] naive_cost = 17, optim_cost = 12
tests/infer/test_enum.py::test_elbo_hmm_exact[5] naive_cost = 23, optim_cost = 16
tests/infer/test_enum.py::test_elbo_hmm_exact[10] naive_cost = 53, optim_cost = 36
tests/infer/test_enum.py::test_elbo_hmm_exact[20] naive_cost = 113, optim_cost = 76
tests/infer/test_enum.py::test_elbo_hmm_exact[30] naive_cost = 173, optim_cost = 116
```

# Questions for reviewers

- [ ] Is this even the right guide? What I really want is to be able to enumerate model sites, and currently `TraceEnum_ELBO` enumerates over guide sites, so I'm replicating part of the model in the guide.

## Tested

- [x] added the above HMM example to `test_enum.py`, checking correctness of loss, gradients, and reduction in cost.

### DIFF
--- a/pyro/infer/traceenum_elbo.py
+++ b/pyro/infer/traceenum_elbo.py
@@ -25,8 +25,9 @@ def _compute_dice_elbo(model_trace, guide_trace):
                 for name, site in trace.nodes.items()
                 if site["type"] == "sample"}
 
+    # Collect log(p(z,x)) and -log(q(z)) terms.
     costs = OrderedDict()
-    exactly_cancelled = set()
+    exactly_cancelled = set()  # terms satisfying p(z,x) == q(z)
     for name, site in guide_trace.nodes.items():
         if site["type"] == "sample":
             if site["infer"].get("exact"):
@@ -37,7 +38,11 @@ def _compute_dice_elbo(model_trace, guide_trace):
         if site["type"] == "sample" and name not in exactly_cancelled:
             costs.setdefault(ordering[name], []).append(site["log_prob"])
 
-    return Dice(guide_trace, ordering).compute_expectation(costs)
+    # Compute q(z) factors.
+    dice = Dice(guide_trace, ordering)
+
+    # Compute the expectation E z~q [ log(p(z,x)) - log(q(z)) ]
+    return dice.compute_expectation(costs)
 
 
 class TraceEnum_ELBO(ELBO):
--- a/pyro/infer/traceenum_elbo.py
+++ b/pyro/infer/traceenum_elbo.py
@@ -26,12 +26,16 @@ def _compute_dice_elbo(model_trace, guide_trace):
                 if site["type"] == "sample"}
 
     costs = OrderedDict()
-    for name, site in model_trace.nodes.items():
-        if site["type"] == "sample":
-            costs.setdefault(ordering[name], []).append(site["log_prob"])
+    exactly_cancelled = set()
     for name, site in guide_trace.nodes.items():
         if site["type"] == "sample":
-            costs.setdefault(ordering[name], []).append(-site["log_prob"])
+            if site["infer"].get("exact"):
+                exactly_cancelled.add(name)
+            else:
+                costs.setdefault(ordering[name], []).append(-site["log_prob"])
+    for name, site in model_trace.nodes.items():
+        if site["type"] == "sample" and name not in exactly_cancelled:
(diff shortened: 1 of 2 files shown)


pyro/infer/traceenum_elbo.py
```diff
@@ -26,12 +26,16 @@ def _compute_dice_elbo(model_trace, guide_trace):
...
     for name, site in guide_trace.nodes.items():
         if site["type"] == "sample":
-            costs.setdefault(ordering[name], []).append(-site["log_prob"])
+            if site["infer"].get("exact"):
+                exactly_cancelled.add(name)
+            else:
+                costs.setdefault(ordering[name], []).append(-site["log_prob"])
+    for name, site in model_trace.nodes.items():
+        if site["type"] == "sample" and name not in exactly_cancelled:
+            costs.setdefault(ordering[name], []).append(site["log_prob"])
```
jpchen: some inline comments here would be helpful. trace* elbo implementations have changed pretty dramatically with additional configurations that its hard to follow without backtracking through PRs

fritzo: Sure, I've added some comments.

fritzo:
> @neerajprad It would be nice if there was some way of annotating this in the model itself, so that we didn't have to abstract out the shared component and reuse in the guide

It is already possible to annotate the `guide` without factoring out the shared component (note it is the `guide` not the `model` that is annotated). The reason I factored out the shared code in the test was to emphasize that the shared code should be identical. Additionally @jpchen and @alicanb and I have found it helpful to factor out shared components so as to minimize code divergence over time.

fritzo:
@martinjankowiak I realize this is a bit kludgey, but I couldn't think of a cleaner alternative. Even equality comparisons won't work, since we need not only the log_prob values but also their gradients to exactly cancel. I'm happy to brainstorm alternatives offline.

neerajprad:
> The reason I factored out the shared code in the test was to emphasize that the shared code should be identical. Additionally @jpchen and @alicanb and I have found it helpful to factor out shared components so as to minimize code divergence over time.

This is definitely a much better design than repeating the same code in the model and guide. I was wondering if it was possible to annotate the sites in the model and not have to specify them in the guide (which will throw an error in `check_model_guide_match`), but thinking about it, this isn't possible in general, and may lead to more complications. Explicit is probably better.

fritzo:
> @neerajprad possible to annotate the sites in the model and not have to specify them in the guide?

Yes, after discussion with @martinjankowiak I'll try a different approach.

fritzo:
Closed in favor of https://github.com/uber/pyro/pull/1328

PR #1331, opened 2018-08-23
Model and EIG estimation for linear model with sigmoid link function

ae-foster:
Changes
 - add a sigmoid model (with slope governed by a random variable `k`)
 - add a BA guide and tests for BA convergence for estimating the EIG


ae-foster:
Checked- tests and examples working fine

PR #1333, opened 2018-08-24
Speed up pickling of dmm; use pickling in einsum

fritzo:
This PR does three things to speed up experiments:
1. Refactors the dmm example by moving unpickling logic into a `load_data()` function in `polyphonic_data_loader.py` so that it can be used by other examples.
2. Switches to `pickle.HIGHEST_PROTOCOL` which is about 1000x faster to load the dmm data (13seconds --> tens of milliseconds). This should be safe because we use these pickles as a local per-machine cache; we do not require portability because we don't move these pickles around.
3. Adds a `cached_paths` context manager that caches einsum paths via pickling. This is useful when running scripts that train models with large enumeration graphs.

## Tested

- [x] added a unit test
- [x] successfully used all of these tricks in the hmm-tutorial branch
- [x] verified that pickle is recreated when switching Python 2 --> 3 --> 2

### DIFF
--- a/examples/dmm/polyphonic_data_loader.py
+++ b/examples/dmm/polyphonic_data_loader.py
@@ -13,7 +13,7 @@ however, the original source of the data seems to be the Institut fuer Algorithm
 und Kognitive Systeme at Universitaet Karlsruhe.
 """
 
-from os.path import exists, join
+import os
 
 import numpy as np
 import six.moves.cPickle as pickle
@@ -24,9 +24,15 @@ from observations import jsb_chorales
 
 # this function processes the raw data; in particular it unsparsifies it
 def process_data(base_path, filename, T_max=160, min_note=21, note_range=88):
-    output = join(base_path, filename)
-    if exists(output):
-        return
+    output = os.path.join(base_path, filename)
+    if os.path.exists(output):
+        try:
+            with open(output, "rb") as f:
+                return pickle.load(f)
+        except (ValueError, UnicodeDecodeError):
+            # Assume python env has changed.
+            # Recreate pickle file in this env's format.
+            os.remove(output)
 
     print("processing raw polyphonic music data...")
     data = jsb_chorales(base_path)
--- a/examples/dmm/polyphonic_data_loader.py
+++ b/examples/dmm/polyphonic_data_loader.py
@@ -44,13 +44,20 @@ def process_data(base_path, filename, T_max=160, min_note=21, note_range=88):
                 slice_length = len(note_slice)
                 if slice_length > 0:
                     processed_dataset[split]['sequences'][seq, t, note_slice] = np.ones((slice_length))
-    pickle.dump(processed_dataset, open(output, "wb"))
+    pickle.dump(processed_dataset, open(output, "wb"), pickle.HIGHEST_PROTOCOL)
     print("dumped processed data to %s" % output)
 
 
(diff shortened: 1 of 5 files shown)


examples/dmm/polyphonic_data_loader.py
```diff
@@ -44,13 +44,20 @@ def process_data(base_path, filename, T_max=160, min_note=21, note_range=88):
                 slice_length = len(note_slice)
                 if slice_length > 0:
                     processed_dataset[split]['sequences'][seq, t, note_slice] = np.ones((slice_length))
-    pickle.dump(processed_dataset, open(output, "wb"))
+    pickle.dump(processed_dataset, open(output, "wb"), pickle.HIGHEST_PROTOCOL)
```
neerajprad: Will you be able to import this in your tutorial?

examples/dmm/polyphonic_data_loader.py
```diff
@@ -44,13 +44,20 @@ def process_data(base_path, filename, T_max=160, min_note=21, note_range=88):
...
 # this logic will be initiated upon import
 base_path = './data'
 process_data(base_path, "jsb_processed.pkl")
+jsb_file_loc = "./data/jsb_processed.pkl"
+
+
+# ingest training/validation/test data from disk
+def load_data():
+    with open(jsb_file_loc, "rb") as f:
+        return pickle.load(f)
```
neerajprad: Pickling / unpickling is incompatible between python 2 and 3 due to ambiguity in bytestring decoding ([stackoverflow issue](https://stackoverflow.com/questions/28218466/unpickling-a-python-2-object-with-python-3)). We can include the `encoding='latin-1'` for python 3 to ensure compatibility. Another thing would be to force remove the dataset and download again if you change your virtual env.

examples/dmm/polyphonic_data_loader.py
fritzo: I have switched from a tutorial to an example, so yes.

pyro/ops/einsum/__init__.py
```diff
@@ -35,4 +39,23 @@ def contract(equation, *operands, **kwargs):
...
+    """
+    Context manager to load and save cached paths. This is most useful outside
+    of the training loop, so that optimized einsum paths can be saved across
+    training runs. Note that this saves even after user interrupt with CTRL-C.
+
+    :param str filename: path to a pickle file where the cache will be stored
+    """
+    if os.path.exists(filename):
+        with open(filename, 'rb') as f:
+            _PATH_CACHE.update(pickle.load(f))
```
neerajprad: This is neat! I am guessing that this is particularly for the `einsum` computation and not the other tensor computations keyed in by `id`?

fritzo: That's right, this is purely for caching einsum paths (the optimized operation schedules, not actual tensor comptuations; no tensors will be cached)

examples/dmm/polyphonic_data_loader.py
fritzo: Good point. I'll simply remove the file if it is incompatible. `HIGHEST_PROTOCOL` is much faster, and that's the most common use case.

neerajprad: Sounds good!

fritzo:
Thanks for reviewing!

Issue #1272, opened 2018-07-27
DMM throwing nans in pytorch 1.0


jpchen:

eb8680:
@jpchen @neerajprad what's the status of this in PyTorch 1.0?  Can this be closed?

fehiepsi:
I just quickly go through bug tagged issue and confirm that this is still happen in PyTorch 1.0.

PR #1335, opened 2018-08-25
Fix enumeration inside model (wrt scale, mask)

fritzo:
~~Blocked by #1334~~ 
Blocking #1338

This fixes our handling of enumeration of model variables, based on separating `poutine.mask` from `poutine.scale`. There are two crucial fixes in this PR:
1. Move all log factors into a single `logsumproductexp()` operation so that the dependent marginal can be computed.
2. Apply subsampling `scale` outside of the `logsumproductexp()`, and apply `mask` inside. This was motivated by the need for both `scale` and `mask` in the [HMM example](https://github.com/uber/pyro/blob/hmm-tutorial/examples/hmm.py).

## Tested

- [x] added scale tests
- [x] added masking tests
- [x] verified that this PR reduces the HMM example's loss-per-observation from ~1e8 to <10


fritzo:
I'm happy to walk reviewers over the new `_compute_model_costs()` code.

pyro/infer/traceenum_elbo.py
```diff
@@ -22,36 +22,61 @@
...
-            logprobs_t = sum((logprobs_u for u, logprobs_u in enum_logprobs.items() if u <= t), [])
-            if not logprobs_t:
-                marginal_costs[t] = costs_t
+        for t, sites_t in cost_sites.items():
+            # TODO split log_factors into connected components wrt shared tensor dims.
+            log_factors = []
+            scales = set()
+            for site in sites_t:
+                if site["log_prob"].dim() <= -enum_boundary:
+                    # For site do not depend on an enumerated variable, procede as usual.
```
neerajprad: nit: `proceed`

pyro/infer/traceenum_elbo.py
```diff
@@ -22,36 +22,61 @@
...
-                target_shape = target_shape[enum_boundary:] if enum_boundary else ()
-                marginal_costs[t].append(logsumproductexp(logprobs_t + [cost_t], target_shape))
+            for u, sites_u in enum_sites.items():
+                # TODO refine this coarse dependency ordering using time and tensor shapes.
+                if u <= t:
+                    for site in sites_u:
+                        logprob = site["unscaled_log_prob"]
+                        log_factors.append(logprob)
+                        scales.add(site["scale"])
+            # This is only correct if all enumerated things share a common subsampling scale.
```
neerajprad: Thanks for adding these comments! It was almost like a walkthrough.

pyro/infer/traceenum_elbo.py
```diff
@@ -22,36 +22,61 @@
...
+            # Note that we use a cheap weak comparison by id rather than tensor value, because
```
neerajprad: Where exactly are we comparing tensors by `id`?

jpchen:
looks good, does this fix #1272 ?

fritzo:
> @martinjankowiak a test where you differentiate through scale?

Sure, I can add such a test.

pyro/infer/traceenum_elbo.py
fritzo: This is subtle. `scales` is a `set`. `set` objects compare elements by `.__hash__`-then-`is`-then-`.__eq__`. Since PyTorch tensors hash by id, we only ever compare by `.__hash__` and `is`. This all happens at `scales.add(site["scale"])`.

neerajprad: Ahh..I see. Thanks for explaining!

pyro/infer/traceenum_elbo.py
fritzo: Thank @jpchen who requested them on an earlier review :smile:

fritzo:
Should be ready to merge, pending final review.

PR #1338, opened 2018-08-26
Add an HMM example using message-passing enumeration

fritzo:
Blocked by #1334 #1335 

This adds a small tutorial demonstrating how to use message-passing enumeration of discrete variables in a model. The example progresses through four different models of increasing complexity, all using a one-line autoguide for Bayesian EM / Baum-Welch inference.

## Tested

- [x] added cheap smoke tests to tests/test_examples.py
- [x] verified that log likelihood is competitive with published models on this dataset


examples/hmm.py
```diff
@@ -0,0 +1,286 @@
+"""
+This example shows how to marginalize out discrete model variables in Pyro.
+
+This combines Stochastic Variational Inference (SVI) with an
+Expectation-Maximization (EM) algorithm, where we compute exact expectation of
+the ELBO wrt a subset of model variables.
```
martinjankowiak: well not quite, since the exact expectation is inside of the elbo

examples/hmm.py
```diff
@@ -0,0 +1,286 @@
...
+# narrow treewidth, therefore admitting efficient inference by message passing.
+# Pyro's TraceEnum_ELBO will find an efficient message passing scheme if one
+# exists.
+def model_1(sequences, lengths, args, batch_size=None, include_prior=True):
+    num_sequences, max_length, data_dim = sequences.shape
+    assert lengths.shape == (num_sequences,)
+    assert lengths.max() <= max_length
+    with poutine.mask(mask=torch.tensor(include_prior)):
+        probs_x = pyro.sample("probs_x",
+                              dist.Dirichlet(0.9 * torch.eye(args.hidden_dim) + 0.1)
```
martinjankowiak: comment on prior choices?

examples/hmm.py
```diff
@@ -0,0 +1,286 @@
...
+                                  .independent(1))
+        probs_y = pyro.sample("probs_y",
+                              dist.Beta(0.1, 0.9)
+                                  .expand([args.hidden_dim, 2, data_dim])
+                                  .independent(3))
+    tones_iarange = pyro.iarange("tones", data_dim, dim=-1)
+    with pyro.iarange("sequences", len(sequences), batch_size, dim=-2) as batch:
+        lengths = lengths[batch]
+        x, y = 0, 0
```
martinjankowiak: is it useful to do something like in the dmm? `x_prev =0 .... x_prev = x`? more lines of code but clarifies the structure. `x=pyro.sample(...x...)` might be confusing to the pyro beginner

fritzo:
Should be ready to merge after final review.

examples/hmm.py
```diff
@@ -0,0 +1,299 @@
...
+    sequences = torch.tensor(data['train']['sequences'], dtype=torch.float32)
+    lengths = torch.tensor(data['train']['sequence_lengths'], dtype=torch.long)
+    if args.truncate:
+        lengths.clamp_(max=args.truncate)
+    num_observations = float(lengths.sum())
+    pyro.set_rng_seed(0)
+    pyro.clear_param_store()
+    pyro.enable_validation(True)
+
+    # We'll train using Bayesian Baum-Welch, i.e. MAP estimation while
```
martinjankowiak: can we not call this 'Bayesian Baum-Welch' just call it MAP Baum-Welch?

PR #1342, opened 2018-08-28
Refactor OED code

ae-foster:
- cleanup BA and DV guides
- remove duplicated code
- merge example from presentation graphs branch


pyro/ops/matrix_utils.py
```diff
@@ -0,0 +1,102 @@
...
+
+import math
+
+import torch
+from torch.autograd import grad
+
+from pyro.util import warn_if_nan
+
+
+def rinverse(M, sym=False):
```
fritzo: Does this work for all matrices, or only symmetric positive definite matrices?

ae-foster: all matrices!

PR #1343, opened 2018-08-29
Fix model enumeration across iaranges

fritzo:
Addresses #915 

This attempts to fix our handling of model-side enumeration of discrete variables across multiple `iarange` contexts, e.g.
```py
x = pyro.sample("x", Bernoulli(0.4), infer={'enumerate': 'parallel'}).long()
pyro.sample("obs_outer", Normal(0., scales[x]), obs=outer_data)
with pyro.iarange("data", len(data)):
    y = pyro.sample("y", Bernoulli(0.2), infer={'enumerate': 'parallel'}).long()
    pyro.sample("obs_inner", Normal(locs[y], scales[x]), obs=inner_data)
```

This combines approach suggested by @eb8680 and @neerajprad: recursively contract dimensions:
- contract enumeration dimensions via `logsumproductexp()`
- contract `iarange` dimensions via `.sum()`

This PR adds an explicit check to verify that `iarange` nesting (of enumerated variables) has a tree structure. I believe it is generally impossible to efficiently apply message-passing inference to non-tree-shaped `iarange` nestings due to the presence of loops.

## Tested

- [x] added lots of tests of iaranges
- [x] added lots of tests of nested iaranges
- [x] added death tests to `test_valid_models.py`
- [x] confirmed that hmm example still runs in reasonable time


pyro/infer/traceenum_elbo.py
```diff
@@ -40,43 +41,58 @@ def _compute_model_costs(model_trace, guide_trace, ordering):
...
+                for log_factor in log_factors_t:
+                    scaled_log_factors.append(log_factor if denom == 1 else log_factor / denom)
+            marginal_cost = logsumproductexp(scaled_log_factors, target_shape)
+
+            # Contract iarange dimensions via product (in log space).
+            lower = frozenset.intersection(*log_factors.keys())
+            for frame in upper - lower:
+                marginal_cost = marginal_cost.sum(frame.dim, keepdim=True)
+            while marginal_cost.dim() and marginal_cost.shape[0] == 1:
+                marginal_cost.squeeze_(0)
```
jpchen: if i followed this correctly, youll have still singleton dims within `marginal_cost` (because of the previous `sum`)?

tests/infer/test_enum.py
```diff
@@ -1593,6 +1593,146 @@ def guide(data):
...
+
+@pytest.mark.parametrize('scale', [1, 10])
+def test_elbo_enumerate_ordinals_2(scale):
+    #  Guide    Model
+    #    a -----> b
+    #    |        |
+    #  +-|--------|----------------+
+    #  | V        V                |
+    #  | c -----> d -----> e   N=2 |
+    #  +---------------------------+
```
jpchen: your ascii diagram drawings are second to none

pyro/infer/traceenum_elbo.py
fritzo: Yes, this is typical when `iaranges` skip dimensions, e.g. when `vectorize_particles` adds an `iarange` to the left of all other `iarange`s, and say we're only enumerating over a global variable. For example in the following we may want to compare multiple models by sampling `which` model we believe is useful.
```py
which = pyro.sample("which", Categorical(...))
with iarange("data", len(data)):
    pyro.sample("obs", Normal(locs[which], scales[which]), obs=data)
```
If we run this with `num_particles=10, vectorize_particles=True`, then the final `marginal_cost` will have shape `(num_particles, 1)`.

fritzo:
Ok I believe this is correct now. I'm happy to walk any reviewers through the logic.

pyro/infer/traceenum_elbo.py
```diff
@@ -20,6 +19,88 @@
...
+            tu = t & u
+            if tu not in log_factors:
+                log_factors[tu] = []
+                pending.append(tu)
+
+    # Collect all enumeration dimensions.
+    enum_dims = defaultdict(set)
+    for t, terms in log_factors.items():
+        for term in terms:
+            for dim, size in enumerate(term.shape):
```
neerajprad: nit: Consider replacing the for loop with:

```python
    enum_dims[t] |= set((i for i in range(-term.dim(), enum_boundary)
                         if term.shape[i] > 1))
```

pyro/infer/traceenum_elbo.py
```diff
@@ -20,6 +19,88 @@
...
+        shape.reverse()
+        shape = tuple(shape)
+        term = logsumproductexp(terms, shape)
+        if not log_factors:
+            break
+
+        # Eliminate iarange dims via .sum() contractions.
+        frames = frozenset.intersection(*(leaf - t for t in log_factors if t < leaf))
+        assert frames
+        for frame in frames:
```
neerajprad: Is it worth sending these to `sumproduct`?

pyro/infer/traceenum_elbo.py
```diff
@@ -20,6 +19,88 @@
...
+            term = term.sum(frame.dim, keepdim=True)
+        parent = leaf - frozenset([frame])
+        log_factors[parent].append(term)
```
neerajprad: Is there a possibility that `log_factors[parent]` does not exist. e.g. when we have nested iarange with no log factors in the `outer` iarange:

```python
with iarange("outer", ..):
    with iarange("inner", ..):
        pyro.sample(..)
```

neerajprad: I may be missing something - I'm not sure if that will address this. Consider if `log_factors` consists of `frozenset()` and `frozenset(["a", "b"])`. That shouldn't change anything at the top. It seems like we are going to move up only one level in `parent = leaf - frozenset([frame])` and the remaining ordinal (either `frozenset(["a"])` or `frozenset(["b"])`) won't be in the list of `log_factors`.

fritzo: Oh you're right, that should be `parent = leaf - frames`. I'll add a test with a double jump that would have caught that.

fritzo:
Ok, I now believe this is correct (again).

pyro/infer/traceenum_elbo.py
```diff
@@ -20,6 +19,132 @@
...
+        leaf_terms = log_factors.pop(leaf)
+        leaf_dims = enum_dims.pop(leaf)
+        remaining_dims = set.union(set(), *enum_dims.values())
+        contract_dims = leaf_dims - remaining_dims
+        contract_frames = (frozenset.intersection(*(leaf - t for t in log_factors if t < leaf))
+                           if log_factors else frozenset())
+        parent = leaf - contract_frames
+        enum_dims[parent] |= leaf_dims & remaining_dims
+        log_factors.setdefault(parent, [])
+        for terms, dims in _partition_terms(leaf_terms, contract_dims):
```
fritzo: @neerajprad this differs from `TraceLogProbEvaluator` in that it now splits contraction terms into connected components wrt the shared-enum-dim graph. This is needed in examples where we need to delay the `sumproduct` (and its resulting broadcast) until the parent ordinal.

See the HMM example for motivation: the observations `y` are in an iarange that we need to contract *before* combining with the `x`s outside of the `iarange`. If we put all of the `y`s in a single `sumproduct`, then we would need to preserve all the dimensions for the next ordinal up. In this PR we instead contract the `y`s separately and move each of them to the parent ordinal.

tests/infer/test_enum.py
```diff
@@ -1514,29 +1524,34 @@ def hand_model(data):
...
 @pytest.mark.parametrize('scale', [1, 10])
 @pytest.mark.parametrize('num_samples,num_masked',
-                         [(1, 1), (3, 3), (3, 1)],
+                         [(1, 1), (2, 2), (3, 2)],
                          ids=["single", "batch", "masked"])
 def test_elbo_enumerate_iarange_3(num_samples, num_masked, scale):
+    #  +-----------------------+
+    #  | x ----> y ----> z     |
+    #  |                     N |
+    #  +-----------------------+
```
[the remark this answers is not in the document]
fritzo: I don't think of these tests as not so much a list of specific regression tests exercising edge cases. Rather I think of these as an exhaustive enumeration of models up to a given complexity bound.

I don't think it would be helpful to add text describing the history of the kind of bug each test originally caught.

fritzo:
Ok @eb8680 I've added a unit test for `_partition_terms()` and moved related functions to a new file `pyro/infer/contract.py`. I don't think verbose comments on tests will be helpful, but feel free to add text you'd like to remember.

tests/infer/test_enum.py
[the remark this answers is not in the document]
fritzo: Ideally we'd automatically generate these tests by enumerating ASTs up to a given complexity. In this PR I did that manually. The first set of tests is all iarange nestings of 3 variables and 1 iarange; the next set of tests should be all iarange nestings with two iaranges and at most 4 variables (with duplicates removed).

eb8680:
> I don't think verbose comments on tests will be helpful

Each enumeration PR has added hundreds of lines of test code to cover behavior we'd missed in the previous iterations.  Having a even a little bit of extra context for new tests makes it much easier to for me as a reviewer to evaluate algorithmic changes.

tests/infer/test_enum.py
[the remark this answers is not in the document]
eb8680: That would be neat, but it sounds like it could get complex enough that we'd need tests for those tests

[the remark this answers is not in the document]
fritzo: Ok I've added some comments.

PR #1344, opened 2018-08-29
Update GMM tutorial to discuss model enumeration

fritzo:
This updates the Gaussian Mixture Model tutorial to demonstrate both model-side enumeration and guide-side enumeration.

I've also switched to random initialization, demonstrating a nice pattern for using the param store this way. This addresses the many forum questions we've had, where the mixture converged to a trivial solution.


tutorial/source/gmm.ipynb
```diff
@@ -56,7 +60,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Maximum likelihood approach\n",
+    "## MAP estimation\n",
     "\n",
     "Let's start by optimizing model parameters `weights`, `locs`, and `scale`, rather than treating them as random variables with priors. Our model will learn global mixture weights, the location of each mixture component, and a shared scale that is common to both components. Our guide will learn soft assignment weights of each point.\n",
```
alicanb: Text seems to be outdated.

tutorial/source/gmm.ipynb
```diff
@@ -248,71 +250,189 @@
...
...7lyJUx1khDiFQyzZ7MDOF+lCuVbtVKdI2xQ76eLFi+7coV7584prhHCQulAy2g0cuTIEerWrZtwm16vp27dugQGBiZ6n8DAwBe2B2jQoMFL2+/atQt3d3dKlixJjx49ePDgwSs7YmNjiYyMfOEi0g9Ne7YosYHOnVvToEFx1Ukijfj7t8LJqRwQwaef+qnOEUIk5skTmDMHgJJDhiiOEbaqWufOVHZ2JhaYJ0sHCCuhdKB1//59TCYTHh4eL9zu4eFBcHBwovcJDg7+z+19fHxYtGgRAQEB/Pjjj+zevZuGDRtiMpkSfUw/Pz9cXV0TLvnz53/DZybS0pw5F/n77yicnECOGMhY7Oz0fP215aT6w4enERh4Q3GREOKfjvv5cT8sDAoVgiZNVOcIG6XT6+ndti0AMwMCiI+JUVwkhBUcOpgaPv30U5o0aUK5cuVo2rQpGzZs4NChQ+zatSvR7YcMGUJERETC5ebT48iF9TMaTfTt2xwoTsuW+8mXT3WRSGvDhjUkW7b3gFjatZMp/YWwJvTtvQAAIABJREFUJprZTLtx48gPbK1Xz3LMrxCp5JPx48ml03HTZGLtMJkUS6indKDl5uaGwWAgJCTkhdtDQkLw9PRM9D6enp7J2h6gSJEiuLm5cfny5US/7ujoiIuLywsXkT58+eWvxMaeQaeLYfTokqpzhAJ6vY5Jk8YBcOXKIlauPKm4SAjxzM5JkzgdG4sBqPbNN6pzhI1zcnWlW/XqVARybN5sObdACIWUDrQcHBzw9vYmICAg4Taz2UxAQADVq1dP9D7Vq1d/YXuAP//885XbA9y6dYsHDx6QO3fulAkXViEs7AkLFljesWrU6BsKF5YJTzKqjh2rkC/fJ4BGz56yV0sIazF5/HgAOpQpQ7ZChdTGiAxh2OLFHLG35/0zZ+Dvv1XniAxO+aGDvr6+zJ07l4ULF3Lu3Dl69OhBVFQUnTp1AqB9+/YMee7k2b59+7Jlyxb8/f05f/48I0eO5PDhw/Tq1QuAx48fM2jQIPbv38+1a9cICAjg448/plixYjRo0EDJcxSpo3376ZhMtzAY8rN4cS/VOUKx+fPHoNN1JTT0J3buVF0jhDi5ciXrQ0LQAb3HjVOdIzIIp8KF0XXpYrnyww9qY0SGp3yg1bp1ayZMmMDw4cPx8vLi+PHjbNmyJWHCixs3bnD37t2E7
```
neerajprad: You can use a progress bar here (if you prefer), by:

```python
from tqdm import trange
...
for i in trange(300 if not smoke_test else 2):
    ...
```
You can use a better looking progress bar by importing `tqdm` from `tqdm.autonotebook`, but that would also throw an experimental warning which you might want to filter.

neerajprad:
>  Does it seem reasonable to first train globals with model-side enumeration, then train local variables with guide-side enumeration?

Seem reasonable. Just for my understanding - does this have any advantages in terms of not getting stuck in a local optimum (compared to training all parameters using guide side enumeration as earlier)?

fritzo:
> does this have any advantages in terms of not getting stuck in a local optimum

No advantage. Model-side enumeration should provide lower-variance gradient estimates, but it does not affect the loss surface, so local optima remain. In practice the optimization problem in this tutorial is very susceptible to local optima, and I had to tweak the learning parameters and seed to get it to work. I'm open to other options e.g. automatic random restarts in a loop until a non-degenerate solution is found.

eb8680:
Re: the first two questions, I'd personally like to see a bunch of different configurations of model and guide side enumeration together in the same tutorial.

> other options e.g. automatic random restarts in a loop until a non-degenerate solution is found

What about using a higher-order optimizer, which would also show off our infinitely differentiable message-passing implementation?

fritzo:
Thanks for the detailed reviews! I've cleaned up the prose and added random best-of-100 initialization which seems to fix any robustness issues.

PR #1345, opened 2018-08-29
Add progress bar for MCMC runs

neerajprad:
This adds a progress bar for HMC / NUTS. This has become increasingly important for a few reasons:
 - Making sure that HMC is making progress (and checking for diagnostic information) without cluttering the console or notebooks with logging statements. 
 - Additionally, much of this has been designed so that with minor modifications, we can reuse this for parallel samplers initialized via multiprocessing, and get synchronized output.


neerajprad:
Example output:

```
  $ python examples/eight_schools/mcmc.py
Warmup:  45%|██████████████▎                 | 894/2000 [00:15<00:16, 68.22it/s, Step size=0.408, Acceptance rate=0.947]
```

pyro/infer/mcmc/logger.py
```diff
@@ -0,0 +1,111 @@
...
+from tqdm.autonotebook import tqdm
+
+LOG_MSG = "LOG"
+TQDM_MSG = "TQDM"
+
+
+class TqdmHandler(logging.StreamHandler):
+    """
+    Handler that synchronizes the log output with the
+    :class:`~tqdm.tqdm` progress bar.
```
neerajprad: We need to redirect all logging to `tqdm.write`, without which progress bar and logging output will be interspersed.

pyro/infer/mcmc/logger.py
```diff
@@ -0,0 +1,111 @@
...
+        except (KeyboardInterrupt, SystemExit):
+            raise
+        except Exception:
+            self.handleError(record)
+
+
+class MetadataFilter(logging.Filter):
+    """
+    Adds auxiliary information to log records, like `chain_id` and
+    `msg_type`.
```
neerajprad: Adding this auxiliary information (`chain_id`, `msg_type`) will be helpful in the multiprocessing case, where there will be a separate listener thread in the main process, that will listen for logging events and log them appropriately.

neerajprad:
Build failing due to https://github.com/travis-ci/travis-ci/issues/1337; will have a work-around for travis.

pyro/infer/mcmc/logger.py
```diff
@@ -0,0 +1,111 @@
...
+    to handle both progress bar updates and regular `logging`
+    messages.
+
+    :param log_handler: default log handler for logging
+        output.
+    :param progress_bar: If provided, diagnostic information
+        is updated using the bar.
+    """
+    def __init__(self, log_handler, progress_bar=None):
+        logging.Handler.__init__(self)
```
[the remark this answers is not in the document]
neerajprad: Not much; we can just replace the `log_handler` which is currently just a version of StreamHandler synchronized to tqdm to a queuehandler and handle all the logging events in a separate listener thread.

neerajprad:
> lgtm - this is pretty cool if it actually can synchronize across parallel sampling. so for my benefit in case i want to suppress output, the logger by default logs to stdout and the progbar is stderr is that correct?

Actually, the logger by default logs to `stderr` and so does the progress bar. This is because a lot of other tools rely on stdout for their functioning like `doctest`. So if you want to disable logs, you can set the handler lever to error, or change the handler to a NullHandler when you initialize the logger. We can also have a separate argument to MCMC that disables logging, if there's a need for it.

pyro/infer/mcmc/logger.py
```diff
@@ -0,0 +1,117 @@
...
+            msg = self.format(record)
+            self.flush()
+            tqdm.write(msg, file=sys.stderr)
+        except (KeyboardInterrupt, SystemExit):
+            raise
+        except Exception:
+            self.handleError(record)
+
+
+class MCMCLoggingHandler(logging.Handler):
```
fritzo: This seems more general than `MCMC`, e.g. I often use `sys.stdout.write('.');sys.stdout.flush()` to simulate progress bar for SVI in notebooks. Would it be reasonable to move this file to `pyro.infer.logger` and rename this class `InferenceLoggingHandler` or something, so we can use it for other inference algorithms? Or should I simply use `tqdm` directly in those notebooks?

fritzo: Ah, I see that although this class is general, its usage is specialized to MCMC below. Fine to leave it here.

pyro/infer/mcmc/logger.py
```diff
@@ -0,0 +1,117 @@
...
+    Handler that synchronizes the log output with the
+    :class:`~tqdm.tqdm` progress bar.
+    """
+    def emit(self, record):
+        try:
+            msg = self.format(record)
+            self.flush()
+            tqdm.write(msg, file=sys.stderr)
+        except (KeyboardInterrupt, SystemExit):
+            raise
```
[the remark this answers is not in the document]
neerajprad: You are right. Will change this.

pyro/infer/mcmc/logger.py
neerajprad: That's a good idea. Let me try to refactor this to make it more general.

PR #1347, opened 2018-08-30
add ZIP distribution

alicanb:
questions for reviewers:
- This definitely needs more tests, looking for suggestions
- I choose subclassing `TorchDistribution` directly. Other option was creating `ZeroInflatedDistribution` abstract class, for easily implementing ZINB as well, but it wasn't generic enough to provide enough code reuse imo. Do you agree?

### DIFF
--- a/pyro/distributions/zip.py
+++ b/pyro/distributions/zip.py
@@ -1,0 +1,57 @@
+from __future__ import absolute_import, division, print_function
+
+import torch
+from torch.distributions import constraints
+from torch.distributions.utils import broadcast_all, lazy_property
+
+from pyro.distributions import TorchDistribution
+
+
+class ZIP(TorchDistribution):
+    """
+    A Zero Inflated Poisson distribution.
+
+    :param torch.Tensor gate: probability of extra zeros.
+    :param torch.Tensor rate: rate of poisson distribution.
+    """
+    arg_constraints = {'gate': constraints.unit_interval, 'rate': constraints.positive}
+    support = constraints.nonnegative_integer
+
+    def __init__(self, gate, rate, validate_args=None):
+        self.gate, self.rate = broadcast_all(gate, rate)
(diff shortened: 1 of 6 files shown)


fritzo:
cc @OptimusLime

pyro/distributions/zip.py
```diff
@@ -0,0 +1,57 @@
+from __future__ import absolute_import, division, print_function
+
+import torch
+from torch.distributions import constraints
+from torch.distributions.utils import broadcast_all, lazy_property
+
+from pyro.distributions import TorchDistribution
+
+
+class ZIP(TorchDistribution):
```
fritzo: nit: I'm tempted to name this `ZeroInflatedPoisson` and let users `import ZeroInflatedPoisson as ZIP`

pyro/distributions/zip.py
```diff
@@ -0,0 +1,57 @@
...
+        batch_shape = self.gate.shape
+        event_shape = torch.Size()
+        super(ZIP, self).__init__(batch_shape, event_shape, validate_args)
+
+    def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
+        gate, rate, value = broadcast_all(self.gate, self.rate, value)
+        log_prob = (-gate).log1p() + (rate.log() * value) - rate - (value + 1).lgamma()
+        log_prob[value == 0] = (gate[value == 0] + (1 - gate[value == 0]) * torch.exp(-rate[value == 0])).log()
```
fritzo: Let's pull out the common subexpression
```
zeros = (value == 0)
```

pyro/distributions/zip.py
jpchen: +1 for more descriptive names, it's also consistent with the rest of the distribution names

alicanb:
Yep, good to go!

PR #1351, opened 2018-09-01
bugfix for AIR example

giorgiop:
The option `no-baseline` sets a variable `bl_value = None` which is later squeezed [here](https://github.com/uber/pyro/blob/dev/examples/air/air.py#L236).
This is now fixed. I also include a breaking test that now passes.

### DIFF
--- a/examples/air/air.py
+++ b/examples/air/air.py
@@ -233,7 +233,7 @@ class AIR(nn.Module):
         # Sample presence.
         z_pres = pyro.sample('z_pres_{}'.format(t),
                              dist.Bernoulli(z_pres_p * prev.z_pres).independent(1),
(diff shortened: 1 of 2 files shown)


examples/air/air.py
```diff
@@ -233,7 +233,7 @@ def guide_step(self, t, n, prev, inputs):
         # Sample presence.
         z_pres = pyro.sample('z_pres_{}'.format(t),
                              dist.Bernoulli(z_pres_p * prev.z_pres).independent(1),
-                             infer=dict(baseline=dict(baseline_value=bl_value.squeeze(-1))))
+                             infer=bl_value)
```
fritzo: Can you rename this variable from `bl_value` to `infer_dict` to clarify that it is a configuration dict rather than a tensor?

PR #1352, opened 2018-09-01
New IAF implementation with inverse operation

stefanwebb:
This pull request contains a new IAF implementation that uses a slightly different formula (also from the IAF paper). This matches the TF implementation and should improve performance I believe on the DMM example. The old version of IAF is moved to class `InverseAutoregressiveFlowStable`.

Explicit inverses are calculated for both `InverseAutoregressiveFlow` and `InverseAutoregressiveFlowStable` when required.

I've added in MaskedAutoregressiveFlow and NeuralAutoregressiveFlow. For these, the numerical gradient doesn't seem to match the analytical one, but I don't think it's due to an incorrect formula. I think it's caused by limitations in the method used to calculate the numerical gradient. I think that the gradient calculation for MAF could work if I use dx/dy rather than dy/dx (which has to be calculated by inverting the operation). I'm not sure about NAF...

I've separated out the tests for autoregressive flows from those for autoregressive networks.


martinjankowiak:
@stefanwebb thanks! can you please split this into at least two PRs? maybe peeling off the NeuralAutoregressive* bits? it's difficult to review 700 lines of code in one go

stefanwebb:
Let me have another go at submitting this PR and the FlipFlow one tomorrow. I'll break this one into three and remove the IAF stuff from the FlipFlow one so they're all independent

martinjankowiak:
great, thanks. maybe while you're at it rename flipflow to something like `PermutedDistribution`? doesn't something like that seem a better fit?

stefanwebb:
Yeah I agree

fritzo:
@stefanwebb what's the status of this PR? No hurry, I just want to make sure you're unblocked.

stefanwebb:
Nothing is blocking me at the moment, except an important conference deadline tomorrow. I look forward to getting back on this on Friday though

karalets:
Hey,

I see you closed this issue. However, I was quite fond of it. Any updates
on how this is going?

Cheers,

T.

stefanwebb:
Ah sorry, yes I'm breaking it up into several pull requests, which I'll do in sequence

stefanwebb:
About to do the first one...

stefanwebb:
P.s. I'm glad you liked it!

PR #1356, opened 2018-09-03
Detect invalid guide structure in model-guide enumeration

fritzo:
Addresses #915 
Pair coded with @eb8680 

This adds a check to handle the changing of iarange contexts during model-side enumeration.

## Tested

- [x] added a death test to ensure invalid model+guide structure is detected
- [x] added a test verifying that sequential guide enumeration + parallel model enumeration combine

### DIFF
--- a/pyro/infer/traceenum_elbo.py
+++ b/pyro/infer/traceenum_elbo.py
@@ -29,6 +29,18 @@ def _check_shared_scale(scales):
                          "but found {} different scales.".format(len(scales)))
 
 
+def _check_model_guide_enumeration_constraint(model_enum_sites, guide_trace):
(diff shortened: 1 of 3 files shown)


pyro/infer/traceenum_elbo.py
```diff
@@ -29,6 +29,18 @@ def _check_shared_scale(scales):
                          "but found {} different scales.".format(len(scales)))
 
 
+def _check_model_guide_enumeration_constraint(model_enum_sites, guide_trace):
```
eb8680: This check may be overly conservative - we should work out the minimal version of the criterion at some point.

Issue #1226, opened 2018-07-13
Reinstate and improve Search inference


eb8680:
`pyro.infer.Search` was removed in #1044 for performance and maintenance reasons, but it's being used informally in multiple places in Pyro and other projects, including in the RSA examples (#1198 ) and in `contrib.oed` (#1224 ).  We should collect the various copies into one place and add broadcasting-based parallel search as well, since most of the applications have static structure.

Tasks:
- [ ] Add dependent enumeration (#915) -based `Search` that creates a sequential `TracePosterior`
- [ ] Extend `Search` to support mixed parallel and sequential enumeration, as in `TraceEnum_ELBO`
- [ ] Move `HashingMarginal` from `examples/rsa/search_inference.py` to `pyro.infer` (proposed in #845 and finished in #1198)

fritzo:
@eb8680 it would be nice to implement message passing for `Search` next week, at the same time we implement message passing for SVI and HMC. Is there any preparation we can do this week to set us up to implement the harder parts next week?

eb8680:
One big blocker is actually not having a batched `EmpiricalMarginal`/`HashingMarginal`, but I think that would be a bit of a pain to implement as well since `Empirical` isn't vectorized.

neerajprad:
> One big blocker is actually not having a batched EmpiricalMarginal/HashingMarginal, but I think that would be a bit of a pain to implement as well since Empirical isn't vectorized.

What do you think we need to vectorize Empirical - can we generate bootstrap samples off of some initial samples to fill arbitrary batches on the left, or do you foresee needing something more sophisticated?

PR #1357, opened 2018-09-03
Implement contract_to_tensor() helper for querying marginals

fritzo:
Addresses #915 
Pair coded with @eb8680 

This implements a function `contract_to_tensor(tensor_tree, sum_dims, target_ordinal)` that performs message passing to compute requested marginals at any position in a tensor tree. Whereas `contract_tensor_tree()` contract down the tree to narrow `iarange` contexts for use in computing marginal log likelihood for `Dice` (following the `TraceLogProbEvaluator`'s downward reductions), this new `contract_to_tensor()` can propagate down and up and across the tree.

The motivation is to be able to query marginalized-out sample sites for their marginal distributions, e.g. for predicting document topics in LDA. Much work remains to refactor `_compute_mode_costs()` into a user-facing helper that uses this operation.

@martinjankowiak: Note this is like the batched einsum we discussed, but uses our internal unpacked tensor representation rather than einsum's packed representation of high-dimensional tensors. The packing-unpacking logic lives in [opt_sumproduct](https://github.com/uber/pyro/blob/239ed68/pyro/ops/sumproduct.py#L136), and would have been difficult to factor out.

## Tested

- [x] added unit tests with a hand-constructed example
- [x] this is a refactoring of `_contract_component()` which is covered by test_enum.py

### DIFF
--- a/pyro/infer/contract.py
+++ b/pyro/infer/contract.py
@@ -90,19 +90,26 @@ def _partition_terms(terms, dims):
         yield component_terms, component_dims
 
 
-def _contract_component(tensor_tree, sum_dims):
+def _contract_component(tensor_tree, sum_dims, target_ordinal=None):
     """
     Contract out ``sum_dims`` in a tree of tensors in-place, via message
-    passing. This reduces all tensors down to the minimum shared iarange
-    context. This function should be deterministic.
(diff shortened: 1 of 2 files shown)


eb8680:
Also a necessary prerequisite for #1226

PR #1359, opened 2018-09-04
Implement TraceEnum_ELBO.compute_marginals()

fritzo:
Addresses #915 

This exposes model-enumerated variables via a `TraceEnum_ELBO.compute_marginals()` method.

This PR aims to be the minimum viable interface to expose latent variables that have been marginalized out, e.g. the hidden states of an HMM or the document topics in LDA. This PR does not attempt to address more complex enumeration such as:
- guide enumeration (either sequential or parallel)
- num_particles

We should design the interface to be forwards-compatible with these additional features

## Questions for reviewers

- [x] Is this interface reasonable?
- [x] Should this live in the `TraceEnum_ELBO` class or elsewhere?

## Tested

- [x] tests with `Bernoulli`
- [x] tests with `Categorical`
- [x] tests with `OneHotCategorical`
- [x] death test for parallel guide enumeration (should raise `NotImplementedError`)
- [x] death test for sequential guide enumeration (should raise `NotImplementedError`)
- [x] death test for sequential `num_particles > 1` (should raise `NotImplementedError`)
- [x] death test for vectorized `num_particles > 1` (should raise `NotImplementedError`)
- [x] test reading latent state of an HMM


pyro/infer/contract.py
```diff
@@ -227,4 +220,15 @@ def contract_to_tensor(tensor_tree, sum_dims, target_ordinal):
     _contract_component(tensor_tree, sum_dims, target_ordinal)
     t, terms = tensor_tree.popitem()
     assert t == target_ordinal
-    return sum(terms)
+    term = sum(terms)
+
+    # Broadcast to any missing iaranges via .expand().
```
fritzo: I've moved broadcasting logic to happen outside of `sumproduct` so as to improve sharing.

pyro/infer/traceenum_elbo.py
```diff
@@ -261,6 +295,21 @@ def loss_and_grads(self, model, guide, *args, **kwargs):
         warn_if_nan(loss, "loss")
         return loss
 
+    def compute_marginals(self, model, guide, *args, **kwargs):
+        """
+        Computes marginal distributions at each model-enumerated sample site.
```
eb8680: Do we want to default to computing marginals for all enumeration sites, or only a single user-specified site at a time?  We could also compute and cache all marginals and only return one, since requesting them in a different order could reduce sharing.

pyro/infer/traceenum_elbo.py
```diff
@@ -261,6 +295,21 @@ def loss_and_grads(self, model, guide, *args, **kwargs):
...
+
+        :returns: a dict mapping site name to marginal ``Distribution`` object
+        :rtype: OrderedDict
+        """
+        result = None
+        for model_trace, guide_trace in self._get_traces(model, guide, *args, **kwargs):
```
eb8680: This will still do guide enumeration (as opposed to sampling from previously guide-enumerated variables)

pyro/infer/traceenum_elbo.py
fritzo: How about we pass in a list of requested sites, or say a `query_fn(msg)` as we do with `poutine.block`?

pyro/infer/traceenum_elbo.py
fritzo: Right. If we constrain this interface to be forwards-compatible with a future implementation that can enumerate guide sites, then this method will need to decide behavior based on `guide_trac.nodes[name]['infer']['enumerate']`. I believe the only forwards-compatible behavior is to raise a `NotImplementedError` if any guide site is set to be enumerated (I still need to write some of these error-checks in this PR).

If a user really wants guide-side enumeration during inference and guide sampling during prediction, then she can construct two different `TraceEnum_ELBO` classes; this will still work after we eventually implement querying for guide-enumeration-marginals.

pyro/infer/traceenum_elbo.py
eb8680: > list of requested sites

This seems reasonable for now, since it's what `EmpiricalMarginal` currently takes as a query as well.  The computational question of whether to compute all marginals or only one seems more immediate.

tests/infer/test_contract.py
```diff
@@ -1,14 +1,68 @@
...
+    if isinstance(x, (dict, OrderedDict)):
+        if len(x) != len(y):
+            return False
+        if any(key not in y for key in x):
+            return False
+        return all(deep_equal(x[key], y[key]) for key in x)
+    raise TypeError(type(x))
+
+
+def assert_immutable(fn):
```
fritzo: @eb8680 @neerajprad I though you'd like this :smile:

pyro/infer/traceenum_elbo.py
fritzo: Ok, it appears to be pretty clean and cheap to compute and return marginals for all sites; it should be only as expensive as O(1) SVI steps, since we're `sharing_intermediates()` across all sites. How about we implement the all-sites version in this PR, and if in a future application we find computational burden, we can add a `kwarg` like `sites` or `_sites`?

fritzo:
Ok @eb8680 this is ready for review again. Let me know if you want me to walk you through any changes over zoom.

pyro/ops/sumproduct.py
```diff
@@ -16,10 +16,31 @@ def zip_align_right(xs, ys):
...
+    """
+    if hasattr(tensor, '_pyro_memoized_squeeze'):
+        return tensor._pyro_memoized_squeeze
     result = tensor.squeeze()
-    tensor._memoized_squeeze = result
+    tensor._pyro_memoized_squeeze = result
+    return result
+
+
+def memoized_sum_keepdim(tensor, dim):
```
fritzo: These memoized functions are needed to improve sharing via `shared_intermediates`.

PR #1363, opened 2018-09-05
Add dict-like interface to the ParamStore

fritzo:
Addresses #1234 

This adds a dict-like interface to our `ParamStore`, while preserving all existing methods for backwards compatibility. I think a reasonable refactoring plan is to preserve the old inferface through our 0.2.x releases, and delete it in our 0.3 release.

The one deviation from dict behavior is in the `setdefault` method which:
1. takes an optional `constraint` argument, and
2. allows the value to be a callable for lazy evaluation (a la `defaultdict`)

Note that I've also fixed an arguable bug in param initialization whereby we were forgetting to `.detach()` unconstrained parameters (constrained parameters were correctly detached via `torch.no_grad`). This arguably sketchy functionality was relied on by `test_em_nested_inside_svi`, but the em-inside-svi pattern has been obsoleted by `MixedMultiOptimizer`, so I've merely deleted the old test.

## Tested

- [x] refactored to use the new dict-interface inside old-style methods; covered by existing tests

### DIFF
--- a/pyro/params/param_store.py
+++ b/pyro/params/param_store.py
@@ -30,6 +30,9 @@ class ParamStoreDict(object):
     - parameters can be saved and loaded from disk using `save` and `load`.
     """
 
+    # -------------------------------------------------------------------------------
+    # New dict-like interface
+
     def __init__(self):
         """
         initialize ParamStore data structures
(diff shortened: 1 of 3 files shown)


jpchen:
> test_em_nested_inside_svi, but the em-inside-svi pattern has been obsoleted by ?MixedMultiOptimizer

not for internal applications..?

fritzo:
>> @fritzo but the em-inside-svi pattern has been obsoleted

> @jpchen not for internal applications..?

In internal applications we're not passing EM params through the param store; the `test_em_nested_inside_svi` was merely an early attempt at a more Pyronic version of our internal applications, and eventually we'll refactor our internal code to use `MixedMultiOptimizer`

fritzo:
Ok I've reverted the `.detach()` change due to mysteriously failing tests, so this should now be pure refactoring. Ready to merge pending tests and reviewer approval.

pyro/params/param_store.py
```diff
@@ -46,18 +49,133 @@ def clear(self):
...
+        Set the constrained value of an existing parameter, or the value of a
+        new unconstrained parameter. To declare a new parameter with
+        constraint, use :meth:`setdefault`.
+        """
+        # store constraint, defaulting to unconstrained
+        constraint = self._constraints.setdefault(name, constraints.real)
+
+        # compute the unconstrained value
+        with torch.no_grad():
+            # FIXME should we .detach() the new_constrained_value?
```
fritzo: I tried to `.detach()` here but gave up due to failures in `test_em_nested_in_svi` and `tests/integration_tests/test_tracegraph_elbo.py::RaoBlackwellizationTests.test_iarange_in_elbo_with_superfluous_rvs`. I wasn't able to determine the reason behind the latter failure, so I'm avoiding the unnecessary change.

jpchen: bugs 1 
fritz 0

pyro/params/param_store.py
```diff
@@ -46,18 +49,133 @@ def clear(self):
...
+                init_constrained_value = init_constrained_value()
+
+            # set the initial value
+            self[name] = init_constrained_value
+
+        # get the param, which is guaranteed to exist
+        return self[name]
+
+    # -------------------------------------------------------------------------------
+    # Old non-dict interface
```
eb8680: Looking forward to getting rid of the rest of this!

PR #1361, opened 2018-09-04
Barber-Agakov, and OED refactoring

ae-foster:
New:
 - Barber-Agakov estimator of EIG
 - models and guides for sigmoid link function linear model

Refactored:
 - consistent API for different EIG estimators
 - improved code re-use in guides
 - extended support of EIG estimators: include integrating out random effects, supporting lists of variable names as well as single names

PR #1364, opened 2018-09-05
Better EIG testing

ae-foster:
Ports some code from `eig_estimation_benchmarking.py` to `test_eig.py` to be run on CI


ae-foster:
Await #1361

eb8680:
Moved to #1371

PR #1365, opened 2018-09-05
Add sequential sigmoid example

ae-foster:
This adds the example from the presentation using OED sequentially and plotting the final posterior entropy.


ae-foster:
Await #1361

PR #1369, opened 2018-09-07
Pin to pytest 3.7

fritzo:
This attempts to avoid CI errors in numpy-scipy version warnings.

Example failing builds:
https://travis-ci.org/uber/pyro/jobs/425494514
https://travis-ci.org/uber/pyro/jobs/425502606


fritzo:
> do we want to add a comment or just leave it pinned for the indefinite future

if you need a new feature some day, unpin pytest and hope they fixed the bug

PR #1370, opened 2018-09-07
Implement TreeCat model and inference

fritzo:
Addresses #1824 

This adds a pyro.contrib.treecat module implementing the [TreeCat](https://github.com/posterior/treecat) model of sparse heterogeneous tabular data.

TreeCat is interesting in Pyro both technically (as it exercises `TraceEnum_ELBO`'s message-passing inference for discrete latent variables), and for practical modeling, since its feature models can be arbitrary probabilistic programs. The model also makes cute use of `@poutine.broadcast` and `plate`, using four different `plate`s with a budget of only `max_iarange_nesting=1`. The `Feature` models are also neat in that the `.sample_group()` method is intended to be called with nonstandard interpretation inside a `plate` over mixture components.

This implementation is roughly based on the original [numba-based TreeCat](https://github.com/posterior/treecat) implementation. Differences include:
- this uses SVI for MAP inference of feature distributions (rather than collapsed Gibbs sampling)
- this supports pluggable hierarchical feature models a la the [posterior/distributions](https://github.com/posterior/distributions) used in [loom](https://github.com/posterior/loom)

## Tested

- [x] smoke test for inference
- [x] smoke test for prediction
- [x] unit tests for `Feature` objects
- [x] end-to-end example
- [ ] fix feature parameter mixing
- [ ] write paper comparing to competing algorithms


martinjankowiak:
mark september 6th 2018 as the day @fritzo's ulterior agenda surfaced... 😁

fritzo:
Moving to sandbox.

MaxwellRebo:
@fritzo what ever became of this?

fritzo:
@MaxwellRebo inference kind of worked but was very slow. @martinjankowiak and @jpchen and I tested imputation against some simple baselines and TreeCat was worse or no better at each baseline, and took >=100x longer to achieve those results. While this work did not lead to a releasable model, it did lead to a number of low-level improvements and speedups in Pyro machinery that have been released and are useful in other models.

If you are interested in CrossCat generalizations, I hear @fsaad has been thinking about generalizations based on sum product networks.

Issue #1375, opened 2018-09-11
Update to Pytorch 1.0


fritzo:
Pyro currently works with PyTorch 0.4.0 (0.4.1 support was abandoned #1267).

## Tasks

- [x] #1431 Update code
- [x] Switch networkx dependency to 0.2.2 (see #1191)
- [ ] Finish release notes
- [ ] Cut a 0.2.2 release

fritzo:
Closing in favor of #1570

PR #1376, opened 2018-09-11
Update to PyTorch 0.5.0

fritzo:
Addresses #1375 

## Tasks

- [ ] Update .travis.yml
- [ ] Remove obsolete patches for 0.4.1 release (@fritzo)
- [ ] Update jit wrappers (@neerajprad)
- [ ] Fix remaining tests
- [ ] Update docker files


pyro/distributions/torch_patch.py
```diff
@@ -218,15 +55,6 @@ def _einsum(equation, operands):
         y, x = operands
         return (x.unsqueeze(1) * y).sum(0).transpose(0, 1)
 
-    # work around torch.einsum's limitation to 26 letters
-    symbols = sorted(set(equation) - set(',->'))
```
neerajprad: We unfortunately still need this to avoid `RuntimeError: only lowercase letters a-z allowed as indices`.

fritzo: Hmm I thought I moved that logic upstream to `opt_einsum` [here](https://github.com/dgasmith/opt_einsum/blob/3d72bba/opt_einsum/parser.py#L88) and [here](https://github.com/dgasmith/opt_einsum/blob/3d72bba4087a56411b19ce30c1a68a70773113dd/opt_einsum/backends/torch.py#L41). Where are you seeing errors? Maybe it's missing from our `torch_log` backend.

neerajprad:
Migrated to #1431.

Issue #1358, opened 2018-09-04
JitTrace*ELBO does not work with irregular batch sizes


neerajprad:
With https://github.com/pytorch/pytorch/pull/10952, our VAE example becomes runnable, except that it fails on the last data batch in the epoch which has size `96` instead of the usual `128` (runs fine with regular batch sizes).

A minimal example that fails is below.

```python
import torch
from torch.distributions import constraints

import pyro
import pyro.distributions as dist
from pyro.infer import SVI, JitTrace_ELBO
from pyro.optim import Adam


def model(data):
    loc = pyro.param("loc", torch.tensor(0.0))
    scale = pyro.param("scale", torch.tensor(1.0), constraint=constraints.positive)
    with pyro.iarange("data", data.shape[0], dim=-2):
        pyro.sample("x",
                    dist.Normal(loc, scale).expand([data.shape[0]]),
                    obs=data)


def guide(data):
    pass


pyro.clear_param_store()
elbo = JitTrace_ELBO()
inference = SVI(model, guide, Adam({"lr": 1e-6}), elbo)
inference.step(torch.ones(10))
inference.step(torch.ones(3))  # fails; see error below
```

which results in the same error:

<details><Summary> Stack Trace </Summary>

```
clang: error: unsupported option '-fopenmp'
clang: error: unsupported option '-fopenmp'
warning: pytorch jit fuser failed to compile with openmp, trying without it...
Traceback (most recent call last):
  File "examples/expand.py", line 61, in <module>
    svi()
  File "examples/expand.py", line 58, in svi
    inference.step(torch.ones(3))
  File "/Users/npradhan/workspace/pyro_dev/pyro/pyro/infer/svi.py", line 96, in step
    loss = self.loss_and_grads(self.model, self.guide, *args, **kwargs)
  File "/Users/npradhan/workspace/pyro_dev/pyro/pyro/infer/trace_elbo.py", line 202, in loss_and_grads
    loss, surrogate_loss = self._loss_and_surrogate_loss(*args)
  File "/Users/npradhan/workspace/pyro_dev/pyro/pyro/ops/jit.py", line 59, in __call__
    ret = self.compiled[argc](*params_and_args)
  File "/Users/npradhan/miniconda2/envs/pytorch-master/lib/python3.6/site-packages/torch/nn/modules/module.py", line 477, in __call__
    result = self.forward(*input, **kwargs)
[...]

PR #1383, opened 2018-09-12
Override hash and eq for CondIndepStackFrame

neerajprad:
This overrides the `__eq__`, `__hash__` for `CondIndepStackFrame` so that the size arg can be either a PyTorch scalar or a python int type. 

**Use case:** Allows us to use data-structures that depend on `CondIndepStackFrame` in the JIT, e.g. in `TraceLogProbEvaluator` or `TraceEnum_ELBO`. 

Consider a model which has a data dependent iarange like `pyro.iarange("data", size=data.shape[0])`. When this is JITed, the `size` arg in the tuple will not be a python int but rather a PyTorch scalar. Since these are used to form keys that index into the different log prob terms per independence context, this either gives rise to an unrelated failure downstream, or a subtle bug (e.g. when the data structure keyed on these ordinals is using a `defaultdict`).

### DIFF
--- a/pyro/poutine/indep_messenger.py
+++ b/pyro/poutine/indep_messenger.py
@@ -2,6 +2,8 @@ from __future__ import absolute_import, division, print_function
 
 from collections import namedtuple
 
+import torch
+
 from .messenger import Messenger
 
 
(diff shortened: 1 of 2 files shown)


neerajprad:
This still doesn't fix #1358 (trying to look into that now), but enables `TraceLogProbEvaluator` to work as expected.

PR #1384, opened 2018-09-12
Allow for device kwarg in sumproduct

neerajprad:
Needed to run examples on CUDA - sometimes prob is a CPU tensor and cost is a CUDA tensor in `Dice.compute_expectation()` e.g. in `ss-vae`. This fix is wrt the current PyTorch release, and can be safely merged.

### DIFF
--- a/pyro/ops/sumproduct.py
+++ b/pyro/ops/sumproduct.py
@@ -44,7 +44,7 @@ def memoized_sum_keepdim(tensor, dim):
     return result
 
 
-def sumproduct(factors, target_shape=(), optimize=True):
+def sumproduct(factors, target_shape=(), optimize=True, device="cpu"):
     """
     Compute product of factors; then sum down extra dims and broadcast up
     missing dims so that result has shape ``target_shape``.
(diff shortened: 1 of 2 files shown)


pyro/ops/sumproduct.py
```diff
@@ -44,7 +44,7 @@ def memoized_sum_keepdim(tensor, dim):
     return result
 
 
-def sumproduct(factors, target_shape=(), optimize=True):
+def sumproduct(factors, target_shape=(), optimize=True, device="cpu"):
```
fritzo: What do you think about switching to default to `device=None` so that this works with `torch.set_default_tensor_type()` even when `device` is omitted?

neerajprad: Yeah, good point. Will change.

pyro/ops/sumproduct.py
```diff
@@ -90,10 +95,11 @@ def logsumproductexp(log_factors, target_shape=(), optimize=True):
     for t in log_factors:
         (numbers if isinstance(t, Number) else tensors).append(t)
     if not tensors:
-        return torch.tensor(float(sum(numbers))).expand(target_shape)
+        return torch.tensor(float(sum(numbers)),
+                            device=device).expand(target_shape)
     if numbers:
         number_part = sum(numbers)
-        tensor_part = logsumproductexp(tensors, target_shape)
+        tensor_part = logsumproductexp(tensors, target_shape, device=device)
```
fritzo: nit: there's no need to pass device here since all args are tensors.

Issue #1381, opened 2018-09-11
EKFState.update() fails on GPU 


neerajprad:
`tests/contrib/tracking/test_ekf.py::test_EKFState_with_NcpContinuous` fails on CUDA with

```
        K_prefix = self._cov.mm(H.transpose(-1, -2))
>       dx = K_prefix.mm(torch.gesv(dz, S)[0]).squeeze(1)  # K*dz
E       RuntimeError: invalid argument 2: b should be 2 dimensional at /opt/conda/conda-bld/pytorch_1524590031827/work/aten/src/THC/generic/THCTensorMathMagma.cu:67

pyro/contrib/tracking/extended_kalman_filter.py:190: RuntimeError
```

cc. @jpchen

fritzo:
Was this fixed by #1385?

PR #1385, opened 2018-09-12
fix shape error on cuda

jpchen:
addresses #1381

### DIFF
--- a/pyro/contrib/tracking/extended_kalman_filter.py
+++ b/pyro/contrib/tracking/extended_kalman_filter.py
@@ -187,7 +187,7 @@ class EKFState(object):
         S = H.mm(P).mm(H.transpose(-1, -2)) + R  # innovation cov
 
         K_prefix = self._cov.mm(H.transpose(-1, -2))
-        dx = K_prefix.mm(torch.gesv(dz, S)[0]).squeeze(1)  # K*dz
+        dx = K_prefix.mm(torch.gesv(dz.unsqueeze(1), S)[0]).squeeze(1)  # K*dz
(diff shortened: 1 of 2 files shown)


pyro/contrib/tracking/extended_kalman_filter.py
```diff
@@ -187,7 +187,7 @@ def update(self, measurement):
         S = H.mm(P).mm(H.transpose(-1, -2)) + R  # innovation cov
 
         K_prefix = self._cov.mm(H.transpose(-1, -2))
-        dx = K_prefix.mm(torch.gesv(dz, S)[0]).squeeze(1)  # K*dz
+        dx = K_prefix.mm(torch.gesv(dz.unsqueeze(1), S)[0]).squeeze(1)  # K*dz
```
neerajprad: Just for my understanding - could you comment on why this is failing on the GPU but works fine on CUDA?

tests/contrib/tracking/test_ekf.py
```diff
@@ -11,8 +11,6 @@
 from tests.common import assert_equal, assert_not_equal
```
neerajprad: You'll have to remove the pytest and os imports to make the linter happy.

pyro/contrib/tracking/extended_kalman_filter.py
jpchen: seems to be a pytorch bug, for some reason on cuda the `gesv` operation will properly handle `(N)` tensors but on cuda they need to be `(N, 1)` tensors.  they both return the same `(N, 1)` output

neerajprad:
Thanks for fixing this, @jpchen!

Issue #1374, opened 2018-09-10
Low count accuracy of AIR model


thematrixduo:
Hi,

I have been experimenting with the Attend Infer Repeat model. I downloaded the code from examples and run the code with the args provided in the tutorial (http://pyro.ai/examples/air.html), which is:

python main.py -n 200000 -blr 0.1 --z-pres-prior 0.01 --scale-prior-sd 0.2 --predict-net 200 --bl-predict-net 200 --decoder-output-use-sigmoid --decoder-output-bias -2 --seed 287710

But the highest count accuracy I get is around 76%. Could someone let me know if the args are not correct? I did not change any part of the code.

Also I am using pyro version 0.2.1 and pytorch version 0.4.0

Many thanks

null-a:
> Could someone let me know if the args are not correct?

The command line args in the tutorial are the ones I used, and it looks like you're using the same.

As well as answers to eb8680's questions, I'd be interested in hearing what other values for count accuracy you obtained, if that's possible. Thanks.

thematrixduo:
@eb8680 @null-a 
I let the inference run till the end (I believe it is specified by -n 200000). I am using the count accuracy function provided in 'main.py'. I just specified the args '--eval-every' to 1000. I have attached a plot of the count accuracy obtained.
![acc_plot](https://user-images.githubusercontent.com/8089252/45356239-a5806600-b5ba-11e8-9bb2-e2b819eaf611.jpg)

null-a:
@thematrixduo How many times have you run inference? What other values for final count accuracy have you obtained? Setting the random seed doesn't make this deterministic, so some variance is expected. If you're seeing consistently poor results, I'll try running it myself and see if I can spot anything odd going on. Thanks.

thematrixduo:
@null-a I have tried at least 5 times but for none of them did the count accuracy go above 80%. I have attached another plot. 
![acc_plot_2](https://user-images.githubusercontent.com/8089252/45372635-6a475c80-b5e5-11e8-8d74-7c4a8d3b9cf6.jpg)

null-a:
@thematrixduo It looks like performance may have degraded after c99ea673. If you apply the following patch (to v0.2.1 say) and then re-try I think you will see performance similar to that reported in the tutorial.

```diff
diff --git a/examples/air/main.py b/examples/air/main.py
index df89df2..09bdf25 100644
--- a/examples/air/main.py
+++ b/examples/air/main.py
[...]

PR #1388, opened 2018-09-14
Fix AIR example bugs.

null-a:
This PR includes the fix already mentioned in #1374 as well as a fix for what appears to be a second regression in the AIR example. I've run inference a few times and am now regularly seeing a count accuracy of ~90% after ~6K steps,  suggesting that the problem reported in #1374 is fixed. (That issue reported count accuracy never getting above 80%.)

### DIFF
--- a/examples/air/air.py
+++ b/examples/air/air.py
@@ -282,7 +282,7 @@ class AIR(nn.Module):
         if self.baseline_scalar is not None:
             bl_value = bl_value * self.baseline_scalar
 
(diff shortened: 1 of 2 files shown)


fritzo:
How expensive would it be to add a regression test for this?

eb8680:
> How expensive would it be to add a regression test for this?

Regression testing the AIR example seems infeasible, but we should add a small self-contained test that would have caught this.

PR #1394, opened 2018-09-18
Add Kalman Filter Tutorial

jpchen:
adding a basic tutorial on the (extended) kalman filter, mostly an example of how to use the library.  ended up removing all the (verbose) derivations, and just stuck to language semantics.

i used karl's [plotting code](https://github.com/Standard-Cognition/recursive-bayesian-filtering/blob/master/stats_tools.py#L372) to generate the nice plot.


jpchen:
(not really an anchor example but the `tutorial` label is gone so thats the next best :) )

tutorial/source/ekf.ipynb
```diff
@@ -0,0 +1,213 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Nonlinear Estimation: Extended Kalman Filter\n",
+    "\n",
+    "What if our system is non-linear, eg in GPS navigation?  Consider this non-linear system:\n",
+    "\n",
+    "$$ X_{k+1} = \\mathbf{f}(X_k) + \\mathbf{W}_k $$\n",
```
fritzo: For some reason, the `f` equation is not rendering. Does it render in your local copy?

jpchen: hm it loads locally, ill build the html to double check

jpchen:
thanks for the comment, as discussed ill rewrite the tutorial to focus on a differentiable ekf

jpchen:
this passes on python 3, i not sure what the problem is. going to remove the smoke test for now:
```
platform darwin -- Python 3.6.2, pytest-3.6.4, py-1.5.3, pluggy-0.6.0 -- /Users/jpchen/anaconda/envs/pyro3/bin/python
cachedir: .pytest_cache
rootdir: /Users/jpchen/Uber/pyro, inifile: setup.cfg
plugins: xdist-1.22.2, forked-0.2, cov-2.5.1, nbval-0.9.1
collected 6 items

tutorial/source/ekf::ipynb::Cell 0 PASSED                                                                                                                                                                                            [ 16%]
tutorial/source/ekf::ipynb::Cell 1 PASSED                                                                                                                                                                                            [ 33%]
tutorial/source/ekf::ipynb::Cell 2 PASSED                                                                                                                                                                                            [ 50%]
tutorial/source/ekf::ipynb::Cell 3 PASSED                                                                                                                                                                                            [ 66%]
tutorial/source/ekf::ipynb::Cell 4 PASSED                                                                                                                                                                                            [ 83%]
tutorial/source/ekf::ipynb::Cell 5 PASSED
```

eb8680:
Don't forget to add a link in `tutorials/source/index.rst`

tutorial/source/ekf.ipynb
```diff
@@ -0,0 +1,247 @@
...
+    "states= ekf_dist.get_states(zs)"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/html"
+   },
+   "source": [
+    "<center><figure>\n",
```
fritzo: Can you also embed the plotting code? I think it's cleaner to publish executed notebooks with cell contents so that users can fork the notebooks. If you really prefer this embedded-html idiom, then at least paste the plotting code into a non-executable code block so users can fork and re-plot.

pyro/contrib/tracking/dynamic_models.py
```diff
@@ -438,16 +442,17 @@ def process_noise_cov(self, dt=0.):
...
+                d = self._dimension
+                dt2 = dt*dt
+                dt3 = dt2*dt
+                dt4 = dt2*dt2
+                Q = self.sa2.new_zeros(d, d)
+                Q[:d//2, :d//2] = 0.25 * dt4 * eye_like(self.sa2, d//2)
+                Q[:d//2, d//2:] = 0.5 * dt3 * eye_like(self.sa2, d//2)
+                Q[d//2:, :d//2] = 0.5 * dt3 * eye_like(self.sa2, d//2)
+                Q[d//2:, d//2:] = dt2 * eye_like(self.sa2, d//2)
+                Q = Q * self.sa2
```
fritzo: Shouldn't this line be outside of `torch.no_grad`?

pyro/contrib/tracking/dynamic_models.py
```diff
@@ -364,22 +367,23 @@ def process_noise_cov(self, dt=0.):
...
+                d = self._dimension
+                dt2 = dt * dt
+                dt3 = dt2 * dt
+                Q = torch.zeros((d, d))
+                eye = eye_like(self.sa2, d//2)
+                Q[:d//2, :d//2] = dt3 * eye / 3.0
+                Q[:d//2, d//2:] = dt2 * eye / 2.0
+                Q[d//2:, :d//2] = dt2 * eye / 2.0
+                Q[d//2:, d//2:] = dt * eye
+                Q = Q * q
```
fritzo: Shouldn't this line be outside of `torch.no_grad()`?

pyro/contrib/tracking/distributions.py
```diff
@@ -37,14 +37,34 @@ def __init__(self, x0, P0, dynamic_model, measurement_cov, time_steps=1, dt=1.,
...
         batch_shape = x0.shape[:-1]
-        event_shape = (time_steps,) + x0.shape[-1:]
+        event_shape = (time_steps,) + (x0.shape[-1] // 2,)
         super(EKFDistribution, self).__init__(batch_shape, event_shape,
                                               validate_args=validate_args)
 
     def rsample(self, sample_shape=torch.Size()):
         raise NotImplementedError('TODO: implement forward filter backward sample')
 
+    def get_states(self, value):
```
fritzo: nit: I would rename this `filter_states()` or similar. `get_*()` suggests accessing an attribute, whereas you're actually performing computation here. Moreover you are only filtering forward, not filtering-forward-and-smoothing-backward, so it would be good to emphasize that you are returning filtered states rather than posterior states (only the final state will be a posterior marginal). In the future we could add a `filter_smooth_states()` that computes exact marginals at each time point.

tutorial/source/ekf.ipynb
```diff
@@ -0,0 +1,247 @@
...
+    "def model(data):\n",
+    "    # TODO replace this with HalfNormal\n",
+    "    R = pyro.sample('pv_cov', dist.HalfCauchy(2e-6)) * torch.eye(4)\n",
+    "    Q = pyro.sample('measurement_cov', dist.HalfCauchy(1e-6)) * torch.eye(2)\n",
+    "    # observe the measurements\n",
+    "    pyro.sample('track_{}'.format(i), EKFDistribution(xs_truth[0], R, ncv,\n",
+    "                                                      Q, time_steps=num_frames),\n",
+    "                obs=data)\n",
+    "    \n",
+    "guide = AutoDelta(model)"
```
fritzo: nit: add a line coment `# use a delta guide for MAP estimation of parameters`

tutorial/source/ekf.ipynb
```diff
@@ -0,0 +1,247 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def model(data):\n",
+    "    # TODO replace this with HalfNormal\n",
```
fritzo: nit: remove TODO? `HalfCauchy` seems fine to me. Alternatively `Chi2` would be appropriate.

tutorial/source/ekf.ipynb
```diff
@@ -0,0 +1,247 @@
...
+    "dzs = pyro.sample('dzs', dist.MultivariateNormal(mean, cov).expand((num_frames,)))\n",
+    "# compute measurement means\n",
+    "zs = xs_truth[:, :2] + dzs"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We'll use a [Delta autoguide](http://docs.pyro.ai/en/0.2.1-release/contrib.autoguide.html#autodelta) to learn the position and measurement covariances. The `EKFDistribution` computes the joint log density of all of the EKF states given a tensor of sequential measurements."
```
fritzo: nit: "to learn MAP estimates of the position and measurement covariances"

tutorial/source/ekf.ipynb
fritzo: If the plotting code is outside of Pyro, you could still link to the package and add `import` statements in a non-executable code block.

jpchen: the reason i didnt include the plotting code is not because i prefer embedding, it's because it was really verbose formatting the data properly to use standard cognition's plotting code.  if i wanted actual reusable plotting code, i would have to rewrite most of it since pyro distributions handle the cholesky computations under the hood, etc.  the AIR tutorial also doesnt have all of its plotting code available, even on github, but i can do it if you think it significantly detracts from the tutorial.

tutorial/source/ekf.ipynb
jpchen: halfcauchy seems too fat? i only used it because we dont have halfnormal in 0.4.0 but i had to shrink its width significantly. ill put a more aesthetic comment though

tutorial/source/ekf.ipynb
fritzo: Ok, that sounds like a reasonable compromise.

pyro/contrib/tracking/dynamic_models.py
```diff
@@ -364,21 +367,22 @@ def process_noise_cov(self, dt=0.):
...
-            Q[:d//2, d//2:] = dt2 * eye / 2.0
-            Q[d//2:, :d//2] = dt2 * eye / 2.0
-            Q[d//2:, d//2:] = dt * eye
-            Q *= q
+            with torch.no_grad():
+                # q: continuous-time process noise intensity with
+                # units length^2/time^3 (m^2/s^3). Choose ``q`` so that
+                # changes in velocity, over a sampling period ``dt``,
+                # are roughly ``sqrt(q*dt)``.
+                q = self.sa2 * dt
```
fritzo: This should be outside of `torch.no_grad()`. Please re-run results and verify convergence after fixing grad errors. Also, doesn't this introduce an extra factor of `dt`?

jpchen:
thanks for the detailed review and catching these subtle grad bugs!

Issue #1393, opened 2018-09-18
Clean up _Subsample, _subsample


fritzo:
PyTorch is much easier to use than it was when we first wrote `_Subsample` and friends. Let's
- [x] Replace `torch.randperm(...)[:...]` -> `torch.multinomial(replacement=False)`
- [x] Replace `use_cuda` kwarg -> `device`, which is more idiomatic and supports multi-gpu.
- [x] Replace `torch.LongTensor(list(range(...))).cuda()` -> `torch.arange(..., device=device)`
- [x] Replace `torch.zeros(1)` -> `torch.tensor(0., device=device)`

See also #1392

neerajprad:
Should be addressed in #1392. I initially was targeting `dev`, but our `torch-0.4.0` is quite far behind when it comes to scalar support, and I was having to write work-arounds and removing them in our `pytorch-0.5.0` branch.

PR #1395, opened 2018-09-19
Add device keyword for iarange and irange

neerajprad:
Addresses #1393. The deprecated `use_cuda` arg will be removed in the next major release.

### DIFF
--- a/pyro/primitives.py
+++ b/pyro/primitives.py
@@ -200,9 +200,8 @@ class iarange(object):
         right. If not specified, ``dim`` is set to the rightmost dim that is
         left of all enclosing ``iarange`` contexts.
     :param bool use_cuda: Optional bool specifying whether to use cuda tensors
-        for `subsample` and `log_prob`. Defaults to `torch.Tensor.is_cuda`.
-        .. note:: This will be deprecated in `version=0.3`. Please use the
-        ``device`` arg instead.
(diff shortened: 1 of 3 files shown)


pyro/primitives.py
```diff
@@ -92,15 +92,22 @@ class _Subsample(Distribution):
     Internal use only. This should only be used by `iarange`.
     """
 
-    def __init__(self, size, subsample_size, use_cuda=None):
+    def __init__(self, size, subsample_size, use_cuda=None, device=None):
         """
         :param int size: the size of the range to subsample from
         :param int subsample_size: the size of the returned subsample
         :param bool use_cuda: whether to use cuda tensors
```
fritzo: Maybe we can say
```
:param bool use_cuda: DEPRECATED, use `device` instead
```

neerajprad: I have added that to `iarange` and `irange` docs since they'll be visible, but maybe I should mention it up front.

fritzo: oh that's fine, since those are the only public docs.

neerajprad:
> Do you think it's worth updating all the other examples and tutorials in this PR?

I didn't really find any other use cases actually. I think the reason for that is that we only need to do this when subsampling i.e. using the indices to index into the data that is there on the GPU (which we don't often do using `iarange` but rather using custom or PyTorch's data loading utils).

PR #1398, opened 2018-09-20
Remove tensor constructors that use existing tensors

neerajprad:
`new_tensor` or `torch.tensor(source_tensor)`throws a UserWarning on current PyTorch master when it is fed with a source tensor (a python scalar, list and numpy array work fine).

```python
>>> t = torch.ones(3)
>>> t.new_tensor(torch.ones(2))
__main__:1: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
tensor([1., 1.])

```

This changes such invocations to either clone the source tensor (and use `.type` and `.to` methods to specify placement), or uses the `device` and `dtype` args to specify placement at the time of construction.

### DIFF
--- a/pyro/distributions/delta.py
+++ b/pyro/distributions/delta.py
@@ -55,7 +55,7 @@ class Delta(TorchDistribution):
 
     def log_prob(self, x):
         v = self.v.expand(self.shape())
-        log_prob = x.new_tensor(x == v).log()
+        log_prob = (x == v).type(x.dtype).log().to(x.device)
         log_prob = sum_rightmost(log_prob, self.event_dim)
         return log_prob + self.log_density
 
--- a/pyro/distributions/delta.py
+++ b/pyro/distributions/delta.py
@@ -55,7 +55,7 @@ class Delta(TorchDistribution):
 
     def log_prob(self, x):
         v = self.v.expand(self.shape())
-        log_prob = (x == v).type(x.dtype).log().to(x.device)
+        log_prob = (x == v).type(x.dtype).log()
         log_prob = sum_rightmost(log_prob, self.event_dim)
         return log_prob + self.log_density
 
--- a/pyro/distributions/iaf.py
+++ b/pyro/distributions/iaf.py
@@ -79,7 +79,7 @@ class InverseAutoregressiveFlow(Transform):
         sample from the base distribution (or the output of a previous flow)
         """
         mean, scale = self.module.arn(x)
-        scale = self.module.sigmoid(scale + self.module.sigmoid_bias.type(x.dtype).to(x.device))
+        scale = self.module.sigmoid(scale + self.module.sigmoid_bias.type(x.dtype))
 
         y = scale * x + (1 - scale) * mean
         self._add_intermediate_to_cache(x, y, 'x')
--- a/pyro/distributions/iaf.py
+++ b/pyro/distributions/iaf.py
@@ -79,7 +79,7 @@ class InverseAutoregressiveFlow(Transform):
         sample from the base distribution (or the output of a previous flow)
         """
         mean, scale = self.module.arn(x)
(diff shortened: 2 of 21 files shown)


pyro/distributions/delta.py
```diff
@@ -55,7 +55,7 @@ def rsample(self, sample_shape=torch.Size()):
 
     def log_prob(self, x):
         v = self.v.expand(self.shape())
-        log_prob = x.new_tensor(x == v).log()
+        log_prob = (x == v).type(x.dtype).log().to(x.device)
```
fritzo: I'm surprised the `.to(x.device)` is needed.

pyro/distributions/iaf.py
```diff
@@ -79,7 +79,7 @@ def _call(self, x):
         sample from the base distribution (or the output of a previous flow)
         """
         mean, scale = self.module.arn(x)
-        scale = self.module.sigmoid(scale + scale.new_tensor(self.module.sigmoid_bias))
+        scale = self.module.sigmoid(scale + self.module.sigmoid_bias.type(x.dtype).to(x.device))
```
fritzo: ditto: is the `.to(x.device)` really needed?

tests/contrib/gp/test_models.py
```diff
@@ -331,9 +331,9 @@ def f(x):
         return 2 * x + 3 + 5 * torch.sin(7 * x)
 
     tensor_holder = torch.tensor([])
-    X = tensor_holder.new_tensor(torch.arange(100.))
+    X = tensor_holder.new_tensor(range(100))
```
fritzo: consider changing to
```py
X = torch.arange(100, dtype=tensor_holder.dtype, device=tensor_holder.device)
```
here and below

pyro/distributions/delta.py
neerajprad: `.to` is only useful for the multi gpu case, and can be avoided if x and v are on the same device (which I suppose they should be unless we are doing something funny). I think we can remove it.

neerajprad:
This should be good to merge, pending any further comments.

tests/distributions/test_rejector.py
```diff
@@ -45,8 +45,8 @@ def compute_elbo_grad(model, guide, variables):
 @pytest.mark.parametrize('factor', [0.25, 0.5, 1.0])
 def test_rejector(rate, factor):
     num_samples = 100000
-    rates = torch.tensor(torch.tensor(rate).expand(num_samples, 1), requires_grad=True)
-    factors = torch.tensor(torch.tensor(factor).expand(num_samples, 1), requires_grad=True)
+    rates = torch.tensor(rate).expand(num_samples, 1).requires_grad_(True)
```
fritzo: We should never use `.expand(...).requires_grad_()`, since the resulting tensor will be a view. I believe these tests can be made sane by simply removing the `.requires_grad_()`.

fritzo: (If you actually do require grad, you should `.expand(...).contiguous().requires_grad_()`)

tests/distributions/test_rejector.py
```diff
@@ -61,9 +61,8 @@ def test_rejector(rate, factor):
 @pytest.mark.parametrize('factor', [0.25, 0.5, 1.0])
 def test_exponential_elbo(rate, factor):
     num_samples = 100000
-    rates = torch.tensor(rate).expand(num_samples, 1).requires_grad_(True)
-    factors = torch.tensor(factor).expand(num_samples, 1).requires_grad_(True)
-
+    rates = torch.tensor(rate).expand(num_samples, 1)
```
fritzo: How about `torch.Tensor(num_samples, 1).fill_(rate).requires_grad_()` for tests that require grad?

neerajprad:
I suppose this should be good to merge now.

PR #1400, opened 2018-09-21
Implement EKFDistribution

jpchen:
added an ekf distribution which wraps the EKFState in a Pyro distribution for tutorial.

added smoke tests

paircoded with @fritzo

### DIFF
--- a/pyro/contrib/tracking/distributions.py
+++ b/pyro/contrib/tracking/distributions.py
@@ -29,14 +29,14 @@ class EKFDistribution(TorchDistribution):
                        'x0': constraints.real_vector}
     has_rsample = True
 
-    def __init__(self, x0, P0, dynamic_model, measurement_cov, dt=1., validate_args=None):
+    def __init__(self, x0, P0, dynamic_model, measurement_cov, time=1, dt=1., validate_args=None):
         self.x0 = x0
         self.P0 = P0
         self.dynamic_model = dynamic_model
         self.measurement_cov = measurement_cov
         self.dt = dt
         batch_shape = x0.shape[:-1]
-        event_shape = x0.shape[-1:]
+        event_shape = (time,) + x0.shape[-1:]
         super(EKFDistribution, self).__init__(batch_shape, event_shape,
                                               validate_args=validate_args)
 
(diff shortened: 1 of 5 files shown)


pyro/contrib/tracking/distributions.py
```diff
@@ -0,0 +1,46 @@
...
+import pyro.distributions as dist
+from pyro.distributions.torch_distribution import TorchDistribution
+from pyro.contrib.tracking.extended_kalman_filter import EKFState
+from pyro.contrib.tracking.measurements import PositionMeasurement
+
+
+class EKFDistribution(TorchDistribution):
+    r"""
+    EKF Distribution
+
```
fritzo: Could you
1. add docstrings that link to the input datatypes, and
2. ensure this class shows up on docs.pyro.ai

pyro/contrib/tracking/distributions.py
```diff
@@ -0,0 +1,46 @@
...
+        self.dt = dt
+        batch_shape = x0.shape[:-1]
+        event_shape = x0.shape[-1:]
+        super(EKFDistribution, self).__init__(batch_shape, event_shape,
+                                              validate_args=validate_args)
+
+    def rsample(self, sample_shape=torch.Size()):
+        raise NotImplementedError('TODO: implement forward filter backward sample')
+
+    def log_prob(self, value):
```
fritzo: Could you add a docstring indicating that `value` corresponds to a sequence of observations of shape `(time_steps, event_size)`? We may even want to increase the `event_shape` of this distribution to include a dimension for time. WDYT?

pyro/contrib/tracking/distributions.py
```diff
@@ -0,0 +1,46 @@
...
+        state = EKFState(self.dynamic_model, self.x0, self.P0, time=0.)
+        result = 0.
+        zero = value.new_zeros(self.event_shape)
+        for measurement_mean in value:
+            state = state.predict(self.dt)
```
fritzo: on second thought, maybe we should avoid update on the first time step, so that `x0` coincides with `time=0.`?

PR #1403, opened 2018-09-22
Use contract_to_tensor in HMC for discrete site enumeration

neerajprad:
This uses the `contract_tree` operation when `use_einsum=True` to evaluate trace log probability in HMC and NUTS. This is implemented as a separate class `TraceEinsumEvaluator` in addition to the earlier class (now renamed to `TraceTreeEvaluator`).

### Tests:
 - Since this is mostly a refactor, existing tests with `use_einsum=True` should cover this, and all pass.
 - Added the new class to `test_valid_models`, results are identical between the two classes.

### DIFF
--- a/pyro/infer/contract.py
+++ b/pyro/infer/contract.py
@@ -126,6 +126,8 @@ def _contract_component(tensor_tree, sum_dims, target_ordinal=None):
     for t, terms in tensor_tree.items():
         dims_tree[t] = set.union(set(), *(sum_dims[term] for term in terms))
 
+    enum_boundary = max(set.union(*sum_dims.values(), {float("-inf")}))
+
     # Recursively combine terms in different iarange contexts.
     while len(tensor_tree) > 1 or any(dims_tree.values()):
         leaf = max(tensor_tree, key=lambda t: len(t ^ target_ordinal))
@@ -153,7 +155,7 @@ def _contract_component(tensor_tree, sum_dims, target_ordinal=None):
                 for dim in dims:
                     shape[dim] = 1
                 shape.reverse()
-                while shape and shape[-1] == 1:
+                while shape and len(shape) >= -enum_boundary and shape[-1] == 1:
                     shape.pop()
                 shape.reverse()
                 shape = tuple(shape)
--- a/pyro/infer/contract.py
+++ b/pyro/infer/contract.py
@@ -126,7 +126,7 @@ def _contract_component(tensor_tree, sum_dims, target_ordinal=None):
     for t, terms in tensor_tree.items():
         dims_tree[t] = set.union(set(), *(sum_dims[term] for term in terms))
 
-    enum_boundary = max(set.union(*sum_dims.values(), {float("-inf")}))
+    enum_boundary = max(set.union({float("-inf")}, *sum_dims.values()))
 
     # Recursively combine terms in different iarange contexts.
     while len(tensor_tree) > 1 or any(dims_tree.values()):
--- a/pyro/infer/mcmc/util.py
+++ b/pyro/infer/mcmc/util.py
@@ -164,6 +164,11 @@ class TraceEinsumEvaluator(object):
         Populate the ordinals (set of ``CondIndepStack`` frames)
         and enum_dims for each sample site.
         """
+        if not self.has_enumerable_sites:
+            return
+        if self.max_iarange_nesting == float("inf"):
+            raise ValueError("Finite value required for `max_iarange_nesting` when model "
+                             "has discrete (enumerable) sites.")
         model_trace.compute_log_prob()
         self.ordering = {name: frozenset(site["cond_indep_stack"])
                          for name, site in model_trace.nodes.items()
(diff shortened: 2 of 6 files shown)


pyro/infer/contract.py
```diff
@@ -126,6 +126,8 @@ def _contract_component(tensor_tree, sum_dims, target_ordinal=None):
     for t, terms in tensor_tree.items():
         dims_tree[t] = set.union(set(), *(sum_dims[term] for term in terms))
 
+    enum_boundary = max(set.union({float("-inf")}, *sum_dims.values()))
```
fritzo: I'm not sure this is the right level of abstraction for this function, especially there is no mention of enumeration in this file. Do you think it's possible to move this up the stack a bit? Will this be fixed by `ring.broadcast()` in #1405? What was the error that this fixed?

pyro/infer/mcmc/util.py
```diff
@@ -153,3 +132,81 @@ def log_prob(self, model_trace):
...
+
+        :return: log pdf of the trace.
+        """
+        with shared_intermediates():
+            if not self.has_enumerable_sites:
+                return model_trace.log_prob_sum()
+            log_probs = self._get_log_factors(model_trace)
+            sum_dims = {site["log_prob"]: self._enum_dims[name] for name, site in model_trace.nodes.items()
+                        if site["type"] == "sample" and not isinstance(site["fn"], _Subsample)}
+            return contract_to_tensor(log_probs, sum_dims, frozenset()).sum()
```
fritzo: I'm surprised the `.sum()` is needed. Does #1405 fix this?

neerajprad: The issue is that the tensor returned is not a scalar and has singleton dimensions (I think the same as the number of `iarange`s in the model). Let me check with #1405.

pyro/infer/contract.py
neerajprad: Also, just checked that this still errors on #1405 - the failing test is `tests.infer.mcmc.test_valid_models.test_all_discrete_sites_log_prob`.

pyro/infer/mcmc/util.py
neerajprad: So this is fixed in #1405, we wouldn't need the `.sum()` with that.

pyro/infer/contract.py
fritzo: Hmm it would be good to have a failing test in `test/infer/test_contract.py`. I'll try adding one, assuming the cause is `frame.size == 1`.

Note the original motivation for aggressively squeezing was to maximize sharing via `shared_intermediates`. My mental model is not precise enough to gauge how much extra sharing we get.

neerajprad: Great, it works now! I'll take another look at #1405.

neerajprad:
Let us get #1405 merged first; I will rebase and update this PR.

pyro/infer/mcmc/util.py
```diff
@@ -153,3 +132,81 @@ def log_prob(self, model_trace):
...
+        Returns the log pdf of `model_trace` by appropriately handling
+        enumerated log prob factors.
+
+        :return: log pdf of the trace.
+        """
+        with shared_intermediates():
+            if not self.has_enumerable_sites:
+                return model_trace.log_prob_sum()
+            log_probs = self._get_log_factors(model_trace)
+            sum_dims = {site["log_prob"]: self._enum_dims[name] for name, site in model_trace.nodes.items()
```
fritzo: nit: Would this be cheaper as
```py
{model_trace.nodes[name]["log_prob"]: dims for name, dims in self._enum_dims.items()}
```

pyro/infer/mcmc/util.py
```diff
@@ -153,3 +132,81 @@ def log_prob(self, model_trace):
...
+        Populate the ordinals (set of ``CondIndepStack`` frames)
+        and enum_dims for each sample site.
+        """
+        if not self.has_enumerable_sites:
+            return
+        if self.max_iarange_nesting == float("inf"):
+            raise ValueError("Finite value required for `max_iarange_nesting` when model "
+                             "has discrete (enumerable) sites.")
+        model_trace.compute_log_prob()
+        self.ordering = {name: frozenset(site["cond_indep_stack"])
```
fritzo: I believe ordinals should only contain vectorized frames:
```py
frozenset(f for f in site["cond_indep_stack"] if f.vectorized)
```

pyro/infer/mcmc/util.py
```diff
@@ -153,3 +132,81 @@ def log_prob(self, model_trace):
...
+                         for name, site in model_trace.nodes.items()
```
fritzo: nit: Consider fusing these two loops:
```py
for name, site in model_trace.nodes.items():
    if site["type"] == "sample" and not isinstance(site["fn"], _Subsample):
        self._enum_dims[...] = ...
        self.ordering.setdefault(..., []).append(...)
```

pyro/infer/mcmc/util.py
```diff
@@ -153,3 +132,81 @@ def log_prob(self, model_trace):
...
+
+    def log_prob(self, model_trace):
+        """
+        Returns the log pdf of `model_trace` by appropriately handling
+        enumerated log prob factors.
+
+        :return: log pdf of the trace.
+        """
+        with shared_intermediates():
+            if not self.has_enumerable_sites:
```
fritzo: nit: I think this can go outside of `shared_intermediates()`

pyro/infer/mcmc/util.py
neerajprad: Done; in HMC, there shouldn't be non-vectorized frames though.

Issue #1380, opened 2018-09-11
Work around torch.cuda.Tensor maximum dim limit


fritzo:
Much of Pyro's enumeration code requires sparse tensor dimensions, i.e. only a few dimensions of size > 1 but those dimensions are sparsely distributed, e.g. `shape=(3,1,1,1,1,1,14,1,1,1,1,1,1,1,1)`. However torch CUDA tensors have a maximum dim limit (25?). This is preventing message-passing algorithms from running in CUDA on large factor graphs #1379 .

## Proposed solution

One approach would be to create named dimensions rather than globally-positioned dimensions. We're already doing something similar with recycled `iarange`s, and we're already using named dimensions under-the-hood when using `einsum`.

## Tasks

- [x] #1405 Implement packed version of `contract_to_tensor`, e.g. `ubersum`
- [ ] track enumeration dims used by each tensor (like `cond_indep_stack` but not a stack)
- [ ] track dependencies to determine when an enum dim can be recycled. This is a difficult task and @eb8680 and @fritzo may implement multiple mechanisms, e.g.
    - static analysis for dependency tracking
    - context manager
    - python `del` statement
    - declare number of downstream dependencies of each node ("ref counting")
- [ ] switch from `UnpackedLogRing` to `PackedLogRing` in `contract_to_tensor`, when used in `traceenum_elbo.py`
- [ ] implement dim recycling

fritzo:
This is now higher priority since @colesbury has notified us that PyTorch plans to limit tensors to 64 dimensions.

Issue #1391, opened 2018-09-17
Write tech report on enumeration


fritzo:
See also #1355 

## Coding tasks

- [x] #1405 Create a general `ubersum` interface
- [x] #1408 Relax diamond constraint as much as possible
- [x] #1417 Create simple reference implementation with clear semantics
- [x] #1417 Add `ubersum` tests for computing marginals

fritzo:
@martinjankowiak would  you be able to start a draft with rough paper structure? I think you can already write a substantial related work section.

fritzo:
forthcoming

PR #1405, opened 2018-09-22
Refactor towards packed tensor representations

fritzo:
Addresses #1380, #1391 

This refactoring mostly shuffles around logic in `_contract_component()` and `contract_to_tensor()` so those two functions can work with both packed and unpacked tensors. This achieved by moving operations into a new `TensorRing` abstract base class with two implementations: `Packed` and `Unpacked`.

I've also refactored `contract_to_tensor()` by moving broadcasting logic out of `_contract_component()` and into `contract_to_tensor()`. This allows `contract_to_tensor()` to perform connected components decomposition before summing out enumeration dims (the decomposition was already done by `contract_tensor_tree()`).

## Tested

- [x] refactoring is exercised by existing unit tests
- [x] tests for ubersum-as-einsum
- [x] tests for ubersum with multiple outputs
- [x] tests for ubersum with batch dims and product reductions
- [x] tests for `iarange`s of size=1
- [x] death tests for non-tree structure

### DIFF
--- a/pyro/infer/contract.py
+++ b/pyro/infer/contract.py
@@ -95,22 +95,29 @@ class UnpackedLogRing(TensorRing):
         return [d for d in range(-term.dim(), 0) if term.size(d) > 1]
 
     def sumproduct(self, terms, dims):
-        assert all(dim < 0 for dim in dims)
-        if not dims:
-            return sum(terms)
-        shape = list(broadcast_shape(*set(x.shape for x in terms)))
-        for dim in dims:
-            shape[dim] = 1
-        shape.reverse()
-        while shape and shape[-1] == 1:
-            shape.pop()
-        shape.reverse()
-        shape = tuple(shape)
-        return logsumproductexp(terms, shape)
+        key = 'sumproduct', frozenset(id(x) for x in terms), frozenset(dims)
+        if key in self._cache:
+            return self._cache[key]
+
+        if dims:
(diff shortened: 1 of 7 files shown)


pyro/infer/contract.py
```diff
@@ -95,22 +95,29 @@ def dims(self, term):
...
+        while term.dim() and term.size(0) == 1:
+            term = term.squeeze(0)
+        self._save_tensor(term)
+        self._cache[key] = term
+        return term
 
     def product(self, term, ordinal):
         for frame in sorted(ordinal, key=lambda f: -f.dim):
-            if term.shape[frame.dim] != 1:
+            if -frame.dim <= term.dim() and term.size(frame.dim) != 1:
```
neerajprad: Any reason why this is changed to `.size` now?

fritzo: #1406

neerajprad: Interesting! We changed from `.size` to `.shape` in the last release, but I suppose that was only for the sake of consistency, rather than perf.

fritzo:
> Does this fix #1380?

This does not fix #1380, but is a step towards fixing #1380. I'll update #1380 with a task list of remaining work.

I think this is ready to merge, pending review. I'll be making further updates to relax some restrictions, as we work on #1391 .

PR #1407, opened 2018-09-24
fix docstring for renyi elbo

martinjankowiak:
make defaults consistent with docstring

### DIFF
--- a/pyro/infer/renyi_elbo.py
+++ b/pyro/infer/renyi_elbo.py
@@ -14,12 +14,13 @@ from pyro.util import check_if_enumerated, warn_if_nan
 class RenyiELBO(ELBO):
     r"""
     An implementation of Renyi's :math:`\alpha`-divergence variational inference
-    follows reference [1].
+    following reference [1].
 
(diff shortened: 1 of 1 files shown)


pyro/infer/renyi_elbo.py
```diff
@@ -14,12 +14,13 @@
...
-    reference [1], depending on the dataset, :math:`\alpha < 0` might give better
-    results. In the special case :math:`\alpha = 0`, we have important weighted
-    lower bound derived in reference [2].
+    In order for the objective to be a strict lower bound, we require
+    :math:`\alpha \ge 0`. Note, however, that according to reference [1], depending
+    on the dataset :math:`\alpha < 0` might give better results. In the special case
+    :math:`\alpha = 0`, the objective function is that of the important weighted
+    autoencoder derived in reference [2].
 
     .. note:: Setting :math:`\alpha < 1` gives a better bound than the usual ELBO.
```
martinjankowiak: @fehiepsi what is 'a better bound' supposed to mean here?

Issue #1355, opened 2018-09-02
Add tutorial on enumeration


fritzo:
As `TraceEnum_ELBO` gains functionality, we should describe how to use it.

Note that the gmm tutorial mentions enumeration, but I'd like a tutorial that focuses on the language aspects rather than a single model. We can point the the gmm tutorial and hmm example from the enumeration tutorial.

## Previous discussions on forum.pyro.ai

- https://forum.pyro.ai/t/question-on-local-global-random-variable-in-pyro-hmm-example/444/9
- https://forum.pyro.ai/t/traceenum-elbo-parallel-enumeration-fails-with-error-about-cartesian-products/452/6

## Topics

- [ ] restrictions on `iarange` dependency structure of `TraceEnum_ELBO`
- [ ] guide-side vs model-side enumeration
- [ ] restrictions on `iarange` nesting structure in model-side enumeration
- [ ] replacing a problematic `iarange` with an `irange` (see [this comment](https://github.com/uber/pyro/pull/1343/files#r214530423))
- [ ] restrictions on model-guide dependencies
- [ ] how to estimate computational complexity
- [ ] using `with cached_paths()` for large graphs
- others?

mbabadi:
Chiming in my experience -- I have learned most of what I know about Pyro by working through the realistic examples (e.g. VAE, DMM, AIR, etc.); those are _great_, and that's perhaps the best way for newcomers to dive in. That being said, I had to go through http://pyro.ai/examples/tensor_shapes.html multiple times in order to understand the logic of enumeration, nested `iarange`, etc.

It would be great to have a formal guide/tutorial on the language aspects, perhaps using abstract graphical models that showcase typical examples of dependency structures (rather than concrete models) to facilitate the discussion.

Anyway, I love where Pyro is headed! it is so liberating to be able to just think about models and worry less about logsumexps!

PR #1408, opened 2018-09-25
Relax diamond restriction on iarange nesting for enumeration

fritzo:
Addresses #1391 

This partially relaxes the no-loops restriction on `iarange` nesting during enumeration. After this PR, there is still a restriction involving `iarange` nesting, and I believe this is optimal for the complexity class. I am working on a parsimonious statement of this restriction in #1391 .

The change in this PR was basically to compute the `parent` ordinal independently for each connected component of tensors at each ordinal, rather than compute a single `parent` for each ordinal.

## Tested

- [x] added a regression tests for `ubersum`
- [x] added a regression test for `TraceEnum_ELBO`

### DIFF
--- a/pyro/infer/contract.py
+++ b/pyro/infer/contract.py
@@ -280,19 +280,23 @@ def _contract_component(ring, tensor_tree, sum_dims):
                 pending.append(tu)
 
     # Collect contraction dimensions by ordinal.
-    dims_tree = defaultdict(set)
+    dim_to_ordinal = {}
     for t, terms in tensor_tree.items():
-        dims_tree[t] = set.union(set(), *(sum_dims[term] for term in terms))
+        for term in terms:
+            for dim in sum_dims[term]:
+                dim_to_ordinal[dim] = dim_to_ordinal.get(dim, t) & t
+    dims_tree = defaultdict(set)
+    for dim, t in dim_to_ordinal.items():
+        dims_tree[t].add(dim)
 
     # Recursively combine terms in different iarange contexts.
     while any(dims_tree.values()):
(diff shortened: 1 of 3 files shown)


pyro/infer/contract.py
```diff
@@ -213,52 +213,6 @@ def broadcast(self, term, ordinal):
         return term
 
 
-def _check_tree_structure(tensor_tree):
-    # Check for absence of diamonds, i.e. collider iaranges.
-    # This is required to avoid loops in message passing.
-    #
-    # For example let f1, f2 be two CondIndepStackFrames, and consider
```
fritzo: This is a nice figure but is no longer correct. I'll try to add something like it to both the tutorial #1355 and the tech report #1391 .

tests/infer/test_enum.py
```diff
@@ -2328,12 +2328,141 @@ def guide(data):
         elbo.differentiable_loss(model_iarange_iarange, guide, data)
 
 
+@pytest.mark.parametrize('scale', [1, 10])
+def test_elbo_enumerate_iaranges_7(scale):
```
fritzo: The following example motivates this PR.

neerajprad:
Looks great! I checked that this works with one of the dag examples in `mcmc.test_valid_models`, as well. I would like to dig in a bit more, and understand how the `Expected tree-structured iarange nesting..` warning is triggered with this refactor in another pass shortly.

fritzo:
@neerajprad let me first add a test for `ubersum("ae,be,abe->", ..., batch_dims="ab")` and make sure we can handle this.

Issue #1397, opened 2018-09-20
Master task - test clean up


neerajprad:
- [ ] Filter warnings when running doctest.
 - [ ] Change cuda tests to not use xdist - multiple threads run out of GPU memory; moreover the parallelization rarely improves the run time.
 - [ ] Investigate connection refused error in AIR tutorial.
 - [ ] Remove progress bar when using pytest - problematic with multiple threads.
 - [ ] Skip slow tests - investigate why HMC tests are running slower on current PyTorch master.

PR #1410, opened 2018-09-25
Minor testing related fixes

neerajprad:
Addresses #1397. 
 - Remove warnings from doctest.
 - Cuda tests run in a single process. Many of our tests construct fairly big tensors which results in OOM errors otherwise, besides not being any faster.
 - Disable progress bar in HMC when pytest-xdist is used.
 - Remove deprecated `pandas.DataFrame.as_matrix` method.
 - Do not throw error if the cached path in einsum was dumped using a higher protocol.


neerajprad:
This should be good to merge.

PR #1417, opened 2018-09-27
Add naive_ubersum reference implementation + tests

### LINKED ISSUE #1414: test_contract failing on PyTorch master
Some of our einsum tests (`test_ubersum_sizes`) are failing on PyTorch master:

e.g.

```python
X = torch.randn(2, 1)
Y = torch.randn(1, 3)
Z = torch.randn(3, 4)
ubersum('ab,bc,cd->a,b,c,d', X, Y, Z, batch_dims='ad')
```

throws an exception,

```
        if len(operands) == 1 and isinstance(operands[0], (list, tuple)):
            # the old interface of passing the operands as one list argument
            operands = operands[0]
>       return torch._C._VariableFunctions.einsum(equation, operands)
E       RuntimeError: size of dimension does not match previous size, operand 1, dim 0
```

The issue seems to be that PyTorch release does implicit broadcasting of operands whereas the new einsum does not. e.g.

**PyTorch Release**

```python
In [7]: torch.einsum('a,a->', [torch.tensor([1.]), torch.tensor([1., 1.])])
Out[7]: tensor(2.)
```

**PyTorch Master:**

```python
In [2]: torch.einsum('a,a->', [torch.tensor([1.]), torch.tensor([1., 1.])])
```
```
<ipython-input-2-65721fd81f1e> in <module>()
----> 1 torch.einsum('a,a->', [torch.tensor([1.]), torch.tensor([1., 1.])])

~/miniconda2/envs/pytorch-master/lib/python3.6/site-packages/torch/functional.py in einsum(equation, *operands)
    239         # the old interface of passing the operands as one list argument
    240         operands = operands[0]
--> 241     return torch._C._VariableFunctions.einsum(equation, operands)
    242
    243

RuntimeError: size of dimension does not match previous size, operand 1, dim 0
```

cc. @fritzo

fritzo:
Addresses #1391
Fixes #1414

This clarifies the semantics of `ubersum()` by adding a simple reference implementation. The `naive_ubersum()` reference implementation is slow but handles both polynomial-time and exponential-time contractions, whereas `ubersum()` currently raises `NotImplementedError` for exponential-time contractions. It's actually very very slow; it is good for testing but wouldn't be able to serve as a fallback implementation for `ubersum()`'s NotImplemented cases.

This PR also fixes a shape bug in the `torch_log` backend (causing #1414) and adds regression tests.

## Tested

- [x] add test for agreement between `contract()` and `naive_ubersum()`
- [x] add test for agreement between `ubersum()` and `naive_ubersum()`
- [x] parametrize existing tests to also test `naive_ubersum()`
- [x] add tests for preserved batched dimensions
- [x] mark `ubersum()` discrepancies as xfail

### DIFF
--- a/pyro/ops/contract.py
+++ b/pyro/ops/contract.py
@@ -1,11 +1,15 @@
 from __future__ import absolute_import, division, print_function
 
+import itertools
+import logging
 from abc import ABCMeta, abstractmethod
 from collections import OrderedDict, defaultdict
 
+import opt_einsum
 import torch
 from opt_einsum import shared_intermediates
 from six import add_metaclass
+from six.moves import map
 
 from pyro.distributions.util import broadcast_shape
 from pyro.ops.einsum import contract
@@ -308,11 +312,12 @@ def _contract_component(ring, tensor_tree, sum_dims):
             else:
                 parent = frozenset.union(*(t for t, d in dims_tree.items() if d & remaining_dims))
                 if parent == leaf:
-                    raise ValueError("Expected tree-structured iarange nesting, but found "
-                                     "dependencies on independent iaranges [{}]. "
-                                     "Try converting one of the iaranges to an irange (but beware "
-                                     "exponential cost in the size of that irange)"
-                                     .format(', '.join(getattr(f, 'name', str(f)) for f in leaf)))
+                    raise NotImplementedError(
+                        "Expected tree-structured iarange nesting, but found "
+                        "dependencies on independent iaranges [{}]. "
+                        "Try converting one of the iaranges to an irange (but beware "
+                        "exponential cost in the size of that irange)"
+                        .format(', '.join(getattr(f, 'name', str(f)) for f in leaf)))
                 contract_frames = leaf - parent
                 term = ring.product(term, contract_frames)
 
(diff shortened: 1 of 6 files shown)


tests/ops/test_contract.py
```diff
@@ -364,7 +403,7 @@ def test_ubersum_collide_error():
     x = torch.randn(a, c)
     y = torch.randn(b, d)
     z = torch.randn(a, b, c, d)
-    with pytest.raises(ValueError, match='Expected tree-structured iarange nesting'):
+    with pytest.raises(NotImplementedError, match='Expected tree-structured iarange nesting'):
```
neerajprad: Would it be correct that `naive_ubersum` would handle these error cases now, albeit in exponential time?

pyro/ops/contract.py
```diff
@@ -487,3 +492,97 @@ def ubersum(equation, *operands, **kwargs):
...
+    inputs = inputs.split(',')
+    output_dims = set(output)
+    keep_dims = output_dims - batch_dims
+    sizes = {dim: size
+             for dims, term in zip(inputs, operands)
+             for dim, size in zip(dims, term.shape)}
+
+    # It is unclear what the desired behavior is in this case.
+    if batch_dims & output_dims:
+        raise NotImplementedError
```
neerajprad: It will be nice if `naive_ubersum` could handle cases that are a superset of those handled by `ubersum` (though I suppose it depends on what is the purpose of this reference implementation). One thing we could do is to first call `ubersum` and only go forward with the naive implementation if the case returns a `NotImplementedError`.

neerajprad: On second thoughts, that defeats the point of having a reference implementation. Could you explain the difficulty in handling the case where some batch dimensions are not contracted?

tests/ops/test_contract.py
fritzo: That is correct.

neerajprad:
@fritzo - pushing a fix for the dependency issue.

fritzo:
Ok, I think this is ready to merge pending review. I'd like to address discrepancies with `ubersum` vs `naive_ubersum` in a follow-up PR, if that's ok with reviewers.

pyro/ops/contract.py
```diff
@@ -487,3 +525,113 @@ def ubersum(equation, *operands, **kwargs):
...
+    # intersection over all batch contexts of tensors in which the dim appears.
+    dim_to_ordinal = {}
+    for dims in map(set, inputs):
+        ordinal = dims & batch_dims
+        for dim in dims - batch_dims:
+            dim_to_ordinal[dim] = dim_to_ordinal.get(dim, ordinal) & ordinal
+    for dim in output_dims - batch_dims:
+        missing_dims = dim_to_ordinal[dim] - output_dims
+        if missing_dims:
+            raise ValueError(u"It is nonsensical to preserve a batched dim without preserving "
```
neerajprad: Should this be preserving a "non-batched" dim?

fritzo: I'm trying to distinguish a dim-that-is-batched (i.e. a local enumeration dim) from a batch dim (i.e. an iarange dim). Martin suggested using the term `plate dim`. I guess we can hone that language in the paper then adjust the code.

neerajprad: > I'm trying to distinguish a dim-that-is-batched (i.e. a local enumeration dim) from a batch dim (i.e. an iarange dim)

Makes sense. `plate dim` sounds like a reasonable choice.

PR #1422, opened 2018-09-28
Batch hidden sample in hmm example

jaberg:
Hi - I'm new to Pyro, iarange etc. but I was wondering if the hmm example might have a bug in the way the hidden state is sampled. Does it?

If so, the bug might also be in the other model_{2,3,4} functions too.

### DIFF
--- a/examples/hmm.py
+++ b/examples/hmm.py
@@ -72,7 +72,7 @@ def model_1(sequences, lengths, args, batch_size=None, include_prior=True):
     tones_iarange = pyro.iarange("tones", data_dim, dim=-1)
     with pyro.iarange("sequences", len(sequences), batch_size, dim=-2) as batch:
         lengths = lengths[batch]
-        x = 0
+        x = 0 * batch
         for t in range(lengths.max()):
             with poutine.mask(mask=(t < lengths).unsqueeze(-1)):
                 # On the next line, we'll overwrite the value of x with an updated


jaberg:
I can't sign the CLA right now, but if this is legit and someone else doesn't mind making the fix throughout this file, they can go for it and consider this a bug report rather than a contribution.

fritzo:
Thanks for reading the example in detail! In fact `x = 0` is a valid initial value due to Pyro's broadcasting semantics. Note that we're never storing or sampling from the initial value of `x`; the only place it is used is in [line 81](https://github.com/uber/pyro/blob/f9923f205fa8b1d993f1e374b384860ec6941f44/examples/hmm.py#L81)
```py
x = pyro.sample("x_{}".format(t), dist.Categorical(probs_x[x]),
```
in this case
```
probs_x.shape == (hidden_dim, hidden_dim)
probs_x[0].shape == (hidden_dim,)
``
and using [`@poutine.broadcast`](http://docs.pyro.ai/en/0.2.1-release/poutine.html#pyro.poutine.broadcast) Pyro knows to expand the `Categorical` distribution up to `batch` shape.

This may seem too clever. Maybe it is. But the subsequent `probs_x[x]` indexing really make use of broadcasting.

**Comment by jaberg** [comment]
Hi @fritzo ,

I'm not sure this case is closed...
```
posterior = pyro.infer.Importance(model_1, guide, num_samples=10)
#sequences, lengths are TEST DATA
post = posterior.run(sequences, lengths, args)
assert list(post.exec_traces[0].nodes['y_0']['value'].shape) == [sequences.shape[0], sequences.shape[2]]
assert list(post.exec_traces[0].nodes['x_0']['value'].shape) == [sequences.shape[0]]
```
Importance sampling is admittedly not the best way to do HMM inference, but the two assertions should pass right?  What happens in the example is that the second assertion fails because there isn't a different value for x for every sequence in the batch.  When running the code with the change I made, the log probabilities are also in a very different range, which I thought was consistent with there being more sampled variables (one for each sequence in the batch).

The two assertions you mention in your reply are correct, but the potential issue here is with x, not probs_x.

jaberg:
Thanks for taking the time to explain. It's still blowing my mind. Any insight you can provide will be appreciated as I wrap my head around pyro's iaranges, broadcasting, etc.

1. Is it the case that each `exec_trace[i]` is a [ancestral] sample from the model?
2. A HMM calls for a different progression of hidden states for each visible sequence in a batch, is that what's expressed in model_1?
3. If that is what's expressed by model_1, where can those samples for x be found, if they are not in `exec_trace[i].nodes['x_0']['value']` tensor ?
4. why would you want to broadcast log prob values up to the batch_shape?  Again I can see that you'd want to broadcast the probability of the whole probs_x matrix (under the dirichlet), but I can't see wanting to broadcast the probability of any of the x[t] values (sampled from the various categoricals).
5. Is replacing `x = 0` with `x = 0 * batch` unnecessary or actually inducing a wrong model? The training code runs both ways, and it seems to lead two different behaviours, in terms of the range of log probs and the dimensions of what ends up in exec_trace.

fritzo:
Could you ask these questions on http://forum.pyro.ai so that others may benefit from extended conversations? I feel like closed github comments are difficult to discover.

jaberg:
Certainly, thanks!

Issue #1427, opened 2018-10-01
Flakiness in test_paths in CI


neerajprad:
It seems like some of the parametrized tests are failing. @fritzo: Is this expected? If the flop counts are not deterministic, should we test within a broader margin? 

Failing CI build - https://travis-ci.org/uber/pyro/jobs/435351412.

```bash
tests/ops/einsum/test_paths.py:166: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
equation = 'chd,bde,agbc,hiad,hgc,hgi,hiad->'
shapes = [(3, 4, 2), (3, 2, 2), (4, 3, 3, 3), (4, 2, 4, 2), (4, 3, 3), (4, 3, 2), ...]
    def _test_path(equation, shapes):
        inputs, output = equation.split('->')
        inputs = inputs.split(',')
        dim_sizes = {dim: size for dims, shape in zip(inputs, shapes)
                     for dim, size in zip(dims, shape)}
        operands = [torch.randn(shape).log() for shape in shapes]
        logging.debug(equation)
    
        # Compute path using Pyro.
        pyro_time = -timeit.default_timer()
        pyro_path = optimize(inputs, output, dim_sizes)
        _, pyro_info = opt_einsum.contract_path(equation, *operands, path=pyro_path)
        pyro_time += timeit.default_timer()
        pyro_info = '\n'.join(pyro_info.splitlines()[1:7])
    
        assert sum(map(len, pyro_path)) - len(pyro_path) + 1 == len(inputs)
        path = ssa_to_linear(linear_to_ssa(pyro_path))
        assert path == pyro_path
    
        # Compute path using opt_einsum's greedy method.
        opt_time = -timeit.default_timer()
        opt_path, opt_info = opt_einsum.contract_path(equation, *operands, path='greedy')
        opt_time += timeit.default_timer()
        opt_info = '\n'.join(opt_info.splitlines()[1:7])
    
        # Check path quality.
        logging.debug(u'Pyro path took {}s:\n{}'.format(pyro_time, pyro_info))
        logging.debug(u'opt_einsum took {}s:\n{}'.format(opt_time, opt_info))
        pyro_flops = float(re.search('Optimized FLOP count:(.*)', pyro_info).group(1))
        opt_flops = float(re.search('Optimized FLOP count:(.*)', opt_info).group(1))
[...]

PR #1428, opened 2018-10-01
Xfail flaky assert in test_path

neerajprad:
See #1427 for details.


neerajprad:
Should be good to merge.

PR #1430, opened 2018-10-01
Update einsum optimizer to match upstream

fritzo:
Mirrors https://github.com/dgasmith/opt_einsum/pull/60

This PR merely mirrors changes suggested in upstream review by @jcmgray.


tests/ops/einsum/test_paths.py
```diff
@@ -138,9 +137,12 @@ def _test_path(equation, shapes):
...
-    opt_info = '\n'.join(opt_info.splitlines()[1:7])
 
     # Check path quality.
+    _, pyro_info = opt_einsum.contract_path(equation, *operands, path=pyro_path)
+    # Wrap in text_type to avoid breaking after https://github.com/dgasmith/opt_einsum/pull/61
+    pyro_info = '\n'.join(text_type(pyro_info).splitlines()[1:7])
+    opt_info = '\n'.join(text_type(opt_info).splitlines()[1:7])
     logging.debug(u'Pyro path took {}s:\n{}'.format(pyro_time, pyro_info))
     logging.debug(u'opt_einsum took {}s:\n{}'.format(opt_time, opt_info))
     pyro_flops = float(re.search('Optimized FLOP count:(.*)', pyro_info).group(1))
```
neerajprad: Was the non-determinism in the test inputs independent of the issue with `pyro_flops <= opt_flops * 1.5 + 2.0`, or would we expect this inequality to hold now?

fritzo: I believe the inequalities should hold now, since I've fixed the nondeterminism bug. Would you like me to re-enable these as assertions (rather than error messages)?

tests/ops/einsum/test_paths.py
```diff
@@ -162,7 +164,7 @@ def _test_path(equation, shapes):
 def test_contract(equation):
     inputs, output = equation.split('->')
     inputs = inputs.split(',')
-    symbols = set(output).union(*inputs)
+    symbols = sorted(set(output).union(*inputs))
```
fritzo: This fixes the nondeterminism bug #1427

tests/ops/einsum/test_paths.py
neerajprad: Thanks for removing the error message!

PR #1431, opened 2018-10-04
Update to using pytorch-1.0-rc

neerajprad:
This is simply reopening #1376 with an updated branch name. Note that we will keep this as the base PR until PyTorch 1.0 is released. We should merge any changes from dev directly, but any additional functionality that uses `pytorch-1.0` as a base branch should be a regular PR.

## Tasks

- [x] Remove obsolete patches for 0.4.1 release (@fritzo)
- [x] Update `pyro.distributions` to remove wrappers for `enumerate_support`, `expand` 
- [x] Udate pyro.ops.jit.trace
- [x] Update jit wrappers (@neerajprad)

NOTE: The slowdown that we observed and noted in https://github.com/pytorch/pytorch/issues/12190 still continues, and the unit tests are slower. My hope is that this will be resolved when https://github.com/pytorch/pytorch/pull/13420 merges.


neerajprad:
`test_memory.py` is throwing a very strange DeprecationWarning in `c10.reduce_op`, but I have not been able to replicate it locally (trying still).

pyro/distributions/diag_normal_mixture.py
```diff
@@ -67,7 +67,22 @@ def __init__(self, locs, coord_scale, component_logits):
         self.dim = locs.size(-1)
         self.categorical = Categorical(logits=component_logits)
         self.probs = self.categorical.probs
-        super(MixtureOfDiagNormals, self).__init__(batch_shape=batch_shape, event_shape=(self.dim,))
+        super(MixtureOfDiagNormals, self).__init__(batch_shape=torch.Size(batch_shape),
+                                                   event_shape=torch.Size((self.dim,)))
+
+    def expand(self, batch_shape, _instance=None):
```
martinjankowiak: @neerajprad why are these expand methods necessary?

pyro/distributions/relaxed_straight_through.py
```diff
@@ -29,10 +29,6 @@ class RelaxedOneHotCategoricalStraightThrough(RelaxedOneHotCategorical):
     [2] Categorical Reparameterization with Gumbel-Softmax,
         Eric Jang, Shixiang Gu, Ben Poole
     """
-    def __init__(self, temperature, probs=None, logits=None, validate_args=None):
```
martinjankowiak: what's going on here? where did the temperatures go?

neerajprad: Since its just calling its parent with the same arguments, its not needed. We can just rely on the parent's init method.

pyro/distributions/diag_normal_mixture.py
neerajprad: We don't have `ReshapedDistribution` anymore, and we have implemented a `.expand` method for each of the distributions (in PyTorch). This is used when we call `dist.expand()`. If we are only overriding any existing distribution without changing its `__init__` method or using TransformedDistribution, we don't need to define it though.

neerajprad: This was the original issue - https://github.com/uber/pyro/issues/1337. And this was the pytorch PR - https://github.com/pytorch/pytorch/pull/11341#pullrequestreview-153078931.

neerajprad:
I expect that the timeout in `test_examples` should be resolved by #1571.

tests/test_examples.py
```diff
@@ -71,27 +71,35 @@
...
+    xfail_jit('contrib/gp/sv-dkl.py --epochs=1 --num-inducing=4 --jit'),
+    xfail_jit('dmm/dmm.py --num-epochs=1 --jit'),
+    xfail_jit('dmm/dmm.py --num-epochs=1 --num-iafs=1 --jit'),
+    xfail_jit('eight_schools/mcmc.py --num-samples=500 --warmup-steps=100 --jit'),
+    xfail_jit('eight_schools/svi.py --num-epochs=1 --jit'),
+    xfail_jit('examples/contrib/gp/sv-dkl.py --epochs=1 --num-inducing=4 --jit'),
+    xfail_jit('hmm.py --num-steps=1 --truncate=65 --model=1 --jit'),
+    xfail_jit('hmm.py --num-steps=1 --truncate=65 --model=2 --jit'),
+    xfail_jit('hmm.py --num-steps=1 --truncate=65 --model=3 --jit'),
+    xfail_jit('hmm.py --num-steps=1 --truncate=65 --model=4 --jit'),
```
fritzo: This is missing model=5

neerajprad:
I have put the current dev on `pytorch-0.4.0` branch. This should be good to merge once #1571 is merged into this (and hopefully takes care of the build timeout issues).

neerajprad:
Update - @jpchen just mentioned that this will break our doc generation script. He is working on a PR that fixes that and which we can merge into this branch before the final dev merge.

jpchen:
the [build](https://readthedocs.org/projects/pyro-ppl/builds/8159615/) is failing due to memory limits. i reached out to rtd to request more as this is a blocker for us when pytorch releases. we can still merge this knowing this will break the [dev docs](http://docs.pyro.ai/en/pytorch-1.0/)

fritzo:
@neerajprad will this be ready to merge as soon as conflicts are resolved?

neerajprad:
Just merged with dev. This should be good to merge once tests pass.

neerajprad:
> @neerajprad what's the status of the performance issues as of the 11/28 nightly build? Are they mostly resolved as suggested in pytorch/pytorch#12190 ?

The 11/28 nightly build was not available when I checked late last night, but the issue seems to be mostly resolved in my local build. So either its a platform specific issue or the way the wheels are built. Once this PR merges, I will update to the 11/28 build separately and update on that issue.

eb8680:
> Once this PR merges, I will update to the 11/28 build separately and update on that issue.

Do we want to merge this before double-checking?

neerajprad:
> Do we want to merge this before double-checking?

It is an external issue that mostly affects HMC more than other things. It will be nice to have this in dev, and be able to make other major refactoring changes. We expect the perf issues to be resolved either in the final release or in the pre release. But we shouldn't merge this if in the worst case, we don't want to release using pytorch-1.0 if the perf issues persist!

neerajprad:
Let me quickly check if 11/28 is available, but I would like to merge this soon so as not to accumulate more merge conflicts.

neerajprad:
The new wheel seems to throw an error saying `typing not found` when we import `torch`. I'm not sure where it was getting the package from earlier, looking into it.

neerajprad:
I think we can already see from the unit tests that the slowdown still persists, and many tests (mostly HMC) are taking almost twice as long. Note that unit test stage on this branch is expected to take longer because we have more tests, but even individual tests are significantly slower. I have not been able to replicate this locally though (yet).

PR #1439, opened 2018-10-12
Implementation of Inference Compilation

wsgharvey:
Contains an implementation of inference compilation (see <https://arxiv.org/abs/1610.09900>). This allows the user to specify a model and a guide, and amortize inference by using traces from the model to train the guide to propose approximate posterior distributions. During inference, this approximate posterior is used as a proposal distribution for importance sampling.

This implementation differs from the paper in that the guide program must be specified by the user, instead of being automatically generated to fit the model. Also, the model and guide are run separately during inference. This is different to the implementation in the paper, where they are run together, which allows the guide program to make use of the parameters at each sample statement in the model when proposing a posterior distribution for that sample statement.

An example using this implementation is available at <https://github.com/wsgharvey/pyro-infcomp-examples>.


pyro/infer/csis.py
```diff
@@ -0,0 +1,161 @@
...
+    program to minimise KL(model posterior || guide), and inference with
+    importance sampling.
+
+    **Reference**
+    "Inference Compilation and Universal Probabilistic Programming" `pdf https://arxiv.org/pdf/1610.09900.pdf`
+
+    :param model: probabilistic model defined as a function. Must accept
+        keyword arguments with names of observed sites in model.
+    :param guide: guide function which is used as an approximate posterior. Must
+        accept keyword arguments with names of observed sites in model.
```
eb8680: nit: instead of one keyword argument per site, what about requiring the guide to take a single keyword argument `observations`, a dictionary whose keys are names of observed sites?  That feels like a more idiomatic way to write guides that take lots of observed sites with mechanically generated names, e.g. time series models.

pyro/infer/csis.py
```diff
@@ -0,0 +1,161 @@
...
+        Take a gradient step on the loss function. Any args or kwargs are
+        passed to the model and guide.
+        """
+        loss = self.loss(grads=True)
+        self.optim.step()
+        self.optim.zero_grad()
+
+        return torch_item(loss)
+
+    def loss(self, grads=False, batch=None):
```
eb8680: If you refactor this a little bit to look more like `infer.Trace_ELBO`, it'll be much easier to add support for the PyTorch JIT (which currently doesn't support calling `backward` within compiled functions).  In particular, you can rename `CSIS.loss` to `loss_and_grads`, define a `differentiable_loss_particle` method that computes `particle_loss` and call that inside the main loop in `loss_and_grads`.

tests/infer/test_csis.py
```diff
@@ -0,0 +1,70 @@
...
+    csis = pyro.infer.CSIS(model,
+                           guide,
+                           torch.optim.Adam(guide.parameters()),
+                           num_inference_samples=100)
+    # observations chosen so that proposal distribution and true posterior will both have zero mean
+    posterior = csis.run(y1=torch.tensor(-1.0),
+                         y2=torch.tensor(1.0))
+    assert_equal(len(posterior.exec_traces), 100)
+    marginal = pyro.infer.EmpiricalMarginal(posterior, "x")
+    assert_equal(marginal.mean, torch.tensor(0.0), prec=0.25)
```
eb8680: This seems like a high tolerance, was the test failing with a lower one?

pyro/poutine/uncondition_messenger.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import absolute_import, division, print_function
+
+from .messenger import Messenger
+
+
+class UnconditionMessenger(Messenger):
```
eb8680: Could you add a simple test or two for this to `tests/poutine/test_poutines.py`?  It seems like a generally useful utility, and tests help with maintainability.

tests/infer/test_csis.py
```diff
@@ -0,0 +1,70 @@
...
+    return x
+
+
+class Guide(nn.Module):
+    def __init__(self):
+        super(Guide, self).__init__()
+        self.linear = torch.nn.Linear(1, 1, bias=False)
+        self.std = torch.nn.Parameter(torch.tensor(1.))
+
+    def forward(self, y1, y2):
```
eb8680: Call `pyro.module(<name>, self)` here for compatibility with the Pyro param store and the suggested changes to `CSIS.step`.

tests/infer/test_csis.py
```diff
@@ -0,0 +1,70 @@
...
+        mean = self.linear((y1+y2).view(1, 1))[0, 0]
+        pyro.sample("x", dist.Normal(mean, self.std))
+
+
+@pytest.mark.init(rng_seed=7)
+def test_csis_sampling():
```
eb8680: Call `pyro.clear_param_store()` in the test for compatibility with the Pyro param store

tests/infer/test_csis.py
```diff
@@ -0,0 +1,70 @@
...
+    # observations chosen so that proposal distribution and true posterior will both have zero mean
+    posterior = csis.run(y1=torch.tensor(-1.0),
+                         y2=torch.tensor(1.0))
+    assert_equal(len(posterior.exec_traces), 100)
+    marginal = pyro.infer.EmpiricalMarginal(posterior, "x")
+    assert_equal(marginal.mean, torch.tensor(0.0), prec=0.25)
+
+
+@pytest.mark.init(rng_seed=7)
+def test_csis_parameter_update():
```
eb8680: Call `pyro.clear_param_store()` in the test for compatibility with the Pyro param store.

wsgharvey:
Thanks for the comments. I've also used this on a tool to infer data from images of bar charts (<https://github.com/wsgharvey/pyro-plot-reader>). For this, the performance was okay - not great but not as big a bottleneck as rendering an image with matplotlib every time the model was run.

And yeah, I'd be happy to do an example/tutorial.

pyro/poutine/uncondition_messenger.py
```diff
@@ -0,0 +1,22 @@
...
+        super(UnconditionMessenger, self).__init__()
+
+    def _pyro_sample(self, msg):
+        """
+        :param msg: current message at a trace site.
+
+        Samples value from distribution, irrespective of whether or not the
+        node has an observed value.
+        """
+        msg["value"] = msg["fn"](*msg["args"], **msg["kwargs"])
```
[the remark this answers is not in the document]
wsgharvey: The reason I didn't do it in this way before was so that I could use `model_trace.observation_nodes` to iterate over the "observations" in `CSIS._get_matched_trace`. Since I don't think this will work with your suggestion, would it be reasonable to add another flag like:
```
if msg["is_observed"]:
    msg["is_observed"] = False
    msg["value"] = None
    msg["done"] = False
    msg["was_observed"] = True
```
I just want to check this is sufficiently Pyro-ey and won't mess up any other parts.

pyro/infer/csis.py
```diff
@@ -0,0 +1,161 @@
...
+        self.kwargs = {}
+
+    def set_validation_batch(self, validation_batch_size=20):
+        """
+        Samples a batch of model traces and stores it as an object property.
+        """
+        self.validation_batch = [self._sample_from_joint()
+                                 for _ in range(validation_batch_size)]
+
+    def set_args(self, *args, **kwargs):
```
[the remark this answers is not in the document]
wsgharvey: Sure, I did it that way as it seemed messy to ask the user to continually input arguments during compilation and inference when they always need to be the same for correctness. `functools.partial` sounds like a good solution though.

pyro/infer/csis.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        :returns: loss estimated using validation batch
+        :rtype: float
+
+        Calculates loss on validation batch. `set_validation_batch` must have
+        been called previously. Can be used to track loss in a less noisy way
+        during training.
+        """
+        if self.validation_batch is None:
+            raise ValueError("Validation batch not set.")
```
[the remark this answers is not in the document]
wsgharvey: Good point, will change

pyro/infer/csis.py
wsgharvey: That's a good idea, I'll change that

pyro/poutine/uncondition_messenger.py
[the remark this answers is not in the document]
eb8680: Ah, I see - adding that new flag is fine, except instead of using `msg["was_observed"]` use `msg["infer"]["was_observed"]`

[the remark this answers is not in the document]
wsgharvey: Great, will do

pyro/infer/csis.py
wsgharvey: Sure, I've now refactored it to use `loss_and_grads` and `differentiable_loss_particle`. I'm not sure exactly how similar they should be though - e.g. `CSIS.loss_and_grads` still takes in an argument specifying whether to take gradients, unlike `Trace_ELBO.loss_and_grads`. Is this a problem?

pyro/infer/csis.py
```diff
@@ -0,0 +1,161 @@
...
+            batch_size = self.training_batch_size
+        else:
+            batch_size = len(batch)
+
+        loss = 0
+        for model_trace in batch:
+            guide_trace = self._get_matched_trace(model_trace)
+            particle_loss = -guide_trace.log_prob_sum() / batch_size
+            if grads:
+                torch_backward(particle_loss)
```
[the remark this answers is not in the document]
wsgharvey: I'm not quite sure how to do this. Are you suggesting using `torch.autograd.grad` instead of `torch.autograd.backward`?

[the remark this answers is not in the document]
eb8680: If you grab the guide parameters from `guide_trace` and do something like this, you won't accumulate gradients in model parameters:
```python
guide_grads = torch.autograd.grad(particle_loss, guide_params)
for guide_grad, guide_param in zip(guide_grads, guide_params):
    guide_param.grad = guide_param.grad + guide_grad if guide_param.grad else guide_grad
```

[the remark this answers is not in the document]
wsgharvey: I see, thanks!

wsgharvey:
Sure, I'll add the example now

ngoodman:
Btw it’d be interesting to test this against the auto guides we currently have available in Pyro!

PR #1441, opened 2018-10-13
Update importance.py (new features) 

robsalomone:
Added methods for the Importance class that are often used with importance sampling methods. Useful on their own, and for future features such as Sequential Monte Carlo and Annealed Importance Sampling.

In all cases they are computed in a numerically stable manner using the log scale.

1. Estimate the normalizing constant of the target distribution. (this assumes that the log_weights are computed using the normalized density of the guide, otherwise it is an estimator of the ratio of normalizing constants). This is also used at each stage in Sequential Monte Carlo to construct the final normalizing constant estimator, likewise for Annealed Importance Sampling.

2. Compute the normalized sample weights.

3. Compute the (Importance Sampling) effective sample size (which is different to the one used for MCMC). This is useful as a diagnostic for the efficacy of an importance sampling density. It is also used in Sequential Monte Carlo algorithms to determine resampling times.

For further background and an example of where such things are used, see for example: https://www.stats.ox.ac.uk/~doucet/delmoral_doucet_jasra_sequentialmontecarlosamplersJRSSB.pdf
which discusses these aspects in the context of SMC Samplers.

### DIFF
--- a/pyro/infer/importance.py
+++ b/pyro/infer/importance.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
-
+import torch
+import warnings
 import logging
 
 import pyro.poutine as poutine
@@ -7,7 +8,6 @@ import pyro.poutine as poutine
 from .abstract_infer import TracePosterior
 
 logger = logging.getLogger(__name__)
-
 
 class Importance(TracePosterior):
     """
@@ -44,3 +44,43 @@ class Importance(TracePosterior):
                 poutine.replay(self.model, trace=guide_trace)).get_trace(*args, **kwargs)
             log_weight = model_trace.log_prob_sum() - guide_trace.log_prob_sum()
             yield (model_trace, log_weight)
+
+    def get_log_normalizer(self):
+        """
+        Estimator of the normalizing constant of the target distribution. 
+        (mean of the unnormalized weights)
+        """
+        log_w = torch.tensor(self.log_weights)
+        log_num_samples = torch.log(torch.tensor(self.num_samples, dtype=torch.float))
+        return torch.logsumexp(log_w - log_num_samples,0)
+    
+    def get_normalized_weights(self, log_scale=False):
+        """
+        Compute the normalized importance weights. 
+        """
+        log_w = torch.tensor(self.log_weights)
+        log_w_norm = log_w - torch.logsumexp(log_w,0)
+        return log_w_norm if log_scale else torch.exp(log_w_norm)
+   
+    def get_ESS(self): 
+        """
+        Compute (Importance Sampling) Effective Sample Size (ESS). 
+        """ 
+        # check to make sure list is not empty
+        if self.log_weights:
+            log_w_norm = self.get_normalized_weights(log_scale=True)
+            ess = torch.exp(-torch.logsumexp(2*log_w_norm,0))
+        else:
+            logger.warn("The log_weights list is empty, effective sample size is zero.")
+            ess = 0 
+        return ess
+        
+        # tests to make sure it matches with simpler (but 
+        # numerically unstable) ways of computing.
+        # Can be removed. 
+       
+        # print("ESS (log-scale computation (main)) {:3f}".format(ess))
+        # ess_check1 = (torch.sum(torch.exp(log_w))**2) / (torch.sum(torch.exp(2*log_w)))
+        # print("ESS (semi-log-scale computation) {:3f}".format(ess_check1))
+        # ess_check2 = (torch.sum(torch.exp(log_w))**2) / (torch.sum(torch.exp(2*log_w)))
+        # print("ESS (numerically unstable computation) {:3f}".format(ess_check2))
--- a/pyro/infer/importance.py
+++ b/pyro/infer/importance.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import, division, print_function
 import torch
-import warnings
 import logging
 
 import pyro.poutine as poutine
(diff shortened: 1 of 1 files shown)


pyro/infer/importance.py
```diff
@@ -44,3 +44,33 @@ def _traces(self, *args, **kwargs):
...
             log_weight = model_trace.log_prob_sum() - guide_trace.log_prob_sum()
             yield (model_trace, log_weight)
+
+    def get_log_normalizer(self):
+        """
+        Estimator of the normalizing constant of the target distribution.
+        (mean of the unnormalized weights)
+        """
+        log_w = torch.tensor(self.log_weights)
+        log_num_samples = torch.log(torch.tensor(self.num_samples, dtype=torch.float))
```
jpchen: this `dtype` be the type of `log_w`

pyro/infer/importance.py
```diff
@@ -44,3 +44,33 @@ def _traces(self, *args, **kwargs):
...
+    def get_ESS(self):
+        """
+        Compute (Importance Sampling) Effective Sample Size (ESS).
+        """
+        # check to make sure list is not empty
+        if self.log_weights:
+            log_w_norm = self.get_normalized_weights(log_scale=True)
+            ess = torch.exp(-torch.logsumexp(2*log_w_norm, 0))
+        else:
+            logger.warn("The log_weights list is empty, effective sample size is zero.")
```
jpchen: can you `import warnings` up top and use `warnings.warn()` here? (and also above though not in the scope of this PR)

pyro/infer/importance.py
```diff
@@ -44,3 +44,33 @@ def _traces(self, *args, **kwargs):
...
+        log_w = torch.tensor(self.log_weights)
+        log_num_samples = torch.log(torch.tensor(self.num_samples, dtype=torch.float))
+        return torch.logsumexp(log_w - log_num_samples, 0)
+
+    def get_normalized_weights(self, log_scale=False):
+        """
+        Compute the normalized importance weights.
+        """
+        log_w = torch.tensor(self.log_weights)
+        log_w_norm = log_w - torch.logsumexp(log_w, 0)
```
jpchen: i think you might need to use the `logsumexp` in `pyro.distributions.util` for backwards compatibility

pyro/infer/importance.py
```diff
@@ -44,3 +44,33 @@ def _traces(self, *args, **kwargs):
...
+        return log_w_norm if log_scale else torch.exp(log_w_norm)
+
+    def get_ESS(self):
+        """
+        Compute (Importance Sampling) Effective Sample Size (ESS).
+        """
+        # check to make sure list is not empty
+        if self.log_weights:
+            log_w_norm = self.get_normalized_weights(log_scale=True)
+            ess = torch.exp(-torch.logsumexp(2*log_w_norm, 0))
```
jpchen: same comment

pyro/infer/importance.py
robsalomone: Sure, will make the change.

pyro/infer/importance.py
robsalomone: Sure. Will also fix the other one.

pyro/infer/importance.py
robsalomone: Can you please elaborate? 

self.num_samples is an int by default; log_w is a tensor of floats.

jpchen: i was thinking if default_tensor_type is `torch.DoubleTensor` then `log_w` will be a `torch.DoubleTensor`, but `log_num_samples` is a `torch.FloatTensor` so the following line will throw an error like `Expected object of type torch.DoubleTensor but found type torch.FloatTensor`.

robsalomone: It isn't throwing an error, but its an easy fix. Will change it to be the same type.

jpchen: have you tried changing the default tensor type? on my system (pytorch 0.4.0) i get an error:

```
In [3]: torch.set_default_tensor_type('torch.DoubleTensor')

In [4]: log_w = torch.tensor([0.1])

In [5]: log_num_samples = torch.log(torch.tensor(2, dtype=torch.float))

In [6]: log_w - log_num_samples
<ipython-input-6-1e3c76f98fb1> in <module>()
----> 1 log_w - log_num_samples

RuntimeError: Expected object of type torch.DoubleTensor but found type torch.FloatTensor for argument #3 'other'
```

pyro/infer/importance.py
robsalomone: This is an easy fix. However, I just noticed that pyro.distributions.util just calls torch.logsumexp; are you sure the change is needed?

pyro/infer/importance.py
robsalomone: I'll look into it.

pyro/infer/importance.py
jpchen: note the `except: ...` for earlier versions, torch doesnt support `logsumexp` which is why there's a hand implementation underneath :)

pyro/infer/importance.py
robsalomone: I tested with PyTorch 1.0.0rc1 and didn't get any issues, 0.4.0 must be less forgiving with types. 

Anyways, this was easily fixed to be the default tensor type by simply multiplying num_samples by 1. in the latest commit.

pyro/infer/importance.py
robsalomone: Odd, I do not see the "except:" in my utils.py, only the following: 

```

import numbers
from contextlib import contextmanager

import torch
import torch.distributions as torch_dist
from torch import logsumexp
from torch.distributions.utils import broadcast_all


_VALIDATION_ENABLED = False

log_sum_exp = logsumexp  # DEPRECATED
```

In any case, now logsumexp from this pyro.distributions.util is used.

PR #1442, opened 2018-10-14
Add mass matrix adaptation for HMC

fehiepsi:
This pull request is based on https://github.com/uber/pyro/pull/1178, which has been inactive for a while. So credit is given to @LoganWalls.

The design choice is based on many discussions at https://github.com/uber/pyro/issues/1137, https://github.com/uber/pyro/pull/1178, and https://github.com/uber/pyro/pull/1203.

In summary, I added comments to explain reasons for some parameters and made the following changes:
+ Add `inverse_mass_matrix` argument to `velocity_verlet`, which is used together with a `_kinetic_grad` function to find derivative of kinetic energy w.r.t. r. Because this is a full matrix/vector, we have to pack/unpack r dict while computing.
+ Add window adaptation scheme: heavy computation will happen only at the end of each window (from Stan). More on this is explained in the implementation of HMC.
+ `r_dist` is no longer a dictionary of distributions. Now, it is a distribution (Normal if full_mass = False, MVN if full_mass = True).
+ DualAveraging will be reset at the end of each window adaptation (from Stan).

Where to pack the dicts:
+ Pack z to feed into Welford scheme.
+ Pack r to calculate kinetic energy.
+ Pack r to calculate derivative of kinetic energy (in integrator module).

Where to unpack the tensors:
+ Unpack r_grads_flat after calculating derivative of kinetic energy, for the purpose of updating z.
+ Unpack r_flat from a sample of r_dist.

Where we do heavy computation:
+ When sampling momentum r from the inverse of mass matrix (which we obtained from Welford scheme), we have to do the combo inverse+Cholesky. These heavy computations are just happened at the end of each window adaptation (during warmup phases).

TODO:
- [x] Change window sizes when num_steps are small.
- [x] Add test and fix bugs.
- [x] Rebase the branch when https://github.com/uber/pyro/pull/1431 is merged.

Blocked by https://github.com/uber/pyro/pull/1431.


eb8680:
@fehiepsi can you change the base branch to `pytorch-1.0` for easier reviewing?

fehiepsi:
@eb8680 First time to know that I can do it. Thanks! ^^

eb8680:
> it will be great to get this merged into dev sooner, so that we can keep dev and pytorch-1.0 in sync as much as possible.

In that case, @fehiepsi if you don't need the changes in `pytorch-1.0` can you remove them from your branch and rebase against `dev`? Sorry for the confusion

fehiepsi:
It is good to know that there is no problem with pytorch 1.0rc. I will make the change soon.

neerajprad:
@fehiepsi - It will be great if you could also post some numbers on how much longer our unit tests / examples take with mass matrix adaptation enabled.

fehiepsi:
@neerajprad FYI, with `test_logistic_regression`, I get (running 2 times)
No adapt: 46s
Adapt step_size: 16s
Adapt step_size + mass_matrix diag: 14s
Adapt step_size + mass_matrix full: 13s

I think that there is no regression in the implementation. Both packing and unpacking are pretty fast comparing to other computations such as log_prog, grad,...

pyro/infer/mcmc/hmc.py
```diff
@@ -188,20 +215,108 @@ def _find_reasonable_step_size(self, z):
...
+                self._step_size_adapt_scheme.reset()
+
+            if self.adapt_mass_matrix:
+                self._inverse_mass_matrix = self._mass_matrix_adapt_scheme.get_covariance()
+                self._update_r_dist()
+                self._mass_matrix_adapt_scheme.reset()
+
+            self._adapt_window = 2 * self._adapt_window
+            if self._adapt_window_ending == self._adapt_mass_matrix_phase_ending:
+                self._adapt_mass_matrix_phase_ending = False
```
neerajprad: It seems like this is wrongly being set to a boolean type here, or it should be `self._adapt_mass_matrix_phase`?

pyro/infer/mcmc/hmc.py
```diff
@@ -233,29 +349,18 @@ def setup(self, *args, **kwargs):
...
-            _, log_step_size_avg = self._adapted_scheme.get_state()
-            self.step_size = math.exp(log_step_size_avg)
-            self.num_steps = max(1, int(self.trajectory_length / self.step_size))
+        self._validate_trace(trace)
+        mass_matrix_size = sum(self._r_numels.values())
+        if self.full_mass:
+            self._inverse_mass_matrix = eye_like(site_value, mass_matrix_size)
+        else:
+            self._inverse_mass_matrix = site_value.new_ones(mass_matrix_size)
+        self._update_r_dist()
```
neerajprad: Since we are assuming an identity mass matrix by default, would it simplify the code if we used that as the argument (`mass_matrix_structure="identity"`), instead of maintaining two boolean args `adapt_mass_matrix` and `full_max`?

neerajprad:
> I think that there is no regression in the implementation.

In that case, can we set the default to be diagonal mass matrix? If there are any perf issues or other corner cases, we should defer this to another PR. How is this playing with the step size adaptation?

pyro/infer/mcmc/hmc.py
fehiepsi: Ah, nice catch, thanks!

pyro/infer/mcmc/hmc.py
fehiepsi: I think that maintaining `adapt_mass_matrix` has longer benefit. Later, we can let users set initial `inverse_mass_matrix` and decide if they want to adapt it or not.

pyro/infer/mcmc/mcmc.py
```diff
@@ -22,10 +22,10 @@ class MCMC(TracePosterior):
         during the warmup phase are discarded.
     """
 
-    def __init__(self, kernel, num_samples, warmup_steps=0):
+    def __init__(self, kernel, num_samples, warmup_steps=None):
         self.kernel = kernel
-        self.warmup_steps = warmup_steps
         self.num_samples = num_samples
+        self.warmup_steps = warmup_steps if warmup_steps is not None else num_samples // 2  # Stan
```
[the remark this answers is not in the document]
fehiepsi: Do you want to move `num_samples` too? (the initialization of warmup_steps depends on num_samples by default)

fehiepsi:
I like the idea of setting the defaults to adapt stepsize+diagonal (so default parameters will be the same as Stan). I can't do throughout experiments to decide how adapt_mass_matrix and adapt_step_size play together. In Stan, both of them are turned on or turned off at the same time (they just have a parameter `adapt_engaged` to decide if we want to adapt or not).

Btw, do you observe that tqdm's progressbar floats whole screen (it creates new bar for each iteration)?

neerajprad:
> Edit: Never mind, it works on normal terminal (but not on tmux)

Will take a look at it. The only time I have seen it doing that behavior is when the screen width is too small, which I have tried to fix in #1443.

pyro/infer/mcmc/mcmc.py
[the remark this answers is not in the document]
neerajprad: Hmm..I think let uss keep it here only for now then, and just document the default initialization in the docstring.

neerajprad:
> I like the idea of setting the defaults to adapt stepsize+diagonal (so default parameters will be the same as Stan) so I go ahead and enable them by default. I can't do throughout experiments to decide how adapt_mass_matrix and adapt_step_size play together. In Stan, both of them are turned on or turned off at the same time (they just have a parameter adapt_engaged to decide if we want to adapt or not).

I think leaving it on by default is a good idea. We should add a test case where just using NUTS with the default parameters (i.e. mass matrix adaptation enabled) gives better results than no mass matrix adaptation - say in the case of a highly correlated gaussian. We could even just test the mass matrix estimation with "full" vs "diagonal" covariance given enough number of warmup steps.

neerajprad:
@fehiepsi - Would this interact with parallel chaining? My understanding is that each chain should run its own adaptation scheme, but I am curious to know if other packages handle this differently, i.e. use the warmup iterations across all chains to decide on the step size or mass matrix. Any ideas how Stan does this?

pyro/infer/mcmc/nuts.py
```diff
@@ -250,8 +258,7 @@ def sample(self, trace):
 
         # Temporarily disable distributions args checking as
         # NaNs are expected during step size adaptation.
-        dist_arg_check = False if self._adapt_phase else pyro.distributions.is_validation_enabled()
-        with dist.validation_enabled(dist_arg_check):
+        with optional(pyro.validation_enabled(False), self._adapt_phase):
```
fehiepsi: related https://github.com/uber/pyro/pull/1196

fehiepsi:
@neerajprad I will check and get back to you about chain soon. For a test case to check performance of adapting, we might create a linear regression model of two variables y = ax + b, where a, b are two latent variables. Assume that x's values are around 100, then a and b should be highly negative correlated (a small change in `a` will lead to a big change in `b` in opposite direction). Do you think that it makes sense?

fehiepsi:
> > Edit: Never mind, it works on normal terminal (but not on tmux)
> 
> Will take a look at it. The only time I have seen this behavior is when the screen width is too small, which I have tried to fix in #1443.

Yup, that's the reason. I enlarged the tmux window to capture full message of an iteration, then it behaved correctly.

neerajprad:
> Yup, that's the reason. I enlarged the tmux window to capture full message of an iteration, then it behaved correctly.

Even on very narrow width, we should just eliminate the progress bar, rather than having this jumpy output. I will send a fix separately for this.

fehiepsi:
> Would this interact with parallel chaining? My understanding is that each chain should run its own adaptation scheme, but I am curious to know if other packages handle this differently, i.e. use the warmup iterations across all chains to decide on the step size or mass matrix. Any ideas how Stan does this?

From their Reference Manual and discussions around web, it seems that each chain runs its own adaptation scheme. :)

neerajprad:
> For a test case to check performance of adapting, we might create a linear regression model of two variables y = ax + b, where a, b are two latent variables. Assume that x's values are around 100, then a and b should be highly negative correlated (a small change in a will lead to a big change in b in opposite direction). Do you think that it makes sense?

That sounds reasonable, though I am not sure how much of a problem it will be in a low dimensional setting - curious to see what you get.

fehiepsi:
@neerajprad I just decided to try it out on baseball example. It works well btw. :)

tests/infer/mcmc/test_nuts.py
```diff
@@ -80,13 +90,23 @@ def model(data):
...
 
-    nuts_kernel = NUTS(model, step_size=0.0855)
+    nuts_kernel = NUTS(model, step_size, adapt_step_size, adapt_mass_matrix, full_mass)
     mcmc_run = MCMC(nuts_kernel, num_samples=500, warmup_steps=100).run(data)
     posterior = EmpiricalMarginal(mcmc_run, sites='beta')
     assert_equal(rmse(true_coefs, posterior.mean).item(), 0.0, prec=0.1)
 
 
-def test_beta_bernoulli():
+@pytest.mark.parametrize(
```
neerajprad: Our HMC tests are consuming more test time than earlier, see below for some of our most expensive unit tests. Should we just use the default parameters (i.e. adaptation = True) on most of these tests, and only use the full parameterized test on one example each in HMC and NUTS? I think that should give us a good balance between coverage and test times, at least until we make HMC faster. :) 

```python
136.68s call     tests/infer/mcmc/test_nuts.py::test_beta_bernoulli[0.02-False-True-False]
107.39s call     tests/infer/mcmc/test_nuts.py::test_gamma_normal[0.01-False-True-False]
51.55s call     tests/infer/mcmc/test_nuts.py::test_nuts_conjugate_gaussian[dim=10_chain-len=4_num_obs=1]
40.83s call     tests/infer/mcmc/test_nuts.py::test_logistic_regression[0.0855-False-True-False]
39.48s call     tests/infer/mcmc/test_hmc.py::test_beta_bernoulli[None-1-None-True-False-False]
29.94s call     tests/infer/test_abstract_infer.py::test_posterior_predictive
29.25s call     tests/infer/mcmc/test_hmc.py::test_gamma_normal[None-1-None-True-False-False]
24.20s call     tests/infer/mcmc/test_hmc.py::test_logistic_regression[None-1-None-True-False-False]
23.35s call     tests/infer/mcmc/test_hmc.py::test_bernoulli_latent_model[True]
23.29s call     tests/infer/mcmc/test_nuts.py::test_beta_bernoulli[0.02-False-False-False]
22.85s call     tests/infer/mcmc/test_nuts.py::test_bernoulli_latent_model
21.22s call     tests/infer/mcmc/test_nuts.py::test_nuts_conjugate_gaussian[dim=10_chain-len=3_num_obs=1]
20.41s call     tests/infer/mcmc/test_nuts.py::test_gamma_beta
19.91s call     tests/infer/mcmc/test_hmc.py::test_hmc_conjugate_gaussian[dim=10_chain-len=4_num_obs=1]
19.60s call     tests/infer/mcmc/test_hmc.py::test_dirichlet_categorical[None-1-None-True-False-False]
```

pyro/infer/mcmc/hmc.py
```diff
@@ -188,20 +219,109 @@ def _find_reasonable_step_size(self, z):
...
+            direction_new = 1 if self._direction_threshold < -delta_energy else -1
         return step_size
 
+    def _configure_adaptation(self, trace):
+        self._adapt_phase = True
+        self._adapt_window = self._adapt_initial_window
+
+        # from Stan, for small warmup_steps
+        if self._warmup_steps < 20:
+            self._adapt_start_buffer = self._warmup_steps
```
neerajprad: Since we are basically not doing mass matrix adaptation when warmup steps is smaller than this, can we `return` early? This is more from the point of code readability, since attributes like `self._adapt_mass_matrix_phase_ending` will have wrongly set values (won't be used though). 

There is also a lot of state setting that is happening in this part, which might be tougher to maintain / refactor in the future. We should think about ways this can be refactored - say by constructing an adaptation schedule via `build_adaptation_schedule(warmup_steps)` which depending on the options provided returns something like `[(10, "start_buffer"), (20, "window_1"), (40, "window_2")..]` etc. Then we can peek / pop off the items from this queue and do the required operations. We can defer this to a separate refactoring PR though.

pyro/infer/mcmc/trace_kernel.py
```diff
@@ -8,11 +8,12 @@
 @add_metaclass(ABCMeta)
 class TraceKernel(object):
 
-    def setup(self, *args, **kwargs):
+    def setup(self, warmup_steps=None, *args, **kwargs):
```
neerajprad: Remove `=None`.

pyro/infer/mcmc/hmc.py
fehiepsi: Yup, I have thought about this for a while but did not come up with a good solution though. :(

pyro/infer/mcmc/nuts.py
neerajprad: Nice!

pyro/infer/mcmc/hmc.py
neerajprad: Np at all! It is working great, we can refactor to make things less mutable in a later PR.

fehiepsi:
@neerajprad Thanks for reviewing! This is ready when tests pass.

Issue #1093, opened 2018-04-24
Future areas of work / improvement for HMC and NUTS


neerajprad:
Please create a separate issue if you are working on a major task (e.g. mass matrix adaptation, or parallel chaining), so that all task specific discussion is contained within that issue.

### Enhancements

**Minor:**
- [x] Deciding on a suitable number of warmup iterations. These are currently taken as input arguments, but `adapt_step_size=True` should have a reasonable default number of warmup iterations, if not specified by the user. e.g. we could default to 50% (as Stan), in which case if `num_samples=100` then we would automatically run 100 warmup iterations if none are specified by the user.
- [x] Ability to set target acceptance probability during step size adaptation - currently fixed at `0.8`. This will be specially useful to bias the adaptation towards smaller step sizes to explore problematic posteriors with regions of high curvature.
- [x] Ability to set max_tree_depth to trade off accuracy for speed.
- [ ] Detect divergent transitions, and log them for analysis. Throwing a `NaN` here during sampling as we might do now with validation check enabled, is not very useful to the end user.
- [x] Implement a progress bar so that the logging info does not clutter the screen (specially in notebooks).

**Major:**
 - [x] Multiple parallel chains for HMC (and/or NUTS). For HMC, this could be as simple as using `poutine.broadcast` to run parallel chains, similar in spirit to parallelizing ELBO computation over `num_particles` in #1176, or (preferably) use `torch.distributed` to implement a more general (applicable to NUTS) and scalable solution.
- [x] Adapt the mass matrix during the warmup phase, alongside the step size parameter. Currently this is assumed to be a diagnormal.
- [x] Better initialization strategies, ~e.g. generating the initial trace after running ADVI, MAP~ (EDIT: this can be done by the user independently and the trace so generated can be specified via `initial_trace`). In addition, providing the option to the user to specify an initial trace to the NUTS/HMC kernel.
- [x] Use multinomial instead of slice sampling in NUTS.
- [x] Enumerate over discrete latents. #1128 
- [ ] (Low priority) Add support for other MCMC algorithms like Gibbs sampling, and use these in conjunction with the HMC/NUTS kernel to allow sampling from models with discrete latent variables.
- [ ] Parallel chains on CUDA. **NOTE**: This feature is partially supported. We need to hold traces in the workers until it terminates or the main process does not need them any more. But this way will somehow violate the main reason for using generator: to resolve memory issue for large model. A better mechanism for when to store traces, when to clear it in workers should be implemented.

[...]

PR #1443, opened 2018-10-15
Support parallel chains in HMC

neerajprad:
Addresses #1093. 

This uses the python `multiprocessing` library to spin up separate python interpreters each running its own MCMC chain in parallel. I had to make some modifications to the `logging` module to make it work with parallel runners by using a logging handler that uses a `multiprocess.Queue` for logging events. On windows (also when we want to use CUDA), only the "spawn" context is available (no "fork"), which puts certain additional constraints on both the internals as well as the model callable, in that a lot of the code that is exercised needs to be picklable.

Some notes:
 - There is no change to the external interface. Multiprocessing with parallel workers will only kick in when `num_chains > 1`. We should be able to swap out the internals later to use tensor based parallelization, for instance.
 - The majority of change is restricted to the external `MCMC` wrapper. The only internal change is that the poutine wrapping of the model (`poutine.broadcast` or `poutine.enum`) happens in the setup method instead of the constructor. The reason for this is that our poutine code uses a lot of closures which makes it difficult to pickle poutine wrapped models. By moving this to `setup`, the poutine wrapping happens inside each of the processes so as not to be an issue. Once we make our handlers picklable, poutine wrapped models can be sent directly by the user.
 - Logging is handled asynchronously using a separate thread that consumes events from the logging `queue` into which each of the workers put in diagnostics / other logging messages. I found this to be the only reliable way of getting sane logging output with processes running in parallel. Most of the changes in `logger.py` relate to this change.

e.g. invocation (8 schools example)
```bash
   [pytorch-36] npradhan@npradhan:~/workspace/pyro_dev/pyro (hmc-chain-tqdm)
  $ python examples/eight_schools/mcmc.py --num-chains 4
Warmup [0]:  24%|██████▊                     | 305/1250 [00:07<00:23, 41.01it/s, Step size=0.435, Acceptance rate=0.941]
Warmup [1]:  24%|██████▌                     | 294/1250 [00:07<00:22, 42.32it/s, Step size=0.401, Acceptance rate=0.946]
Warmup [2]:  27%|███████▊                     | 337/1250 [00:07<00:17, 51.50it/s, Step size=0.59, Acceptance rate=0.947]
Warmup [3]:  25%|███████                     | 318/1250 [00:07<00:19, 47.73it/s, Step size=0.402, Acceptance rate=0.931]
```

#### TODOS:
 - Check execution on:
   - [x] Windows, Python 2/3 
   - [x] Mac OS X, Python 2/3
   - [x] Linux, Python 2/3
 - [ ] Check if CUDA backend works with `spawn` context - will need to redesigned due to https://github.com/pytorch/pytorch/issues/10375
 - [x] Deterministic (different for each chain) random number seeding w.r.t. single chain.
 - [x] Clean exit of child processes with KeyboardInterrupt.
 - [x] Close all processes and cleanup if there is an exception in one of the child processes.
 - [x] Pickling issue on dynamically generated torch distributions - #1456 

#### Later PR:
 - [ ] Chain level stats / diagnostics.


fritzo:
@eb8680 it looks like `functools.partial(my_poutine, **kwargs)` is pickleable, whereas the closure `lambda f: my_poutine(f, **kwargs)` is not pickleable. Can we simply switch from closures to `functools.partial` in our poutines?

neerajprad:
As I discussed with @eb8680, I have worked around the poutine pickling issue, so I am not currently blocked on that. It will be great to have `Messenger` and handlers like `condition`, `enum` etc. be picklable eventually, so that poutine wrapped models can be sent as args directly.

neerajprad:
I am looking at some cleanup related issue in linux, but this should be good to review now. I will make changes for CUDA in a separate PR.

neerajprad:
@eb8680 - Could you help test the baseball example on windows (just running `python examples/baseball.py`)? Multiprocessing is quite different in windows, so just wanted to make sure that we know of any issues beforehand.

pyro/infer/mcmc/logger.py
```diff
@@ -93,25 +172,35 @@ def initialize_progbar(warmup_steps, num_samples, min_width=100, max_width=120,
                         position=pos, file=sys.stderr, disable=disable)
 
     if getattr(progress_bar, "ncols", None) is not None:
-        progress_bar.ncols = min(min_width, progress_bar.ncols)
-        progress_bar.ncols = max(max_width, progress_bar.ncols)
+        progress_bar.ncols = max(min_width, progress_bar.ncols)
+        progress_bar.ncols = min(max_width, progress_bar.ncols)
```
neerajprad: This was just wrong earlier, so just correcting that.

eb8680:
> Could you help test the baseball example on windows

Seems to work correctly for me, at least with Python 3.

neerajprad:
> I'd recommend adding a disclaimer somewhere about a relative lack of windows testing.

Thanks, @eb8680. I will add that disclaimer. KeyboardInterrupt should result in exiting from the main thread very fast, and we shouldn't be leaving orphan processes behind (it might take a few seconds for them to be cleaned up though). I have tested that the cleanup happens as expected in OS X, and will run a few tests on linux too.

neerajprad:
Thanks for reviewing! This should be good to merge.

PR #1446, opened 2018-10-15
Allow other operation labels in Poutine

eb8680:
This PR simplifies and generalizes the dispatch mechanism in `poutine.Messenger` to allow handlers for other types of operations (beyond `"sample"` and `"param"`).  In particular:
1. It defines a wrapper function `poutine.runtime.effectful` and uses it to implement `pyro.param`.  `effectful` takes a type label (e.g. `"sample"`) and a callable.
2. `poutine.runtime.apply_stack` and `poutine.Messenger` use the type label passed to `effectful`, which is stored in `msg["type"]` to dispatch to the appropriate operation: `getattr(self, "_pyro_{}".format(msg["type"]))(msg)`

I was using this elsewhere to experiment with lazy evaluation and dependency tracking.


pyro/poutine/messenger.py
```diff
@@ -114,13 +114,37 @@ def _process_message(self, msg):
         Process the message by calling appropriate method of itself based
         on message type. The message is updated in place.
         """
-        return getattr(self, "_pyro_{}".format(msg["type"]))(msg)
+        if hasattr(self, "_pyro_{}".format(msg["type"])):
```
fritzo: nit: avoid formatting the string multiple times
```py
method_name = "_pyro_{}".format(msg["type"])
if hasattr(self, method_name):
    return getattr(self, method_name)(msg)
```
or maybe
```py
return getattr(self, "_pyro_{}".format(msg["type"]), lambda msg: None)(msg)
```

pyro/poutine/messenger.py
```diff
@@ -114,13 +114,37 @@ def _process_message(self, msg):
...
+            return getattr(self, "_pyro_{}".format(msg["type"]))(msg)
+        return None
 
     def _postprocess_message(self, msg):
+        if hasattr(self, "_pyro_post_{}".format(msg["type"])):
```
fritzo: ditto

pyro/poutine/messenger.py
```diff
@@ -114,13 +114,37 @@ def _process_message(self, msg):
...
+        @SomeMessengerClass.register
+        def some_function(...)
+        """
+        if fn is None:
+            return lambda x: cls.register(x, type=type, post=post)
 
-    def _pyro_param(self, msg):
-        return None
+        if type is None:
+            type = fn.__code__.co_name
```
fritzo: What is the difference between `fn.__name__` and `fn.__code__.co_name`? Should I start avoiding `fn.__name__`?

pyro/poutine/runtime.py
```diff
@@ -196,3 +151,52 @@ def am_i_wrapped():
...
+
+    if getattr(fn, "__wrapped", None):
+        return fn
+    else:
+        if type is None:
+            type = fn.__code__.co_name
+
+        assert type != "message", "cannot use 'message' as keyword"
+
+        def _fn(*args, **kwargs):
```
fritzo: Do you know how expensive it is to create a closure each sample statement? I do not know, but I'm a little worried about our gradual increase of Python overhead over time. It is nice to find zero-abstraction-cost alternatives whenever possible.

pyro/poutine/runtime.py
```diff
@@ -196,3 +151,52 @@ def am_i_wrapped():
     :returns: bool
     """
     return len(_PYRO_STACK) > 0
+
+
+def effectful(fn=None, type=None):
```
fritzo: Can we memoize this function to reduce computational overhead of abstraction?

pyro/poutine/runtime.py
eb8680: Inner function overhead is discussed a little [here](https://stackoverflow.com/questions/7839632/is-there-an-overhead-when-nesting-functions-in-python), not sure about closures but I guess the cost is one extra variable lookup (a `LOAD_FAST` VM op) each for `fn` and `type` relative to the existing situation for `sample`.

I think most of our Python overhead comes from the dispatching (`getattr`) in `poutine.runtime.apply_stack`, and this is unlikely to be large in comparison after `effectful` is memoized.  I'll do some profiling when I get a chance.

pyro/poutine/messenger.py
eb8680: I think `fn.__code__.co_name` will refer to the name in the original code object, whereas `__name__` can refer to an intermediate variable name?  It's probably not a great idea to provide a default `type` anyway, I should just remove that.

pyro/poutine/runtime.py
eb8680: Hmm, memoization seems to be causing a memory leak in `ss_vae_M2.py`.

fritzo: Ok, we can avoid memoizing. I guess that would be premature optimization anyway.

pyro/poutine/runtime.py
eb8680: I've reverted the definition of `pyro.sample`, so that performance should be affected.  In the other use cases I imagine for this, `effectful` is only called once per function/operation definition, rather than once per function call.

eb8680:
Good to merge pending review.  There should be no performance impact.

pyro/poutine/runtime.py
```diff
@@ -196,3 +153,53 @@ def am_i_wrapped():
...
+    """
+    :param fn: function or callable that performs an effectful computation
+    :param str type: the type label of the operation, e.g. `"sample"`
+
+    Wrapper for calling :func:~`pyro.poutine.runtime.apply_stack` to apply any active effects.
+    """
+    if fn is None:
+        return functools.partial(effectful, type=type)
+
+    if getattr(fn, "__wrapped", None):
```
fritzo: nit: Could we name this `._is_effectful`?

pyro/poutine/runtime.py
```diff
@@ -196,3 +153,53 @@ def am_i_wrapped():
...
+        return fn
+    else:
```
fritzo: nit: no need for `else:` since you've returned early

pyro/poutine/runtime.py
```diff
@@ -196,3 +153,53 @@ def am_i_wrapped():
...
+        assert type is not None, "must provide a type label for operation {}".format(fn)
+        assert type != "message", "cannot use 'message' as keyword"
+
+        def _fn(*args, **kwargs):
```
fritzo: 1. Consider using `@functools.wraps(fn)`
2. Would it be reasonable to pull out `name` from `*args`? I.e. are all foreseeable statements named? If so, then we could define exactly `param = effectful(_PYRO_PARAM_STORE.get_param, type="param")`.

pyro/poutine/messenger.py
```diff
@@ -114,13 +114,70 @@ def _process_message(self, msg):
...
+            return lambda x: cls.register(x, type=type, post=post)
+
+        if type is None:
+            raise ValueError("An operation type name must be provided")
+
+        setattr(cls, "_pyro_" + ("post_" if post else "") + type, staticmethod(fn))
+        return fn
+
+    @classmethod
+    def unregister(cls, fn=None, type=None, post=None):
```
fritzo: Consider removing the unused `post` kwarg?

pyro/poutine/runtime.py
eb8680: > I.e. are all foreseeable statements named?

No, actually I'd expect most other statements to not be named.

PR #1447, opened 2018-10-15
Add partial pooling (no logit) baseball example

fehiepsi:
The new adapt matrix scheme seems work well with baseball example. So I add a new model for this example which (according to @neerajprad ) has been non-stable previously. There might be some discussions around this which are different from HMC's implementation. So I make it separately from #1442.


examples/baseball.py
```diff
@@ -253,22 +272,38 @@ def main(args):
...
-    evaluate_log_predictive_density(partially_pooled, posterior_partially_pooled, baseball_dataset)
+    evaluate_log_predictive_density(partially_pooled_with_logit,
+                                    posterior_partially_pooled_with_logit, baseball_dataset)
 
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Baseball batting average using HMC")
-    parser.add_argument("-n", "--num-samples", nargs="?", default=1200, type=int)
-    parser.add_argument("--warmup-steps", nargs='?', default=300, type=int)
+    parser.add_argument("-n", "--num-samples", nargs="?", default=200, type=int)
```
neerajprad: I am guessing this was changed for testing? Should we revert the original defaults back.

neerajprad:
Merging this so that I can pull in these changes in #1443.

fehiepsi:
I just wondered if we really need high num_samples and warmup_steps for
this small model. I tested the models with these default numbers and it
gave results aligned with the tutorial in Stan. With the old numbers, the
partial_pooling_with_logit model gives small step_size after warmup, which
in turns makes NUTS runs slowly. Maybe I keep num-samples and decrease
warmup-steps to 200?

neerajprad:
> With the old numbers, the partial_pooling_with_logit model gives small step_size after warmup, which
in turns makes NUTS runs slowly. 

On that model, I have seen some chains get stuck making no progress due to having a large step size (all proposals getting rejected). So I suppose, having a smaller step size is desirable to explore most of the posterior mass, though it might be suboptimal for other regions where a larger step size will suffice. I will try to increase it with parallel chaining and even if a few chains are slower than others, it should still be faster than a single chain.

fehiepsi:
Oh, maybe with mass matrix adaptation, things are better. I don't face such issue yet.

PR #1450, opened 2018-10-16
Resolve the problem of Renyi elbo with reject sampling

fehiepsi:
In [this paper](https://arxiv.org/abs/1806.01851) by @martinjankowiak and @fritzo, they mentioned that we can drop score function term at the cost of small bias. So I go ahead and drop it in the implementation of Renyi elbo. The [test with ShapeAugmentedGamma](https://github.com/uber/pyro/blob/dev/tests/infer/test_inference.py#L228) passes now.

### DIFF
--- a/pyro/infer/renyi_elbo.py
+++ b/pyro/infer/renyi_elbo.py
@@ -161,13 +161,11 @@ class RenyiELBO(ELBO):
                     elbo_particle = elbo_particle - log_prob_sum.detach()
 
                     if not is_identically_zero(entropy_term):
+                        # XXX for Rejector distribution which both entropy_term and
+                        # score_function_term are not zero, we can drop score_function_term
+                        # at the cost of small bias (https://arxiv.org/abs/1806.01851)
                         surrogate_elbo_particle = surrogate_elbo_particle - log_prob_sum
-
-                        if not is_identically_zero(score_function_term):
-                            # link to the issue: https://github.com/uber/pyro/issues/1222
-                            raise NotImplementedError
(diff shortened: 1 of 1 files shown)


martinjankowiak:
@fehiepsi i'm a bit confused about the purpose of this PR. if you're using a rejector type distribution (which you don't explicitly check for in the code) it's true that one can often drop the score function term at the cost of small bias, but at least for the rejector type distributions that we have implemented, we _also_ have fully pathwise distributions implemented. if someone were to try renyi + discrete random variables they'll now get incorrect results, no? (instead of `NotImplementedError`)

fehiepsi:
@martinjankowiak I don't know what is going on with discrete variables so I can't say. I believe you know much better than me about these stuffs so I will close this pull request. :)

fehiepsi:
@martinjankowiak I re-open this pull request for further discussion. In the first pull request for Renyi, I wrote a derivation for non-reparameterizable variable [here](https://gist.github.com/fehiepsi/7916400d8380cb05d21e859681b7a694). In [test.inferece](https://github.com/uber/pyro/blob/dev/tests/infer/test_inference.py), we have various tests for non-reparameterizable. If I understand right, then for fully pathwise distributions, we just use reparameterizable version of Renyi. This pull request deals with the distributions which have both `score_function` and `entropy_term`. To my knowledge, only Rejectors have both terms nonzero and from the the paper of you and Fritz, it seems that we can omit `score_function` term in gradient estimation. That's why I made this pull request.

Issue #1458, opened 2018-10-17
Value sampled is not in distribution's support for baseball example


neerajprad:
The `partially_pooled` model in `baseball.py` seems to sporadically fail with the exception that the value argument is not within the support for the Pareto distribution. cc. @fehiepsi 

Taking a look at this, but I am temporarily removing this from `test_examples.py` to unblock the build.

```
Traceback (most recent call last):
  File "/home/travis/build/uber/pyro/pyro/poutine/trace_struct.py", line 128, in log_prob_sum
    log_p = site["log_prob_sum"]
KeyError: 'log_prob_sum'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/travis/build/uber/pyro/examples/baseball.py", line 309, in <module>
    main(args)
  File "/home/travis/build/uber/pyro/examples/baseball.py", line 272, in main
    .run(partially_pooled, at_bats, hits)
  File "/home/travis/build/uber/pyro/pyro/infer/abstract_infer.py", line 84, in run
    for tr, logit in self._traces(*args, **kwargs):
  File "/home/travis/build/uber/pyro/pyro/infer/mcmc/mcmc.py", line 51, in _traces
    for trace in self._gen_samples(self.num_samples, trace):
  File "/home/travis/build/uber/pyro/pyro/infer/mcmc/mcmc.py", line 35, in _gen_samples
    trace = self.kernel.sample(trace)
  File "/home/travis/build/uber/pyro/pyro/infer/mcmc/nuts.py", line 276, in sample
    direction, tree_depth, energy_current)
  File "/home/travis/build/uber/pyro/pyro/infer/mcmc/nuts.py", line 159, in _build_tree
    direction, tree_depth-1, energy_current)
  File "/home/travis/build/uber/pyro/pyro/infer/mcmc/nuts.py", line 178, in _build_tree
    direction, tree_depth-1, energy_current)
  File "/home/travis/build/uber/pyro/pyro/infer/mcmc/nuts.py", line 178, in _build_tree
    direction, tree_depth-1, energy_current)
  File "/home/travis/build/uber/pyro/pyro/infer/mcmc/nuts.py", line 155, in _build_tree
    return self._build_basetree(z, r, z_grads, log_slice, direction, energy_current)
  File "/home/travis/build/uber/pyro/pyro/infer/mcmc/nuts.py", line 130, in _build_basetree
[...]

PR #1459, opened 2018-10-17
Comment out baseball example from test_examples

neerajprad:
This is to resolve the build failure until #1458 is resolved.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ jobs:
         - python: 2.7
           env: STAGE=examples
           script:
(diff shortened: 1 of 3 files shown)


neerajprad:
> if everything is fixed why is this failure sporadic?

@jpchen - I forgot about the coverage test which fails if any of our examples aren't in `test_examples`. I am only bypassing the specific model that is sporadically failing. I'm not sure about the source of non determinism in that, because we are setting the random seed.

neerajprad:
Ready to merge.

PR #1462, opened 2018-10-17
Fix pickling of dynamically imported torch distributions

### LINKED ISSUE #1456: Make dynamic distribution mixin wrappers in torch.py picklable
We load torch distributions and create distribution mixin classes dynamically in `torch.py`. These distribution instances are not picklable however.

```python
>>> pickle.dumps(Pareto(1., 1.))
_pickle.PicklingError: Can't pickle <class 'pyro.distributions.torch._PyroDist'>: it's not the same object as pyro.distributions.torch._PyroDist
```
We may need to provide more help to `pickle` to be able pickle / unpickle these instances. Note that this is not currently an issue with most distribution classes since they have a small wrapper in Pyro (containing `.expand` method), but will be in `pyro==0.3` where we are constructing these wrappers automatically.

### Why do we need this?

This will be useful in the multiprocessing context where workers send back traces to the main process. All elements in the trace (inc. distribution instances) should be picklable in that case. An alternative design would be to not send back the full trace, but only the empirical distribution itself, but this is worth exploring only if the overhead of trace serialization/deserialization is considerable.

fritzo:
Fixes #1456 

Note that pickling`torch.Size()` requires pickle protocol >= 2.

`Gumbel` currently fails due to its use of an inverse transform. I believe this can be fixed upstream by adding `__getstate__` and `__setstate__` methods to `InverseTransform`.

## Tested

- Added a `pickle.dumps()` test for all Pyro distributions.

### DIFF
--- a/pyro/distributions/torch.py
+++ b/pyro/distributions/torch.py
@@ -313,11 +313,8 @@ for _name, _Dist in torch.distributions.__dict__.items():
     try:
         _PyroDist = locals()[_name]
     except KeyError:
-
-        class _PyroDist(_Dist, TorchDistributionMixin):
-            pass
-
-        _PyroDist.__name__ = _name
+        _PyroDist = type(_name, (_Dist, TorchDistributionMixin), {})
+        _PyroDist.__module__ = __name__
         locals()[_name] = _PyroDist
 
     _PyroDist.__doc__ = '''
--- a/tests/distributions/test_pickle.py
+++ b/tests/distributions/test_pickle.py
@@ -1,0 +1,80 @@
+from __future__ import absolute_import, division, print_function
+
+import inspect
+
+import pytest
+import six.moves.cPickle as pickle
+import torch
+
+import pyro.distributions as dist
+from pyro.distributions.torch_distribution import TorchDistributionMixin
+from tests.common import xfail_param
+
+# Collect distributions.
+BLACKLIST = [
+    dist.TorchDistribution,
+    dist.ExponentialFamily,
+    dist.OMTMultivariateNormal,
+]
+XFAIL = {
+    dist.Gumbel: xfail_param(dist.Gumbel, reason='cannot pickle weakref'),
+}
+DISTRIBUTIONS = [d for d in dist.__dict__.values()
+                 if isinstance(d, type)
+                 if issubclass(d, TorchDistributionMixin)
+                 if d not in BLACKLIST]
+DISTRIBUTIONS.sort(key=lambda d: d.__name__)
+DISTRIBUTIONS = [XFAIL.get(d, d) for d in DISTRIBUTIONS]
+
+# Provide default args if Dist(1, 1, ..., 1) is known to fail.
+ARGS = {
+    dist.AVFMultivariateNormal: [torch.zeros(3), torch.eye(3), torch.rand(2, 4, 3)],
+    dist.Bernoulli: [0.5],
+    dist.Binomial: [2, 0.5],
+    dist.Categorical: [torch.ones(2)],
+    dist.Delta: [torch.tensor(0.)],
+    dist.Dirichlet: [torch.ones(2)],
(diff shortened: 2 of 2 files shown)


tests/distributions/test_pickle.py
```diff
@@ -0,0 +1,80 @@
...
+            # Python 2.6-3.5
+            spec = inspect.getargspec(Dist.__init__)
+            nargs = len(spec.args) - 1 - (len(spec.defaults) if spec.defaults else 0)
+        args = (1,) * nargs
+    try:
+        dist = Dist(*args)
+    except:
+        pytest.skip(reason='cannot construct distribution')
+
+    # Note that pickling torch.Size() requires protocol >= 2
```
neerajprad: Surprisingly, I haven't found this to be an issue with multiprocessing in python 2.7, so my guess is that the ForkingPickler is doing something more than the default pickler that uses protocol 0.

pyro/distributions/torch.py
```diff
@@ -313,11 +313,8 @@ def expand(self, batch_shape):
     try:
         _PyroDist = locals()[_name]
     except KeyError:
-
-        class _PyroDist(_Dist, TorchDistributionMixin):
-            pass
-
-        _PyroDist.__name__ = _name
+        _PyroDist = type(_name, (_Dist, TorchDistributionMixin), {})
+        _PyroDist.__module__ = __name__
```
neerajprad: Wow..this was neat!

tests/distributions/test_pickle.py
```diff
@@ -0,0 +1,80 @@
...
+            spec = list(inspect.signature(Dist.__init__).parameters.values())
+            nargs = sum(1 for p in spec if p.default is p.empty) - 1
+        except AttributeError:
+            # Python 2.6-3.5
+            spec = inspect.getargspec(Dist.__init__)
+            nargs = len(spec.args) - 1 - (len(spec.defaults) if spec.defaults else 0)
+        args = (1,) * nargs
+    try:
+        dist = Dist(*args)
+    except:
```
neerajprad: You'll need to add `Exception` to make the linter happy.

tests/distributions/test_pickle.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+@pytest.mark.parametrize('Dist', DISTRIBUTIONS)
+def test_pickle(Dist):
+    if Dist in ARGS:
+        args = ARGS[Dist]
+    else:
+        # Optimistically try to initialize with Dist(1, 1, ..., 1).
+        try:
+            # Python 3.6+
```
jpchen: should i add python 3.6 to travis? (either in addition to or in lieu of 3.5). not sure which is more common among users, but i use 3.6.2 locally.

jpchen: new builds will, which is why im thinking of replacing 3.5 since apparently (from this PR) there are breaking changes 3.5 -> 3.6, which we generally wont catch

neerajprad: > should i add python 3.6 to travis? (either in addition to or in lieu of 3.5). not sure which is more common among users, but i use 3.6.2 locally.

Should we just move to testing on Python 3.6? It is not ideal, but I think a vast majority of users would either by using Python 2.7 or 3.6. Doing testing on both 3.5 and 3.6 will take up too much time, and not worth it for each commit.

jpchen: ok based on this thread it seems like everyones ok testing on 3.6. ill make the appropriate changes and submit a PR

fehiepsi: @neerajprad with python 3.6, we don't have to sort the keys in dict ;)

neerajprad: Life would indeed be much better if we just dropped python 2 support completely! 😆 PyMC3 is doing that.

PR #1463, opened 2018-10-17
add `retain graph` option to elbo

jpchen:
### Overview
this adds a `retain_graph` kwarg to `*Elbo` classes.  The motivation for this is #1394 where the user needs to retain the graph if they cache any intermediate variables. Not a breaking change.

### Tests
None, this is for discussion atm.

### DIFF
--- a/pyro/infer/elbo.py
+++ b/pyro/infer/elbo.py
@@ -50,10 +50,12 @@ class ELBO(object):
                  num_particles=1,
                  max_iarange_nesting=float('inf'),
                  vectorize_particles=False,
-                 strict_enumeration_warning=True):
+                 strict_enumeration_warning=True,
+                 retain_graph=False):
         self.num_particles = num_particles
         self.max_iarange_nesting = max_iarange_nesting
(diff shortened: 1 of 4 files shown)


pyro/infer/util.py
```diff
@@ -35,13 +35,13 @@ def torch_item(x):
     return x if isinstance(x, numbers.Number) else x.item()
 
 
-def torch_backward(x):
+def torch_backward(x, retain_graph=False):
```
fritzo: Default value should be `None`

pyro/infer/elbo.py
```diff
@@ -50,10 +50,12 @@ def __init__(self,
                  num_particles=1,
                  max_iarange_nesting=float('inf'),
                  vectorize_particles=False,
-                 strict_enumeration_warning=True):
+                 strict_enumeration_warning=True,
+                 retain_graph=False):
```
fritzo: Default value should be `None`, following PyTorch.

PR #1464, opened 2018-10-17
Fuse broadcast+iarange -> plate

fritzo:
Addresses #1465

This PR
- replaces `iarange` with `plate` everywhere;
- removes all usage of `@broadcast`, which now happens automatically via `plate`; and
- renames `max_iarange_nesting` to `max_plate_nesting`.

For backwards-compatibility, `iarange`, `broadcast`, and `max_iarange_nesting` are still available, and `iarange` and `max_iarange_nesting` raise `DeprecationWarning`s explaining the new usage.

This also refactors `plate` to be a `Messenger` rather than a handler, so we can continue to use the `pyro.plate("name", len(data))` syntax. We can later revert back to a handler by doing a `callable` or `isinstance(args[0], str)` check; however in this PR I've tried to make `plate` as close as possible to `iarange`.

## Tasks

- [x] Make `plate` closer to `iarange`
- [x] Replace `iarange` with `plate`
- [x] Remove `@broadcast` everywhere
- [x] Rename `max_iarange_nesting` -> `max_plate_nesting`
- [x] Update tensor shapes tutorial
- [x] Add deprecation warnings for `iarange` and `max_iarange_nesting`


eb8680:
Oof, 70 files changed.  When/how do we want this merged?  The sooner the better, or wait until closer to release?  Should we try and get outstanding PRs merged first?

fritzo:
> Oof, 70 files changed

I think this should not cause merge conflicts. Most changes were merely single-word replacements, and `iarange` and `@broadcast` still exist for backwards compatibility. @neerajprad WDYT?

neerajprad:
I think the merge conflicts in other PRs should be relatively simple to fix, so 👍 from my end!

pyro/infer/mcmc/hmc.py
```diff
@@ -93,13 +93,13 @@ def __init__(self,
...
-                 max_iarange_nesting=float("inf"),
+                 max_plate_nesting=float("inf"),
                  experimental_use_einsum=False):
         # Wrap model in `poutine.enum` to enumerate over discrete latent sites.
         # No-op if model does not have any discrete latents.
         self.model = poutine.enum(config_enumerate(model, default="parallel"),
-                                  first_available_dim=max_iarange_nesting)
-        # broadcast sample sites inside iarange.
+                                  first_available_dim=max_plate_nesting)
+        # broadcast sample sites inside plate.
```
neerajprad: Can we also remove the `poutine.broadcast` in the model below?

pyro/contrib/glmm/glmm.py
```diff
@@ -220,7 +220,7 @@ def bayesian_linear_model(design, w_means={}, w_sqrtlambdas={}, re_group_sizes={
             alpha, beta = re_alphas[name], re_betas[name]
             group_p = alpha.shape[-1]
             G_prior = dist.Gamma(alpha.expand(tau_shape + (group_p,)),
-                                 beta.expand(tau_shape + (group_p,)))
+                                 beta.expand(tau_shape + (group_p,))).independent(1)
```
fritzo: @eb8680 @ae-foster this appears to be a bug caught by the new `@broadcast`-by-default. Can you confirm that this is correct?

fritzo:
@neerajprad could you please review the updated tensor shapes tutorial?

tutorial/source/tensor_shapes.ipynb
```diff
@@ -528,22 +524,22 @@
...
-    " - We would also like to use the `pyro.iarange` construct to parallelize the ELBO estimator over [num_particles](http://docs.pyro.ai/en/latest/inference_algos.html#pyro.infer.elbo.ELBO). This is done by wrapping the contents of model/guide inside an outermost `pyro.iarange` context."
+    " - We have separated out the sampling function which returns the tensors corresponding to the active pixels. Modularizing the model code into components is a common practice, and helps with maintainability of large models.\n",
+    " - We would also like to use the `pyro.plate` construct to parallelize the ELBO estimator over [num_particles]s(http://docs.pyro.ai/en/latest/inference_algos.html#pyro.infer.elbo.ELBO). This is done by wrapping the contents of model/guide inside an outermost `pyro.plate` context."
```
neerajprad: Additional s `[num_particles]s(..)` is breaking the link.

tutorial/source/tensor_shapes.ipynb
```diff
@@ -610,10 +606,7 @@
...
     "        fun(observe=False, sample_fn=sample_fn)\n",
     "\n",
-    "    if broadcast:\n",
-    "        model = poutine.broadcast(model)\n",
-    "        guide = poutine.broadcast(guide)\n",
-    "    test_model(model, guide, TraceEnum_ELBO(max_iarange_nesting=3))\n",
+    "    test_model(model, guide, TraceEnum_ELBO(max_plate_nesting=3))\n",
     "\n",
     "test_model_with_sample_fn(sample_pixel_locations_no_broadcasting)\n",
     "test_model_with_sample_fn(sample_pixel_locations_automatic_broadcasting, broadcast=True)\n",
```
neerajprad: Just noticing this - we probably don't need the broadcast arg anymore?

tutorial/source/tensor_shapes.ipynb
```diff
@@ -528,22 +524,22 @@
...
-    "test_model(model4, guide4, TraceEnum_ELBO(max_iarange_nesting=2))"
+    "test_model(model4, guide4, TraceEnum_ELBO(max_plate_nesting=2))"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Automatic broadcasting via broadcast poutine<a class=\"anchor\" id=\"Automatic-broadcasting-via-broadcast-poutine\"></a>\n",
+    "### Automatic broadcasting inside pyro.plate<a class=\"anchor\" id=\"Automatic-broadcasting-inside-pyro-plate\"></a>\n",
```
neerajprad: I think while this section is good as an appendix, it probably needs to be reworded slightly (maybe minor tweaks to the example too) in light of recent changes. I am happy to do so in a separate PR, and incorporate any suggestions you have.

tutorial/source/tensor_shapes.ipynb
jpchen: yeah plus the var name is actually `num_particles` not `num_particle`. also `latest` is no longer a branch since it is the same as `dev` so if you'll need to change all `http://docs.pyro.ai/en/latest/...` to `http://docs.pyro.ai/en/dev/...`

pyro/contrib/glmm/glmm.py
ae-foster: Yep you are right @fritzo - the `group_p` dimension should be under `independent(1)`

ae-foster: Nice catch :)

tutorial/source/tensor_shapes.ipynb
fritzo: SGTM. I've reworded a bit, and I think further changes can be pushed to a follow-up PR.

fritzo:
Yikes, it looks like `@broadcast`-by-default is catching a bunch of previously silent bugs in our integration tests. I'll try to fix these, but I may end up marking a bunch of tests xfail and asking original authors to fix.

fritzo:
OK, I think this is ready to merge, pending tests and review.

PR #1466, opened 2018-10-17
Modify turning condition for nuts

fehiepsi:
With the introduction of mass matrix, we have to modify the turning condition. It seems not straightforward, as discussed in section A.4.2. of [A Conceptual Introduction to
Hamiltonian Monte Carlo](https://arxiv.org/abs/1701.02434). I will follow that section for the implementation (and compare the current implementation in Stan).

- [x] Add a flag to eliminate starting point from candidates of a trajectory
- [x] Implement new turning condition
- [x] Test
- [x] Debug the slowness (or why stepsize is so small when adapting)
- [x] Debug: gaussian conjugate tests fail

It would be fun btw. After this, I will add the multinomial sampling option for nuts. Hope that it would not be so complicated. ^^

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,8 +5,8 @@ all: docs test
 install: FORCE
 	pip install -e .[dev,profile]
 
-uninstall: FORCE
-	pip uninstall pyro-ppl
+reinstall: FORCE
+	pip uninstall -y pyro-ppl && pip install . --no-deps
 
 docs: FORCE
 	$(MAKE) -C docs html
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,9 @@ all: docs test
 install: FORCE
 	pip install -e .[dev,profile]
 
+uninstall: FORCE
+	pip uninstall pyro-ppl
+
 docs: FORCE
 	$(MAKE) -C docs html
 
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,6 @@ all: docs test
 install: FORCE
 	pip install -e .[dev,profile]
 
-reinstall: FORCE
-	pip uninstall -y pyro-ppl && pip install . --no-deps
-
 docs: FORCE
 	$(MAKE) -C docs html
 
--- a/pyro/infer/mcmc/nuts.py
+++ b/pyro/infer/mcmc/nuts.py
@@ -127,14 +127,15 @@ class NUTS(HMC):
         r_right_flat = torch.cat([r_right[site_name].reshape(-1) for site_name in sorted(r_right)])
         # TODO: change to torch.dot for pytorch 1.0
         if self.full_mass:
-            if ((r_sum - r_left_flat) * (self._inverse_mass_matrix.matmul(r_left_flat))).sum() > 0:
-                if ((r_sum - r_right_flat) * (self._inverse_mass_matrix.matmul(r_right_flat))) \
-                        .sum() > 0:
-                    return False
+            if (((r_sum - r_left_flat) * (self._inverse_mass_matrix.matmul(r_left_flat)))
+                    .sum() > 0 and
+                    ((r_sum - r_right_flat) * (self._inverse_mass_matrix.matmul(r_right_flat)))
+                    .sum() > 0):
+                return False
         else:
-            if (self._inverse_mass_matrix * (r_sum - r_left_flat) * r_left_flat).sum() > 0:
-                if (self._inverse_mass_matrix * (r_sum - r_right_flat) * r_right_flat).sum() > 0:
-                    return False
+            if ((self._inverse_mass_matrix * (r_sum - r_left_flat) * r_left_flat).sum() > 0 and
+                    (self._inverse_mass_matrix * (r_sum - r_right_flat) * r_right_flat).sum() > 0):
+                return False
         return True
 
     def _build_basetree(self, z, r, z_grads, log_slice, direction, energy_current):
--- a/pyro/infer/mcmc/nuts.py
+++ b/pyro/infer/mcmc/nuts.py
@@ -240,8 +240,8 @@ class NUTS(HMC):
         #     (z, r) ~ Uniform({(z', r') in trajectory | p(z', r') >= u}).
         #
         # For more information about slice sampling method, see [3].
-        # For another version of NUTS which uses multinomial sampling instead of slice sampling, see
-        # [2].
+        # For another version of NUTS which uses multinomial sampling instead of slice sampling,
+        # see [2].
 
         # Rather than sampling the slice variable from `Uniform(0, exp(-energy))`, we can
         # sample log_slice directly using `energy`, so as to avoid potential underflow or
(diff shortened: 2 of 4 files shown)


pyro/infer/mcmc/nuts.py
```diff
@@ -115,19 +116,27 @@ def __init__(self,
...
-            diff_left += (dz * r_left[name]).sum()
-            diff_right += (dz * r_right[name]).sum()
-        return diff_left < 0 or diff_right < 0
+        # Set a flag to decide if we want to eliminate the initial point from the candidates to
+        # choose uniformly along the trajectory. In [1], this flag is True, but in Stan, they set
+        # it to False (implicitly).
+        self._eliminate_starting_point = True
+
+    def _is_turning(self, r_left, r_right, r_sum):
+        # We follow the strategy in Section A.4.2 of [2] for this implementation.
```
neerajprad: Would also suggest including this [reference](https://arxiv.org/pdf/1304.1920.pdf) which derives the termination criterion explicitly.

fehiepsi: In [2], the author also derives the criterion too. I find it easier to understand than differential geometry style (though it is a nice language).

pyro/infer/mcmc/nuts.py
```diff
@@ -283,11 +294,13 @@ def sample(self, trace):
...
                 rand = pyro.sample("rand_t={}_treedepth={}".format(self._t, tree_depth),
                                    dist.Uniform(torch.zeros(1), torch.ones(1)))
-                if rand < new_tree.size / tree_size:
+                if ((tree_size > 0) and (rand < new_tree.size / tree_size)) \
+                        or ((tree_size == 0) and (new_tree.size > 0)):
                     accepted = True
                     z = new_tree.z_proposal
 
-                if self._is_turning(z_left, r_left, z_right, r_right):  # stop doubling
+                r_sum += new_tree.r_sum
```
fehiepsi: what a bug!!! it took me a lot of time to detect it. We should never do shortcut assignment for tensor.

pyro/infer/mcmc/nuts.py
neerajprad: Oh..its the same derivation. I was looking at 4.2 rather than the appendix.

tests/infer/mcmc/test_nuts.py
```diff
@@ -90,8 +128,8 @@ def model(data):
 @pytest.mark.parametrize(
     "step_size, adapt_step_size, adapt_mass_matrix, full_mass",
     [
-        (0.02, False, False, False),
-        (0.02, False, True, False),
+        (0.1, False, False, False),
+        (0.1, False, True, False),
```
fehiepsi: the old step_size works too but it is slow

tests/infer/mcmc/test_nuts.py
```diff
@@ -14,19 +14,57 @@
...
-T2 = T(*TEST_CASES[2].values)._replace(num_samples=800, warmup_steps=200)
-TEST_CASES[2] = pytest.param(*T2, marks=pytest.mark.skipif(
-    'CI' in os.environ or 'CUDA_TEST' in os.environ,
-    reason='Slow test - skip on CI/CUDA'))
-T3 = T(*TEST_CASES[3].values)._replace(num_samples=1000, warmup_steps=200)
-TEST_CASES[3] = pytest.param(*T3, marks=[
-    pytest.mark.skipif('CI' in os.environ or 'CUDA_TEST' in os.environ,
-                       reason='Slow test - skip on CI/CUDA')]
-)
+TEST_CASES = [
```
neerajprad: +1, this has been bugging me too for a while. 😄

pyro/infer/mcmc/nuts.py
neerajprad: Yeah, in place tensor ops can give non deterministic results, in many cases. Just curious - what was the cause of failure here?

tests/infer/mcmc/test_nuts.py
fehiepsi: yeah, me too ^^

fehiepsi:
@neerajprad from the gaussian test, I can see that using mass matrix makes things easier to sample. Now we don't need many samples to pass the test.

Makefile
```diff
@@ -5,8 +5,8 @@ all: docs test
 install: FORCE
 	pip install -e .[dev,profile]
 
-uninstall: FORCE
-	pip uninstall pyro-ppl
+reinstall: FORCE
```
neerajprad: When do you need this? If you do an editable install via -e, it should pick up any local changes.

fehiepsi: whoa, I have uninstalled and installed again and again to test examples, so I made this command. Thx @neerajprad!

Makefile
```diff
@@ -5,9 +5,6 @@ all: docs test
 install: FORCE
 	pip install -e .[dev,profile]
```
neerajprad: let us put the uninstall option back in.

pyro/infer/mcmc/nuts.py
```diff
@@ -115,19 +116,32 @@ def __init__(self,
...
+        self._eliminate_starting_point = True
+
+    def _is_turning(self, r_left, r_right, r_sum):
+        # We follow the strategy in Section A.4.2 of [2] for this implementation.
+        r_left_flat = torch.cat([r_left[site_name].reshape(-1) for site_name in sorted(r_left)])
+        r_right_flat = torch.cat([r_right[site_name].reshape(-1) for site_name in sorted(r_right)])
+        # TODO: change to torch.dot for pytorch 1.0
+        if self.full_mass:
+            if ((r_sum - r_left_flat) * (self._inverse_mass_matrix.matmul(r_left_flat))).sum() > 0:
+                if ((r_sum - r_right_flat) * (self._inverse_mass_matrix.matmul(r_right_flat))) \
```
neerajprad: nit: why not just `and` this condition?

pyro/infer/mcmc/nuts.py
fehiepsi: @neerajprad When tree_depth = 0, we create a new_tree with 1 element. Hence r_sum = r_left + r_right, but the shortcut assignment gives r_sum = r_left. Because of this, `is_turning` will return True (r_sum - r_left = 0), which makes NUTS not run at all or just run with 1 velocity verlet step.

fehiepsi:
@neerajprad Do you think that it is better to keep the old terminate condition (it is not a bug indeed) by using a hidden flag `_riemannian_turning_condition`? It might be good to have different versions to compare. If you think that it is good to have that flag, I will add it in the multinomial pull request. :)

neerajprad:
> Do you think that it is better to keep the old terminate condition (it is not a bug indeed) by using a hidden flag _riemannian_turning_condition? 

The old terminating condition with the identity mass matrix?  Do you mean that it is a valid terminating condition in terms of preserving detailed balance, even though it might not generate the longest trajectories? Unless you are already observing cases where the old terminating condition is yielding better results, I would be more inclined to just remove that option. In any case, this should be equivalent to our old code with mass matrix adaptation disabled, so we can always compare against that.

fehiepsi:
Never mind, ignoring it is totally fine to me. :)

PR #1469, opened 2018-10-18
Fix non determinism in HMC / NUTS

neerajprad:
Fixes the non-deterministic results in #1458, and in HMC/NUTS tests in general. The issue is in `torch.dot` which can give very slightly different results on different runs. This seems to be fixed on master but is an issue with `torch==0.4.0`. I am reverting to using element-wise product + sum, which seems to be stable. We can move to using the dot product in pyro==0.3.0.

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -156,9 +156,9 @@ class HMC(TraceKernel):
     def _kinetic_energy(self, r):
         r_flat = torch.cat([r[site_name].reshape(-1) for site_name in sorted(r)])
         if self.full_mass:
-            return 0.5 * r_flat.dot(self._inverse_mass_matrix.matmul(r_flat))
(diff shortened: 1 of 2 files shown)


pyro/infer/mcmc/hmc.py
```diff
@@ -156,9 +156,9 @@ def _compute_trace_log_prob(self, model_trace):
     def _kinetic_energy(self, r):
         r_flat = torch.cat([r[site_name].reshape(-1) for site_name in sorted(r)])
         if self.full_mass:
-            return 0.5 * r_flat.dot(self._inverse_mass_matrix.matmul(r_flat))
+            return 0.5 * (r_flat * (self._inverse_mass_matrix.matmul(r_flat))).sum()
```
fehiepsi: LGTM, if it is possible, it is better to add a TODO comment here for easy to remember (because this pattern is hard to grep).

neerajprad:
Should be good to merge.

PR #1471, opened 2018-10-18
Guess max_plate_nesting if it is not provided

### LINKED ISSUE #1455: [FR] Default to max_plate_nesting="auto"
@mbabadi requested an easier-to-use interface for `TraceEnum_ELBO`, where `max_plate_nesting` could be inferred automatically.

One possible solution would be to set the default to "auto" rather than `float('inf')`, and use `EscapeMessenger` to optimistically start at zero and then on encountering a `plate` exceeding headroom: break, increment, and retry. This should work for most models and will have only `O(1)` overhead overall since it should only happen at the first SVI step. Ideally `max_plate_nesting` would almost never need to be specified.

fritzo:
Resolves #1455 
Blocked by #1464 

This eliminates the need for users to specify the `max_plate_nesting` arg to `ELBO` in cases where model and guide structure are static (or at least max plate nesting is static). The guesswork has low overhead: it runs the (model,guide) pair only on the first SVI.step(), it samples rather than enumerates, it does not use num_particles, and it avoids computing log_prob if validation is disabled. For future JIT compatibility, we may need to later avoid jitting until `max_plate_nesting` is known.

## Tested

- [x] updated test_valid_models.py
- [x] removed `max_plate_nesting` arg from many tests in tests/infer/test_enum.py
- [ ] update tensor shapes tutorial

### DIFF
--- a/pyro/infer/elbo.py
+++ b/pyro/infer/elbo.py
@@ -1,11 +1,16 @@
 from __future__ import absolute_import, division, print_function
 
+import logging
 import warnings
 from abc import ABCMeta, abstractmethod
 
 from six import add_metaclass
 
 import pyro
+import pyro.poutine as poutine
+from pyro.infer.util import is_validation_enabled
+from pyro.poutine.util import prune_subsample_sites
+from pyro.util import check_site_shape
 
 
 @add_metaclass(ABCMeta)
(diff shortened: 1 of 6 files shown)


pyro/infer/elbo.py
```diff
@@ -62,14 +68,47 @@ def __init__(self,
...
-            if self.num_particles > 1:
-                if self.max_plate_nesting == float('inf'):
-                    raise ValueError("Automatic vectorization over num_particles requires " +
-                                     "a finite value for `max_plate_nesting` arg.")
-                self.max_plate_nesting += 1
+        if self.vectorize_particles and self.num_particles > 1:
+            self.max_plate_nesting += 1
         self.strict_enumeration_warning = strict_enumeration_warning
 
+    def _guess_max_plate_nesting(self, model, guide, *args, **kwargs):
```
neerajprad: It will be nice to be able to use this from HMC too. Is it possible to decouple this from the ELBO class and have guide be optional?

fritzo: I'd recommend just forking and simplifying this method in `HMC`. Two similar methods seem simpler than a complex conglomerate method filled with conditionals, imo.

fritzo:
> add a note ... to the tensor shapes tutorial

Sure, I'll update the tutorial after #1464 merges to avoid ipynb conflicts

PR #1472, opened 2018-10-18
Multinomial sampling for NUTS

fehiepsi:
Recently, NUTS has been improved using multinomial sampling to sample points in a trajectory (see section A.2.1 of https://arxiv.org/abs/1701.02434). This pull request implements that feature.

- [x] Sketch a working version
- [x] Refactor with a flag `use_multinomial_sampling` or similar name to turn on/off the multinomial sampling.
- [x] Test and fix bugs

After this pull request and #1443, I think that our HMC/NUTS implementations have achieved the current state of Stan/PyMC3 (algorithmic side). Cheer!


pyro/infer/mcmc/nuts.py
```diff
@@ -110,17 +120,12 @@ def __init__(self,
...
         # here u is the "slice" variable introduced at the `self.sample(...)` method.
         # Denote E_p = -log p(z, r), E_u = -log u, the third condition is equivalent to
-        #     sliced_energy := E_p - E_u >= some constant =: max_sliced_energy.
+        #     sliced_energy := E_p - E_u > some constant =: max_sliced_energy.
         # This also suggests the notion "diverging" in the implemenation:
         #     when the energy E_p diverges from E_u too much, we stop doubling.
         # Here, as suggested in [1], we set dE_max = 1000.
         self._max_sliced_energy = 1000
 
-        # Set a flag to decide if we want to eliminate the initial point from the candidates to
```
fehiepsi: In Stan, multinomial sampling version of NUTS includes initial point while slice sampling ignores initial point. To compromise, I remove this flag to avoid confusion. (a developer of Stan said that "including the original point is important for avoiding periodicity.")

neerajprad:
@fehiepsi - You will need to update with `dev` to pull in the latest changes.

jpchen:
@rohitsingh0812 i think with this PR we can revisit nuts comparisons between stan and pyro

pyro/infer/mcmc/nuts.py
```diff
@@ -51,6 +55,10 @@ class NUTS(HMC):
     :param bool adapt_mass_matrix: A flag to decide if we want to adapt mass
         matrix during warm-up phase using Welford scheme.
     :param bool full_mass: A flag to decide if mass matrix is dense or diagonal.
+    :param bool use_multinomial_sampling: A flag to decide if we want to sample
+        candidates along its trajectory using "multinomial sampling" or using
+        "slice sampling". Slice sampling is used in the original NUTS paper [1],
+        while multinomial sampling is suggested in [2].
```
neerajprad: We should mention that this is the default, and if set to False, the sampler uses slice sampling instead.

pyro/infer/mcmc/nuts.py
neerajprad: Does this cause any issues with slice sampling? I am guessing not because that's what we had earlier too.

pyro/infer/mcmc/nuts.py
```diff
@@ -150,24 +155,28 @@ def _build_basetree(self, z, r, z_grads, log_slice, direction, energy_current):
             z, r, self._potential_energy, self._inverse_mass_matrix, step_size, z_grads=z_grads)
         r_new_flat = torch.cat([r_new[site_name].reshape(-1) for site_name in sorted(r_new)])
         energy_new = potential_energy + self._kinetic_energy(r_new)
-        sliced_energy = energy_new + log_slice
 
-        # As a part of the slice sampling process (see below), along the trajectory
-        #     we eliminate states which p(z, r) < u, or dE > 0.
-        # Due to this elimination (and stop doubling conditions),
-        #     the size of binary tree might not equal to 2^tree_depth.
-        tree_size = 1 if sliced_energy <= 0 else 0
```
neerajprad: We should modify or remove the comment below now, since we aren't handling this as a special case.

pyro/infer/mcmc/nuts.py
```diff
@@ -209,12 +225,16 @@ def _build_tree(self, z, r, z_grads, log_slice, direction, tree_depth, energy_cu
         # For the special case that the sizes of each half are both 0,
         #     we choose the proposal from the first half
         #     (any is fine, because the probability of picking it at the end is 0!).
-        if tree_size != 0:
```
neerajprad: We can remove the comment above?

pyro/infer/mcmc/nuts.py
```diff
@@ -268,16 +288,22 @@ def sample(self, trace):
         # Rather than sampling the slice variable from `Uniform(0, exp(-energy))`, we can
         # sample log_slice directly using `energy`, so as to avoid potential underflow or
         # overflow issues ([2]).
-        slice_exp_term = pyro.sample("slicevar_exp_t={}".format(self._t),
```
neerajprad: Move the above comment to the else block.

pyro/infer/mcmc/nuts.py
```diff
@@ -150,24 +155,28 @@ def _build_basetree(self, z, r, z_grads, log_slice, direction, energy_current):
...
-            accept_prob = energy_new.new_tensor(0.0)
+        # This is obtained by setting energy_new to +inf when it is nan.
+        energy_new = energy_new.new_tensor(float("inf")) if torch_isnan(energy_new) else energy_new
+        sliced_energy = energy_new + log_slice
+        diverging = (sliced_energy > self._max_sliced_energy)
+        delta_energy = energy_new - energy_current
+        accept_prob = (-delta_energy).exp().clamp(max=1.0)
+
+        if self.use_multinomial_sampling:
+            tree_size = -sliced_energy
```
neerajprad: This variable name is unintuitive for the multinomial sampling case - we should consider renaming this and adding a comment as to what it corresponds to in the multinomial and slice sampling case.

pyro/infer/mcmc/nuts.py
```diff
@@ -198,7 +207,14 @@ def _build_tree(self, z, r, z_grads, log_slice, direction, tree_depth, energy_cu
         other_half_tree = self._build_tree(z, r, z_grads, log_slice,
                                            direction, tree_depth-1, energy_current)
 
-        tree_size = half_tree.size + other_half_tree.size
+        if self.use_multinomial_sampling:
+            # TODO: for pytorch 1.0, we don't need to check these conditions
+            if half_tree.size != float("-inf") or other_half_tree != float("-inf"):
```
neerajprad: nit: Reverting the order of the conditionals seems more natural.

```python
if half_tree.size == float("-inf") and other_half_tree == float("-inf"):
    return half_tree.size  # -inf
else:
    return logsumexp(torch.stack([half_tree.size, other_half_tree.size]), dim=0)
```

pyro/infer/mcmc/nuts.py
```diff
@@ -150,24 +155,28 @@ def _build_basetree(self, z, r, z_grads, log_slice, direction, energy_current):
...
-        # As a part of the slice sampling process (see below), along the trajectory
-        #     we eliminate states which p(z, r) < u, or dE > 0.
-        # Due to this elimination (and stop doubling conditions),
-        #     the size of binary tree might not equal to 2^tree_depth.
-        tree_size = 1 if sliced_energy <= 0 else 0
         # Special case: Set diverging to True and accept prob to 0 if the
         # diverging trajectory returns `NaN` energy (e.g. in the case of
         # evaluating log prob of a value simulated using a large step size
         # for a constrained sample site).
-        if torch_isnan(energy_new):
```
neerajprad: We should modify or remove the comment below now, since we aren't handling this as a special case.

pyro/infer/mcmc/nuts.py
fehiepsi: I have added a comment at the definition of TreeInfo. Could you suggest me another name for it? :)

pyro/infer/mcmc/nuts.py
fehiepsi: It seems that we don't have to check these conditions. It will never happen because we return first half_tree when it is turning or diverging.

pyro/infer/mcmc/nuts.py
neerajprad: How about using `weight` - that might convey the desired use case?

pyro/infer/mcmc/nuts.py
neerajprad: Great, let us remove the comment in that case!

pyro/infer/mcmc/nuts.py
fehiepsi: That is a good name!

fehiepsi: Thanks @neerajprad !

pyro/infer/mcmc/nuts.py
fehiepsi: Ah, when using slice_sampling, if sliced_energy > 0, we also ignore the point (though it is not both turning and diverging). So there are cases tree_size = 0.

PR #1474, opened 2018-10-19
Guess max_plate_nesting if it is not provided

### LINKED ISSUE #1455: [FR] Default to max_plate_nesting="auto"

neerajprad:
Resolves #1455 
Blocked by #1464 

This eliminates the need for users to specify the `max_plate_nesting` arg to `ELBO` in cases where model and guide structure are static (or at least max plate nesting is static). The guesswork has low overhead: it runs the (model,guide) pair only on the first SVI.step(), it samples rather than enumerates, it does not use num_particles, and it avoids computing log_prob if validation is disabled. For future JIT compatibility, we may need to later avoid jitting until `max_plate_nesting` is known.

## Tested

- [x] updated test_valid_models.py
- [x] removed `max_plate_nesting` arg from many tests in tests/infer/test_enum.py
- [x] update tensor shapes tutorial
- [x] guess in `Trace_ELBO` and `TraceGraph_ELBO` if vectorize_particles=True

\* Refer to the original PR - https://github.com/uber/pyro/pull/1471 for earlier comments. This is just reinstating the same, since it was wrongly merged into its base PR.


fritzo:
@neerajprad Could you please review? I am ready to merge pending review and tests.

tests/infer/test_valid_models.py
```diff
@@ -966,7 +980,7 @@ def model():
             assert q.shape == (8, 5, 7, 6)
 
     guide = config_enumerate(model, expand=expand) if enumerate_ else model
-    assert_ok(model, guide, Elbo())
+    assert_ok(model, guide, Elbo(max_plate_nesting=4))
```
neerajprad: Does this example not work with plate guessing?

fritzo: This test does not work because there are shape assertions in the model. Those assertions assume the model is being enumerated, and they fail when the model is run via `_guess_max_plate_nesting()`. I couldn't see a way to conditionally enable the assertions.

neerajprad: Ahh..I see. Thanks for explaining!

PR #1478, opened 2018-10-20
Make HMC try to find a valid initial trace

fehiepsi:
As title.

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -334,7 +334,7 @@ class HMC(TraceKernel):
         assert pos == r_flat.size(0)
         return r, r_flat
 
-    def _validate_trace(self, trace):
+    def _get_valid_prototype_trace(self, trace):
         trace_eval = TraceEinsumEvaluator if self.use_einsum else TraceTreeEvaluator
         self._trace_prob_evaluator = trace_eval(trace,
                                                 self._has_enumerable_sites,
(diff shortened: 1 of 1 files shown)


ngoodman:
How does it search? Rejection?

fehiepsi:
@ngoodman We just take random samples from the prior until we get a valid one (non NaN energy). I think there is no need for other sampling/search at this step.

pyro/infer/mcmc/hmc.py
```diff
@@ -366,7 +372,7 @@ def setup(self, warmup_steps, *args, **kwargs):
             self._r_shapes[name] = site_value.shape
             self._r_numels[name] = site_value.numel()
 
-        self._validate_trace(trace)
+        self._get_valid_prototype_trace(trace)
```
neerajprad: Would suggest a minor refactoring as follows:
 - Make the `_get_valid_prototype_trace` not take in any args, and just return a valid trace, or throw a `ValueError`
 - Move the call to this method early in setup, and use this trace for any subsequent setup. That way, we can terminate early if we don't find a good value for the trace.

   ```python
   def setup(..):
       trace = self._get_valid_prototype_trace()
       # subsequent operations
       ...
   ```

fehiepsi: In the setup, we need to know if we have `_has_enumerable_sites` to feed into `trace_prob_evaluator` . So I think that we need to understand the structure before finding a valid trace.

neerajprad: I see - I think it might call for a slightly bigger reorganization. Let us just rename the method to `_set_valid_prototype_trace` to clearly designate that it is setting an attribute.

PR #1484, opened 2018-10-22
Fix CUDA tests on dev

neerajprad:


pyro/infer/mcmc/nuts.py
```diff
@@ -326,7 +326,8 @@ def sample(self, trace):
                 else:
                     new_tree_prob = new_tree.weight / tree_weight
                 rand = pyro.sample("rand_t={}_treedepth={}".format(self._t, tree_depth),
-                                   dist.Uniform(torch.zeros(1), torch.ones(1)))
+                                   dist.Uniform(new_tree_prob.new_tensor(0.),
```
neerajprad: With multinomial sampling, `new_tree_prob` can be a CUDA tensor.

jpchen: shouldnt this use default tensor type which is cuda on gpu?

neerajprad: If we set the default tensor type to CUDA, then that will work. But we will have issues when the data (/ model) is on the GPU, without changing the default tensor type. This is the case for our examples, for instance.

PR #1489, opened 2018-10-23
Add `CensoredDistribution`

ae-foster:
We currently have `TransformedDistribution` which works with *invertible* transformations of random variables. The new `CensoredDistribution`, added here, makes the following non-invertible transformation
```
X ~ dist
X[X >= upper_lim] = upper_lim
X[X <= lower_lim] = lower_lim
```
The new pdf uses the `log(cdf)` values at `upper_lim` and `lower_lim`, and the original pdf elsewhere. This is a valid probability density function, albeit with respect to a new base measure.

These distributions find applications in `contrib.oed` work (they make interesting OED problems because censoring leads to lower / zero information gain)

### DIFF
--- a/pyro/distributions/censored.py
+++ b/pyro/distributions/censored.py
@@ -1,6 +1,5 @@
 import torch
 from torch.distributions import constraints
-from torch.distributions.utils import _sum_rightmost
 
 from pyro.distributions.torch_distribution import TorchDistribution
 
@@ -9,7 +8,8 @@ class CensoredDistribution(TorchDistribution):
 
     def __init__(self, base_distribution, upper_lim=float('inf'), lower_lim=float('-inf'), validate_args=None):
         # Log-prob only computed correctly for univariate base distribution
-        assert base_distribution.event_dim == 0 or base_distribution.event_dim == 1 and base_distribution.event_shape[0] == 1
+        assert base_distribution.event_dim == 0 or (
+                base_distribution.event_dim == 1 and base_distribution.event_shape[0] == 1)
         self.base_dist = base_distribution
         self.upper_lim = upper_lim
         self.lower_lim = lower_lim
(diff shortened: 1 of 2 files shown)


pyro/distributions/censored.py
```diff
@@ -0,0 +1,71 @@
...
+        with torch.no_grad():
+            x = self.base_dist.sample(sample_shape)
+            x[x > self.upper_lim] = self.upper_lim
+            x[x < self.lower_lim] = self.lower_lim
+            return x
+
+    def rsample(self, sample_shape=torch.Size()):
+        x = self.base_dist.sample(sample_shape)
+        x[x > self.upper_lim] = self.upper_lim
+        x[x < self.lower_lim] = self.lower_lim
```
fritzo: `return x` and add a test that would have caught this.

fritzo:
cc @alicanb who prototyped `TruncatedDistribution` in https://github.com/probtorch/pytorch/pull/121

ae-foster:
Thanks @fritzo ! I will get on and write some tests

activatedgeek:
Hey @ae-foster, can you point me to some references where Censored Distribution was required?

ae-foster:
I'm going to close this PR unmerged. And it's not because I'm too lazy to write the test (ok it's partly that).

The main reason is that I didn't use this code as it stands in `pyro.contrib.oed`. I want to point out what the issue with this code is, in case anyone tries something similar in future. The issue is quite simply the numerical stability of `torch.log(cdf)`. Log-cdf values, like log-pdf values can be large and negative. Unfortunately, taking the explicit `log` of floating point values will run into problems in this regime. What is needed is an asymptotic expansion of the cdf or its log. Fortunately, for certain distributions (e.g. the Normal) we can find such an asymptotic expansion. The resulting algorithm blends between `log(cdf)` and this asymptotic expansion to avoid the numerical instabilities. Unfortunately, this is distribution specific, thus precluding the use of a `CensoredDistribution` tool.

For details of my new approach, see https://github.com/ae-foster/pyro/blob/oed-master/pyro/distributions/censored_sigmoid_normal.py

PR #1492, opened 2018-10-23
PermutationFlow

stefanwebb:
A simple flow that reorders the elements of a random vector.

Also, I've separated the tests for MADE and flows.

I'm trying to make my pull requests more incremental, so new IAF implementation will follow after this one.


martinjankowiak:
can you please make this `PermutedDistribution` since it doesn't actually flow?

stefanwebb:
Well, it's more of a bijection than a distribution. If we were to name it `PermutedDistribution` would that lead to confusion? People could think it inherits from `TorchDistribution` rather than `Transform`. What do you think?

stefanwebb:
How about calling it `Permute`? Would it be too drastic a design change to separate out the transformations from the distributions by putting them into a different module (like in TensorFlow)?

martinjankowiak:
yeah sorry i'd go with `PermutationTransform` or `Permute` for now

would need to gather more evidence for/against before considering larger design change

pyro/distributions/permutation_flow.py
```diff
@@ -0,0 +1,78 @@
...
+    A normalizing flow that reorders the input dimensions, that is, multiplies the input by a permutation matrix.
+    This is useful in between IAF transforms to increase the flexibility of the resulting distribution and
+    stabilize learning. Whilst not being an autoregressive flow, the log absolute determinate of the Jacobian is
+    easily calculable as 0. Note that reordering the input dimension between two layers of IAF is not equivalent
+    to reordering the dimension inside the MADE networks that those IAFs use; using a PermutationFlow results in a
+    distribution with more flexibility.
+
+    Example usage:
+
+    >>> from pyro.nn import AutoRegressiveNN
```
fritzo: Can you please add this class to docs/source/distributions.rst, then `make docs` and ensure docs are generated correctly? I believe you'll need to indent this code block for docs to be generated correctly.

pyro/distributions/permutation_flow.py
```diff
@@ -0,0 +1,78 @@
...
+    >>> iaf_dist.sample()  # doctest: +SKIP
+        tensor([-0.4071, -0.5030,  0.7924, -0.2366, -0.2387, -0.1417,  0.0868,
+                0.1389, -0.4629,  0.0986])
+
+    :param permutation: a permutation ordering that is applied to the inputs.
+    :type permutation: torch.LongTensor
+
+    """
+
+    codomain = constraints.real
```
fritzo: Add:
```py
bijective = True
```

pyro/distributions/permutation_flow.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __init__(self, permutation):
+        super(PermutationFlow, self).__init__()
+
+        self.permutation = permutation
+
+        # Calculate the inverse permutation order
+        self.inv_permutation = torch.empty(permutation.shape, dtype=torch.long)
+        self.inv_permutation[permutation] = torch.arange(permutation.size(0), dtype=torch.long)
```
fritzo: I think you should also define `self.sign` depending on the sign of the permutation.

pyro/distributions/permutation_flow.py
```diff
@@ -0,0 +1,78 @@
...
+
+    codomain = constraints.real
+
+    def __init__(self, permutation):
+        super(PermutationFlow, self).__init__()
+
+        self.permutation = permutation
+
+        # Calculate the inverse permutation order
+        self.inv_permutation = torch.empty(permutation.shape, dtype=torch.long)
```
fritzo: I think you'll need to add `device=permutation.device` to the following two lines.

pyro/distributions/permutation_flow.py
stefanwebb: Would this be a useful bit of information to a user do you think?

stefanwebb:
I've made all the suggested changes except adding in `self.sign`. Do you mean the parity of the permutation? Like decompose it into transpositions?

pyro/distributions/permutation_flow.py
stefanwebb: `make docs` seems to work now with the new class in `distributions.rst`

pyro/distributions/permutation_flow.py
stefanwebb: Done

pyro/distributions/permutation_flow.py
stefanwebb: Done

pyro/distributions/permutation_flow.py
fritzo: Oh my bad, I confused the monotonicity sign with the jacobian sign, and I think you're right that monotonicity sign does not make sense for this transform.

fritzo:
@stefanwebb can you please merge dev into this PR to fix linting errors #1496 ? No need to address `sign` stuff; I was confused.

stefanwebb:
Its all good now

pyro/distributions/permute.py
```diff
@@ -0,0 +1,80 @@
...
+    bijective = True
+
+    def __init__(self, permutation):
+        super(PermuteTransform, self).__init__()
+
+        self.permutation = permutation
+
+        # Calculate the inverse permutation order
+        self.inv_permutation = torch.empty_like(permutation, dtype=torch.long)
+        self.inv_permutation[permutation] = torch.arange(permutation.size(0), dtype=torch.long,
```
fritzo: nit: If this is only sometimes used, consider making it lazy
```py
from torch.distributions.utils import lazy_property

class PermuteTransform(Transform):
    ...

    @lazy_property
    def inv_permutation(self):
        result = torch.empty_like(self.permutation, dtype=torch.long)
        result[self.permutation] = torch.arange(permutation.size(0), dtype=torch.long,
                                                device=self.permutation.device)
        return result
```

pyro/distributions/permute.py
```diff
@@ -0,0 +1,80 @@
...
+from torch.distributions.transforms import Transform
+from torch.distributions import constraints
+
+from pyro.distributions.util import copy_docs_from
+
+
+@copy_docs_from(Transform)
+class PermuteTransform(Transform):
+    """
+    A normalizing flow that reorders the input dimensions, that is, multiplies the input by a permutation matrix.
```
martinjankowiak: can we please get rid of the word flow entirely

pyro/distributions/permute.py
```diff
@@ -0,0 +1,80 @@
...
+    This is useful in between IAF transforms to increase the flexibility of the resulting distribution and
```
martinjankowiak: can we please link to IAF docs here

pyro/distributions/permute.py
```diff
@@ -0,0 +1,80 @@
...
+
+        Inverts y => x.
+        """
+
+        return y[..., self.inv_permutation]
+
+    def log_abs_det_jacobian(self, x, y):
+        """
+        Calculates the elementwise determinant of the log Jacobian, i.e. log(abs([dy_0/dx_0, ..., dy_{N-1}/dx_{N-1}])).
+        Note that this type of flow is not autoregressive, so the log Jacobian is not the sum of the previous
```
martinjankowiak: again, please axe 'flow'

pyro/distributions/permute.py
stefanwebb: I think it would be very rarely used only sometimes, when you want to sample but not score samples from a normalizing flow

stefanwebb: Done!

stefanwebb:
Thanks!

PR #1493, opened 2018-10-24
add citation to README

jpchen:
do i need to add "to appear" to the journal?


neerajprad:
You will need to pull in the flake8 changes from dev.

Issue #1460, opened 2018-10-17
Fix TraceEnum_ELBO.compute_marginals(), .sample_posterior()


fritzo:
Currently `TraceEnum_ELBO.compute_marginals()`, and `.sample_posterior()` incorrectly handle `iarange`s. The issue is in `contract_to_tensor`, where information is not correctly passed across local variables that are coupled by a global variable. The error can be detected in discrepancy between `naive_ubersum()` and `ubersum()`.

**Currently working:** when enumerated variables are either all local or all global
**Currently failing:** when there is communication between global and local enumerated variables.

PR #1495, opened 2018-10-24
Unify ubersum logic (step 1/2)

fritzo:
DO NOT MERGE
Addresses #1460 

This pure-refactoring PR switches to use an in-house implementation of `ContractExpression` for `pyro.ops.contract.contract()`. We still use `opt_einsum` for sharing logic. A follow-up PR will add ubersum's batch functionality to `pyro.ops.einsum.paths.optimize()` and `ContractExpression`.

## Questions for reviewers

- [x] @neerajprad can you suggest ways to check performance impact?

## Tested

- [x] added tests comparing the in-house implementation to `opt_einsum.contract()`.


pyro/ops/einsum/contract.py
```diff
@@ -0,0 +1,61 @@
...
+def _einsum(equation, *operands, **kwargs):
+    backend = kwargs.pop('backend', 'numpy')
+    if backend == 'torch':
+        # provide np.einsum interface for torch.einsum.
+        return torch.einsum(equation, operands)
+    einsum = getattr(importlib.import_module(backend), 'einsum')
+    return einsum(equation, *operands, **kwargs)
+
+
+class ContractExpression(object):
```
fritzo: This is the entirety of the new logic, similar to opt_einsum's [ContractExpression](https://github.com/dgasmith/opt_einsum/blob/dbf284b/opt_einsum/contract.py#L575).

neerajprad:
> can you suggest ways to check performance impact?

@fritzo - I have mostly just been using timeit (for small snippets on ipython, or quickly timing tests), and cProfiler and line profiler to do more detailed profiling. I just tested your changes on the HMC example, and there doesn't seem to be any regression on that (if anything it is a tiny bit faster).

If you have a canonical example that you would like to monitor for perf regression, let me know. I can add it to `test_benchmark.py` which creates a more isolated environment and runs via a cron every night on my workstation. :)

pyro/ops/einsum/contract.py
eb8680: It looks like the main piece of functionality in `opt_einsum` left out of this version is constant terms.  Were we ever using those?

fritzo: We were not using constant terms. I don't foresee us using those. Hopefully when `ubersum` settles down we can try to move this back upstream.

fritzo:
Closing this for now, as it appears not to be the shortest path to fixing #1460.

PR #1497, opened 2018-10-24
fix broken code+link format in gmm tutorial

fehiepsi:
The current [gmm tutorial](http://pyro.ai/examples/gmm.html) is rendered wrong for code+link format. This fixes that error.


fritzo:
@fehiepsi can you please merge dev into this PR to fix recent change in flake8

PR #1498, opened 2018-10-25
New version of IAF and inversion operator

stefanwebb:
The default version of IAF uses a slightly different equation with better properties. The old version of IAF is renamed `InverseAutoregressiveFlowStable`. Both now have an inversion operator that is called when the inverse is not in the cache.


stefanwebb:
Is it possible to put display mode equations in the documentation?

eb8680:
> Is it possible to put display mode equations in the documentation?

Yep, see e.g. the `pyro.contib.glmm` docs: https://github.com/uber/pyro/blob/dev/pyro/contrib/glmm/glmm.py#L130

stefanwebb:
There might be a chance for a small amount of code reuse... Let me see what a `InverseAutoregressiveFlowBase` class for this would look like

stefanwebb:
Cool, I'll put the basic equation that the flows use in the docs

stefanwebb:
I've tried but I don't think it does simplify things, so I'd like to leave it as it is. The docs have the equations in them now

PR #1500, opened 2018-10-28
Refactor sum_dims to be a set rather than dict

fritzo:
This pure-refactoring PR prepares for #1460.
- Before this PR, `sum_dims` was a dict mapping each tensor to the set of dims in that tensor that should be sum-contracted out.
- After this PR, `sum_dims` is simply the set of all dims to be sum-contracted out.

@eb8680 this will conflict with your [collapse-messenger](https://github.com/uber/pyro/compare/collapse-messenger) branch; you should be able to resolve easily by looking at the changes to `BackwardSampleMessenger`.

## Tested

### DIFF
--- a/pyro/ops/contract.py
+++ b/pyro/ops/contract.py
@@ -95,7 +95,10 @@ class UnpackedLogRing(TensorRing):
     Ordinals are frozensets of ``CondIndepStackFrame``s.
     """
     def dims(self, term):
-        return [d for d in range(-term.dim(), 0) if term.size(d) > 1]
+        key = 'dims', id(term)
+        if key not in self._cache:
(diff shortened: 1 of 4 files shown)


pyro/ops/contract.py
```diff
@@ -95,7 +95,10 @@ class UnpackedLogRing(TensorRing):
     Ordinals are frozensets of ``CondIndepStackFrame``s.
     """
     def dims(self, term):
-        return [d for d in range(-term.dim(), 0) if term.size(d) > 1]
+        key = 'dims', id(term)
```
fritzo: previously we computed `dims` once and cached in `sum_dims`. Since we now call this more often, I've added memoization logic (as is already done in the other `TensorRing` implementation).

fritzo:
Thanks for reviewing!

PR #1501, opened 2018-10-29
Use outer product instead of matmul in welford algorithm

neerajprad:
Using `torch.ger` is more computationally efficient than `matmul` for computing outer product in the welford scheme, an online algorithm for computing covariance (used in HMC/NUTS for mass matrix estimation).

### DIFF
--- a/pyro/ops/welford.py
+++ b/pyro/ops/welford.py
@@ -29,7 +29,7 @@ class WelfordCovariance(object):
         if self.diagonal:
             self._m2 += delta_pre * delta_post
         else:
-            self._m2 += torch.matmul(delta_post.unsqueeze(-1), delta_pre.unsqueeze(-2))
+            self._m2 += torch.ger(delta_post, delta_pre)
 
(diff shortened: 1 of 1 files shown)


pyro/ops/welford.py
```diff
@@ -29,7 +29,7 @@ def update(self, sample):
         if self.diagonal:
             self._m2 += delta_pre * delta_post
         else:
-            self._m2 += torch.matmul(delta_post.unsqueeze(-1), delta_pre.unsqueeze(-2))
+            self._m2 += torch.ger(delta_post, delta_pre)
```
fritzo: Weird, I don't understand why there was a `matmul` at all, rather than the simpler
```py
delta_post.unsqueeze(-1) * delta_pre.unsqueeze(-2)
```

neerajprad: The version that relies on broadcasting is a bit slower (since I guess it goes through the expand operation first) than matmul, but I doubt this was written with perf considerations in mind. 😄

PR #1503, opened 2018-10-30
remove numpy usage in main package

fehiepsi:
Given that most numeric calculations can be done with pytorch, we should avoid relying on `numpy` if possible.

### DIFF
--- a/pyro/ops/einsum/paths.py
+++ b/pyro/ops/einsum/paths.py
@@ -4,7 +4,7 @@ import heapq
 import itertools
 from collections import defaultdict
 
-import numpy as np
+import torch
 
 
 def ssa_to_linear(ssa_path):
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -54,6 +54,7 @@
 EXTRAS_REQUIRE = [
     'jupyter>=1.0.0',
     'matplotlib>=1.3',
+    'numpy>=1.7',
```
fritzo: I'll still need numpy in an upcoming PR, so I'll need to revert this line.

pyro/ops/einsum/paths.py
```diff
@@ -15,7 +15,7 @@ def ssa_to_linear(ssa_path):
         >>> ssa_to_linear([(0, 3), (2, 4), (1, 5)])
         [(0, 3), (1, 2), (0, 1)]
     """
-    ids = np.arange(1 + max(map(max, ssa_path)), dtype=np.int32)
+    ids = torch.arange(1 + max(map(max, ssa_path)), dtype=torch.long)
```
fritzo: Please revert. This should be kept as close as possible to the upstream version.

fehiepsi: @fritzo sorry, I didn't know about that. Do you want me to revert in another pull request?

fritzo: no worries. would you mind opening a tiny PR that reverts this file and setup.py. The other changes are nice, thanks for cleaning up!

neerajprad: Sorry, I should have checked with you on the change in paths.py. Thanks for reverting this part, @fehiepsi.

PR #1506, opened 2018-10-31
Return a deterministic trace generator when using HMC with parallel chains

### LINKED ISSUE #1505: HMC gives non deterministic results with parallel chaining
Even though the `ParallelSampler` yields the same set of traces (each worker gives deterministic results), the order in which the traces are consumed might be different, which gives different results when we use `TracePredictive`. 

One way to get deterministic results from HMC/NUTS with `num_chains > 1` would be to store intermediate results, and yield results from each of the workers in a round-robin fashion.

neerajprad:
Fixes #1505. 

While each of the workers in `mcmc._ParallelSampler` returns deterministic outputs, the order of the traces returned by `_ParallelSampler._traces` generator is non-deterministic (it depends on how fast/slow each of the workers is). This gives deterministic stats on the latents, but non-deterministic results when the `TracePosterior` object is consumed by `TracePredictive`.

As a solution, we are yielding traces from each of the chains in a round-robin fashion and using a queue to hold the intermediate traces from the workers. 

Note that this is not a final interface and may need a slight refactoring (or more design discussion) when we would like to compute and provide diagnostics that need chain information. In that case, we may have to provide an optional chain id to the parent `TracePosterior` class.

### DIFF
--- a/pyro/infer/mcmc/mcmc.py
+++ b/pyro/infer/mcmc/mcmc.py
@@ -35,7 +35,7 @@ def logger_thread(log_queue, warmup_steps, num_samples, num_chains):
     try:
         while True:
             try:
-                record = log_queue.get_nowait()
+                record = log_queue.get(1)
             except queue.Empty:
                 continue
             if record is None:
@@ -154,7 +154,7 @@ class _ParallelSampler(TracePosterior):
                 w.start()
             while active_workers:
                 try:
-                    chain_id, val = self.result_queue.get_nowait()
+                    chain_id, val = self.result_queue.get(1)
                 # This can happen when the worker process has terminated.
                 # See https://github.com/pytorch/pytorch/pull/5380 for motivation.
                 except socket.error as e:
--- a/pyro/infer/mcmc/mcmc.py
+++ b/pyro/infer/mcmc/mcmc.py
@@ -35,7 +35,7 @@ def logger_thread(log_queue, warmup_steps, num_samples, num_chains):
     try:
         while True:
             try:
-                record = log_queue.get(1)
+                record = log_queue.get_nowait()
             except queue.Empty:
                 continue
             if record is None:
@@ -154,7 +154,7 @@ class _ParallelSampler(TracePosterior):
                 w.start()
             while active_workers:
                 try:
-                    chain_id, val = self.result_queue.get(1)
+                    chain_id, val = self.result_queue.get_nowait()
                 # This can happen when the worker process has terminated.
                 # See https://github.com/pytorch/pytorch/pull/5380 for motivation.
                 except socket.error as e:
--- a/pyro/infer/mcmc/mcmc.py
+++ b/pyro/infer/mcmc/mcmc.py
@@ -8,7 +8,7 @@ import socket
 import sys
 import threading
 import warnings
-from collections import OrderedDict
+from collections import OrderedDict, deque
 
 import six
 from six.moves import queue
@@ -136,6 +136,16 @@ class _ParallelSampler(TracePosterior):
             if w.is_alive():
                 w.terminate()
 
+    @staticmethod
+    def _next(results_buffer, sample_idx):
+        val = results_buffer[sample_idx].popleft()
+        if val is None:
+            results_buffer.pop(sample_idx)
+            sample_idx -= 1
+        sample_idx = (sample_idx + 1) % len(results_buffer) if len(results_buffer) \
+            else None
+        return results_buffer, sample_idx, val
+
     def _traces(self, *args, **kwargs):
         # Ignore sigint in worker processes; they will be shut down
         # when the main process terminates.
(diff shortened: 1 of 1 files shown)


pyro/infer/mcmc/mcmc.py
```diff
@@ -162,10 +167,23 @@ def _traces(self, *args, **kwargs):
...
                     raise val
-                elif val is not None:
-                    yield val
-                else:
+                results_buffer[chain_id - 1].append(val)
+                if val is None:
                     active_workers -= 1
+                while results_buffer[cur_idx]:
+                    val = results_buffer[cur_idx].popleft()
+                    if val:
```
fritzo: what does `if val` mean here? I.e. what types and values can `val` take on? Can you more precisely test `if val is not None`?

pyro/infer/mcmc/mcmc.py
```diff
@@ -162,10 +167,23 @@ def _traces(self, *args, **kwargs):
                 if isinstance(val, Exception):
                     # Exception trace is already logged by worker.
                     raise val
-                elif val is not None:
-                    yield val
-                else:
+                results_buffer[chain_id - 1].append(val)
```
fritzo: It seems a little weird to `.append(None)` and then not yield it. Would it be simpler to rearrange so that you
```py
if val is None:
    active_worker -= 1
else:
    results_buffer[chain_id - 1].append(val)
    ...
```

pyro/infer/mcmc/mcmc.py
```diff
@@ -144,6 +144,11 @@ def _traces(self, *args, **kwargs):
         # restore original handler
         signal.signal(signal.SIGINT, sigint_handler)
         active_workers = self.num_chains
+        # To yield a deterministic ordering, we hold intermediate traces
+        # from each of the workers in its own queue in `results_buffer`
+        # and yield these in a round robin fashion.
+        cur_idx = 0
```
fritzo: nit: how about `current_chain` or `chain_idx`? I was a bit confused whether this indexed over workers or chains or time

neerajprad:
> Outside of this PR, I'm concerned that we're spin locking rather than blocking. Why not block?

Unfortunately, the issue with a blocking `queue.get` is that it also disregards KeyboardInterrupt, so will lead to situations where the user will need to kill off the processes manually to recover. We can however specify a small timeout, say `0.1s` so as to catch a KeyboardInterrupt - I am not doing that since the main process is pretty lightweight anyways, it just handles logging and aggregating traces from the workers and exits the loop as soon as any of the worker crashes or there is a KeyboardInterrupt.

pyro/infer/mcmc/mcmc.py
neerajprad: This is the round robin index, i.e. the index of the queue in the buffer we are currently blocked on and will be incremented with each `yield`. This might not be the same as the currently sampled chain, which is given by `chain_id`. Maybe we can call this `buffer_idx`?

pyro/infer/mcmc/mcmc.py
neerajprad: Thanks for the suggestion. The while loop can be outside, and this should work.

pyro/infer/mcmc/mcmc.py
neerajprad: This isn't needed - I will remove it.

pyro/infer/mcmc/mcmc.py
fritzo: How are these numbers related: number of workers, number of chains, and number of buffers?

fritzo:
> the main process is pretty lightweight anyways

But a process that spin locks still uses a full cpu core, right? But ok if you have 56 cores I guess 1 is lightweight...

pyro/infer/mcmc/mcmc.py
neerajprad: They are the same - `chain_id` refers to the id of the chain/worker from where the last trace is sampled. `(chain_id, trace)` is sent as a tuple by each of the workers. As soon as this sample is received it is put into its own queue in the buffer. Then we keep yielding samples from the buffer (popping off the queue indexed by `buffer_idx`, which is incremented) until we are blocked, at which point we call `.get_nowait` on the shared queue again.

neerajprad:
> But a process that spin locks still uses a full cpu core, right? Ok if you have 56 cores I guess 1 is lightweight...

That is true - I added a timeout of 1 s., and it seems to handle keyboard interrupts fine without deadlocking, so let me address it in this PR.

neerajprad:
Thanks for reviewing, @fritzo. There are still some rare cases when the logger thread is not able to clean up on KeyboardInterrupt, but it is not related to changing to a blocking get with timeout, so I will try to address that separately.

neerajprad:
For some reason, this seems to hang on travis, so I will need to debug to see what's happening.

fritzo:
@neerajprad feel free to remove the block-vs-spin change to unblock this PR. It's a minor issue and we can address it another time.

neerajprad:
Thanks, @fritzo. I have reverted the block with timeout change, but will address this in another PR.

PR #1507, opened 2018-10-31
Fix remaining issues in ubersum

### LINKED ISSUE #1460: Fix TraceEnum_ELBO.compute_marginals(), .sample_posterior()

fritzo:
Resolves #1460 

This fixes all known isses with `ubersum()`, `contract_to_tensor()`, and `contract_tensor_tree()`. The main issue was that we were simply broadcasting rather than performing forward-backward message passing up and down the plate tree. This PR completes our implementation of tensor message passing.

One subtle issue is that whereas `ubersum()` naturally aggregates across all batch slices (so that each batch of the output depends on all batches of the input), tensor message passing can avoid passing messages among batches that are not connected by an "upstream variable" (e.g. as a result of vectorized particles). To bridge this behavior gap, I've added an `modulo_total=True` flag to `ubersum()` to perform the cheaper tensor message passing algorithm. Thus users need to opt-in to this slightly subtle behavior.

## Tested

- [x] added lots of new tests to test_contract.py
- [x] refactored `contract_to_tensor()` and `contract_tensor()` are exercised by test_enum.py

### DIFF
--- a/pyro/infer/mcmc/util.py
+++ b/pyro/infer/mcmc/util.py
@@ -206,4 +206,4 @@ class TraceEinsumEvaluator(object):
         with shared_intermediates():
             log_probs = self._get_log_factors(model_trace)
             sum_dims = set.union(*self._enum_dims.values())
-            return contract_to_tensor(log_probs, sum_dims, frozenset())
+            return contract_to_tensor(log_probs, sum_dims)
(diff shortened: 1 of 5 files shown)


fritzo:
Thanks for reviewing @eb8680 and @martinjankowiak !

PR #1510, opened 2018-11-01
Minor optimizations to pyro.ops.contract

fritzo:
This makes some minor optimizations relevant to large tensors as used in e.g. the hmm example. This PR reduces the time spent in `contract_to_tensor()` from 19.4sec to 17.6sec.
```sh
python -m cProfile -o hmm.prof examples/hmm.py -n 200
snakeviz hmm.prof
```

## Tested

- this is a pure refactoring, hence covered by existing tests
- I've profiled using the above commands

### DIFF
--- a/pyro/ops/contract.py
+++ b/pyro/ops/contract.py
@@ -144,10 +144,10 @@ class TensorRing(object):
             return self._cache[key]
 
         term_sum = self.sumproduct([term], dims)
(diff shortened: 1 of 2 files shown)


fritzo:
The benefit is coming from the minor changes to contract.py. I think the memoization is only needed in guide enumeration with multiple plate nesting, whereas our hmm example does model enumeration.

PR #1512, opened 2018-11-02
Further optimize ubersum using modulo_toal

fritzo:


pyro/ops/contract.py
```diff
@@ -495,7 +491,10 @@ def contract_to_tensor(tensor_tree, sum_dims, target_ordinal=None, target_dims=N
     contracted_terms = []
 
     # Split this tensor tree into connected components.
+    modulo_total = bool(target_dims)
```
fritzo: @martinjankowiak the following few lines are the only new logic required for optimization.

fritzo:
> do we have any tests for modulo_total?

Yes, its correctness is already assessed by existing tests. I've also added one new test.

pyro/ops/contract.py
```diff
@@ -56,12 +56,14 @@ class TensorRing(object):
...
+    def _hash_by_id(self, tensor):
         """
-        Saves a tensor in the cache so that ``id(tensor)`` can be used as a
-        key in the cache without risk if the id being recycled.
+        Returns the id of a tensor and saves the tensor so that this id can be
+        used as a key in the cache without risk of the id being recycled.
         """
-        self._cache['tensor', id(tensor)] = tensor
+        result = id(tensor)
+        assert self._cache.setdefault(('tensor', result), tensor) is tensor
```
neerajprad: Is the cache bounded in any way, since we might be holding tensors in memory now that might have been previously GCd?

fritzo: The cache is not bounded. This is the einsum sharing cache whose lifetime is the duration of a single `ubersum()` call (or `contract_to_tensor()` call or `svi.step()`). That lifetime is required anyway for `.backward()`; while some CPU tensors may have been deleted, their underlying storage would still remain until after `.backward()`.

Also, any previous recycling was incorrect. I believe it is only the `modulo_total` trick in this PR that allowed recycling, and that is why I've included both bugfix + feature in this PR.

neerajprad:
@martinjankowiak - are you good with merging this?

martinjankowiak:
@neerajprad i'd just like to better understand where we are testing `modulo_total=False`
@fritzo ?

fritzo:
`modulo_total=False` is not implemented.

PR #1514, opened 2018-11-05
Added NUTS baseball tutorial

cglazner:
Added a tutorial notebook closely following the baseball example, in response to https://github.com/uber/pyro/issues/979. I did notice the problem in https://github.com/uber/pyro/issues/1458 cropping up on some runs but it works with this seed. Passes `make test-tutorials`.


neerajprad:
Thanks a lot for taking out the time to contribute this tutorial! I had a few general comments which I will leave here:

### Content:
 - I think we should mention upfront that users should refer to Bob Carpenter’s tutorial for details on the modeling choices; and that this tutorial will be largely focused on Pyro internals that are relevant for HMC, or MCMC, more generally (e.g. classes like `EmpiricalMarginal` or `TracePredictive` are not covered in other tutorials, and it will be great to use this as an avenue to educate our users about some of these classes).
 - Since HMC is still under development and lacks certain features like convergence diagnostics, we should mention this upfront and direct the users to https://github.com/uber/pyro/issues/1093, to set correct expectations on the level of support available, and known issues. Most of our internal efforts are focused around SVI, and HMC/NUTS has primarily been sustained through open source contributions! 
 - More details in *other functions* as they are introduced - e.g. `poutine.condition`, `EmpiricalMarginal`, `TracePredictive` (I think you have covered this later).

### Formatting:
 - Show the full dataset, which is quite small.
 - Align the last equation within each set, omit numbering.
 - Tabular result for each of the models - show predictions, posterior predictive density.

If you add me as a contributor to your branch, I will be happy to send you PRs addressing some of these, keeping this as the base PR. 

@fehiepsi has contributed a bulk of the HMC/NUTS code so it will be great to get his feedback as well.

fehiepsi:
@cglazner Thanks for making a tutorial for NUTS! I have several comments:
+ It is better to display output of the cells.
+ I think that we should add a section to quickly review HMC/NUTS, instead of citing reference.
+ The paragraph mentioning baseball data should be more revised. I think you should display the data, state the problem, and how to approach.
+ You can replace `.expand_by([num_players]).independent(1)` by using `with pyro.plate(...)`.
+ It is better to use `pyro.sample(..., obs=...)` instead of `pyro.condition(...)`. This way we don't have to explain what `pyro.condition` do.
+ It would be nice if we can have some plots. :)
+ Add a conclusion section to compare models.

fehiepsi:
@cglazner Thanks for addressing comments! The content looks great overall! It would be better if you address some weird lines such as SEJveChjaGlsZHJlbj0oSW50UHJvZ3Jlc3ModmFsdWU9MCwgZGVzY3JpcHRpb249dSdXYXJtdXAgWzFdJywgbWF4PTE1MCwgc3R5bGU9UHJvZ3Jlc3NTdHlsZShkZXNjcmlwdGlvbl93aWR0aD3igKY=
I guess this might come from tqdm?
It would be better if you address other formatting issues: only display relevant column of data and combine output of `run and evaluate` into 1 table: rows are person name, columns are model name, entries are mean +- std.

Btw, I think that @neerajprad @eb8680 will have better ideas how things should incorporate into pyro tutorials.

neerajprad:
> @fehiepsi , I like your suggestion to use pyro.plate but I got some pickling errors when I used the multiprocessing MCMC methods. I will change to use plate if I can figure that out.

Thanks @cglazner. I can look into this, let me try to run your notebook, and report back.

neerajprad:
@cglazner - I checked that using `pyro.plate` would indeed throw pickling error on pytorch-0.4. However, device serialization has been fixed in https://github.com/pytorch/pytorch/pull/7713, so we can make the change once we move to the latest version of PyTorch. I will create an issue as a reminder.

tutorial/source/baseball.ipynb
```diff
@@ -0,0 +1,3238 @@
...
+   "outputs": [
+    {
+     "data": {
+      "application/vnd.jupyter.widget-view+json": {
+       "model_id": "764f0d366ab846f4897fed2ce937e78d",
+       "version_major": 2,
+       "version_minor": 0
+      },
+      "text/plain": [
+       "SEJveChjaGlsZHJlbj0oSW50UHJvZ3Jlc3ModmFsdWU9MCwgZGVzY3JpcHRpb249dSdXYXJtdXAgWzFdJywgbWF4PTE1MCwgc3R5bGU9UHJvZ3Jlc3NTdHlsZShkZXNjcmlwdGlvbl93aWR0aD3igKY=\n"
```
neerajprad: I'm not quite sure why the bar is getting serialized this way for you. I am getting the following. Could you try updating your version of tqdm?

```
      "text/plain": [
       "HBox(children=(IntProgress(value=0, description='Warmup [1]', max=150), HTML(value='')))"
      ]
```

tutorial/source/baseball.ipynb
```diff
@@ -0,0 +1,3238 @@
...
...dx04TJGMnnut7OpBqrIUJpXbV3GRSta+OZ2a+yjTpvtTIZLLdXtxbOnb4TVHXHWdCaI18je0AtKy1BLkhHOjOWcAsNMvlghTfS2xaYSs1g057yXnpYofp84AVGAHzxxjJd/5kFHw9fB7YtXtlJU1s5Wy37rliSJh/wcHUzZqdPBslYeAFes7XB+nmrMAqy4Rb6oeOLIgJNKPl68NVSa37mwm8FUjof29vPs8SGu+fsHeM83dzqP6zW4poaxaIlaMy2qxX7KYxal/lXV0BuhqKcT9Uq7oSCU4qO9VTwLgKs3dBLwCd963OpfqWvKkpGAY7iOD1ZP758odU21HmWqlLptSr9lHtAWD5XtuGsZi6QrA0nnPw+n8879y1qjDIzlGM1Y7YDTueK0eBZg7Vh22bnt+gJvpcDlCOZ9JCOVizPg9/HzD1wHwPYDp/n6Lw858pKWrqoZi/ZEuCxHe6CGZwFw+dp2OhJh7t5xlOs2LeGxA6cJ2IbMi4hw5fpO/vn+PYxk8mXjZL0t1ff1jzhNFxPhQNVOnt7dV7OdOjyUytGeCNvGwutZxLj36eOTmhS2mM55L0G/j2Weoq/7nunjqSNDHDg1yrolSSe4vW1VGw/s7ufImZRrZxpmaXOEYwNp4zbQ5gAAHlZJREFUCkpVyJAAV6y3jEV7PFQWz5gsOgX90OkUl6yannV25bpOmqNB/uWn+9nTN8yZsRyP7j/lTHrc3z+KCKyokvYL1iapqKzrQrNnzQynrZqsWMhPNOjHJ+OQoULlxmJVe5wfPnWcVLbgxI16angWzdEgl6xq42E7Y7Hdkzqr2xHB1I3FuGIWIrJBRG4VkXtF5H79NaXfPMtYzQRLrmP/cKYiEwpw1TZYF9psvkgmX3SCzU767ECKfn0hnCbPojUWchru6Qu3DlQNp/MV8Qov7Z7urzqNtbqxCJXVRQyO5Wqm9vp9wssv6Ob+3Sd46ace5Me7+nj7VWsqDJdmm70r3fH8gGswVMTZ6Z0ayZLJFzh4atTJt4+F/PVlKFuS04tTG8J0rlAxWnJ5W5RcQU2pknsxnPPVWNEW43lXHr/enOw+bmWi6XYe21a2Albcwn2x6WmJcmwobfcGqzxflrVEWdMRH1cdxXhY3WHNboD6abMTIRTwceO5S3nsudNEgn7++Nq1DKXzjvTz3MlRelujNYPpTk1UlZYfekCTi
```
neerajprad: Unfortunately, this isn't correct, since chain information is not preserved when traces are yielded. I would suggest simply removing this part, and we can update our tutorial once chain diagnostics and any chain related refactoring is implemented.

neerajprad: This is a very good point, but I am wondering if there is an alternate visualization that we could use instead of a line plot that might be less confusing - e.g. plotting the distribution of batting averages in test/training and the posteriors. It will also be nice to plot the actuals vs predictions as a 90% CI for each of the players.

neerajprad: I would also suggest using seaborn to get better defaults for plotting. [example tutorial](https://github.com/uber/pyro/blob/6d34843959bf9553b922a9fc642b68e89b237f85/tutorial/source/reg_example-gdp_terrain_ruggedness.ipynb).

tutorial/source/baseball.ipynb
cglazner: The base64 hash appeared in tqdm 4.27 and 4.28 but 4.26 was ok.

neerajprad: Hmm..maybe we should just remove our dependency on tqdm. I will work on a simple progress bar implementation and see if I can get it to work well with jupyter.

neerajprad:
Thanks for updating! I will take a closer look at this over the break.

fritzo:
closing as abandoned.

Issue #1511, opened 2018-11-02
HMM model using HMC/NUTS is slow


neerajprad:
Since we started using einsum to evaluate the joint log density of the trace with discrete parameters, we can do discrete site enumeration across many more models in HMC/NUTS without going OOM. 

For HMMs, however, NUTS / HMC is extremely slow to the point of being unusable beyond a few time steps, say 10. Refer to this [test](https://github.com/uber/pyro/blob/dev/tests/infer/mcmc/test_nuts.py#L243) for profiling.

There are a few issues that I noticed:
 - ~NUTS crawls to a halt due to extremely small step size in the warmup phase itself. I suppose the issue might be with starting from a random initialization, but I am not sure. It might be worth initializing with the MAP estimate and then running NUTS to not hang with unworkably small step size. I think we have seen this issue with a few other models too. e.g. #1470. cc. @fehiepsi~ I don't observe this issue on pytorch 1.0.
 - Converting the values returned from the integrator into a model trace (`_get_trace`) also takes more than 3s. While comparatively small, I believe this can be optimized if we assume our models to be static by assuming a different data structure inside HMC, so that we do not need to run the model each time. 
 - The trace log density evaluation (and gradient computation) takes the bulk of the time, as expected. It is not immediately clear how this can be improved given that we need to call this many times per integrator step to generate even a single sample, and it does seem like NUTS will continue to be slow until we can make this much faster. Profiling viz below.

<img width="1134" alt="screen shot 2018-11-01 at 5 24 41 pm" src="https://user-images.githubusercontent.com/1762463/47887135-142fb200-ddfb-11e8-919e-ceb38f0c39dc.png">

fehiepsi:
@neerajprad I tested with pytorch 0.4.0 and didn't catch the slowness. As I mentioned in #1487, trace + Gamma/Dirichlet + pytorch 1.0rc will give wrong grads in backward pass. And that bug is not related to HMC/NUTS. I don't know what I can do more with that issue so I skip it and implement statistics such as effective number of samples / Gelman-Rubin convergence diagnostic instead.

fehiepsi:
[...]

PR #1516, opened 2018-11-05
Cache potential energy and gradient of last trace in HMC/NUTS

neerajprad:
Minor optimization as a result of investigating #1511. 

Since the log density and gradient computation can take much longer with traces containing discrete sites, it is best to avoid it whenever possible. This includes a minor optimization - i.e. caching the potential energy and grads of the last seen trace (we already do caching within the same kernel `sample` step but not across steps).

**Perf Impact:**
The benefits are typically higher for HMC than NUTS, because NUTS tends to build longer chains and therefore the impact of caching the potential energy and gradient values for the beginning of the chain can be small. On the HMM example with 30 time steps with NUTS, this results in a reduction in runtime from 3min 28 seconds to 3min 19 seconds.

### DIFF
--- a/pyro/ops/integrator.py
+++ b/pyro/ops/integrator.py
@@ -21,35 +21,23 @@ def velocity_verlet(z, r, potential_fn, inverse_mass_matrix, step_size, num_step
         Here :math:`M` can be a 1D tensor (diagonal matrix) or a 2D tensor (dense matrix).
     :param float step_size: step size for each time step iteration.
     :param int num_steps: number of discrete time steps over which to integrate.
-    :return tuple (z_next, r_next): final position and momenta, having same types as (z, r).
-    """
-    z_next = z.copy()
-    r_next = r.copy()
-    for _ in range(num_steps):
-        z_next, r_next, z_grads, potential_energy = _single_step_inplace_verlet(z_next,
-                                                                                r_next,
-                                                                                potential_fn,
-                                                                                inverse_mass_matrix,
-                                                                                step_size,
-                                                                                z_grads)
-    return z_next, r_next, z_grads, potential_energy
-
-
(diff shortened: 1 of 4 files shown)


pyro/ops/integrator.py
```diff
@@ -56,30 +46,38 @@ def single_step_velocity_verlet(z, r, potential_fn, inverse_mass_matrix, step_si
...
+        r[site_name] = r[site_name] + 0.5 * step_size * (-z_grads[site_name])  # r(n+1)
+
+    return z, r, z_grads, potential_energy
 
 
 def _potential_grad(potential_fn, z):
     z_keys, z_nodes = zip(*z.items())
     for node in z_nodes:
-        node.requires_grad = True
+        node.requires_grad_(True)
```
fehiepsi: +1

pyro/ops/integrator.py
```diff
@@ -56,30 +46,38 @@ def single_step_velocity_verlet(z, r, potential_fn, inverse_mass_matrix, step_si
     """
     z_next = z.copy()
     r_next = r.copy()
-    z_grads = _potential_grad(potential_fn, z_next)[0] if z_grads is None else z_grads
+    return _single_step_inplace_verlet(z_next, r_next, potential_fn, inverse_mass_matrix, step_size, z_grads)
```
fehiepsi: nit: we might call `velocity_verlet` with `num_steps=1` here

neerajprad: Good point. Yeah, from what I can see, we now don't need `single_step_velocity_verlet`; will refactor.

neerajprad:
This is good to merge, pending any further comments.

neerajprad:
@jpchen - could you help merge this PR (reviewed by @fehiepsi)? Feel free to review too!

PR #1521, opened 2018-11-08
Implement pyro.markov to recycle enum dims

### LINKED ISSUE #1380: Work around torch.cuda.Tensor maximum dim limit

fritzo:
Resolves #1380 
Pair coded with @eb8680 

This PR aims to resolve the upcoming dimension limits in PyTorch 1.0, which would otherwise break parallel enumeration of many variables, e.g. examples/hmm.py. The solution is to recycle enum dims and use symbolic names to distinguish colliding dims.

**Solution part 1** is to add a tiny bit of annotation `pyro.markov` to user code, which can be done in a variety of ways. The core idea is to provide a windowed dependency arena, so that when variables go out of context they are assumed conditionally independent, and their enumeration dimensions can be reused. We still keep a unique symbolic name for each dim. The symbolic and numeric dims are managed jointly by a new `_ENUM_ALLOCATOR`, an outermoset `EnumMessenger`, and zero or more `MarkovMessenger`s that can be nested inside the `EnumMessenger`.

**Solution part 2**  The second part is to switch implementation of contraction machinery from `UnpackedLogRing` to `PackedLogRing` with named dimensions. This PR also introduces a pyro.ops.packed module to handle a few packed tensor operations outside of the `contract_*` functions. (I may pull this out into a separate PR if this PR is too large).

## Limitations

- `MarkovMesenger` does not work correctly with `poutine.block`; @eb8680 may add support for block in a later PR.

## Tested

- [x] added lots of tests to test_valid_models.py
- [x] update tests/infer/test_enum.py to use `pyro.markov`
- [x] update examples/hmm.py to use `pyro.markov`


examples/hmm.py
```diff
@@ -72,15 +72,15 @@ def model_1(sequences, lengths, args, batch_size=None, include_prior=True):
     with pyro.plate("sequences", len(sequences), batch_size, dim=-2) as batch:
         lengths = lengths[batch]
         x = 0
-        for t in range(lengths.max()):
+        for t in pyro.markov(range(lengths.max())):
```
martinjankowiak: 💯

fritzo:
Ready to merge as soon as tests pass.

fritzo:
@srush as of this PR you should be able to run CRFs and HMMs in CUDA without hitting dimension limits. E.g. I noticed your EMNLP tutorial was limited to 20 time steps, but by changing your [HiddenMarkovModel](https://colab.research.google.com/drive/1b522fQGUdFI8Wl840D_XsevYYdbIAlrl#scrollTo=StxHKIYJqVXO) to use `pyro.markov` you should be able to use any limit:
```diff
- for t in range(T):
+ for t in pyro.markov(range(T)):
```

Issue #1520, opened 2018-11-08
Refactoring MCMC module


neerajprad:
We have added a lot of important features to HMC while addressing #1093. This has come at the expense of class complexity, particularly for HMC and MCMC classes. It will be great to refactor some of this code to make it amenable to extend, understand and debug.
 - [x] Remove `use_experimental_einsum` flag and make `TraceEinsumEvaluator` the default in HMC/NUTS.
 - [x] Refactoring the adaptation code to be separate from the HMC class. This currently toggles a lot of mutable attributes that makes debugging difficult. 
 - [x] Making it easy for the user to specify an initial trace to begin sampling from (based on @fehiepsi's suggestion in #1511).
 - [ ] Own implementation of progress bar that is specialized to our use case than having to work around tqdm's (which is a great general purpose progress bar otherwise) quirks. - Low Priority
 - [x] Make progress bar consistent when number of decimals of step_size change. - Low Priority
 - [x] Blocked call with timeout when reading from the shared queue in ParallelSampler.
 - [x] Refactoring the TracePosterior / ParallelSampler classes to support chain level diagnostics.
 - [ ] Refactoring to support parallel chains on CUDA - we will need to hold the traces temporarily within each worker until its read by the main process. Ref. to discussions [#1](https://discuss.pytorch.org/t/invalid-device-pointer-using-multiprocessing-with-cuda/28023/6), [#2](https://discuss.pytorch.org/t/cuda-tensors-on-multiprocessing-queue/28626/2)
 - [ ] Any other?

neerajprad:
Removing from the milestone - the only major task left is to support parallel chaining with CUDA.

fehiepsi:
I think we can merge this and https://github.com/uber/pyro/issues/1093. In issue 1093, it seems that there is only one remaining issue: "Detect divergent transitions, and log them for analysis. Throwing a NaN here during sampling as we might do now with validation check enabled, is not very useful to the end user."

PR #1522, opened 2018-11-08
Abstract out HMC adaptation into its own class

neerajprad:
Addresses #1520. 

This refactors the step size and mass matrix adaptation into its own class. The motivation for this is to reduce the complexity in the HMC class that arises out of maintaining many stateful attributes during adaptation. The logic that previously resided here has been refactored to keep mutable attributes to a minimum, and moved to `WarmupAdapter`. 

We are doing this by constructing an adaptation schedule instead and then moving within this schedule to take appropriate actions as needed. One minor departure from earlier might be that the slow adaptation windows are always monotonically increasing, i.e. the last slow window (before the end buffer) cannot be smaller than the penultimate one. This is to ensure that we are collecting enough number of samples for estimating the covariance.

### Testing
Test is added to check if the adaptation schedule is built correctly. Remaining tests should pass as is (or with minor changes) since this is a straight up refactor.

### TODOs:
 - [x] Ensure that there is no perf regression.
 - [x] Understand test failures, if any; ensure that logic is largely the same as earlier.

### DIFF
--- a/pyro/infer/mcmc/adaptation.py
+++ b/pyro/infer/mcmc/adaptation.py
@@ -1,0 +1,163 @@
+import math
+import torch
+from collections import namedtuple
+
+import pyro.distributions as dist
+from pyro.ops.dual_averaging import DualAveraging
+from pyro.ops.welford import WelfordCovariance
+
+adapt_window = namedtuple("adapt_window", ["start", "end"])
+
+
+class WarmupAdapter(object):
+    r"""
+    Adapts tunable parameters, namely step size and mass matrix, during the
+    warmup phase. This class provides lookup properties to read the latest
+    values of ``step_size`` and ``inverse_mass_matrix``. These values are
+    periodically updated when adaptation is engaged.
+    """
+
+    def __init__(self,
+                 step_size,
+                 adapt_step_size=False,
+                 target_accept_prob=0.8,
+                 adapt_mass_matrix=False,
+                 is_diag_mass=True):
+        self.adapt_step_size = adapt_step_size
+        self.adapt_mass_matrix = adapt_mass_matrix
+        self.target_accept_prob = target_accept_prob
+        self.is_diag_mass = is_diag_mass
+        self._step_size = step_size if step_size is not None else 1
+        self._adaptation_disabled = not (adapt_step_size or adapt_mass_matrix)
+        if adapt_step_size:
+            self._step_size_adapt_scheme = DualAveraging()
+        if adapt_mass_matrix:
+            self._mass_matrix_adapt_scheme = WelfordCovariance(diagonal=is_diag_mass)
+
+        # We separate warmup_steps into windows:
+        #   start_buffer + window 1 + window 2 + window 3 + ... + end_buffer
+        # where the length of each window will be doubled for the next window.
+        # We won't adapt mass matrix during start and end buffers; and mass
+        # matrix will be updated at the end of each window. This is helpful
+        # for dealing with the intense computation of sampling momentum from the
+        # inverse of mass matrix.
+        self._adapt_start_buffer = 75  # from Stan
+        self._adapt_end_buffer = 50  # from Stan
+        self._adapt_initial_window = 25  # from Stan
+        self._current_window = 0  # starting window index
+
+        # configured later on setup
+        self._warmup_steps = None
+        self._inverse_mass_matrix = None
+        self._r_dist = None
+        self._adaptation_schedule = []
+
+    def _build_adaptation_schedule(self):
+        adaptation_schedule = []
+        # from Stan, for small warmup_steps < 20
+        if self._warmup_steps <= 20:
+            adaptation_schedule.append(adapt_window(0, self._warmup_steps - 1))
+            return adaptation_schedule
+
+        start_buffer_size = self._adapt_start_buffer
+        end_buffer_size = self._adapt_end_buffer
+        init_window_size = self._adapt_initial_window
+        if (self._adapt_start_buffer + self._adapt_end_buffer
+                + self._adapt_initial_window > self._warmup_steps):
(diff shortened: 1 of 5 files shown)


pyro/infer/mcmc/hmc.py
```diff
@@ -416,27 +343,21 @@ def sample(self, trace):
             energy_current = self._kinetic_energy(r) + potential_energy if potential_energy is not None \
                 else self._energy(z, r)
         delta_energy = energy_proposal - energy_current
+        # Set accept prob to 0.0 if delta_energy is `NaN` which may be
+        # the case for a diverging trajectory when using a large step size.
+        if torch_isnan(delta_energy):
+            accept_prob = delta_energy.new_tensor(0.0)
+        else:
+            accept_prob = (-delta_energy).exp().clamp(max=1.)
```
fehiepsi: +1 for moving the check to here, it is easier to follow the code

pyro/infer/mcmc/hmc.py
```diff
@@ -227,102 +209,36 @@ def _find_reasonable_step_size(self, z):
...
-        r_flat = pyro.sample(name, self._r_dist)
+        r_dist = self._adapter.r_dist
+        # For finding a reasonable step size before adapter is initialized.
+        if r_dist is None:
+            inv_mass_matrix = self._initial_mass_matrix
+            loc = inv_mass_matrix.new_zeros(inv_mass_matrix.size(0))
+            if self.full_mass:
+                r_dist = dist.MultivariateNormal(loc, precision_matrix=inv_mass_matrix)
+            else:
+                r_dist = dist.Normal(loc, inv_mass_matrix.rsqrt())
```
fehiepsi: I think that we can move this logic to Adapter too.

neerajprad: `r_dist` is already moved to adapter. This is one part that I wish could be more cleanly separted - the issue here is that some components of the adapter (those that rely on the model trace like `inverse_mass_matrix` etc.) are not initialized until we hit `setup._configure_adaptation` but that method itself requires an `initial_step_size` for which we need to hit `find_reasonable_step_size`. None of the alternatives seemed cleaner :( - one of which would be to initialize the adapter in three stages: constructing it, then initializing the mass matrix and then finally setting the initial step size.

neerajprad: Btw @fehiepsi - Feel free to directly commit to this PR if you notice something off!

fehiepsi: I think that you can revise the logic of `adapter.configure` like:
```
if warmup_steps is not None:
    update schedule
if step_size is not None:
    update step_size
if inv_mass_matrix is not None:
    update inv_mass_matrix
```
+ then remove `_configure_adaptation`, putting the following logic 
```
self._adapter.configure(warmup_steps=warmup_steps, inv_mass_matrix=initial_mass_matrix)
if self.adapt_step_size:
    with pyro.validation_enabled(False):
          initial_step_size = self._find_reasonable_step_size(trace)  # trace instead of z
    self._adapter.configure(step_size=initial_step_size)
```
at the end of `setup`.

This way, you can remove `self._initial_mass_matrix` in HMC I guess.

neerajprad:
Thanks a lot for the detailed review, @fehiepsi! Let me try to address all your comments, and report back if needed.

pyro/infer/mcmc/adaptation.py
```diff
@@ -0,0 +1,183 @@
...
+                return
+
+            if self._current_window == num_windows - 1:
+                self._current_window += 1
+                self._end_adaptation()
+                return
+
+            if self.adapt_step_size:
+                self._step_size_adapt_scheme.prox_center = math.log(10 * self.step_size)
+                self._step_size_adapt_scheme.reset()
```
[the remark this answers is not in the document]
neerajprad: +1. I will do that. `_reset_step_size` is also called from `configure`, so will be best to keep it.

pyro/infer/mcmc/hmc.py
neerajprad: I think at this point, let us just have separate setters - and then assert that we have a well specified configuration inside `configure` method. The issue with multiple optional parameters is that we won't be able to verify that the configuration is incomplete. Will make the change.

pyro/infer/mcmc/adaptation.py
```diff
@@ -0,0 +1,183 @@
...
+        :param int t: time step, beginning at 0.
+        :param dict z: latent variables.
+        :param float accept_prob: acceptance probability of the proposal.
+        """
+        if t > self._warmup_steps or self._adaptation_disabled:
+            return
+        window = self._adaptation_schedule[self._current_window]
+        num_windows = len(self._adaptation_schedule)
+        mass_matrix_adaptation_phase = self.adapt_mass_matrix and \
+            (0 < self._current_window < num_windows - 1)
```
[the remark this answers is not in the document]
neerajprad: That was the intention. :) I think having a lot of mutable state allows for the most efficient code in python, but that comes at the expense of debuggability and hidden bugs. As such, I have tried to keep these attributes to a minimum. In this case, as you rightly pointed out, storing this won't affect the runtime.

pyro/infer/mcmc/adaptation.py
```diff
@@ -0,0 +1,183 @@
...
+        if self.adapt_step_size:
+            self._update_step_size(accept_prob)
+        if mass_matrix_adaptation_phase:
+            z_flat = torch.cat([z[name].reshape(-1) for name in sorted(z)])
+            self._mass_matrix_adapt_scheme.update(z_flat.detach())
+        if t == window.end:
```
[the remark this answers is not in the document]
neerajprad: This part is smaller now, and I think it is easier to read if this logic is all in one place, so I am leaving it here.

PR #1525, opened 2018-11-10
 Add ops stats for statistical functions

fehiepsi:
This pull request implements the following stats:

- [x] ~density~ it seems unnecessary
- [x] resample
- [x] quantile
- [x] percentage interval
- [x] highest posterior density interval
- [x] widely applicable information criterion
- [x] ~model selection~ unnecessary
- [x] tests comparing with Stan/PyMC3


fehiepsi:
Writing tests for all of these functions would be exhausting. >"< So I'll split this into sequential pull requests.

fehiepsi:
@neerajprad I add several stats ops which ultimately will be called by methods in Empirical class. I will add WAIC in another PR (together with effective number of params) because I think that it should be a method of TracePosterior instance, rather than a method of Empirical instance.

pyro/ops/stats.py
```diff
@@ -1,18 +1,21 @@
...
-    var_between = chain_mean.var(dim=0)
-
     chain_var = input.var(dim=0)
     var_within = chain_var.mean(dim=0)
-    var_estimator = (N - 1) / N * var_within + var_between
+    var_estimator = (N - 1) / N * var_within
+    if input.size(1) > 1:
+        chain_mean = input.mean(dim=0)
+        var_between = chain_mean.var(dim=0)
+        var_estimator = var_estimator + var_between
```
fehiepsi: Different from arviz, in Stan, they allow computing n_eff with [num_chains = 1](https://github.com/stan-dev/stan/blob/develop/src/stan/analyze/mcmc/compute_effective_sample_size.hpp#L49).

neerajprad:
@fehiepsi - Is this blocked on #1557?

fehiepsi:
@neerajprad It is not blocked by that PR and also does not block other PRs. So please take this low priority. There is a failed test but it is easy to fix though.

fehiepsi:
@neerajprad I have fixed the test and updated some nits at docs. This is good to review now. :)

tests/ops/test_stats.py
```diff
@@ -1,12 +1,90 @@
...
+def _pi(x, dim=0):
+    return pi(x, prob=0.8, dim=dim)
+
+
+def _hpdi(x, dim=0):
+    return hpdi(x, prob=0.8, dim=dim)
+
+
+@pytest.mark.parametrize('statistics', [_quantile, _pi, _hpdi])
+@pytest.mark.parametrize('sample_shape', [(), (3,), (2, 3)])
```
neerajprad: Any reason why we can't add these to the parameters below?

tests/ops/test_stats.py
```diff
@@ -1,12 +1,90 @@
...
+        assert_equal(torch.unique(z.reshape(-1)).numel(), z.numel())
+    assert_equal(y.shape, torch.Size([num_samples, 2]))
+    assert_equal(z.shape, torch.Size([2, num_samples]))
+    assert_equal(y.mean(dim=0), torch.tensor([3., 5.]), prec=0.1)
+    assert_equal(z.mean(dim=1), torch.tensor([3., 5.]), prec=0.12)
+    assert_equal(y.std(dim=0), torch.tensor([4., 6.]), prec=0.1)
+    assert_equal(z.std(dim=1), torch.tensor([4., 6.]), prec=0.1)
+
+
+@pytest.mark.init(rng_seed=3)
```
neerajprad: Why does this need any rng seed?

tests/ops/test_stats.py
```diff
@@ -1,12 +1,90 @@
...
+
+    num_samples = 5000
+    y = resample(x, num_samples=num_samples, replacement=replacement)
+    z = resample(x.t(), num_samples=num_samples, dim=1, replacement=replacement)
+    if not replacement:
+        assert_equal(torch.unique(y.reshape(-1)).numel(), y.numel())
+        assert_equal(torch.unique(z.reshape(-1)).numel(), z.numel())
+    assert_equal(y.shape, torch.Size([num_samples, 2]))
+    assert_equal(z.shape, torch.Size([2, num_samples]))
+    assert_equal(y.mean(dim=0), torch.tensor([3., 5.]), prec=0.1)
```
neerajprad: Could you also run these tests on CUDA and confirm that they pass?

tests/infer/test_abstract_infer.py
```diff
@@ -44,3 +47,30 @@ def nested():
         nested()
 
     assert len(tp.trace.nodes) == 0
+
+
+@pytest.mark.init(rng_seed=3)
```
neerajprad: How variable are the results with the default seed? Is it possible to relax the thresholds instead? A consequence of adding as many tests as we have now, is that we need to be careful in ensuring that they can run robustly under different environments - CUDA, platform etc, otherwise maintaining them will be challenge.

pyro/ops/stats.py
```diff
@@ -1,30 +1,36 @@
...
     # input has shape N x C x sample_shape
     N = input.size(0)
-    chain_mean = input.mean(dim=0)
-    var_between = chain_mean.var(dim=0)
-
     chain_var = input.var(dim=0)
     var_within = chain_var.mean(dim=0)
-    var_estimator = (N - 1) / N * var_within + var_between
+    var_estimator = (N - 1) / N * var_within
+    if input.size(1) > 1:
```
neerajprad: Will this ever be false, since it seems that we are asserting below that `input.size(chain_dim) >= 2`?

pyro/ops/stats.py
```diff
@@ -213,3 +218,108 @@ def effective_sample_size(input, chain_dim=0, sample_dim=1):
...
+    """
+    Draws ``num_samples`` samples from ``input`` at dimension ``dim``.
+
+    :param torch.Tensor input: the input tensor.
+    :param int num_samples: the number of samples to draw from ``input``.
+    :param int dim: dimension to draw from ``input``.
+    :returns torch.Tensor: samples drawn randomly from ``input``.
+    """
+    weights = input.new_ones(input.size(dim))
+    indices = torch.multinomial(weights, num_samples, replacement)
```
neerajprad: Nice!

tests/ops/test_stats.py
fehiepsi: @neerajprad `y` is generated uniformly. But I observed that the generated samples are flaky somehow (even if we use a number much larger than `2000`).

tests/ops/test_stats.py
fehiepsi: These statistics generate results with shapes changed. The `autocovariance` e.g. gives the same shape.

tests/infer/test_abstract_infer.py
fehiepsi: @neerajprad As long as AutoDelta is able to find the model's MAP points. Then we can get the correct WAIC. However, depending on seed (hence on initial point of AutoDelta), it took us longer to converge. I'll find good lr and num_steps which ensures converging at with different seeds though.

pyro/ops/stats.py
fehiepsi: @neerajprad Both `split_gelman_rubin` and `effective_sample_size` don't require `input.size(chain_dim) >= 2`. So we can get `diagnostics` in HMC even with `num_chains=1`.

pyro/infer/abstract_infer.py
```diff
@@ -87,6 +89,44 @@ def run(self, *args, **kwargs):
...
+            parameters.
+        """
+        if not self.exec_traces:
+            return {}
+        obs_node = None
+        log_likelihoods = []
+        for trace in self.exec_traces:
+            obs_nodes = trace.observation_nodes
+            if len(obs_nodes) > 1:
+                raise ValueError("Infomation criterion calculation only works for models "
```
neerajprad: It seems that we should be able to flatten all our observations and support `len(obs_nodes) > 1`, right? Not saying we should do that, but just want to understand where this restriction is coming from.

tests/ops/test_stats.py
```diff
@@ -116,3 +194,20 @@ def test_diagnostics_ok_with_sample_shape(diagnostics, sample_shape):
         assert_equal(diagnostics(a, chain_dim=1, sample_dim=0), y)
         assert_equal(diagnostics(b, chain_dim=-1, sample_dim=0), y)
         assert_equal(diagnostics(c, sample_dim=-1), y)
+
+
+def test_waic():
+    x = - torch.arange(1., 101).log().reshape(25, 4)
```
neerajprad: Are there any restrictions to the shape of `obs_nodes`? Can we test that other arbitrary shapes do not throw an error?

tests/ops/test_stats.py
fehiepsi: @neerajprad It is sad that I don't have a GPU machine at my hand.

neerajprad: Np at all. Let me test and get back on this.

pyro/infer/abstract_infer.py
fehiepsi: @neerajprad I can't find any documentation for that but your suggestion is quite reasonable! Will make a change if you (or other users in the future) need it.

tests/ops/test_stats.py
fehiepsi: There is no restriction for it. I'll make a test soon.

pyro/infer/abstract_infer.py
```diff
@@ -87,6 +89,44 @@ def run(self, *args, **kwargs):
         self._categorical = Categorical(logits=torch.tensor(self.log_weights))
         return self
 
+    def information_criterion(self, pointwise=False):
```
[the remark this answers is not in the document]
fehiepsi: This belongs to TracePosterior class indeed. People uses this with MAP inference too.

pyro/infer/abstract_infer.py
fehiepsi: Ah, indeed I don't know what we should do with the case of many observations. In its theory, each sample gives a likelihood. And waic is calculated from a batch of samples (with a batch of likelihoods). If we allow many observations, I am not sure that we should concatenate their likelihoods, or sum them, or take average of them. >"<

tests/ops/test_stats.py
neerajprad: We might have to bump this to 0.2 for CUDA.

tests/ops/test_stats.py
```diff
@@ -37,7 +115,7 @@ def test_cummin():
 
 @pytest.mark.parametrize('statistics', [autocorrelation, autocovariance, _cummin])
 @pytest.mark.parametrize('sample_shape', [(), (3,), (2, 3)])
-def test_statistics_ok_with_sample_shape(statistics, sample_shape):
+def test_statistics_B_ok_with_sample_shape(statistics, sample_shape):
```
neerajprad: Not for this PR, but `test_effective_sample_size` gives 107.85 which is very different from the expected value of 134.5. I remember you had a fix for CPU, but I am not sure why that's not working for CUDA.

fehiepsi: It might be that cuda uses a different fft library. I'll try to find some cuda resources to test it.

pyro/infer/abstract_infer.py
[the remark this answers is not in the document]
neerajprad: TracePosterior is also used by Importance though so if we put it there we need to add checks to validate that the weights are all uniform, or incorporate weights into this method.

tests/ops/test_stats.py
neerajprad: I see; I think we can defer it to later.

fehiepsi:
@neerajprad I will replace the random procedure to generate correlated samples by some deterministic one. CPU and GPU give different random numbers given the same seed: https://discuss.pytorch.org/t/best-practices-for-seeding-random-numbers-on-gpu/18751

tests/infer/test_abstract_infer.py
```diff
@@ -44,3 +46,30 @@ def nested():
...
+
+    def model():
+        mu = pyro.sample("mu", dist.Normal(kcal_mean, 1))
+        log_sigma = pyro.sample("log_sigma", dist.Normal(kcal_logstd, 1))
+        with pyro.plate("plate"):
+            pyro.sample("kcal", dist.Normal(mu, log_sigma.exp()), obs=kcal)
+
+    delta_guide = AutoLaplaceApproximation(model)
+
+    svi = SVI(model, delta_guide, optim.Adam({"lr": 0.05}), loss=Trace_ELBO(), num_samples=5000)
```
fehiepsi: @neerajprad I change `num_samples` to 5000 and delete the seed to have more stable result. I test this will different runs in CPU. Hope that it would be fine for GPU too.

tests/infer/test_abstract_infer.py
fehiepsi: @neerajprad Correction: with this lr, model converges for different seeds. But with different seeds, we have different posterior samples. That causes different results in waic. Sorry for confusion. :(

pyro/infer/abstract_infer.py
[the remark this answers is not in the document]
fehiepsi: I have added it in the lastest commit.

tests/infer/test_abstract_infer.py
neerajprad: Don't worry about the GPU, @fehiepsi.

fehiepsi:
Yes, it is ready if tests pass. Thanks a lot for your detailed reviews!

PR #1526, opened 2018-11-11
seed contrib.bnn with a hiddenlayer distribution

martinjankowiak:
this is a distribution that represents a single hidden layer in a bayesian neural network. see doc string. it can also be used in other settings like bayesian logistic regression (which after all is just a very simple neural network). 

both the `rsample` and `log_prob` method have accompanying tests. this code has also been tested in various experiments i've done in a different context.

### DIFF
--- a/pyro/contrib/bnn/hidden_layer.py
+++ b/pyro/contrib/bnn/hidden_layer.py
@@ -1,0 +1,175 @@
+from __future__ import absolute_import, division, print_function
+
+import math
+
+import torch
+from torch.distributions import Categorical
+from torch.distributions.utils import lazy_property
+
+from pyro.contrib.bnn.utils import adjoin_ones_vector
+from pyro.distributions.torch_distribution import TorchDistribution
+
+
+class HiddenLayer(TorchDistribution):
+    r"""
+    :param torch.Tensor X: B x D dimensional mini-batch of input
+    :param torch.Tensor A_scale: D x H     dimensional (diagonal covariance) or
+                                 D x D x H dimensional (cholesky factorization) or
+                                 D x K x H dimensional (mixture of diagonal normals)
+    :param torch.Tensor A_logits: H x K dimensional (for mixture of diagonal normals)
+    """
+    has_rsample = True
+
+    def __init__(self, X=None, A_mean=None, A_scale=None, A_logits=None, non_linearity="relu",
+                 KL_factor=1.0, A_prior_scale=1.0, include_hidden_bias=True,
+                 weight_space_sampling=False, relu_shift=None, relu_scale=None, N_KL_samples=5000):
+        self.X = X
+        self.dim_X = X.size(-1)
+        self.dim_H = A_mean.size(-1)
+        # assert A_mean.shape == A_scale.shape
+        #assert A_mean.size(0) == self.dim_X
+        self.A_mean = A_mean
+        self.A_scale = A_scale
+        self.A_logits = A_logits
+        self.non_linearity = non_linearity
+        if A_scale.dim()==2:
+            self.A_covariance = "diagonal"
+        elif A_scale.dim() == 3 and self.A_logits is not None:
+            self.A_covariance = "mixture"
+            self.K = A_scale.size(-2)
+            self.A_categorical = Categorical(logits=A_logits)
+        elif A_scale.dim() == 3 and self.A_logits is None:
+            self.A_covariance = "cholesky"
+        self.KL_factor = KL_factor
+        self.A_prior_scale = A_prior_scale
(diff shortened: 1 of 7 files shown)


pyro/contrib/bnn/hidden_layer.py
```diff
@@ -0,0 +1,122 @@
...
+
+import torch
+from torch.distributions.utils import lazy_property
+import torch.nn.functional as F
+
+from pyro.contrib.bnn.utils import adjoin_ones_vector
+from pyro.distributions.torch_distribution import TorchDistribution
+
+
+class HiddenLayer(TorchDistribution):
```
fritzo: Should this also inherit from `torch.nn.Module`?

pyro/contrib/bnn/utils.py
```diff
@@ -0,0 +1,16 @@
+import torch
```
fritzo: nit: from future import ...

By ensuring all files use the same programming language, we make it safer to debug and refactor by moving code among files.

tests/contrib/bnn/test_hidden_layer.py
```diff
@@ -0,0 +1,51 @@
...
+    out2_mean, out2_var = out2.mean(0), out2.var(0)
+
+    assert_equal(out1_mean, out2_mean, prec=0.003)
+    assert_equal(out1_var, out2_var, prec=0.003)
+    return
+
+
+@pytest.mark.parametrize("non_linearity", [F.relu])
+@pytest.mark.parametrize("include_hidden_bias", [True, False])
+def test_hidden_layer_log_prob(non_linearity, include_hidden_bias, B=2, D=3, H=2, N=1000000):
```
fritzo: Is `N` vestigial?

pyro/contrib/bnn/hidden_layer.py
```diff
@@ -0,0 +1,122 @@
...
+    r"""
+    This distribution is a basic building block in a Bayesian neural network. It represents
```
fritzo: nit: comments wider than 80chars print poorly in jupyter notebooks. feel free to disregard this nit.

pro tip: in vim position cursor at top of paragraph and type `gq<ENTER>` to automatically wrap.

pyro/contrib/bnn/hidden_layer.py
martinjankowiak: maybe, i am always wary of multiple inheritance because i'm not always sure how it works

pyro/contrib/bnn/utils.py
martinjankowiak: done

pyro/contrib/bnn/hidden_layer.py
martinjankowiak: although probably not. since X changes from mini-batch to mini-batch in a typical use case you remake these on the fly as you go. so we don't really need any of the stateful enhancements that `nn.Module` bbrings

karalets:
Let me have a look at this for a couple of days guys.
And As JP said we an implementation for this, but it never made it into the
repo, so I would look at this with fresh eyes. I also used a hidden layer
dist to be able to do local rep, so strategy looks sound here.

pyro/contrib/bnn/hidden_layer.py
```diff
@@ -0,0 +1,110 @@
...
+        KL_A -= self.dim_X * self.dim_H
+        KL_A += torch.pow(self.A_scale / self.A_prior_scale, 2.0).sum()
+        KL_A -= 2.0 * torch.log(self.A_scale / self.A_prior_scale).sum()
+        return 0.5 * KL_A
+
+    def rsample(self, sample_shape=torch.Size()):
+        # note: weight space sampling is only meant for testing
+        if self.weight_space_sampling:
+            A = self.A_mean + torch.randn(sample_shape + self.A_scale.shape).type_as(self.A_mean) * self.A_scale
+            activation = torch.matmul(self.X, A)
```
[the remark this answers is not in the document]
martinjankowiak: of course it works with subsampling. thus the `KL_factor` argument

martinjankowiak:
@karalets bump

karalets:
Ok you thanksgiving ghoul. I am already looking at it.

fritzo:
OK LGTM. I'll merge this now, but we should examine one issue in a follow-up PR:

Instead of hacking KL into the `.log_prob()` method and manually scaling by `KL_factor`, it would be interesting to try to make this work with `TraceMeanField_ELBO` by
1. raising `NotImplementedError` from `.log_prob()`, then
2. registering a KL divergence between some pair of distributions, using the KL you've implemented here.

PR #1528, opened 2018-11-12
Implement diagnostic stats

fehiepsi:
This is a split of #1525 to make it easier for reviewing and comments.

This pull request implements autocorrelation, autocovariance, gelman_rubin, effective sample size.

I have implemented these functions with `batch` support in mind. So I think that it is more efficient than implements in Stan and PyMC3. In addition, they work with arbitrary sample shape (not as singleton sample as in other implementation). There is a minor difference at the computation of initial positive sequence: the other implementations truncate the sequence to make it positive, then build a monotone sequence from the truncated sequence. In my implementation, I clamp the sequence by 0, then build a monotone sequence from the clamped sequence. They are just the same in my opinion (correct me if I'm wrong).


pyro/ops/stats.py
```diff
@@ -0,0 +1,160 @@
...
+        while m % 5 == 0:
+            m //= 5
+        if m == 1:
+            return N
+        N += 1
+
+
+def autocorrelation(input):
+    """
+    Computes the autocorrelation of samples. The last dimension is the number of
```
neerajprad: I think it will be more natural to have the first dimension represent the different samples. This is how the `Empirical` distribution would represent it at least, and it will be nice to use it directly with these diagnostic functions.

pyro/ops/stats.py
```diff
@@ -0,0 +1,160 @@
...
+    chain_var = input.var(dim=1)
+    var_within = chain_var.mean(dim=0)
+    var_estimator = (N - 1) / N * var_within + var_between
+    return var_within, var_estimator
+
+
+def gelman_rubin(input):
+    """
+    Computes R-hat over chains of samples. The first two dimensions of `input`
+    is `C x N` where `C` is the number of chains and `N` is the number of samples.
```
neerajprad: Not part of this PR, but we will need to refactor our `MCMC` (and, even `TracePosterior`) classes to aggregate samples by chain. I will start thinking about that.

pyro/ops/stats.py
fehiepsi: I'll add a parameter `dim` instead. By default, it will be `-1`.

pyro/ops/stats.py
fehiepsi: In that case, it will be good to have two arguments `chain_dim`, `sample_dim`. :D

neerajprad: Sounds good! If it's too much of an effort to generalize, I think the choice of `chain_dim` and `sample_dim` is fine. This is anyways what would have been the default convention to have chain be leftmost followed by samples.

pyro/ops/stats.py
neerajprad: Sounds good! If its too much effort generalizing though, I wouldn't bother with it since there aren't too many parts of the code that would be calling these diagnostic functions (at least for the time being); so feel free to just change the default if that's easier.

fehiepsi: @neerajprad to make things consistence, I will follow your suggestion by setting `dim=0` by default. This is easier than the case with both chain and sample size. I just simply transpose dim with -1 for Fourier transform, then transpose the result back.

pyro/ops/stats.py
fehiepsi: In this case, I have to add two dummy dimensions at the front of input, then transpose sample and chain dimensions to these two dummies. Finally, I have to squeeze them. This way it is easier to follow the code without worrying about the positions of dimensions.

fehiepsi:
@neerajprad It is not easy to test efficient_sample_size throughout though. I tried various chains and found that the result is a bit different to arviz. This might be expected because the way I calculate autocorrelation is different to them (I follow Stan way).

neerajprad:
> @neerajprad It is not easy to test efficient_sample_size throughout though. I tried various chains and found that the result is a bit different to arviz. This might be expected because the way I calculate autocorrelation is different to them (I follow Stan way).

Np, @fehiepsi. This looks great overall, and I'll go through the computation logic in detail tonight.

pyro/ops/stats.py
```diff
@@ -0,0 +1,208 @@
...
+    :param torch.Tensor input: the input tensor.
+    :param int dim: the dimension to calculate autocorrelation.
+    :returns torch.Tensor: autocorrelation of `input`.
+    """
+    return autocorrelation(input) * input.var(dim, unbiased=False, keepdim=True)
+
+
+def _cummin(input):
+    """
+    Computes cummulative minimum of input at dimension `dim=0`.
```
neerajprad: Can we mention that `input` is expected to be a vector instead?

pyro/ops/stats.py
```diff
@@ -0,0 +1,208 @@
...
+    input = input.transpose(0, sample_dim + 2).transpose(1, chain_dim + 2)
+
+    N_half = input.size(0) // 2
+    new_input = torch.stack([input[:N_half], input[-N_half:]], dim=1)
+    new_input = new_input.reshape((N_half, -1) + input.shape[2:])
+    split_rhat = gelman_rubin(new_input)
+    return split_rhat.squeeze(max(sample_dim, chain_dim)).squeeze(min(sample_dim, chain_dim))
+
+
+def _fft_next_good_size(N):
```
neerajprad: Are you able to run these tests using the night PyTorch build. I get `RuntimeError: fft: ATen not compiled with MKL support`, so this will likely also fail when we merge into `pytorch-1.0` branch.

pyro/ops/stats.py
```diff
@@ -0,0 +1,208 @@
...
+    Rho_k = rho_k if N % 2 == 0 else rho_k[:-1]
+    Rho_k = Rho_k.reshape((N // 2, 2) + Rho_k.shape[1:]).sum(dim=1)
+
+    # separate the first index
+    Rho_init = Rho_k[0]
+
+    if Rho_k.size(0) > 1:
+        # Theoretically, Rho_k is positive, but due to noise of correlation computation,
+        # Rho_k might not be positive at some point. So we need to truncate (ignore first index).
+        Rho_positive = Rho_k[1:].clamp(min=0)
```
neerajprad: Maybe I'm misunderstanding, but I would imagine that this will turn the entire sequence after the occurrence of a negative element to 0, which won't be the case if you were to remove the elements.

tests/ops/test_stats.py
```diff
@@ -0,0 +1,85 @@
...
+
+def test_split_gelman_rubin():
+    x = torch.rand(4, 10, 5)
+    r_hat1 = gelman_rubin(x.reshape(4, 2, 5, 5).reshape(8, 5, 5), sample_dim=1, chain_dim=0)
+    r_hat2 = split_gelman_rubin(x, sample_dim=1, chain_dim=0)
+    assert_equal(r_hat1, r_hat2)
+
+
+def test_autocorrelation():
+    x = torch.arange(10.)
```
neerajprad: I understand that you have already validated the output wrt Stan / arviz?

pyro/ops/stats.py
```diff
@@ -0,0 +1,208 @@
...
+    chain_dim = input.dim() + chain_dim if chain_dim < 0 else chain_dim
+    assert chain_dim != sample_dim
+    input = input.reshape((1, 1) + input.shape)
+    input = input.transpose(0, sample_dim + 2).transpose(1, chain_dim + 2)
+
+    N_half = input.size(0) // 2
+    new_input = torch.stack([input[:N_half], input[-N_half:]], dim=1)
+    new_input = new_input.reshape((N_half, -1) + input.shape[2:])
+    split_rhat = gelman_rubin(new_input)
+    return split_rhat.squeeze(max(sample_dim, chain_dim)).squeeze(min(sample_dim, chain_dim))
```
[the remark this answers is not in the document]
fehiepsi: There is an effort for doing that but it is closed https://github.com/pytorch/pytorch/pull/4692#issuecomment-394970171

pyro/ops/stats.py
fehiepsi: According to [rfft docs](https://pytorch.org/docs/stable/torch.html#torch.rfft), I think you need to install pytorch with MKL.

tests/ops/test_stats.py
fehiepsi: Yes, all the tests are against arviz. For efficient_sample_size, the results are a little bit different with different random seed (to generate `input`): the difference are 54 (ours) vs 57 (theirs), 95 vs 97, 992 vs 996 (kind of that).

tests/ops/test_stats.py
```diff
@@ -0,0 +1,88 @@
...
+    for i in range(1, x.size(0)):
+        y[i] = min(x[i], y[i-1])
+
+    assert_equal(_cummin(x), y)
+
+
+@pytest.mark.init(rng_seed=3)
+def test_effective_sample_size():
+    x = torch.empty(2000)
+    x[0].normal_(0, 1)
```
neerajprad: nit: why not just `x = torch.randn(2000)`?

tests/ops/test_stats.py
```diff
@@ -0,0 +1,88 @@
...
+    y = torch.empty(x.shape)
+    y[0] = x[0]
+    for i in range(1, x.size(0)):
+        y[i] = min(x[i], y[i-1])
+
+    assert_equal(_cummin(x), y)
+
+
+@pytest.mark.init(rng_seed=3)
+def test_effective_sample_size():
```
neerajprad: Can we make this a parameterized test and add tests for other sample shapes, e.g. when we have more than a single latent site? Can we similarly check for different shapes in other tests?

tests/ops/test_stats.py
```diff
@@ -0,0 +1,88 @@
...
+    x = torch.empty(2000)
+    x[0].normal_(0, 1)
+    for i in range(1, x.size(0)):
+        x[i].normal_(0.8 * x[i-1], 1)
+    x = x.reshape(4, x.size(0) // 4)
+
+    try:
+        assert_equal(effective_sample_size(x, sample_dim=1, chain_dim=0).item(), 268, prec=2)
```
neerajprad: I am not sure why, but this is throwing an exception on my Mac. This seems like a relatively big discrepancy. Would you expect that? If so, we should expand the precision appropriately.

```
>               assert x == approx(y, abs=prec), msg
E               AssertionError: 284.49502516376975 vs 268
```

PS: I realized that the pip wheels do not ship with MKL, so I installed from conda which has MKL included.

pyro/ops/stats.py
```diff
@@ -0,0 +1,215 @@
...
+    input_tril = input.unsqueeze(0).repeat((N,) + (1,) * input.dim())
+    triu_mask = input.new_ones(N, N).triu(diagonal=1).reshape((N, N) + (1,) * (input.dim() - 1))
+    triu_mask = triu_mask.expand((N, N) + input.shape[1:]) > 0.5
+    input_tril.masked_fill_(triu_mask, input.max())
+    return input_tril.min(dim=1)[0]
+
+
+def effective_sample_size(input, sample_dim=0, chain_dim=1):
+    """
+    Computes effective sample size of input.
```
neerajprad: I think we should also document what kind of input sizes / constraints we expect. e.g. if the latents have different shapes, this would not work. Are these diagnostics well behaved with multidimensional (but equally sized) latents, or is there an expectation that they will all be unidimensional?

pyro/ops/stats.py
neerajprad: Makes sense, thanks for explaining @fehiepsi.

pyro/ops/stats.py
neerajprad: The pip wheels do not ship with MKL unfortunately, but the conda package does, which also implies that we need to run most of these tests locally since CI will end up skipping.

tests/ops/test_stats.py
fehiepsi: This way we can create a correlated sequence

pyro/ops/stats.py
fehiepsi: I try to make these methods have the same interface as `torch.foo(input, ...)`. This just accepts 1 input tensor. If you want to apply it for multiple latents, you have to put it in a loop. The other way is to flatten then merge all latents into a 1D tensor, then apply this function.

tests/ops/test_stats.py
fehiepsi: This is expected I guess. I will make a notebook to show the different between Stan and arviz soon.

pyro/ops/stats.py
```diff
@@ -0,0 +1,208 @@
...
+
+    chain_var = input.var(dim=0)
+    var_within = chain_var.mean(dim=0)
+    var_estimator = (N - 1) / N * var_within + var_between
+    return var_within, var_estimator
+
+
+def gelman_rubin(input, sample_dim=0, chain_dim=1):
+    """
+    Computes R-hat over chains of samples. It is required that
```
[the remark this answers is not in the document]
fehiepsi: Yes, maybe we resample if the chains have different size (I can't figure out when it happens:-?)

tests/ops/test_stats.py
fehiepsi: Could you rerun the test @neerajprad ? I detected the bug (missing dim when centerize signal) and found that the new result is quite agree with arviz.

neerajprad: Awesome, this passes now.

pyro/ops/stats.py
neerajprad: I see; so input can be arbitrary shaped but the expectation is that it represents samples only from a single latent site.

neerajprad:
@fehiepsi - I am happy to merge this; let me know if it looks good to  you.

fehiepsi:
@neerajprad This looks good to me, as long as you satisfied with all the tests (I see that testing is so important now; without your suggestions, I have missed two bugs in the last commit).

neerajprad:
Thanks for adding the additional tests for different shaped inputs, it also helped me better understand the logic!

PR #1530, opened 2018-11-13
Small test fixes for CI

neerajprad:
- I was finding instances where our travis log output was getting truncated due to large size; changing the NUTS/HMC tests to use `.debug` instead of `.info`. If I continue to see issues, I'll follow up with similar fixes for some of our other tests.
 - tests in `nn` were getting run at every stage, so just adding a stage marker to ensure that they only run in the unit test stage.

### DIFF
--- a/tests/test_memory.py
+++ b/tests/test_memory.py
@@ -63,7 +63,7 @@ def test_networkx_copy():
     g = nx.DiGraph()
     expected = count_objects_of_type(nx.DiGraph)
     for _ in range(10):
-        h = g.fresh_copy()
+        h = g.__class__()
         h.__dict__.clear()
         del h
         counts.append(count_objects_of_type(nx.DiGraph))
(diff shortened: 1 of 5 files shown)


tests/test_memory.py
```diff
@@ -63,7 +63,7 @@ def test_networkx_copy():
     g = nx.DiGraph()
     expected = count_objects_of_type(nx.DiGraph)
     for _ in range(10):
-        h = g.fresh_copy()
+        h = g.__class__()
```
neerajprad: `fresh_copy` is deprecated.

tests/test_util.py
```diff
@@ -6,7 +6,7 @@
 import torch
 from pyro import util
 
-pytestmark = pytest.mark.stage('test_util')
+pytestmark = pytest.mark.stage('unit')
```
neerajprad: These tests were not running at all, since we don't have a `test_util` stage in `.travis.yml`.

Issue #1453, opened 2018-10-16
[FR] Error messages should be more specific


fritzo:
Users at PROBPROG 2018 requested more-specific error messages when models are invalid.

Specific improvements include:
- [ ] Print sample site information when `Distribution` inputs are invalid. Currently it requires a debugger to see what line of the model is affected.
- [ ] On error, print contextual information: the current `cond_indep_stack`, the portion of stack trace under `svi.step()`, whether the model or guide is being executed, `_DIM_ALLOCATOR` sate, tracing state, etc. This could be accomplished via a `try: except: raise` block to add information to a caught error.
- others?

kamathhrishi:
Can I work on this issue ?

kamathhrishi:
@fritzo I would like to go for this first. What exceptions do we handle under exception blocks? I noticed PEP8 format didn't allow me to create a blank except statement that could allow me to handle any kind of error.

fritzo:
@kamathhrishi  I think the cleanest way might be to add a new context manager to pyro.poutine.runtime
```py
# in pyro/poutine/runtime.py
class LogRuntimeOnError(object):
    def __enter__(self):
        return self
    def __exit__(self, typ, value, traceback):
        if typ is not None:
            # TODO log runtime information here, e.g.
            logging.error('PYRO_STACK:')
            for handler in PYRO_STACK:
                logging.error(str(handler))
            # ...log other things...
```
Then we can use this debugger in `pyro.infer.enum.get_importance_trace`
```diff
  # in pyro/infer/enum.py
+ @pyro.poutine.runtime.LogRuntimeOnError
  def get_importance_trace(graph_type, ...):
      ...
```
and probably somewhere else for HMC. @eb8680 does this seem reasonable?

kamathhrishi:
@fritzo Thank you for the leads. I am unfamiliar with the codebase, is there any way I could learn about it quickly any relevant paper or document?

fritzo:
@kamathhrishi I recommend starting with the [tutorials](http://docs.pyro.ai) and reading [minipyro.py](http://pyro.ai/examples/minipyro.html) for an architectural overview. Feel free to put up a draft pull request following my above comment, and we can discuss changes on the PR.

PR #1534, opened 2018-11-13
Add pretty printing of trace shapes on error

fritzo:
Addresses #1453 

Adds tensor_shapes.ipynb-style pretty printing of trace shapes on error. For example in the hmm error fixed by the recent #1521 , this would have printed
```
$ python examples/hmm.py
      977 Loading data
     1054 ----------------------------------------
     1054 Training model_1 on 229 sequences
     1067 Step	Loss
...
ValueError: Error while packing tensors at site 'y_0':
  Invalid tensor shape.
  Allowed dims: -3, -2, -1
  Actual shape: (16, 1, 8, 88)
  Try adding shape assertions for your model's sample values and distribution parameters.
Trace Shapes:
 Param Sites:
Sample Sites:
 probs_x dist            | 16 16
        value            | 16 16
     log_prob            |
 probs_y dist            | 16 88
        value            | 16 88
     log_prob            |
     x_0 dist       8  1 |
        value    16 1  1 |
     log_prob    16 8  1 |
     y_0 dist 16  1 8 88 |
        value       8 88 |
     log_prob 16  1 8 88 |
```

### DIFF
--- a/pyro/poutine/trace_struct.py
+++ b/pyro/poutine/trace_struct.py
@@ -136,9 +136,10 @@ class Trace(networkx.DiGraph):
                         log_p = site["fn"].log_prob(site["value"], *site["args"], **site["kwargs"])
                     except ValueError:
                         _, exc_value, traceback = sys.exc_info()
+                        shapes = self.print_shapes(last_site=site["name"])
                         six.reraise(ValueError,
-                                    ValueError("Error while computing log_prob_sum at site '{}': {}"
-                                               .format(name, exc_value)),
+                                    ValueError("Error while computing log_prob_sum at site '{}':\n{}\n"
+                                               .format(name, exc_value, shapes)),
                                     traceback)
                     log_p = scale_and_mask(log_p, site["scale"], site["mask"]).sum()
                     site["log_prob_sum"] = log_p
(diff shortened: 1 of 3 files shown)


pyro/poutine/trace_struct.py
```diff
@@ -306,10 +320,91 @@ def pack_tensors(self, plate_to_symbol=None):
...
                 six.reraise(ValueError,
-                            ValueError("Error while packing tensors at site '{}':\n  {}"
-                                       .format(site["name"], exc_value)),
+                            ValueError("Error while packing tensors at site '{}':\n  {}\n{}"
+                                       .format(site["name"], exc_value, shapes)),
                             traceback)
 
-        self.plate_to_symbol = plate_to_symbol
-        self.symbol_to_dim = symbol_to_dim
+    def print_shapes(self, title='Trace Shapes:', last_site=None):
```
neerajprad: nit: Would suggest calling this method `repr_shapes` or something like that, which suggests that it would return a string representation.

neerajprad:
We had added error message checks in `test_trace_log_prob_err_msg ` which you'll need to change as well.

PR #1542, opened 2018-11-15
Fix CUDA tests on master

neerajprad:


pyro/ops/contract.py
```diff
@@ -17,7 +17,7 @@
 def _finfo(tensor):
     # This can be replaced with torch.finfo once it is available
     # https://github.com/pytorch/pytorch/issues/10742
-    return np.finfo(torch.empty(torch.Size(), dtype=tensor.dtype).numpy().dtype)
+    return np.finfo(torch.empty(torch.Size(), dtype=tensor.dtype).cpu().numpy().dtype)
```
fritzo: let's avoid GPU here:
```py
np.finfo(torch.empty(torch.Size(), dtype=tensor.dtype, device='cpu').numpy().dtype)
```

tests/infer/test_valid_models.py
```diff
@@ -1300,6 +1301,7 @@ def guide():
 
 
 @pytest.mark.parametrize('markov', [False, True])
+@pytest.mark.skipif("CUDA_TEST" in os.environ, reason="https://github.com/uber/pyro/issues/1380")
```
fritzo: This issue is already resolved. If this is still failing, try changing `range(20)` to `range(15)`.

neerajprad: I had to use`range(11)` - anything above that fails.

fritzo: Oh I see, that makes sense since each time step allocates a new dim, so `range(11)` should allocate 22 dims.

neerajprad:
This should be good to merge.

PR #1543, opened 2018-11-17
added limited support for analytic kl divergences for elbos

### LINKED ISSUE #688: Use kl_divergence() in SVI

martinjankowiak:
Resolves #688
pair coded with @fritzo.

## summary

this adds support for an `SVI` object that supports analytic kl divergences. this `ELBO`
assumes that:

- everything is reparameterized (have checks for this)
- a certain dependency condition is satisfied. we have an approximate check for this, and describe the exact requirement in the docstring. since mean field guides is a common case in which this condition is satisfied, we use mean field as the name of the class

## reviewers note:
- ~~do we need any more tests?~~
- ~~i removed some `xfail_param` statements in test_gradient. if it is important to reinstate these please let me know how to do so (i couldn't figure it out).~~

## tested

- [x] gradient tests
- [x] autoguide tests with auxiliary guide variables
- [x] death tests for warning about invalid model structure


martinjankowiak:
@fritzo do you want to add the masked support directly into this PR?

fritzo:
@martinjankowiak one of the autoguide tests is failing with `TraceMeanField_ELBO`, can you help diagnose?

pyro/infer/trace_mean_field_elbo.py
```diff
@@ -0,0 +1,81 @@
...
+    Note for advanced users:
+
+    The mean field condition is a sufficient but not necessary condition for
+    this estimator to be correct. The precise condition is that for every
+    latent variable `z` in the guide, its parents in the model must not include
+    any latent variables that are descendants of `z` in the guide. Here
+    'parents in the model' and 'descendants in the guide' is with respect
+    to the corresponding (statistical) dependency structure. For example, this
+    condition is always satisfied if the model and guide have identical
+    dependency structures.
```
fritzo: Would it be sufficient and not too inconvenient to require the model and guide have the same variable ordering, and warn otherwise? This is easy to check from the traces.

fehiepsi: @fritzo I guess the trace just gives the "coding" order, not "structure" order.

martinjankowiak: @fritzo i think that would be reasonable. probably warn over error/assert

martinjankowiak:
@fritzo not sure what's going on with the autoguide test (we should take a look together), but looking at this makes me realize that this implementation will break for subsampled latent variables (no problems with subsampled observations)

pyro/infer/trace_mean_field_elbo.py
```diff
@@ -0,0 +1,121 @@
...
+def _check_mean_field_requirement(model_trace, guide_trace):
+    """
+    Checks that the guide and model sample sites are ordered identically.
+    This is sufficient but not necessary for correctness.
+    """
+    model_sites = [name for name, site in model_trace.nodes.items()
+                   if site["type"] == "sample" and name in guide_trace.nodes]
+    guide_sites = [name for name, site in guide_trace.nodes.items()
+                   if site["type"] == "sample" and name in model_trace.nodes]
+    assert set(model_sites) == set(guide_sites), "programmer error"
```
martinjankowiak: 'programmer error' seems a potentially confusing assert msg, no?

fritzo: if find it standard, and it should never be surfaced to users, but ok I'll remove it.

martinjankowiak:
might need to adjust the precision on failing test down a bit?

fehiepsi:
Love the latest change, we only check things when validation is enable.

fritzo:
@gpleiss hopefully this will speed up your GPyTorch+Pyro models. Let us know if you have any issues!
```diff
- SVI(model, guide, optim, Trace_ELBO())
+ SVI(model, guide, optim, TraceMeanField_ELBO())
```

gpleiss:
Awesome! Excited to try it out :)

royf:
Very cool, thanks!

Are there plans to support non-reparametrizable variables?
I have a use case where variables cannot be reparametrized, and where training without analytic KL cannot work.
So I cannot use Pyro until it supports analytic KL of non-reparametrizable variables.

PR #1544, opened 2018-11-17
Revise gp module

fehiepsi:
This pull request revises gp module. I will sum up all the changes here:

## Changes in API:

- Get value of a parameter of Reparameterized class:
```py
# previously
gpmodel.get_param("noise")

# now
gpmodel.noise
```

- GP model is indeed a `nn.Module` now. So no need to use SVI. If user wants to use SVI, simply use `pyro.module(...)` to cast its parameters to Pyro ParamStoreDict
```py
# previously
loss = Trace_ELBO()
optimizer = pyro.optim.Adam({"lr": 0.01})
svi = SVI(gpmodel.model, gpmodel.guide, optimizer, loss)
for i in range(1000):
    svi.step()

# now
optimizer = torch.optim.Adam(gpmodel.parameters(), lr=0.01)
loss_fn = Trace_ELBO.differentiable_loss
for i in range(1000):
    optimizer.zero_grad()
    loss = loss_fn(gpmodel.model, gpmodel.guide)
    loss.backward()
    optimizer.step()
```

- `gpmodel.optimize()` method is removed. The training step is quite similar to how to train a PyTorch module, so it is good to keep the interface simple. If needed, users can use the helper `contrib.gp.util.train` to quickly train a gpmodel though.

- No requirement to specify names for kernel/model. The problem of name confliction (for combined kernels) is resolved thanks to `contrib.autoname.name_count` context manager.

## Notes for reviewers

- Although there is a lot of file changes, most of them are just a replace from `self.get_param("foo")` to `self.foo`.

- Reparameterized class has been completely revised. It will no longer interact with Pyro ParamStoreDict. We can still use SVI the following way though it is unnecessary now:
```py
def model():
    pyro.module(gpmodel)
    gpmodel.model()

def guide():
    pyro.module(gpmodel)
    gpmodel.guide()
```

- Experimental support for autoguide. This is required by a user in Pyro Forum a few months back. I also intend to use it in GPLVM tutorial.

## Todo

- [x] Use `contrib.autoname`. So no need to define name for each module.
- [x] No need to use `self.get_param("foo")` to get value for parameter `foo` in Reparameterized class, just simply uses `self.foo`. The previous pattern `foo = self.get_param("foo")` is a little bit weird.
- [x] Implement `autoguide` method for parameters in Reparameterized class, which mimics the pattern of `autoguide` module.
- [x] ~Use KL in #1543 by default in `optimize` function. This will help the model converges much faster.~ It seems that `whitening` does a good job for optimization. KL does not help much.
- [x] Fix bugs, tests
- [x] Fix docs
- [x] Fix sv-dkl and gp, bo tutorials


pyro/infer/__init__.py
```diff
@@ -32,4 +32,5 @@
     "TracePosterior",
     "TracePredictive",
     "Trace_ELBO",
+    "TraceMeanField_ELBO",
```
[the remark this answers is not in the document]
fehiepsi: Thanks! It seems like there is a few redundant while merging other PRs. I'll remove them.

pyro/contrib/gp/util.py
```diff
@@ -1,139 +1,237 @@
...
-        """
-        Sets a constraint to a parameter.
+        self._priors[name] = prior
+        self.autoguide(name, dist.Delta)
 
-        :param str param: Name of the parameter.
-        :param ~torch.distributions.constraints.Constraint constraint: A PyTorch
-            constraint. See :mod:`torch.distributions.constraints` for a list of
-            constraints.
+    def autoguide(self, name, dist_constructor):
```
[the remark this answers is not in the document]
fehiepsi: It would be nice if we can use `contrib.autoguide` directly. However, there are a few conflicts here:
- `contrib.autoguide` will flatten the parameter then build a guide on it. Here we keep the shape of parameter as-is.
- `contrib.autoguide` works with ParamStoreDict so cann't  be trained with torch.optim.

pyro/contrib/gp/models/model.py
```diff
@@ -55,14 +53,16 @@ class GPModel(Parameterized):
...
 
-    + Using a variational inference (e.g. :class:`~pyro.infer.svi.SVI`) on the pair
-      :meth:`model`, :meth:`guide` as in `SVI tutorial
-      <http://pyro.ai/examples/svi_part_i.html>`_:
+    + Using a variational inference on the pair :meth:`model`, :meth:`guide`:
 
-        >>> optimizer = pyro.optim.Adam({"lr": 0.01})
-        >>> svi = SVI(gpr.model, gpr.guide, optimizer, loss=Trace_ELBO())
+        >>> optimizer = torch.optim.Adam(gpr.parameters(), lr=0.01)
+        >>> loss_fn = Trace_ELBO().differentiable_loss
```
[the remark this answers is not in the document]
fehiepsi: @eb8680 I found it does not help much while testing the sv-dkl example. In addition, I am not sure that TraceMeanField_ELBO will work with `infer={"is_auxiliary": True}` in autoguide so I let it `Trace_ELBO` by default.

pyro/contrib/gp/kernels/kernel.py
```diff
@@ -73,89 +66,6 @@ def _slice_input(self, X):
         else:
             raise ValueError("Input X must be either 1 or 2 dimensional.")
 
-    def add(self, other, name=None):
```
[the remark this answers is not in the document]
fehiepsi: I remember that you suggested that. :D While revising, I found that it is better to be more explicit that we are creating a new `nn.Module`, and that `nn.Module` has two children (hence `.parameters()` will catch all parameters).

pyro/contrib/gp/kernels/kernel.py
```diff
@@ -73,89 +66,6 @@ def _slice_input(self, X):
...
-        Returns the subkernel corresponding to ``name``.
-
-        :param str name: Name of the subkernel.
-        :returns: A subkernel.
-        :rtype: Kernel
-        """
-        if name in self._subkernels:
-            return self._subkernels[name]
-        else:
-            raise KeyError("There is no subkernel with the name '{}'.".format(name))
```
fehiepsi: This method is useful with the old API (which interacts things by `name`). Now, `name` is removed so this is not useful anymore. We can access subkernels's parameters like accessing parameters of children of a `nn.Module`.

fehiepsi:
@martinjankowiak Would you mind helping me revise the tutorials. Then content is almost the same. The difference at second gpr loss is due to: previously, we do MAP inference on constrained space; but it should be in unconstrained space like in `contrib.autoguide`. I make that change, and the inference gives a tiny different result.

martinjankowiak:
@fehiepsi sure. can you be more specific about the changes you'd like me to pay attention to? both in the code and in the tutorial?

fehiepsi:
@martinjankowiak Thanks! For the changes in tutorials, I sum up the details in the first comment. Basically, I replace the VI procedure (not use SVI interface, but use `torch.optim` directly) and make a change at how to get value of a parameter. The inference results are the same (except for the MAP inference, which I explained above).

examples/contrib/oed/gp_bayes_opt.py
```diff
@@ -33,7 +34,9 @@ def update_posterior(self, X, y):
         X = torch.cat([self.gpmodel.X, X])
         y = torch.cat([self.gpmodel.y, y])
         self.gpmodel.set_data(X, y)
-        self.gpmodel.optimize(loss=TraceEnum_ELBO(strict_enumeration_warning=False).differentiable_loss)
+        optimizer = torch.optim.Adam(self.gpmodel.parameters())
+        gp.util.train(self.gpmodel, optimizer,
+                      loss_fn=TraceEnum_ELBO(strict_enumeration_warning=False).differentiable_loss)
```
fehiepsi: @ae-foster I have just revised the GP module to make it easier to use. However, I notice that the `oed` example failed because of my change. I look over the example and it seems to me that this together with a similar one in `ab_test.py` are the only places to change. But the test is still failing :(. Now, when a gpmodel is trained, it keeps trained values for parameters (previously, calling `pyro.clear_param_store()` will clear everything). I suspect that it causes some issues for this example. What do you think about it?

pyro/contrib/gp/util.py
```diff
@@ -1,139 +1,237 @@
...
         else:
-            return self._registered_params[param]
-
-    def _register_param(self, param, mode="model"):
+            unconstrained_value = pyro.sample("{}_latent".format(name), guide.independent(),
+                                              infer={"is_auxiliary": True})
+            transform = biject_to(self._priors[name].support)
+            value = transform(unconstrained_value)
+            log_density = transform.inv.log_abs_det_jacobian(value, unconstrained_value)
+            p = pyro.sample(name, dist.Delta(value, log_density.sum(), event_dim=value.dim()))
```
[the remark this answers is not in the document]
fehiepsi: Can I defer this to a later PR? This Reparameterized class is non-GP though. If you think that it is useful, we might move it to `pyro.nn.Reparameterized` in the future. :)

examples/contrib/oed/gp_bayes_opt.py
fehiepsi: By the way, why did you use `loss=TraceEnum_ELBO(strict_enumeration_warning=False).differentiable_loss` instead of `loss=TraceEnum_ELBO(strict_enumeration_warning=False)`? This is another hint to me.

pyro/contrib/gp/models/model.py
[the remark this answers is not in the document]
martinjankowiak: we've encountered a few bugs along the way, but as far as we know tracemeanfield_elbo works with auto guides

tutorial/source/bo.ipynb
```diff
@@ -164,7 +166,9 @@
     "    X = torch.cat([gpmodel.X, x_new]) # incorporate new evaluation \n",
     "    y = torch.cat([gpmodel.y, y])     \n",
     "    gpmodel.set_data(X, y)\n",
-    "    gpmodel.optimize()  # optimize the GP hyperparameters using default settings"
+    "    # optimize the GP hyperparameters using Adam optimizer\n",
+    "    optimizer = optim.Adam(gpmodel.parameters())\n",
```
martinjankowiak: i'm surprised you use adam with default parameters; i've usually found this is too much momentum in this setting

fehiepsi: @martinjankowiak this is just to replicate the old behaviour. I agree that we should use another cheaper one (such as LBFGS) or with a higher `lr`. If you want a change, I'll try it after debugging `oed`.

pyro/contrib/gp/models/model.py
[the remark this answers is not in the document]
fehiepsi: Thanks @martinjankowiak ! It is good to know.
@eb8680 Do you want to make `TraceMeanField_ELBO` by default? Either way is fine to me.

martinjankowiak:
@fehiepsi i'm happy to take a closer look at and clean-up the discussion in gp.ipynb. but let me do that in a follow-up PR after this is merged

examples/contrib/oed/ab_test.py
```diff
@@ -102,7 +102,6 @@ def main(num_vi_steps, num_bo_steps):
         gpmodel = gp.models.GPRegression(
             X, y, gp.kernels.Matern52(input_dim=1, lengthscale=torch.tensor(10.)),
             noise=torch.tensor(noise), jitter=1e-6)
-        gpmodel.optimize(loss=TraceEnum_ELBO(strict_enumeration_warning=False).differentiable_loss)
```
fehiepsi: @ae-foster I have resolved the issue. In addition, I remove this optimize step because with `pyro.clear_param_store` a few lines below, gpmodel is reset.

fehiepsi:
Finally, it seems that all issues have been resolved. To all reviewers, this PR is ready to review now. :)

fehiepsi:
I have used `pytorch 1.0` for a while. Now, the tests passed but I catch an issue: Before merging 1.0 branch, test_inference for GPR and SGPR are quite fast. After merging, they are pretty slow (took a minute to finish). Not sure why because I used the same pytorch 1.0 environment before and after merging.

pyro/contrib/gp/models/model.py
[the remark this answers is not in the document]
eb8680: Yes, analytic KLs should be the default for GPs

neerajprad:
> @eb8680 I have used pytorch 1.0 for a while. Now, the tests passed but I catch an issue: Before merging 1.0 branch, test_inference for GPR and SGPR are quite fast. After merging, they are pretty slow (took a minute to finish). Not sure why because I used the same pytorch 1.0 environment before and after merging.

Hmm..are you able to reproduce this locally? Some of the HMC tests also seemed slower, but I have not been able to reliably reproduce it locally. Could you point me to how I can reproduce it?

fehiepsi:
@neerajprad FYI `pytest -vs tests/contrib/gp/test_models.py::test_inference[GPRegression_y1D]`. Now it took more than 2 minutes on my machine. :(

neerajprad:
@fehiepsi - I have figured out the reason (or at least part of the reason) for the slowness. Both the `log_prob` and gradient computation for the multivariate normal distribution seem extremely slow on the expanded instances (i.e. instances constructed from `distribution.expand` method). Before the merge we were using `ReshapedDistribution`. This doesn't seem to impact HMC and most other tests, so it might be that some operation in `mvn.log_prob` are extremely slow for expanded instances. The slowdown is quite unacceptable though - we might have to put in our patched mvn implementation in Pyro or reintroduce `ReshapedDistribution` if there is evidence that other distributions are similarly slow.

fehiepsi:
You are right! The computation for mvn’s logprob in pytorch is ineffective with expanded operator (because unbroadcasted_scale_tril is expanded during the way). I’m not sure if this also happens for other distributions. But I’ll try to make a pull request in pytorch for that.

Btw the performance is not relevant to this pull request, so this is still ready for reviewing.

fehiepsi:
I just made a change to separate jitter and noise:
+ previously, `noise` is constrained greater than jitter, and noise will be added to the diagonal of kernel(X)
+ now, `noise` is constrained positive, and (noise + jitter) will be added to the diagonal of kernel(X)
This way, users don't have to worry about creating a prior for noise with some positive lower bound (to avoid singularity).

fehiepsi:
@eb8680 Sorry for making the changes to so many files! It is a consequence of revising Reparameterized class, which leads to deleting redundant things in all files of GPs. If I split into 2 PRs "drop name requirement" and "drop the rely on ParamStoreDict", then each of them also contains many changed files. Beside the clean up, I will comment below all the change.

pyro/contrib/gp/models/gplvm.py
```diff
@@ -62,80 +56,30 @@ class GPLVM(Parameterized):
...
-        X_scale_tril_shape = self.X_loc.shape + (C,)
-        Id = eye_like(self.X_loc, C)
-        X_scale_tril = Id.expand(X_scale_tril_shape)
-        self.X_scale_tril = Parameter(X_scale_tril)
-        self.set_constraint("X_scale_tril", constraints.lower_cholesky)
-
-        self._call_base_model_guide = True
+        self.X = Parameter(self.base_model.X)
+        self.set_prior("X", dist.Normal(0, 1).expand(self.X.shape).independent(self.X.dim()))
+        self.autoguide("X", dist.Normal)
```
fehiepsi: "X" is a parameter with a prior. With `autoguide` method, there is no need to implement `pyro.sample` statements for "X" in `self.model` and `self.guide`.

pyro/contrib/gp/models/gpr.py
```diff
@@ -65,44 +65,38 @@ class GPRegression(GPModel):
...
-                 name="GPR"):
-        super(GPRegression, self).__init__(X, y, kernel, mean_function, jitter, name)
+    def __init__(self, X, y, kernel, noise=None, mean_function=None, jitter=1e-6):
+        super(GPRegression, self).__init__(X, y, kernel, mean_function, jitter)
 
-        noise = self.X.new_ones(()) if noise is None else noise
+        noise = self.X.new_tensor(1.) if noise is None else noise
         self.noise = Parameter(noise)
-        self.set_constraint("noise", torchdist.constraints.greater_than(self.jitter))
+        self.set_constraint("noise", torchdist.constraints.positive)
```
fehiepsi: Constraint for noise is changed to `positive`. This is the only change for all GP algorithms.

pyro/contrib/gp/models/gpr.py
```diff
@@ -65,44 +65,38 @@ class GPRegression(GPModel):
...
     def model(self):
         self.set_mode("model")
 
-        noise = self.get_param("noise")
-
-        N = self.X.shape[0]
+        N = self.X.size(0)
         Kff = self.kernel(self.X)
-        Kff.view(-1)[::N + 1] += noise  # add noise to diagonal
+        Kff.view(-1)[::N + 1] += self.jitter + self.noise  # add noise to the diagonal
```
fehiepsi: because `noise` is positive with no lower bound, to avoid singularity, we have to add `jitter` here

pyro/contrib/gp/models/gplvm.py
fehiepsi: Here the autoguide is `Normal` instead of `MVN`. It is just for simplicity of the model.

examples/contrib/gp/sv-dkl.py
```diff
@@ -91,42 +93,39 @@ def main(args):
                                   is_training_set=False,
                                   shuffle=True)
 
-    cnn = CNN().cuda() if args.cuda else CNN()
-
-    # optimizer in SVI just works with params which are active inside its model/guide scope;
-    # so we need this helper to mark cnn's parameters active for each `svi.step()` call.
-    def cnn_fn(x):
-        return pyro.module("CNN", cnn)(x)
```
fehiepsi: because we don't rely on ParamStoreDict, we don't need this wrapper now.

pyro/contrib/gp/models/sgpr.py
```diff
@@ -112,12 +109,10 @@ def __init__(self, X, y, kernel, Xu, noise=None, mean_function=None, approx=None
             raise ValueError("The sparse approximation method should be one of "
                              "'DTC', 'FITC', 'VFE'.")
 
+    @autoname.scope(prefix="SGPR")
```
fehiepsi: Creating a scope is in unnecessary now. It is just useful when we want to combine many gp models.

pyro/contrib/gp/util.py
```diff
@@ -1,139 +1,246 @@
...
-        self._registered_params = {}
+        self._constraints = OrderedDict()
+        self._priors = OrderedDict()
+        self._guides = OrderedDict()
+        self._mode = None
 
-        self.name = name
+    def __call__(self, *args, **kwargs):
+        with autoname.name_count():
+            return super(Parameterized, self).__call__(*args, **kwargs)
```
fehiepsi: Thanks to `autoname.name_count`, if we set priors to variances of the subkernels of `Sum(kern1, kern2)` where both of them are RBF, then latent name of `variance` will be `RBF__1/variance`, `RBF__2/variance` respectively.

pyro/contrib/gp/util.py
```diff
@@ -1,139 +1,246 @@
...
-            p = pyro.sample(param_name, prior)
-        else:  # prior != None and mode = "guide"
-            MAP_param_name = param_name + "_MAP"
-            # initiate randomly from prior
-            MAP_param = pyro.param(MAP_param_name, prior)
-            p = pyro.sample(param_name, dist.Delta(MAP_param))
-
-        self._registered_params[param] = p
+        if name in self._priors:
+            with autoname.scope(prefix=self._get_name()):
```
fehiepsi: The `scope` is put here instead of `pyro.param_with_module_name(module_name, param_name)` for simplicity, and to compatible with `autoname.name_count` decorator.

fehiepsi:
Given that reviewing this might be painful, though most of them is just minor points, it is better to close this and split this into several PRs in the future.

martinjankowiak:
@fehiepsi do you mean to get this into the release?

fehiepsi:
Hi Martin, I’m splitting this into several PRs. It might not need to be in release though I’ll try to make it possible. :)

martinjankowiak:
@fehiepsi no worries. i just won't have much time next week so this weekend would be my last chance to look at anything more carefully. but it can wait!

PR #1545, opened 2018-11-18
Revise baseball example with diagnostics info

fehiepsi:
This full request does the following:
- Make baseball example easier to follow (by removing conditioned model and renaming several variables). For predictions based on posterior, now the user only needs  to use one `TracePredictive` instance (not two instances: one for `model`, another for `conditioned_model` as before).
- Add chain diagnostics to the summary. To be able to achieve this, I modify Parallel sampler in mcmc a bit to trace the indices of samples for each chain. (addressing #1520)
- Fix the warning when using `tqdm.autonotebook` and fix the error `RuntimeError: received 0 items of ancdata` (when running multiprocessing in slow computer).
- `num_samples` in MCMC is now the number of samples in each chain.

If you obverse effective sample size > (num_samples * num_chains), then it is totally fine. Discussion about that can be found [here](https://andrewgelman.com/2018/01/18/measuring-speed-stan-incorrectly-faster-thought-cases-due-antithetical-sampling/).

### DIFF
--- a/examples/baseball.py
+++ b/examples/baseball.py
@@ -5,7 +5,6 @@ import logging
 import math
 import os
 
-import numpy as np
 import pandas as pd
 import torch
 
@@ -16,6 +15,7 @@ from pyro.distributions.util import logsumexp
 from pyro.infer import EmpiricalMarginal
 from pyro.infer.abstract_infer import TracePredictive
 from pyro.infer.mcmc import MCMC, NUTS
+from pyro.ops.stats import effective_sample_size, split_gelman_rubin
 
 """
 Example has been adapted from [1]. It demonstrates how to do Bayesian inference using
@@ -64,34 +64,37 @@ DATA_URL = "https://d2fefpcigoriu7.cloudfront.net/datasets/EfronMorrisBB.txt"
 # ===================================
 
 
-def fully_pooled(at_bats):
+def fully_pooled(at_bats, hits):
     r"""
     Number of hits in $K$ at bats for each player has a Binomial
     distribution with a common probability of success, $\phi$.
 
     :param (torch.Tensor) at_bats: Number of at bats for each player.
+    :param (torch.Tensor) hits: Number of hits for the given at bats.
     :return: Number of hits predicted by the model.
     """
     phi_prior = Uniform(at_bats.new_tensor(0), at_bats.new_tensor(1))
     phi = pyro.sample("phi", phi_prior)
-    return pyro.sample("obs", Binomial(at_bats, phi))
+    return pyro.sample("obs", Binomial(at_bats, phi), obs=hits)
 
 
-def not_pooled(at_bats):
+def not_pooled(at_bats, hits):
     r"""
     Number of hits in $K$ at bats for each player has a Binomial
     distribution with independent probability of success, $\phi_i$.
 
     :param (torch.Tensor) at_bats: Number of at bats for each player.
+    :param (torch.Tensor) hits: Number of hits for the given at bats.
     :return: Number of hits predicted by the model.
     """
     num_players = at_bats.shape[0]
+    # TODO: use pyro.plate when pytorch 1.0 is released
     phi_prior = Uniform(at_bats.new_tensor(0), at_bats.new_tensor(1)).expand_by([num_players]).independent(1)
     phi = pyro.sample("phi", phi_prior)
-    return pyro.sample("obs", Binomial(at_bats, phi))
+    return pyro.sample("obs", Binomial(at_bats, phi), obs=hits)
 
 
-def partially_pooled(at_bats):
+def partially_pooled(at_bats, hits):
     r"""
     Number of hits has a Binomial distribution with independent
     probability of success, $\phi_i$. Each $\phi_i$ follows a Beta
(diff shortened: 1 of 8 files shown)


examples/baseball.py
```diff
@@ -147,29 +143,47 @@ def get_site_stats(array, player_names):
...
+                                                                 r_hat=r_hat[i].numpy())
     return site_stats
 
 
+def chain_diagnostics(samples, chain_indices):
+    assert samples.size(0) == chain_indices.numel()
+    chain_samples = samples[chain_indices.reshape(-1)].reshape(
+        chain_indices.shape + samples.shape[1:])
+    n_eff = effective_sample_size(chain_samples, sample_dim=1, chain_dim=0)
+    r_hat = split_gelman_rubin(chain_samples, sample_dim=1, chain_dim=0)
```
fehiepsi: @neerajprad to avoid confusion, should we make chain_dim=0, sample_dim=1 by default in `stats`?

neerajprad: I think so too - I think that would be a better convention to choose.

fehiepsi: I'll do it in #1525 if this pull request is fine in its own. Otherwise, I will make the change here too. :)

neerajprad: Sure, let us address it in #1525.

pyro/infer/mcmc/mcmc.py
```diff
@@ -137,6 +135,17 @@ def terminate(self):
                 w.terminate()
 
     def _traces(self, *args, **kwargs):
+        chain_indices = [[] for _ in range(self.num_chains)]
+        try:
```
neerajprad: I think we should modify the TracePosterior class itself to store chain information, and plug in diagnostics there. I am fine with merging this if we need it earlier, but we should refactor this soon. I can take a look at it this week.

examples/baseball.py
```diff
@@ -147,29 +143,47 @@ def get_site_stats(array, player_names):
...
-def summary(traces, sites, player_names, transforms={}):
+def summary(trace_posterior, sites, player_names, transforms={}, diagnostics=True):
     """
     Return summarized statistics for each of the ``sites`` in the
     traces corresponding to the approximate posterior.
     """
-    marginal = EmpiricalMarginal(traces, sites).get_samples_and_weights()[0].numpy()
+    marginal = EmpiricalMarginal(trace_posterior, sites).get_samples_and_weights()[0]
+
+    if diagnostics and trace_posterior.sampler.num_chains > 1:
```
neerajprad: With `num_chains=1`, sampler will not have `num_chains` set, and this will throw an error.

examples/baseball.py
```diff
@@ -100,41 +105,32 @@ def partially_pooled(at_bats):
...
     :return: Number of hits predicted by the model.
     """
     num_players = at_bats.shape[0]
     loc = pyro.sample("loc", Normal(at_bats.new_tensor(-1), at_bats.new_tensor(1)))
     scale = pyro.sample("scale", HalfCauchy(scale=at_bats.new_tensor(1)))
     alpha = pyro.sample("alpha", Normal(loc, scale).expand_by([num_players]).independent(1))
-    return pyro.sample("obs", Binomial(at_bats, logits=alpha))
-
-
-def conditioned_model(model, at_bats, hits):
```
neerajprad: How does this work when we need to sample from the posterior predictive? Does this rely on us calling the model with `hits=None`. Do the results remain the same with this refactoring?

examples/baseball.py
```diff
@@ -53,6 +52,9 @@
     path lengths in Hamiltonian Monte Carlo", (https://arxiv.org/abs/1111.4246)
 """
 
+# work around with the error "RuntimeError: received 0 items of ancdata"
```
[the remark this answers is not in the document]
fehiepsi: yup, it is in linux. This error happens every time I use multiprocessing with pytorch. Things run fine for while, then bump, this error happens. T T

examples/baseball.py
fehiepsi: you are right, I'll change to trace_posterior.num_chains instead.

examples/baseball.py
fehiepsi: It seems so. Why do you expect that things will be different? I'll check it btw.

fehiepsi: @neerajprad They give identical results for the fully pooled model.

fehiepsi: Yes, passing `obs=None` makes it a latent site, not observe site. The difference between two versions is at the `_RETURN` node. With conditioning, model returns sample at `obs`. With `obs=None`, model returns `None`. I remember so.

neerajprad: I think that's fine for as long as we aren't using the return value.

neerajprad:
@fehiepsi - I am fine merging this PR as is, let me know if this is good to merge? I'll refactor the `TracePosterior` class to store chain indices, so we may have to change some of this code soon, but it will be good to have this in dev as a comparison.

fehiepsi:
@neerajprad Yes, this is good to merge. We can discuss what is the best way in your refactoring pull request. :+1:

PR #1547, opened 2018-11-19
Getting parameters from TransformModule & TransformedDistribution

stefanwebb:
This PR is about discussing a better method for getting parameters out of a transformed distribution.

Previously, you had to do e.g., `x = PlanarFlow(2); x.module.parameters()` to extract the parameters from a single transform. To get the parameters out of a sequence of `Transform`s was tricky because you had to extract the `.module`s from each transform and it was decoupled from `TransformedDistribution`.

I'm suggesting a two-part fix. Firstly, a new class `TransformModule` that turns transforms into true `nn.Module`s, so you can do e.g. `x = PlanarFlow(2); x.parameters()` if `PlanarFlow` inherits from `TransformModule`. Secondly, also making `TransformedDistribution` a true `nn.Module`. Then you can simply pass the parameters from the transformed distribution to e.g. an optimizer by doing `flow_dist = TransformedDistributions(base_dist, flows); optimizer = torch.optim.Adam(flow_dist.parameters())`

### DIFF
--- a/examples/dmm/dmm.py
+++ b/examples/dmm/dmm.py
@@ -150,7 +150,7 @@ class DMM(nn.Module):
 
         # if we're using normalizing flows, instantiate those too
         self.iafs = [InverseAutoregressiveFlow(AutoRegressiveNN(z_dim, [iaf_dim])) for _ in range(num_iafs)]
-        self.iafs_modules = nn.ModuleList([self.iafs])
+        self.iafs_modules = nn.ModuleList(self.iafs)
 
         # define a (trainable) parameters z_0 and z_q_0 that help define the probability
         # distributions p(z_1) and q(z_1)
--- a/examples/dmm/dmm.py
+++ b/examples/dmm/dmm.py
@@ -150,7 +150,7 @@ class DMM(nn.Module):
 
         # if we're using normalizing flows, instantiate those too
         self.iafs = [InverseAutoregressiveFlow(AutoRegressiveNN(z_dim, [iaf_dim])) for _ in range(num_iafs)]
-        self.iafs_modules = nn.ModuleList([iaf.module for iaf in self.iafs])
+        self.iafs_modules = nn.ModuleList([self.iafs])
 
         # define a (trainable) parameters z_0 and z_q_0 that help define the probability
         # distributions p(z_1) and q(z_1)
--- a/pyro/distributions/torch.py
+++ b/pyro/distributions/torch.py
@@ -282,9 +282,24 @@ class StudentT(torch.distributions.StudentT, TorchDistributionMixin):
             return type(self)(df, loc, scale, validate_args=validate_args)
 
 
-class TransformedDistribution(torch.distributions.TransformedDistribution, TorchDistributionMixin):
+class TransformedDistribution(torch.distributions.TransformedDistribution, TorchDistributionMixin, torch.nn.Module):
+    def __init__(self, *args, **kwargs):
+        torch.distributions.TransformedDistribution.__init__(self, *args, **kwargs)
+        TorchDistributionMixin.__init__(self)
+        torch.nn.Module.__init__(self)
+        self.transforms = torch.nn.ModuleList(self.transforms)
+
     def expand(self, batch_shape):
         return super(TransformedDistribution, self).expand(batch_shape)
+
+
+class TransformModule(torch.distributions.Transform, torch.nn.Module):
+    def __init__(self):
+        torch.distributions.Transform.__init__(self)
+        torch.nn.Module.__init__(self)
+
+    def __hash__(self):
+        return super(torch.nn.Module, self).__hash__()
 
 
 class Uniform(torch.distributions.Uniform, TorchDistributionMixin):
@@ -324,6 +339,7 @@ for _name, _Dist in torch.distributions.__dict__.items():
 
     __all__.append(_name)
 
+__all__.append('TransformModule')
 
 # Create sphinx documentation.
 __doc__ = '\n\n'.join([
--- a/pyro/distributions/torch.py
+++ b/pyro/distributions/torch.py
@@ -287,15 +287,6 @@ class TransformedDistribution(torch.distributions.TransformedDistribution, Torch
         return super(TransformedDistribution, self).expand(batch_shape)
 
 
-class TransformModule(torch.distributions.Transform, torch.nn.Module):
-    def __init__(self):
-        torch.distributions.Transform.__init__(self)
-        torch.nn.Module.__init__(self)
-
-    def __hash__(self):
-        return super(torch.nn.Module, self).__hash__()
-
-
 class Uniform(torch.distributions.Uniform, TorchDistributionMixin):
     def expand(self, batch_shape):
         try:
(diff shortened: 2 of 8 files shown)


pyro/distributions/torch.py
```diff
@@ -282,11 +282,28 @@ def expand(self, batch_shape):
             return type(self)(df, loc, scale, validate_args=validate_args)
 
 
-class TransformedDistribution(torch.distributions.TransformedDistribution, TorchDistributionMixin):
+class TransformedDistribution(torch.distributions.TransformedDistribution, TorchDistributionMixin, torch.nn.Module):
```
neerajprad: What are the specific methods we would like to inherit from `nn.Module`? I am wondering if it safe to inherit off of `Module` for all `TransformedDistribution` and tying ourselves up in terms of future compatibility. If it is only a couple of methods like `.parameters()` and `.cuda()`, we could provide implementations for these (and use composition when needed to call `.module.cuda()` under the hood, for instance).

pyro/distributions/torch.py
```diff
@@ -282,11 +282,28 @@ def expand(self, batch_shape):
...
+    def __init__(self, *args, **kwargs):
+        torch.distributions.TransformedDistribution.__init__(self, *args, **kwargs)
+        TorchDistributionMixin.__init__(self)
+        torch.nn.Module.__init__(self)
+
+        # Not all transforms have parameters, so only want to extract nn.Module ones
+        self.transform_modules = torch.nn.ModuleList([t for t in self.transforms if isinstance(t, torch.nn.Module)])
```
neerajprad: This will register the parameters of the transform for use in `.parameters` and `.cuda`. Are there cases where we would have to separately grab the base distribution's parameters? If so, it might give a false impression to the user that they have transferred their parameters to the GPU by calling `transformed_dist.cuda()`, but they'll get an error later because their base distribution's params will still be on the CPU.

pyro/distributions/torch.py
stefanwebb: I think basically all the main ones, like `apply`, `cpu`, `cuda`, `load_state_dict`, `named_parameters`, `parameters`, `state_dict`, `to`. IMHO, I think future compatibility wouldn't be a problem since `nn.Module` is one of the main abstraction mechanisms in PyTorch and it's been around for so long

pyro/distributions/torch.py
stefanwebb: Hmm, at the moment you can't do `.cuda()` or `.parameters()` on a distribution (Fritz explains the rationale [here](https://github.com/pytorch/pytorch/issues/7795#issuecomment-391807661)). I guess the difficulty is created because a transformed distribution mixes objects with and without trainable parameters... I've added in an extra bit of code that adds a base distribution that is also an `nn.Module` to the module list

stefanwebb: I agree it would be confusing if a user calls `.cuda()` and their base distribution in on the `.cpu()`. Although I can't see a good solution. It doesn't seem possible to add a `.cuda()` to distributions in general for the reasons in that issue above. On the other hand, there's a need to have `.cuda()` for normalizing flows, since they have trainable parameters... What about overriding `.cuda()`, `.to()` and `.cpu()` in `TransformedDistribution` to raise an exception? Then the preferred way to transform your normalizing flow to the GPU would be to call `.cuda()` on each `TransformModule` separately (or in a list comprehension)?

stefanwebb: The more I think about this, the more I prefer not making `TransformedDistribution` an `nn.Module`, keeping it lightweight like `Distribution`, and doing all the work with `TransformModule`. I was just looking over the DMM example in Pyro, and I think in a typical use case, you would make an `nn.ModuleList` of `Transforms` inside your model (inheriting from `nn.Module`), and create a `TransformedDistribution` on the fly as needed

neerajprad: I agree - I think that will be a cleaner, more straightforward abstraction!

examples/dmm/dmm.py
```diff
@@ -150,7 +150,7 @@ def __init__(self, input_dim=88, z_dim=100, emission_dim=100,
 
         # if we're using normalizing flows, instantiate those too
         self.iafs = [InverseAutoregressiveFlow(AutoRegressiveNN(z_dim, [iaf_dim])) for _ in range(num_iafs)]
-        self.iafs_modules = nn.ModuleList([iaf.module for iaf in self.iafs])
+        self.iafs_modules = nn.ModuleList([self.iafs])
```
neerajprad: Remove the square brackets, since `self.iafs` is already a list.

pyro/distributions/torch.py
```diff
@@ -287,6 +287,15 @@ def expand(self, batch_shape):
         return super(TransformedDistribution, self).expand(batch_shape)
 
 
+class TransformModule(torch.distributions.Transform, torch.nn.Module):
+    def __init__(self):
+        torch.distributions.Transform.__init__(self)
+        torch.nn.Module.__init__(self)
+
+    def __hash__(self):
+        return super(torch.nn.Module, self).__hash__()
```
neerajprad: nit: I think it will be simpler to just use `hash(self)`.

stefanwebb: This won't work because you need to call the `.__hash__()` from the parent class. Doing `.hash(self)` enters an infinite recursion. Is there a neater way to call the parent class method?

examples/dmm/dmm.py
stefanwebb: Ah sorry, yes of course! :)

pyro/distributions/torch.py
neerajprad: Sorry, what you have is correct! I was confused - hash(self) would just call `__hash__` method (as it should, I thought it would use `object.__hash__`) itself and go into infinite recursion.

pyro/distributions/torch.py
```diff
@@ -324,6 +333,7 @@ def expand(self, batch_shape):
 
     __all__.append(_name)
 
+__all__.append('TransformModule')
```
neerajprad: I think this can be removed since you are adding this in `distributions.__init__.py`. 

nit: I think a better place for this class might be `torch_distribution.py` which has other mixin classes too.

pyro/distributions/torch.py
fritzo: We may need to avoid using `@lazy_property`, or add something to reset all lazy properties when calling `.cpu()`, `.cuda()`, and `.to()`.

pyro/distributions/torch.py
```diff
@@ -287,6 +287,15 @@ def expand(self, batch_shape):
         return super(TransformedDistribution, self).expand(batch_shape)
 
 
+class TransformModule(torch.distributions.Transform, torch.nn.Module):
+    def __init__(self):
+        torch.distributions.Transform.__init__(self)
```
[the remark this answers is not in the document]
stefanwebb: Sounds good! I've made all these changes

stefanwebb:
I also added in docs for `TransformModule`

pyro/distributions/torch_patch.py
```diff
@@ -45,6 +45,19 @@ def _torch_dirichlet_grad(x, concentration, total):
     return unpatched_fn(x, concentration, total)
 
 
+@_patch('torch.distributions.transforms.Transform.__init__')
```
fritzo: Could you just add a comment
```py
# this can be removed when super(...).__init__() is added upstream
```

stefanwebb: Done!

Issue #1451, opened 2018-10-16
[FR] Add tutorial on implementing new effects


fritzo:
Users at PROBPROG 2018 requested a tutorial on high-level Pyro architecture including advanced features like poutines.  This issue proposes adding a tutorial on implementing a new effect handler.

Whereas #1553 should aim to explain Pyro's architecture in a simplified way, this tutorial should prepare developers to make changes to Pyro.

robsalomone:
Hi guys, has there been any progress with this issue? Would been keen to see any current drafts of tutorials and provide feedback. This sort of tutorial really is necessary I think to get the scientific community on board to implement new methodology they are working on (or in general, to encourage and make contributions easier).

eb8680:
@robsalomone sorry, I've been busy with other things and haven't had time to get to this.  It'll definitely be included with the [0.3.0 release](https://github.com/uber/pyro/milestone/5), which will hopefully be out in a few weeks.  I agree it's important for encouraging contributors and I'll cc you when I open a PR with a draft.

eb8680:
@robsalomone have a look at [minipyro.py](https://github.com/uber/pyro/blob/dev/pyro/contrib/minipyro.py) and #1629

PR #1553, opened 2018-11-20
Sketch mini-pyro implementation

fritzo:
Addresses #1451, #1554

This implements "Pyro in 150 lines" including an example using SVI to fit a normal-normal model.

My objective is to create a minimal example, and then add some deeper poutine usage in subsequent PRs. I'd like to keep total line count under 500, and add documentation as needed. In lieu of a hackathon, I suggest all Pyro team members fit minified versions of their ideas into this example (in separate PRs).

## Questions for reviewers

- [x] @eb8680 can you suggest minimal changes to this poutine stuff? I'd be happy merging this first version and letting you take a second pass.

## Tested

- [x] added to test/test_examples.py

### DIFF
--- a/examples/mini_pyro.py
+++ b/examples/mini_pyro.py
@@ -12,7 +12,7 @@ PARAM_STORE = {}
 
 
 class Messenger(object):
-    def __init__(self, fn):
+    def __init__(self, fn=None):
         self.fn = fn
 
     def __enter__(self):
@@ -22,6 +22,12 @@ class Messenger(object):
         assert PYRO_STACK[-1] is self
         PYRO_STACK.pop()
 
+    def process_message(self, msg):
+        pass
+
+    def postprocess_message(self, msg):
+        pass
+
     def __call__(self, *args, **kwargs):
         with self:
             return self.fn(*args, **kwargs)
@@ -29,11 +35,12 @@ class Messenger(object):
 
 class trace(Messenger):
     def __enter__(self):
+        super(trace, self).__enter__()
         self.trace = OrderedDict()
-        return super(trace, self).__enter__()
+        return self.trace
 
-    def process_msg(self, msg):
-        self.trace[msg["name"]] = msg
+    def postprocess_message(self, msg):
+        self.trace[msg["name"]] = msg.copy()
 
     def get_trace(self, *args, **kwargs):
         self(*args, **kwargs)
@@ -45,7 +52,7 @@ class replay(Messenger):
         self.guide_trace = guide_trace
         super(replay, self).__init__(fn)
 
-    def process_msg(self, msg):
+    def process_message(self, msg):
         if msg["name"] in self.guide_trace:
             msg["value"] = self.guide_trace[msg["name"]]["value"]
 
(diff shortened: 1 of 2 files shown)


examples/mini_pyro.py
```diff
@@ -0,0 +1,148 @@
...
+    def __init__(self, model, guide, optim, loss):
+        self.model = model
+        self.guide = guide
+        self.optim = optim
+        self.loss = loss
+
+    def step(self, *args, **kwargs):
+        loss = self.loss(self.model, self.guide, *args, **kwargs)
+        loss.backward()
+        params = list(PARAM_STORE.values())
```
eb8680: We should only grab params that appear in `loss`

examples/mini_pyro.py
```diff
@@ -0,0 +1,148 @@
...
+            return self.fn(*args, **kwargs)
+
+
+class trace(Messenger):
+    def __enter__(self):
+        self.trace = OrderedDict()
+        return super(trace, self).__enter__()
+
+    def process_msg(self, msg):
+        self.trace[msg["name"]] = msg
```
eb8680: `self.trace[msg["name"]] = msg.copy()`

eb8680:
Also, in terms of organization, maybe we should make this into a notebook and automatically extract a script with `nbconvert`?

fritzo:
> maybe we should make this into a notebook

While I like the idea of publishing this as a tutorial that shows up on docs.pyro.ai, I find it frustrating to edit, refactor, test, profile, and import jupyter notebooks. @jpchen Is there a way we can make a nice html page out of this file?

jpchen:
probably. i think the tutorials webpage is a good home for it. not sure how well it plays with sphinx as a standalone html. at the very least it can be exported into a one-cell notebook using an `nbformat` script

fritzo:
> missing ... block, plate/markov, and parallel enumeration.

Yeah I'd really like to see `block`, since users seem most confused about poutines and messengers. I feel like `plate` might be useful to see how broadcasting works, but I think this file will be most helpful if we focus on the nonstandard-interpretation aspects and leave tensor shape explanation for elsewhere. Similarly, enumeration would be nice to demonstrate nonstandard interpretation, but I think it's worth implementing a limited version that assumes a single plate nesting, so as to focus on the PL aspects.

eb8680:
> users seem most confused about poutines and messengers

I guess it depends on the goals of this example - if the goal is to explain effect handlers, then that's fine, but if the goal is to explain how Pyro works then it seems like many more people will want to know how `TraceEnum_ELBO` works, since that's currently our main selling point and most useful feature.  That means at least explaining `plate` and parallel enumeration via broadcasting and `ubersum`.

martinjankowiak:
my two cents is that this should mostly focus on examples of non-standard interpretation stuff and that 2-3 instantiations of that concept should be sufficient. it'd be a shame if too much complexity scared people off or confused them. if we want tutorials that explain how, e.g., traceenum works, i think that should go elsewhere.

eb8680:
> my two cents is that this should mostly focus on examples of non-standard interpretation stuff and that 2-3 instantiations of that concept should be sufficient. it'd be a shame if too much complexity scared people off or confused them

I think you're probably right that building all the way up to enumeration would be too much material, but on the other hand, without a meatier example of some kind as payoff at the end I'm not sure who this tutorial would be useful/interesting for.  What do you think we should build up to instead?

PR #1556, opened 2018-11-21
Move nonstrict name functionality into contrib.autoname

eb8680:
This PR moves some little-used autonaming functionality out of `TraceMessenger` and into a new function `contrib.autoname.scoping.name_count`, thereby decluttering `TraceMessenger` considerably and removing the `strict_names` keyword argument to `poutine.trace`.  It also changes the counter suffix from `_i` to `__i` to reduce collisions with user-specified names and uses iteration instead of recursion.

This is a breaking change to `TraceMessenger` and `poutine.trace`, though I doubt anyone (except me) was actually using this feature given that it was never really documented and was disabled by default.

Before:
```python
def model():
    for i in range(3):
        pyro.sample("x", dist.Bernoulli(0.5))

tr = poutine.trace(model, strict_names=False).get_trace()
assert "x" in tr
assert "x_0" in tr
assert "x_1" in tr
```

After
```python
@contrib.autoname.name_count
def model():
    for i in range(3):
        pyro.sample("x", dist.Bernoulli(0.5))

tr = poutine.trace(model).get_trace()
assert "x" in tr
assert "x__0" in tr
assert "x__1" in tr
```


fehiepsi:
@eb8680 this is a nice feature! I am going to use it for GP clean up. Just wonder if we can start enumerate from 1, instead of 0?

eb8680:
@fehiepsi sure, feel free to open a PR.

PR #1557, opened 2018-11-21
 Add diagnostics method to TracePosterior class

neerajprad:
Addresses #1560. 

This refactors the `mcmc` module as follows:
 - Holds chain id information in `TracePosterior`, and ensures that we are able to deterministically sample from the `TracePosterior` object in the presence of multiple chains. This allows us to remove the gnarly code that I introduced in #1506 for this purpose. 
 - Adds a diagnostics method to an MCMC specific `EmpiricalMarginal` which returns stats from @fehiepsi's `ops.stats` (currently effective sample size and gelman rubin statistic), but more will be added. I initially tried to share the code between the MCMC and Importance `EmpiricalMarginal` but it turned out to be quite convoluted getting existing oed and importance code to work given that the new implementation does not override Empirical and has slightly different interface. As such, I have tried to not refactor the importance and search code. There is more code sharing that could be done, but the benefits are questionable.

The baseball example is modified to use this and is a minor refactoring over @fehiepsi's previous PR - #1545. 

### TODO:
 - [x] Ensure that deterministic sampling order is preserved despite samples being stored out of order in `EmpiricalMarginal`.
 - [x] Fix `mean`, `variance` in `Empirical` in the presence of multiple chains.
 - [x] Test diagnostics is correctly computed for different latent shapes.
 - [x] Test diagnostics is correctly computed when run on CUDA.


fehiepsi:
Thanks @neerajprad ! All the changes make sense to me. I guess you would also need to adapt svi _traces to this change? When you finish, I'll go for another pass.

neerajprad:
@fehiepsi - I think it will be better to move this into the empirical distribution. That way anyone should be able to access the tensor directly and be able to see samples for each chain (e.g. if we would like to have something like traceplot). I will make that change in this PR.

fehiepsi:
@neerajprad - Agree! As we discussed on Monday, a restriction is EmpiricalMarginal requires samples have a same shape. I am waiting to see your idea. Here is an api I have in mind
```
mcmc = MCMC(...).run()
marginal = EmpiricalMarginal(mcmc, sites=["a", "b", "c"])  # sites a, b, c does not need to have a same shape; hence empirical marginal is just a container of Empirical for each site
marginal.effective_sample_size()  # return a dict of `n_eff` with keys "a", "b", "c"
marginal.gelman_rubin()  # return a dict of `r_hat` with keys "a", "b", "c"
marginal.sites["a"]  # return a empirical distribution for "a"
marginal.sites["a"].effective_sample_size()  # return n_eff with keys "a"
```

neerajprad:
> Basically, I think that all the computation should go under empirical distribution class. And EmpiricalMarginal is just a convenient wrapper for a list of sites.

I think that's a good idea! In that case the restriction on the empirical distribution does not matter, we can have a separate empirical distribution for each of the marginals.

neerajprad:
Since that will be a breaking change, I created #1557 to gather more opinions / suggestions.

pyro/distributions/empirical.py
```diff
@@ -165,11 +182,23 @@ def variance(self):
         deviation_squared = torch.pow(self._samples - self.mean, 2)
         return self._weighted_mean(deviation_squared)
 
-    def get_samples_and_weights(self):
-        self._finalize()
-        return self._samples, self._log_weights
-
+    @_finalized
+    def get_data(self, flatten=False):
```
fehiepsi: This name might confuse users for what `data` means in the following context:
```
marginal = inference.run(data).marginal()
marginal.empirical()["a"].get_data()
```
How about `get_support`? or `get_underlined_samples`?

pyro/infer/mcmc/mcmc.py
```diff
@@ -278,8 +251,61 @@ def __init__(self, kernel, num_samples, warmup_steps=0,
...
 
     def _traces(self, *args, **kwargs):
         for sample in self.sampler._traces(*args, **kwargs):
             yield sample
+
+    def marginal(self, sites=None):
+        return _EmpiricalMarginal(self, sites)
+
+
+class _EmpiricalMarginal(object):
```
fehiepsi: I think we can make this inherited from `abstract_infer.EmpiricalMarginal`.

pyro/infer/mcmc/mcmc.py
```diff
@@ -278,8 +251,61 @@ def __init__(self, kernel, num_samples, warmup_steps=0,
...
+    """
+    Marginal distribution for MCMC that provides a a marginal over one or more
+    latent sites as well as the return values of the TracePosterior's model.
+
+    :param TracePosterior trace_posterior: a TracePosterior instance representing
+        a Monte Carlo posterior.
+    :param list sites: optional list of sites for which we need to generate
+        the marginal distribution. Note that for multiple sites, the shape
+        for the site values must match (needed by the underlying ``Empirical``
+        class).
```
fehiepsi: nit: no need for shape matching.

pyro/infer/mcmc/mcmc.py
```diff
@@ -278,8 +251,61 @@ def __init__(self, kernel, num_samples, warmup_steps=0,
...
+            return self._diagnostics
+        for site in self.sites:
+            self._diagnostics[site] = OrderedDict([
+                ("n_eff", stats.effective_sample_size(self._marginals[site].get_data())),
+                ("r_hat", stats.split_gelman_rubin(self._marginals[site].get_data()))
+            ])
+        return self._diagnostics
+
+    def empirical(self):
+        return self._marginals
```
fehiepsi: I think we can make this a property.

fehiepsi: Or we can make EmpiricalMarginal a dict-like object. If you agree, I'll do that in another pull request.

pyro/distributions/empirical.py
```diff
@@ -25,49 +47,47 @@ def __init__(self, validate_args=None):
...
         """
-        self._finalize()
         if self._samples is None:
             return 0
+        if self._num_chains > 1:
+            return self._samples[:2].numel()
         return self._samples.size(0)
 
-    def add(self, value, weight=None, log_weight=None):
+    def add(self, value, weight=None, log_weight=None, chain_id=0):
```
fehiepsi: missing docs for `chain_id`

pyro/infer/mcmc/mcmc.py
```diff
@@ -120,8 +121,8 @@ def __init__(self, kernel, num_samples, warmup_steps, num_chains, mp_context):
     def init_workers(self, *args, **kwargs):
         self.workers = []
         for i in range(self.num_chains):
-            worker = _Worker(i + 1, self.result_queue, self.log_queue, self.kernel,
-                             self.num_samples[i], self.warmup_steps)
+            worker = _Worker(i, self.result_queue, self.log_queue, self.kernel,
+                             self.num_samples, self.warmup_steps)
```
fehiepsi: +1

pyro/distributions/empirical.py
neerajprad: I just realized that we have ``.enumerate_support()`` method, so let us use that directly and remove this altogether.

pyro/infer/mcmc/mcmc.py
neerajprad: `EmpiricalMarginal` inherits from the `Empirical` distribution, so that won't work (at least not without creating other problems). The main issue is that this needs to hold an `Empirical` distribution per latent, therefore it cannot inherit from `Empirical` distribution without relaxing the shape requirements for latents.

neerajprad: Also, if we make `EmpiricalMarginal` not inherit from Empirical, it creates issues with importance sampling.

pyro/infer/mcmc/mcmc.py
neerajprad: Hmm..I'm not sure if that will work. Can you elaborate?

fehiepsi:
Sure ;)

pyro/infer/mcmc/mcmc.py
fehiepsi: This is what I have in mind (not sure it is indeed useful at all):

```
def items(self):
    for name in self._marginals:
        yield name, self[name]

def keys(self):
    return self._marginals.keys()

def values(self):
    for name, empirical in self.items():
        yield empirical

def __bool__(self):
    return bool(self._marginals)

def __len__(self):
    return len(self._marginals)

def __contains__(self, name):
    return name in self._marginals

def __iter__(self):
    return iter(self.keys())

def __delitem__(self, name):
    self.sites.remove(value)
    del self._marginals[name]
    if self._diagnostics:
        del self._diagnostics[name]

def __getitem__(self, name):
    return self._marginals[name]
    
def __setitem__(self, name, new_empirical):
    if name not in self.sites:
        self.sites.append(name)
    self._marginals[name] = new_empirical
    if self._diagnostics:
         self._diagnostics[name] = OrderedDict([...])
```

neerajprad: We have changed the API as follows:

Earlier:
```python
mcmc_run = MCMC(hmc_kernel, ..).run(data)
marginal = EmpiricalMarginal(mcmc_run, sites=["a", "b"]).get_dist_and_values()[0]
# this will only work if a and b had the same shape
# version below does not have this issue.
samples_a = marginal[:, 1]
samples_b = marginal[:, 2]
```

Now:
```python
mcmc_run = MCMC(hmc_kernel, ..).run(data)
marginal = mcmc_run.marginal(["a", "b"])
samples_a = mcmc_run.empirical["a"].enumerate_support()
samples_b = mcmc_run.empirical["b"].enumerate_support()
```

The second version has an additional indirection, because `.marginal()` is not a distribution unlike `EmpiricalMarginal`, and we need to call `.empirical` to get the underlying empirical distribution. We can provide a convenience method called `.support` which can do this under the hood though:

```python
mcmc_run = MCMC(hmc_kernel, ..).run(data)
marginal = mcmc_run.marginal(["a", "b"]).support
samples_a = marginal["a"]
samples_b = marginal["b"]
```

I don't completely follow how the dict interface will simplify this.

fehiepsi: @neerajprad I think that
```
mcmc_run = MCMC(hmc_kernel, ..).run(data)
marginal = mcmc_run.marginal(["a", "b"])
samples_a = marginal.empirical["a"].enumerate_support()
samples_b = marginal.empirical["b"].enumerate_support()
```
is pretty clean! Please don't worry about `dict`.

pyro/distributions/empirical.py
```diff
@@ -151,7 +163,7 @@ def mean(self):
                              "or ``torch.float64``. If these are samples from a " +
                              "`Categorical` distribution, consider converting to a " +
                              "`OneHotCategorical` distribution.")
-        return self._weighted_mean(self._samples)
+        return self._weighted_mean(self._flatten_chain(self._samples))
```
neerajprad: Note that to keep the implementation simple, this simply gives the `mean`/`variance` across all chains. If the user needs to implement chain specific mean they can get all samples via `.enumerate_support(flatten=False)`

eb8680:
I think the interfaces are fine and I'd be happy to merge this as-is, but I wonder if it makes sense to factorize the internals slightly differently:
1. Have a basic vectorized `Empirical` distribution that would be constructed from already-`_finalize`d sample and weight tensors, whose `batch_shape` is the same as a `Categorical` constructed from the weights and whose `event_shape` is determined by aligning the `batch_shape` and the sample tensor shape on the left.
2. Move all of the `_finalize` logic into `_Marginal.populate_traces`, so that the full sample and weight tensors are constructed there; for each site, construct a single sample/weight tensor pair whose leftmost dimension is the chain dimension and use that to construct an `Empirical` marginal for the site
3. Move the diagnostic logic into `_Marginal` (or into an `MCMCMarginal` distribution that inherits from `Empirical` that is constructed for each site and knows which dimension is the chain dimension)

@neerajprad you may have already tried something like this and found problems with it, WDYT?

neerajprad:
> @neerajprad you may have already tried something like this and found problems with it, WDYT?

I think it's a good idea to implement (1) and (2), and have consistent semantics for vectorizing! For (3), I think at that point we might as well just put that logic in `EmpiricalMarginal`, and each site's marginal can be an instance of that. `Let me try to refactor, and get back on this.

neerajprad:
This should be good for review again, with the design changed as per @eb8680's suggestions.

pyro/distributions/empirical.py
```diff
@@ -62,89 +44,47 @@ def sample_size(self):
...
-        idxs = self._categorical.sample(sample_shape=sample_shape)
-        return self._samples[idxs]
+        sample_idx = self._categorical.sample(sample_shape)
+        return self._samples[sample_idx]
 
     def log_prob(self, value):
         """
         Returns the log of the probability mass function evaluated at ``value``.
         Note that this currently only supports scoring values with empty
-        ``sample_shape``, i.e. an arbitrary batched sample is not allowed.
```
neerajprad: @eb8680 - This PR is getting a bit unwieldy, but I will generalize this to an arbitrary `sample_shape` in a follow-up and add some tests. Currently samples are restricted to be of shape `batch_shape + event_shape` but it should be easy to have arbitrary batch dims on the left.

pyro/distributions/empirical.py
```diff
@@ -155,21 +95,20 @@ def mean(self):
...
+        mean = self.mean.unsqueeze(self._aggregation_dim)
+        deviation_squared = torch.pow(self._samples - mean, 2)
         return self._weighted_mean(deviation_squared)
 
-    def get_samples_and_weights(self):
-        self._finalize()
-        return self._samples, self._log_weights
+    @property
+    def weights(self):
+        return self._log_weights
```
fehiepsi: I think we need to do some math here to get the weight. Otherwise, we might rename `weights` to `log_weights`.

neerajprad: I think we can rename this to `log_weights`, since `Empirical` only accepts `log_weights`. I'll change to use `log_weights` everywhere, we don't need to make the interface confusing when the user can use a one liner to transform their weights.

PR #1558, opened 2018-11-21
Fixed bug where Empirical.mean() and Empirical.variance() could return nan

adam-coogan:
Fixed bug in https://github.com/uber/pyro/issues/1548.

Previously, `Empirical.mean()` and `Empirical.variance()` returned `nan` if the largest sample probability was rounded to zero. The `Empirical._weighted_mean()` helper now normalizes by dividing by the sum of the sample probabilities, with an algebraic simplification to avoid multiplying by `max_val.exp()`. A simple test was added to verify the bug has been fixed.

The same `_weighted_mean()` and usage pattern appeared in two other files, which have been similarly corrected.

### DIFF
--- a/examples/rsa/search_inference.py
+++ b/examples/rsa/search_inference.py
@@ -125,12 +125,10 @@ class HashingMarginal(dist.Distribution):
             return d
 
     def _weighted_mean(self, value, dim=0):
-        weights = self._dist_and_values()[0].logits
-        for _ in range(value.dim() - 1):
-            weights = weights.unsqueeze(-1)
-        max_val = weights.max(dim)[0]
-        normalization = (value.new_tensor([1.]) * (weights - max_val.unsqueeze(-1)).exp()).sum(dim=dim)
-        return (value * (weights - max_val.unsqueeze(-1)).exp()).sum(dim=dim) / normalization
+        weights = self._log_weights.reshape([-1] + (value.dim() - 1) * [1])
+        max_weight = weights.max(dim=dim)[0]
+        relative_probs = (weights - max_weight).exp()
+        return (value * relative_probs).sum(dim=dim) / relative_probs.sum(dim=dim)
 
     @property
     def mean(self):
--- a/examples/rsa/search_inference.py
+++ b/examples/rsa/search_inference.py
@@ -6,18 +6,18 @@ Adapted from: http://dippl.org/chapters/03-enumeration.html
 
 from __future__ import absolute_import, division, print_function
 
+import collections
+
+import six
 import torch
+from six.moves import queue
 
 import pyro
 import pyro.distributions as dist
 import pyro.poutine as poutine
-
 from pyro.infer.abstract_infer import TracePosterior
 from pyro.poutine.runtime import NonlocalExit
 
-import six
-from six.moves import queue
-import collections
 if six.PY3:
     import functools
 else:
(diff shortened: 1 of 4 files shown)


examples/rsa/search_inference.py
```diff
@@ -129,18 +129,19 @@ def _weighted_mean(self, value, dim=0):
         for _ in range(value.dim() - 1):
             weights = weights.unsqueeze(-1)
         max_val = weights.max(dim)[0]
```
fritzo: Hmm the original code looks suspicious to me, in that we're calling `.max(dim)` but then `unsqueeze(-1)`. Shouldn't we instead `.max(dim=dim, keepdim=True)`? (also eliminating common subexpressions:)
```py
max_val = weights.max(dim, keepdim=True)[0]
probs = (weights - max_val).exp()
return (value * probs).sum(dim=dim) / probs.sum(dim=dim)
```
(BTW thanks for this fix, and sorry to bog down review with cleanup that's not entirely related to your PR!)

tests/distributions/test_empirical.py
```diff
@@ -86,3 +86,10 @@ def test_weighted_mean_var(event_shape, dtype):
         with pytest.raises(ValueError):
             empirical_dist.mean
             empirical_dist.variance
+
```
neerajprad: You'll need another line here to get the lint test to pass (you could run `make lint` to see if there are any other linting issues).

adam-coogan: Thanks, oops, thought I'd done that.

examples/rsa/search_inference.py
adam-coogan: Good call on `probs = ...`!

Due to pytorch's broadcasting that doesn't work, and I don't think `keepdim` has an effect. The original `unsqueeze()` calls are used to add `value.dim()-1` singleton dimensions to `weights`, ensuring that it aligns with `value` along dimension 0. Maybe it'd be a bit cleaner to write:

```python
weights = weights.reshape([-1] + (value.dim() - 1) * [1])
```

?

neerajprad: I think the unsqueezes or reshapes are needed, but the remaining can be changed as per @fritzo's suggestion (which you already did, I see). 

> I am also curious whether there's a reason for _weighted_mean() to have the dim argument.

I think it was put in to support the possibility of multiple chains, with `dim=0` corresponds to the chain, but it will likely need other fixes to work correctly.

examples/rsa/search_inference.py
```diff
@@ -129,18 +129,19 @@ def _weighted_mean(self, value, dim=0):
...
-        return max_val.exp() * (value * (weights - max_val.unsqueeze(-1)).exp()).sum(dim=dim)
+        normalization = (value.new_tensor([1.]) * (weights - max_val.unsqueeze(-1)).exp()).sum(dim=dim)
```
neerajprad: Can we simplify this too as below?

examples/rsa/search_inference.py
fritzo: Thanks for explaining!

neerajprad:
@adam-coogan - Are you using `EmpiricalMarginal` directly, or via other algorithms like SVI, HMC? The reason I ask is because we might be making a few changes that will affect the interface (in #1557), but these are internal changes and shouldn't affect the public API.

adam-coogan:
I don't think so. Eventually I'll be implementing reversible jump MCMC. I've been trying to familiarize myself with pyro by different inference algorithms to a Gaussian mixture model and came across this issue.

examples/rsa/search_inference.py
adam-coogan: Ok, I've pushed a version using `reshape()`. `_weighted_mean()` will break if `dim != 0`, but I guess I'll leave it as is given @neerajprad's comment.

adam-coogan:
Yeah, happy to contribute, thanks for the reviews!

ahmadsalim:
This is great, I was just experiencing the bug myself before I saw the PR! Thanks!

PR #1559, opened 2018-11-21
Add example of a neural hmm

fritzo:

## Tested

- [x] added to tests/test_examples.py


examples/hmm.py
```diff
@@ -212,6 +213,70 @@ def model_4(sequences, lengths, args, batch_size=None, include_prior=True):
...
+        self.data_dim = data_dim
+        super(TonesGenerator, self).__init__()
+        self.x_to_hidden = nn.Linear(args.hidden_dim, args.nn_dim)
+        self.y_to_hidden = nn.Linear(args.nn_channels * data_dim, args.nn_dim)
+        self.conv = nn.Conv1d(1, args.nn_channels, 3, padding=1)
+        self.hidden_to_logits = nn.Linear(args.nn_dim, data_dim)
+        self.relu = nn.ReLU()
+
+    def forward(self, x, y):
+        if y.dim() < 2:
```
martinjankowiak: can we have some comments about one_hot, broadcasting in hidden state, etc.?

martinjankowiak: why `y.dim()` check?

examples/hmm.py
```diff
@@ -286,6 +351,8 @@ def main(args):
     parser.add_argument("-n", "--num-steps", default=50, type=int)
     parser.add_argument("-b", "--batch-size", default=8, type=int)
     parser.add_argument("-d", "--hidden-dim", default=16, type=int)
+    parser.add_argument("-nn", "--nn-dim", default=48, type=int)
```
martinjankowiak: please fix or remove `capacity` bit

fritzo: Hmm I believe this should still be a correct computation of capacity, i.e. the total number of parameters. Can you explain in person?

eb8680:
@fritzo @martinjankowiak is this good to merge?

Issue #1519, opened 2018-11-07
LBFGS does not work with SVI


fehiepsi:
LBFGS requires us define a `closure` to be called in its `step` method. Defining a closure to call it with `loss_and_grads` is easy but I can't figure out how to incorporate it with the way PyroOptim handles parameters (creating an optimizer for each param).

fehiepsi:
@eb8680 will we support it? LBFGS is a very convenient optimizer for models with small dataset! I have [used it to optimize a lot of models](https://github.com/fehiepsi/rethinking-pyro/blob/master/notebooks/rethinking.py#L49). Many of them can not be trained with SVI.

eb8680:
@fehiepsi sorry, I only closed the issue because I thought it was resolved by the PR #1518. 

`pyro.optim.PyroOptim` is a klugy wrapper for handling varying amounts or orders of parameters.  I don't see how we could modify it to take a closure argument, but I also think the way you're using `torch.optim` in your linked example is fine - if anything, we should be encouraging users to use that idiom instead of `svi.step`, since it's more explicit.

fehiepsi:
Please find a solution for this issue and discussions why such solution is not appropriate to merged in #1562.

PR #1562, opened 2018-11-23
Implement StaticSVI

fehiepsi:
This pull request implements StaticSVI, which is an SVI interface for model & guide which does not create new params dynamically. Hence LBFGS works with this inference (address #1519).

### DIFF
--- a/tests/infer/test_static_svi.py
+++ b/tests/infer/test_static_svi.py
@@ -1,0 +1,59 @@
+from __future__ import absolute_import, division, print_function
+
+import pytest
+import torch
+
+import pyro
+import pyro.distributions as dist
+import pyro.optim as optim
+from pyro.infer import StaticSVI, Trace_ELBO
+
+from tests.common import assert_equal
+
+
+def test_inference():
+    alpha0 = torch.tensor(1.0)
+    beta0 = torch.tensor(1.0)  # beta prior hyperparameter
+    data = torch.tensor([0.0, 1.0, 1.0, 1.0])
+    n_data = len(data)
+    data_sum = data.sum()
+    alpha_n = alpha0 + data_sum  # posterior alpha
+    beta_n = beta0 - data_sum + torch.tensor(float(n_data))  # posterior beta
+    log_alpha_n = torch.log(alpha_n)
+    log_beta_n = torch.log(beta_n)
+
+    def model():
+        p_latent = pyro.sample("p_latent", dist.Beta(alpha0, beta0))
+        with pyro.plate("data", n_data):
+            pyro.sample("obs", dist.Bernoulli(p_latent), obs=data)
+        return p_latent
+
+    def guide():
+        alpha_q_log = pyro.param("alpha_q_log", log_alpha_n + 0.17)
+        beta_q_log = pyro.param("beta_q_log", log_beta_n - 0.143)
+        alpha_q, beta_q = torch.exp(alpha_q_log), torch.exp(beta_q_log)
+        pyro.sample("p_latent", dist.Beta(alpha_q, beta_q))
+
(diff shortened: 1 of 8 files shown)


tests/infer/test_static_svi.py
```diff
@@ -0,0 +1,59 @@
...
+            pyro.sample("obs", dist.Bernoulli(p_latent), obs=data)
+        return p_latent
+
+    def guide():
+        alpha_q_log = pyro.param("alpha_q_log", log_alpha_n + 0.17)
+        beta_q_log = pyro.param("beta_q_log", log_beta_n - 0.143)
+        alpha_q, beta_q = torch.exp(alpha_q_log), torch.exp(beta_q_log)
+        pyro.sample("p_latent", dist.Beta(alpha_q, beta_q))
+
+    adam = optim.Adam({"lr": .001})
```
fritzo: Can you also add a test using `LBFGS`?

eb8680:
I don't really think this is necessary.  In this situation, we should encourage users to write their models as `nn.Module`s and use Pyro losses and `torch` optimizers together directly instead of going through the SVI and `PyroOptim` interfaces:
```python
class MyModel(nn.Module):
    def __init__(self, ...):
        ...

    def model(self, batch):
        ....

    def guide(self, batch):
        ...

model = MyModel(...)

elbo = pyro.infer.Trace_ELBO()
optim = torch.optim.SGD(model.parameters())  # or LBFGS etc
for batch in data:
    optim.zero_grad()
    elbo.loss_and_grads(model.model, model.guide, batch)
    optim.step()

fehiepsi:
@eb8680 Do you mean that we'll have a Pyro `nn.Module` which will catch all parameters from `model` and `guide` when we call `mymodel.parameters()`? Without that, users have to know about `poutine.trace(param_only=True)` to capture all params for pytorch optimizer. In addition, `SVI.run()` is quite convenient to get samples from posterior. Otherwise, users will have to learn about `poutine.trace(...)`, `poutine.replay(...)`,...

tests/infer/test_static_svi.py
fehiepsi: @fritzo I have added another test for it. Using LBFGS for this model is quite flaky.

eb8680:
@fehiepsi `module.parameters()` is a method of `torch.nn.Module` that PyTorch users would be familiar with.  Most of our tutorials, e.g. the VAE tutorial, already wrap the model and guide as methods of a single `nn.Module` and never call `pyro.param` explicitly in the model or guide, but they use `pyro.module`  (which calls `module.named_parameters()` under the hood) to pass the parameters to the Pyro parameter store.  I don't think that extra layer of indirection is buying us much, and I think the snippet above is less opaque and more consistent with PyTorch idioms.

We shouldn't try to fix leaky abstractions by introducing more indirection, we should just deprecate/remove them.  `SVI`, `Trace`, `ParamStoreDict`, and `PyroOptim` are some of the worst offenders.  On that note, re: `trace` and `replay` for parameter capture and posterior sampling instead of `SVI`, I actually think it'd be better to do that.  We can always provide a couple of tiny wrappers that remove boilerplate without obfuscation.

fehiepsi:
@eb8680 I took me a while to have a feeling that I understand what you mean. :) Did you mean that we define all the parameters ahead, so we don't need to use ParamStoreDict, PyroOptim? If that is the case, then I am in with the future deprecation of these wrappers (together with SVI of course). Please correct me if my understanding is incorrect.

About this PR, how about moving this to contrib for a while?

fehiepsi:
For that purpose, I vote to close this PR. I'll use that pattern in a GP tutorial instead. :)

PR #1564, opened 2018-11-24
Implement ubersum adjoint algorithms: map, sample, marginal

fritzo:
Pair coded with @eb8680 

This PR implements backward algorithms for `ubersum()` to accomplish a few tasks that were previously accomplished by more complex, much more expensive means.

Before this PR, `ubersum()` supported only the `LogRing` backend, which could propagate probabilities to a single target plate. This PR adds three new backends (at both the lower `einsum()` level, and the higher `ubersum()` level):
- **Marginal:** this can be used to compute marginals simultaneously at all sample sites, reducing the complexity of guide-to-model enumeration from quadratic to linear in the the tree size (e.g. hmm length).
- **MAP:** this can be used to compute a joint MAP estimate at all enumerated sample sites. This is needed for serving e.g. Viterbi estimates from CRF and HMM models.
- **Sample:** this can be used to compute joint samples at all enumerated sample sites. This is useful for e.g. querying TreeCat models.

Note that this PR only implements low-level `ubersum()` machinery. It is left for subsequent PRs to integrate this machinery into `TraceEnum_ELBO`, `CollapseMessenger`, etc. After these subsequent refactoring PRs, we should be able to delete much of our `shared_intermediates` logic, much of our caching code, and all of `pyro.ops.einsum.paths`.

## Design

These  backward (aka adjoint) algorithms are implemented very similar to PyTorch autograd adjoint algorithms:
1. we call `requires_backward(x)` on inputs `x` whose backward results we want, similar to PyTorch's `x.requires_grad_()`
2. we run a forward computation via `result = ubersum(...)`
3. we run a backward computation via `result._pyro_backward()`, similar to PyTorch's `.backward()`
4. we extract results from each input via an attribute `x._pyro_backward_result`, similar to PyTorch's `x.grad`

## Tested

- [x] einsum tests for Marginal
- [x] einsum tests for MAP
- [x] einsum tests for Sample
- [x] ubersum tests for Marginal
- [x] ubersum tests for MAP
- [x] ubersum tests for Sample

### DIFF
--- a/pyro/ops/contract.py
+++ b/pyro/ops/contract.py
@@ -379,7 +379,7 @@ def ubersum(equation, *operands, **kwargs):
     # Compute outputs, sharing intermediate computations.
     results = []
     with shared_intermediates(cache) as cache:
-        ring = Ring(cache)
+        ring = Ring(cache, dim_to_size=dim_to_size)
         for output in outputs:
             sum_dims = set(output).union(*inputs) - set(batch_dims)
             term = contract_to_tensor(tensor_tree, sum_dims,
--- a/pyro/ops/contract.py
+++ b/pyro/ops/contract.py
@@ -76,7 +76,8 @@ def _contract_component(ring, tensor_tree, sum_dims, target_dims):
     This function should be deterministic.
     This function has side-effects: it modifies ``tensor_tree``.
 
-    :param Ring ring: an algebraic ring defining tensor operations.
+    :param pyro.ops.rings.Ring ring: an algebraic ring defining tensor
+        operations.
     :param OrderedDict tensor_tree: a dictionary mapping ordinals to lists of
         tensors. An ordinal is a frozenset of ``CondIndepStack`` frames.
     :param set sum_dims: the complete set of sum-contractions dimensions
@@ -206,6 +207,8 @@ def contract_to_tensor(tensor_tree, sum_dims, target_ordinal=None, target_dims=N
         preserved in the result.
     :param dict cache: an optional :func:`~opt_einsum.shared_intermediates`
         cache.
+    :param pyro.ops.rings.Ring ring: an optional algebraic ring defining tensor
+        operations.
     :returns: a single tensor
     :rtype: torch.Tensor
     """
--- a/pyro/ops/contract.py
+++ b/pyro/ops/contract.py
@@ -158,6 +158,8 @@ class LogRing(Ring):
     Dims are characters (string or unicode).
     Ordinals are frozensets of characters.
     """
+    _backend = 'pyro.ops.einsum.torch_log'
+
     def __init__(self, cache=None, dim_to_size=None):
         super(LogRing, self).__init__(cache=cache)
         self._dim_to_size = {} if dim_to_size is None else dim_to_size
@@ -166,7 +168,7 @@ class LogRing(Ring):
         inputs = [term._pyro_dims for term in terms]
         output = ''.join(sorted(set(''.join(inputs)) - set(dims)))
         equation = ','.join(inputs) + '->' + output
-        term = contract(equation, *terms, backend='pyro.ops.einsum.torch_log')
+        term = contract(equation, *terms, backend=self._backend)
         term._pyro_dims = output
         return term
 
@@ -195,6 +197,29 @@ class LogRing(Ring):
         result._pyro_dims = term._pyro_dims
         self._cache[key] = result
         return result
+
+
+class MapRing(LogRing):
+    """
+    Ring of forward-maxsum backward-argmax operations.
+    """
+    _backend = 'pyro.ops.einsum.torch_map'
+    # TODO implement product backward
+
+
+class SampleRing(LogRing):
+    """
+    Ring of forward-sumproduct backward-sample operations in log space.
+    """
+    _backend = 'pyro.ops.einsum.torch_sample'
+    # TODO implement product backward
+
+
+_BACKEND_TO_RING = {
+    'pyro.ops.einsum.torch_log': LogRing,
+    'pyro.ops.einsum.torch_map': MapRing,
+    'pyro.ops.einsum.torch_sample': SampleRing,
+}
 
 
 def _partition_terms(ring, terms, dims):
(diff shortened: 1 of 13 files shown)


pyro/ops/rings.py
```diff
@@ -0,0 +1,242 @@
...
+from pyro.ops.einsum.adjoint import Backward
+
+
+def _finfo(tensor):
+    # This can be replaced with torch.finfo once it is available
+    # https://github.com/pytorch/pytorch/issues/10742
+    return np.finfo(torch.empty(torch.Size(), dtype=tensor.dtype, device="cpu").numpy().dtype)
+
+
+@add_metaclass(ABCMeta)
```
fritzo: This class was moved verbatim from contract.py

pyro/ops/rings.py
```diff
@@ -0,0 +1,242 @@
...
+        term_sum = self.sumproduct([term], dims)
+        global_part = self.product(term_sum, ordinal)
+        local_part = self.sumproduct([term, self.inv(term_sum)], set())
+        assert local_part._pyro_dims == term._pyro_dims
+        result = global_part, local_part
+        self._cache[key] = result
+        return result
+
+
+class LogRing(Ring):
```
fritzo: This class was moved verbatim from contract.py

pyro/ops/einsum/util.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    Creates a tensordot implementation from an einsum implementation.
+    """
+    def __init__(self, einsum):
+        self.einsum = einsum
+
+    # Copyright (c) 2014 Daniel Smith
+    # This function is copied and adapted from:
+    # https://github.com/dgasmith/opt_einsum/blob/a6dd686/opt_einsum/backends/torch.py
+    def __call__(self, x, y, axes=2):
```
fritzo: This method was moved nearly verbatim from torch_log.py; it has merely been wrapped in a class.

pyro/ops/einsum/torch_map.py
```diff
@@ -0,0 +1,54 @@
...
+
+import operator
+
+from six.moves import reduce
+
+from pyro.ops import packed
+from pyro.ops.einsum.adjoint import Backward, einsum_backward_sample, transpose, unflatten
+from pyro.ops.einsum.util import Tensordot
+
+assert transpose  # pacify flake8
```
neerajprad: nit: We have used `# noqa F403` to pacify flake8's import not used error in other modules. Can we use the same idiom?

pyro/ops/einsum/adjoint.py
```diff
@@ -0,0 +1,142 @@
...
+    if hasattr(a, '_pyro_backward'):
+        result._pyro_backward = _TransposeBackward(a, axes)
+        result._pyro_name = getattr(a, '_pyro_name', '?') + "'"
+    return result
+
+
+def einsum_backward_sample(operands, sample1, sample2):
+    """
+    Cuts down samples to pass on to subsequent steps.
+    This is used various in ``_EinsumBackward.__call__()`` methods.
```
neerajprad: typo - `in various`

pyro/ops/einsum/torch_marginal.py
```diff
@@ -0,0 +1,53 @@
...
+    result = pyro.ops.einsum.torch_log.einsum(equation, *operands)
+
+    if any(hasattr(x, '_pyro_backward') for x in operands):
+        result._pyro_backward = _EinsumBackward(equation, operands)
+    return result
+
+
+tensordot = Tensordot(einsum)
+
+__all__ = ["transpose", "einsum", "tensordot"]
```
neerajprad: I think `flake8` should recognize this usage pattern. What version are you using?

pyro/ops/einsum/adjoint.py
```diff
@@ -0,0 +1,142 @@
...
+        target = self.target()
+        target._pyro_backward_result = message
+        return ()
+
+
+def require_backward(tensor):
+    """
+    Marks a tensor as a leaf in the adjoint graph.
+    """
+    tensor._pyro_backward = _LeafBackward(tensor)
```
neerajprad: Just for my understanding - is the backward pass repeated when we call `_pyro_backward()` twice? Would it ever be beneficial to cache it?

pyro/ops/contract.py
```diff
@@ -355,7 +186,7 @@ def contract_tensor_tree(tensor_tree, sum_dims, cache=None):
 
 
 def contract_to_tensor(tensor_tree, sum_dims, target_ordinal=None, target_dims=None,
-                       cache=None, dim_to_size=None):
+                       cache=None, ring=None):
```
neerajprad: We should update the docstring with the `ring` kwarg.

pyro/ops/einsum/torch_map.py
fritzo: I dislike the noqa F403 for two reasons:
1. it is ambiguous in that only one of the imports is unused here (`transpose`), and `noqa` allows other extraneous imports on the same `import` line
2. on long import lines that are wrapped, the `#noqa f403` can end up appearing on only one line, indeed in this PR appeared on a line different from the `transpose` it was intended to block

But I'll concede if you feel strongly that `f403` is a better solution 😄

pyro/ops/einsum/adjoint.py
fritzo: Yes it should be repeated; the caching should happen at the `opt_einsum.shared_intermediates` layer. I haven't verified that this works yet, since I haven't needed to rely on it.

pyro/ops/einsum/torch_map.py
fritzo: Nevermind, it looks like `__all__ = ...` works fine. Thanks for suggesting!

neerajprad: I don't feel strongly at all, and your reasons are all valid, but it will be nice to maintain consistency whichever way it is. Also, you might not even need it (I think flake8 picks up `__all__` usage e.g. in `distributions.__init__`).

pyro/ops/einsum/adjoint.py
neerajprad: Got it, thanks for explaining!

PR #1571, opened 2018-11-27
Fix build timeout due to dataset download issues

neerajprad:
The [build](https://travis-ci.org/uber/pyro/jobs/460036655) on the `pytorch-1.0` branch seems to be timing out very frequently in `test_examples`. It seems that the issue is simply that the download for the MNIST dataset is timing out or is being throttled. 

I have uploaded the dataset to our S3 bucket instead, and patched `torchvision.datasets.MNIST` to use these URLs. I have moved the very useful `patch` decorator to `utils` (which @fritzo might not approve of). Will merge this into the pytorch-1.0 branch if the change is approved.


pyro/contrib/examples/util.py
```diff
@@ -2,9 +2,23 @@
...
 import torchvision.datasets as datasets
 from torch.utils.data import DataLoader
 from torchvision import transforms
+from torchvision.datasets import MNIST
+
+from pyro.util import patch_dependency
+
+
+@patch_dependency('torchvision.datasets.MNIST', torchvision)
+class _MNIST(MNIST):
```
fritzo: Just to be safe in the presence of Jupyter notebook autoreloads , I'd recommend:
```py
class _MNIST(getattr(MNIST, '_pyro_unpatched', MNIST)):
```

pyro/util.py
```diff
@@ -342,3 +342,22 @@ def deep_getattr(obj, name):
     Throws an AttributeError if bad attribute
     """
     return functools.reduce(getattr, name.split("."), obj)
+
+
+def patch_dependency(target, root_module=torch):
```
fritzo: This introduces a cyclic dependency, since pyro.poutine depends on pyro.distributions depends on pyro.util depends on pyro.poutine. Could we keep this in pyro.distributions (which should not depend on anything outside of distributions), say in pyro.distributions.util or a new module pyro.distributions.patch?

neerajprad:
Thanks for reviewing! This can be merged into pytorch-1.0 branch.

pyro/util.py
fritzo: Thanks. I'm not sure how much of an issue it actually is, but I'd like to avoid any potential issues combining cyclic dependencies with hot patching :smile:

Issue #269, opened 2017-10-16
lift to memoize samples


jpchen:
taken from #266 :
lift doesn't quite work properly when param is called with the same name twice in a stochastic function. This is not allowed for sample but is currently fine to do with param, since the second call just retrieves the same value from the parameter store.

eb8680:
@jpchen this seems like an easy fix, do you want to get this into 0.3?

jpchen:
yeah i can do this, it's not a release blocker though.

PR #1576, opened 2018-11-28
memoize sample statements in lift

jpchen:
Addresses #269

Essentially this PR memoizes repeated `param` statements when wrapped by `lift`.  Added a simple sanity test.


pyro/poutine/lift_messenger.py
```diff
@@ -28,6 +28,7 @@ def __init__(self, prior):
         """
         super(LiftMessenger, self).__init__()
         self.prior = prior
+        self._samples_cache = {}
```
eb8680: You should also clear `self._samples_cache` in `__enter__` and `__exit__`

Issue #1554, opened 2018-11-20
Create mini-version of Pyro for didactic purposes


fritzo:
The Architecture of Open Source Applications project hosts tiny educational versions of many pieces of software https://github.com/aosabook/500lines (recommended reading!). I think it would be helpful to have a similar examples/mini_pyro.py to demonstrate how e.g. how Pyro's effects library works.

## Tasks

- [x] #1553 `trace`, `replay`, `SVI`, `elbo`, and a simple example
- [x] #1578 rendered version on pyro.ai
- [x] #1583 comments throughout the file
- [x] #1589 `plate` with broadcasting
- [ ] #1599 `block` (@eb8680)

---- triaged ----
- ~~enumeration~~
- ~~`markov`~~

## Budget

All of pyro plus comments plus an example should fit into:
- 500 lines of code
- ~~1 single python file~~ 2 python files = 1 for example + 1 for library

PR #1578, opened 2018-11-28
Split minipyro into contrib + example

fritzo:
Addresses #1554 

This splits `examples/mini_pyro.py` into `examples/minipyro.py` + `pyro/contrib/minipyro`.

Note this PR does not add any docstrings; that is deferred to a follow-up PR.

### Pros:
- The contrib/minipyro.py package will now show up on docs.pyro.ai
- Users can `import pyro.contrib.minipyro as pyro` to try out the minimal version (e.g. for debugging)
- The example is now tested on both mini and full pyro :smile:

### Cons:
- we no longer locate the example and implementation in the same file.
- it is kind of weird to distribution minipyro with pyro.

## Questions for reviewers

- [ ] Does this seem reasonable?

## Tested

- [x] added test to tests/test_examples.py


tutorial/source/mini_pyro.py
```diff
@@ -0,0 +1 @@
+../../examples/mini_pyro.py
```
jpchen: does this build with `make html`?

fritzo: oops no, I accidentally left a stray simlink; will delete...

fritzo:
ready to merge

PR #1582, opened 2018-11-29
[Do not merge; Testing] Reintroduce reshaped dist to check perf regression

neerajprad:
Refer to the discussion in #1544. This may not be needed for all distributions but I want to get our CI runtimes with this change. cc. @fehiepsi

### DIFF
--- a/pyro/distributions/torch_distribution.py
+++ b/pyro/distributions/torch_distribution.py
@@ -1,12 +1,15 @@
 from __future__ import absolute_import, division, print_function
 
+import numbers
+
 import torch
 from torch.distributions import biject_to, constraints, transform_to
 from torch.distributions.kl import kl_divergence, register_kl
 
 import pyro.distributions.torch
 from pyro.distributions.distribution import Distribution
-from pyro.distributions.util import broadcast_shape, scale_and_mask
+from pyro.distributions.score_parts import ScoreParts
+from pyro.distributions.util import broadcast_shape, scale_and_mask, sum_rightmost
 
 
 class TorchDistributionMixin(Distribution):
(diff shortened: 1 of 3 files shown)


pyro/distributions/torch_distribution.py
```diff
@@ -221,6 +254,142 @@ def check(self, value):
...
+
+    def sample(self, sample_shape=torch.Size()):
+        return self.base_dist.sample(sample_shape + self.sample_shape)
+
+    def rsample(self, sample_shape=torch.Size()):
+        return self.base_dist.rsample(sample_shape + self.sample_shape)
+
+    def log_prob(self, value):
+        shape = broadcast_shape(self.batch_shape, value.shape[:value.dim() - self.event_dim])
+        return sum_rightmost(self.base_dist.log_prob(value), self.reinterpreted_batch_ndims).expand(shape)
```
fehiepsi: This should be the way to do! Expand parameters first, then computing log_prob might be ineffective if log_prob contains expensive operators.

cc @fritzo for a throughout discussion about this problem

fehiepsi:
I know the bug. Will make a PR at pytorch soon.

neerajprad:
> I know the bug. Will make a PR at pytorch soon.

The PR will likely not make it into release. So could you patch the mvn distribution in `torch.py` in Pyro as well?

neerajprad:
> Sounds great, thanks @fehiepsi.

pyro/distributions/torch_distribution.py
neerajprad: At a primary glance, it doesn't seem like we should be introducing the ReshapedDistribution back, but instead patch the mvn and lowrank mvn distributions, as you have done below. I'll gather a few test statistics and close this PR.

neerajprad:
Closing this, as #1588 resolves the slowdown with a change localized to mvn rather than reintroducing `ReshapedDistribution`.

Issue #1584, opened 2018-11-29
GMM tutorial broken in current dev build


kgero:
### Environment
Mac OS, Python 3.6.1. Originally installed pyro with pip. 

torch=0.4.1 pyro=0.2.1

### Issue 
I was having the same issues as areshytko (see on the forum [here](https://forum.pyro.ai/t/gmm-example-in-documentation-gives-warning-about-from-traceenum-elbo/394/5)) when trying to run the GMM tutorial (http://pyro.ai/examples/gmm.html). 

```
UserWarning: TraceEnum_ELBO found no sample sites configured for enumeration. If you want to enumerate sites, you need to @config_enumerate or set infer={"enumerate": "sequential"} or infer={"enumerate": "parallel"}? If you do not want to enumerate, consider using Trace_ELBO instead.
warnings.warn('TraceEnum_ELBO found no sample sites configured for enumeration. '
```

While it would run, the results were bad -- no clear convergence, and couldn't correctly find the two clusters (put everything into one cluster).

Since it seems like it was fixed recently, I installed pyro from the dev branch. However, then the code would not run at all. (Apologies for not including the actual error; I tried to revert but then it was an issue with using torch==0.4.1 and not torch==0.4.0 (couldn't find `logsumexp`) and I couldn't figure out exactly.

I fixed this issue by reverting to commit `4c89c99e1026a4d87e53b2f2e813b41b78f6a439` (which a friend of mine was on, and she had successfully run the tutorial). Also seems like commit `43155bf967921f9ddccd2998d97186f0c2ff2f7a` works fine (for the tutorial) but the next one doesn't (with the update to pytorch==1.0.0).

neerajprad:
The `dev` branch in Pyro is currently tracking the PyTorch pre-release to prepare us for a release next week. I think the example should run if you were to install pytorch [pre-release](https://pytorch.org/), but let me know if you face any issues.

neerajprad:
My recommendation would be to use the `pytorch-0.4.0` branch (which has all recent changes) which is compatible with pytorch version 0.4.0, for another week or so until release, so that you don't have to deal with constant flux on both PyTorch and Pyro as they prepare for release!

neerajprad:
Closing this, please feel free to reopen if you continue to see issues.

Issue #1566, opened 2018-11-24
 'dmm.polyphonic_data_loader' has no attribute 'load_data'


dongfeng951:
When I run hmm.py example, there is an error. 
AttributeError: module 'dmm.polyphonic_data_loader' has no attribute 'load_data'
Then I check the polyphonic_data_loader.py, and I cannot find load_data(). Is there anything missing?
Thank you!

neerajprad:
Please use the `dev` branch of Pyro to run the `hmm` example. You should be able to see `load_data` in `poloyphonic_data_loader` on the dev branch. Note that this will be part of the next release in a few weeks.

jpchen:
i like (3). i was actually thinking of adding a README to examples as a short-term remedy, since most people discover them through the github webpage.

dongfeng951:
Hi, Thank you very much!

neerajprad:
I think (3) is best (i.e. each example has an associated minimum version), since examples might build on functionality that require the `dev` branch and it might not be feasible to port those changes to master always.

Issue #1563, opened 2018-11-23
module 'pyro' has no attribute 'plate'


dongfeng951:
When I run lda.py in the example directory, there is an error.

Traceback (most recent call last):
    File "lda.py", line 137, in <module>
        main(args)
    File "lda.py", line 105, in main
        true_topic_weights, true_topic_words, data = model(args=args)
    File "lda.py", line 35, in model
        with pyro.plate("topics", args.num_topics):
AttributeError: module 'pyro' has no attribute 'plate'

What is the problem?
Thanks.

fehiepsi:
I think you need to use Pyro dev: `pip install git+https://github.com/uber/pyro`.

jpchen:
all examples need to be run with the `dev` version of pyro

dongfeng951:
Thanks! I will try it.

beckybai:
Does pyro 0.3 needs pytorch 1.0?

neerajprad:
> Does pyro 0.3 need pytorch 1.0? 

That's right. We are waiting for pytorch 1.0 release before releasing pyro 0.3, but you should be able to use the pytorch pre-release version with pyro 0.3.

Issue #1552, opened 2018-11-20
hmm.py example not working


tongjiyiming:
Just want to run the hmm.py example, and got the following error:
```
      691 Loading data
      722 ----------------------------------------
      722 Training model_1 on 229 sequences
Traceback (most recent call last):
  File "hmm.py", line 295, in <module>
    main(args)
  File "hmm.py", line 244, in main
    guide = AutoDelta(poutine.block(model, expose_fn=lambda msg: msg["name"].startswith("probs_")))
TypeError: block() got an unexpected keyword argument 'expose_fn'
```
look like the function argument is been updated. Then, I changed the "expose_fn" to "expose" as I found in Pyro docs. It still breaks:
```
      682 Loading data
      711 ----------------------------------------
      711 Training model_1 on 229 sequences
Traceback (most recent call last):
  File "hmm.py", line 295, in <module>
    main(args)
  File "hmm.py", line 245, in main
    guide = AutoDelta(poutine.block(model, expose=lambda msg: msg["name"].startswith("probs_")))
  File "/opt/conda/lib/python3.6/site-packages/pyro/poutine/handlers.py", line 209, in block
    hide_types=hide_types, expose_types=expose_types)
  File "/opt/conda/lib/python3.6/site-packages/pyro/poutine/block_messenger.py", line 79, in __init__
    assert set(hide).isdisjoint(set(expose)), \
TypeError: 'function' object is not iterable
```
anyone can check what is the problem?

fritzo:
Hi @tongjiyiming can you verify that your Pyro version matches the example version? I suspect you're using an older Pyro with a newer examples/hmm.py. The example currently works on dev branch.

tongjiyiming:
I got "0.2.1".
So, should I install the dev version?

fritzo:
Yes, the example only works with dev branch for now. We're planning another release in a few weeks, just after PyTorch 1.0 is released.

tongjiyiming:
unfortunately, after I upgrade to dev branch by
```
pip install git+https://github.com/uber/pyro.git#egg=project[extras]
```
It still failed with:
```

tongjiyiming:
Thank you! It works now.

PR #1587, opened 2018-11-29
Bump Pyro version to 0.3.0; enable version checks

fritzo:
Replaces #1568 
Addresses #1584, #1566, #1563, #1552, #1523, ...

- Bump Pyro version to 0.3.0
- `assert torch.__version__.startswith('1.')` in torch_patch.py
- `assert pyro.__version__.startswith('0.3.0')` in all examples and tutorials


examples/air/main.py
```diff
@@ -246,6 +246,7 @@ def per_param_optim_args(module_name, param_name):
 
 
 if __name__ == '__main__':
+    assert pyro.__version__.startswith('0.3.0')
```
jpchen: '0.3' probably works since patches shouldnt break examples

fritzo: While `starswith('0.3')` would work while `pyro.__version__ == 0.3.0`, we'll eventually need `startswith('0.3.1')` when we bump to `pyro.__version__ == 0.3.1`. It seems simpler to pin the whole version.

Issue #1586, opened 2018-11-29
Perf bug in multivariate normal due to inefficient .expand


neerajprad:
Ref. https://github.com/uber/pyro/pull/1582. 

There is a perf bug in MVN, and lowrank MVN `log_prob` methods on expanded instances (i.e. instances coming from `mvn.expand()`). @fehiepsi already has a patch for it. I am running our test suite to see if this might also affect other distributions.

fehiepsi:
#1588 should fix some slowness. Another thing related to mvn (but I guess it also happens for other distributions too) which is not relevant to this bug, but I want to put here for record:
- With event_shape < 50, pytorch 0.4 is faster than pytorch 1.0
- With event_shape > 50, pytorch 1.0 is starting outperforming pytorch 0.4

neerajprad:
> With event_shape < 50, pytorch 0.4 is faster than pytorch 1.0

The pytorch devs did mention that there is some tensor overhead that was added after `0.4` due to which small models are likely to be slower. https://github.com/pytorch/pytorch/pull/9320 was an attempt to rectify that.

fehiepsi:
Following your link, I can see there are many interesting stories about the performance of 0.4 vs 1.0. Thanks @neerajprad !

fehiepsi:
This is already fixed.

Issue #1540, opened 2018-11-15
sv-dkl example timing out with pytorch-1.0


neerajprad:
I am seeing an increasing number of builds ([example](https://travis-ci.org/uber/pyro/jobs/455675436)) being aborted in `test_examples` with the error that `No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.` on the `pytorch-1.0` branch. This seems to mostly get triggered in `python examples/contrib/gp/sv-dkl.py`. 

The problem usually resolves if we rebuild, but it is worth investigating as we are not seeing this on the `dev` branch. If the issue is due to #1421, we should make this test example faster so as not to stall the build.

```
python examples/contrib/gp/sv-dkl.py --epochs=1 --num-inducing=4
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

fritzo:
I've seen a couple stalled builds on dev branch recently, so this may merely be a Travis issue. I'll grab details next time I see this.

neerajprad:
I think the issue is most likely that the torchvision dataloader is deadlocking in pytorch-1.0 for some reason. I'm still investigating it.

fehiepsi:
I have tried various assumptions in #1612 without successful. Here are my observations:

+ This seems like a DataLoader problem: I disable `forward` and `backward`, failed travis can be seen  at https://travis-ci.org/uber/pyro/jobs/465717303.
+ Because `vae` example also requires MNIST dataloader but it didn't fail in Travis. I thought that the torchvision Normalize transform might be the problem. However, the problem is still happened when I removed it. :(
+ The loss in failed run is different from the loss in passed run by a small margin. Loss is consistent in failed run and also consistent in passed run.

neerajprad:
> The loss in failed run is different from the loss in passed run by a small margin. Loss is consistent in failed run and also consistent in passed run.

Maybe we can try `shuffle=False` and see if that at least gives the same loss.

fehiepsi:
It seems that there is a regression in `conditional` function. In the commit which I mentioned above (about DataLoader problem), I forget to disable `forward` pass of `test` function. So DataLoader might not be a problem. I'm investigating more.

fehiepsi:
Finally, I can isolate the slowness: `nn.Conv2d`.
[...]

PR #1588, opened 2018-11-29
temporary fix expanded mvn

fehiepsi:
Temporary fix for #1586 to make CI faster.

### DIFF
--- a/pyro/distributions/torch.py
+++ b/pyro/distributions/torch.py
@@ -1,14 +1,86 @@
 from __future__ import absolute_import, division, print_function
 
+import math
+
 import torch
 from torch.distributions import constraints, kl_divergence, register_kl
+from torch.distributions.multivariate_normal import _batch_diag, _batch_mv
+from torch.distributions.lowrank_multivariate_normal import (_batch_lowrank_mahalanobis,
+                                                              _batch_lowrank_logdet)
+from torch.distributions.utils import _standard_normal
 
 from pyro.distributions.torch_distribution import IndependentConstraint, TorchDistributionMixin
 from pyro.distributions.util import sum_rightmost
 
 
+class LowRankMultivariateNormal(torch.distributions.LowRankMultivariateNormal,
+                                TorchDistributionMixin):
+    # Resolve the issue https://github.com/uber/pyro/issues/1586
+    def __init__(self, loc, cov_factor, cov_diag, validate_args=None):
+        super(LowRankMultivariateNormal, self).__init__(loc, cov_factor, cov_diag, validate_args)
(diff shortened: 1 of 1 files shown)


pyro/distributions/torch.py
```diff
@@ -1,15 +1,87 @@
 from __future__ import absolute_import, division, print_function
 
+import math
+
 import torch
 from torch.distributions import constraints, kl_divergence, register_kl
+from torch.distributions.multivariate_normal import _batch_mv
+from torch.distributions.lowrank_multivariate_normal import (_batch_lowrank_mahalanobis,
+                                                             _batch_lowrank_logdet)
+from torch.distributions.utils import _standard_normal
```
neerajprad: Should we also override the kl implementations?

neerajprad:
Also, could you remove the `skipif` parameter from `test_examples` for GP. I had to add that due to #1540, and I believe that the cause was most likely this slowdown.

pyro/distributions/torch.py
fehiepsi: Because we just introduce KL for SVI recently, it is unlikely to have some tests with kl affected by this issue. If the fix in pytorch is not able to be in 1.0, I agree with you that we should override KL too.

neerajprad:
There are some failures in `test_autoguide`, but the unit test time here seems to match what we observed in #1582 that reintroduces ReshapedDistribution, so I'll close that other PR.

neerajprad:
@fehiepsi - I am not sure why the example slowdown is only happening for python 2.7, but it is great to see that the slowdown issues are resolved for python 3. If you think it is unrelated to this PR, feel free to reintroduce the skip marker.

fehiepsi:
@neerajprad I tested with python 2.7 and didnt observe. I don't know too. :(

neerajprad:
The only thing I can imagine is that it is loading the torch distribution class instead of the patched class. One thing you could try to debug would be to put in a call attribute `_pyro_patched=True` in your patched class, and `assert dist.MultivariateNormal._pyro_patched` somewhere in `gpr.py`. At least that will help in narrowing down the problem.

neerajprad:
@fehiepsi - Can I push to your branch to help debug?

fehiepsi:
I think it is better to close this. Gaining time is negligible to debugging time. Eventually we will delete this fix in a few days. :(

fehiepsi:
Sure, I just add you to my fork. Please don't worry much about this issue.

neerajprad:
Thanks @fehiepsi, I'll take a look. I am not too concerned about the few mins we are saving on unit testing, but I am not sure if your fix will make it into release. :( In that case, we will need this patch for sure.

PR #1591, opened 2018-11-30
add fix for torch 1.0 on RTD

jpchen:


pyro/distributions/torch_patch.py
```diff
@@ -2,7 +2,9 @@
 
 import torch
 
-assert torch.__version__.startswith('1.')
+if 'READTHEDOCS' not in os.environ:
```
fritzo: import os

jpchen: thanks. you can see the the thorough testing this went through

PR #1597, opened 2018-11-30
Modify custom objectives tutorial

jpchen:
- adds ToC
- adds multi-optimizer example


tutorial/source/custom_objectives.ipynb
```diff
@@ -115,7 +133,23 @@
     "sgd.zero_grad()\n",
     "```\n",
     "\n",
-    "### Example: Custom ELBO\n",
+    "Pyro also provides this as a [MultiOptimizer](http://docs.pyro.ai/en/dev/optimization.html?highlight=multi%20optimizer#module-pyro.optim.multi) which wraps Pyro optimizers.\n",
```
martinjankowiak: what's "this" ?

martinjankowiak: do we really want this multi-optimizer stuff? kind of orthogonal?

jpchen: i mean it's an aside to that particular example, since it so happens that that functionality is built-in. i dont think it detracts from the rest, and it's a nice place to showcase an otherwise obscure feature. wdyt?

PR #1599, opened 2018-12-01
Add block to minipyro and a few comments

### LINKED ISSUE #1554: Create mini-version of Pyro for didactic purposes

eb8680:
Resolves #1554 

This PR adds `block` to `contrib.minipyro`, and adds a few more descriptive comments.  I also reversed the order of the actual `PYRO_STACK` to be consistent with the `minipyro` one.

### DIFF
--- a/pyro/contrib/minipyro.py
+++ b/pyro/contrib/minipyro.py
@@ -96,7 +96,7 @@ class replay(Messenger):
 # Sites hidden by block will only have the handlers below block on the PYRO_STACK applied,
 # allowing inference or other effectful computations to be nested inside models.
 class block(Messenger):
-    def __init__(self, fn, hide_fn):
+    def __init__(self, fn=None, hide_fn=lambda msg: True):
         self.hide_fn = hide_fn
         super(block, self).__init__(fn)
 
@@ -144,6 +144,7 @@ def sample(name, fn, obs=None):
     return msg["value"]
 
 
+# param is an effectful version of PARAM_STORE.setdefault
 def param(name, init_value=None):
 
     def fn(init_value):
@@ -204,7 +205,7 @@ class SVI(object):
         # further tracing occurs inside of `loss`.
         with trace() as param_capture:
             # We use block here to allow tracing to record parameters only.
-            with block(fn=None, hide_fn=lambda msg: msg["type"] == "sample"):
+            with block(hide_fn=lambda msg: msg["type"] == "sample"):
                 loss = self.loss(self.model, self.guide, *args, **kwargs)
         # Differentiate the loss.
         loss.backward()
--- a/pyro/contrib/minipyro.py
+++ b/pyro/contrib/minipyro.py
@@ -18,7 +18,7 @@ from collections import OrderedDict
 import torch
 
 # Pyro keeps track of two kinds of global state:
-# i)  The Poutine stack, which enables non-standard interpretations of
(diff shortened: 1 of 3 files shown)


pyro/contrib/minipyro.py
```diff
@@ -78,6 +92,19 @@ def process_message(self, msg):
             msg["value"] = self.guide_trace[msg["name"]]["value"]
 
 
+# block allows the selective application of effect handlers to different parts of a model.
+# Sites hidden by block will only have the handlers below block on the PYRO_STACK applied,
+# allowing inference or other effectful computations to be nested inside models.
+class block(Messenger):
+    def __init__(self, fn, hide_fn):
```
fritzo: How about defaulting `fn=None, hide_fn=lambda msg: True`? Then we can use the syntax
```py
with block():
    ...do stuff...
```
Plus it doesn't come out of our 500-line budget :wink:

pyro/contrib/minipyro.py
```diff
@@ -32,16 +32,19 @@ def get_param_store():
...
+# The base effect handler class (called Messenger here for consistency with Pyro).
 class Messenger(object):
     def __init__(self, fn=None):
         self.fn = fn
 
+    # Effect handlers push themselves onto the PYRO_STACK.
+    # Handlers earlier in the PYRO_STACK are applied first.
     def __enter__(self):
-        PYRO_STACK.append(self)
+        PYRO_STACK.insert(0, self)
```
fritzo: IMO it's really gross to use Python's native stack datatype in reverse order of the intended use (as indicated by the `.pop()` method). This use is also slower due to linear-time cost of popping from the front, vs constant time cost of popping from the back. I think we should instead update full pyro to match this file.

eb8680: Sure, I just wanted `minipyro` to be consistent with actual Pyro.

pyro/contrib/minipyro.py
```diff
@@ -98,31 +125,36 @@ def __iter__(self):
...
+def apply_stack(msg):
+    for pointer, handler in enumerate(reversed(PYRO_STACK)):
         handler.process_message(msg)
+        if msg.get("stop"):
+            break
     if msg["value"] is None:
-        msg["value"] = fn()
-    for handler in PYRO_STACK:
+        msg["value"] = msg["fn"](*msg["args"])
+    for handler in PYRO_STACK[-pointer-1:]:
```
fritzo: note: I think you could have used `reversed(enumerate(PYRO_STACK))` above, and then this would be merely `PYRO_STACK[pointer:]`?

Issue #1406, opened 2018-09-24
Misc cleanup for 0.3 release


fritzo:
- [x] replace `.new_tensor()` with `.detach().clone()` where possible
- [x] replace `.independent()` with `.any_less_confusing_name()` (@jpchen)
- [x] replace `.empty(...).fill_(value)` with `torch.full(shape, value)` or `torch.full_like(x, value)`
- [x] replace `len(tensor.shape)` with `tensor.dim()`
- [x] use `plate(subsample=)` in examples to get correctly scaled ELBO loss (see #1437)
- [x] rename various D_term, W_term while using LowRankMVN
- [x] add docs for `pyro.ops`
- [x] check versioning for examples
- others?

---- triaged ----
- (optional?) remove remaining uses of `unittest.TestCase`?
- replace `tensor.shape[n]` with `tensor.size(n)` (which is a tiny bit cheaper)
- ~~rename `Trace_ELBO` and friends?~~

jpchen:
closing as most have been addressed, the rest are low priority, too much effort, or not targeted for this release.

PR #1602, opened 2018-12-01
independent -> to_event

jpchen:
Addressing #1406

based on discussions, `to_event` seems like the most concise, clearest, not-incorrect name.  `reinterpreted_batch_dims` is what it's called in the `Independent` distribution but that's simply not practical for writing models where one would need to append it most of their distributions.


pyro/distributions/torch_distribution.py
```diff
@@ -118,6 +118,10 @@ def independent(self, reinterpreted_batch_ndims=None):
             reinterpreted_batch_ndims = len(self.batch_shape)
         return pyro.distributions.torch.Independent(self, reinterpreted_batch_ndims)
 
+    def independent(self, reinterpreted_batch_ndims=None):
```
fritzo: 💯

tests/distributions/test_mask.py
```diff
@@ -70,9 +70,9 @@ def test_kl_divergence():
     mask = torch.tensor([[0, 1], [1, 1]]).byte()
     p = Normal(torch.randn(2, 2), torch.randn(2, 2).exp())
     q = Normal(torch.randn(2, 2), torch.randn(2, 2).exp())
-    expected = kl_divergence(p.independent(2), q.independent(2))
-    actual = (kl_divergence(p.mask(mask).independent(2),
-                            q.mask(mask).independent(2)) +
-              kl_divergence(p.mask(~mask).independent(2),
-                            q.mask(~mask).independent(2)))
+    expected = kl_divergence(p.to_event(2), q.independent(2))
```
fritzo: Missed one.

jpchen:
> .add_event_dims(x)

i think @fritzo's suggested `shift_event_dims` would be more correct because no additional dims are being added. and i dropped the `dims` cause of verbosity/its contextually evident

PR #1605, opened 2018-12-02
Add Parameterized class to `pyro.contrib.gp`

fehiepsi:
This PR adds the Parameterized class to `pyro.contrib.gp`. The good points of this class are demonstrated in its docs. Basically, it is a nn.Module which can set constraints and priors to its parameters. I believe that there are many things to be improved, but I hope this can be served as a basic one. Note that this does not rely on any `gp` model/algorithm, but in a future PR, all gp models will depend on this class.

Please feel free to suggest a better place for this class. It origins from GP module, so I can move it there if `pyro.nn` is not a good place for it. **Edit:** I just saw that this depends on `contrib.autoname`, so moving it to `contrib` is a better place.

### DIFF
--- a/pyro/contrib/gp/parameterized.py
+++ b/pyro/contrib/gp/parameterized.py
@@ -14,8 +14,8 @@ from pyro.distributions.util import eye_like
 
 class Parameterized(nn.Module):
     """
-    A wrapper of :class:`nn.Module` whose parameters can be set constraints, set
-    priors.
+    A wrapper of :class:`torch.nn.Module` whose parameters can be set
+    constraints, set priors.
 
     Under the hood, we move parameters to a buffer store and create "root"
     parameters which are used to generate that parameter's value. For example,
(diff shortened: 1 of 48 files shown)


fehiepsi:
@fritzo Beside adding `autoguide`, this simplifies `gp`, not adding complexity!

About GPyTorch, its models work with PyTorch parameters, not ParamStoreDict. In that aspect, this version of Reparameterized class is closer to GPyTorch than previously. To use Pyro SVI there, we need to cast all `nn.Parameter` to ParamStoreDict (by calling `pyro.module`), create a model and guide with `pyro.sample(...)` statements in them (to trigger KL loss). They already [implemented it](https://github.com/cornellius-gp/gpytorch/blob/master/examples/09_Pyro_Integration/Pyro_Simple_GP_Classification.ipynb) as far as I can see.

About migrating to GPyTorch, it is a "far" plan. We might not eventually migrate to GPyTorch if it is unnecessary and they don't like the interface of `contrib.gp`. To me, Pyro GP is closer to what are presented in GP textbooks and most GP papers which I have read (except SKI related ones). So GPyTorch devs can easily grasp the stuffs which they think are useful from Pyro GP.

If we need performance, we might implement a "fast" multivariate normal distribution (?!) which uses [BlackBoxMatrixMatrix](https://arxiv.org/abs/1809.11165) to do matrix_solve and log_det. I am not sure if that will be enough to achieve a better performance as presented in GPyTorch paper. cc @jacobrgardner

Btw, I am not good at selling things so I can't say much more about the flexibility of `contrib.gp`. What I focus is to bring the convenience for users who like to play around GP stuffs (like making deep GP, deep kernel, gplvm, trained with HMC,...), not for "power" users who already understands all technical points and requires a fast version for production. Even that my focus might be deviated from what Pyro devs want, I believe that this PR will make the transition (if any) to GPyTorch easier than before (in terms of no more depending on ParamStoreDict, which requires `self.get_param("foo")` every time we need to get value of a parameter).

fehiepsi:
I think that it is better to make a separate package, which relies on Pyro and incorporates changes in API. It will reduce the load of reviewing gp stuffs for Pyro devs. WDYT?

pyro/contrib/gp/parameterized.py
```diff
@@ -0,0 +1,262 @@
...
+    def __init__(self):
+        super(Parameterized, self).__init__()
+        self._constraints = OrderedDict()
+        self._priors = OrderedDict()
+        self._guides = OrderedDict()
+        self._mode = None
+
+    def set_constraint(self, name, constraint):
+        """
+        Sets a constraint to a parameter.
```
fritzo: nit: "Sets the constraint of an existing parameter"

pyro/contrib/gp/parameterized.py
```diff
@@ -0,0 +1,262 @@
...
+        self.register_parameter("{}_unconstrained".format(name), p_unconstrained)
+        # due to precision issue, we might get f(f^-1(x)) != x
+        # so it is necessary to transform back
+        p = transform_to(constraint)(p_unconstrained)
+        self.register_buffer(name, p.detach())
+        self._constraints[name] = constraint
+
+    def set_prior(self, name, prior):
+        """
+        Sets a prior to a parameter.
```
fritzo: nit: "Sets the prior distribution of an existing parameter"

pyro/contrib/gp/parameterized.py
```diff
@@ -0,0 +1,262 @@
...
+            self.register_parameter("{}_loc".format(name), loc)
+            self.register_parameter("{}_scale".format(name), scale)
+            dist_args = {"loc", "scale"}
+        elif dist_constructor is dist.MultivariateNormal:
+            loc = Parameter(biject_to(self._priors[name].support).inv(p).detach())
+            identity = eye_like(loc, loc.size(-1))
+            scale_tril = Parameter(identity.repeat(loc.shape[:-1] + (1, 1)))
+            self.register_parameter("{}_loc".format(name), loc)
+            self.register_parameter("{}_scale_tril".format(name), scale_tril)
+            dist_args = {"loc", "scale_tril"}
```
fritzo: Can we also check
```py
else:
    raise NotImplementedError("unsupported distribution type: {}".format(dist_constructor))
```

fehiepsi:
Thanks a lot for your review @fritzo ! I like to the idea of moving this class to `contrib.module`. About making this into 0.3, it is less priority to me, so feel free to make your choice.

pyro/contrib/gp/parameterized.py
fehiepsi: There already has a check for it from above. But I agree that there should be something to put here (because I used `elif`). Do you have any suggestion?

fritzo:
```py
else:
    raise NotImplementedError
```

fehiepsi:
It is great. Thanks for merging this, @fritzo !

PR #1606, opened 2018-12-02
Make noise is GPR/SGPR constraints.positive by default

fehiepsi:
This PR makes a small change at GPR/SGPR models:
+ Previously: noise's constraint is `constraints.greater_than(jitter)`. We only add noise to the diagonal of k(X, X).
+ Now: noise's constraint is `constraints.positive`. We will add noise+jitter to the diagonal of k(X, X).

This change does not affect much the inference and the term is added to k(X, X) is always guaranteed to greater than `jitter`.


fehiepsi:
It seems that travis is flaky?

PR #1609, opened 2018-12-02
add helper gp.util.train

fehiepsi:
This PR adds a helper `gp.util.train` to train gpmodel (instead of calling `gpmodel.optimize(...)`, we will use `util.train(gpmodel, ...)`. This is to make the training procedure closer to PyTorch way.

This is just a move the method `.optimize()` to `gp.util`, so I mark it easy.

### DIFF
--- a/pyro/contrib/gp/models/model.py
+++ b/pyro/contrib/gp/models/model.py
@@ -58,7 +58,7 @@ class GPModel(Parameterized):
       <http://pyro.ai/examples/svi_part_i.html>`_:
 
         >>> optimizer = pyro.optim.Adam({"lr": 0.01})
-        >>> svi = SVI(gpr.model, gpr.guide, optimizer, loss=Trace_ELBO())
+        >>> svi = pyro.infer.SVI(gpr.model, gpr.guide, optimizer, loss=pyro.infer.Trace_ELBO())
         >>> for i in range(1000):
         ...     svi.step()  # doctest: +SKIP
 
@@ -149,7 +149,7 @@ class GPModel(Parameterized):
             >>> Xu = torch.tensor([[1., 0, 2]])  # inducing input
             >>> likelihood = gp.likelihoods.Gaussian()
             >>> vsgp = gp.models.VariationalSparseGP(X, y, kernel, Xu, likelihood)
-            >>> svi = SVI(vsgp.model, vsgp.guide, optimizer, Trace_ELBO())
+            >>> svi = pyro.infer.SVI(vsgp.model, vsgp.guide, optimizer, pyro.infer.Trace_ELBO())
             >>> batched_X, batched_y = X.split(split_size=10), y.split(split_size=10)
             >>> for Xi, yi in zip(batched_X, batched_y):
             ...     vsgp.set_data(Xi, yi)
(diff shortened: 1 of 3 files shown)


pyro/contrib/gp/models/model.py
```diff
@@ -60,7 +58,7 @@ class GPModel(Parameterized):
       <http://pyro.ai/examples/svi_part_i.html>`_:
 
         >>> optimizer = pyro.optim.Adam({"lr": 0.01})
-        >>> svi = SVI(gpr.model, gpr.guide, optimizer, loss=Trace_ELBO())
+        >>> svi = pyro.infer.SVI(gpr.model, gpr.guide, optimizer, loss=pyro.infer.Trace_ELBO())
```
neerajprad: `SVI`, and `Trace_ELBO` should already be available in the doctest namespace, so you don't need to use absolute path. If you are adding it to be explicit, that's fine though!

pyro/contrib/gp/models/model.py
```diff
@@ -151,7 +149,7 @@ def set_data(self, X, y=None):
             >>> Xu = torch.tensor([[1., 0, 2]])  # inducing input
             >>> likelihood = gp.likelihoods.Gaussian()
             >>> vsgp = gp.models.VariationalSparseGP(X, y, kernel, Xu, likelihood)
-            >>> svi = SVI(vsgp.model, vsgp.guide, optimizer, Trace_ELBO())
+            >>> svi = pyro.infer.SVI(vsgp.model, vsgp.guide, optimizer, pyro.infer.Trace_ELBO())
```
neerajprad: Same here.

pyro/contrib/gp/util.py
```diff
@@ -274,3 +276,27 @@ def conditional(Xnew, X, kernel, f_loc, f_scale_tril=None, Lff=None, full_cov=Fa
...
+    :param ~pyro.contrib.gp.models.GPModel gpmodule: A GP module.
+    :param ~pyro.optim.PyroOptim optimizer: A Pyro optimizer.
+        By default, we use Adam with ``lr=0.01``.
+    :param ~pyro.infer.ELBO loss: A Pyro loss instance.
+        By default, ``loss=TraceMeanField_ELBO()``.
+    :param int num_steps: Number of steps to run SVI.
+    :returns: a list of losses during the training procedure
+    :rtype: list
+    """
+    optimizer = optim.Adam({"lr": 0.01}) if optimizer is None else optimizer
```
neerajprad: Any reason to override the default of `0.001`? I think users are probably used to that default learning rate for Adam.

pyro/contrib/gp/models/model.py
fehiepsi: I also think so, but triggered errors for doctest at this place. :( I'll test it again.

neerajprad: Try running `make doctest` for the correct command.

fehiepsi: I can't run `make doctest` in my local machine because of the error:`ImportError: Error importing plugin "tests.doctest_fixtures": No module named 'tests.doctest_fixtures'`. In last commit, I remove `pyro.infer` and this error is triggered in Travis: `UNEXPECTED EXCEPTION: NameError("name 'SVI' is not defined",)` at these places

fehiepsi: @neerajprad I can import `tests.common` and can do other `make foo` except doctest. ^^! I have installed all required pytest plugins and used `pytest=3.7.4` as suggested in `setup.py` file. I installed most of my packages using conda instead of `pip`. That might be the problem.

Issue #1581, opened 2018-11-28
OED example throwing NANs, flaking CI tests


fritzo:
Example error log:
https://travis-ci.org/uber/pyro/jobs/460986415

<details>

```
=================================== FAILURES ===================================
_____________ test_cpu[contrib/oed/ab_test.py --num-vi-steps=1000] _____________
example = ['contrib/oed/ab_test.py', '--num-vi-steps=1000']
    @pytest.mark.parametrize('example', CPU_EXAMPLES)
    def test_cpu(example):
        logger.info('Running:\npython examples/{}'.format(example))
        example = example.split()
        filename, args = example[0], example[1:]
        filename = os.path.join(EXAMPLES_DIR, filename)
>       check_call([sys.executable, filename] + args)
tests/test_examples.py:137: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
popenargs = (['/home/travis/virtualenv/python2.7.14/bin/python', '/home/travis/build/uber/pyro/examples/contrib/oed/ab_test.py', '--num-vi-steps=1000'],)
kwargs = {}, retcode = 1
cmd = ['/home/travis/virtualenv/python2.7.14/bin/python', '/home/travis/build/uber/pyro/examples/contrib/oed/ab_test.py', '--num-vi-steps=1000']
    def check_call(*popenargs, **kwargs):
        """Run command with arguments.  Wait for command to complete.  If
        the exit code was zero then return, otherwise raise
        CalledProcessError.  The CalledProcessError object will have the
        return code in the returncode attribute.
    
        The arguments are the same as for the Popen constructor.  Example:
    
        check_call(["ls", "-l"])
        """
        retcode = call(*popenargs, **kwargs)
        if retcode:
            cmd = kwargs.get("args")
            if cmd is None:
                cmd = popenargs[0]
>           raise CalledProcessError(retcode, cmd)
E           CalledProcessError: Command '['/home/travis/virtualenv/python2.7.14/bin/python', '/home/travis/build/uber/pyro/examples/contrib/oed/ab_test.py', '--num-vi-steps=1000']' returned non-zero exit status 1
/opt/python/2.7.14/lib/python2.7/subprocess.py:186: CalledProcessError
------------------------------ Captured log call -------------------------------
test_examples.py           133 INFO     Running:
[...]

PR #1614, opened 2018-12-02
Enable oed example

fehiepsi:
I test it locally and it seems that things are still fine. #1581


fritzo:
That test seems to be flaky, meaning that it sometimes passes and sometimes fails. Unless you have a good reason to believe the flakiness has been fixed, we should keep it marked xfail to avoid slowing down other unrelated PRs.

Issue #1532, opened 2018-11-13
Hmc, Nuts gaussian chain tests are flaky and slow


fritzo:
These tests are flaky and slow and are impeding development, e.g.
```
FAILED tests/infer/mcmc/test_nuts.py::test_nuts_conjugate_gaussian[dim=5_chain-len=9_num_obs=1]
FAILED tests/infer/mcmc/test_hmc.py::test_hmc_conjugate_gaussian[dim=10_chain-len=4_num_obs=1]
```
Can we replace some of these with less flaky tests?

neerajprad:
Is the issue that these tests are giving non-deterministic results (we had a few instances of this, but that has been resolved), or that they are extremely sensitive to small changes in the algorithm? I think at this point there is very little signal that we get from these conjugate gaussian tests, so I would be inclined to skip them altogether if I am unable to make them faster. They are still useful for profiling and checking if inference has improved due to a better chosen mass matrix etc., so we can keep them around.

fehiepsi:
@neerajprad FYI, pyro NUTS is 20x slower than pymc3 NUTS. This is unacceptable :(

fehiepsi:
I updated pytorch to lastest version and turned on `jit_compile`. Things are faster. Now, pyro is 4x slower than pymc.
You can get profiles file here: https://1drv.ms/f/s!AiPrcBQmpf6EgSW6A-7-QD77Y5Ax

eb8680:
@fehiepsi can you clarify what models you're using and what you mean by "compiling in pymc" and "without compiling, pyro is a little bit faster than pymc"?  Are you saying your model in Pyro is slower with the JIT than without in the latest version of PyTorch?

fehiepsi:
@eb8680 FYI, here are scripts to replicate results in pyro (1.0 branch) and pymc3:
```
# Source: Statistical Rethinking with Python and PyMC3
import numpy as np
import pandas as pd
import theano
# uncomment the following line to disable jit
#theano.config.mode = 'FAST_COMPILE'
import pymc3 as pm

np.random.seed(0)
rugged_df = (pd.read_csv('rugged.csv', sep=';')
             .assign(log_gdp=lambda df: np.log(df.rgdppc_2000))
             .dropna(subset=['log_gdp']))

with pm.Model() as model:
    a = pm.Normal('a', 0., 100.)
    bR = pm.Normal('bR', 0., 10.)
    bA = pm.Normal('bA', 0., 10.)
[...]

PR #1616, opened 2018-12-02
Tiny cleanup nuts test

fehiepsi:
~This PR fixes various bugs reported in #1532 and #1511.~

This PR adds a clearer message for the reason why JIT is slow.

### DIFF
--- a/tests/infer/mcmc/test_hmc.py
+++ b/tests/infer/mcmc/test_hmc.py
@@ -83,8 +83,8 @@ TEST_CASES = [
     ),
     T(
         GaussianChain(dim=10, chain_len=4, num_obs=1),
-        num_samples=1600,
-        warmup_steps=300,
+        num_samples=800,
+        warmup_steps=200,
         hmc_params={'step_size': 0.46,
                     'num_steps': 5},
         expected_means=[0.20, 0.40, 0.60, 0.80],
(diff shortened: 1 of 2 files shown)


tests/infer/mcmc/test_nuts.py
```diff
@@ -53,20 +53,20 @@
     ),
     T(
         GaussianChain(dim=5, chain_len=2, num_obs=10000),
-        num_samples=800,
-        warmup_steps=200,
+        num_samples=400,
+        warmup_steps=20,
```
fehiepsi: this test is slow, so I reduce warmup_steps to the minimum of adaptation steps.

tests/infer/mcmc/test_hmc.py
```diff
@@ -123,7 +123,6 @@ def rmse(t1, t2):
     'fixture, num_samples, warmup_steps, hmc_params, expected_means, expected_precs, mean_tol, std_tol',
     TEST_CASES,
     ids=TEST_IDS)
-@pytest.mark.skip(reason='Slow test (https://github.com/pytorch/pytorch/issues/12190)')
```
neerajprad: I think we should skip this on CI for sure. All of these together add several minutes to testing time and provide little in the way of additional information (the code paths they test are covered by faster tests). They are very useful for local profiling though, so if they all take more than a minute I would keep the skip flag.

tests/infer/mcmc/test_nuts.py
```diff
@@ -96,7 +96,6 @@ def jit_idfn(param):
     'fixture, num_samples, warmup_steps, expected_means, expected_precs, mean_tol, std_tol',
     TEST_CASES,
     ids=TEST_IDS)
-@pytest.mark.skip(reason='Slow test (https://github.com/pytorch/pytorch/issues/12190)')
```
neerajprad: Same here - I think they can't be made much faster, and they will still add 5 mins to travis, so let us only keep these for local runs.

tests/infer/mcmc/test_hmc.py
fehiepsi: You are right! I should look at CI log before making this PR. So we'll keep this for profiling. I'll change the title for this PR though (clearer skip messages)

neerajprad:
You'll need to merge the latest dev.

Issue #1418, opened 2018-09-27
Test failures using PyTorch master


neerajprad:
While most test failures on `pytorch-0.5.0` are resolved, the following tests are still failing with the latest master branch. Some of these are recent failures. 

CPU Tests:
 - [x] #1417 tests/ops/test_contract.py::test_ubersum_sizes (#1414)
 - [x] #1590 test_svi_irregular_batch_size[JitTraceEnum_ELBO]
 - [x] many HMC/NUTS tests with step size adaptation (#1487) wontfix - this is an upstream pytorch issue.
 - [ ] tests/contrib/oed/test_ewma.py::test_ewma - Assertion failures (#1373)
 - [ ] test_eig_lm[model5-design5-y-w-donsker_varadhan_eig-args5-True-0.3] (#1373)
 - [ ] test_eig_lm[model10-design10-y-w1-donsker_varadhan_eig-args10-True-0.3] (#1373)
 - [x] tests/infer/test_enum.py::test_mixture_of_diag_normals (Assertion failures) (#1425)

fehiepsi:
Closed because these tests have been solved in #1617.

Issue #1373, opened 2018-09-08
Test OED code on CUDA


ae-foster:
As suggested by @eb8680, we ought to verify that `contrib.oed` and `contrib.glmm` examples and tests work correctly on GPU

martinjankowiak:
i am shocked by @ae-foster's impudence but i guess i can do it....

neerajprad:
Just a heads up that the following tests are failing already with PyTorch master with/without CUDA (it seems like there is a difference of 1 between the actual and expected values, so it's worth looking into this in case there is some bug upstream in PyTorch that we can fix before release):

 - `test_eig_lm[model10-design10-y-w1-donsker_varadhan_eig-args10-True-0.3]`
 - `test_eig_lm[model5-design5-y-w-donsker_varadhan_eig-args5-True-0.3]`

fehiepsi:
Tests in `oed` and `glmm` have been resolved in #1617. So I change the title of this to reflect a better situation. We still need to implement a CUDA version for oed example.

PR #1617, opened 2018-12-03
fix cpu/gpu oed's eig&ewma tests

fehiepsi:
This PR addresses bugs announced in #1373 and #1418.

We get those bugs because:
+ EmwaLog uses old style of `torch.autograd.Function`, which might not be supported anymore (it still works in some cases but there will be many edge cases around the old style).
+ EmwaLog is initialized with `torch.tensor`, so GPU tests will fail.

### DIFF
--- a/pyro/contrib/oed/eig.py
+++ b/pyro/contrib/oed/eig.py
@@ -5,10 +5,11 @@ import torch
 
 import pyro
 from pyro import poutine
-from pyro.contrib.oed.search import Search
-from pyro.infer import EmpiricalMarginal, Importance, SVI
 from pyro.contrib.autoguide import mean_field_guide_entropy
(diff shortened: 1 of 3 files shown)


pyro/contrib/oed/eig.py
```diff
@@ -334,7 +335,22 @@ def loss_fn(design, num_particles):
     return loss_fn
 
 
-class EwmaLog(torch.autograd.Function):
+class _EwmaLogFn(torch.autograd.Function):
```
fritzo: Can you add a dosctring explaining how this class differs from the usual `torch.log()`?

fehiepsi: We have doc string at EwmaLog though. I didn't change it so it might not show up in git diff.

PR #1620, opened 2018-12-04
Fix hmm example when run on the GPU

neerajprad:
This seems like a recent change (regression?)

```ipython
In [1]: import torch

In [2]: import numpy as np

In [3]: torch.tensor(np.array([1.]), dtype=torch.float32)
Out[3]: tensor([1.])

In [4]: torch.set_default_tensor_type('torch.cuda.FloatTensor')

In [5]: torch.tensor(np.array([1.]), dtype=torch.float32)
<ipython-input-5-5b6e0594b346> in <module>()
----> 1 torch.tensor(np.array([1.]), dtype=torch.float32)

RuntimeError: device.has_value() ? device.value().type() == type.device_type() : other.device().type() == type.device_type() ASSERT FAILED at /opt/conda/conda-bld/pytorch-nightly_1543828543801/work/torch/csrc/utils/tensor_new.cpp:94, please report a bug to PyTorch.
```

### DIFF
--- a/examples/hmm.py
+++ b/examples/hmm.py
@@ -200,7 +200,7 @@ def model_4(sequences, lengths, args, batch_size=None, include_prior=True):
         # Note the broadcasting tricks here: we declare a hidden torch.arange and
         # ensure that w and x are always tensors so we can unsqueeze them below,
         # thus ensuring that the x sample sites have correct distribution shape.
-        w = x = sequences.new_tensor(0, dtype=torch.long)
+        w = x = torch.tensor(0, dtype=torch.long)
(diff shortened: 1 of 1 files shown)


neerajprad:
Let us not merge this yet. I'm seeing some odd behavior, and don't completely understand why this example is failing.

neerajprad:
I have pasted a minimal example above, but I am not sure if this is a regression on the pytorch side.

neerajprad:
The PyTorch fix seems to be in dev, I will check with tomorrow's nightly build again.

PR #1621, opened 2018-12-04
Update GP use the new Parameterized class

fehiepsi:
This is the final piece of my GP revising work. Most of these are to remove `name` and `self.get_param(...)` patterns because we use the new Parameterized class. In addition, SVI is not suitable for Pyro GP any more.

TODO

- [x] Fix tutorial


pyro/contrib/gp/parameterized.py
```diff
@@ -162,7 +170,7 @@ def autoguide(self, name, dist_constructor):
         if dist_constructor is dist.Delta:
             p_map = Parameter(p.detach())
             self.register_parameter("{}_map".format(name), p_map)
-            self.set_constraint("{}_map".format(name), self._priors[name].support)
+            self.set_constraint("{}_map".format(name), _get_independent_support(self._priors[name]))
```
fehiepsi: we don't need to create `unconstrained` parameter for real independent constraint.

pyro/contrib/gp/util.py
```diff
@@ -278,25 +146,34 @@ def conditional(Xnew, X, kernel, f_loc, f_scale_tril=None, Lff=None, full_cov=Fa
     return (loc, cov) if full_cov else (loc, var)
 
 
-def train(gpmodule, optimizer=None, loss=None, num_steps=1000):
+def train(gpmodule, optimizer=None, loss_fn=None, retain_graph=None, num_steps=1000):
```
fehiepsi: `train` method now use PyTorch optimizer and elbo.differentiable_loss.

tests/contrib/gp/test_models.py
```diff
@@ -26,50 +23,61 @@
...
+            VariationalSparseGP,
+            X, y1D, _kernel(), _likelihood()
+        ),
+        T(
+            VariationalSparseGP,
+            X, y2D, _kernel(), _likelihood()
+        ),
+    ]
+
+    return TEST_CASES
```
fehiepsi: Previously, kernel and likelihood are global variables, so each tests will update their parameters. To prevent that, we need to generate new kernel and likelihood for each test. This is different to before: where learned params are stored in ParamStoreDict and are cleared after each test.

PR #1282, opened 2018-08-02
[Do Not Merge] Bayesian Regression Tutorial

neerajprad:
Opening a PR to track and easily view the current draft of the bayesian regression example. This is mostly for @eb8680. Please feel free to edit directly, but give a heads up so as to avoid painful merge conflicts.

PR #1623, opened 2018-12-04
bayesian regression tutorial part i

jpchen:
these intro two tutorials are a rehash of  #1282 and the existing tutorial.

This tutorial covers:
- linear regression
- bayesian linear regression via `random_module`
- autoguide (diagnormal)

part 2 shortly to follow

**TODO:**
- [x] posterior marginal plot 
- [x] explain autoguides
- [x] add proper links
- [x] add ref


robsalomone:
Hi @jpchen , am happy to give feedback on this if you can give me a few days.

jpchen:
@robsalomone we are releasing this week, so the earlier the better. also happy to address any comments you have in a follow up PR

eb8680:
What do you think about having `random_module` and `autoguide` in part 2 instead of part 1?  It seems like we should introduce those as tools for eliminating boilerplate rather than the default idiom for regression, since they don't work that well for nonlinear regression.

robsalomone:
> @robsalomone we are releasing this week, so the earlier the better. also happy to address any comments you have in a follow up PR

Hi @jpchen , sounds good, as this is already merged now I'll give feedback down the line and you can follow up.

Issue #1622, opened 2018-12-04
baseball.py example freezing on travis intermittently


neerajprad:
I noticed in https://github.com/uber/pyro/pull/1619 that the baseball.py example in `test_examples.py` froze midway and the build terminated due to `no output received`. 

My guess is that this is an issue with python multiprocessing not playing well on travis. If we continue to see this, I will default to running a single chain on travis.

fritzo:
Maybe if `multiprocessing.cpu_count() == 1` we should execute synchronously rather than in parallel?

neerajprad:
Currently, we would throw a warning in that case (I haven't seen such a warning on travis though), and set the number of processes to max cpu count. I just noticed though that we would still end up spawning a separate process rather than executing synchronously. I'll make that fix.

fehiepsi:
@neerajprad Do you think that this causes `sv-dkl` fail? It's "maybe" to me. ^^

neerajprad:
@fehiepsi - `sv-dkl` is just extremely slow. It doesn't freeze. You can see from the console output that it is progressing very slowly and hence times out because the entire stage takes longer than 50 mins. It also has a much higher failure rate (fails more than 50% of the times). This example on the other hand, doesn't fail often, but when it fails it just freezes and doesn't emit anything on to the console for 10 mins after which travis kills the job. My suspicion is that there is some deadlock that happens when we use multiprocessing. Other people have had issues with multiprocessing on travis, so I am pretty sure this example would work fine with `num_chains=1`.

fritzo:
One pattern I've often used in the past is
```py
def parallel_map(fun, args, num_threads=None):
    args = list(args)
    if num_threads is None:
        num_threads = multiprocessing.cpu_count()
    if len(args) < 2 or num_threads == 1:
        return list(map(fun, args))
    return multiprocessing.Pool(num_threads).map(fun, args)

```

fehiepsi:
@neerajprad I agree that it is extremely slow frequently in Travis recently. What I mean is that there might be some relevant to how much bandwidth Travis gives it in those cases. In our example tests, there are just two tests which consumes a lot of CPU cores: `sv-dkl` and `ss-vae`. It might not affect much other tests but for these two tests, they will be heavily affected. The thing is that `ss-vae` still passed, which makes me wondering. Because I can't replicate the problem locally in my machine, so all of these are just hypothesis. ^^

[...]

PR #1624, opened 2018-12-04
Fix bug with num chain initialization in MCMC

neerajprad:
Addresses #1520, #1622 (testing).
 - We were not setting num chains to cpu count, only raising a warning. This fixes this and adds a test to ensure that we run synchronously if cpu count is 1.
 - Using `mp.Queue()` with a timeout on `get`, instead of `Manager().Queue()`. Using the former is marginally faster. I was using the `Manager()` earlier because it fared better cleaning up after `KeyboardInterrupt`, but I have verified that the former works as well after some of our refactoring.

### DIFF
--- a/pyro/infer/mcmc/mcmc.py
+++ b/pyro/infer/mcmc/mcmc.py
@@ -242,8 +242,9 @@ class MCMC(TracePosterior):
         if num_chains > 1:
             cpu_count = mp.cpu_count()
             if num_chains > cpu_count:
-                warnings.warn("`num_chains` is more than CPU count - {}. "
-                              "Resetting num_chains to CPU count.".format(cpu_count))
+                warnings.warn("num_chains={} is more than cpu_count={}. "
+                              "Resetting number of chains to CPU count."
+                              .format(num_chains, cpu_count))
                 num_chains = cpu_count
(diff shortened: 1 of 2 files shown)


pyro/infer/mcmc/mcmc.py
```diff
@@ -146,7 +146,7 @@ def _traces(self, *args, **kwargs):
                 w.start()
             while active_workers:
                 try:
-                    chain_id, val = self.result_queue.get_nowait()
+                    chain_id, val = self.result_queue.get(timeout=5)
```
fehiepsi: I wonder if this will make any difference?

neerajprad: We are blocking and not infinitely looping until an item is available. This means fewer accesses to the internal lock (and hence it is slightly faster) as compared to the non blocking version; but the main thread cannot do anything else while it is blocked. In our case, the main thread just loops back, so this is completely fine. The only reason why we had a non blocking version was to exit on `KeyboardInterrupt`, but having a timeout works quite well too.

fehiepsi: Thanks for explaining @neerajprad !

PR #1627, opened 2018-12-04
Add short tutorial on using the jit compiler

fritzo:
I had to modify `Delta` and `AutoContinuous` to suppress unnecessary compiler warnings.

This also fixes indentation in one cell in the custom_losses tutorial.

### DIFF
--- a/pyro/contrib/autoguide/__init__.py
+++ b/pyro/contrib/autoguide/__init__.py
@@ -362,7 +362,8 @@ class AutoContinuous(AutoGuide):
             unconstrained_value = latent[..., pos:pos + size].view(unconstrained_shape)
             yield site, unconstrained_value
             pos += size
-        assert pos == latent.size(-1)
+        if not torch._C._get_tracing_state():
+            assert pos == latent.size(-1)
 
     def __call__(self, *args, **kwargs):
         """
--- a/pyro/distributions/delta.py
+++ b/pyro/distributions/delta.py
@@ -35,7 +35,7 @@ class Delta(TorchDistribution):
         event_shape = v.shape[batch_dim:]
         if isinstance(log_density, numbers.Number):
             log_density = v.new_empty(batch_shape).fill_(log_density)
-        elif log_density.shape != batch_shape:
+        elif validate_args and log_density.shape != batch_shape:
             raise ValueError('Expected log_density.shape = {}, actual {}'.format(
                 log_density.shape, batch_shape))
         self.v = v
--- a/tutorial/source/custom_objectives.ipynb
+++ b/tutorial/source/custom_objectives.ipynb
@@ -78,10 +78,10 @@
     "\n",
     "```python\n",
     "def my_custom_L2_regularizer(my_parameters):\n",
-    "   reg_loss = 0.0\n",
-    "   for param in my_parameters:\n",
-    "       reg_loss = reg_loss + param.pow(2.0).sum()\n",
-    "   return reg_loss  \n",
+    "    reg_loss = 0.0\n",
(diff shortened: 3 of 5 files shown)


fritzo:
> with the simple ELBO from minipyro

Not sure I understand, since `pyro.ops.jit.trace` does not work with minipyro. Are you suggesting copying the minipyro `elbo` implementation into the tutorial?

Maybe we could instead show how to use `pyro.ops.jit.trace` to jit one of the custom losses from the custom loss tutorial?

eb8680:
> Are you suggesting copying the minipyro elbo implementation into the tutorial?

Yeah, or putting a JIT section into the custom loss tutorial as you suggest.

neerajprad:
All minor suggestions, feel free to ignore some/all as you see fit:
 - We should also mention about the `ignore_jit_warnings` flag in the Elbo / HMC classes.
 - You can think of this as a "static mode", where PyTorch usually operates in "eager mode". - I was a little confused when I read this - would suggest changing to “as opposed to“ or “instead of” in place of “where”.
 - `Note that since there were three different sequence lengths, compilation was triggered three times.` - Maybe we can put some kind of print statement in the model to demonstrate number of compilation passes, like `print(“.”)`.

neerajprad:
You'll also need to add this to `test_examples`.

fritzo:
@eb8680 I can't see a simple way to use `pyro.ops.jit.trace` due to the `model,guide` args, hence I'd rather discourage users from using that directly. Feel free to add a section in a later PR.

@neerajprad I've reworded as suggested. I haven't printed '.' because for some reason PyTorch runs each model twice during compilation (despite our specifying `check_trace=False`), and it seems like explaining why would only add to confusion. This is not added to test_examples because it is a tutorial rather than a test.

neerajprad:
> This is not added to test_examples because it is a tutorial rather than a test.

Sorry, I forgot that we grep on smoke_test and run all the notebook files in travis instead of explicitly invoking them.

eb8680:
> I can't see a simple way to use pyro.ops.jit.trace due to the model,guide args

Does `jit_loss = pyro.ops.jit.trace(functools.partial(loss, model, guide))` not work?

fritzo:
> Does jit_loss = pyro.ops.jit.trace(functools.partial(loss, model, guide)) not work?

Then it's a pain to use the jitted function, you can't use SVI and have to use optimizers directly. I think it's best to fix `pyro.ops.jit.trace` to simply handle non-tensor args. If you know how to use functools.partial in a simple way, please do so in a follow-up PR.

eb8680:
> Then it's a pain to use the jitted function, you can't use SVI and have to use optimizers directly.

That seems like a feature, not a bug.  I can add a section to the custom loss tutorial in a followup PR.

neerajprad:
> invocation using HMC/NUTS for the existing example?

I think it should be straightforward to use this in the NUTS kernel, will try and report back.

fritzo:
> @jpchen be nice to stick an assert between the non-jitted computation and the jitted

I feel like that would be too much code (at least 30% more code) and would detract from the tutorial.

> @jpchen invocation using HMC/NUTS for the existing example?

@neerajprad Could you add an HMC/NUTS section, either in this PR or in a follow-up PR?

neerajprad:
> @neerajprad Could you add an HMC/NUTS section, either in this PR or in a follow-up PR?

I am happy to add it in a follow-up PR.

neerajprad:
Are there any changes you would like to make to this one, before we merge this?

fritzo:
Well I'd like to merge and iterate :smile:

PR #1630, opened 2018-12-05
Bayesian Regression Part II

jpchen:
this tutorial covers:
- writing a manual guide
- using HMC
- comparing posteriors

**TODO**
- [x] mvn transforms in guide
- [x] more descriptions on dist choices
- [x] references for algorithms


neerajprad:
@jpchen - Can we restrict the scope of this and just cover HMC and the DiagNormal autoguide for now? I think we should have all the material and plots for that.

jpchen:
> Can we restrict the scope of this and just cover HMC and the DiagNormal autoguide for now

you mean diagnormal guide? the autoguide was in the first part. i can do the mvn one we have the plots for comparisons already, i just need to use a better variational initialization i will update it and you can see if it's necessary

neerajprad:
> you mean diagnormal guide? the autoguide was in the first part. i can do the mvn one we have the plots for comparisons already, i just need to use a better variational initialization i will update it and you can see if it's necessary

Sorry, I completely missed that we ended up using the autoguide in the first part itself. My impression was that we were going to not invoke auto-guides in that. I think the diagnormal guide works great as a comparison against HMC, as it is instructive to see the implications of the modeling assumptions i.e. independent latents, and the under-dispersion due to Klqp loss. Are we planning to do that first before comparing again the mvn guide?

jpchen:
>Are we planning to do that first before comparing again the mvn guide? I am not sure if comparing directly against the mvn guide has the same educational value.

the flow is `diagnormal -> hmc -> mvn` with comparisons at each `->`. after discussing with you offline decided to not do the fancier thing as it requires too much clutter code.

tutorial/source/bayesian_regression_ii.ipynb
```diff
@@ -0,0 +1,728 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from __future__ import absolute_import, division, print_function\n",
+    "\n",
+    "import argparse\n",
```
fritzo: nit: no need for argparse in a notebook, right?

tutorial/source/bayesian_regression_ii.ipynb
```diff
@@ -0,0 +1,728 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%matplotlib inline\n",
+    "logging.basicConfig(format='%(message)s', level=logging.INFO)\n",
+    "# Enable validation checks\n",
+    "pyro.enable_validation(True)\n",
```
fritzo: `assert pyro.__version__ == ...` as we do in other tutorials

tutorial/source/bayesian_regression_ii.ipynb
```diff
@@ -0,0 +1,728 @@
...
+    "          guide, \n",
+    "          optim.Adam({\"lr\": .005}), \n",
+    "          loss=Trace_ELBO(), \n",
+    "          num_samples=1000)\n",
+    "is_cont_africa, ruggedness, log_gdp = train[:, 0], train[:, 1], train[:, 2]\n",
+    "pyro.clear_param_store()\n",
+    "num_iters = 8000 if not smoke_test else 2\n",
+    "for i in range(num_iters):\n",
+    "    elbo = svi.step(is_cont_africa, ruggedness, log_gdp)\n",
+    "    if i % 100 == 0:\n",
```
fritzo: nit: maybe every 400 iters would be less verbose?

tutorial/source/bayesian_regression_ii.ipynb
```diff
@@ -0,0 +1,728 @@
...
...wEiMUEZFk+9nFDkJBH/duDN16rSDgY3dlPieaexmPRNMYnYhMp0RBUuLR5xvJ93v50L3rF33urzywEWvh6NWeJEQmIiKp0D4wxo3eUV5XU4TbdWcji/s3hRiPxDh1vS9N0YnIbJQoSNK19Y/x0/PtfOxgNQHf4stiqosCbCnL4ejVHqIxm4QIRUQk2U409+IysHd9wYz31ocCVORnc/yaWmKLrCQqZpakmCpcA3iuvpOYhQ8fWPxswpT7NhbxzbprNHQOsbUsNxEhiohIikRjllPX+9halktO1syvHsYYdlfm8+T5dga0zFRkxdCMgiSVtZYT13rZEAqwsXjhTkdz2VKWQ5bHxbmW/gRGJyIiqXC1e5jBsci8LVB3VOQBcKFtIFVhicgClChIUrX2j9E5NM7+ZfbH9rpdbC/P5dXWAS0/EhFZZc7fHMDjMmybZ0a4NDeLoqCP861KFERWCiUKklTnW/sxwM51ecsea3dlPqPhKE1dw8sPTEREUsJay8WbA2wuzcHnmftrhzGGnRV5NHYOMzim5UciK4FqFCSpXm0dYGNxkGCW5466haXYXJqL22W43D7I5tKZOzWLiMjKc6l9kN6RMG/aWrrgsVvLc3nhShd1jT28dWdZCqITkfloRkGSpmtonI7BcXZWLH82AcDncVEdCnClcygh44mISPI9dd7ZMHNbxcKNKKpDATwuw+HG7mSHJSJx0IyCJM2ltkHgtQK1eM0387ClNIcnz7czNB5ZVmwiIpIaz9V3UlngJy/bu+CxXreL9aEAhxuUKIisBJpRkKS50jFEUdBHKOhL2Ji1Jc6So4YOzSqIiKx0Q+MRTjb3LWq5aG1JkAttA/SNTCQxMhGJhxIFSYpILEZT13DCawkqC/1ke100qqBZRGTFe6Whm0jMLupeUFOcg7XwSmNPEiMTkXgoUZCkaO4ZYSIaY0uCEwWXMawvDHC9ZySh44qISOK9eKWLbK+LDaFA3OdUhfz4vW5eUZ2CSNopUZCkaOgYxmWgpiTx3YmqQwHaB8bUPk9EZIV74XIn920qwuOO/+uGx+XiwMZC1SmIrABKFCQprnYPU5HvJ9vrTvjY1aEAFjh9Xbs0i4isVDf7R2noHOb1m4sXfe79NUVcah+ke2g8CZGJSLyUKEjCTURiXO8ZYWNR/FPNi7E+FMAAJ5p7kzK+iIgs3wuXuwB4cMv
```
fritzo: ditto, every 400 iters

fritzo: nit: the last two plots are a little busy with all of HMC + diag + multivariate. I think it would be clearer to simply show HMC + multivariate.

tutorial/source/bayesian_regression_ii.ipynb
```diff
@@ -0,0 +1,728 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Utility function\n",
+    "def summary(traces, sites):\n",
+    "    marginal = EmpiricalMarginal(traces, sites)._get_samples_and_weights()[0].detach().cpu().numpy()\n",
```
neerajprad: This can be simplified to `EmpiricalMarginal(..).enumerate_support().cpu().numpy`. 

@eb8680 - what do you think about the marginals for both SVI and MCMC to return what we currently have as `_Marginal` within MCMC (without the diagnostics which MCMC can override). While `EmpiricalMarginal` works here, it wouldn't work when latents have different shapes. Moreover, we now have this ugly behavior where `.marginal()` on SVI returns `EmpiricalMarginal` and `.marginal()` for MCMC returns an internal `_Marginal` class.

eb8680: I'm fine with updating `SVI` to match `MCMC`.  We can always revisit these APIs further after the release.

jpchen: +1, in part one i have to manually run the empirical distribution to avoid shape errors

jpchen:
>  Again, this looks great, an I expect we'll use these plots in future slide decks and tutorials!

+1, most of these are repurposing/tweaking plots that @neerajprad  made!

PR #1633, opened 2018-12-05
Minor fixes to tqdm progress bars used for MCMC logging

neerajprad:
Addresses #1520. 

I have shelved my current effort to reimplement progress bars in the `progbar` branch, and tried to get around some of tqdm's limitations instead (https://github.com/tqdm/tqdm/issues/630, https://github.com/tqdm/tqdm/issues/649). This was because while I was able to solve some of these limitations, I was introducing new ones which tqdm handles well like synchronizing logging output with progress bar updates.
 - Remove remaining time counter which isn't that useful specially during the warmup phase because step size is being constantly tuned. Also reduces the width of the description, making it a bit more unlikely to bump into issues like in https://github.com/tqdm/tqdm/issues/630. 
 - Diagnostics are formatted with fixed width so as to reduce jumpy behavior.
 - Slightly different configuration between jupyter notebook environment and otherwise, inc. cleaning up and bar formatting.

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -411,6 +411,6 @@ class HMC(TraceKernel):
 
     def diagnostics(self):
         return OrderedDict([
-            ("Step size", self.step_size),
-            ("Acceptance rate", self._accept_cnt / self._t)
+            ("step size", "{:.2e}".format(self.step_size)),
+            ("acc. rate", "{:.3f}".format(self._accept_cnt / self._t))
         ])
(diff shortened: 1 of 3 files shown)


pyro/infer/mcmc/hmc.py
```diff
@@ -411,6 +411,6 @@ def sample(self, trace):
 
     def diagnostics(self):
         return OrderedDict([
-            ("Step size", self.step_size),
-            ("Acceptance rate", self._accept_cnt / self._t)
+            ("step size", "{:.2e}".format(self.step_size)),
+            ("acc. rate", "{:.3f}".format(self._accept_cnt / self._t))
```
fehiepsi: I have wanted this for a while. Your trick `"{:.2e}".format(...)` is +1000 to me.

PR #1636, opened 2018-12-06
Implement forward-filter backward-sample poutine

fritzo:
Pair coded with @eb8680 

This factors out an `infer_discrete` poutine from the `collapse-messenger` branch. The future `collapse` messenger is a thin wrapper around `infer_discrete`; I will rebase that and the `mis_ssm` branches once this merges.

The new `infer_discrete` poutine is a stand-alone inference algorithm in models with only discrete latent variables (observed variables can have any type). Applications include Viterbi-like MAP estimation and forward-filter backward-sample inference. Eventually `TraceEnum_ELBO.smaple_posterior()` will use `infer_discrete`, reducing its complexity from quadratic to linear.

The algorithm is parametrized by a `temperature` variable that can be either temperature=1 for sampling from the posterior, or temperature=0 for MAP estimation.

## Tested

- added unit tests with a variety of model shapes

### DIFF
--- a/tutorial/source/enumeration.ipynb
+++ b/tutorial/source/enumeration.ipynb
@@ -13,6 +13,7 @@
     "\n",
     "- Pyro implements automatic enumeration over discrete latent variables.\n",
     "- This strategy can be used alone or inside SVI (via [TraceEnum_ELBO](http://docs.pyro.ai/en/dev/inference_algos.html#pyro.infer.traceenum_elbo.TraceEnum_ELBO)), HMC, or NUTS.\n",
+    "- The standalone [infer_discrete](http://docs.pyro.ai/en/dev/inference_algos.html#pyro.infer.discrete.infer_discrete) can generate samples or MAP estimates.\n",
     "- Annotate a sample site `infer={\"enumerate\": \"parallel\"}` to trigger enumeration.\n",
     "- If a sample site determines downstream structure, instead use `{\"enumerate\": \"sequential\"}`.\n",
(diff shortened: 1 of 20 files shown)


martinjankowiak:
cool! should we use the name `posterior` for something rather specific?

fritzo:
I welcome a better name suggestion :smile:

fritzo:
How about one of the following:
- `infer_discrete`
- `transduce_discrete`
- `impute_discrete`
- `impute`
- `sample_posterior`

I think this might also better belong in `pyro.infer` than in `pyro.poutine`, since it is really a high-level inference algorithm built from smaller poutines...

tutorial/source/enumeration.ipynb
```diff
@@ -207,6 +210,49 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "### Examining discrete latent states <a class=\"anchor\" id=\"Examining-discrete-latent-states\"></a>\n",
+    "\n",
+    "While enumeration in SVI allows fast learning of parameters like `p` above, it does not give access to predicted values of the discrete latent variables like `x,y,z` above. We can access these using a standalone `infer_discrete` poutine. In this case the guide was trivial, so we can simply wrap the model in `infer_discrete`. We need to pass a `first_available_dim` argument to tell `infer_discrete` which dimensions are available for enumeration; this is related to the `max_plate_nesting` arg of `TraceEnum_ELBO` via\n",
```
eb8680: Can you mention that score-function gradients through `infer_discrete` won't be correct? Also a nit: "standalone `infer_discrete` poutine" -> "standalone `infer_discrete` handler"

PR #1638, opened 2018-12-06
Using cache from Transform and event_dim=1

stefanwebb:
This pull request suggests the following two changes in the normalizing flow library:

1. Using the caching facility in the `Transform` class rather than implementing it ourselves in `InverseAutoregressiveFlow` etc. I think this is a good idea for several reasons. Firstly, it only stores 1 value so the cache won't explode if you keep sampling but don't score (and thus pop off the cache). Secondly, it makes it possible to invert the forward and inverse operator and have everything work as expected. So you can produce MAF from IAF by calling `maf.inv()`

2. Setting `event_dim=1` in `InverseAutoregressiveFlow` because these flows operate on vectors. Now `log_abs_det_jacobian` returns a scalar rather than a vector, like it should. I think this change may break how IAF is used in the DMM example though...


martinjankowiak:
hmm, i don't seem to be able to reproduce the dmm error

eb8680:
> hmm, i don't seem to be able to reproduce the dmm error

I didn't try it myself, I just took the error message from Travis.

stefanwebb:
I believe all of these failed tests are due to changing the `event_dim` of IAF from 0 to 1, not the change for the cache. Although my current understanding of the internals may not be deep enough to fix it without some substantial help... This might be a good opportunity to dig into the workings of Pyro and the autoguide code in particular

stefanwebb:
@martinjankowiak does Pyro require caching multiple sampled values from normalizing flows at a time before they're scored? I need to put that logic back in if that's the case. I think there should be a class that takes care of this, because the code would be shared across all normalizing flows...

stefanwebb:
@martinjankowiak I haven't run it yet, but the assert in line 253 of `dmm.py` (`assert z_dist.event_shape == ()`) would only be thrown in the case that the IAFs are enabled

martinjankowiak:
no theres no such use case currently

stefanwebb:
Okay, how about we keep the simple caching and I'll try and sort out the compatibility issues with the change in `event_dim`

fritzo:
@stefanwebb are you still working on this PR, or can we close it?

stefanwebb:
No I still intend to finish it off. Will look into it now

stefanwebb:
I think this PR is good to go!

examples/dmm/dmm.py
```diff
@@ -250,16 +254,22 @@ def guide(self, mini_batch, mini_batch_reversed, mini_batch_mask,
...
-                assert z_dist.batch_shape == (len(mini_batch), self.z_q_0.size(0))
+                    assert z_dist.event_shape == ()
+                    assert z_dist.batch_shape == (len(mini_batch), self.z_q_0.size(0))
 
                 # sample z_t from the distribution z_dist
                 with pyro.poutine.scale(scale=annealing_factor):
-                    z_t = pyro.sample("z_%d" % t,
-                                      z_dist.mask(mini_batch_mask[:, t - 1:t])
-                                            .to_event(1))
+                    if len(self.iafs) > 0:
```
martinjankowiak: lgtm but can we add a comment about the presence/absence of `to_event` in the two cases?

stefanwebb: Okay, I'd added a few comments here:

`# in output of normalizing flow, all dimensions are correlated (event shape is not empty)`
`...`
`# when no normalizing flow used, ".to_event(1)" indicates latent dimensions are independent`

PR #1641, opened 2018-12-06
Show minipyro code on tutorials page

fritzo:
This also fixes a latex error in csis.ipynb that had been breaking the tutorials build.

screen shot 2018-12-06 at 12 31 35 pm


fritzo:
@jpchen this also suggests a way to add some of examples/*.py to our website.

jpchen:
>this also suggests a way to add some of examples/*.py to our website.

is this something we want to do?

fritzo:
> is [adding examples to the website] something we want to do?

Yes, I think it would be nice add a toc section for "Code Examples" and expose some of examples/*.py on pyro.ai/examples (with links to github, as in this PR). I think that would aid discoverability.

Any other votes?

martinjankowiak:
fine by me if it's a well chosen subset (lda and sparse_gamma_def, yes, dmm. no; etc)

eb8680:
> Any other votes?

One consideration to keep in mind is that we'd have to keep the examples pinned to Pyro release compatibility, rather than Pyro `dev`.  That might not be so bad when we release more frequently, though.

PR #1629, opened 2018-12-05
Add effect handler tutorial

eb8680:
Resolves #1451 

This PR adds a tutorial that explains the `Messenger` API in more detail and shows a couple of short examples of using existing handlers and implementing new ones.  It is intended as a follow-up piece to `minipyro` and does not attempt to re-explain machinery like `pyro.poutine.runtime.apply_stack` in detail.

PR #1642, opened 2018-12-06
Update effect handler tutorial

eb8680:
This PR responds to @jpchen's comments on #1629 and @martinjankowiak's request for more comments in `minipyro`.  Among other things, it adds a section on the handler stack to the effect handler tutorial.

### DIFF
--- a/pyro/contrib/minipyro.py
+++ b/pyro/contrib/minipyro.py
@@ -125,26 +125,50 @@ class PlateMessenger(Messenger):
         return range(self.size)
 
 
+# apply_stack is called by pyro.sample and pyro.param.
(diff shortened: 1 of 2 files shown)


eb8680:
> i'm wondering if we should put this in a separate "advanced" section to denote it's not a prerequisite to using the language.

Sure, feel free to open a PR reorganizing the tutorial headers

fritzo:
> "advanced"

I kind of like this in the intro section. I'm guessing most users will first see some easy stuff and then see some sophisticated stuff and think "ok, I get this easy stuff, and if I ever have spare time I can dig into the sophisticated stuff".

PR #1643, opened 2018-12-06
Fix an important bug in gp likelihood

fehiepsi:
While making gplvm tutorial, I observe that there is a bug in the likelihood module. The `scale` of Normal distribution should be `variance.sqrt()`, not `variance`. This explains why when I used TraceMeanField_ELBO in sv-dkl example, the performance does not improved.

I have tested `sv-dkl` examples and after 18 epochs, it reachs 99% accuaracy. This is unable to obtain when training with the bug.

In addition, I enhance sgpr a little bit to deal with the precision issue (clamping the trace_term, which is theoretically always nonnegative, by 0).


examples/contrib/gp/sv-dkl.py
```diff
@@ -116,7 +116,7 @@ def main(args):
 
     optimizer = torch.optim.Adam(gpmodule.parameters(), lr=args.lr)
 
-    elbo = infer.JitTrace_ELBO() if args.jit else infer.Trace_ELBO()
+    elbo = infer.JitTraceMeanField_ELBO() if args.jit else infer.TraceMeanField_ELBO()
```
fehiepsi: now TraceMeanField_ELBO works better than Trace_ELBO

tutorial/source/gp.ipynb
```diff
@@ -658,7 +658,7 @@
...
...w/KPgWszxu8fV8CamoFFwIXAb8M/5I0Zb8b9+zP84z87LJeFcVbg+saF/2htUHtR7EPSQbAmvNHLwj68tBj2ITB10H+yR7XPgGuBH2e0HzCuEDUO6vs4cFdYPuB93L8fC/1+z1YfcB9wKrCSgSCIbB/m8lPKU0P9b8x+XaEtUmEK4HTgWaDR3d8GCI8NYVgUtd8G/AOQCusTga3u3pulhv31hf5tYXwhTQe6gZ+F6aufmtloimQfuvta4BvAauBt0vtkMcW1D/sd7T6L+r30KdK/ZXOIWoa1RjO7Aljr7i8P6iqK+o5WKQdBthu3RnqJlJmNAX4NfN7dtx9qaJa2gtVuZpcDG9x98RHWEMW+LSN9eP5Ddz8d2EV6WmMow70PxwPzSE9XTAFGA3MPUUPR/ftk6Joiq9XMbgV6gbv6m4aoZdhqNLNRwK3A/8zWPUQdxfj3vV8pB0EX6Tm8fs3AuohqwczKSYfAXe5+f2heb2aTQ/9kYENoH+7azwWuMLOVwD2kp4duA2rNrCxLDfvrC/01wOYC1tf/ml3u/mxYv490MBTLPrwIeMvdu929B7gfOIfi2of9jnafRfJeCidULwc+6WE+pUhqPJ504L8c3jPNwAtmNqlI6jtqpRwEzwOt4aqNCtIn5BZEUYiZGXA7sMTdv5nRtQDov3rgetLnDvrbrwtXIJwFbOs/lC8Ed7/F3ZvdfSrp/fSou38SeAy4aoj6+uu+Kowv6G837v4OsMbMTghNc4A3KJLMw0clAAABKklEQVR9SHpK6CwzGxX+vvvrK5p9mOFo99nDwCVmNj4c+VwS2grGzC4Dvghc4e67B9V+TbjqahrQCjzHML7f3f1Vd29w96nhPdNF+mKQdyiifXhUoj5JUcgf0mfwl5G+muDWCOv4IOnDwFeAl8LPR0jPCS8COsLjhDDegO+Hul8F2oax1g8xcNXQdNJvsk7g34HK0F4V1jtD//Rhqu00oD3sx/8gffVF0exD4H8BbwKvAf9G+sqWSPchcDfpcxY9pP/DuuFY9hnpefrO8PMXw1BjJ+k59f73y48yxt8aalwKzM1oL8j7PVt9g/pXMnCyOJJ9mOuPPlksIhJzpTw1JCIiR0BBICIScwoCEZGYUxCIiMScgkBEJOYUBCIiMacgEBGJOQWBiEjM/X/U3q9iAnV+wwAAAABJRU5ErkJggg==\n",
```
fehiepsi: The current plot is at http://pyro.ai/examples/gp.html#More-Sparse-GPs. Because of the bug, two standard deviations cover all points, which should indicate that something is wrong.

pyro/contrib/gp/models/sgpr.py
```diff
@@ -139,6 +139,7 @@ def model(self):
                 D = D + Kffdiag - Qffdiag
             else:  # approx = "VFE"
                 trace_term = (Kffdiag - Qffdiag).sum() / self.noise
+                trace_term = trace_term.clamp(min=0)
```
fehiepsi: This is to guarantee `trace_term` nonnegative. If it is negative, `pyro.sample("trace_term", dist.Bernoulli(probs=torch.exp(-trace_term / 2.))` will throw error because `probs > 1`.

PR #1645, opened 2018-12-06
Fix the flakiness of oed example

fehiepsi:
This PR should fix the flakiness reported at #1581.

cc @fritzo


examples/contrib/oed/ab_test.py
```diff
@@ -84,9 +84,9 @@ def true_ape(ns):
     return torch.tensor(true_ape)
 
 
-def main(num_vi_steps, num_bo_steps):
+def main(num_vi_steps, num_bo_steps, seed):
 
-    pyro.set_rng_seed(42)
+    pyro.set_rng_seed(seed)
```
fehiepsi: To replicate the `NaN` issue, we can run the test with `seed=5`.

examples/contrib/oed/gp_bayes_opt.py
```diff
@@ -89,6 +89,8 @@ def opt_differentiable(self, differentiable, num_candidates=5):
             x_init = self.gpmodel.X.new_empty(1).uniform_(
                 self.constraints.lower_bound, self.constraints.upper_bound)
             x, y = self.find_a_candidate(differentiable, x_init)
+            if torch.isnan(y):
+                continue
```
fehiepsi: Here we find 5 candidates given 5 different initial points. If an initial point is so bad, we will get `NaN`. As the result,
the `mvalue, argmin = torch.min(torch.cat(values), dim=0)` line will return `NaN`. We should eliminate `NaN` points from candidates.

fritzo: This will not produce the correct number of candidates. Could you change the loop to be something like
```py
max_tries = int(1e6)
for i in range(max_tries):
    if len(candidates) == num_candidates:
        break
    ...rest of loop...
```

fehiepsi: I think that it is unnecessary but I'll do that. Your suggestion can help in a very rare case that all 5 candidates give NaN (which should imply that something is wrong with the model, instead of wasting resources for `1e6` loop). We set several initializations just to overcome the local minimum problem.

fritzo: I'm happy merging this now if you want to fix this in a later lower-priority PR.

fehiepsi: Thanks! I’ll do that in later PR if necessary.

PR #1646, opened 2018-12-06
Fix infer_discrete to work under the PyTorch jit

fritzo:
This fixes a few jit warnings and patches `opt_einsum` (until https://github.com/dgasmith/opt_einsum/pull/77 is released) so that `infer_discrete` works with the PyTorch jit.

## Tested

- added an example of a jitted viterbi decoder


fritzo:
Thanks for reviewing!

PR #1648, opened 2018-12-07
Fix ubersum to be compatible with pytorch jit

fritzo:
Some failures remain until https://github.com/pytorch/pytorch/issues/14875 is fixed.

## Tested

- added a test of ubersum examples
- added a test running `TraceEnum_ELBO` on a nontrivial contraction

### DIFF
--- a/tests/ops/test_contract.py
+++ b/tests/ops/test_contract.py
@@ -9,6 +9,7 @@ import pytest
 import six
 import torch
 
+import pyro.ops.jit
 from pyro.distributions.util import logsumexp
 from pyro.ops.contract import _partition_terms, contract_tensor_tree, contract_to_tensor, naive_ubersum, ubersum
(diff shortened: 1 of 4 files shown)


tests/ops/test_contract.py
```diff
@@ -401,6 +402,28 @@ def test_ubersum(equation, batch_dims):
...
+        pytest.skip()
+
+    @pyro.ops.jit.trace
+    def jit_ubersum(*operands):
+        return ubersum(equation, *operands, batch_dims=batch_dims, modulo_total=True)
+
+    actual = jit_ubersum(*operands)
+
+    if not isinstance(actual, tuple):
+        pytest.xfail(reason="https://github.com/pytorch/pytorch/issues/14875")
```
eb8680: Do we want to xfail here or just patch `pyro.ops.jit.trace` until the PyTorch issue is resolved?

PR #1652, opened 2018-12-07
Add infer_discrete section to GMM tutorial

fritzo:
This adds a new section on `infer_discrete` to the GMM tutorial. This is nice because `infer_discrete` is how one would typically serve a classifier model using Pyro, when the model was originally trained using SVI with model-side enumeration.


neerajprad:
This is really nice!! Just a small nit:

“For an deeper look “ -> “For a deeper look”

PR #1656, opened 2018-12-07
Tutorial reorganzation

jpchen:
screen shot 2018-12-07 at 11 59 43 am
(photo is cropped but 4 examples are on there)

### DIFF
--- a/tutorial/source/RSA-hyperbole.ipynb
+++ b/tutorial/source/RSA-hyperbole.ipynb
@@ -5,7 +5,6 @@
    "metadata": {},
    "source": [
     "# Understanding Hyperbole using RSA\n",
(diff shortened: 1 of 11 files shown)


tutorial/source/RSA-hyperbole.ipynb
```diff
@@ -5,7 +5,6 @@
    "metadata": {},
    "source": [
     "# Understanding Hyperbole using RSA\n",
-    "(This tutorial written by Noah Goodman and Eli Bingham, following Kao, Wu, Bergen, and Goodman (2014).)\n",
```
jpchen: i still think we should attribute authors in contributed tutorials, but we can revive this discussion after 0.3

PR #1662, opened 2018-12-10
remove bad notes from hmm example

martinjankowiak:


examples/hmm.py
```diff
@@ -469,7 +475,7 @@ def main(args):
     # Finally we evaluate on the test dataset.
     logging.info('-' * 40)
     logging.info('Evaluating on {} test sequences'.format(len(data['test']['sequences'])))
-    sequences = torch.tensor(data['test']['sequences'], dtype=torch.float32)
+    sequences = torch.tensor(data['test']['sequences'], dtype=torch.float32)[..., present_notes]
```
fritzo: Note that this makes our perplexity numbers differ from other published results. Do you think it's worth adding a warning comment?

martinjankowiak: sure

martinjankowiak: done

PR #1666, opened 2018-12-13
Add option to use sv-dkl for Binary MNIST

fehiepsi:
This PR adds an option to allow users do classification for Binary MNIST.

~With pytorch 1.0, the result is nondeterministic (maybe due to `torch.cholesky` or `torch.trtrs`) but~ After 16 epochs, I can get the accuracy 99.41%, which is a little bit better than the result in the paper. Note that this is achieved without modifying cnn architecture (copied from pytorch mnist example) and learning rate, seed,... So users can get better results with some efforts of tuning these values.

In addition, this PR also adds sv-dkl example to the examples page. I have revised the introduction for that purpose.


examples/contrib/gp/sv-dkl.py
```diff
@@ -87,10 +105,13 @@ def main(args):
...
                                   data_dir=data_dir,
-                                  batch_size=args.batch_size,
+                                  batch_size=args.test_batch_size,
                                   dataset_transforms=[transforms.Normalize((0.1307,), (0.3081,))],
                                   is_training_set=False,
-                                  shuffle=True)
+                                  shuffle=False)
+    if args.cuda:
+        train_loader.num_workers = 1
+        test_loader.num_workers = 1
```
fehiepsi: this is to train faster in GPU (18s vs 22s for each epoch)

examples/contrib/gp/sv-dkl.py
```diff
@@ -102,14 +123,27 @@ def main(args):
...
-    # use MultiClass likelihood for 10-class classification problem
-    likelihood = gp.likelihoods.MultiClass(num_classes=10)
-    # Because we use Categorical distribution in MultiClass likelihood, we need GP model returns
-    # a list of probabilities of each class. Hence it is required to use latent_shape = 10.
+    batches = []
+    for i, (data, _) in enumerate(train_loader):
+        batches.append(data)
+        if i >= ((args.num_inducing - 1) // args.batch_size):
+            break
+    Xu = torch.cat(batches)[:args.num_inducing].clone()
```
fehiepsi: This fixes the logic of the old code, which is wrong when batch_size < num_inducing. We need to draw enough batches.

examples/contrib/gp/sv-dkl.py
fritzo: Do you think this might fix https://github.com/uber/pyro/issues/1540 ? Should we set both to 1 always, to avoid hanging on CI?

neerajprad: @fehiepsi - The default value of num_workers is 0, which corresponds to not invoking multiprocessing at all. Did you try with a higher number for `num_workers` say 2, or 4? 

@fritzo - By default, pytorch's dataloader does not invoke multiprocessing, so we should be good. @fehiepsi has already isolated the cause for #1540 which is platform dependent flaky regression `nn.conv2D` (probably related - https://github.com/pytorch/pytorch/issues/14456).

fehiepsi: @neerajprad I thought that it is just useful to load data while GPU is doing its stuff so didnt try a higher number. Will check it tonight. :) In MNIST pytorch example, they set numworker=1 and pin memory is True when using GPU. I don’t see any difference with pin memory so I don’t put it here.
@fritzo I isolated the bug at 
https://travis-ci.org/fehiepsi/travis_test/builds/467326092
with just a simple cnn operator. In some specific Travis system, it is fast. In the other ones it is slow.

fehiepsi:
Thanks for reviewing @fritzo ! I’ll add that test.

examples/contrib/gp/sv-dkl.py
fehiepsi: @neerajprad I set a higher num workers but it does not help for speed

neerajprad: Np, I was only curious if it might. Looks good!

PR #1667, opened 2018-12-13
Refactor polyphonic data loader to load other music datasets

neerajprad:
Tasks:
 - [x] Ensure that muse data can be pickled efficiently. Currently pickling the entire numpy array consumes more than 4GB and breaks. Resolved - dumping the sequences and constructing the full dense representation on load.
 - [x] Check that DMM, HMM examples run giving similar / same results as earlier.
 - [x] Test examples on CUDA to ensure that refactoring has not broken CUDA compatibility.

### DIFF
--- a/examples/dmm/dmm.py
+++ b/examples/dmm/dmm.py
@@ -271,7 +271,7 @@ def main(args):
     log = get_logger(args.log)
     log(args)
 
-    data = poly.load_data()
+    data = poly.load_data(poly.JSB_CHORALES)
     training_seq_lengths = data['train']['sequence_lengths']
     training_data_sequences = data['train']['sequences']
     test_seq_lengths = data['test']['sequence_lengths']
--- a/examples/dmm/dmm.py
+++ b/examples/dmm/dmm.py
@@ -279,12 +279,12 @@ def main(args):
     val_seq_lengths = data['valid']['sequence_lengths']
     val_data_sequences = data['valid']['sequences']
     N_train_data = len(training_seq_lengths)
-    N_train_time_slices = float(np.sum(training_seq_lengths))
+    N_train_time_slices = float(torch.sum(training_seq_lengths))
     N_mini_batches = int(N_train_data / args.mini_batch_size +
                          int(N_train_data % args.mini_batch_size > 0))
 
     log("N_train_data: %d     avg. training seq. length: %.2f    N_mini_batches: %d" %
-        (N_train_data, np.mean(training_seq_lengths), N_mini_batches))
+        (N_train_data, training_seq_lengths.float().mean(), N_mini_batches))
 
     # how often we do validation/test evaluation during training
     val_test_frequency = 50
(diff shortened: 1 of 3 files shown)


examples/dmm/polyphonic_data_loader.py
```diff
@@ -14,19 +14,46 @@
...
+PIANO_MIDI = dset("piano_midi",
+                  "https://d2fefpcigoriu7.cloudfront.net/datasets/polyphonic/piano_midi.pickle",
+                  "piano_midi.pkl",
+                  3857)
+
+# FIXME: t_max is too large for pickle.dump
+MUSE_DATA = dset("muse_data",
+                 "https://d2fefpcigoriu7.cloudfront.net/datasets/polyphonic/muse_data.pickle",
+                 "muse_data.pkl",
+                 4273)
```
neerajprad: Since `T_max` is much higher for this dataset, the unsparse pickle file gets too large (anything more than 4GB throws an error on Python 3 in mac - [Stackoverflow issue](https://stackoverflow.com/questions/31468117/python-3-can-pickle-handle-byte-objects-larger-than-4gb)). While there are ways to get around this, I think we should think of alternative ways of storing / reading in this data so as to not have to write / load such big objects in the first place. I will need to understand the code a bit better to figure out what to do here. 

@fritzo, @martinjankowiak - Feel free to hack on this branch in the meantime. Note that only this dataset is failing, but the other two are also probably loading in large objects in memory.

examples/dmm/dmm.py
```diff
@@ -294,17 +294,19 @@ def main(args):
     # package repeated copies of val/test data for faster evaluation
     # (i.e. set us up for vectorization)
     def rep(x):
-        y = np.repeat(x, n_eval_samples, axis=0)
-        return y
+        rep_shape = torch.Size([x.size(0) * n_eval_samples]) + x.size()[1:]
```
martinjankowiak: we would probably just use `vectorize_particles=True` now.... (needn't be in this PR)

examples/hmm.py
```diff
@@ -414,14 +414,14 @@ def main(args):
...
-    data = poly.load_data()
+    data = poly.load_data(poly.JSB_CHORALES)
 
     logging.info('-' * 40)
     model = models[args.model]
     logging.info('Training {} on {} sequences'.format(
         model.__name__, len(data['train']['sequences'])))
-    sequences = torch.tensor(data['train']['sequences'], dtype=torch.float32)
-    lengths = torch.tensor(data['train']['sequence_lengths'], dtype=torch.long)
+    sequences = data['train']['sequences']
```
martinjankowiak: where in the preprocessing did this become an array and not a list of sequences?

neerajprad: This is happening inside `load_data` when you call `load_data(poly.JSB_CHORALES)`.

examples/dmm/dmm.py
neerajprad: I didn't realize this was being used in that context, and probably tried being too clever coming up with a torch solution for this. Let us defer this to a separate PR.

neerajprad:
I think the examples failed because travis was using an outdated cache. I have cleared the cache and restarted the examples now.

fritzo:
I'm happy merging if @martinjankowiak is happy.

Issue #1549, opened 2018-11-20
More tutorials on Gaussian Processes


fehiepsi:
I think that it is good to have more examples to illustrate the flexibility and the scalable of GP module. I have two examples which come into mind, which I hope that we will have them before next release.

- [ ] Deep Gaussian Process. This is also a good example to illustrate various ways to use `mean_function` and how to use LBFGS.
- [x] Latent variable model. This [paper](https://academic.oup.com/bioinformatics/advance-article/doi/10.1093/bioinformatics/bty533/5047752) has various examples on how to use it on pseudotime estimation of cell dataset. This is a good chance to illustrate the usefulness of Reparameterized class.

AlejandroCatalina:
I thought of opening a new issue but I figured this one worked just fine!

First of all I want to congratulate all contributors because pytorch + pyro look really really nice, I'm looking forward to implementing actual models with them!

Nonetheless, I came here looking for a nice probabilistic implementation, and, as always, the first example I look at is the gaussian process. In this regard, I've seen the discussion about adding examples of how to learn the hyper parameters with HMC pop up in plenty of places, but haven't seem the actual examples. Is that because I'm too bad at finding them or because that's a blind spot in the documentation? 

I've seen the HMC tests and tried to run myself a HMC chain to learn the hyper parameters of the GP with no success. My code looks like this:

```python
N = 20
X = dist.Uniform(0.0, 5.0).sample(sample_shape=(N,))
y = 0.5 * torch.sin(3*X) + dist.Normal(0.0, 0.2).sample(sample_shape=(N,))

kernel = gp.kernels.RBF(input_dim=1, variance=torch.tensor(5.),
                        lengthscale=torch.tensor(10.))
gpr = gp.models.GPRegression(X, y, kernel, noise=torch.tensor(1.))

gpr.kernel.set_prior("lengthscale", dist.LogNormal(0.0, 1.0))
gpr.kernel.set_prior("variance", dist.LogNormal(0.0, 1.0))
# we reset the param store so that the previous inference doesn't interfere with this one
pyro.clear_param_store()

hmc_kernel = HMC(gpr.model)
mcmc_run = MCMC(hmc_kernel, num_samples=10000, warmup_steps=2000)
post_trace = defaultdict(list)
for trace, _ in mcmc_run._traces():
    variance_name = param_with_module_name(kernel.name, "variance")
    post_trace["variance"].append(trace.nodes[variance_name]["value"])
[...]

PR #1668, opened 2018-12-13
Tutorial on deep gaussian process

fehiepsi:
This PR tracks the progress of making deep gaussian process tutorial. It will be great to be able to answer the following questions in this tutorial:

- how to stack many GPs layers and do inference
- how to do inference with HMC
- how to make a mean function


martinjankowiak:
@fehiepsi i think we need to think carefully about which/how many tutorials we want. every tutorial we add has a cost in both: i) review and ii) maintenance. also, at the end of the day pyro is not a GP library, so it doesn't make sense to have excessively many GP tutorials. maybe we haven't reached that point yet, but it is something to keep in mind. also, if we move towards closer integration with gpytorch (which i am convinced we should) then tutorials will in any case require rewriting once the code integration is complete. broadly, i think we should think more carefully about where we want to take `pyro.contrib.gp` before adding much in the way of code/tutorials.

fehiepsi:
Hi Martin, thanks for your feedback! I have thought about the issues which you proposed. I intend to make this last one to show the flexibility of Pyro GP. Pyro GP has most features which other GP libraries have (not including GPyTorch). The features in this tutorial (HMC training, stacking layers) are hidden in the tests, so it is good idea to let users know. Anyway, I am persuaded that it is better to write a blog for it instead of putting in Pyro tutorials. So I will close this and the issue #1549.

RPrudden:
@fehiepsi did you ever end up writing this blog post?

fehiepsi:
@RPrudden Sorry, I don't have time recently. I should keep my words to write a blog-post about it. I will do  this or the next weekend and let you know.

fehiepsi:
@RPrudden You can find a simple tutorial [here in my blog](https://fehiepsi.github.io/blog/deep-gaussian-process/). It includes doing HMC/SVI for deep GPs. Hope that you will enjoy it. :)

PR #1670, opened 2018-12-14
Use native inverse for large matrices in rinverse

vishwakftw:
As specified in title.

cc: @fritzo @neerajprad

### DIFF
--- a/pyro/ops/linalg.py
+++ b/pyro/ops/linalg.py
@@ -26,20 +26,7 @@ def rinverse(M, sym=False):
     elif M.shape[-1] == 3:
         return inv3d(M, sym=sym)
     else:
-        # Use blockwise inversion
-        d = M.shape[-1]//2
-        A, B, C, D = M[..., :d, :d], M[..., :d, d:], M[..., d:, :d], M[..., d:, d:]
-        Ainv = rinverse(A, sym=sym)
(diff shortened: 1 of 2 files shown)


pyro/ops/linalg.py
```diff
@@ -5,43 +5,6 @@
...
-        inv = M.new_empty(M.shape)
-        inv[..., 0, 0] = M[..., 1, 1]
-        inv[..., 1, 1] = M[..., 0, 0]
-        inv[..., 0, 1] = -M[..., 0, 1]
-        inv[..., 1, 0] = -M[..., 1, 0]
-        return inv / det.unsqueeze(-1).unsqueeze(-1)
-    elif M.shape[-1] == 3:
-        return inv3d(M, sym=sym)
-    else:
-        # Use blockwise inversion
```
fritzo: I think it's probably safe to use `torch.inverse` here.

vishwakftw: Oh, sorry about this. I was unaware of this regression, which is why I made the change.

fritzo: No worries, I'm glad we can now use `torch.inverse()` for large matrices. As usual, we find ourselves in a neglected region of performance space, operating on large batches of small matrices :-/

PR #1678, opened 2018-12-19
Resolve the slowness of HMC during init adaptation phase

### LINKED ISSUE #1673: Pyro NUTS inference slow when compared to the same PyMC3 model
For the same models below, the PyMC3 model finishes within a second, whereas the Pyro model has an extremely slow rate of progress. I came across this perf issue while working on a larger model, and isolated it to the example below. I am not sure if this is a perf bug, or whether PyMC3 uses some adaptation tricks, but it is worth looking into and learning from. The cause of the slowdown is the very small step size in Pyro (~10^-9), whereas in PyMC3, the step size throughout is of the order of ~ 10^-3. This is also not a case of bad initialization, the same can be observed for different random seeds.

Also refer to #1470. 

```python
import pymc3 as pm
import pyro

import torch


def pm_model(data):
    with pm.Model() as model:
        alpha = pm.HalfCauchy('alpha', beta=2)
        beta = pm.HalfCauchy('beta', beta=2)
        l = pm.Gamma('l', alpha, beta)
        pm.Poisson('obs', l, observed=data)
    return model


def pyro_model(data):
    alpha = pyro.sample('alpha', dist.HalfCauchy(2.))
    beta = pyro.sample('beta', dist.HalfCauchy(2.))
    l = pyro.sample('l', dist.Gamma(alpha, beta))
    pyro.sample('obs', dist.Poisson(l), obs=data)


def get_samples_pymc(data, num_samples=200, warmup_steps=200):
    data = data.numpy()
    with pm_model(data):
        trace = pm.sample(draws=num_samples, tune=warmup_steps, chains=1)


def get_samples_pyro(data, num_samples=200, warmup_steps=200):
    data = data
    nuts_kernel = NUTS(pyro_model,
                       adapt_step_size=True,
                       adapt_mass_matrix=True,
                       jit_compile=True,
                       full_mass=True)
    mcmc_run = MCMC(nuts_kernel,
                    num_samples=num_samples,
                    warmup_steps=warmup_steps,
                    num_chains=1).run(data)


data = torch.tensor(4805497.)
get_samples_pymc(data)  # finishes in a second.
get_samples_pyro(data)  # orders of magnitude slower.
```

fehiepsi:
This PR tries to resolve #1673 by setting a high prox_center value during the first window_adaptation. 

In addition, this also fixes some bugs:

- Missing cache for HMC. Not sure why HMC tests passed previously.
- Better definition of tree_depth (for debugging)
- Feed python float into DualAveraging scheme. Previously it is a mixed of float and tensor_scalar, which in turns forces `step_size` be a tensor_scalar, not float.

More details are in comments.

Todo:

- [x] Clean the PR (lots of printing are added for debugging)

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -387,7 +387,6 @@ class HMC(TraceKernel):
                                                                               z_grads=z_grads)
             # apply Metropolis correction.
             energy_proposal = self._kinetic_energy(r_new) + potential_energy_new
-
         delta_energy = energy_proposal - energy_current
         # Set accept prob to 0.0 if delta_energy is `NaN` which may be
         # the case for a diverging trajectory when using a large step size.
@@ -407,6 +406,7 @@ class HMC(TraceKernel):
         self._t += 1
 
         # get trace with the constrained values for `z`.
+        z = z.copy()
         for name, transform in self.transforms.items():
             z[name] = transform.inv(z[name])
         return self._get_trace(z)
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -104,10 +104,8 @@ class HMC(TraceKernel):
             self.trajectory_length = step_size * num_steps
         else:
             self.trajectory_length = 2 * math.pi  # from Stan
-        self.adapt_step_size = adapt_step_size
         self._jit_compile = jit_compile
         self._ignore_jit_warnings = ignore_jit_warnings
-        self.full_mass = full_mass
         self._target_accept_prob = 0.8  # from Stan
         # The following parameter is used in find_reasonable_step_size method.
         # In NUTS paper, this threshold is set to a fixed log(0.5).
@@ -142,15 +140,11 @@ class HMC(TraceKernel):
         return self._trace_prob_evaluator.log_prob(model_trace)
 
     def _kinetic_energy(self, r):
-        # TODO: revert to `torch.dot` in pytorch==1.0
-        # See: https://github.com/uber/pyro/issues/1458
         r_flat = torch.cat([r[site_name].reshape(-1) for site_name in sorted(r)])
-        if self.full_mass:
-            return 0.5 * (r_flat * (self.inverse_mass_matrix.matmul(r_flat))).sum()
+        if self.inverse_mass_matrix.dim() == 2:
+            return 0.5 * self.inverse_mass_matrix.matmul(r_flat).dot(r_flat)
         else:
-            k = 0.5 * (self.inverse_mass_matrix * (r_flat ** 2)).sum()
-            #print("kinetic energy", k)
-            return k
+            return 0.5 * self.inverse_mass_matrix.dot(r_flat ** 2)
 
     def _potential_energy(self, z):
         if self._jit_compile:
@@ -194,10 +188,7 @@ class HMC(TraceKernel):
         return self._compiled_potential_fn(*vals)
 
     def _energy(self, z, r):
-        k = self._kinetic_energy(r)
-        p = self._potential_energy(z)
-        #print("energy", k, p, k+p)
-        return k + p
+        return self._kinetic_energy(r) + self._potential_energy(z)
 
     def _reset(self):
         self._t = 0
@@ -210,21 +201,23 @@ class HMC(TraceKernel):
         self._initial_trace = None
         self._has_enumerable_sites = False
         self._trace_prob_evaluator = None
+        self._z_last = None
         self._potential_energy_last = None
         self._z_grads_last = None
         self._warmup_steps = None
 
-    def _find_reasonable_step_size(self, z):
+    def _find_reasonable_step_size(self):
         step_size = self.step_size
 
         # We are going to find a step_size which make accept_prob (Metropolis correction)
         # near the target_accept_prob. If accept_prob:=exp(-delta_energy) is small,
         # then we have to decrease step_size; otherwise, increase step_size.
+        z, potential_energy, z_grads = self._fetch_from_cache()
         r, _ = self._sample_r(name="r_presample_0")
-        energy_current = self._energy(z, r)
-        z_new, r_new, z_grads, potential_energy = velocity_verlet(
-            z, r, self._potential_energy, self.inverse_mass_matrix, step_size)
-        energy_new = potential_energy + self._kinetic_energy(r_new)
+        energy_current = self._kinetic_energy(r) + potential_energy
+        z_new, r_new, z_grads_new, potential_energy_new = velocity_verlet(
+            z, r, self._potential_energy, self.inverse_mass_matrix, step_size, z_grads=z_grads)
+        energy_new = self._kinetic_energy(r_new) + potential_energy_new
         delta_energy = energy_new - energy_current
         # direction=1 means keep increasing step_size, otherwise decreasing step_size.
         # Note that the direction is -1 if delta_energy is `NaN` which may be the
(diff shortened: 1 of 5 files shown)


neerajprad:
> Missing fetch_from_cache for HMC. Not sure why HMC tests passed previously.

We fetch_from_cache in both, I don't see any changes related to that in this PR?

pyro/infer/mcmc/hmc.py
```diff
@@ -307,10 +321,13 @@ def initial_trace(self):
...
         for i in range(self._max_tries_initial_trace):
             trace_log_prob_sum = self._compute_trace_log_prob(trace)
+            z = {name: node["value"].detach() for name, node in self._iter_latent_nodes(trace)}
+            #print("in configure", z)
+            #print(trace_log_prob_sum)
             if not torch_isnan(trace_log_prob_sum) and not torch_isinf(trace_log_prob_sum):
                 self._initial_trace = trace
                 return trace
-            trace = poutine.trace(self.model).get_trace(*self._args, **self._kwargs)
+            trace = poutine.trace(self.model).get_trace(self._args, self._kwargs)
```
neerajprad: This seems to be a regression?

pyro/infer/mcmc/hmc.py
```diff
@@ -398,6 +415,7 @@ def sample(self, trace):
         if rand < accept_prob:
             self._accept_cnt += 1
             z = z_new
+            self._cache(potential_energy_new, z_grads_new)
```
fehiepsi: We should cache pe and its grad here.

pyro/infer/mcmc/hmc.py
fehiepsi: Sorry, I'll go for a pass to clean these diffs. Currently, NUTS is fast in double precision for your example, but slow in single precision. I'm investigating it.

pyro/infer/mcmc/hmc.py
neerajprad: Good catch! It is not doing any caching across sample steps since we recompute if the gradients / pe computation is None. That's why the tests weren't failing, but we weren't making use of this optimization for HMC.

neerajprad:
This is great, @fehiepsi! Thanks for digging into this, I am taking another look now.

pyro/infer/mcmc/hmc.py
```diff
@@ -232,11 +230,15 @@ def _find_reasonable_step_size(self, z):
...
         # keep scale step_size until accept_prob crosses its target
         # TODO: make thresholds for too small step_size or too large step_size
+        t = 0
         while direction_new == direction:
+            t += 1
             step_size = step_size_scale * step_size
-            z_new, r_new, z_grads, potential_energy = velocity_verlet(
-                z, r, self._potential_energy, self.inverse_mass_matrix, step_size)
-            energy_new = potential_energy + self._kinetic_energy(r_new)
+            r, _ = self._sample_r(name="r_presample_{}".format(t))
```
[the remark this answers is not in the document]
neerajprad: Interesting, makes sense.

pyro/infer/mcmc/hmc.py
```diff
@@ -343,38 +333,50 @@ def _initialize_model_properties(self):
...
+
+    def _initialize_sampling(self, trace):
+        z = {name: node["value"].detach() for name, node in self._iter_latent_nodes(trace)}
+        # automatically transform `z` to unconstrained space, if needed.
+        for name, transform in self.transforms.items():
+            z[name] = transform(z[name])
+        potential_energy = self._potential_energy(z)
+        self._cache(z, potential_energy, None)
+        if self._adapter.adapt_step_size:
+            self._adapter.reset_step_size_adaptation()
```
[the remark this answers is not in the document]
neerajprad: Hmm..why don't we do any state setup that is trace dependent, when we set the initial trace instead? Would suggest a minor change:
 - Renaming this to `_initialize_step_size` to make the objective clearer.
 - Calling this from inside `initial_trace` instead of inside `sample`. Note that you will need to call this from both the setter and the getter.

pyro/infer/mcmc/nuts.py
```diff
@@ -347,14 +354,18 @@ def sample(self, trace):
                         tree_weight = tree_weight + new_tree.weight
 
         if self._t < self._warmup_steps:
-            accept_prob = new_tree.sum_accept_probs / new_tree.num_proposals
+            accept_prob = sum_accept_probs / num_proposals
```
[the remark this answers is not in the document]
neerajprad: Does this seem to affect any of our results?

pyro/infer/mcmc/hmc.py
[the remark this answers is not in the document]
fehiepsi: Nice solution! Thanks @neerajprad !

pyro/infer/mcmc/nuts.py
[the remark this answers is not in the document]
fehiepsi: In the old approach, if tree is diverging, accept_prob for the new_tree will be smaller than the whole tree, which will lead to a smaller step_size for the next step. However, I don't observe any improvement with this change. All tests pass just well. :(

tests/infer/mcmc/test_hmc.py
```diff
@@ -272,7 +272,7 @@ def test_initial_trace(monkeypatch):
     true_coefs = torch.arange(1., dim + 1.)
     labels = dist.Bernoulli(logits=(true_coefs * data).sum(-1)).sample()
     # mock values to replay - from right to left
-    trace_log_prob_replay = [-5, float('NaN'), float('Inf')]
+    trace_log_prob_replay = [-5, -5, float('NaN'), float('Inf')]
```
fehiepsi: we need to compute potential energy for initializing step_size.

neerajprad:
> So in my opinion, we should encourage users reduce max_tree_depth if they want speed, and use double precision for tricky posterior.

I agree, that makes sense. We should, in that case, make `max_tree_depth` a configurable param for NUTS (in a separate PR perhaps).

pyro/infer/mcmc/hmc.py
```diff
@@ -309,13 +299,15 @@ def initial_trace(self):
...
                 self._initial_trace = trace
+                self._initialize_step_size()  # this method also caches z and its potential energy
                 return trace
             trace = poutine.trace(self.model).get_trace(*self._args, **self._kwargs)
         raise ValueError("Model specification seems incorrect - cannot find a valid trace.")
 
     @initial_trace.setter
     def initial_trace(self, trace):
         self._initial_trace = trace
+        self._initialize_step_size()
```
neerajprad: Sorry, my suggestion wouldn't work because `setup` is always called after we set the `intial_trace`, but there is a much simpler solution for this. Let us remove this from `initial_trace` completely, and just put it in `setup()` towards the very end. However, you will need to make one small change to `_initialize_step_size` :

```
    def _initialize_step_size(self):
        z = {name: node["value"].detach()
             for name, node in self._iter_latent_nodes(**self.initial_trace**)}  # not self._initial_trace
        ....
```

This will set the initial trace if it is not already specified, and should be much simpler overall.

neerajprad: The failing test should pass once you make this change.

neerajprad: Just pushed this to your branch - I hope that is okay!

fehiepsi: No worry, feel free to do so. :)

fehiepsi:
I think that the last commit raises another bug when users set initial_trace after setup. The cache and step size will be set for the trace generated from `setup`, not initial_trace.

neerajprad:
> I think that the last commit raises another bug when users set initial_trace after setup. 

This shouldn't happen though, because setup should only be called by `MCMC` and not by the user. I think we should also specify the initial trace inside as an argument to `MCMC` rather than it being a `TraceKernel` attribute, but we should make these interface changes separately.

fehiepsi:
> we should also specify the initial trace as an argument to `MCMC` rather than it being a `TraceKernel` attribute, but we should make these interface changes separately.

Agree!

PR #1680, opened 2018-12-20
Minor CUDA related fixes

neerajprad:


tests/test_examples.py
```diff
@@ -61,8 +61,6 @@
 
 CUDA_EXAMPLES = [
     'air/main.py --num-steps=1 --cuda',
-    xfail_param('baseball.py --num-samples=200 --warmup-steps=100 --num-chains=2 --cuda',
```
jpchen: ?

neerajprad: Will put this back when we have CUDA with parallel chains in. Currently we don't have `--cuda` option for the baseball example, so removing this from the tests.

examples/dmm/polyphonic_data_loader.py
```diff
@@ -94,6 +94,7 @@ def load_data(dataset):
         for k, v in dset.items():
             sequences = v["sequences"]
             dset[k]["sequences"] = pad_sequence(sequences, batch_first=True).type(torch.Tensor)
+            dset[k]["sequence_lengths"] = v["sequence_lengths"].to(device=torch.Tensor().device)
```
neerajprad: Fails non-deterministically for the HMM example using `--cuda` otherwise, because the sequence lengths are used to define the mask (I am not sure why this passes sometimes). cc. @martinjankowiak

PR #1686, opened 2018-12-28
Expose target_accept_prob and max_tree_depth for HMC/NUTS

fehiepsi:
Address some minor issues at #1093: expose target_accept_prob and max_tree_depth

There are a few small silent bugs which are also fixed in this PR. I'll comment at them.

### DIFF
--- a/pyro/infer/mcmc/mcmc.py
+++ b/pyro/infer/mcmc/mcmc.py
@@ -244,7 +244,7 @@ class MCMC(TracePosterior):
         CUDA.
     :param bool disable_progbar: Disable progress bar and diagnostics update.
     """
-    def __init__(self, kernel, num_samples, warmup_steps=0,
+    def __init__(self, kernel, num_samples, warmup_steps=None,
                  num_chains=1, mp_context=None, disable_progbar=False):
         self.warmup_steps = warmup_steps if warmup_steps is not None else num_samples // 2  # Stan
         self.num_samples = num_samples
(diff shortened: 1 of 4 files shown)


pyro/infer/mcmc/mcmc.py
```diff
@@ -257,10 +257,10 @@ def __init__(self, kernel, num_samples, warmup_steps=0,
                               .format(num_chains, available_cpu))
                 num_chains = available_cpu
         if num_chains > 1:
-            self.sampler = _ParallelSampler(kernel, num_samples, warmup_steps,
+            self.sampler = _ParallelSampler(kernel, num_samples, self.warmup_steps,
```
fehiepsi: this is a bug when `warmup_steps` is None, we should use `self.warmup_steps` here.

pyro/infer/mcmc/mcmc.py
```diff
@@ -244,9 +244,9 @@ class MCMC(TracePosterior):
         CUDA.
     :param bool disable_progbar: Disable progress bar and diagnostics update.
     """
-    def __init__(self, kernel, num_samples, warmup_steps=0,
+    def __init__(self, kernel, num_samples, warmup_steps=None,
                  num_chains=1, mp_context=None, disable_progbar=False):
-        self.warmup_steps = warmup_steps if warmup_steps is not None else num_samples // 2  # Stan
+        self.warmup_steps = num_samples if warmup_steps is None else warmup_steps  # Stan
```
fehiepsi: Correct the behaviour from Stan: Stan uses `iter` and set default `warmup = iter // 2`. Iter in Stan includes warmup, not separated from warmup. Ref: https://pystan.readthedocs.io/en/latest/api.html#pystan.stan. Because `warmup_steps=None` is a bug previously, this change does not affect any user.

pyro/infer/mcmc/mcmc.py
```diff
@@ -275,12 +275,15 @@ class MCMCMarginals(Marginals):
     def diagnostics(self):
         if self._diagnostics:
             return self._diagnostics
+        support = self.support()
         for site in self.sites:
+            if self._trace_posterior.num_chains == 1:
+                site_support = support[site].unsqueeze(0)
```
fehiepsi: Bug: when `num_chains=1`, we have to create a new dimension to calculate diagnostics statistics.

Issue #1685, opened 2018-12-28
JIT trace does not work correctly with pyro.plate


fehiepsi:
I am not sure if this is a bug or I am using "pyro.plate" in a wrong way.

The following code shows the issue. I got this issue while trying to train a HMM model. Things work well for non-jit version. If we call `torch.autograd.grad(loss_jit(x), (x,))` two times, it will throw RuntimeError: "Can not go backward 2 times for the graph".

```
import pyro
import pyro.distributions as dist
import pyro.poutine as poutine
import torch

def model():
    with pyro.plate("plate", 2):
        pyro.sample("x", dist.Normal(0, 1))
    #pyro.sample("x", dist.Normal(0, 1).expand((2,)).to_event())  # this works!

trace = poutine.trace(model).get_trace()

def loss(x):
    trace.nodes["x"]["value"] = x
    return poutine.trace(poutine.replay(model, trace=trace)).get_trace().log_prob_sum()

x = torch.tensor([0., 1.], requires_grad=True)
loss_jit = torch.jit.trace(loss, x, check_trace=False)
print(loss_jit(x))  # -2.3379
print(loss_jit(x))  # -4.6758
```

fehiepsi:
Excellent! Thanks a lot @eb8680 for a clear explanation! Things work smoothly now. :)

fritzo:
Nice sleuthing, @eb8680! We should fix that line in `Trace.log_prob_sum()` to use `result = result + _` rather than `+=`.

fehiepsi:
Closed via #1687.

PR #1687, opened 2018-12-28
Fix initial trace get NaN energy

fehiepsi:
In initial_trace method, we only check for NaN of log_prob_sum in constrained space. Sometimes, after transformed to unconstrained space, latent variables will get NaN values, which in turn makes initial_trace's potential energy get NaN value.

This PR prevents that case by only selecting traces with non-NaN potential energy.

I also want to take this chance to incorporate the fix of @eb8680 for #1685.

### DIFF
--- a/pyro/poutine/trace_struct.py
+++ b/pyro/poutine/trace_struct.py
@@ -146,7 +146,7 @@ class Trace(networkx.DiGraph):
                     if is_validation_enabled():
                         warn_if_nan(log_p, "log_prob_sum at site '{}'".format(name))
                         warn_if_inf(log_p, "log_prob_sum at site '{}'".format(name), allow_neginf=True)
-                result += log_p
+                result = result + log_p
         return result
 
     def compute_log_prob(self, site_filter=lambda name, site: True):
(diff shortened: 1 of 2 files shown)


pyro/poutine/trace_struct.py
```diff
@@ -146,7 +146,7 @@ def log_prob_sum(self, site_filter=lambda name, site: True):
                     if is_validation_enabled():
                         warn_if_nan(log_p, "log_prob_sum at site '{}'".format(name))
                         warn_if_inf(log_p, "log_prob_sum at site '{}'".format(name), allow_neginf=True)
-                result += log_p
+                result = result + log_p
```
eb8680: This is probably the single most traversed line in the entire codebase, so we should make sure there's no performance impact.  cc @neerajprad

martinjankowiak: there also may be some reason this is in-place here. @fritzo is always grumbling about in-place ops, has seen these lines of code many times, so maybe there was some particular reason 🤔

neerajprad: Since `log_p` is scalar, perf impact shouldn't be a concern (we can check the hmc test runtimes). I think this should be safe, but it would be best to get a green light from @fritzo.

fritzo: I think it is a good idea to change this, and I don't foresee a performance impact.

fritzo:
LGTM on the changes to `trace_struct.py`, but I defer to @neerajprad for changes to `hmc.py`

PR #1688, opened 2018-12-29
Fix various docs format

fehiepsi:
As title.

### DIFF
--- a/docs/source/ops.rst
+++ b/docs/source/ops.rst
@@ -19,6 +19,12 @@ Utilities for HMC
     :show-inheritance:
     :member-order: bysource
 
+.. automodule:: pyro.ops.welford
+    :members:
+    :undoc-members:
+    :show-inheritance:
+    :member-order: bysource
+
 Newton Optimizers
 -----------------
 
@@ -38,3 +44,12 @@ Tensor Contraction
     :member-order: bysource
 
 .. autofunction:: pyro.ops.contract.ubersum
+
+Statistical Utilities
+---------------------
+
+.. automodule:: pyro.ops.stats
+    :members:
+    :undoc-members:
+    :show-inheritance:
+    :member-order: bysource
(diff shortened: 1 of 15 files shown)


pyro/infer/importance.py
```diff
@@ -2,7 +2,6 @@
 import torch
 import warnings
 
-from pyro.distributions.util import logsumexp
```
fehiepsi: This leads to a wrong display for logsumexp http://docs.pyro.ai/en/0.3.0-release/inference_algos.html#pyro.infer.importance.logsumexp

pyro/primitives.py
```diff
@@ -273,9 +273,9 @@ def module(name, nn_module, update_module_params=False):
 
 
 def random_module(name, nn_module, prior, *args, **kwargs):
-    """
+    r"""
     Places a prior over the parameters of the module `nn_module`.
-    Returns a distribution (callable) over `nn.Module`s, which
+    Returns a distribution (callable) over `nn.Module`\s, which
```
fehiepsi: This fixes the format at http://docs.pyro.ai/en/0.3.0-release/primitives.html#pyro.random_module. I don't know a better solution.

docs/source/ops.rst
```diff
@@ -38,3 +44,12 @@ Tensor Contraction
...
 .. autofunction:: pyro.ops.contract.ubersum
+
+Statistical Utilities
+---------------------
+
+.. automodule:: pyro.ops.stats
+    :members:
+    :undoc-members:
+    :show-inheritance:
+    :member-order: bysource
```
fehiepsi: Docs for stats are missing. There aren't docs for many other ops too but it might be an intention so I don't touch them.

fritzo:
Thanks for cleaning up docs, @fehiepsi !

PR #1689, opened 2018-12-29
Fix make doctest

fehiepsi:
It has been a while that I can't run doctest locally. Somehow we have to tell pytest to add current directory to PYTHONPATH to load plugin. Travis already did that at https://github.com/uber/pyro/blob/dev/.travis.yml#L7 so there is no error in Travis.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ scrub: FORCE
 	find tutorial -name "*.ipynb" | xargs python tutorial/source/cleannb.py
 
 doctest: FORCE
-	pytest -p tests.doctest_fixtures --doctest-modules -o filterwarnings=ignore pyro
+	python -m pytest -p tests.doctest_fixtures --doctest-modules -o filterwarnings=ignore pyro
 
 format: FORCE
 	isort --recursive *.py pyro/ examples/ tests/ profiler/*.py docs/source/conf.py


Makefile
```diff
@@ -22,7 +22,7 @@ scrub: FORCE
 	find tutorial -name "*.ipynb" | xargs python tutorial/source/cleannb.py
 
 doctest: FORCE
-	pytest -p tests.doctest_fixtures --doctest-modules -o filterwarnings=ignore pyro
+	python -m pytest -p tests.doctest_fixtures --doctest-modules -o filterwarnings=ignore pyro
```
neerajprad: @fehiepse - I have not faced this on linux or mac. Does installing using `pip install -e .` fix this issue? Your solution is fine - `python -m` will place the current directory on the path, but I wonder why I am not able to replicate it. Regardless, I see no harm in this change.

fehiepsi: When I created a new environment with only Pyro, doctest works fine. After debugging, I detect the reason is "I installed gpflow in the same environment with pyro". GPflow creates a `tests` folder in python `site-packages`, so pytest tried to find plugin `doctest_fixtures` in that folder instead of in `pyro/tests` folder.

You can quickly do `pip install gpflow --no-deps` to see this issue.

Because I have known the reason and I can fix that problem by removing that `tests` folder, I let it up to you to merge or close this PR. :)

neerajprad: Makes sense, thanks for explaining, @fehiepsi.

Issue #1683, opened 2018-12-22
Add docs for abstract_infer's methods


fehiepsi:
Currently, some methods in `abstract_infer` classes (and their derived classes) miss docs explaining its purpose (as reported in [forum](https://forum.pyro.ai/t/how-to-begin-learning-probabilistic-programming/519/3), so we need to add them. I collect some of them here for reference:

- [Marginals](http://docs.pyro.ai/en/0.3.0-release/inference_algos.html#pyro.infer.abstract_infer.Marginals) `empirical` and `support`.
- [TracePosterior](http://docs.pyro.ai/en/0.3.0-release/inference_algos.html#pyro.infer.abstract_infer.TracePosterior.marginal) marginal method.

PR #1691, opened 2018-12-29
Add docs for Abstract Infer classes

fehiepsi:
Addresses #1683. My last PR, I don't want to spam PR during the end of the year. Happy new year! :dancer:

### DIFF
--- a/pyro/infer/abstract_infer.py
+++ b/pyro/infer/abstract_infer.py
@@ -134,7 +134,7 @@ class Marginals(object):
 
         :param bool flatten: A flag to decide if we want to flatten `batch_shape`
             when the marginal distribution is collected from the posterior with
-            ``num_chains > 1``.
+            ``num_chains > 1``. Defaults to False.
         :returns: a dict with keys are sites' names and values are sites' supports.
         :rtype: :class:`OrderedDict`
         """
@@ -150,7 +150,7 @@ class Marginals(object):
     @property
     def empirical(self):
         """
-        A dictionary of sites' names and theirs corresponding :class:`EmpiricalMarginal`
+        A dictionary of sites' names and their corresponding :class:`EmpiricalMarginal`
         distribution.
 
         :type: :class:`OrderedDict`
--- a/pyro/infer/abstract_infer.py
+++ b/pyro/infer/abstract_infer.py
@@ -129,6 +129,15 @@ class Marginals(object):
                            for site in self.sites}
 
     def support(self, flatten=False):
+        """
+        Gets support of this marginal distribution.
+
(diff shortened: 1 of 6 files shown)


pyro/infer/abstract_infer.py
```diff
@@ -278,4 +302,7 @@ def _traces(self, *args, **kwargs):
             yield (replayed_trace, 0., 0)
 
     def marginal(self, sites=None):
+        """
+        Gets marginal distribution from posterior.
+        """
         return self.posterior.marginal(sites)
```
fehiepsi: not sure why we return marginal of posterior (not predictive) here so I leave it as is.

pyro/infer/abstract_infer.py
```diff
@@ -140,6 +149,12 @@ def support(self, flatten=False):
 
     @property
     def empirical(self):
+        """
+        A dictionary of sites' names and theirs corresponding :class:`EmpiricalMarginal`
```
jpchen: "their"

pyro/infer/abstract_infer.py
```diff
@@ -129,6 +129,15 @@ def _populate_traces(self, trace_posterior, validate):
                            for site in self.sites}
 
     def support(self, flatten=False):
+        """
+        Gets support of this marginal distribution.
+
+        :param bool flatten: A flag to decide if we want to flatten `batch_shape`
+            when the marginal distribution is collected from the posterior with
+            ``num_chains > 1``.
```
jpchen: "defaults to False"

pyro/infer/mcmc/mcmc.py
```diff
@@ -268,11 +268,23 @@ def _traces(self, *args, **kwargs):
...
+        :returns: A :class:`MCMCMarginals` class instance.
+        :rtype: :class:`MCMCMarginals`.
+        """
         return MCMCMarginals(self, sites)
 
 
 class MCMCMarginals(Marginals):
     def diagnostics(self):
+        """
+        Gets some diagnostics statistics such as effective number size and
```
jpchen: "effective sample size" is clearer imo

fehiepsi: Thanks @jpchen! I'll address them soon. :)

fritzo:
@eb8680 can you please confirm that we want to add docs to all these things? I recall you wanted to remove things, and we should avoid adding docs to any classes or methods planned for deprecation.

eb8680:
> confirm that we want to add docs to all these things

Since the methods documented in this PR are already in 0.3, I'm fine with merging this for now.  We can revisit later if we decide to do a cleanup release.

PR #1695, opened 2019-01-03
Reduce running time for nuts tests in Travis

fehiepsi:
All issues which we noticed related to jit have been solved before 0.3 release. This PR cleans up the `mark_jit` functionality. In addition, I fixed a mistake which I made in the last PR.


tests/infer/mcmc/test_hmc.py
```diff
@@ -16,18 +15,6 @@
 logger = logging.getLogger(__name__)
 
 
-def mark_jit(*args, **kwargs):
-    jit_markers = kwargs.pop("marks", [])
-    jit_markers += [
-        pytest.mark.skipif('CI' in os.environ,
-                           reason='slow test'),
-        pytest.mark.skipif('CUDA_TEST' in os.environ,
```
neerajprad: - Have we tested that these run (and finish within reasonable time) on CUDA?
 - I am also not sure if we should run all HMC JIT tests on CI. HMC tests even with the JIT flag are quite expensive. The additional signal by enabling the JIT flag is low (it covers the same code path essentially) and may not justify adding several minutes to unit test times. I think whitelisting one or two should be good enough for coverage, and we can change the reason above to something other than "slow test", if the slowness issues are fixed i.e. JIT is not slower than the non JIT test.

fehiepsi: Oppss... I have interpreted this marker wrongly. I thought they have no functionality now. >"<

pyro/infer/mcmc/mcmc.py
```diff
@@ -277,8 +277,9 @@ def diagnostics(self):
             return self._diagnostics
         support = self.support()
         for site in self.sites:
+            site_support = support[site]
             if self._trace_posterior.num_chains == 1:
-                site_support = support[site].unsqueeze(0)
+                site_support = site_support.unsqueeze(0)
```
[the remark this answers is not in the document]
fritzo: 1. Let's add a regression test for this.
2. Let's try to keep PRs focused on single goals.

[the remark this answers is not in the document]
fehiepsi: @fritzo I'll fix this mistake in another PR.

tests/infer/mcmc/test_hmc.py
```diff
@@ -263,7 +261,7 @@ def model(data):
                      jit_compile=jit, ignore_jit_warnings=True)
     mcmc_run = MCMC(hmc_kernel, num_samples=600, warmup_steps=200).run(data)
     posterior = mcmc_run.marginal("y_prob").empirical["y_prob"].mean
-    assert_equal(posterior, y_prob, prec=0.05)
+    assert_equal(posterior, y_prob, prec=0.06)
```
fehiepsi: @neerajprad this is the only failed test for hmc in CUDA

tests/infer/mcmc/test_nuts.py
```diff
@@ -279,7 +285,7 @@ def model(data):
     assert_equal(posterior.mean, y_prob, prec=0.05)
 
 
-@pytest.mark.parametrize("jit", [False, mark_jit(True)], ids=jit_idfn)
+@pytest.mark.parametrize("jit", [True], ids=jit_idfn)
```
fehiepsi: I don't see any test which is slower with `jit_compile=True`. And this is the slowest test on Travis, so we can put a test for jit here.

tests/infer/mcmc/test_nuts.py
```diff
@@ -175,7 +173,7 @@ def model(data):
     true_probs = torch.tensor([0.9, 0.1])
     data = dist.Bernoulli(true_probs).sample(sample_shape=(torch.Size((1000,))))
     nuts_kernel = NUTS(model, step_size, adapt_step_size, adapt_mass_matrix, full_mass)
-    mcmc_run = MCMC(nuts_kernel, num_samples=500, warmup_steps=100).run(data)
+    mcmc_run = MCMC(nuts_kernel, num_samples=400, warmup_steps=200).run(data)
```
fehiepsi: the test for `adapt_mass_matrix=True` and `adapt_step_size=False` is slow; so I increase warmup_steps to force NUTS learn a better mass_matrix.

fehiepsi:
I think that after #1696 is resolved, we can safely use `jit=True` to improve running time on Travis.

tests/infer/mcmc/test_nuts.py
```diff
@@ -230,7 +228,7 @@ def model(data):
     nuts_kernel = NUTS(model, jit_compile=jit, ignore_jit_warnings=True)
     mcmc_run = MCMC(nuts_kernel, num_samples=500, warmup_steps=200).run(data)
     posterior = mcmc_run.marginal(['alpha', 'beta']).empirical
-    assert_equal(posterior['alpha'].mean, true_alpha, prec=0.06)
+    assert_equal(posterior['alpha'].mean, true_alpha, prec=0.08)
```
fehiepsi: to pass CUDA test

tests/infer/mcmc/test_hmc.py
```diff
@@ -226,7 +224,7 @@ def model(data):
     assert_equal(posterior.mean, true_probs, prec=0.05)
 
 
-@pytest.mark.parametrize("jit", [False, mark_jit(True)], ids=jit_idfn)
+@pytest.mark.parametrize("jit", [True], ids=jit_idfn)
```
fehiepsi: this test is slow in Travis, so we only use `jit` version. We keep this pytest.mark to have a better message in Travis.

fehiepsi:
@fritzo @neerajprad this is ready to review. Travis slowest hmc/nuts tests should be under 20s now. We can improve it by marking `jit=True` for other tests but I think that it is not necessary.

tests/infer/mcmc/test_hmc.py
neerajprad: I think given that the JIT functionality is relatively stable now, it is okay to use it as a default option without having the test name indicate it as so. I don't feel about it strongly though, so would defer to your judgment.

Issue #1696, opened 2019-01-03
jit_compile in NUTS/HMC gives different results in GPU


fehiepsi:
I enable CUDA_TEST flag and observe that we get different results between jit and no-jit for the following models: `bernoulli_latent_model` and `gamma_beta`. In CPU, we get identical results. The following command shows the issue:
```
CUDA_TEST=1 PYRO_TENSOR_TYPE=torch.cuda.DoubleTensor pytest -vs tests/infer/mcmc/test_nuts.py -k gamma_beta
```

PR #1697, opened 2019-01-03
Add argument jit_optimize for hmc/nuts

fehiepsi:
This PR adds an optional `jit_optimize` flag to get a deterministic behaviour for using JIT in GPU. (refer #1696 for a context)

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -59,6 +59,8 @@ class HMC(TraceKernel):
         optimized executable trace in the integrator.
     :param bool ignore_jit_warnings: Flag to ignore warnings from the JIT
         tracer when ``jit_compile=True``. Default is False.
+    :param bool jit_optimize: Optional parameter denoting whether to enable
+        ``optimize`` flag for PyTorch JIT. Defaults to True.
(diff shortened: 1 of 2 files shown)


pyro/infer/mcmc/hmc.py
```diff
@@ -187,7 +191,8 @@ def compiled(*zi):
             return potential_energy
 
         with pyro.validation_enabled(False), optional(ignore_jit_warnings(), self._ignore_jit_warnings):
-            self._compiled_potential_fn = torch.jit.trace(compiled, vals, check_trace=False)
+            self._compiled_potential_fn = torch.jit.trace(compiled, vals, check_trace=False,
+                                                          optimize=self._jit_optimize)
```
fritzo: @neerajprad Do you think it would be more useful to add a `jit_options` dict and call
```py
torch.jit.trace(compiled, vals, check_trace=False, **jit_options)
```
This costs an extra level of indirection (`HMC(..., jit_options={'optimize': False})`) but will avoid duplication of possibly multiple future jit options in the `HMC` class.

neerajprad: +1. It will be much better to consolidate all JIT options into a `jit_options` dict!

PR #1698, opened 2019-01-03
Fix mcmc diagnostics method

fehiepsi:
This fixes a regression of my last PR. In addition, I add a test to verify that this method works as expected.

### DIFF
--- a/tests/infer/mcmc/test_mcmc.py
+++ b/tests/infer/mcmc/test_mcmc.py
@@ -34,7 +34,7 @@ class PriorKernel(TraceKernel):
 
 def normal_normal_model(data):
     x = pyro.param('loc', torch.tensor([0.0]))
-    y = pyro.sample('x', dist.Normal(x, torch.tensor([1.0])))
+    y = pyro.sample('y', dist.Normal(x, torch.tensor([1.0])))
     pyro.sample('obs', dist.Normal(y, torch.tensor([1.0])), obs=data)
     return y
 
@@ -49,6 +49,16 @@ def test_mcmc_interface():
     sample_std = marginal.variance.sqrt()
     assert_equal(sample_mean, torch.tensor([0.0]), prec=0.08)
     assert_equal(sample_std, torch.tensor([1.0]), prec=0.08)
+
+
+@pytest.mark.parametrize("num_chains", [1, 2])
+def test_mcmc_diagnostics(num_chains):
+    data = torch.tensor([1.0])
+    kernel = PriorKernel(normal_normal_model)
+    mcmc = MCMC(kernel=kernel, num_samples=10, num_chains=1).run(data)
+    diagnostics = mcmc.marginal(["y"]).diagnostics()
+    assert diagnostics["y"]["n_eff"].shape == torch.Size([1])
+    assert diagnostics["y"]["r_hat"].shape == torch.Size([1])
 
 
 @pytest.mark.parametrize("num_chains, cpu_count", [
--- a/pyro/infer/mcmc/mcmc.py
+++ b/pyro/infer/mcmc/mcmc.py
@@ -277,8 +277,9 @@ class MCMCMarginals(Marginals):
             return self._diagnostics
         support = self.support()
         for site in self.sites:
+            site_support = support[site]
             if self._trace_posterior.num_chains == 1:
-                site_support = support[site].unsqueeze(0)
+                site_support = site_support.unsqueeze(0)
             site_stats = OrderedDict()
             try:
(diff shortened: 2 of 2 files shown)


tests/infer/mcmc/test_mcmc.py
```diff
@@ -51,6 +51,16 @@ def test_mcmc_interface():
     assert_equal(sample_std, torch.tensor([1.0]), prec=0.08)
 
 
+@pytest.mark.parametrize("num_chains", [1, 2])
+def test_mcmc_diagnostics(num_chains):
+    data = torch.tensor([1.0])
+    kernel = PriorKernel(normal_normal_model)
+    mcmc = MCMC(kernel=kernel, num_samples=10, num_chains=1).run(data)
```
fehiepsi: oppss, `num_chains=num_chains`, my brain is foggy recently I guess >"< srrrrrrrrrrrr

neerajprad:
> the test will fail with a weird error relating pytorch multiprocessing.

That's odd..does it fail on the CPU as well?

fehiepsi:
Yes, it failed in CPU (in my system). I just set `num_chains=num_chains` and I expect that you can see the error message in Travis.

fehiepsi:
@neerajprad this is not relevant to the above issue: when I replace `x = pyro.param(...)` by `x = tensor(...)`, I get the error `local variable `val` referenced before assignment at [this line](https://github.com/uber/pyro/blob/dev/pyro/infer/mcmc/mcmc.py#L165). I guess that I have triggered [this line](https://github.com/uber/pyro/blob/dev/pyro/infer/mcmc/mcmc.py#L160).

tests/infer/mcmc/test_mcmc.py
```diff
@@ -51,6 +51,16 @@ def test_mcmc_interface():
     assert_equal(sample_std, torch.tensor([1.0]), prec=0.08)
 
 
+@pytest.mark.parametrize("num_chains", [1, 2])
```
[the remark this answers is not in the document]
fehiepsi: Nice observation! That's why I don't get any error in Travis with last PR. I'll make pytest marks here and open an issue for the error which we caught in this PR.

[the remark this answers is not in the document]
neerajprad: @fehiepsi - Yes, for as long as the test passes locally, it is fine. You can add a skipif marker here for travis. Btw..this isn't a bug, so no issue is required, it is just a check within the MCMC class itself so that we don't dispatch to multiple chains and become slower. Since travis machines do not have multiple cores, it is fine to do any testing for multiple chains locally. We we have a real need for testing parallel chains in travis, we can add additional arguments to force dispatch even when `num_chains > num_cpus` later.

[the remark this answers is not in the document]
fehiepsi: Agree, we don't need to test multi-chain in Travis.

tests/infer/mcmc/test_mcmc.py
```diff
@@ -51,6 +55,20 @@ def test_mcmc_interface():
     assert_equal(sample_std, torch.tensor([1.0]), prec=0.08)
 
 
+@pytest.mark.parametrize("num_chains", [
+    1,
+    pytest.param(2, marks=[pytest.mark.skipif("CI" in os.environ, reason="CI only provides 1 CPU"),
+                           pytest.mark.skip(reason="https://github.com/uber/pyro/issues/1699")])
```
neerajprad: Let us not skip this for local testing. Alternatively, we can also capture the UserWarning if "CI" is in the environment.

fehiepsi: In my system, the queue stuck after 18 steps (there are totally 20 steps: 10 for warmups and 10 for samples), and there is no error raised. I have to mark this `skip` to not affect other people. I think we can remove it later when #1699 is resolved. It can be that the problem just happens in my system (despite that I have restarted it).

neerajprad: Thanks for explaining. I think we should resolve this separately. For some reason, I am getting an assertion error even with `num_chains = 1`

```
>       assert diagnostics["y"]["n_eff"].shape == torch.Size([1])
E       assert torch.Size([]) == torch.Size([1])
E         Right contains more items, first extra item: 1
E         Full diff:
E         - torch.Size([])
E         + torch.Size([1])
E         ?             +

tests/infer/mcmc/test_mcmc.py:68: AssertionError
```

neerajprad: Okay, I see the issue. When MKL isn't available, we return `torch.tensor(float('nan'))`. Let us change that to return the correctly shaped NaN tensor.

PR #1704, opened 2019-01-07
Import pyro.optim in gp_bayes_opt

### LINKED ISSUE #1684: make test: no attribute 'optim' error in 'examples/contrib/oed/ab_test.py'
### Issue Description
On the latest dev branch, `make test` gives the following error:

_
examples/contrib/oed/ab_test.py:12: in <module>
    from gp_bayes_opt import GPBayesOptimizer
examples/contrib/oed/gp_bayes_opt.py:11: in <module>
    class GPBayesOptimizer(pyro.optim.multi.MultiOptimizer):
E   AttributeError: module 'pyro' has no attribute 'optim'

### Environment
 - OS and python version: CentOS Linux 7 (Core); Python 3.7.1
 - PyTorch version, or if relevant, output of `pip freeze`: PyTorch 1.0.0
 - Pyro version: output of `python -c 'import pyro; print pyro.__version__'`: pyro 0.3.0+9adbdb7

### Code Snippet

```
make install
make format
make test
```

neerajprad:
Fixes #1684. 

Note that the pattern `import pyro; adam=pyro.optim.Adam()` cannot be used. Instead we should use `import pyro.optim; adam=pyro.optim.Adam()`. This is because `optim` is not loaded by default in `pyro.__init__.py`.


neerajprad:
We will need to update pytest, which is needed by pytest-benchmark used for perf regression testing. Will do so in another PR.

Issue #1707, opened 2019-01-08
[FR] Conjugate distributions


fritzo:
It would be nice to have some basic conjugate distributions. These are especially useful when all local variables can be integrated out. Note that samplers can use the existing distributions, and we need only implement some new math for the `.log_prob()` methods.

It would also be nice to be able to sample the marginalized variable from its complete conditional, e.g.
```py
def sample_hidden(self, value, sample_shape=()):
    # sample hidden variable conditioned on params and value
```

In the long-term we'd prefer to automatically detect conjugate relationships from programs without users needing to manually specify conjugate-pair distributions (as pointed out by @martinjankowiak). Thus the pairs in this issue can serve as short-term implementations for numerical testing, and we can work on better conjugacy detection that compiles down to these pairs.

## Tasks
- [x] #1708 `BetaBinomial` (needed by @fritzo, @neerajprad )
- [x] `GammaPoisson`
- [x] #1734 `DirichletMultinomial`
- [others](https://en.wikipedia.org/wiki/Conjugate_prior)?

vishwakftw:
Could I work on NormalNormal?

vishwakftw:
Thank you @fritzo. I will have an implementation available in a few hours for the `loc` conjugated normal.

fritzo:
Closing this issue until we work out design issues in #1723 .

Here are the reasons I am closing this issue:
1. We're still working out the conjugacy interface, and we welcome contributions to the [design doc](https://docs.google.com/document/d/1625HLlJUyaNxQRng3LsixVFqOuUYkK8S6otGVgYEhos/edit#), which is still in progress.
2. We're implementing [a library](https://github.com/pyro-ppl/funsor) to enable [Birch](https://birch-lang.org)-style [delayed sampling](https://arxiv.org/abs/1708.07787) in Pyro. That library will eventually wrap implementations of conjugate math in Pyro and PyTorch, but we'd like to implement more of that library so we know exactly what interfaces to create.
3. Regarding specific distributions, the `BetaBinomial`, `DirichletMultinomial`, and `GammaPoisson` have the property that multiple observations can be trivially combined into a single observation (by adding the observations and adding `total_count`s). Other distributions have nontrivial sufficient statistics, and e.g. `NormalNormal`. It may make sense for e.g. `NormalNormal` to be a multivariate distribution with event dim.
[...]

PR #1708, opened 2019-01-09
Implement BetaBinomial distributions

fritzo:
Addresses #1707 
Joint work with @neerajprad .

This implements a `BetaBinomial` distribution that analytically integrates out the `probs` in a `Beta->Binomial` pair.

## Questions for reviewers

- [x] When we broadcast `value` in `log_prob()`, should the hidden variable be shared along the broadcast dimension? Similarly, when we pass a `sample_shape` to `.sample()`, should this broadcast the hidden variable, or only the final `Binomial` value?

## Tested

- [x] `log_prob` (since scipy doesn't implement it, added a basic check to ensure that log prob values over support add up to 1.)
- [x] `sample`
- [x] `mean`
- [x] `variance`
- [x] `expand`


neerajprad:
> When we broadcast value in log_prob(), should the hidden variable be shared along the broadcast dimension? Similarly, when we pass a sample_shape to .sample(), should this broadcast the hidden variable, or only the final Binomial value?

The current behavior seems intuitive to me - i.e. broadcasting both the hidden variable and the Binomial together. 

How do you foresee this being used in SVI and MCMC? If the users are interested in the distribution of the hidden variables (for MCMC), do you think we should provide convenience utilities to populate these directly from the posterior distribution into the trace? I don't see a clean way of doing this at the moment.

fritzo:
> How do you foresee this being used in SVI and MCMC?

I think we can add a method `.sample_hidden()` (whose interface is described in a base class `CollapsedDistribution`). We already follow this pattern with collapsed discrete distributions:
1. `@config_infer` to convert a model to collapsed form
2. Run SVI blind to the collapsed variables
3. Use `infer_discrete(replay(model, guide_trace))` to query the collapsed variables.

(Actually I'm unclear how to perform 3. using MCMC)

pyro/distributions/conjugate.py
```diff
@@ -0,0 +1,64 @@
...
+        batch_shape = torch.Size(batch_shape)
+        new._beta = self._beta.expand(batch_shape)
+        new.total_count = self.total_count.expand_as(new._beta.concentration0)
+        super(BetaBinomial, new).__init__(batch_shape, validate_args=False)
+        new._validate_args = self._validate_args
+        return new
+
+    def sample(self, sample_shape=()):
+        probs = self.Beta.sample(sample_shape)
+        return Binomial(probs).sample()
```
fehiepsi: Though I don't know if using Conjugate distributions will be helpful at all in HMC context, I would like to suggest a way to query collapsed variables: we return here a sample with a specific attribute `_collapsed_sample` which stores the sample from Beta distribution (which is `probs`). A specific method `poutine.get_uncollapsed_trace` will be used instead of `poutine.get_trace` when users want to get the samples from Beta distribution. In that method, we add an additional node for this collapsed site. WDYT?

neerajprad:
@fritzo - I am fine merging this. Let me know if this seems reasonable. I would like to add gamma-poisson, and follow that up with some other changes that make it usable inside MCMC.

PR #1709, opened 2019-01-10
add second-order hmm to hmm example

martinjankowiak:


examples/hmm.py
```diff
@@ -22,6 +22,7 @@
 
 import torch
 import torch.nn as nn
+from torch.distributions import constraints
```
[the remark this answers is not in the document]
fritzo: Hmm, I guess since the expectation is inside the log, this might be better called "collapsed SVI" or "collapsed SGVB". There are no sampled latent variables, and the only stochasticity is in data subsampling. @martinjankowiak I defer to your judgement as to what to call these algorithms.

tests/test_examples.py
```diff
@@ -40,6 +40,8 @@
     'hmm.py --num-steps=1 --truncate=10 --model=3',
     'hmm.py --num-steps=1 --truncate=10 --model=4',
     'hmm.py --num-steps=1 --truncate=10 --model=5',
+    'hmm.py --num-steps=1 --truncate=10 --model=6',
```
fritzo: Could you also add these to the `CUDA_EXAMPLES` below?

examples/hmm.py
```diff
@@ -509,5 +569,6 @@ def main(args):
     parser.add_argument("-p", "--print-shapes", action="store_true")
     parser.add_argument('--cuda', action='store_true')
     parser.add_argument('--jit', action='store_true')
+    parser.add_argument('--raftery-parameterization', action='store_true')
```
fritzo: Could you add an alias `-r` or `-rp`?
```py
parser.add_argument('-r', '--raftery-parameterization', action='store_true')
```

tests/test_examples.py
martinjankowiak: done

examples/hmm.py
martinjankowiak: done

PR #1712, opened 2019-01-18
Add artwork/ folder containing logos

fritzo:
This adds Pyro logos to the repo, to make it easier to find canonical logos for talks, etc.

Files were created by @ibrahimhaddad who suggested we add them to the repo.


fritzo:
@briankhsieh suggests instead creating this as a repo under a new org. Let's wait to do that until we settle on a new org.

PR #1713, opened 2019-01-19
Adding Gamma-Poisson distribution

neerajprad:
Addresses #1707.

Adds a gamma-poisson distribution, building on top of #1708. Tests are similar to the ones added in #1708.


pyro/distributions/conjugate.py
```diff
@@ -85,3 +85,59 @@ def enumerate_support(self, expand=True):
...
+        return new
+
+    def sample(self, sample_shape=()):
+        rate = self._gamma.sample(sample_shape)
+        return Poisson(rate).sample()
+
+    def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
+        return torch.lgamma(self.concentration + value) + self.concentration * self.rate.log() \
```
fritzo: Hmm I'm confused because usually for the Gamma distribution, `beta=rate` and `alpha=concentration`, whereas [this reference](http://www.math.wm.edu/~leemis/chart/UDR/PDFs/Gammapoisson.pdf) suggests we should have
```py
lgamma(rate + value) + value * concentration.log() \
- lgamma(rate) - (rate + value) * (1 + concentration).log() \
- lgamma(value + 1)
```
Are you sure you haven't swapped concentration <-> rate?

neerajprad: They use some tricky notation there, which tripped me up too! 😄 

Here is their reference to the [gamma distribution](http://www.math.wm.edu/~leemis/chart/UDR/PDFs/Gamma.pdf) which should clarify the relationship with the parameters for `torch.distributions.gamma`.
 - pdf - shape parameter beta = torch.dist - concentration
 - pdf - scale parameter alpha = torch.dist - 1 / rate

neerajprad:
>  I suggest at least noting that in the docstring, and mentioning the relationship between the two parametrizations.

Will add to the docstring. Thanks for reviewing!

pyro/distributions/conjugate.py
fritzo: "tricky" is being generous

fehiepsi: Wikipedia has a good [reference](https://en.wikipedia.org/wiki/Conjugate_prior) for conjugate prior. According to it, this formula is correct. I think that we should add some reference here to avoid confusion. And I prefer to group `lgamma` operators together (called it `binomial_coef` e.g.) to make it easier to verity this formula, but this is just a nit.

fritzo: :+1: to using `_log_beta()` to group `lgamma` operations, since that's already defined in this file.

neerajprad: The `lgamma` operations are grouped in `log_beta`, but note that I had to take out a `(self.concentration + value)` term in the denominator (due to `lgamma(x + 1)` in the original), so this might not be that much clearer.

neerajprad:
> Note that GammaPoisson is equivalent to (a different parametrization of) NegativeBinomial

Done.

pyro/distributions/conjugate.py
```diff
@@ -85,3 +85,63 @@ def enumerate_support(self, expand=True):
...
+
+class GammaPoisson(TorchDistribution):
+    r"""
+    Compound distribution comprising of a gamma-poisson pair, also referred to as
+    a gamma-poisson mixture. The ``rate`` parameter for the
+    :class:`~pyro.distributions.Poisson` distribution is unknown and randomly
+    drawn from a :class:`~pyro.distributions.Gamma` distribution.
+
+    .. note:: This can be treated as an alternate parametrization of the
+        :class:`~pyro.distributions.NegativeBinomial` (``total_count``, ``probs``)
```
fritzo: I don't think `NegativeBinomial` is implemented yet. However I think it's worth mentioning the relation to `NegativeBinomial` (or implementing it as a simple wrapper) so that people can find it rather than needing to reimplement it.

pyro/distributions/conjugate.py
```diff
@@ -85,3 +85,63 @@ def enumerate_support(self, expand=True):
...
+        return new
+
+    def sample(self, sample_shape=()):
+        rate = self._gamma.sample(sample_shape)
+        return Poisson(rate).sample()
+
+    def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
+        return -_log_beta(self.concentration, value + 1) - (self.concentration + value).log() + \
```
fritzo: nit: eliminate the common subexpression?
```py
post_value = self.concentration + value
return (-_log_beta(self.concentration, value + 1) - post_value.log() +
        self.concentration * self.rate.log() - post_value * (1 + self.rate).log())
```

pyro/distributions/conjugate.py
neerajprad: [Negative Binomial](https://pytorch.org/docs/stable/_modules/torch/distributions/negative_binomial.html) is implemented in `torch.distributions`! Do you think we should call that internally instead?

fritzo: Any idea why it doesn't appear in our docs? Even in dev http://docs.pyro.ai/en/dev/distributions.html

Re-implementing `GammaPoisson` is fine by me.

neerajprad: @jpchen is just updating our rtd build so it should show up soon.

fritzo: Thanks for explaining.

fehiepsi:
I guess this is an accident by github, so I reopen this PR.

neerajprad:
> I guess this is an accident by github, so I reopen this PR.

Haha..it seems `resolves` also closes other PRs.

PR #1714, opened 2019-01-22
Add mixed effect HMM experiment code

eb8680:
cc @martinjankowiak @fritzo


eb8680:
Oops, changing base to my fork to avoid spam

PR #1716, opened 2019-01-23
Uncollapse compound distributions in traces generated from MCMC

neerajprad:
**NOTE:** This is a sketch, and we will need to iron out the API issues before this can be merged.

This adds the ability to collapse latent sites in compound distributions (i.e. latent observed conjugate pair) during inference in HMC, and uncollapse it later by sampling from the exact posterior distribution followed by an observed site with the sampled parameters. Note that this requires the user to use compound distributions directly in the model. Later, when we can do conjugacy detection via `config_infer`, this can be relaxed.

**Design**: The original implementation could not handle latents within arbitrary plates, which has been rectified now. This has however added some complexity to these classed due to which I have moved them to `contrib`.  
 - We are passing in the name of the site to the constructor for the distribution, and have two methods, `pin_latent` and `observe` to specify the location for the collapsed latent site and the observed site. e.g. below. Note that the reason to pass in the name is to be able to do site matching later when sampling from the posterior, but an alternate design might be for the user to specify the latent and observed sample in their models with the requirement that they use a pre-established naming convention. Another complication with this design choice would be dynamically adjusting the shape of the latent (which is currently happening in `.pin_latent()`). EDIT: I think the latter is a much better design choice though, and I am exploring work-arounds to not have to pass in name into the constructor.

   ```python
   b = BetaBinomial("betabinom", ...)
   with pyro.plate("dim0", ...):
       b.pin_latent()  # Adds a collapsed site named "betabinom.latent"
       ...
       with pyro.plate("data", ...):
           b.observe(data)  # Add a sample(observed) site called "betabinom"
   ```
 - When we call `.pin_latent()`, the latent size is adjusted so that all the other methods like `.sample` and `.log_prob` use the correctly shaped latent distribution.

**Changes:**
 - Introduces `_posterior_latent` and `_posterior_predictive` methods to conjugate distributions - the former is the posterior distribution of the parameter values given the observation, and the latter takes in parameters sampled from the posterior to give the posterior predictive distribution. Note that this uncollapsing will be done by the user outside of HMC, to retrieve samples from the  posterior distribution of the (originally collapsed) latent site. 
 - Minor changes to the HMC class so that in case the model only comprises of compound distributions (i.e. there are only observe statements), we do not fail but return a deterministic trace.
 - An `infer_conjugate` poutine that collapses latent sites with the infer kwarg `collapse=True`, , and uncollapses these by sampling them from the exact posterior when the model is wrapped under `infer_conjugate` poutine.


neerajprad:
Please hold off on reviewing this - after discussion with @fritzo, we concluded that we will need more information in our compound distributions to infer the shape of the latent, and this implementation isn't general in that it assumes that the latent shape is the shape of the original parameters that was used to construct the distribution.

neerajprad:
@fritzo - This has been updated with the changes that I mentioned that allow for sampling latent values inside arbitrary plates, but this has forced me to make certain design decisions that don't align well with the rest of distributions. I plan to use this branch for the MOBB model, but would like to discuss if you see a way we can integrate this into `distributions/conjugate.py`.

neerajprad:
I will put up a new PR with a simplified interface for discussion purpose.

PR #1718, opened 2019-01-24
Add adjoint ubersum script to examples/

fritzo:
This adds a simple `ubersum` example demonstrating how to use adjoint algorithms via `backend=...`, `require_backward()`, `._pyro_backward()`, and `._pyro_backward_result`.

@neerajprad this is a simplified and documented version of our script from https://github.com/uber/pyro/tree/profile-ubersum .

## Tested

- added to `test_examples.py` for both CPU and CUDA. CPU tests take under 5 seconds.


fritzo:
@justinchiu here's how to use the adjoint interface to compute MAP estimates and samples.

martinjankowiak:
should we consider renaming this now that we're exposing it in a more user-facing way?

fritzo:
> @martinjankowiak renaming this

We could rename `ubersum` to `einsum` and rename `batch_dims` to `plates`? (Of course keeping the old version for backwards compatibility.) Or did you mean renaming the name of the example script?

martinjankowiak:
the former. we should especially do plate dims for consistency with the arxiv paper that will be up sooner or later

examples/ubersum.py
```diff
@@ -0,0 +1,192 @@
...
+    return _jit_adjoint(equation, *operands, backend='pyro.ops.einsum.torch_sample', **kwargs)
+
+
+def jit_marginal(equation, *operands, **kwargs):
+    return _jit_adjoint(equation, *operands, backend='pyro.ops.einsum.torch_marginal', **kwargs)
+
+
+def time_fn(fn, equation, *operands, **kwargs):
+    iters = kwargs.pop('iters')
+    _CACHE.clear()  # Avoid memory leaks.
```
neerajprad: Just to clarify from our discussion yesterday - this isn't just to avoid memory leaks, but to get the correctly compiled function for different plate sizes too, right?

fritzo: Clearning was only needed for correctness by `naive_ubersum` in our discussion yesterday. non-naive Ubersum never slices things up, and so still be correct for different sized tensors.

fritzo:
Ok, I'll change `batch_dims` to `plates`. I think it's still a bit early to rename `ubersum` to `einsum` because`ubersum` uses the `torch_log` backend by default and doesn't even implement the usual linear backend and `ubersum` is only implemented for `modulo_total=True`. @martinjankowiak Is that ok?

fritzo:
@neerajprad I'd rather rename `batch_dims` -> `plates` in a follow-up PR. As I mentioned, I'm hesitant to rename `ubersum` -> `einsum` at this point.

PR #1720, opened 2019-01-25
Rename ubersum -> einsum

fritzo:
Addresses discussion in #1718 

This PR:
- renames `ubersum` to `einsum`
- renames the `batch_dims` kwarg to `plates`
- renames internal comments and classes to reflect the new naming convention
- adds a `LinearRing` so that `einsum` can default to `backend='torch'`, making it almost backwards compatible with `torch.einsum` for non-plated inputs. Note that `pyro.ops.contract.einsum` is not backwards compatible with `torch.einsum` because the former returns a tuple of tensors whereas the latter returns a single tensor.

An `ubersum` with the old behavior is preserved for backwards compatibility, but raises `DeprecationWarning`.

## Tested

- added a test of linear vs log ring
- added a linear example to the profiling script examples/einsum.py (~10x faster than log backend)
- most tests continue to use the old `ubersum` wrapper


fritzo:
@martinjankowiak @eb8680 are you ok with `einsum` as the name?

martinjankowiak:
:+1:

neerajprad:
Merging this as @eb8680 expressed his approval on chat!

Issue #1717, opened 2019-01-24
Bug in Empirical.sample() with multiple MCMC chains 


adam-coogan:
It seems like the marginals for MCMC runs with multiple chains are not being computed correctly. Here is a simple example:
```python
pyro.set_rng_seed(18)
def model(obs=None):
    a = pyro.sample("a", dist.Normal(0., 1.))
    pyro.sample("y", dist.Normal(a**2, 1.), obs=obs)

y_obs = torch.tensor(1.)
kernel = NUTS(model)
# kernel = HMC(model)  # doesn't work with this kernel either
mcmc_run = MCMC(kernel, num_samples=20, warmup_steps=1, num_chains=2).run(y_obs)

posterior = mcmc_run.marginal("a").empirical["a"]
# posterior = EmpiricalMarginal(mcmc_run, 'a')  # this doesn't work either
posterior.mean
posterior.sample()

>>> ...
>>> RuntimeError: index 19 is out of bounds for dim with size 2
```
When I look at `posterior._samples`, it contains a tensor of size `[2, 20]`. Calling `posterior.sample()` generates a number between 0 and 19, which leads to the error. I'd think `MCMC.marginal()` should concatenate the results from the chains together. Is the right place to do this `_ParallelSampler.terminate()`?

Environment info:
 - macOS Mojave 10.14.2 (18C54)
 - python 3.6.5
 - torch.__version__ is 1.0.0
 - pyro.__version__ is 0.3.0+0830f7ed

neerajprad:
Thanks for reporting, this is indeed a bug. It should pull out `[0, 2]` and `[1, 20]` entries from the sample tensor instead, i.e. a sample from each of the chains. I will push a fix for this, and additionally would like to revisit some of the assumptions we made to ensure that vectorized samples are fully supported by all methods.

neerajprad:
Should be fixed in #1721.

PR #1721, opened 2019-01-29
Fix bug in vectorized empirical.sample()

neerajprad:
Addresses #1717. 

This fixes a bug in `EmpiricalDistribution.sample()` with vectorized samples. This will be the case when we have samples from multiple chains in MCMC. Note that this case is limited to having samples of shape `[num_chains, num_samples]` but the class can otherwise handle arbitrary vectorization except the `log_prob` method which can't score broadcasted samples.

 - Added documentation on how vectorization is handled by the class.
 - Added some more tests, and hardened our existing test to catch such erroneous behavior.

### DIFF
--- a/pyro/distributions/empirical.py
+++ b/pyro/distributions/empirical.py
@@ -37,7 +37,7 @@ class Empirical(TorchDistribution):
     >>> with pyro.validation_enabled():
     ...     emp_dist.log_prob(batch_sample).shape
     Traceback (most recent call last):
-     ...
+    ...
     ValueError: ``value.shape`` must be torch.Size([2, 10])
 
     :param torch.Tensor samples: samples from the empirical distribution.
--- a/pyro/distributions/empirical.py
+++ b/pyro/distributions/empirical.py
@@ -1,8 +1,5 @@
 from __future__ import absolute_import, division, print_function
 
-import operator
-
-from six.moves import reduce
 import torch
 from torch.distributions import constraints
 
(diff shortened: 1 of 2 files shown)


pyro/distributions/empirical.py
```diff
@@ -47,8 +59,23 @@ def sample_size(self):
...
+        sample_idx = self._categorical.sample([num_samples])
+        for _ in range(len(self._samples.shape) - sample_idx.dim()):
+            sample_idx = sample_idx.unsqueeze(-1)
+        sample_idx = sample_idx.permute(dim_order)
+        sample_idx = sample_idx.expand(self.batch_shape + torch.Size([-1]) + self.event_shape)
+        samples = self._samples.gather(self._aggregation_dim, sample_idx)
+        # At this point, samples have `num_samples` values at the aggregation dim.
+        # Permute the ordering (and reshape) so that `sample_shape` is leftmost.
+        # dim_order.insert(self._aggregation_dim, dim_order.pop(0))
+        return samples.permute(dim_order).reshape(sample_shape + self.batch_shape + self.event_shape)
```
fehiepsi: I think that the following version is simpler:
```
idx = self._categorical.sample(sample_shape)  # sample_shape x batch_shape
# correct samples' shape: batch_shape x num_samples x event_shape -> num_samples x batch_shape x event_shape
samples = empirical._samples.unsqueeze(0).transpose(0, self._aggregation_dim + 1).squeeze(self._aggregation_dim + 1)
# make idx's shape compatible to samples' shape
idxs = idx.reshape((-1,) + empirical.batch_shape + (1,) * len(empirical.event_shape))
idxs = idxs.expand((-1,) + empirical.batch_shape + empirical.event_shape)
return samples.gather(0, idxs).reshape(sample_shape + samples.shape[1:])
```

pyro/distributions/empirical.py
```diff
@@ -11,12 +14,21 @@
...
+    requirement for `log_weights` is that its leftmost shape must match that of
+    `samples`. Samples are aggregated along the ``aggregation_dim``, which is the
+    rightmost dim of `log_weights`.
+
+    e.g. If ``samples.shape = torch.Size([2, 3, 10])`` and
+    ``log_weights.shape = torch.Size([2, 3])``, the second dimension corresponds
+    to the `aggregation_dim`. The distribution's `batch_shape` is ``torch.Size([2])``
+    and its `event_shape` is ``torch.Size([10])``. While sampling, we generate
+    a batch of random indices amongst ``[0, 1, 2]``, which are used to index
+    into the aggregation dim to return samples of shape ``torch.Size([2, 10])``.
```
fehiepsi: I think that it will be clearer if we add comments for an example code here.

pyro/distributions/empirical.py
```diff
@@ -11,12 +14,21 @@
 @copy_docs_from(TorchDistribution)
 class Empirical(TorchDistribution):
     r"""
-    Empirical distribution associated with the sampled data.
+    Empirical distribution associated with the sampled data. Note that the shape
+    requirement for `log_weights` is that its leftmost shape must match that of
```
fehiepsi: nit: its shape must match the leftmost shape of

neerajprad: Good catch, will fix.

pyro/distributions/empirical.py
```diff
@@ -11,12 +11,38 @@
...
+    >>> batch_sample.shape
+    torch.Size([100, 2, 10])
+
+    >>> emp_dist.log_prob(single_sample).shape
+    torch.Size([2])
+    >>> # Vectorized samples cannot be scored by log_prob.
+    >>> with pyro.validation_enabled():
+    ...     emp_dist.log_prob(batch_sample).shape
+    Traceback (most recent call last):
+     ...
```
fehiepsi: I am not sure if 5 spaces here is right.

neerajprad: Do you mean the `emp_dist` line? That has 4 spaces if you count w.r.t. the with block.

fehiepsi: I mean the line after Traceback. It has 5 spaces before "..."

PR #1729, opened 2019-02-01
Implement sampling for von Mises distribution based on Best-Fisher algorithm

ahmadsalim:
Hi Pyro Devs!

I have tried to implement sampling for the von Mises distribution based on the algorithm by Best and Fisher that is referenced in the comments.
I have marked this as WIP since I have two aspects I would like to improve on:
* The algorithm is a rejection sampling based one, but it seems that I can't make it fit `Rejector` obviously, since the probability of the resulting sample being accepted is defined independently from the sample itself. Did I miss something that could have made this possible?
* I would like to test the sampling part of the distribution, do you have any recommendations?

Thanks!

### DIFF
--- a/pyro/distributions/von_mises.py
+++ b/pyro/distributions/von_mises.py
@@ -17,22 +17,31 @@ def _eval_poly(y, coef):
     return result
 
 
-_COEF_SMALL = [1.0, 3.5156229, 3.0899424, 1.2067492, 0.2659732, 0.360768e-1, 0.45813e-2]
-_COEF_LARGE = [0.39894228, 0.1328592e-1, 0.225319e-2, -0.157565e-2, 0.916281e-2,
-               -0.2057706e-1, 0.2635537e-1, -0.1647633e-1,  0.392377e-2]
+_I0_COEF_SMALL = [1.0, 3.5156229, 3.0899424, 1.2067492, 0.2659732, 0.360768e-1, 0.45813e-2]
+_I0_COEF_LARGE = [0.39894228, 0.1328592e-1, 0.225319e-2, -0.157565e-2, 0.916281e-2,
+                  -0.2057706e-1, 0.2635537e-1, -0.1647633e-1,  0.392377e-2]
+_I1_COEF_SMALL = [0.5, 0.87890594, 0.51498869, 0.15084934, 0.2658733e-1, 0.301532e-2, 0.32411e-3]
+_I1_COEF_LARGE = [0.39894228, -0.3988024e-1, -0.362018e-2, 0.163801e-2, -0.1031555e-1,
+                  0.2282967e-1, -0.2895312e-1, 0.1787654e-1, -0.420059e-2]
 
 
-def _log_modified_bessel_fn_0(x):
+def _log_modified_bessel_fn(x, order=0):
     """
-    Returns ``log(I0(x))`` for ``x > 0``.
+    Returns ``log(I_order(x))`` for ``x > 0``,
+    where `order` is either 0 or 1.
     """
+    assert order == 0 or order == 1
+
+    small_coeff = [_I0_COEF_SMALL, _I1_COEF_SMALL]
+    large_coeff = [_I0_COEF_LARGE, _I1_COEF_LARGE]
+
     # compute small solution
     y = (x / 3.75).pow(2)
-    small = _eval_poly(y, _COEF_SMALL).log()
+    small = _eval_poly(y, small_coeff[order]).log()
 
     # compute large solution
     y = 3.75 / x
-    large = x - 0.5 * x.log() + _eval_poly(y, _COEF_LARGE).log()
+    large = x - 0.5 * x.log() + _eval_poly(y, large_coeff[order]).log()
 
     mask = (x < 3.75)
     result = large
@@ -65,6 +74,10 @@ class VonMises(TorchDistribution):
         batch_shape = self.loc.shape
         event_shape = torch.Size()
 
+        # Moments
+        self._variance = 1 - (_log_modified_bessel_fn(self.concentration, order=1) -
+                              _log_modified_bessel_fn(self.concentration, order=0)).exp()
+
         # Parameters for sampling
         tau = 1 + (1 + 4 * self.concentration ** 2).sqrt()
         rho = (tau - (2 * tau).sqrt()) / (2 * self.concentration)
@@ -74,7 +87,7 @@ class VonMises(TorchDistribution):
 
     def log_prob(self, value):
         log_prob = self.concentration * torch.cos(value - self.loc)
-        log_prob = log_prob - math.log(2 * math.pi) - _log_modified_bessel_fn_0(self.concentration)
+        log_prob = log_prob - math.log(2 * math.pi) - _log_modified_bessel_fn(self.concentration, order=0)
         return log_prob
 
     def sample(self, sample_shape=torch.Size()):
@@ -105,3 +118,11 @@ class VonMises(TorchDistribution):
             loc = self.loc.expand(batch_shape)
             concentration = self.concentration.expand(batch_shape)
             return type(self)(loc, concentration, validate_args=validate_args)
+
+    @property
+    def mean(self):
+        return self.loc
+
+    @property
+    def variance(self):
+        return self._variance
--- a/pyro/distributions/von_mises.py
+++ b/pyro/distributions/von_mises.py
@@ -126,13 +126,13 @@ class VonMises(TorchDistribution):
     @property
     def mean(self):
         """
-        The provided mean is the :attr:`circular` one.
+        The provided mean is the circular one.
         """
         return self.loc
 
     @property
     def variance(self):
         """
-        The provided variance is the :attr:`circular` one.
+        The provided variance is the circular one.
         """
         return self._variance
--- a/pyro/distributions/von_mises.py
+++ b/pyro/distributions/von_mises.py
@@ -69,7 +69,7 @@ class VonMises(TorchDistribution):
     """
     arg_constraints = {'loc': constraints.real, 'concentration': constraints.positive}
     support = constraints.real
-    has_rsample = True
+    has_rsample = False
 
     def __init__(self, loc, concentration, validate_args=None):
         self.loc, self.concentration = broadcast_all(loc, concentration)
@@ -92,7 +92,8 @@ class VonMises(TorchDistribution):
         log_prob = log_prob - math.log(2 * math.pi) - _log_modified_bessel_fn(self.concentration, order=0)
         return log_prob
 
-    def rsample(self, sample_shape=torch.Size()):
+    @torch.no_grad()
+    def sample(self, sample_shape=torch.Size()):
         # Based on:
         # Best, D. J., and Nicholas I. Fisher.
         # "Efficient simulation of the von Mises distribution." Applied Statistics (1979): 152-157.
@@ -100,9 +101,8 @@ class VonMises(TorchDistribution):
         x = torch.empty(shape, dtype=self.loc.dtype, device=self.loc.device)
         done = torch.zeros(shape, dtype=self.loc.dtype, device=self.loc.device).byte()
         while not done.all():
-            u1 = torch.rand(shape, dtype=self.loc.dtype, device=self.loc.device)
-            u2 = torch.rand(shape, dtype=self.loc.dtype, device=self.loc.device)
-            u3 = torch.rand(shape, dtype=self.loc.dtype, device=self.loc.device)
+            u = torch.rand((3,) + shape, dtype=self.loc.dtype, device=self.loc.device)
+            u1, u2, u3 = u.unbind()
             z = torch.cos(math.pi * u1)
             f = (1 + self._proposal_r * z) / (self._proposal_r + z)
             c = self.concentration * (self._proposal_r - f)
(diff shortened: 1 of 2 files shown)


neerajprad:
Thanks for the PR! Leaving some quick comments, but will take a deeper look at the math later:
 - I think there are a few stages to the accept-reject step here which may not be straightforward to fit in with the simpler Rejector API. Unless the code can be naturally simplified by sub-classing Rejector, I would keep it as is. @martinjankowiak or @fritzo might have better ideas.
 - Reg testing, I think some simple checks would be checking that the mean, variance for the sample matches the analytical expressions (we can add these properties since the class is missing these). For a more involved test, maybe we can follow the approach from [torch/test_distributions.py](https://github.com/pytorch/pytorch/blob/master/test/test_distributions.py#L679) and test the sampler against a reference like scipy's vonmises.

ahmadsalim:
Thank you for the comments!

I will take a look at your suggestions regarding testing and see how I can add mean and variance calculations.

ahmadsalim:
I added calculations for the analytic circular means and variance, and I have smoke tested them against Wolfram Alpha.
Apparently, scipy calculates the arithmetic means and variance for von Mises (i.e. the one you get if you use the standard `.mean()` and `.var()` functions in torch), which is different (and often less useful) than the circular ones.

Do you want me to add ways to calculate the circular mean and variance of samples somewhere, and then test against them? (e.g. according to http://www.fiserlab.org/manuals/procheck/manual/man_cv.html)

neerajprad:
> Actually, I found a better way to test the sampling process.
I have implemented a function that estimates the parameters of von Mises based on a set of samples (works best with a very large number of samples, like 1e6).

I think this sampling-scoring coherence, along with the tests for the mean and circular variance should be fine! Given that we have a sampling method now, I think we should also consider moving this to `torch.distributions` in a later PR, or even as is. cc @fritzo, @jpchen: what do you think?

pyro/distributions/von_mises.py
```diff
@@ -41,6 +51,36 @@ def _log_modified_bessel_fn_0(x):
...
+        return obj
+
+    for i in range(n_iter):
+        bfgs.step(bfgs_closure)
+    return mu, kappa.detach()
+
+
 class VonMises(TorchDistribution):
     """
     A circular von Mises distribution.
```
neerajprad: Time to remove this from the description below: `Currently only :meth:`log_prob` is implemented.`.

pyro/distributions/von_mises.py
```diff
@@ -64,13 +104,43 @@ def __init__(self, loc, concentration, validate_args=None):
...
+
         super(VonMises, self).__init__(batch_shape, event_shape, validate_args)
 
     def log_prob(self, value):
         log_prob = self.concentration * torch.cos(value - self.loc)
-        log_prob = log_prob - math.log(2 * math.pi) - _log_modified_bessel_fn_0(self.concentration)
+        log_prob = log_prob - math.log(2 * math.pi) - _log_modified_bessel_fn(self.concentration, order=0)
         return log_prob
 
+    def sample(self, sample_shape=torch.Size()):
```
neerajprad: It seems like we should also be able to differentiate through this. Correct me if I am wrong. If so, we should make this `rsample`, and add `has_rsample = True`.

pyro/distributions/von_mises.py
```diff
@@ -64,13 +104,43 @@ def __init__(self, loc, concentration, validate_args=None):
...
+        # Based on:
+        # Best, D. J., and Nicholas I. Fisher.
+        # "Efficient simulation of the von Mises distribution." Applied Statistics (1979): 152-157.
+        shape = self._extended_shape(sample_shape)
+        x = torch.empty(shape, dtype=self.loc.dtype, device=self.loc.device)
+        done = torch.zeros(shape, dtype=self.loc.dtype, device=self.loc.device).byte()
+        while not done.all():
+            u1 = torch.empty(shape, dtype=self.loc.dtype, device=self.loc.device).uniform_()
```
neerajprad: Can we not just use `torch.rand(dtype=self.loc.dtype, device=self.loc.device)`?

pyro/distributions/von_mises.py
```diff
@@ -17,22 +18,31 @@ def _eval_poly(y, coef):
...
+_I1_COEF_SMALL = [0.5, 0.87890594, 0.51498869, 0.15084934, 0.2658733e-1, 0.301532e-2, 0.32411e-3]
+_I1_COEF_LARGE = [0.39894228, -0.3988024e-1, -0.362018e-2, 0.163801e-2, -0.1031555e-1,
+                  0.2282967e-1, -0.2895312e-1, 0.1787654e-1, -0.420059e-2]
 
+_COEF_SMALL = [_I0_COEF_SMALL, _I1_COEF_SMALL]
+_COEF_LARGE = [_I0_COEF_LARGE, _I1_COEF_LARGE]
 
-def _log_modified_bessel_fn_0(x):
+
+def _log_modified_bessel_fn(x, order=0):
```
neerajprad: Neat!

pyro/distributions/von_mises.py
```diff
@@ -41,6 +51,36 @@ def _log_modified_bessel_fn_0(x):
     return result
 
 
+def _fit_params_from_samples(samples, n_iter=50):
```
neerajprad: If we are only using this for testing, let us move this to `test_von_mises.py`. If you have reason to think this will be more generally useful, we can add a docstring and remove the leading underscore.

pyro/distributions/von_mises.py
```diff
@@ -64,13 +104,43 @@ def __init__(self, loc, concentration, validate_args=None):
...
+        x = torch.empty(shape, dtype=self.loc.dtype, device=self.loc.device)
+        done = torch.zeros(shape, dtype=self.loc.dtype, device=self.loc.device).byte()
+        while not done.all():
+            u1 = torch.empty(shape, dtype=self.loc.dtype, device=self.loc.device).uniform_()
+            u2 = torch.empty(shape, dtype=self.loc.dtype, device=self.loc.device).uniform_()
+            u3 = torch.empty(shape, dtype=self.loc.dtype, device=self.loc.device).uniform_()
+            z = torch.cos(math.pi * u1)
+            f = (1 + self._proposal_r * z) / (self._proposal_r + z)
+            c = self.concentration * (self._proposal_r - f)
+            accept = (c / u2).log() + 1 - c >= 0
```
neerajprad: Not sure if this will make much difference, but we can make  `accept = (c * (2 - c) - u2) > 0 or (c / u2).log() + 1 - c >= 0` as in the paper.

pyro/distributions/von_mises.py
ahmadsalim: Ah, I understood it as `rsample` being completely reparametrizable (instead of just differentiable). The Von Mises distribution is not reparametrizable w.r.t. concentration which is why I picked `sample`, but I will change it to rsample now.

pyro/distributions/von_mises.py
ahmadsalim: Sure, I will do that :smile:

pyro/distributions/von_mises.py
neerajprad: Sorry, you are right - this is more involved than the accept-reject sampler in algorithm 1 from https://arxiv.org/pdf/1610.05683.pdf.

pyro/distributions/von_mises.py
```diff
@@ -52,25 +64,54 @@ class VonMises(TorchDistribution):
...
+
         super(VonMises, self).__init__(batch_shape, event_shape, validate_args)
 
     def log_prob(self, value):
         log_prob = self.concentration * torch.cos(value - self.loc)
-        log_prob = log_prob - math.log(2 * math.pi) - _log_modified_bessel_fn_0(self.concentration)
+        log_prob = log_prob - math.log(2 * math.pi) - _log_modified_bessel_fn(self.concentration, order=0)
         return log_prob
 
+    def rsample(self, sample_shape=torch.Size()):
```
fritzo: I believe rejection samplers do not admit reprametrization, so you should instead
```py
has_rsample = False

@torch.no_grad()
    ...
```
We can try to wrap this in a `Rejector` in a follow-up PR if you're interested.

pyro/distributions/von_mises.py
```diff
@@ -52,25 +64,54 @@ class VonMises(TorchDistribution):
...
+        # Based on:
+        # Best, D. J., and Nicholas I. Fisher.
+        # "Efficient simulation of the von Mises distribution." Applied Statistics (1979): 152-157.
+        shape = self._extended_shape(sample_shape)
+        x = torch.empty(shape, dtype=self.loc.dtype, device=self.loc.device)
+        done = torch.zeros(shape, dtype=self.loc.dtype, device=self.loc.device).byte()
+        while not done.all():
+            u1 = torch.rand(shape, dtype=self.loc.dtype, device=self.loc.device)
```
fritzo: nit: this would be faster on GPU if you manually fuse
```py
u = torch.rand((3,) + shape, dtype=self.loc.dtype, device=self.loc.device)
u1, u2, u3 = u.unbind()
```

pyro/distributions/von_mises.py
```diff
@@ -79,3 +120,11 @@ def expand(self, batch_shape):
             loc = self.loc.expand(batch_shape)
             concentration = self.concentration.expand(batch_shape)
             return type(self)(loc, concentration, validate_args=validate_args)
+
+    @property
+    def mean(self):
```
fritzo: It's worth noting that this is the circular mean rather than euclidean mean, either in the main class docstring or here. I think you can add an `:attr:` field to the main docstring, as we've done in a couple other places.

ahmadsalim: Thanks for the comments, I will look into how to do this.

ahmadsalim:
Thanks for the review. Is there any procedure that we should follow for moving this upstream? 😄

pyro/distributions/von_mises.py
```diff
@@ -52,25 +64,54 @@ class VonMises(TorchDistribution):
...
         log_prob = self.concentration * torch.cos(value - self.loc)
-        log_prob = log_prob - math.log(2 * math.pi) - _log_modified_bessel_fn_0(self.concentration)
+        log_prob = log_prob - math.log(2 * math.pi) - _log_modified_bessel_fn(self.concentration, order=0)
         return log_prob
 
+    @torch.no_grad()
+    def sample(self, sample_shape=torch.Size()):
+        # Based on:
+        # Best, D. J., and Nicholas I. Fisher.
+        # "Efficient simulation of the von Mises distribution." Applied Statistics (1979): 152-157.
```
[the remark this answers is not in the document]
ahmadsalim: Sure, will do that 😄

tests/distributions/test_von_mises.py
```diff
@@ -4,8 +4,51 @@
...
+    return mu, kappa.detach()
+
+
+@pytest.mark.parametrize('loc', [-math.pi/2.0, 0.0, math.pi/2.0])
+@pytest.mark.parametrize('concentration', [0.01, 0.03, 0.1, 0.3, 1.0, 3.0, 10.0, 30.0, 100.0])
+def test_sample(loc, concentration, n_samples=int(1e6), n_iter=100):
+    prob = VonMises(loc, concentration)
+    samples = prob.sample((n_samples,))
+    mu, kappa = _fit_params_from_samples(samples)
+    assert abs(loc - mu) < 0.1
```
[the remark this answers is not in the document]
ahmadsalim: I have tightened the bounds more for the concentration, but it seems that I can't get the precision to be lower for the following reason:

- For really low concenctration values, it seems that the sampled mean is lower in precision than `1e-2`
- For high concentrantion values, it is hard to infer a tight concentration 
   * you can see that there is no closed formula for inferring the concentration, so I had to rely on an initial estimate + use an optimizer to get within reasonable bounds of the true value

It seems that the asymptotics of the von Mises sampler are not as nice as other traditional distributions (like Normal) and perhaps there is some precision loss for the approximation of the parameter fitting. I have tried hard to find better sampling algorithms, and suggestions are more than welcome if you know of any! 😄

fritzo:
> Is there any procedure that we should follow for moving this upstream [to PyTorch]?

I'd recommend looking deeper into sampler accuracy, but if you're ok, I'd
- move this into torch/distributions/von_mises.py
- register in torch/distributions/__init__.py
- add lots of tests to test/test_distributions.py (mostly shape tests
- add to docs/source/
- cc Neeraj and me on the upstream PR

ahmadsalim:
Thanks for merging it!
I will follow the steps to move the distribution upstream.

I will also look more into the sample accuracy 😄 .

ahmadsalim:
Regarding the accuracy, it seems that I am experiencing the same issues with the numpy sampler.
I tried implementing the other suggested version in Best-Fisher without cosine, but it seems slower and not more accurate.
Note that this seems to be an issue with the mean only when the concentration is low like 0.1.

For the VonMises3D there is a method based on rejection sampling from [Ulrik-Wood](https://www.tandfonline.com/doi/abs/10.1080/03610919408813161), but also closed ones e.g. [Jacob](https://www.mitsuba-renderer.org/~wenzel/files/vmf.pdf) or [Kurz-Hanebeck](https://www.mitsuba-renderer.org/~wenzel/files/vmf.pdf). Apparently, it is easier to sample from the odd dimensional vM-F distributions than from the even dimensional ones. I can take a look at implementing the Jacob one soon.

PR #1734, opened 2019-02-06
Implement Dirichlet-Multinomial distribution

fritzo:
Addresses #1707 

## Tested

- added tests to test_conjugate.py

### DIFF
--- a/pyro/distributions/conjugate.py
+++ b/pyro/distributions/conjugate.py
@@ -1,15 +1,21 @@
 from __future__ import absolute_import, division, print_function
 
+import numbers
+
 import torch
 from torch.distributions import constraints
 from torch.distributions.utils import broadcast_all
 
-from pyro.distributions.torch import Beta, Binomial, Gamma, Poisson
+from pyro.distributions.torch import Beta, Binomial, Dirichlet, Gamma, Multinomial, Poisson
 from pyro.distributions.torch_distribution import TorchDistribution
 
 
 def _log_beta(x, y):
     return torch.lgamma(x) + torch.lgamma(y) - torch.lgamma(x + y)
+
+
+def _log_factorial(x):
+    return torch.lgamma(1 + x)
 
 
 class BetaBinomial(TorchDistribution):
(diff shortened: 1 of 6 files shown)


pyro/distributions/conjugate.py
```diff
@@ -87,6 +93,70 @@ def enumerate_support(self, expand=True):
...
+        return Binomial(self.total_count, probs).sample()
+
+    def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
+        n = self.total_count
+        alpha = self.concentration
+        alpha_sum = self.concentration.sum(-1)
+        return (_log_factorial(n) + torch.lgamma(alpha_sum) - torch.lgamma(n + alpha_sum) +
+                (torch.lgamma(value + alpha) - _log_factorial(value) - torch.lgamma(alpha)).sum(-1))
```
fehiepsi: @fritzo just a side note (for us to keep in mind if it helps for performance): when `value` is sparse, we can avoid computing `torch.lgamma(value + alpha) - _log_factorial(value) - torch.lgamma(alpha)` at zero-value indices of `value`.  However, to do so, we have to flatten `value` and `alpha` (to deal with batch_shape), select nonzero indices, compute the result, then transform back to match the batch_shape. I will address this note in the future when facing this distribution in a real case (if it helps).

pyro/distributions/conjugate.py
```diff
@@ -87,6 +93,70 @@ def enumerate_support(self, expand=True):
...
+        batch_shape = torch.Size(batch_shape)
+        new._dirichlet = self._dirichlet.expand(batch_shape)
+        new.total_count = self.total_count.expand(batch_shape)
+        super(DirichletMultinomial, new).__init__(batch_shape, validate_args=False)
+        new._validate_args = self._validate_args
+        return new
+
+    def sample(self, sample_shape=()):
+        probs = self._dirichlet.sample(sample_shape)
+        return Binomial(self.total_count, probs).sample()
```
neerajprad: This should be Multinomial.

fritzo:
@neerajprad thanks for the detailed review. I've added a new test to conftest.py and fixed a bug in `Multinomial.support` so that these tests pass.

pyro/distributions/conjugate.py
```diff
@@ -87,6 +93,75 @@ def enumerate_support(self, expand=True):
...
+        super(DirichletMultinomial, new).__init__(
+            new._dirichlet.batch_shape, new._dirichlet.event_shape, validate_args=False)
+        new._validate_args = self._validate_args
+        return new
+
+    def sample(self, sample_shape=()):
+        probs = self._dirichlet.sample(sample_shape)
+        total_count = int(self.total_count.max())
+        if not self.total_count.min() == total_count:
+            raise NotImplementedError("Inhomogeneous total count not supported by `enumerate_support`.")
```
neerajprad: enumerate_support --> sample

pyro/distributions/torch_patch.py
```diff
@@ -59,6 +59,16 @@ def _torch_linspace(*args, **kwargs):
     return ret
 
 
+# Fixes a shape error in Multinomial.support with inhomogeneous .total_count
+@patch_dependency('torch.distributions.Multinomial.support')
+@torch.distributions.constraints.dependent_property
+def _Multinomial_support(self):
```
neerajprad: Nice catch, I completely forgot the Multinomial upstream only supports integer total count.

Issue #1725, opened 2019-01-30
Make AbstractInfer classes lazily consume traces


neerajprad:
Our `TracePosterior` sub-classes like MCMC return a generator over traces. The `TracePosterior.run` method greedily consumes from this generator to hold a list with `num_samples` traces. 

If the user does not require access to raw traces, but simply the joint / marginal distribution over some of the sample sites, holding the raw traces in memory is wasteful. By reducing the traces to some terminal representation as they are generated rather than holding all of them in memory, we can reduce the rate at which memory consumption grows with `num_samples` for models that generate large sized tensors or where the initial data is large. This will also be useful with `num_chains > 1` on CUDA for which we need to copy tensors from worker processes to the main process: https://github.com/uber/pyro/pull/1694#issuecomment-456530927.

This is to propose a change in the interface so that we do not consume the traces greedily by calling `.run`. Instead traces can be consumed and relevant site values populated into `EmpiricalDistribution` when we call `.marginal` on these classes. Likewise, TracePredictive will only start running inference when `.marginal` is called. We can still preserve the `.run` method in case users need access to raw traces. 

This will require a few changes to the `EmpiricalDistribution` and `AbstractInfer` classes, but I wanted to open this issue for discussion. @eb8680 - I think you had some additional ideas on refactoring these classes, please feel free to add your thoughts.

In addition to efficiency, this will also have some additional benefits (we should make sure that the refactor addresses these points):
 - SVI shouldn't need a separate `num_samples` kwargs. This has led to some confusion as users sometimes use the default `num_samples=10` and simply resample from these traces when computing the posterior predictive. With lazy evaluation, this will be driven by the `num_samples` arg to `TracePredictive`.
 - The case of mismatched `cond_indep_stacks` between train and test as mentioned by @ahmadsalim in #1770.

fehiepsi:
+1 My use case is the following: I use autoguide, my SVI time is fast but it takes a bit of time (comparing to SVI time) to generate 10000 samples (marginal distribution) from posteriors. I believe that it will be faster if we only generate nodes' values (instead of traces). Currently, I do this way
```
for i in range(10000):
    guide_trace = poutine.trace(guide).get_trace(*args, **kwargs)
    model_poutine = poutine.trace(poutine.replay(model, trace=guide_trace))
    yield model_poutine.get_trace(*args, **kwargs), 1.0, 1.0
```
It will be faster if I can do the following in the `run` method:
```
for i in range(10000):
[...]

PR #1736, opened 2019-02-07
Remove networkx dependency from trace

neerajprad:
This removes dependency on networkx, by making the trace data structure not inherit off it. I have tried to keep the interface more or less the same as that of networkx.
 - Tested that there isn't any perf regression. The `test_tracegraph_elbo` is 20s faster (3m12s vs 3m31s) and GaussianPyramidTests are 45 seconds faster (15m35s vs 16m21s, which are both way too much!). The other tests are faster but not by a big margin, e.g. the HMC tests are faster by 5s.
 - Added test for topological sorting.

Refer to discussion in https://github.com/pyro-ppl/pyro/issues/1725.


pyro/poutine/trace_struct.py
```diff
@@ -103,18 +107,58 @@ def add_node(self, site_name, *args, **kwargs):
...
+    def _dfs(self, site, visited):
+        if site in visited:
+            raise StopIteration
+        for s in self._succ[site]:
+            for node in self._dfs(s, visited):
+                yield node
+        visited.add(site)
+        yield site
+
+    def topsort(self, reverse=False):
```
fritzo: nit: Maybe call this `topological_sort` or `topo_sort`? At first I thought `.topsort(reverse=True)` was equivalent to `.bottomsort()` or something.

tests/poutine/test_trace_struct.py
```diff
@@ -0,0 +1,36 @@
...
+    # 2 3 4
+    #   |
+    #   5
+    #   |
+    #   6
+    #   |
+    #   7
+    (1, [(1, 2), (3, 5), (1, 4), (1, 3), (5, 6), (6, 7)]),
+])
+def test_top_sort(root, edges):
```
fritzo: :+1: nice examples. I'd recommend a stronger test, like
```py
EDGE_SETS = [
   # your edge sets here
]

@pytest.mark.parametrize('edges', [
    perm for edges in EDGE_SETS
    for perm in itertools.permutations(edges)
])
def test_top_sort(edges): 
    tr = Trace()
    for n1, n2 in edges:
        tr.add_edge(n1, n2)
    top_sort = tr.top_sort()

    # check all nodes are accounted for exactly once
    expected_nodes = set(n1 for (n1, n2) in edges) | set(n2 for (n1, n2) in edges)
    assert len(top_sort) == len(expected_nodes)
    assert set(top_sort) == expected_nodes

    # check no edge ordering is violated
    ranks = {n: rank for rank, n in enumerate(top_sort)}
    for n1, n2 in edges:
        assert rank[n1] < rank[n2]
```

neerajprad:
> Nice! Would also be good to check if this helps the performance gap between Pyro and minipyro.

This seems to help actually. On my system (mac with python 3.7), with 5000 steps, with networkx, our runtime was 3.9s (minipyro) vs 6.9s(full), whereas without it, our runtime is is 3.7s (minipyro) vs 6.1s (full).

PR #1739, opened 2019-02-08
Minor fixes for progress bar for MCMC diagnostics

### LINKED ISSUE #1728: [FR] Use normal tqdm in notebooks by default
Currently, Pyro's MCMC uses tqdm's auto_notebook to display progress bar by default. However, I find that it causes more troubles than advantages:
+ auto_notebook uses jupyter widget, which requires users install widget extension.
+ the widget is only available at running time, which leaves some weird messages when sharing results or reopening it. For example, in [this gist](https://gist.github.com/fehiepsi/fa2f7980aef1e7242670a2470694ee2e), it leaves the line

```
HBox(children=(IntProgress(value=0, description='Warmup', max=1300, style=ProgressStyle(description_width='ini…
```

I tried to replace [this line](https://github.com/uber/pyro/blob/dev/pyro/infer/mcmc/logger.py#L7) by `from tqdm import tqdm` but it only works for `num_chains=1`. For multi-chain, the issue https://github.com/tqdm/tqdm/issues/630 still remains.

neerajprad:
Refactoring of mcmc progress bar logging with the following changes:
 - Fixes #1728 - Uses `tqdm` instead of the more experimental ipywidget based implementation in all situations unless we are using multiprocessing in the notebook environment.
 - Fixes an issue with progress bars getting over-written when `.close()` is called.
 - Simplifies the code for single / multi progress bars by moving all of them into a single class that maintains its own tqdm instance (or multiple instances when `num_chains > 1`) under the hood. Given the number of behavioral overrides we had made to tqdm, this consolidates all of those tweaks into a single class.

**In the notebook env with:**

num_chains = 1
screen shot 2019-02-07 at 11 16 27 pm

num_chains = 2
screen shot 2019-02-07 at 11 16 47 pm

### DIFF
--- a/pyro/infer/mcmc/logger.py
+++ b/pyro/infer/mcmc/logger.py
@@ -4,7 +4,8 @@ import os
 import sys
 from collections import OrderedDict
 
-from tqdm.auto import tqdm
+from tqdm import tqdm
+from tqdm.auto import tqdm as tqdm_nb
 
 try:
     get_ipython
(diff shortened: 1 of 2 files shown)


pyro/infer/mcmc/logger.py
```diff
@@ -38,22 +39,72 @@
...
+    def __init__(self, warmup_steps, num_samples, min_width=80, max_width=120,
+                 disable=False, num_bars=1):
+        total_steps = warmup_steps + num_samples
+        # Disable progress bar in "CI"
+        # (see https://github.com/travis-ci/travis-ci/issues/1337).
+        disable = disable or "CI" in os.environ or "PYTEST_XDIST_WORKER" in os.environ
+        bar_format = None
+        if not ipython_env:
+            bar_format = "{l_bar}{bar}| {n_fmt}/{total_fmt} [{elapsed}, {rate_fmt}{postfix}]"
+        pbar_cls = tqdm_nb if num_bars > 1 and ipython_env else tqdm
```
neerajprad: @fehiepsi - This is the line that implements the behavior we discussed in #1728.

fehiepsi: Got it. Thanks @neerajprad !

pyro/infer/mcmc/logger.py
```diff
@@ -38,22 +39,72 @@
...
+class ProgressBar(object):
+    """
+    Initialize progress bars using :class:`~tqdm.tqdm`.
+
+    :param int warmup_steps: Number of warmup steps.
+    :param int num_samples: Number of MCMC samples.
+    :param int min_width: Minimum column width of the bar.
+    :param int max_width: Maximum column width of the bar.
+    :param int pos: Position of the bar (e.g. in the case of
+        multiple parallel samplers).
```
fehiepsi: nit: this `pos` parameter is not available in init method.

neerajprad: Oops..sorry, it shouldn't be there. Will remove.

PR #1740, opened 2019-02-08
Implement sparse computation for DirichletMultinomial.log_prob()

fritzo:
As recommended by @fehiepsi , this PR implements a sparse version of the `DirichletMultinomial.log_prob()` computation, toggled by a constructor arg `is_sparse`.

Note this works with the jit and with cuda, but not with jit+cuda at the same time.

Note we could further speed things up by using a sparse data representation, but that is beyond the scope of this PR.

## Tested

- added a unit test for correctness
- profiled on real sparse data, where this offers 2x speedup on my entire model (most of the time was being taken by `torch.lgamma`)

### DIFF
--- a/pyro/distributions/conjugate.py
+++ b/pyro/distributions/conjugate.py
@@ -104,11 +104,13 @@ class DirichletMultinomial(TorchDistribution):
     :param float or torch.Tensor concentration: concentration parameter (alpha) for the
         Dirichlet distribution.
     :param int or torch.Tensor total_count: number of Categorical trials.
+    :param bool is_sparse: Whether to assume value is mostly zero when computing
+        :meth:`log_prob`, which can speed up computation when data is sparse.
     """
     arg_constraints = {'concentration': constraints.positive, 'total_count': constraints.nonnegative_integer}
     support = Multinomial.support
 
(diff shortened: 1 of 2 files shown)


pyro/distributions/conjugate.py
```diff
@@ -146,8 +150,16 @@ def log_prob(self, value):
...
         alpha_sum = self.concentration.sum(-1)
-        return (_log_factorial(n) + torch.lgamma(alpha_sum) - torch.lgamma(n + alpha_sum) +
-                (torch.lgamma(value + alpha) - _log_factorial(value) - torch.lgamma(alpha)).sum(-1))
+        if self.is_sparse:
+            value, alpha = torch.broadcast_tensors(value, alpha)
+            elementwise = torch.zeros_like(value)
+            mask = (value != 0)
+            value = value[mask]
+            alpha = alpha[mask]
+            elementwise[mask] = torch.lgamma(value + alpha) - _log_factorial(value) - torch.lgamma(alpha)
```
fehiepsi: neat!

neerajprad:
Good to merge?

fritzo:
Yes, ready to merge.

PR #1743, opened 2019-02-09
Improve sparse DirichletMultinomial.log_prob()

fritzo:
This improves on #1740 by also using a sparse computation for the `total_count` portion of the computation. This is useful for models with a large batch of low-dimensional Dirichlet-Multinomial features, where only a few features are present in each observation.

## Tested
- covered by existing unit test
- verified ~10% speedup on CPU on an example dataset

### DIFF
--- a/pyro/distributions/conjugate.py
+++ b/pyro/distributions/conjugate.py
@@ -14,8 +14,17 @@ def _log_beta(x, y):
     return torch.lgamma(x) + torch.lgamma(y) - torch.lgamma(x + y)
 
 
-def _log_factorial(x):
-    return torch.lgamma(1 + x)
+def _log_beta_1(alpha, value, is_sparse):
+    if is_sparse:
+        mask = (value != 0)
+        value, alpha, mask = torch.broadcast_tensors(value, alpha, mask)
+        result = torch.zeros_like(value)
+        value = value[mask]
+        alpha = alpha[mask]
+        result[mask] = torch.lgamma(1 + value) + torch.lgamma(alpha) - torch.lgamma(value + alpha)
+        return result
+    else:
+        return torch.lgamma(1 + value) + torch.lgamma(alpha) - torch.lgamma(value + alpha)
 
 
 class BetaBinomial(TorchDistribution):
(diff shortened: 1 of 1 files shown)


pyro/distributions/conjugate.py
```diff
@@ -14,8 +14,17 @@ def _log_beta(x, y):
...
+    if is_sparse:
+        mask = (value != 0)
+        value, alpha, mask = torch.broadcast_tensors(value, alpha, mask)
+        result = torch.zeros_like(value)
+        value = value[mask]
+        alpha = alpha[mask]
+        result[mask] = torch.lgamma(1 + value) + torch.lgamma(alpha) - torch.lgamma(value + alpha)
+        return result
+    else:
+        return torch.lgamma(1 + value) + torch.lgamma(alpha) - torch.lgamma(value + alpha)
```
fehiepsi: +1 I think that this method will be handy for other discrete conjugate distributions too (e.g. it can be used in BetaBinomial class).

pyro/distributions/conjugate.py
```diff
@@ -147,19 +156,9 @@ def sample(self, sample_shape=()):
...
-            value, alpha = torch.broadcast_tensors(value, alpha)
-            elementwise = torch.zeros_like(value)
-            mask = (value != 0)
-            value = value[mask]
-            alpha = alpha[mask]
-            elementwise[mask] = torch.lgamma(value + alpha) - _log_factorial(value) - torch.lgamma(alpha)
-        else:
-            elementwise = torch.lgamma(value + alpha) - _log_factorial(value) - torch.lgamma(alpha)
-        return _log_factorial(n) + torch.lgamma(alpha_sum) - torch.lgamma(n + alpha_sum) + elementwise.sum(-1)
+        return (_log_beta_1(alpha.sum(-1), value.sum(-1), self.is_sparse) -
```
fehiepsi: Sorry, I need to verify the math at this.

fehiepsi: @fritzo If I understand correctly, then the reason for using `value.sum(-1)` instead of total_count here is to make it a tensor?

fritzo: The two are identical. We may allow `total_count=None` in the future (as is done by `torch.distributions.Multinomial`), and it's nice to use the `value.sum(-1)` version in anticipation.

PR #1744, opened 2019-02-09
Implement NormalNormalLocConjugate

vishwakftw:
Changelog:

- Implement the Normal-Normal conjugate distribution combination where the mean of the likelihood normal is sampled from a Normal prior.

Test Plan:

- Added conjugate distribution specific tests in test_conjugate.py

Address one action point in #1707 

cc: @fritzo


fritzo:
Hi @vishwakftw, while this PR looks correct, I think we need to work on the conjugacy interface a little more before this distribution will be useful. I'll comment on the original issue #1707 with details.

I'm going to close this PR for now while we work out interface issues. And design help is welcome!

PR #1745, opened 2019-02-09
[WIP] Add sampling, mean and variance for von Mises-Fisher distribution

ahmadsalim:
Hi Pyro Devs!

I tried to implement the sampling algorithm for VonMises 3D as discussed in #1729.
I still need to add some tests, especially since I am unsure about the variance-covariance matrix.

### DIFF
--- a/pyro/distributions/von_mises_3d.py
+++ b/pyro/distributions/von_mises_3d.py
@@ -59,12 +59,12 @@ class VonMises3D(TorchDistribution):
         base_tensor = torch.tensor([0., 0., 1.], dtype=self._mean.dtype, device=self._mean.device)
         base_mean_cross = base_tensor.cross(self._mean, dim=-1)
         bmn = base_mean_cross.norm(2, -1)
-        base_mean_inner = base_tensor @ self._mean
+        base_mean_inner = torch.matmul(base_tensor, self._mean)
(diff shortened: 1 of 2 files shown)


yulkang:
Hi @ahmadsalim - it seems it's been a while since you made the pull request - is there a known issue with this commit?

ahmadsalim:
@yulkang I don't think there is anything wrong per se, I just have not done any extensive testing yet 😄. I have not really been needing this distribution that much recently and I have been busy with some other things at work, so I forgot about it.
I will see if I can find time to do some testing at some point soon.

ahmadsalim:
So I have been looking at this issue and I got a bit stuck.
It seems that I need higher-order version of the modified Bessel function to estimate the parameter for testing. 

However I only could find a stable way of computing that in "numerical recipes in C" (where the code is copyrighted).
I think GNU Scientific Library also has an implementation, but the code there is GPL licensed, which is incompatible with the license for this project.

Any ideas on other sources where I could find the relevant algorithm?

Thanks in advance!

ahmadsalim:
Sorry, I think I found a relevant source. 😄

fritzo:
Closing in favor of #2736 . Feel free to reopen.

Issue #1692, opened 2019-01-01
Wishart / InverseWishart / LKJ priors


rachtsingh:
Would it be possible to implement Wishart / InverseWishart / LKJ priors? 

gpytorch has them [already](https://github.com/cornellius-gp/gpytorch/blob/master/gpytorch/priors/wishart_prior.py), but when I tried mixing in `TorchDistributionMixin` to get something useable in Pyro, I realized that they don't have a `.sample` method. 

I don't think it's easy to get efficient samplers for the InverseWishart (I think trying to build it via a TransformedDistribution might be too slow), but curious to see other approaches. There's a Tensorflow Probability tutorial on this [here](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/examples/jupyter_notebooks/TensorFlow_Probability_Case_Study_Covariance_Estimation.ipynb), which explains some of the underlying ideas, as well as a note [here](https://www.math.wustl.edu/~sawyer/hmhandouts/Wishart.pdf).

Great work by the way! Pyro is *so* easy to use it's incredible.

eb8680:
You actually don't need a `.sample` method for a distribution unless you want to sample from it - see e.g. [`pyro.distributions.VonMises`](https://github.com/uber/pyro/blob/dev/pyro/distributions/von_mises.py).  Does your use case require sampling from the prior?  Were you getting an error when you tried to use `TorchDistributionMixin` with the GPyTorch versions?

rachtsingh:
I was trying to use HMC (NUTS), which I think requires sampling (I do get an NotImplementedError). On second thought, though this might belong in `torch.distributions` instead of here, so feel free to close.

eb8680:
> I was trying to use HMC (NUTS), which I think requires sampling

That's just for setting the initial trace - you can hack around that for now by defining a dummy sample method for your prior that just returns an arbitrary appropriately shaped value and is in the prior's support.

rachtsingh:
Got it. I just tried it with the patched method, and ran into two issues:

1. We can't merge GPyTorch `Prior` classes and `TorchDistributionMixin` since the former subclasses `nn.Module` but doesn't implement `forward`. So I just recreated the distribution by copying their code (see below).
2. This doesn't work either because:
```
NotImplementedError: Cannot transform _PositiveDefinite constraints
```
which makes sense.

Actually modeling-wise the LKJ prior would be more useful, but I tried that and it also runs into the same error. 

Here's my attempt:

```python
[...]

PR #1746, opened 2019-02-11
Distribution for Cholesky Factor of LKJ Prior

elbamos:
Addresses #1692.

The PR includes:
1. A distribution that generates lower cholesky factors of correlation matrices based on the LKJ Prior.
2. A bijective (see note) transform from unconstrained space to lower cholesky factors of correlation matrices.
3. A constraint appropriate to lower cholesky factors of correlation matrices. 
4. Tests for the constraints and transform.
5. An example.

The code is based on the Stan source. 

Note the following limitations:
1.  The distribution takes a single value of eta and single dimension and cannot be reparameterized.  This limitation allows it to be more efficient by caching part of the log_prob calculation. I believe this is appropriate given the intended use case. (One wouldn't generally want to reparameterize an LKJ distribution.)
2.  The transform is not perfectly bijective when the inputs exceed a certain magnitude.  This is necessary because of a numerical precision issue. The reasons why are discussed in lkj.py.
3.  Currently, one function is dependent on the batched `tril` in pytorch 1.0.1; this restriction can be removed if necessary. 
4.  I don't have a test for the `log_prob` function. It closely tracks the Stan source and produces reasonable output; but I haven't found a known-reliable source for getting the log prob to use in building a test. 

Thanks,

-Amos


fritzo:
To test `log_prob` we could add an empirical nearest-neighbor test, which is a generic method to test agreement between a sampler and density. The idea is that for random samples from a distribution `p(x)`, the distances to nearest neighbors should be Chi-squared distributed with dof = dimension. For details see
- https://github.com/posterior/goftests/blob/master/goftests/__init__.py#L207
- http://projecteuclid.org/download/pdf_1/euclid.aop/1176993668
- http://arxiv.org/pdf/1006.3019v2.pdf
- http://en.wikipedia.org/wiki/Nearest_neighbour_distribution

examples/lkj.py
```diff
@@ -0,0 +1,39 @@
...
+import math
+from pyro.infer.mcmc import HMC, MCMC, NUTS
+from pyro import poutine
+
+"""
+This simple example is intended to demonstrate how to use an LKJ prior with
+a multivariate distribution.
+"""
+
+pyro.enable_validation(True)
```
fritzo: Thanks for adding an example! Would you mind wrapping this in a `main()` function with `argparse` and options for `--num-samples` and `--warmup-steps` (see [examples/baseball.py](https://github.com/pyro-ppl/pyro/blob/dev/examples/baseball.py#L309)). This is a little more work, but it allows us to [test examples](https://github.com/pyro-ppl/pyro/blob/dev/tests/test_examples.py#L19) and to add configuration for `--jit` and `--cuda`.

elbamos: Yes, I will do so, thanks.

elbamos:
I'm not familiar with the empirical nearest-neighbor test.  I will look at the links and attempt to use them to create a test. Thank you.

examples/lkj.py
elbamos: Example should be conformant now.

elbamos:
@fritzo I took a look at the nearest-neighbor distribution approach.  It appears that there's some complexity in adapting the method as a test for the validity of the log_prob calculation.  I'm reluctant to try to write a test based on a formulation I don't have experience with, where I'd be uncertain about the validity of the implementation and not in a position to properly evaluate the test outcomes. 

One way of reviewing the validity of the implementation is by comparison to the Stan source. The relevant files are: https://github.com/stan-dev/math/blob/release/v2.18.1/stan/math/prim/mat/prob/lkj_corr_lpdf.hpp  and https://github.com/stan-dev/math/blob/release/v2.18.1/stan/math/prim/mat/prob/lkj_corr_cholesky_lpdf.hpp 

I continue to be open to alternatives.

pyro/distributions/__init__.py
```diff
@@ -37,6 +38,8 @@
     "GaussianScaleMixture",
     "InverseAutoregressiveFlow",
     "InverseAutoregressiveFlowStable",
+    "CorrLCholeskyLKJPrior",
+    "CorrLCholeskyLKJPrior",
```
neerajprad: Remove the repeated entry here.

examples/lkj.py
```diff
@@ -0,0 +1,74 @@
...
+
+    with pyro.plate("observations", N) as n:
+        obs = pyro.sample("obs", dist.MultivariateNormal(mu, scale_tril=L_Omega), obs=y[n])
+    return obs
+
+def main(args):
+    y = torch.randn(args.n, args.num_variables).to(dtype=torch.double)
+    if args.cuda:
+        y = y.cuda()
+    nuts_kernel = NUTS(model, jit_compile=args.jit)
```
neerajprad: A few things to note - you mentioned about Pyro using really large values when sampling in the unconstrained region, and why that might happen. The way it works is - we start sampling by using an initial sample from the prior, which is transformed to the unconstrained region using a default transform registered for the support (constraint) of the distribution. In this case, the transform is the `CorrLCholeskyTransform`, and the initial sampling starts from what is given by `_inverse`. Subsequent values proposed are from the leapfrog integrator. 

I checked that this example gives a really large initial value for the gradient, i.e. the hamiltonian trajectory diverges for the initial step size (the default I believe is 1) and hence you see this error. If you start from a small step size, you won't face this issue. e.g. in this code, if you change, your invocation of NUTS to `NUTS(model, jit_compile=args.jit, step_size=1e-3)`, the example will run fine without the eps hack. It is possible that there are other numerical stability issues that I haven't looked into yet, but just based on this, I would think that we shouldn't introduce `eps`, since this is probably a corner case that we can either address in HMC, or throw an error asking the user to lower the initial step size, but this doesn't need to a concern of the LKJ prior.

pyro/distributions/lkj.py
```diff
@@ -0,0 +1,197 @@
...
+        for k in range(d-1):
+            alpha -= .5
+            for j in range(k+1, d):
+                concentrations[..., i] = alpha
+                i += 1
+        self._generating_distribution = Beta(concentrations, concentrations)
+        self._eta = eta
+        self._d = d
+        self._lkj_constant = None
+        self._event_shape = torch.Size((d, d))
```
neerajprad: Call the parent constructor here, `super(CorrLCholeskyLKJPrior, self).__init__(batch_shape, event_shape, validate_args=validate_args)`.

pyro/distributions/lkj.py
```diff
@@ -0,0 +1,197 @@
...
+
```
neerajprad: We will also need to define a `.expand` method which is used to return an expanded distribution instance, and skips some of the parameter checks for efficiency. It used by Pyro to do automatic vectorization. [example code](https://github.com/pytorch/pytorch/blob/master/torch/distributions/lowrank_multivariate_normal.py#L111).

pyro/distributions/__init__.py
elbamos: ok, sorry

examples/lkj.py
```diff
@@ -0,0 +1,74 @@
+from __future__ import absolute_import, division, print_function
+
+import math
```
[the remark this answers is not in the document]
elbamos: Linted, thanks!

tests/distributions/test_lkj.py
```diff
@@ -0,0 +1,80 @@
...
+    assert_tensors_equal(y, y_prime, prec=1e-4)
+
+    # test domain
+    assert (transform.domain.check(y_prime) == 1).all()
+
+    # test log_abs_det_jacobian
+    log_det = transform.log_abs_det_jacobian(y, x)
+    assert log_det.shape == y_shape[:-1]
+    if y_shape == ():
+        assert_tensors_equal(torch.autograd.grad(y, (x,), retain_graph=True)[0].abs().log(), log_det)
```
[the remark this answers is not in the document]
elbamos: Correct... The tests were written by @fehiepsi.  I didn't want to alter/delete more of his code than necessary, as a courtesy, and because I figured that if he added it, he must have had something in mind.  (I'm not sure why we care about an empty shape...)  Would you prefer for me to add a test for this situation or to take out the test?

examples/lkj.py
elbamos: Thanks, that explanation helps.  I think the decision about the eps is really a judgment call at the level of the pyro maintainers.   

Let me add two points:

1.  If an out-of-bound value occurs, the result is an unrecoverable LAPACK exception. To me, this suggests that we're better-off enforcing an eps, because of the possibility that somewhere in the sampling some iteration will exceed bounds.  And because otherwise, we will have to advise users that the distribution only works if the step size is reduced as you propose. 

2. If we do choose to keep eps, then the eps has to be larger than it is.  In additional experimentation, I found more cases where, as the matrix grows larger, the eps is insufficient.  My plan (before reading your comment) was to attempt to derive a formula for the minimum eps needed to guarantee that there won't be an LAPACK exception.

Please let me know your preference regarding eps - I'll wait to make the next push until this is resolved.

pyro/distributions/lkj.py
elbamos: Is what's required to support producing multiple samples from the distribution with the same dimensionality and eta?  Or to support producing multiple samples where eta or dimensionality can change?  

If its the latter, that would have a significant performance implication.  The reason is that the bulk of the work in calculating `log_prob` is constant for a given dimensionality and value of `eta`, so we can cache that when its first generated and avoid the repeat work.    

It would also be tricky to support producing samples with different dimensionality in the same batch.

pyro/distributions/lkj.py
elbamos: Thanks, I will address in the context of resolving the comment concerning `.expand`

tests/distributions/test_lkj.py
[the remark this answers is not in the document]
fehiepsi: `autograd.grad` works for a transform from `x` to `y`, not from . It should be fine if you change back to my [original notations](https://github.com/elbamos/pyro/pull/4/files#diff-45451189a07bb3d85b0b27895c9328cdR52).

examples/lkj.py
fehiepsi: I recommend to verify if your distribution gives right distribution. One way to verify is to sample 100000 samples and test if marginal distribution of each off-diagonal element of correlation matrices (generated by taking `x @ x.t` where x is your sampled Cholesky factors) follows a Beta distribution with a correct mean and variance. I get no numerical issues in my code so I suggest you add more tests to see where is the problem. In case it helps, here are some tests to verify your log_prob implementation is correct:
+ you can test if the log_prob of generated **correlation matrices** are constant when eta=1.
+ When eta != 1, you can test if your implemented log_prob agrees with the log_prob of a transformed distribution, where base distribution is your beta distribution and transform is the transform from partial correlation to cholesky factor (you can use _autograd_log_det function to calculate Jacobian of this transform).

tests/distributions/test_lkj.py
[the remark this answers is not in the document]
fehiepsi: In addition, here and there you add prec=1e-4. Pyro's test uses DoubleTensor so I think that adding precision is unnecessary given that the transform is just a combination of some normal functions.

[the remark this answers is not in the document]
fehiepsi: This is the reason why I have to fix the code while adding tests. Without fixing (I mean by just adding tests alone), you might have no clue if the code is wrong or the test is wrong like this case,... You can freely modify the fixed code to your favourite, as long as the tests pass. This might cause a big issue when the test and the code both are wrong, so we need to review. But generally, I think that reviewing the test is easier than reviewing the code.

examples/lkj.py
neerajprad: @elbamos - Good point. An eps value higher than 1e-4 makes me a bit uncomfortable, and I think it is okay to raise an exception. Later, we can add more context to the error message to allow the user to debug further, e.g. lower step size if they are running HMC; or even add padding if we see a need. I am hesitant to add padding in anticipation of certain errors that might arise, because the padding itself might lead to the kind of issues you had pointed to earlier. The only failure case we know of is when we run NUTS, and I think the right place to address that would be the NUTS code itself (in a later PR). e.g. we can wrap the initial sampling part in a `try catch` block and start with a reasonable step size that doesn't diverge off.

pyro/distributions/lkj.py
neerajprad: Torch distributions have a concept of `batch_shape` and `event_shape` to characterize the shape of samples produced. `event_shape` corresponds to the dependent dimensions of a distribution, i.e. those that are summed out when we call `log_prob`, and the batch dims correspond to the independent dimension (though not necessarily identical) that are retained when we call `log_prob`. 

In this case (like in the multivariate normal case), the event shape is `(d, d)`. This distribution is currently not vectorized, i.e. we can't pass in a tensor of `eta` values, but suppose `eta` could have shape (m, n, 1), where the last dim has size 1 and corresponds to the `eta` for the batch. It would then produce samples of shape (m, n, d, d). Then when we call `.expand([i, j, m, n])`, it will merely expand the `eta` parameters as `eta = eta.expand([i, j, m, n])`. However, this only works if the underlying implementation is batched, like the [multivariate normal](https://github.com/pytorch/pytorch/blob/master/torch/distributions/multivariate_normal.py) distribution. I am not sure how feasible that is. Do you think it is possible to batch this distribution and it being able to take in a tensor (of non-identical)`eta`s as mentioned above, and how much work would that be? That will make this distribution play well with Pyro's and PyTorch's batching semantics (and you'll be able to use it inside `pyro.plate` to auto-vectorize).

tests/distributions/test_lkj.py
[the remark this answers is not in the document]
elbamos: @fehiepsi The tests at issue don't actually test anything, and they were never called by your code.  

I changed your notation so it would be identical to the notation used in both the Stan source code and in the original papers implemented by Stan (which was the same notation I had used throughout). 

Ok, I will remove the unnecessary tests.

examples/lkj.py
elbamos: @neerajprad Ok, I will take out the eps. 

@fehiepsi You are wrong.  Your code *did* start to throw the same exception, and it *did* fail tests, as soon as the size of the matrix grew to be greater than 5.  

The LKJ prior does *not* follow a Beta distribution.  Individual elements of the z-vector follow Beta distribution*S*, plural.  But that is only true when sampling from the LKJ Prior.  It doesn't give a basis for testing the log probability.

pyro/distributions/lkj.py
elbamos: The underlying implementation *is* batched, but only for a single, constant value of `eta`.  

If you parameterize the distribution with `d=3` and `eta=1`, and then call `sample` with a sample size of `i, j, k`, it will return a result of shape `i, j,k,d,d`, where each of the `i * j * k` samples is drawn from the LKJ prior where `eta=1`.   It would not be hard to write an `expand` that increased the batch size but kept `eta` constant.

The difficulty arises when `eta` isn't constant.   

If this is something you feel we need, I will do it.  I would ask you to consider seriously, however --- do we really expect this distribution to be called from inside a `plate`?  Its hard for me to imagine a use-case for that, but I recognize that other people's use-cases are different from my own.

neerajprad: Thanks for explaining! I think using this inside `plate` is reasonable, say if you want to specify a prior over the covariance of a batch of multivariate normals. However, it is not obvious to me if there are many cases that need non-identical etas, and I think that's what you were asking. Maybe others can chime in.

To begin with, I think it is reasonable to have a restriction of supporting only iid batching, as you mention. I have the following proposal - that we modify the current constructor to take in tensor `eta` (say with some batch dims `(m, n, ..)`), and we can do a check like `assert eta.max() == eta.min()` in the constructor and throw a `NotImplemented` error for non-identical `eta`. `.expand` would simply involve appending batch dims to `eta` then. If in the future, we see a need to have non-identical eta, the API would remain the same and we can make modifications to the underlying implementation without breaking backward compatibility. Does this sound reasonable?

examples/lkj.py
fehiepsi: I have not touched your distribution implementation so I don't know where is the origin of your issue. The above are just my suggestions.

I just push a branch lkj2 with tests [here](https://github.com/pyro-ppl/pyro/blob/lkj2/tests/distributions/test_lkj.py) for your reference. I have put comments/proofs pretty careful so I think you can understand it. My branch might not work out of box so it just serves as a reference. It does not follow Stan but follow Lewandowski paper. Hope it help!

tests/distributions/test_lkj.py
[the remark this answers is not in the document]
fehiepsi: I see some differences between that version and the version you put here. There, the condition is `if len(x_shape) == 1:` (so it is `len(y_shape) == 1` with your notation). Here the condition is `if y_shape == ():`, so it will be ignored. You can remove the test but I recommend to replace by another test for Jacobian to see if the transform `log_abs_det_jacobian` method is implemented correctly. Here your transform has `event_dim=1` so it is required to test for input with `ndim=1`. If you test for some transforms else with `event_dim=0` (e.g. AffineTransform, TanhTransform) then testing for constant input is enough.

examples/lkj.py
fehiepsi: Here is a [gist](https://gist.github.com/fehiepsi/a254d77fcd024c14c674cba02634731c) which I run your example under plate. At you can see, no error at all.

pyro/distributions/lkj.py
elbamos: Yes, that makes sense.  So we support varying the `batch_size` while maintaining constant `eta`.  I will implement that tonight and include it in a push with the other issues you've raised.   Thanks!

pyro/distributions/lkj.py
elbamos: Should be resolved in what I just pushed.

pyro/distributions/lkj.py
elbamos: Should be resolved by what I just pushed.

elbamos:
@neerajprad All issues raised should now be addressed either by updated code, or in the discussion above.  If I've failed to address any, or created new ones, that was certainly inadvertent, please let me know.

Thanks much, 
-Amos

examples/lkj.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def main(args):
+    y = torch.randn(args.n, args.num_variables).to(dtype=torch.double)
+    nuts_kernel = NUTS(model, jit_compile=args.jit, step_size=1e-5)
+    MCMC(nuts_kernel, num_samples=args.num_samples, warmup_steps=args.warmup_steps, num_chains=args.num_chains).run(y)
+
+
+if __name__ == "__main__":
+    assert pyro.__version__.startswith('0.3.0')
```
neerajprad: We just had a 0.3.1 release yesterday, so this should be `0.3.1` otherwise it will throw an error.

tests/test_examples.py
```diff
@@ -17,6 +17,7 @@
     'air/main.py --num-steps=1',
     'air/main.py --num-steps=1 --no-baseline',
     'baseball.py --num-samples=200 --warmup-steps=100 --num-chains=2',
+    'lkj.py --n=50 --num-chains=1 --warmup-steps=100 --num-samples=200',
```
neerajprad: You'll also need to add this to `TEST_JIT` and `TEST_CUDA`, otherwise, `test_coverage` would complain that the example is not being tested. If JIT or CUDA does not work on this example, you can use `xfail` markers and specify why we need to make an exception for this example. To quickly test locally, you can also run `make test-examples` to see that all checks pass.

elbamos: Thanks; hopefully clear now.  If its not -- I'm not sure this example, which is really only intended to show the boilerplate for combining a correlation l cholesky factor and a vector of variances, is a book that's worth the candle.

neerajprad:
I think the last merge got messed up? I'm seeing unrelated changes from dev now.

pyro/distributions/lkj.py
```diff
@@ -0,0 +1,205 @@
...
+        constant = torch.lgamma(eta.add(0.5 * Km1)).mul(Km1)
+
+        k = torch.linspace(start=1, end=Km1, steps=Km1, dtype=eta.dtype, device=eta.device)
+        constant -= (k.mul(math.log(math.pi) * 0.5) + torch.lgamma(eta.add(0.5 * (Km1 - k)))).sum()
+
+        self._lkj_constant = constant
+        return constant
+
+    def log_prob(self, x):
```
neerajprad: This will have some subtle shape bugs, which makes the distribution batching semantics somewhat inconsistent with the rest of pytorch.

 - When you score a sample from `CorrLCholeskyLKJPrior(3, torch.tensor([1.]))` for example, its shape will be `[1]` instead of `[]` (i.e. it should return a torch scalar) which is the correct shape for a distribution with empty `batch_shape`. 
 - Another error would be when the data to be scored has a smaller shape than the distribution's batch_shape. In that case, other distributions would simply expand the `log_prob`. e.g. `CorrLCholeskyLKJPrior(3, torch.tensor([1.])).expand([5]).log_prob(torch.ones(3, 3)).shape == (5,)`. 

To debug these kind of shape bugs, I strongly recommend adding a fixture to `tests/distributions/conftest.py` - even a single example like below - and running `test_distributions.py`. I think the fixes should be relatively small. e.g. you might have to expand `eta` by `batch_shape` in `log_prob`.

```python
    Fixture(pyro_dist=dist.CorrLCholeskyLKJPrior,
            examples=[
                {'d': 3, 'eta': [1.],
                 'test_data': torch.ones(5, 3, 3)},
            ]),

```

pyro/distributions/lkj.py
```diff
@@ -0,0 +1,205 @@
...
+                i += 1
+        self._gen = Beta(concentrations, concentrations)
+        self.eta = eta
+        self._d = d
+        self._lkj_constant = None
+        super(CorrLCholeskyLKJPrior, self).__init__(torch.Size(), torch.Size((d, d)), validate_args=validate_args)
+
+    def rsample(self, sample_shape=torch.Size()):
+        y = self._gen.rsample(sample_shape=self.batch_shape + sample_shape).detach()
+        z = y.to(dtype=self.eta.dtype).mul(2).add(-1.0)
```
neerajprad: Is the casting to eta's dtype necessary? I would think that `self._gen` was initialized with eta's `dtype` and would therefore return the correct `dtype`. 

Also, I think this method should be `sample` and `has_rsample = False`, since the distribution isn't reparametrized?

elbamos:
> I think the last merge got messed up? I'm seeing unrelated changes from dev now.

I rebased onto dev to pick up the version bump. That's all.

pyro/distributions/lkj.py
```diff
@@ -0,0 +1,205 @@
...
+    :param torch.Tensor eta: A single positive number parameterizing the distribution.
+    """
+    arg_constraints = {"eta": constraints.positive}
+    support = corr_cholesky_constraint
+    has_rsample = True
+
+    def __init__(self, d, eta, validate_args=None):
+        if not torch.is_tensor(eta):
+            raise ValueError("Eta must be a tensor")
+        if any(eta <= 0):
```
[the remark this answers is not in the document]
elbamos: Ok.

pyro/distributions/lkj.py
elbamos: Thank you for that! I apologize that this has been somewhat seriatim - I wasn't familiar with the pyro test/build process before, and the templates I was working from didn't cover this.  I will take care of this new set of issues over the next day or two.

neerajprad:
> I rebased onto dev to pick up the version bump. That's all.

Hmm..I'm not sure what went wrong, but I'd suggest hard resetting HEAD to the version before your rebase (you could use `git reflog`), and then merging (not rebasing) dev. That should fix it. You'll probably need to force-push to update the PR.

pyro/distributions/lkj.py
elbamos: Regarding `sample` vs. `rsample`, that's what I thought at first, to use `sample`. But, looking at some of the example distributions, it seems that for, e.g., Beta, the only reparameterization accepted by `rsample` is the `sample_shape`, so doing it this way seemed like the shortest path to complying with pyro's shape semantics.  See, e.g., https://github.com/pytorch/pytorch/blob/016f21235721d9998b1ebcd5c7948e1049233771/torch/distributions/beta.py#L57   

Whichever method you prefer is fine with me, just let me know. 

Regarding the `dtype`, I added that when I was working through the eps issue. I'll check if its still necessary and take it out if its not.  

Thanks!

pyro/distributions/lkj.py
neerajprad: Please take your time! We value your contributing to Pyro, and we realize that there's a learning curve to becoming an effective contributor (specially since documentation mostly caters to users and not contributors), and we are all here to help. 😄

elbamos:
> > I rebased onto dev to pick up the version bump. That's all.
> 
> Hmm..I'm not sure what went wrong, but I'd suggest hard resetting HEAD to the version before your rebase (you could use `git reflog`), and then merging (not rebasing) dev. That should fix it. You'll probably need to force-push to update the PR.

That could work.  It could also make the situation much worse.  Is the rebased version too difficult to work with?

neerajprad:
> That could work. It could also make the situation much worse. Is the rebased version too difficult to work with?

That was more to have a confidence that we aren't merging in any changes unwittingly with this PR that isn't actually part of it, and for the convenience of anyone following it now or in the future. Anyways, I changed the base branch and the web ui seems to be generating the right diff now.

fehiepsi:
@elbamos You'll need to sign CLA (at the bottom of this PR) and feel free to change label to "awaiting review" when this PR is ready.

pyro/distributions/lkj.py
neerajprad: > Beta, the only reparameterization accepted by `rsample` is the `sample_shape`.

The shape signature for `rsample` is the same as `sample` (i.e. it only takes an additional `sample_shape` arg), except with the additional requirement that the samples are differentiable with respect to the distribution parameters. In case of Beta, you should be able to compute the gradient w.r.t. `concentration1` and `concentration0` of any objective that uses a sample generated via `rsample`. With `sample`, you aren't expected to compute gradients through the samples and will likely see an exception when you try to do so.

neerajprad: Since we can't do this with the lkj distribution, we should set `has_rsample = False` and just have a non-reparametrized `sample` method.

elbamos: Ok. While it wouldn't be particularly difficult to propagate the gradient through `rsample` (although at a not insignificant cost to performance), I think we should be discouraging users from trying to optimize `eta`.   Thanks!

elbamos: I’m now thinking this about this in detail and I think there’s an interesting question. I hope you won’t mind talking it through with me?

While optimizing eta wouldn’t make sense, what could make sense is optimizing the parameters to the generating set of beta distributions. LKJ by design uses betas where the a and b parameters are equal. If we separate those into two vectors, then we could optimize back. 

The optimized distribution would not be an lkj prior. But it would be a generating distribution for (I think) the posterior of the cholesky factor of the correlation matrix.

What do you think?

If we do this, then how would this need to interact with expand and with sampling more than one correlation matrix at a time?  Would we need to expand and detach the paramaters for the generating distribution so each item in the batch is optimized separately?

neerajprad: I think you are right that we might be able to differentiate back maybe with some modifications to the code (I am not completely sure), but I am unclear about the use case you mention. Generally `rsample` is very useful for Variational Inference where the ELBO objective has some expectation term over the latent which we need to differentiate through. We can use the reparametrization trick if our distribution has an `rsample` method, and push the gradient through the expectation to get a low variance gradient estimator. One reason to reparametrize this might be to optimize eta of the variational distribution to maximize the ELBO in VI. I am not sure why we would need to take gradients through a sample otherwise, say in the case of MCMC. Just `log_prob` being differentiable should be sufficient, but maybe I don't fully understand the use case you mention? If we want to attempt to reparametrize LKJ, my suggestion would be to do this in a following PR, just so we can contain the scope of this, and add further tests for `rsample`. 😄

elbamos: Sounds good, thanks!

pyro/distributions/lkj.py
elbamos: I've implemented the fixture and resolved the first issue you raise (when the batch_shape is empty, log_prob returns a scalar; when the batch shape is non-empty, log_prob returns a tensor of batch_shape; when sample shape is non-empty, log_prob returns a tensor of sample_shape; when both batch_shape and sample_shape are non-empty, log_prob returns a tensor of batch_shape + sample_shape).  

I'm a bit confused by 

> * Another error would be when the data to be scored has a smaller shape than the distribution's batch_shape

I'm not sure what this is referring to?  Does this mean that `log_prob` was called with an object of arbitrary shape?  The current behavior is that as long as the last two dimensions of the passed object match the dimensionality of the distribution, it will return a `log_prob` with a shape matching the passed-shape, less the last two dimensions. 

What **should** be the behavior in that case?

eb8680: > What **should** be the behavior in that case?

The shape/broadcasting semantics of PyTorch Distributions (and TensorFlow Distributions) specify that the shape of the output of `.log_prob(x)` is the result of [broadcasting](https://pytorch.org/docs/master/notes/broadcasting.html) the batch shape of the input (as determined by the distribution's `event_shape`) with the distribution's `batch_shape`.

Imagine a distribution with `batch_shape == (3,1,4)` and `event_shape == (5,)`. Here are several examples to illustrate:
- Any input tensor `x` with `x.shape[-1:] != (5,)` would cause an error, because it does not match the distribution's `event_shape`
- Any input tensor `x` with `x.shape[-2] != 4` or `x.shape[-4] != 3` would cause an error, because it does not match the non-empty dimensions of the distribution's `batch_shape`.
- An input tensor `x` with `x.shape == (5,)` would produce `.log_prob(x).shape == (3,1,4)`.
- An input tensor `x` with `x.shape == (2,4,5)` would produce `.log_prob(x).shape == (3,2,4)`.
- An input tensor `x` with `x.shape == (2,3,1,4,5)` would produce `.log_prob(x).shape == (2,3,1,4)`

elbamos: Super-helpful - thank you!

elbamos:
This is ready for review; I was asked to change the label from WIP to Awaiting Review, but I don't think I have the access permissions to do that.

fehiepsi:
@elbamos Sorry, I forgot about it. In the future, I think that editing titles of your PRs (kind of adding "[WIP]" or "[Ready]") will be enough.

fritzo:
> changed the label from WIP to Awaiting Review

GitHub just released a [draft pull request feature](https://github.blog/2019-02-14-introducing-draft-pull-requests/), which should make this easier 🎉

pyro/distributions/lkj.py
```diff
@@ -0,0 +1,210 @@
...
+
+        k = torch.linspace(start=1, end=Km1, steps=Km1, dtype=eta.dtype, device=eta.device)
+        constant -= (k.mul(math.log(math.pi) * 0.5) + torch.lgamma(eta.add(0.5 * (Km1 - k)))).sum()
+
+        self._lkj_constant = constant
+        return constant
+
+    def log_prob(self, x):
+        if x.shape[-1] != self._d or x.shape[-2] != self._d:
+            raise ValueError("CorrLCholeskyLKJPrior with dimensionality {} got {}".format(self._d, x.shape))
```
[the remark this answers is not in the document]
elbamos: Ok.

tests/distributions/test_lkj.py
```diff
@@ -0,0 +1,76 @@
+from __future__ import absolute_import, division, print_function
+
+import pytest
+import torch
+from torch.distributions import biject_to, transform_to
+from pyro.distributions.lkj import (corr_cholesky_constraint, CorrLCholeskyTransform)
+from tests.common import assert_tensors_equal
+
+
+@pytest.mark.parametrize("value_shape", [(1, 1), (3, 3), (5, 5)])
```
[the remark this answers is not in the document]
elbamos: Yes, those three tests make sense.  I will add them and address the other two issues you raised in a push tonight.  Thanks!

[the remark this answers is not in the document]
elbamos: Actually, now that I'm coding this, I'm not so sure about the third test: 

> * For `eta=1., d > 2`, the log_prob of any sample should be a constant, i.e.

This doesn't seem to be correct to me.   If it were correct, its equivalent to saying that the log_prob is determined only by `eta` and `d`, and not by the content of the matrix.   

See also lines 73-78 here: https://github.com/stan-dev/math/blob/983061e8bc4fbbb5bc1bd2d5b029b599c886880b/stan/math/prim/mat/prob/lkj_corr_cholesky_lpdf.hpp#L73

Can you elaborate on what you have in mind?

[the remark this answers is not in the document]
neerajprad: Just briefly looking through their code, it does seem that they special case `eta==1` - `sum(values)` should just be a constant, if I am not wrong.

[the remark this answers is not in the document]
elbamos: There are three terms in the calculation of the log prob.  One part is the calculation of the lkj constant, which is constant for any given pair of `eta` and `d`.  But there are two other terms.  Look at lines 78-79, 82 (together, the second term) and 85 (the third term) of the code I linked to. 

If eta is 1 **and** not an optimizable parameter, then Stan will skip line 85, since that term drops out.   (This corresponds to line 206 in what I'm about to push.)  

The second term of the calculation takes place at 75-79.  This corresponds to lines 201-204 of the code here.  That part of the calculation is applied even if eta is 1 and a constant; see line 82. 

What do you think?

(Its entirely possible that Stan is deviating from the paper; given the choice, I'd place my bet on Stan's calculation being the correct one.)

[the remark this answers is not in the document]
elbamos: I have tried to be faithful to the organizational structure of the Stan implementation throughout, to simplify verifying the code's correctness.  I've tried to deviate only to vectorize and avoided Stan's optimizations around dropping constant terms, since Pyro doesn't track which terms are constant, and the optimization benefit would only be obtained by the combination of a strongly-typed language and a strong compiler anyway.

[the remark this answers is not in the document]
fehiepsi: @neerajprad the paper derived log_prob of correlation matrix, here @elbamos implements Cholesky version. We need to add jacobian of cholesky transform (of correlation matrix) to make it uniform when eta=1.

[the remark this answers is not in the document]
elbamos: @fehiepsi The jacobian of the cholesky transformation is the product of the diagonal elements in the cholesky matrix.  I don't see that taking place in the code.  

While the code does take the log-diagonal elements of the cholesky matrix, I'm pretty sure its doing that in order to perform a calculation on the log-determinant of the original matrix -- consistent with @neerajprad's observation that the density is proportional to the determinant^(eta-1).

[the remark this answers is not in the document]
neerajprad: Hmm..it does seem that you wouldn't get a constant density with their implementation, but look at the top of the file with the comment that `eta == 1 <-> uniform`. 😄Maybe my brain has switched off, but I think that it is worth clearing this because the behavior at `eta==1` is called out in many other implementations / descriptions. I'm not too familiar with the Stan code but is it possible that the `lpdf` also does a jacobian adjustment for the transformation?

pyro/distributions/lkj.py
```diff
@@ -0,0 +1,209 @@
...
+def _transform_to_corr_cholesky(constraint):
+    return CorrLCholeskyTransform()
+
+
+########################################
+# Define distribution
+########################################
+
+# TODO: Modify class to support more than one eta value at a time?
+class CorrLCholeskyLKJPrior(TorchDistribution):
```
neerajprad: On a completely unrelated note, what do you think about renaming this to `LKJCorrPrior` or anything short enough that begins with `LKJ`? A simple name would be memorable and less wordy to use in day to day. That this implements a cholesky version can be read up in the docstrings, and is standard enough in other libraries that we probably can skip it in the name.

elbamos:
I’m fine with that except what if someone implements an lkj prior on a complete matrix rather than a cholesky factor?l

tests/distributions/test_lkj.py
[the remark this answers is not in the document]
fehiepsi: @elbamos I'm not sure that I understand Stan code correctly but it seems to me that the Jacobian adjustment for Cholesky transform is [this term](https://github.com/stan-dev/math/blob/983061e8bc4fbbb5bc1bd2d5b029b599c886880b/stan/math/prim/mat/prob/lkj_corr_cholesky_lpdf.hpp#L79). @neerajprad 's log_prob testing is validate for the case D=2 because jacobian of cholesky transform in that case is 1 (hence, log of it is 0). A way to get that (numerical) Jacobian term is to use `_autograd_log_det` function (in test_lkj.py file). For example
```
# here `sample` is the Cholesky of a correlation matrix
tril_index = sample.new_ones(dimension, dimension).tril(diagonal=-1) > 0.5
# start with tril vector version
sample_tril = sample[tril_index].clone().requires_grad_()
sample_cloned = sample.new_ones(dimension, dimension).tril(diagonal=-1)
sample_cloned_tmp = sample_cloned.clone()
sample_cloned_tmp[tril_index] = sample_tril
sample_cloned_diag = (1 - sample_cloned_tmp.pow(2).sum(-1)).sqrt()
sample_cloned[tril_index] = sample_tril
sample_cloned.view(-1)[::dimension + 1] = sample_cloned_diag
y = sample_cloned.matmul(sample_cloned.transpose(-2, -1))
corr_tril = y[tril_index]
jacobian = _autograd_log_det(corr_tril, sample_tril)
```

neerajprad:
> I’m fine with that except what if someone implements an lkj prior on a complete matrix rather than a cholesky factor?

How about just `LKJCorrCholesky`, in that case? That would be short enough and easy to autocomplete by editors/IDEs.

elbamos:
Ok that works I’ll include it in the next push.

fehiepsi:
For a better context, tensorflow has [LKJ](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/LKJ) distribution which goes along with Lewandowski's paper and implements log_prob using SVD (to compute log_det of a matrix). But I prefer @elbamos's approach to only use Cholesky version because it will be faster when D is large (because no SVD or cholesky decomposition is involved). If only that, then `LKJCholesky` would be enough. However, some frameworks also implement LKJ for covariance matrix :(. So I am in for LKJCorrCholesky with you guys.

tests/distributions/test_lkj.py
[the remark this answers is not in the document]
elbamos: @fehiepsi I would be more than pleased to include an appropriate test. 

However, in my view, a test needs to satisfy the criteria that it can be implemented without doubt as to both (a) the mathematical validity of the test, and (b) the correctness of the test implementation. 

I would also encourage you to read the entirety of the Stan thread you've linked to. In particular, the part where they discuss the effect of the bug.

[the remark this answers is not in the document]
fehiepsi: Yeah, thanks for pointing it out! In HMC, when `eta` is not a parameter, we can remove normalization constant to save computation and to avoid worrying about if it is correctly implemented or not, so it will not affect the inference. That and the lack of tests might be the reason why the bug is lately detected. However, the bug will affect SVI or Important Inference. Anyway, I believe that by adding above tests, I would prefer reading your code than following Stan's code (which I couldn't understand how they derive their terms so I have to rely on Lewandowski et. paper instead :( ). :)

About the math for Cholesky transformation, I write down what I understand here in case it helps
```
# Note about computing Jacobain of the transformation from Cholesky factor to
# correlation matrix:
#
#   Assume C = L@Lt and L = (1 0 0; a \sqrt(1-a^2) 0; b c \sqrt(1-b^2-c^2)),
#   Then off-diagonal lower triangular vector of L is transformed to the off-diagonal
#   lower triangular vector of C by the transform:
#       (a, b, c) -> (a, b, ab + c\sqrt(1-a^2))
#   Hence, Jacobian = 1 * 1 * \sqrt(1 - a^2) = \sqrt(1 - a^2) = L22, where L22
#       is the 2th diagonal element of L
#   Generally, for a D dimensional matrix, we have
#       Jacobian = L22^(D-2) * L33^(D-3) * ... * Ldd^0
```
It seems similar to the term under your log_diagonals's implementation. Though I am not sure if they are exactly the same or not. Maybe some constants else are involved.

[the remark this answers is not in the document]
elbamos: I'm not sure why you believe it would affect SVI or Importance inference...  We're talking about a constant term.  Its also (by the way) a term that's totally separate from and irrelevant to the discussion we're having.   And we are including tests that would spot that bug, if it existed in this code.

The problem with the tests you propose is that I'm not convinced that they are either mathematically correct or implemented properly.  

To be blunt, I'm not sure what you're trying to get at with these tests -- what they purport to be testing, or why you think its important.  

Let's say the Stan code does include the log abs det jacobian of the cholesky decomposition. That would not imply that the calculation is wrong, or wrongly implemented, either in Stan or in this PR.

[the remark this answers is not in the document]
fehiepsi: I don't think that your code is wrong. And I also think that Stan devs have fixed their code. I just suggest a way to test your code. It seems to me that your code will give uniform distribution for the case eta=1 after adjusting Jacobian (because 2(eta-1)... term will be 0). Maybe you get me wrong.

elbamos:
I see. In that case it should be straightforward to modify the test code to account for that case. I’ll give it a try later tonight. Thanks.

tests/distributions/test_lkj.py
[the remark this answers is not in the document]
elbamos: I tried it.  Looks good!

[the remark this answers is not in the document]
fehiepsi: Nice!

pyro/distributions/lkj.py
```diff
@@ -0,0 +1,209 @@
...
+    """
+    arg_constraints = {"eta": constraints.positive}
+    support = corr_cholesky_constraint
+    has_rsample = False
+
+    def __init__(self, d, eta, validate_args=None):
+        if eta.numel() != 1:
+            raise ValueError("eta must be a single number; for a larger batch size, call expand")
+        if d <= 1:
+            raise ValueError("d must be >= 1 in any correlation matrix")
```
neerajprad: nit: error message should be `d must be > 1`.

PR #1750, opened 2019-02-12
Bump to version 0.3.1

fritzo:
minor release


neerajprad:
We will need to update the version string for examples / tutorials. I'll push a commit to this branch, since the assertion will fail otherwise.

PR #1751, opened 2019-02-12
Fix version string for release tagged commits

neerajprad:
We have a sha id that gets appended to the version string to make it easier to collect debug information, specially when users are using Pyro's dev branch. This however also appends the sha id when HEAD is pointing to a release tag, which doesn't play well when we want to upload to pypi. This adds a fix so that if HEAD points to a release tag no sha id is appended to the version string.

@fritzo - Let me know if this seems reasonable, I don't want to unwittingly introduce other complications. 😄


neerajprad:
Good to merge. Let us merge this before bumping up the version number in #1750.

PR #1757, opened 2019-02-13
update docs urls

jpchen:
RTD now supports deploying tags as versions, so we no longer need to cut release branches for each version. changed stale URL links.

### DIFF
--- a/pyro/contrib/minipyro.py
+++ b/pyro/contrib/minipyro.py
@@ -226,7 +226,7 @@ class Adam(object):
 
 # This is a unified interface for stochastic variational inference in Pyro.
 # The actual construction of the loss is taken care of by `loss`.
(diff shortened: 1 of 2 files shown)


pyro/contrib/minipyro.py
```diff
@@ -226,7 +226,7 @@ def __call__(self, params):
 
 # This is a unified interface for stochastic variational inference in Pyro.
 # The actual construction of the loss is taken care of by `loss`.
-# See http://docs.pyro.ai/en/0.3.0-release/inference_algos.html
+# See http://docs.pyro.ai/en/0.3.0/inference_algos.html
```
neerajprad: This should be `0.3.1` too?

jpchen: yep dont know how i missed that...

PR #1765, opened 2019-02-19
Ignore ipynb_checkpoints for flake8 and isort

fehiepsi:
I mainly use jupyterlab for writing scripts/editing files, which leaves many `.ipynb_checkpoints` folders. On the other hand, `flake8` and `isort` act on all files including files in `ipynb_checkpoints` folders. This PR adds ipynb_checkpoints exceptions for these tools.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,6 +12,7 @@ known_third_party = opt_einsum, six, torch, torchvision
 [tool:pytest]
 filterwarnings = error
     ignore::DeprecationWarning
+    once::DeprecationWarning
 
 doctest_optionflags = ELLIPSIS NORMALIZE_WHITESPACE IGNORE_EXCEPTION_DETAIL
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,16 +1,16 @@
 [flake8]
 max-line-length = 120
-exclude = docs/src, build, dist
+exclude = docs/src, build, dist, .ipynb_checkpoints
 
 [isort]
 line_length = 120
 not_skip = __init__.py
+skip_glob = .ipynb_checkpoints
 known_first_party = pyro, tests
 known_third_party = opt_einsum, six, torch, torchvision
 
 [tool:pytest]
 filterwarnings = error
-    ignore::DeprecationWarning
     once::DeprecationWarning
 
 doctest_optionflags = ELLIPSIS NORMALIZE_WHITESPACE IGNORE_EXCEPTION_DETAIL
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -1,16 +1,16 @@
...
 [isort]
 line_length = 120
 not_skip = __init__.py
+skip_glob = .ipynb_checkpoints
 known_first_party = pyro, tests
 known_third_party = opt_einsum, six, torch, torchvision
 
 [tool:pytest]
 filterwarnings = error
-    ignore::DeprecationWarning
```
fehiepsi: @neerajprad I think this line is redundant. What do you think?

fritzo: Can someone explain to me why this is redundant?

fritzo: I believe we should instead remove the `once::DeprecationWarning` line. We really do want to ignore `DeprecationWarning`s

fehiepsi: Yeah, let me see what's different when we use `ignore` instead of `once`.

setup.cfg
```diff
@@ -1,17 +1,17 @@
...
 line_length = 120
 not_skip = __init__.py
+skip_glob = .ipynb_checkpoints
 known_first_party = pyro, tests
 known_third_party = opt_einsum, six, torch, torchvision
 
 [tool:pytest]
 filterwarnings = error
     ignore::DeprecationWarning
-    once::DeprecationWarning
```
fritzo: @fehiepsi Can you please revert this change? It is unrelated to the PR title and breaks previously tuned logging. These kind of tiny "cleanup" tweaks are not necessary. Let's spend time on useful things like GPyTorch integration.

fehiepsi: @fritzo Though it might not be necessary, these works really help me relax my mind after reading hard-to-understand papers. I just do this clean-up job periodically for that reason.

neerajprad: +1. Both `once` and `ignore` would be ignored by pytest. `once` will print the warning only for the first occurrence of such a warning, and `ignore` would not print the warning at all. The reason why it was set to `once` is so that we see warnings at the end of pytest summary about deprecated methods in external dependencies, and can take appropriate action ahead of time. Let us not remove this! :)

fehiepsi: It seems complicated. Haizzz....

fritzo: @fehiepsi understood, I also find it relaxing to clean up after reading papers. Just remember that cleaning up Pyro also incurs review overhead, and reviewing subtle config PRs may not be relaxing to reviewers :wink:

PR #1766, opened 2019-02-19
Correct egg info file in makefile

fehiepsi:
I'm not sure if the egg info filename is consistence with various platforms, but in my linux system, it has the name is `pyro_ppl.egg-info`.

@neerajprad Could you please check the name of egg-info in Mac?


neerajprad:
> @neerajprad Could you please check the name of egg-info in Mac?

Also the same for mac, I think this is OS agnostic. Good catch.

PR #1768, opened 2019-02-19
Clean up requirements

fehiepsi:
This PR cleans up requirements in `setup.py` and `docs/requirement.txt` files.

@fritzo Is it fine to drop requirement for numpy now?

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,5 +2,6 @@ sphinx
 sphinx_rtd_theme
 contextlib2
 graphviz>=0.8
+numpy>=1.7
 opt_einsum>=2.3.2
 tqdm>=4.28
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,9 +1,6 @@
 sphinx
 sphinx_rtd_theme
-numpy>=1.7
 contextlib2
-cloudpickle>=0.3.1
 graphviz>=0.8
-observations>=0.1.4
 opt_einsum>=2.3.2
-tqdm>=4.25
+tqdm>=4.28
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -83,7 +84,6 @@
         # add them to `docs/requirements.txt`
         'contextlib2',
         'graphviz>=0.8',
-        'numpy>=1.7',
```
neerajprad: We could have removed this if all pytorch builds shipped with numpy. The pytorch pip wheel does not install numpy, but will throw an ImportError without numpy, because there are parts of torch that call numpy. I don't see any harm in keeping this in the requirement since it should anyways be installed with pytorch, and our minimum version requirement is an old one.

fehiepsi: Got it, thanks @neerajprad !

fritzo: It's probably worth adding a comment explaining why we need the numpy requirement, since a code comment is more discoverable than a github comment :wink:

eb8680:
Why are the CodeCov comments back?

docs/requirements.txt
```diff
@@ -1,9 +1,7 @@
 sphinx
 sphinx_rtd_theme
-numpy>=1.7
 contextlib2
-cloudpickle>=0.3.1
```
fritzo: @jpchen can you confirm we no longer need cloudpickle or observations?

jpchen: yep LGTM. besides internal applications, i'm not sure which example was using cloudpickle in the first place...

docs/requirements.txt
```diff
@@ -1,9 +1,7 @@
...
 graphviz>=0.8
-observations>=0.1.4
```
jpchen: i think this is still used in the AIR example from #500
@neerajprad should we move off this dependency ?

fehiepsi: You are right, it seems that I miss that tutorial while using "grep".

neerajprad: This is indeed needed by a few tutorials, let us reinstate it back.

PR #1770, opened 2019-02-19
Change how `TracePredictive` works, so it resets plates on resampling

ahmadsalim:
Hi Pyro Devs,

While I was trying to use `TracePredictive` on test data of different size than training data, I noticed two issues:

- The `marginal` uses the posterior one instead of the one that use the trace computed by the predictive posterior when calling `run`.
-  The plates have fixed size from training data, and were not reset before. This meant that one could only run with test data of the same size as the training data. I am now repurposing `SamplePosterior` to allow a change in plate dimensions.

It seems that the tests are breaking now, but I am unsure whether this is because my changes are fundamentally wrong or the tests were relying on the apparently weird behaviour before.


pyro/poutine/posterior_messenger.py
```diff
@@ -0,0 +1,11 @@
+from .replay_messenger import ReplayMessenger
+
+
+class SamplePosteriorMessenger(ReplayMessenger):
+    # This acts like ReplayMessenger but additionally replays cond_indep_stack.
+
+    def _pyro_sample(self, msg):
+        if msg["infer"].get("enumerate") == "parallel":
+            super(SamplePosteriorMessenger, self)._pyro_sample(msg)
+        if msg["name"] in self.trace:
```
fritzo: :+1: this is a nice trick. @martinjankowiak @eb8680 do you think we could use something like this to support subsampling in autoguides?

ahmadsalim: Thanks! Unfortunately, this is something I can not take credit for.
It was just a class I moved from `infer.discrete`, because I found it useful 😄

ahmadsalim: Git history shows it is probably @eb8680 who implemented it.

pyro/poutine/posterior_messenger.py
```diff
@@ -0,0 +1,11 @@
...
+            msg["cond_indep_stack"] = self.trace.nodes[msg["name"]]["cond_indep_stack"]
```
neerajprad: You will also need to call the parent method towards the end, otherwise you wouldn't be replaying, and just sampling from the prior! That should fix the test, I suppose?

```python
super(SamplePosteriorMessenger, self)._pyro_sample(msg)
```

It will be great if you added some tests to `test_abstract_infer`! Existing tests are pretty bare and only exercise MCMC, and it will be good to have some tests for `SVI` too.

ahmadsalim: Thanks!

ahmadsalim: I will try to add some more tests 😄

pyro/infer/abstract_infer.py
```diff
@@ -297,11 +297,12 @@ def _traces(self, *args, **kwargs):
...
             model_trace = self.posterior()
-            replayed_trace = poutine.trace(poutine.replay(self.model, model_trace)).get_trace(*args, **kwargs)
-            yield (replayed_trace, 0., 0)
+            resampler = poutine.resample_posterior(self.model, model_trace)
+            resampled_trace = poutine.trace(resampler).get_trace(*args, **kwargs)
+            yield (resampled_trace, 0., 0)
 
     def marginal(self, sites=None):
         """
         Gets marginal distribution from posterior.
```
fehiepsi: @ahmadsalim Thanks for modifying the behaviour of this method! I personally find this change really helpful! I think we'd need to change the docs to "Gets marginal distribution of this predictive." (or something similar) too in case this change is acceptted. :)

ahmadsalim: Sure, I will fix it!

pyro/poutine/posterior_messenger.py
ahmadsalim: I am trying to add a test with the following guide, but I am running into shape issues. Any ideas what could be wrong? Thanks! 😄 

```python
def guide(num_trials):
  phi_c0 = pyro.param("phi_c0", num_trials.new_ones(num_trials.shape[0]) * 500.0)
  phi_c1 = pyro.param("phi_c1", num_trials.new_ones(num_trials.shape[0]) * 500.0)
  phi_posterior = dist.Beta(concentrantion0=phi_c0, concentrantion1=phi_c1)
  pyro.sample("phi", phi_posterior)
```

If I have it like above, I get an error about invalid shape of `log_prob` and that I should try using `to_event`. If I add `to_event(1)` on the `phi_posterior` I get an error about a mismatch between the shape in the model and the one in the guide. This also seems to happen when I use an automatic guide like `AutoDelta`

neerajprad: I would suggest using a plate over the data dims in both model and guide:

```python
...
with pyro.plate("num_trials", num_trials):
    pyro.sample("phi", phi_posterior)
```

Feel free to commit your test, and we can take another look, if that doesn't work.

ahmadsalim: Putting it in a plate seems to solve the issue I mentioned!
Unfortunately, it seems that with the manual guide, it look like it is sampling from the prior again? Maybe I misspecified something, but I thought that `Beta` was conjugate to `Binomial` and so the guide should work well.

ahmadsalim: The auto guide tests, work however!

neerajprad: One issue is that with `Beta(500, 500)`, you are placing a heavy prior around 0.5. Consider using a weaker prior and higher learning rate.

tests/infer/test_abstract_infer.py
```diff
@@ -31,6 +39,48 @@ def test_posterior_predictive():
...
+    num_trials = torch.ones(5) * 1000
+    num_success = dist.Binomial(num_trials, true_probs).sample()
+    conditioned_model = poutine.condition(model, data={"obs": num_success})
+    opt = optim.Adam(dict(lr=1e-3))
+    loss = Trace_ELBO()
+    guide = beta_guide
+    svi_run = SVI(conditioned_model, guide, opt, loss, num_steps=10000, num_samples=1000).run(num_trials)
+    posterior_predictive = TracePredictive(model, svi_run, num_samples=10000).run(num_trials)
+    marginal_return_vals = posterior_predictive.marginal().empirical["_RETURN"]
+    assert_equal(marginal_return_vals.mean, torch.ones(5) * 700, prec=30)
```
neerajprad: I suggest using `assert_close(rtol=0.05)` (using some suitable value for `rtol`) in these cases, which is easier to read and shows the relative tolerance used at a glance.

tests/infer/test_abstract_infer.py
```diff
@@ -31,6 +39,48 @@ def test_posterior_predictive():
...
+def test_posterior_predictive_svi_manual_guide():
+    true_probs = torch.ones(5) * 0.7
+    num_trials = torch.ones(5) * 1000
+    num_success = dist.Binomial(num_trials, true_probs).sample()
+    conditioned_model = poutine.condition(model, data={"obs": num_success})
+    opt = optim.Adam(dict(lr=1e-3))
+    loss = Trace_ELBO()
+    guide = beta_guide
+    svi_run = SVI(conditioned_model, guide, opt, loss, num_steps=10000, num_samples=1000).run(num_trials)
+    posterior_predictive = TracePredictive(model, svi_run, num_samples=10000).run(num_trials)
```
neerajprad: Is it possible to use a higher learning rate and run for lesser steps? I am asking since these tests will run on each commit to Github, so a faster test suite makes the development workflow faster for everyone. 😄 Also if you have CUDA, could you run these tests on the GPU using the following command? We have a nightly cron job that runs our test suite on the GPU and find finely tuned precision thresholds on tests often get violated when run on the GPU.

```
CUDA_TEST=1 PYRO_TENSOR_TYPE=torch.cuda.DoubleTensor pytest -vs tests/infer/test_abstract_infer.py
```

pyro/poutine/posterior_messenger.py
ahmadsalim: Thanks for the comments! Relaxing the parameter values helped.
Funnily I was actually aiming to select parameter values of the posterior in the guide close to the actual values of the true posterior `300` and `700` I think?

tests/infer/test_abstract_infer.py
ahmadsalim: I tested it with CUDA and it seems to work. I also heavily decreased the number of steps required and increased the learning rate.

pyro/poutine/posterior_messenger.py
```diff
@@ -0,0 +1,12 @@
+from .replay_messenger import ReplayMessenger
+
+
+class SamplePosteriorMessenger(ReplayMessenger):
```
neerajprad: Now that I am thinking about it, I am not sure if this would work for the second case that you wanted to handle, since in that case what we need to do is to run the guide forward using the test data to get the correct batch shapes and then align them with the model. Replaying `cond_indep_stack` wouldn't solve this issue, as far as I can tell. Maybe we should write a test for this?

tests/infer/test_abstract_infer.py
```diff
@@ -28,7 +36,49 @@ def test_posterior_predictive():
...
+def test_posterior_predictive_svi_manual_guide():
+    true_probs = torch.ones(5) * 0.7
+    num_trials = torch.ones(5) * 1000
+    num_success = dist.Binomial(num_trials, true_probs).sample()
+    conditioned_model = poutine.condition(model, data={"obs": num_success})
+    opt = optim.Adam(dict(lr=1.0))
+    loss = Trace_ELBO()
+    guide = beta_guide
+    svi_run = SVI(conditioned_model, guide, opt, loss, num_steps=1000, num_samples=100).run(num_trials)
+    posterior_predictive = TracePredictive(model, svi_run, num_samples=10000).run(num_trials)
```
neerajprad: Let us run this using `.run(num_trials[:3])` and change the assertion below to `assert_close(marginal_return_vals.mean, torch.ones(3) * 700, rtol=0.05)`, to make sure that we are addressing the second case correctly.

pyro/infer/abstract_infer.py
```diff
@@ -285,23 +285,36 @@ class TracePredictive(TracePosterior):
     :param TracePosterior posterior: trace posterior instance holding
         samples from the model's approximate posterior.
     :param int num_samples: number of samples to generate.
+    :param (:obj:`list(str)`, optional) resample_sites: list of sites to resample from posterior.
+    All sites are resampled by default.
```
neerajprad: I think you'll need to indent this line for sphinx to build correctly. You can run `make docs` locally to see if it works fine.

pyro/poutine/posterior_messenger.py
ahmadsalim: Thanks for bringing it up.
I actually also discovered the issue yesterday 😄 (but I was doing some work-related programming first). I will see how to fix the issue.

neerajprad: No worries! Also, there is no need to address everything in this PR. `AbstractInfer` is due a bigger refactor, so this will help clarify what should go into that refactor (in addition to #1725 for instance). I think that making those classes lazy might be the solution to this issue.

ahmadsalim: Actually, I figured out how to do the fix before you responded, so I have tested it and pushed 😄. But a large refactoring to make it work better in the future would be very nice indeed!

pyro/infer/abstract_infer.py
```diff
@@ -285,23 +286,57 @@ class TracePredictive(TracePosterior):
...
     """
-    def __init__(self, model, posterior, num_samples):
+    def __init__(self, model, posterior, num_samples, resample_sites=None):
         self.model = model
         self.posterior = posterior
         self.num_samples = num_samples
+        self.resample_sites = resample_sites
         super(TracePredictive, self).__init__()
 
     def _traces(self, *args, **kwargs):
```
neerajprad: I don't think we should need the `resample_posterior` poutine at all? We can probably simplify this method to:

```python
        if not self.posterior.exec_traces:
            self.posterior.run(*args, **kwargs)
        data_trace = poutine.trace(self.model).get_trace(*args, **kwargs)
        for _ in range(self.num_samples):
            model_trace = self.posterior()
            self._adjust_to_data(model_trace, data_trace)
            replayed_trace = poutine.trace(poutine.replay(self.model, model_trace)).get_trace(*args, **kwargs)
            yield (replayed_trace, 0., 0)
```

pyro/infer/abstract_infer.py
```diff
@@ -285,23 +286,57 @@ class TracePredictive(TracePosterior):
...
+            # Adjust sites under conditionally independent stacks
+            try:
+                site["cond_indep_stack"] = data_trace.nodes[name]["cond_indep_stack"]
+                site["fn"] = data_trace.nodes[name]["fn"]
+                for cis in site["cond_indep_stack"]:
+                    # Select random sub-indices to replay values under conditionally independent stacks.
+                    # Otherwise, we assume there is an dependence of indexes between training data
+                    # and prediction data.
+                    cat = Categorical(logits=site["value"].new_ones(site["value"].size(cis.dim)))
+                    subidxs = cat.sample((cis.size,))
```
neerajprad: You can probably simplify this to `subidxs = Categorical(logits=site["value"].new_ones(cis.size)).sample()`.

pyro/infer/abstract_infer.py
```diff
@@ -285,23 +286,57 @@ class TracePredictive(TracePosterior):
     :param TracePosterior posterior: trace posterior instance holding
         samples from the model's approximate posterior.
     :param int num_samples: number of samples to generate.
+    :param list(str) resample_sites: sites to resample from in posterior predictive distribution.
```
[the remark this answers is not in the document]
ahmadsalim: The use case is if we have some data-local sampling sites that depend on global parameters, and which we want to resample instead of reusing parameters optimized/inferred for other data points in the training set. (E.g., we do a local Bernoulli trial whose parameters we learn globally, but which should be re-evaluated independently of whether we just had a small training set where the output was all `1` in the posterior).

pyro/infer/abstract_infer.py
ahmadsalim: OK, I will remove the `resample_posterior` poutine.

pyro/infer/abstract_infer.py
ahmadsalim: Thanks, will do!

ahmadsalim: Actually, it seems that the change breaks the tests. 
The reason actually is that what I want is that the `logits` to only have one value, which is the number of possible elements from the training posterior and then the sampling from the categorical ensures that we select enough random elements for the prediction.

For example, let us say we trained with `500` samples, then I want it to be a `Categorical(500)`. The sample statement will the adjust itself to the prediction size, e.g. if we have `300` samples in our test data, we will pick 300 random indices between 0 and 499.

pyro/infer/abstract_infer.py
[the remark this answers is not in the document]
neerajprad: I'm still not quite sure if that should belong to `TracePredictive` and I don't fully understand the use case. Maybe we can support such a use case with `exclude_sites` instead. The `resample_sites` might be very confusing to users. I would imagine someone specifying `resamples_sites = ["_RETURN"]`  and get a prior predictive rather than a posterior predictive distribution! It will also be nice if we have a test or an example (not necessarily in this PR), that illustrates the use case.

pyro/infer/abstract_infer.py
neerajprad: Sorry, I had mistyped. I meant - `subidxs = Categorical(logits=site["value"].new_ones(cis.size)).sample([cis.size])`. Thanks for explaining though, that makes sense.

pyro/infer/abstract_infer.py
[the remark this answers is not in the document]
ahmadsalim: It could certainly also be that I am misunderstanding something on how Pyro work, but I will try to add an example test case to clarify my use case 😄.

pyro/infer/abstract_infer.py
ahmadsalim: That seems to work, thanks 😄 .

pyro/infer/abstract_infer.py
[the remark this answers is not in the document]
ahmadsalim: It seems that I could not make up an example that worked like I wanted, so I removed it. Sorry, for the confusion.

pyro/infer/abstract_infer.py
ahmadsalim: Actually, the reason I am using the original is because I overwrite the context independent stack, so it will have the size of the evaluation data. But I want the categorical distribution to choose randomly according to the size of the training data (which I get from `site["value"].size(cis.dim)` instead). I will add a test where the evaluation data is larger than the training data as well 😄 .

neerajprad: Thanks for explaining, this is getting really confusing! :)

pyro/infer/abstract_infer.py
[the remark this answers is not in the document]
neerajprad: Np, if you have a use case that is not covered, I would suggest opening an issue so that we can keep track of it as we look to refactor this class.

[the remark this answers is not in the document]
ahmadsalim: Cool, thanks!

deoxyribose:
Hi, sorry in advance if I'm not following proper etiquette. I get the following error:
```
~/anaconda3/lib/python3.6/site-packages/pyro/infer/abstract_infer.py in _adjust_to_data(self, trace, data_trace)
    318                     # Otherwise, we assume there is an dependence of indexes between training data
    319                     # and prediction data.
--> 320                     subidxs = Categorical(logits=site["value"].new_ones(site["value"].size(cis.dim))).sample([cis.size])
    321                     site["value"] = site["value"].index_select(cis.dim, subidxs)
    322             except KeyError:

~/anaconda3/lib/python3.6/site-packages/torch/distributions/categorical.py in __init__(self, probs, logits, validate_args)
     52             if logits.dim() < 1:
     53                 raise ValueError("`logits` parameter must be at least one-dimensional.")
---> 54             self.logits = logits - logits.logsumexp(dim=-1, keepdim=True)
     55         self._param = self.probs if probs is not None else self.logits
     56         self._num_events = self._param.size()[-1]

RuntimeError: value cannot be converted to type int64_t without overflow: inf
```
The error is in abstract_infer.py in line 320, with the use of tensor.new_ones(), if the tensor is an int, the new tensor will also be int, which causes the error in logsumexp in categorical.py.

ahmadsalim:
@deoxyribose Thanks for catching the error. I am guessing the fix would be to use `torch.ones(site["value"].size(cis.dim), device=site["value"].device, dtype=torch.float)` instead?

ahmadsalim:
@deoxyribose Alternatively, one could consider calling `.float()` afterwards.
In any case, if you like you are welcome to make a new PR with the correction and a test example with integer random variables :smile:.
Alternatively, I can do so myself if you prefer.

Issue #1778, opened 2019-02-28
[discussion] Pyro sets all Parameters to requires_grad=True


cfperez:
### Issue Description
pyro.module() set `requires_grad=True` fo all `nn.Parameters` in a module, even if they were created with `requires_grad=False`.

Using `register_buffer()` solves this problem, but requires changing code that otherwise works in PyTorch.

It is hard to see how this would not be considered unexpected behavior. The question is how important is it, and if it is worth changing the behavior (and risk breaking other code?) or simply logging a warning + adding documentation.

### Environment
 - OSX, Python 3.6
 - PyTorch 1.0
 - Pyro 0.3.1

### Code Snippet

```py
import torch
import pyro

class Module(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.with_grad = torch.nn.Parameter(torch.zeros(1))
        self.no_grad = torch.nn.Parameter(torch.zeros(1), requires_grad=False)

m = Module()
assert m.with_grad.requires_grad
assert not m.no_grad.requires_grad

pyro_m = pyro.module('module', m)
assert pyro_m.with_grad.requires_grad
assert not pyro_m.no_grad.requires_grad
```

fritzo:
Thanks for the report @cfperez . Can you clarify your expected/desired behavior? When a PyTorch param has `.requires_grad == False` does the parameter simply not get updated during training? If that is the case, I guess we could ignore such params in `pyro.module` and possible emit a warning recommending use of `register_buffer()` (so as to maintain plausible backwards compatibility with previous versions of Pyro).

jpchen:
> does the parameter simply not get updated during training

yes, from [the pytorch docs](https://pytorch.org/docs/stable/notes/autograd.html#excluding-subgraphs)

> we could ignore such params in `pyro.module` and possible emit a warning recommending use of `register_buffer()`

i can make this fix; i think this might have been a later pytorch feature we didnt account for - i seem to remember that when `module` was implemented in pyro (~pytorch 0.2), `Parameter`s were defined to be `Variable`s with `requires_grad = True`.

fritzo:
@jpchen I believe the correct behavior is to **skip** calling `pyro.param` inside `pyro.module` whenever the param has `.requires_grad == False`. Specifically, I believe changes should be limited to `pyro.module`, and there should be no changes required to `pyro.param`. Let me know if you disagree. Otherwise, thanks for implementing this!

jpchen:
[...]

PR #1779, opened 2019-03-01
`module` will ignore params with requires_grad=False

jpchen:
Addresses #1778

### DIFF
--- a/pyro/primitives.py
+++ b/pyro/primitives.py
@@ -255,7 +255,7 @@ def module(name, nn_module, update_module_params=False):
                 target_state_dict[param_name] = returned_param
         else:
             warnings.warn("{} was not registered in the param store because".format(param_name) +
-                          "requires_grad=False")
+                          " requires_grad=False")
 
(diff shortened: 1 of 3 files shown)


pyro/primitives.py
```diff
@@ -245,13 +245,17 @@ def module(name, nn_module, update_module_params=False):
...
+        if param_value.requires_grad:
+            # register the parameter in the module with pyro
+            # this only does something substantive if the parameter hasn't been seen before
+            full_param_name = param_with_module_name(name, param_name)
+            returned_param = param(full_param_name, param_value)
+
+            if param_value._cdata != returned_param._cdata:
+                target_state_dict[param_name] = returned_param
+        else:
+            warnings.warn("{} was not registered in the param store because".format(param_name) +
```
neerajprad: nit: space between `because` and `requires_grad` in the warning message.

neerajprad:
The ss-vae example seems to be failing for some reason, but it doesn't seem related.

PR #1783, opened 2019-03-18
elbo with tail adaptive f-divergence

martinjankowiak:
as detailed in this [paper](https://papers.nips.cc/paper/7816-variational-inference-with-tail-adaptive-f-divergence)

invoke as follows: ```TraceTailAdaptive_ELBO(vectorize_particles=True, num_particles=10, tail_adaptive_beta=-1.0)```

### DIFF
--- a/pyro/infer/elbo.py
+++ b/pyro/infer/elbo.py
@@ -45,6 +45,8 @@ class ELBO(object):
         be ignored. Defaults to False.
     :param bool retain_graph: Whether to retain autograd graph during an SVI
         step. Defaults to None (False).
+    :param float tail_adaptive_beta: Exponent beta with -1.0 >= beta > 0.0 for
+        use with `TraceTailAdaptive_ELBO`.
 
     References
 
@@ -62,7 +64,8 @@ class ELBO(object):
                  vectorize_particles=False,
                  strict_enumeration_warning=True,
                  ignore_jit_warnings=False,
-                 retain_graph=None):
+                 retain_graph=None,
+                 tail_adaptive_beta=-1.0):
         if max_iarange_nesting is not None:
             warnings.warn("max_iarange_nesting is deprecated; use max_plate_nesting instead",
                           DeprecationWarning)
@@ -75,6 +78,7 @@ class ELBO(object):
             self.max_plate_nesting += 1
         self.strict_enumeration_warning = strict_enumeration_warning
         self.ignore_jit_warnings = ignore_jit_warnings
+        self.tail_adaptive_beta = tail_adaptive_beta
 
     def _guess_max_plate_nesting(self, model, guide, *args, **kwargs):
         """
--- a/pyro/infer/elbo.py
+++ b/pyro/infer/elbo.py
@@ -45,7 +45,7 @@ class ELBO(object):
         be ignored. Defaults to False.
     :param bool retain_graph: Whether to retain autograd graph during an SVI
         step. Defaults to None (False).
-    :param float tail_adaptive_beta: Exponent beta with -1.0 <= beta < 0.0 for
+    :param float tail_adaptive_beta: Exponent beta with ``-1.0 <= beta < 0.0`` for
         use with `TraceTailAdaptive_ELBO`.
 
     References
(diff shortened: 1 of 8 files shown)


pyro/infer/util.py
```diff
@@ -267,3 +267,11 @@ def compute_expectation(self, costs):
 
         LAST_CACHE_SIZE[0] = count_cached_ops(cache)
         return expected_cost
+
+
+def _check_fully_reparametrized(guide_site):
```
fritzo: If this is used outside of the file in which it defined, then its name should not begin with underscore.

martinjankowiak: thanks, done

tests/infer/test_inference.py
```diff
@@ -52,6 +52,9 @@ def test_elbo_reparameterized(self):
     def test_elbo_analytic_kl(self):
         self.do_elbo_test(True, 3000, TraceMeanField_ELBO())
 
+    def test_elbo_tail_adaptive(self):
+        self.do_elbo_test(True, 3000, TraceTailAdaptive_ELBO(num_particles=10, vectorize_particles=True))
```
fritzo: If `TraceTailAdaptive_ELBO` supports nonreparametrized samplers (as your test_valid_models test suggests), then you should add a numerical test that nonreparametrized distributions work. Could you fork this test case?

martinjankowiak: it doesn't support nonreparametrized samplers (as mentioned in docstring). please see test again. unless i'm missing something?

martinjankowiak:
well i think test_subsample_gradient assumes KL divergence, so not applicable

i can add something with a plate though that makes sense

tests/infer/test_valid_models.py
```diff
@@ -1613,16 +1613,22 @@ def guide():
...
         pyro.sample("x", dist.Normal(0., 2.))
 
     def nonrep_guide():
         pyro.sample("x", fakes.NonreparameterizedNormal(0., 2.))
 
-    assert_ok(model, rep_guide, TraceTailAdaptive_ELBO(vectorize_particles=True, num_particles=2))
-    assert_error(model, rep_guide, TraceTailAdaptive_ELBO(vectorize_particles=False, num_particles=2))
-    assert_warning(model, rep_guide, TraceTailAdaptive_ELBO(vectorize_particles=True, num_particles=1))
-    assert_error(model, nonrep_guide, TraceTailAdaptive_ELBO(vectorize_particles=True, num_particles=2))
+    assert_ok(plateless_model, rep_guide, TraceTailAdaptive_ELBO(vectorize_particles=True, num_particles=2))
```
fritzo: FYI it's generally useful to split assertions into separate tests so that a failure of the first does not prevent testing of all assertions following. Often we can do this with parametrized tests. Feel free to split this up or not.

pyro/infer/trace_tail_adaptive_elbo.py
```diff
@@ -0,0 +1,68 @@
...
+
+import torch
+
+from pyro.infer.trace_elbo import Trace_ELBO
+from pyro.infer.util import is_validation_enabled, check_fully_reparametrized
+
+
+class TraceTailAdaptive_ELBO(Trace_ELBO):
+    """
+    Interface for Stochastic Variational Inference with an adaptive
```
fritzo: nit: it would be nice to split this into three paragraphs: summary, note on nonreparam, note on loss computation.

pyro/infer/elbo.py
```diff
@@ -45,6 +45,8 @@ class ELBO(object):
         be ignored. Defaults to False.
     :param bool retain_graph: Whether to retain autograd graph during an SVI
         step. Defaults to None (False).
+    :param float tail_adaptive_beta: Exponent beta with -1.0 >= beta > 0.0 for
```
jpchen: `>` -> `<`?

martinjankowiak: whoops

fritzo: Also you'll probably want to enclose these in double ticks
```
Exponent beta with ``-1 <= beta < 0`` for ...
```

martinjankowiak: done

martinjankowiak:
should be good to merge; build error is some unrelated flake

PR #1790, opened 2019-03-25
Add a pyro.generic module to enable backend-agnostic Pyro models

fritzo:
This adds a new `pyro.generic` module to enable dynamic interpretation of `pyro.*`, `dist.*` and a couple other modules.

## Motivation
As we build new Pyro backends, it will be nice to write Pyro models that can be used across backends, similar the proposed [NumPy __array_function__ protocol](https://mail.python.org/pipermail/numpy-discussion/2018-June/078127.html) . Currently Pyro models require hard-coding the backend via `import` statements. This PR adds dynamically dispatched `pyro.generic.*` module wrappers that can be reinterpreted a runtime. For example we can
```py
from pyro.generic import backend, pyro
from pyro.generic import distributions as dist

def training_loop():
    ...standard pyro code...

with backend('pyro'):
    expected = training_loop()
with backend('minipyro'):
    actual = training_loop()
assert torch.allclose(actual, expected)
```

## Tests
- updated examples/minipyro.py to use this new interface


pyro/generic.py
```diff
@@ -0,0 +1,85 @@
...
+        'infer': 'pyro.contrib.minipyro',
+        'optim': 'pyro.contrib.minipyro',
+    },
+    'funsor': {
+        'pyro': 'funsor.minipyro',
+        'infer': 'funsor.minipyro',
+        'optim': 'funsor.minipyro',
+        'distributions': 'funsor.distributions',
+    },
+    'numpy': {
```
fritzo: @neerajprad Could you try to target this interface in `numpyro`? Feel free to push to this PR to add other module renames, e.g. `infer`, `optim` or others.

neerajprad:
I think rather than enforcing any constraints on the `distribution`, `optim` or other modules inside of `numpyro`, it may desirable to have a `numpyro.compatibility` module which can target this interface. That way, the code inside of `numpyro` can stay completely functional that works best with JAX without being encumbered by other constraints, and all the boilerplate can reside in `compatibility` module. I think this looks good to me otherwise. 

There are still quite a few things that I'm not completely sure how they should be handled. e.g.
 - there is no global param store, so to get something like `pyro.get_param_store` to work, we'll need to simply cache the optimizer state somewhere towards the end of svi training. 
 - the loop `for ..: svi.step()` will be suboptimal, and should ideally be replaced by `lax.fori_loop` when users are on the JAX backend. This could be an order of magnitude faster!
 - there is no `pyro.module`, because the stax NNs don't have any parameters that need to be registered. All parameters need to be cycled back and forth with the update function.

I'll probably have more inputs when I write the `compatibility` module and get the minipyro example working, and we can have a more detailed discussion then!

pyro/generic.py
fritzo: > @neerajprad I think ... it may desirable to have a numpyro.compatibility module

Sure, do you want to update this dict to be something less restrictive?
```py
'numpy': {
    'pyro': 'numpyro.compat.pyro',
    'distributions': 'numpyro.compat.distributions',
    'infer': 'numpyro.compat.infer',
    'optim': 'numpyro.compat.optim',
}
```

neerajprad:
I added some of these points to https://github.com/pyro-ppl/numpyro/issues/66 in Numpyro. I'll update that as I get to work on the compatibility wrapper.

fritzo:
@eb8680 I'd like to get your feedback before merging, since you've thought about this before.

PR #1792, opened 2019-03-26
AutoIAFGuide bug?

stefanwebb:
I think there is a bug in AutoIAFGuide. I only get learning in the IAF when I set `update_module_params=True` inside the `pyro.module`


pyro/contrib/autoguide/__init__.py
```diff
@@ -265,6 +268,7 @@ class AutoDelta(AutoGuide):
         pyro.param("auto_concentration", torch.ones(k),
                    constraint=constraints.positive)
     """
+
```
martinjankowiak: @stefanwebb can you please try to get your flake8 to not flag these spaces? i don't think we want to add spaces to every part of the codebase you interact with

stefanwebb: Yes! I think this is caused by my system inserting an extra character for Windows line endings (CR + LF) versus Unix line endings that are just LF

PR #1794, opened 2019-03-28
Add a Trace_ELBO wrapper to minipyro

fritzo:
This adds a `Trace_ELBO()` wrapper to pyro.contrib.minipyro to make it easier to write backend-agnostic code. This is important as we implement more backends.

This PR also renames `backend` to `pyro_backend` to avoid conflict with (1) local variables named `backend` and (2) the `pyro` module, because we can't mix fully qualified and partially qualified names:
```py
# This does not work:
import pyro.generic.backend  # to avoid conflict with backend
from pyro.generic import pyro
```

## Tested
- covered by tests/test_examples.py

### DIFF
--- a/examples/minipyro.py
+++ b/examples/minipyro.py
@@ -11,7 +11,7 @@ import argparse
 import torch
 
 # We use the pyro.generic interface to support dynamic choice of backend.
-from pyro.generic import backend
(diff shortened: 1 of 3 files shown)


fritzo:
@neerajprad you may want to copy these changes into numpyro's minipyro

PR #1795, opened 2019-03-29
Householder flow

stefanwebb:
Householder flow is a way of approximating a multivariate Gaussian distribution by passing i.i.d. noise through a series of transformations.


stefanwebb:
This one is ready for review! It's part of my aim to have a complete normalizing flows library in Pyro

pyro/distributions/householder.py
```diff
@@ -0,0 +1,101 @@
...
+        # If you reflect a point around a plane, then the same operation will reflect it back
+        return self._call(y)
+
+    def log_abs_det_jacobian(self, x, y):
+        """
+        Calculates the elementwise determinant of the log jacobian
+        """
+
+        # Householder flow is measure preserving, so log(|detJ|) = 0
+        return torch.zeros(x.size()[:-1])
```
fritzo: I think you'll want `x.new_zeros(x.size()[:-1])` preserve `x.dtype` and `x.device`.

stefanwebb: Thanks for catching that! I've fixed this and the same problem in `permute.py` in the new commit

stefanwebb:
I've just added the ability to apply multiple transforms of Householder flow using the same object, which should cut down overhead when using Householder flow in high-dimensional spaces

pyro/distributions/householder.py
```diff
@@ -0,0 +1,109 @@
...
+    where :math:`\\mathbf{x}` are the inputs, :math:`\\mathbf{y}` are the outputs, and the learnable parameters
+    are :math:`\\mathbf{u}\\in\\mathbb{R}^D` for input dimension :math:`D`.
+
+    The transformation represents the reflection of :math:`\\mathbf{x}` through the plane passing through the
+    origin with normal :math:`\\mathbf{u}`. Together with `TransformedDistribution` this provides a way to
+    create richer variational approximations.
+
+    :math:`D` applications of this transformation are able to transform standard i.i.d. standard Gaussian noise
+    into a Gaussian variable with an arbitrary covariance matrix. With :math:`K<D` transformations, one is able
+    to approximate a full-rank covariance Gaussian distribution.
```
martinjankowiak: rank K?

pyro/distributions/householder.py
```diff
@@ -0,0 +1,109 @@
...
+
+    Example usage:
+
+    >>> base_dist = dist.Normal(torch.zeros(10), torch.ones(10))
+    >>> flows = [HouseholderFlow(10) for _ in range(3)]
+    >>> [pyro.module("my_flow", p) for f in flows] # doctest: +SKIP
```
martinjankowiak: update docstring to create a single flow with multiple reflections?

pyro/distributions/householder.py
```diff
@@ -0,0 +1,109 @@
...
+    event_dim = 1
+
+    def __init__(self, input_dim, count_transforms=1):
+        super(HouseholderFlow, self).__init__(cache_size=1)
+
+        self.input_dim = input_dim
+        assert count_transforms > 0
+        if count_transforms > input_dim:
+            warnings.warn(
+                "{} layers of Householder flow applied to input of dimension {}".format(
```
martinjankowiak: explain why this is unnecessary?

pyro/distributions/householder.py
stefanwebb: Done!

pyro/distributions/householder.py
stefanwebb: Done!

pyro/distributions/householder.py
stefanwebb: I put this in because I don't think there is an advantage to multiplying more than D Householder matrices for input dim D. Full-rank is as high as you can go!

stefanwebb:
Okay, I've made those changes!

pyro/distributions/householder.py
stefanwebb: Ah I see you meant that I should add in an explanatory note to the user. I've done this now!

stefanwebb:
@martinjankowiak this is ready to be looked at again

pyro/distributions/householder.py
```diff
@@ -0,0 +1,111 @@
...
+        :param x: the input into the bijection
+        :type x: torch.Tensor
+
+        Invokes the bijection x=>y; in the prototypical context of a TransformedDistribution `x` is a
+        sample from the base distribution (or the output of a previous flow)
+        """
+
+        y = x
+        for idx in range(self.count_transforms):
+            squared_norm = self.u[idx].pow(2).sum(-1)
```
martinjankowiak: this line can be pulled outside the loop and done in parallel?

`squared_norm = self.u[idx].pow(2).sum(-1)`

stefanwebb: I made this change

PR #1796, opened 2019-03-30
Subsample pyro.param inside plate if event_dim is given

### LINKED ISSUE #238: Params inside iarange
the current behavior of params created inside an `iarange` (aka vectorized map_data) is a bit nonsensical: a vector of params of length `batch_size` will be created and used on every iteration. this means that a param gets reused for whatever data point happens to be selected at that index in a batch. this comes up eg when there is a mean-field model for a local variable.

(there is a reasonable work-around, by creating the vector of needed params globally, and only selecting from it in the iarange.)

a better behavior would be to automatically associate params to their index into the data tensor, pulling out the appropriate params for a given batch (and creating them as needed). it might be tricky to get this working, and especially to do it efficiently.

i'm flagging this as a low priority bug for now. we can think through a solution after launch.

fritzo:
Fixes #238 

This adds a new optional kwarg `event_dim` to `pyro.param` statements. If `event_dim` is not specified, previous behavior is preserved i.e. all dims are considered event dims.

When `event_dim` is specified (e.g. 0, 1, 2), dims to the left are considered batch dims. When a param statement occurs inside a `pyro.plate`, the `event_dim` argument gives Pyro enough information to subsample the parameters (via `.index_select(plate.dim - event_dim, plate.ind)`).

This PR also anticipates named dimensions (possibly coming in PyTorch 1.2). By specifying `event_dim`, Pyro will have enough information to annotate plate names to the dims of a param tensor.

## Tested
- [x] tests of shape in test_valid_models.py
- [x] updated test_gradients.py to use this slick new syntax
- [x] add test of shape error

### DIFF
--- a/pyro/poutine/subsample_messenger.py
+++ b/pyro/poutine/subsample_messenger.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function
 import torch
 
 from pyro.distributions.distribution import Distribution
+from pyro.poutine.util import is_validation_enabled
 from pyro.util import ignore_jit_warnings, jit_compatible_arange
 
 from .indep_messenger import CondIndepStackFrame, IndepMessenger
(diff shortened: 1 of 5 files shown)


pyro/poutine/subsample_messenger.py
```diff
@@ -117,7 +117,7 @@ def _subsample(name, size=None, subsample_size=None, subsample=None, use_cuda=No
         return size, subsample_size, subsample
 
     def _reset(self):
-        self.subsample = None
+        self._indices = None
```
fritzo: This was apparently a typo. I'm not sure when this is used, however.

pyro/poutine/subsample_messenger.py
```diff
@@ -128,3 +129,20 @@ def _process_message(self, msg):
                 with ignore_jit_warnings():
                     msg["scale"] = torch.tensor(msg["scale"])
         msg["scale"] = msg["scale"] * self.size / self.subsample_size
+
+    def _postprocess_message(self, msg):
+        if msg["type"] == "param":
+            event_dim = msg["kwargs"].get("event_dim")
+            if event_dim is not None:
+                dim = self.dim - event_dim
```
eb8680: You'll need to check `self.dim is not None` to handle the case of sequential plates

tests/infer/test_valid_models.py
```diff
@@ -215,6 +215,62 @@ def guide():
...
+    assert_ok(model, guide, Elbo())
+
+
+@pytest.mark.parametrize("subsample_size", [None, 5], ids=["full", "subsample"])
+@pytest.mark.parametrize("Elbo", [Trace_ELBO, TraceGraph_ELBO, TraceEnum_ELBO])
+@pytest.mark.parametrize("shape,ok", [
+    ((), True),
+    ((1,), True),
+    ((10,), True),
+    ((3, 1), True),
```
neerajprad: The logic for param shapes looks very consistent, but just for my understanding, is there a use case that you have in mind for `(k, 1)` shaped params inside an enclosing plate?

fritzo: The `k,1` might happen if there is another plate. I guess we could error if not all batch dims are accounted for.

PR #1799, opened 2019-04-02
Run without raising exceptions when HMC contains no stochastic site

neerajprad:
This makes sure that when HMC is presented with a model that does not contain stochastic sites, it simply returns the trace without throwing any exceptions. Many parts of our code however were designed with the assumption that there is at least a single stochastic site.

This situation can arise when we use conjugate distributions and collapse the latent site. In such cases, models with a single latent variable might have no latent variable when the model is wrapped with a handler that collapses such sites.

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -4,7 +4,6 @@ import math
 from collections import OrderedDict
 
 import torch
-from torch import tensor
 from torch.distributions import biject_to, constraints
 
 import pyro
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -4,6 +4,7 @@ import math
 from collections import OrderedDict
 
 import torch
+from torch import tensor
 from torch.distributions import biject_to, constraints
 
 import pyro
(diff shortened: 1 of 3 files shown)


tests/infer/mcmc/test_mcmc.py
```diff
@@ -86,3 +87,22 @@ def test_num_chains(num_chains, cpu_count, monkeypatch):
         assert isinstance(mcmc.sampler, _SingleSampler)
     else:
         assert isinstance(mcmc.sampler, _ParallelSampler)
+
+
+def _empty_model():
+    return torch.ones([])
```
neerajprad: One really odd thing that I am noticing is that a `torch.tensor` constructor throws an error under JIT, so if we change this to `torch.tensor(1)` we will get an error when `test_empty_sample_sites` is run under JIT (same for line 150 in hmc.py). This is rather strange, but it is not replicable. It runs fine with the debugger and I am unable to isolate this into a minimal example to present to PyTorch devs.

fehiepsi:
@neerajprad Could you add a test with hmc kernel too?

neerajprad:
> @neerajprad Could you add a test with hmc kernel too?

The test is `test_mcmc` has both the kernels.

pyro/infer/mcmc/hmc.py
```diff
@@ -144,10 +144,19 @@ def _iter_latent_nodes(trace):
...
+                value = node.get("value", None)
+                if value is not None:
+                    device = value.device
+            return torch.zeros([], device=device)
         return self._trace_prob_evaluator.log_prob(model_trace)
 
     def _kinetic_energy(self, r):
-        r_flat = torch.cat([r[site_name].reshape(-1) for site_name in sorted(r)])
+        r_flat = torch.cat([r[site_name].reshape(-1) for site_name in sorted(r)]) if r \
+            else torch.tensor([])
```
fritzo: I'd recommend avoiding `torch.tensor([])` if you can; PyTorch does not have full support for empty tensors, i.e. many parts of the codebase appear to assume nonempty tensors. I would not be surprised if this was the line leading to your jit error.

fritzo: For example, can you return `None` instead?

neerajprad: If we return None, we will need to short-circuit many parts of the code by checking that z is empty (or that KE is not None), which I wanted to avoid. But we can just return `0.` here.

That still does not resolve the JIT error when instead of `torch.zeros([])` we use `torch.tensor(0.)` anywhere in the model or the HMC code.

PR #1800, opened 2019-04-02
helper function for fitting generalized pareto distributions

martinjankowiak:
as a first step towards implementing some VI diagnostics along the lines of [Yes, but Did It Work?](https://arxiv.org/abs/1802.02538) we adapt some [code](https://github.com/avehtari/PSIS/blob/master/py/psis.py#L211) for fitting pareto distributions

### DIFF
--- a/pyro/ops/stats.py
+++ b/pyro/ops/stats.py
@@ -3,6 +3,8 @@ from __future__ import absolute_import, division, print_function
 import numbers
 
 import torch
+import math
+import numpy as np
 
 
 def _compute_chain_variance_stats(input):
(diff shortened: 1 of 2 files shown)


pyro/ops/stats.py
```diff
@@ -342,3 +344,63 @@ def waic(input, log_weights=None, pointwise=False, dim=0):
...
+    Zhang, J. and Stephens, M.A. (2009).
+    [2] 'Pareto Smoothed Importance Sampling.'
+    Aki Vehtari, Andrew Gelman, Jonah Gabry
+    https://github.com/avehtari/PSIS/blob/904146236767182270c9718a7b2a30831fe701fe/py/psis.py#L211
+
+    :param torch.Tensor or numpy.ndarray X: the input data X
+    :returns tuple: tuple of floats (k, sigma) corresponding to the fit parameters
+    """
+    if isinstance(X, torch.Tensor):
+        X = X.data.cpu().float().numpy()
```
fritzo: Why the `.float()`? I believe that will convert double-precision to single-precision. Could we instead simply
```py
X = X.detach().cpu().numpy()
```

martinjankowiak: i implemented in numpy because the original code i adapted from is in numpy. i wanted to keep the changes minimal. (for this reason there is no read to review the math). also this isn't something where performance matters much. we feed it ~1000 importance weights, do a quick computation, report a thing, and move on.

pyro/ops/stats.py
```diff
@@ -342,3 +344,63 @@ def waic(input, log_weights=None, pointwise=False, dim=0):
...
+    if X.ndim != 1 or len(X) <= 1:
+        raise ValueError("Invalid input array.")
+
+    X.sort()
+
+    N = len(X)
+    PRIOR = 3
+    M = 30 + int(math.sqrt(N))
+
+    bs = np.arange(1, M + 1, dtype=float) - 0.5
```
martinjankowiak: i was doing float for simplicity so i don't have variable dtype logic downstream

PR #1804, opened 2019-04-04
Add vectorized importance weight helper

eb8680:


pyro/infer/importance.py
```diff
@@ -79,3 +82,30 @@ def get_ESS(self):
             warnings.warn("The log_weights list is empty, effective sample size is zero.")
             ess = 0
         return ess
+
+
+def vectorized_importance_weights(model, guide, num_samples, *args, **kwargs):
+
+    with plate("num_particles_vectorized", num_samples, dim=-10):
```
martinjankowiak: put a note about max plate nesting?

pyro/infer/importance.py
```diff
@@ -79,3 +81,43 @@ def get_ESS(self):
             warnings.warn("The log_weights list is empty, effective sample size is zero.")
             ess = 0
         return ess
+
+
+def vectorized_importance_weights(model, guide, *args, **kwargs):
+
+    num_samples = kwargs.pop("num_samples", 1)
+    max_plate_nesting = kwargs.pop("max_plate_nesting", 7)
```
neerajprad: nit: I think it might be safer to let this throw an exception if `max_plate_nesting` isn't provided than to assume a default value.

PR #1806, opened 2019-04-04
Adding support for exact inference in HMC with conjugate pairs

neerajprad:
This resurrects the previously closed #1716, with a cleaner and more general interface. I have put all the code in `contrib.conjugate` since this handles a restricted set of cases where a single pair of latent and observed sites are conjugate (the more general case will be handled by funsors, and is outside the purview of this PR). In such cases, the model can be written as (e.g. a gamma-poisson pair):

```python
gamma_poisson = GammaPoissonPair()
# Sample latent rate
rate = pyro.sample("rate", gamma_poisson.latent(alpha, beta))
    with pyro.plate("data", N):
        pyro.sample("obs", gamma_poisson.conditional(rate), obs=data)
```

When doing inference we can use `collapse_conjugate(model)` to remove the latent `rate` variable from the trace and replace the observed site with a `GammaPoisson` compound distribution. We can subsequently use the `uncollapse_conjugate(model, trace)` handler which is the same as `replay` except that it also uncollapses any latent sites by instead sampling from the posterior (exact) distribution. Note that this requires knowing which sites are paired as conjugate distributions by the two handlers, which is handled by the `BetaBinomialPair` and `GammaPoissonPair` classes.

### DIFF
--- a/pyro/contrib/conjugate/infer.py
+++ b/pyro/contrib/conjugate/infer.py
@@ -68,8 +68,15 @@ class BetaBinomialPair(object):
         total_count = self._conditional.total_count
         reduce_dims = len(obs.size()) - len(concentration1.size())
         num_obs = reduce(mul, obs.size()[:reduce_dims], 1)
-        total_count = total_count[tuple(slice(1) if i < reduce_dims else slice(None)
-                                  for i in range(total_count.dim()))].reshape(concentration0.shape)
+        # Unexpand total_count to have the same shape as concentration0.
+        # Raise exception if this isn't possible.
+        total_count = total_count[tuple(slice(None) if stride else slice(1)
+                                  for stride in total_count.stride())]
+        try:
+            total_count = total_count.reshape(concentration0.shape)
+        except RuntimeError:
+            raise ValueError("`total_count.dim() = {}` cannot be unexpanded to `concentration0.dim() = {}`."
+                             .format(total_count.dim(), concentration0.dim()))
         summed_obs = sum_leftmost(obs, reduce_dims)
         return dist.Beta(concentration1 + summed_obs,
                          num_obs * total_count + concentration0 - summed_obs,
--- a/pyro/contrib/conjugate/infer.py
+++ b/pyro/contrib/conjugate/infer.py
@@ -7,42 +7,50 @@ from pyro.poutine.messenger import Messenger
 from pyro.poutine.replay_messenger import ReplayMessenger
 
 
-class _LatentBeta(dist.Beta):
+class _Beta(dist.Beta):
     collapsible = True
 
     def __init__(self, parent, *args, **kwargs):
         self.parent = parent
-        super(_LatentBeta, self).__init__(*args, **kwargs)
+        super(_Beta, self).__init__(*args, **kwargs)
+
+    def expand(self, batch_shape, _instance=None):
+        new = self._get_checked_instance(_Beta, _instance)
+        new.parent = self.parent
+        return super(_Beta, self).expand(batch_shape, _instance=new)
 
 
-class _CondBinomial(dist.Binomial):
+class _Binomial(dist.Binomial):
     marginalize_latent = True
 
     def __init__(self, parent, probs, **kwargs):
         self.parent = parent
         self.latent_param = probs
-        super(_CondBinomial, self).__init__(probs=probs, **kwargs)
+        super(_Binomial, self).__init__(probs=probs, **kwargs)
+
+    def expand(self, batch_shape, _instance=None):
+        new = self._get_checked_instance(_Binomial, _instance)
+        new.parent = self.parent
+        return super(_Binomial, self).expand(batch_shape, _instance=new)
 
 
-class BetaBinomialConjugate(object):
+class BetaBinomialPair(object):
     def __init__(self):
         self._latent = None
-        self._conditioned = None
-        self.cond_kwargs = {}
+        self._conditional = None
 
     def latent(self, *args, **kwargs):
-        self._latent = _LatentBeta(self, *args, **kwargs)
+        self._latent = _Beta(self, *args, **kwargs)
         return self._latent
 
-    def conditioned(self, probs, **kwargs):
-        self.cond_kwargs = kwargs
-        self._conditioned = _CondBinomial(self, probs, **kwargs)
-        return self._conditioned
+    def conditional(self, probs, **kwargs):
+        self._conditional = _Binomial(self, probs, **kwargs)
+        return self._conditional
 
     def posterior(self, obs):
         concentration1 = self._latent.concentration1
         concentration0 = self._latent.concentration0
-        total_count = self.cond_kwargs["total_count"]
+        total_count = self._conditional.total_count
         reduce_dims = len(obs.size()) - len(concentration1.size())
         num_obs = reduce(mul, obs.size()[:reduce_dims], 1)
         total_count = total_count[tuple(slice(1) if i < reduce_dims else slice(None)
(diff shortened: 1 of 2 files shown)


pyro/contrib/conjugate/infer.py
```diff
@@ -0,0 +1,167 @@
...
+        return self._conditional
+
+    def posterior(self, obs):
+        concentration1 = self._latent.concentration1
+        concentration0 = self._latent.concentration0
+        total_count = self._conditional.total_count
+        reduce_dims = len(obs.size()) - len(concentration1.size())
+        num_obs = reduce(mul, obs.size()[:reduce_dims], 1)
+        total_count = total_count[tuple(slice(1) if i < reduce_dims else slice(None)
+                                  for i in range(total_count.dim()))].reshape(concentration0.shape)
```
fehiepsi: could you add a comment in the code about the purpose of this computation?

pyro/contrib/conjugate/infer.py
```diff
@@ -0,0 +1,167 @@
...
+        summed_obs = sum_leftmost(obs, reduce_dims)
+        return dist.Beta(concentration1 + summed_obs,
+                         num_obs * total_count + concentration0 - summed_obs,
```
fehiepsi: this is right if total_count is constant but I am not sure if it is still true when we have batch involved (because I don't understand the computation of total count in a few lines above.

pyro/contrib/conjugate/infer.py
neerajprad: This is basically trying to unexpand `total_count` to have the same shape as `concentration0`. I have added a comment and also made it so that it will throw an error if this unexpand fails. I have also changed the test to have non-uniform `total_count`.

pyro/contrib/conjugate/infer.py
neerajprad: Good point. I think the requirement is that `total_count` shouldn't vary beyond the first k dims counting from the right, where `k = concentration0.dim()`. I have changed the `total_count` in the test below to illustrate this. If the `total_count` varies beyond the k dims, then this will throw a ValueError.

pyro/contrib/conjugate/infer.py
```diff
@@ -0,0 +1,167 @@
...
+    def conditional(self, *args, **kwargs):
+        self._conditional = _conditional(dist.Binomial, parent=self)(*args, **kwargs)
+        return self._conditional
+
+    def posterior(self, obs):
+        concentration1 = self._latent.concentration1
+        concentration0 = self._latent.concentration0
+        total_count = self._conditional.total_count
+        reduce_dims = len(obs.size()) - len(concentration1.size())
+        num_obs = reduce(mul, obs.size()[:reduce_dims], 1)
```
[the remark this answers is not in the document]
neerajprad: Seems like this is available in pytorch 1.0 so I just changed this.

pyro/contrib/conjugate/infer.py
fehiepsi: According to [wikipedia](https://en.wikipedia.org/wiki/Conjugate_prior) posterior formula for BetaBinomial, I think that we need to `sum_leftmost` of `total_count` until it matches `concentration0.shape`. I don't understand why we do unexpand then multiply with num_obs here. Could you please elaborate more on it?

fehiepsi: I think that it would be enough to define
```
total_count = sum_leftmost(total_count, len(total_count.size()) - len(concentration1.size()))
```

neerajprad: > I think that we need to sum_leftmost of total_count until it matches concentration0.shape. I don't understand why we do unexpand then multiply with num_obs here. Could you please elaborate more on it?

I think the issue here is that `total_count` will have the same shape as the observed data which might have a larger shape than `concentration0`, so we need to reduce it down appropriately. Suppose we have observations of shape `(5, 2) + concentration0.shape`. Suppose `total_count` is constant = 10. Then in actuality for each element in `concentration0`, we `10 (total_count) * 5 * 2  (left batch size) = 100 (num_obs)` observations, and the number of successes is given by `summed_obs =  sum_leftmost(obs, reduce_dims)`. That should be the same as the wikipedia formula? Let me know if that is still confusing.

fehiepsi: If we do unexpand for a constant tensor total_count=10, then it can not be reshaped into `concentration0.shape`. On the other hand, if `total_count.shape = obs.shape = probs.shape + 1`, then we can't do inference (while mathematically possible) because of the requirement that `total_count` shouldn't vary beyond the first k dims counting from the right.

neerajprad: Note that total_count would be expanded internally by the distribution so it should be possible to reshape it into `concentration0.shape` if the leftmost dims come from expansion, but you are right that it is unnecessary and can be more general. Sorry, I missed your previous comment. I'm not even sure why I even had that in the previous PR. :) Just summing up the left dims should suffice!

neerajprad:
Thanks for a thorough review, @fehiepsi!

pyro/contrib/conjugate/infer.py
```diff
@@ -0,0 +1,164 @@
...
+class UncollapseConjugateMessenger(ReplayMessenger):
+    r"""
+    Extends `~pyro.poutine.replay_messenger.ReplayMessenger` to uncollapse
+    compound distributions.
+    """
+    def _pyro_sample(self, msg):
+        is_collapsible = getattr(msg["fn"], "collapsible", False)
+        if is_collapsible:
+            observed_node, conj_pair = None, None
+            for site_name in self.trace.observation_nodes:
```
fritzo: Could you also save the conditional `.site_name` and lookup
```py
conditional_name = msg["fn"].parent._conditional.site_name
conditional_node = self.trace[conditional_name]
```

pyro/contrib/conjugate/infer.py
```diff
@@ -0,0 +1,164 @@
...
+
+class UncollapseConjugateMessenger(ReplayMessenger):
+    r"""
+    Extends `~pyro.poutine.replay_messenger.ReplayMessenger` to uncollapse
+    compound distributions.
+    """
+    def _pyro_sample(self, msg):
+        is_collapsible = getattr(msg["fn"], "collapsible", False)
+        if is_collapsible:
+            observed_node, conj_pair = None, None
```
fritzo: Hmm wouldn't this still work if the conditional site were sampled rather than observed but if its `msg["value"]` had generated a sample that was later used in a downstream likelihood? This may be rare, but I see no reason to hard-code to observed sites here.

neerajprad: I suppose I was tunnel-visioned into making this work for HMC, where both the Poisson and Binomial need to be observed as they are discrete variables. I think this should work other-wise too. Let me try to refactor, and update.

pyro/contrib/conjugate/infer.py
neerajprad: Hmm..the issue here is that `msg["fn"].parent._conditional.site_name` might not be set by the time we are at the latent node since that occurs further downstream. We also can't use `self.trace[msg["name"]].parent..` since `msg["name"]` won't be there in `self.trace` as it was collapsed out. 

I guess the only way to retrieve the corresponding pair would be to iterate through the trace and do this round-about check, but we don't need to restrict ourselves to observed nodes.

neerajprad:
@jpchen - Any idea why codecov is showing up again? I remember we had disabled it in the config.

jpchen:
>  I remember we had disabled it in the config.

yeah [it is](https://github.com/pyro-ppl/pyro/blob/dev/.codecov.yml#L10) and the interface hasnt changed. it doesnt seem to be appearing for the other PRs though?

pyro/contrib/conjugate/infer.py
fritzo: Thanks for explaining.

PR #1809, opened 2019-04-06
add first CJS ecological model

martinjankowiak:
will follow this up with additional PRs that add more model variants, more comments, and a bit of interpretation of the results of inference

- special thanks to @fritzo for reviewing the model
- special thanks to @OptimusLime for helping with R data conversation


fritzo:
@martinjankowiak or @stefanwebb feel free to merge as soon as tests pass.

tutorial/source/index.rst
```diff
@@ -56,6 +56,7 @@ Welcome to Pyro Examples and Tutorials!
    :maxdepth: 2
    :caption: Code Examples:
 
+   csj
```
fritzo: Where is `tutorial/source/cjs.rst`?

fritzo: It looks like you have completed step 2 before step 1. I'll merge now, but could you please add this file in a follow-up PR?

PR #1814, opened 2019-04-07
DeepELUFlow

stefanwebb:
This is a new normalizing flow similar to DeepSigmoidalFlow, but replacing the sigmoid function with ELU. It is suggested in the NAF paper but they didn't run any experiments. I have found it to be more flexible than DSF however


martinjankowiak:
@stefan this seems close enough to `DeepSigmoidalFlow` that these two could profitably be combined into one class?

stefanwebb:
That's a really good idea! I've refactored so only the nonlinearity is reimplemented between the types of NAF. I also added in a "leaky ReLU NAF", which is mentioned in the paper

PR #1820, opened 2019-04-12
Implement constraints in minipyro

### LINKED ISSUE #1819: [FR] Support constraints in minipyro
Currently constraints in Pyro are handled by a `ParamStoreDict` class. The `ParamStoreDict` interface was simplified in #1363, but the class remains complex to retain backwards compatibility. It would be nice to pull out the minimal logic needed to handle constraints.

We could do this in minipyro in two different ways:
1. Simplify the `ParamStoreDict` to be a simple dict whose values are `(unconstrained_value, constraint)` pairs, and move `transform_to(constraint)` logic into `pyro.param()`.
    Advantages: less code overall
2. Keep the logic in `ParamStoreDict` but simplify its interface.
    Advantages: easier for users to use the `PARAM_STORE` directly

fritzo:
Resolves #1819 

This implements constraints in minipyro, following @eb8680 's suggestion of simplifying the `PARAM_STORE`.

Note that this implementation is not backwards compatible wrt `get_param_store()`, since `get_param_store().items()` will return values that are `(unconstrained_value, constraint)` pairs. But since this version is much simpler, I'm happy to standardize on this narrower API for the param store.

## Tested

- [x] updated examples/minipyro.py
- [x] added a new tests/contrib/test_minipyro.py (forked from funsor)

### DIFF
--- a/examples/minipyro.py
+++ b/examples/minipyro.py
@@ -53,7 +53,8 @@ def main(args):
 
         # Report the final values of the variational parameters
         # in the guide after training.
-        for name, value in pyro.get_param_store().items():
+        for name in pyro.get_param_store():
+            value = pyro.param(name)
             print("{} = {}".format(name, value.detach().cpu().numpy()))
 
         # For this simple (conjugate) model we know the exact posterior. In
--- a/pyro/contrib/minipyro.py
+++ b/pyro/contrib/minipyro.py
@@ -14,6 +14,7 @@ found at examples/minipyro.py.
 from __future__ import absolute_import, division, print_function
 
 from collections import OrderedDict
+import weakref
 
 import torch
 
@@ -25,7 +26,7 @@ import torch
 #     See http://docs.pyro.ai/en/0.3.1/parameters.html
 
 PYRO_STACK = []
-PARAM_STORE = {}
+PARAM_STORE = {}  # maps name -> (unconstrained_value, constraint)
 
 
 def get_param_store():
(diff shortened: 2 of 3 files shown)


fritzo:
@neerajprad after this merges I will copy the changes to funsor.minipyro. At some point we should also port the changes to numpyro.minipyro.

fritzo:
> how important is it to maintain backwards compatibility

I agree with @eb8680 that in the case of *.minipyro it is more important to be able to write simple implementations than to support all previous features. This version is still compatible with the use `[pyro.param(name) for name in pyro.get_param_store()]`, and I think it's reasonable to limit the informal minipyro spec to this narrow usage.

eb8680:
> get_param_store() API

I suppose we could support this by constructing a new dictionary:
```
def get_param_store():
    return {name: pyro.param(name) for name in PYRO_PARAM_STORE.keys()}
```

neerajprad:
> At some point we should also port the changes to numpyro.minipyro.

I think we already have major transforms in numpyro with a similar interface as `torch` (which we can eventually move to `numpy-tfp` or `jax`). Since there isn't any param store though and JIT needs explicit transformations, we will have to apply the transformations as part of the `substitute` handler or within the elbo computation. I will think about it and make an appropriate change for svi in numpyro.

fehiepsi:
> Since there isn't any param store though and JIT needs explicit transformations, we will have to apply the transformations as part of the substitute handler or within the elbo computation. I will think about it and make an appropriate change for svi in numpyro.

Waiting for your ideas on this! Hopefully it can incorporate into hmc too.

PR #1821, opened 2019-04-14
Implement a SpanningTree distribution

fritzo:
Addresses #1824 

This implements a sampler, log_prob evaluator, and enumerator for random spanning trees.

The motivating application is sampling latent structures in models like [TreeCat](https://github.com/posterior/treecat). The sampler in this PR was forked from TreeCat's [numba-based sampler](https://github.com/posterior/treecat/blob/0.1.10/treecat/structure.py#L365).

I've implemented two versions of the sampler: a simple slow Python version and a C++ version which is compiled as a PyTorch extension module. I've tried to keep the Python and C++ sources synchronized as nearly line-by-line translations. To avoid installation issues for this seldom-used extension, I've used PyTorch's just-in-time extension module mechanism and included the `.cpp` file in the Pyro distribution via `MANIFEST.in` (rather than building the extension in setup.py).

## Review requests
- @neerajprad Could you please review the distribution interface and C++ extension module? 
- @martinjankowiak Could you please lightly review the stats? I've added some strong tests, so I'm fairly confident in correctness. I'm happy to explain the MCMC sampler offline.

## Tested
- [x] exact tests for `.log_prob()` and `.enumerate_support()` on small graphs.
- [x] smoke tests for both python and C++ samplers
- [x] goodness-of-fit tests for the samplers. These are **not** run on CI, and are only run locally if the [goftests](https://github.com/posterior/goftests) package is installed. This seems reasonable since these tests are expensive and we don't plan to change the algorithms frequently (there are smoke tests to ensure the code still runs).
- [x] uniform tests suggested by @martinjankowiak

### DIFF
--- a/pyro/distributions/spanning_tree.py
+++ b/pyro/distributions/spanning_tree.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import itertools
+import math
 
 import torch
 from torch.distributions import constraints
@@ -19,6 +20,8 @@ class SpanningTree(TorchDistribution):
     1. The edges constitute a tree, i.e. are connected and cycle free.
     2. Each edge ``(v1,v2) = edges[e]`` is sorted, i.e. ``v1 < v2``.
     3. The entire tensor is sorted in colexicographic order.
+
+    Use :func:`validate_edges` to verify `edges` are correctly formed.
 
     The ``edge_logits`` tensor has one entry for each of the ``V*(V-1)//2``
     edges in the complete graph on ``V`` vertices, where edges are each sorted
@@ -67,6 +70,40 @@ class SpanningTree(TorchDistribution):
         self.num_vertices = V
         self.sampler_options = {} if sampler_options is None else sampler_options
 
+    def validate_edges(self, edges):
+        """
+        Validates a batch of ``edges`` tensors, as returned by :meth:`sample` or
+        :meth:`enumerate_support` or as input to :meth:`log_prob()`.
+
+        :param torch.LongTensor edges: A batch of edges.
+        :raises: ValueError
+        :returns: None
+        """
+        if edges.shape[-2:] != self.event_shape:
+            raise ValueError("Invalid edges shape: {}".format(edges.shape))
+
+        # Verify canonical ordering.
+        if not ((0 <= edges) & (edges < self.num_vertices)).all():
+            raise ValueError("Invalid vertex ids:\n{}".format(edges))
+        if not (edges[..., 0] < edges[..., 1]).all():
+            raise ValueError("Vertices are not sorted in each edge:\n{}".format(edges))
+        if not ((edges[..., :-1, 1] < edges[..., 1:, 1]) |
+                ((edges[..., :-1, 1] == edges[..., 1:, 1]) &
+                 (edges[..., :-1, 0] < edges[..., 1:, 0]))).all():
+            raise ValueError("Edges are not sorted colexicographically:\n{}".format(edges))
+
+        # Verify tree property, i.e. connectivity.
+        V = self.num_vertices
+        for i in itertools.product(*map(range, edges.shape[:-2])):
+            edges_i = edges[i]
+            connected = torch.eye(V, dtype=torch.float)
+            connected[edges_i[:, 0], edges_i[:, 1]] = 1
+            connected[edges_i[:, 1], edges_i[:, 0]] = 1
+            for i in range(int(math.ceil(V ** 0.5))):
+                connected = connected.mm(connected).clamp_(max=1)
+            if not connected.min() > 0:
+                raise ValueError("Edges do not constitute a tree:\n{}".format(edges_i))
+
     @lazy_property
     def log_partition_function(self):
         # By Kirchoff's matrix-tree theorem, the partition function is the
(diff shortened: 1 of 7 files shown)


pyro/distributions/spanning_tree.py
```diff
@@ -0,0 +1,418 @@
...
+    has_enumerate_support = True
+
+    def __init__(self, edge_logits, sampler_options=None, validate_args=None):
+        if edge_logits.is_cuda:
+            raise NotImplementedError("SpanningTree does not support cuda tensors")
+        K = len(edge_logits)
+        V = int(round(0.5 + (0.25 + 2 * K)**0.5))
+        assert K == V * (V - 1) // 2
+        E = V - 1
+        event_shape = (E, 2)
```
martinjankowiak: should be made clearer what return type of sampler is?

pyro/distributions/spanning_tree.py
```diff
@@ -0,0 +1,418 @@
...
+    trees.sort()
+    return trees
+
+
+def enumerate_spanning_trees(V):
+    """
+    Compute the set of spanning trees on V vertices.
+    """
+    if V >= len(_TREE_GENERATORS):
+        raise NotImplementedError
```
martinjankowiak: error msg?

pyro/distributions/spanning_tree.py
```diff
@@ -0,0 +1,418 @@
...
+    v1, v2 = torch.broadcast_tensors(v1, v2)
+    v1 = v1.contiguous().view(-1)
+    v2 = v2.contiguous().view(-1)
+    mask = (v1 < v2)
+    grid = torch.stack((v1[mask], v2[mask]))
+    assert grid.shape == (2, K)
+    return grid
+
+
+def _remove_edge(grid, e2k, neighbors, components, e):
```
martinjankowiak: could you please add comments defining some of the data structures you use? especially `e2k`? maybe also `neighbors` and a few others?

pyro/distributions/spanning_tree.py
```diff
@@ -0,0 +1,418 @@
...
+    valid_edges_buffer = torch.empty(K, dtype=torch.long)
+
+    for e in range(E):
+        k = _remove_edge(grid, e2k, neighbors, components, e)
+        num_valid_edges = _find_valid_edges(components, valid_edges_buffer)
+        valid_edges = valid_edges_buffer[:num_valid_edges]
+        valid_logits = edge_logits[valid_edges]
+        valid_probs = (valid_logits - valid_logits.max()).exp()
+        total_prob = valid_probs.sum()
+        if total_prob > 0:
```
martinjankowiak: can this not be true?

tests/distributions/test_spanning_tree.py
```diff
@@ -0,0 +1,144 @@
...
+from collections import Counter
+
+import pytest
+import torch
+
+import pyro
+from pyro.distributions.spanning_tree import (NUM_SPANNING_TREES, SpanningTree, make_complete_graph, sample_tree_approx,
+                                              sample_tree_mcmc)
+from tests.common import assert_equal, xfail_if_not_implemented
+
```
martinjankowiak: does it make sense to add any tests that check that the two backends are giving the same results? (assuming randomness can be suitably fixed)

pyro/distributions/spanning_tree.py
```diff
@@ -0,0 +1,418 @@
...
+            raise NotImplementedError("SpanningTree does not support cuda tensors")
+        K = len(edge_logits)
+        V = int(round(0.5 + (0.25 + 2 * K)**0.5))
+        assert K == V * (V - 1) // 2
+        E = V - 1
+        event_shape = (E, 2)
+        batch_shape = ()
+        self.edge_logits = edge_logits
+        super(SpanningTree, self).__init__(batch_shape, event_shape, validate_args=validate_args)
+        if self._validate_args:
```
neerajprad: nit: consider moving this check earlier (after computing K).

fritzo: The reason I put it here is because `self._validate_args` is only defined after `super(...).__init__()` (recall that `self._validate_args` is true if either `validate_args is True` or `validate_args is None and Distribution._validate_args is True`).

pyro/distributions/spanning_tree.py
fritzo: Yes, this can happen when all `valid_logits` are `-inf`. If that's the case the mcmc kernel will avoid moving rather than erroring. I added this a couple years ago to avoid errors in real datasets.

tests/distributions/test_spanning_tree.py
fritzo: I'd rather test the correctness of each backend independently and avoid making assumptions about C++ vs Python rngs and data structure ordering (e.g. order of operations can be slightly different when iterating over a Python `set()` versus a C++ `std::unordered_set`).

pyro/distributions/spanning_tree.py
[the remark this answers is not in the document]
fritzo: I've added a description of in the docstring.

fritzo:
OK, I believe I have addressed all review comments so far.

pyro/distributions/spanning_tree.py
neerajprad: Ahh..thanks for clarifying!

fritzo:
Thanks @neerajprad and @martinjankowiak for reviewing this big PR!

PR #1822, opened 2019-04-16
torchify fit_generalized_pareto

martinjankowiak:


pyro/ops/stats.py
```diff
@@ -350,57 +349,46 @@ def fit_generalized_pareto(X):
...
     """
-    if isinstance(X, torch.Tensor):
-        X = X.data.cpu().float().numpy()
+    if not isinstance(X, torch.Tensor) or X.dim() != 1:
+        raise ValueError("Input X must be a 1-dimensional torch tensor")
 
-    if X.ndim != 1 or len(X) <= 1:
-        raise ValueError("Invalid input array.")
+    X = X.cpu().double()
+    X = torch.sort(X, descending=False)[0]
```
jpchen: is the conversion to cpu because `sort()` is only cpu implemented?

martinjankowiak: cuz i want doubles

jpchen: there are double tensors on cuda though?

PR #1826, opened 2019-04-17
add psis diagnostic

martinjankowiak:
follow-up to #1822 

adds a VI diagnostic along the lines of what's described in [Yes, but Did It Work?](https://arxiv.org/abs/1802.02538)

enabled by #1804

### DIFF
--- a/pyro/infer/importance.py
+++ b/pyro/infer/importance.py
@@ -1,9 +1,13 @@
 from __future__ import absolute_import, division, print_function
-import torch
+
+import math
 import warnings
+
+import torch
 
 import pyro
 import pyro.poutine as poutine
+from pyro.ops.stats import fit_generalized_pareto
 
 from .abstract_infer import TracePosterior
 from .enum import get_importance_trace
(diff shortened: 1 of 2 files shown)


tests/infer/test_util.py
```diff
@@ -44,3 +46,22 @@ def test_multi_frame_tensor():
     for name, expected_sum in expected.items():
         actual_sum = actual.sum_to(stacks[name])
         assert_equal(actual_sum, expected_sum, msg=name)
+
+
+@pytest.mark.parametrize('max_particles', [250 * 1000, 500 * 1000])
+@pytest.mark.parametrize('scale_krange', zip([0.50, 0.95], [(0.7, 0.9), (0.05, 0.2)]))
```
fritzo: nit: You can write this directly using pytest:
```py
@pytest.mark.parametrize('scale,krange', [(0.5, (0.7, 0.9)),
                                          (0.95, (0.05, 0.2))])
...
def test_psis_diagnostic(scale, krange, ...):
```

pyro/infer/importance.py
```diff
@@ -140,3 +144,72 @@ def _fn(*args, **kwargs):
     if normalized:
         log_weights = log_weights - torch.logsumexp(log_weights)
     return log_weights, model_trace, guide_trace
+
+
+def psis_diagnostic(model, guide, *args, **kwargs):
```
fritzo: Is this function meaningfully differentiable? If you are certain it is not, then I'd recommend decorating with `@torch.no_grad()`:
```py
@torch.no_grad()
def psis_diagnostic(...):
    ...
```

tests/infer/test_util.py
martinjankowiak: done

pyro/infer/importance.py
martinjankowiak: done

PR #1827, opened 2019-04-17
Fix infer_discrete handling of plate of size 1

fritzo:
This fixes `infer_discrete`'s handling of plates of size 1.

This issue arises because `pyro.ops.packed` ignores dims of size 1, but `infer_discrete` logic required (before this PR) all dims to be accounted for. I've also had to plumb `dim_to_size` through `_make_ring()` inside `infer_discrete`, which we must have forgotten originally (I followed the correct better-tested implementation in ubersum).

## Tested
- added a smoke test
- tested in the contrib-treecat branch as part of #1370

### DIFF
--- a/tests/infer/test_discrete.py
+++ b/tests/infer/test_discrete.py
@@ -35,7 +35,7 @@ def log_mean_prob(trace, particle_dim):
 
 
 @pytest.mark.parametrize('temperature', [0, 1], ids=['map', 'sample'])
-@pytest.mark.parametrize('plate_size', [1, 2])
+@pytest.mark.parametrize('plate_size', [2])
 def test_plate_smoke(temperature, plate_size):
     #       +-----------------+
     #  z1 --|--> z2 ---> x2   |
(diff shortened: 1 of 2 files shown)


tests/infer/test_discrete.py
```diff
@@ -34,6 +34,27 @@ def log_mean_prob(trace, particle_dim):
...
+    #       +-----------------+
+    #  z1 --|--> z2 ---> x2   |
+    #       |               N | for N in {1,2}
+    #       +-----------------+
+
+    @config_enumerate
+    def model():
+        p = pyro.param("p", torch.tensor([0.25, 0.75]))
+        loc = pyro.param("loc", torch.tensor([-1., 1.]))
+        z1 = pyro.sample("z1", dist.Categorical(p))
```
eb8680: `flake8` failed here

fritzo:
@eb8680 the change was a little more involved that I originally realized; should be fixed now.

PR #1834, opened 2019-04-22
Fix tests failing on CUDA

neerajprad:
This fixes all tests failing on CUDA in preparation for a minor release. It also updates the torch requirement to `1.0.1` which seems to be needed by the `LKJCorrCholesky` distribution.

### DIFF
--- a/pyro/distributions/lkj.py
+++ b/pyro/distributions/lkj.py
@@ -202,9 +202,11 @@ class LKJCorrCholesky(TorchDistribution):
         Km1 = self._d - 1
 
         log_diagonals = x.diagonal(offset=0, dim1=-1, dim2=-2)[..., 1:].log()
+        # TODO: Figure out why the `device` kwarg to torch.linspace seems to not work in certain situations,
+        # and a seemingly redundant .to(x.device) is needed below.
         values = log_diagonals * torch.linspace(start=Km1 - 1, end=0, steps=Km1,
                                                 dtype=x.dtype,
-                                                device=x.device).expand_as(log_diagonals)
+                                                device=x.device).expand_as(log_diagonals).to(x.device)
 
         values += log_diagonals.mul(eta.mul(2).add(-2.0))
         values = values.sum(-1) + lp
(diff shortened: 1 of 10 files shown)


tests/infer/test_gradient.py
```diff
@@ -70,10 +70,10 @@ def guide(subsample):
     inference = SVI(model, guide, optim, loss=elbo)
     with xfail_if_not_implemented():
         if subsample_size == 1:
-            inference.loss_and_grads(model, guide, subsample=torch.LongTensor([0]))
```
jpchen: is there a difference here? wont it default to the right device?

pyro/distributions/lkj.py
```diff
@@ -202,9 +202,11 @@ def log_prob(self, x):
         Km1 = self._d - 1
 
         log_diagonals = x.diagonal(offset=0, dim1=-1, dim2=-2)[..., 1:].log()
+        # TODO: Figure out why the `device` kwarg to torch.linspace seems to not work in certain situations,
+        # and a seemingly redundant .to(x.device) is needed below.
```
jpchen: shouldnt we just remove the `device=x.device` then?

tests/infer/test_gradient.py
neerajprad: Its really odd, but it doesn't:

```python
In [1]: import torch

In [2]: torch.set_default_tensor_type('torch.cuda.DoubleTensor')

In [3]: torch.LongTensor(1)
Out[3]: tensor([140263273888632], device='cpu')

```

pyro/distributions/lkj.py
neerajprad: The device arg to `torch.linspace` should work fine, and I'm only facing this issue when I run `test_cuda.py`. Having the `.to` should be redundant in most cases, but removing the `device` arg will be 2-3 times slower.

neerajprad:
Thanks for reviewing, @jpchen and @fritzo.

PR #1837, opened 2019-04-28
Fix issue with categorical and readd `keep_sites` and `keep_plates` for TracePredictive

ahmadsalim:
Hi Pyro Devs!

It seems that I apparently did need to allow an argument to allow choosing which sites should be sampled from the posterior predictive and which sites should be re-sampled in `TracePredictive`.

In particular, this is useful for 'local' parameters that allow prediction for specific data input, where I want to resample for the new data input instead of reuse predictions from the inferred posterior.

### DIFF
--- a/pyro/infer/abstract_infer.py
+++ b/pyro/infer/abstract_infer.py
@@ -281,16 +281,18 @@ class TracePredictive(TracePosterior):
     achieved by constraining latent sites to randomly sampled parameter
     values from the model execution traces and running the model forward
     to generate traces with new response ("_RETURN") sites.
-
     :param model: arbitrary Python callable containing Pyro primitives.
-    :param TracePosterior posterior: trace posterior instance holding
-        samples from the model's approximate posterior.
+    :param TracePosterior posterior: trace posterior instance holding samples from the model's approximate posterior.
     :param int num_samples: number of samples to generate.
+    :param keep_sites: The sites which should be sampled from posterior predictive distribution (default: all)
+    :param keep_plates: The plates which should be sampled from posterior predictive distribution (default: all)
     """
-    def __init__(self, model, posterior, num_samples):
+    def __init__(self, model, posterior, num_samples, keep_sites=None, keep_plates=None):
         self.model = model
         self.posterior = posterior
         self.num_samples = num_samples
+        self.keep_sites = keep_sites
+        self.keep_plates = keep_plates
         super(TracePredictive, self).__init__()
 
     def _traces(self, *args, **kwargs):
@@ -299,9 +301,23 @@ class TracePredictive(TracePosterior):
         data_trace = poutine.trace(self.model).get_trace(*args, **kwargs)
         for _ in range(self.num_samples):
             model_trace = self.posterior().copy()
+            self._remove_dropped_nodes(model_trace)
             self._adjust_to_data(model_trace, data_trace)
             resampled_trace = poutine.trace(poutine.replay(self.model, model_trace)).get_trace(*args, **kwargs)
             yield (resampled_trace, 0., 0)
+
+    def _remove_dropped_nodes(self, trace):
+        if self.keep_sites is None and self.keep_plates is None:
+            return
+        for name, site in list(trace.nodes.items()):
+            if name not in self.keep_sites and name not in self.keep_plates:
+                trace.remove_node(name)
+                continue
+            try:
+                if any(cis.name in self.keep_plates for cis in site["cond_indep_stack"]):
+                    trace.remove_node(name)
+            except KeyError:
+                pass
 
     def _adjust_to_data(self, trace, data_trace):
         for name, site in list(trace.nodes.items()):
@@ -317,9 +333,12 @@ class TracePredictive(TracePosterior):
                     # Select random sub-indices to replay values under conditionally independent stacks.
                     # Otherwise, we assume there is an dependence of indexes between training data
                     # and prediction data.
-                    logits = torch.ones(site["value"].size(cis.dim), device=site["value"].device)
-                    subidxs = Categorical(logits=logits).sample([cis.size])
-                    site["value"] = site["value"].index_select(cis.dim, subidxs)
+                    bdim = cis.dim - site["fn"].event_dim
+                    cat = Categorical(logits=torch.ones(site["value"].size(bdim),
+                                                        device=site["value"].device,
+                                                        dtype=torch.float))
+                    subidxs = cat.sample((cis.size,))
+                    site["value"] = site["value"].index_select(bdim, subidxs)
             except KeyError:
                 pass
 
--- a/pyro/infer/abstract_infer.py
+++ b/pyro/infer/abstract_infer.py
@@ -284,15 +284,13 @@ class TracePredictive(TracePosterior):
     :param model: arbitrary Python callable containing Pyro primitives.
     :param TracePosterior posterior: trace posterior instance holding samples from the model's approximate posterior.
     :param int num_samples: number of samples to generate.
-    :param keep_sites: The sites which should be sampled from posterior predictive distribution (default: all)
-    :param keep_plates: The plates which should be sampled from posterior predictive distribution (default: all)
+    :param keep_sites: The sites which should be sampled from posterior distribution (default: all)
     """
-    def __init__(self, model, posterior, num_samples, keep_sites=None, keep_plates=None):
+    def __init__(self, model, posterior, num_samples, keep_sites=None):
         self.model = model
         self.posterior = posterior
         self.num_samples = num_samples
         self.keep_sites = keep_sites
-        self.keep_plates = keep_plates
         super(TracePredictive, self).__init__()
 
     def _traces(self, *args, **kwargs):
@@ -307,17 +305,12 @@ class TracePredictive(TracePosterior):
             yield (resampled_trace, 0., 0)
 
     def _remove_dropped_nodes(self, trace):
-        if self.keep_sites is None and self.keep_plates is None:
+        if self.keep_sites is None:
             return
         for name, site in list(trace.nodes.items()):
-            if name not in self.keep_sites and name not in self.keep_plates:
+            if name not in self.keep_sites:
                 trace.remove_node(name)
                 continue
-            try:
-                if any(cis.name in self.keep_plates for cis in site["cond_indep_stack"]):
-                    trace.remove_node(name)
-            except KeyError:
-                pass
 
     def _adjust_to_data(self, trace, data_trace):
         for name, site in list(trace.nodes.items()):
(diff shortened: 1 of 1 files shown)


pyro/infer/abstract_infer.py
```diff
@@ -281,16 +281,18 @@ class TracePredictive(TracePosterior):
...
     values from the model execution traces and running the model forward
     to generate traces with new response ("_RETURN") sites.
-
     :param model: arbitrary Python callable containing Pyro primitives.
-    :param TracePosterior posterior: trace posterior instance holding
-        samples from the model's approximate posterior.
+    :param TracePosterior posterior: trace posterior instance holding samples from the model's approximate posterior.
     :param int num_samples: number of samples to generate.
+    :param keep_sites: The sites which should be sampled from posterior predictive distribution (default: all)
+    :param keep_plates: The plates which should be sampled from posterior predictive distribution (default: all)
```
neerajprad: @ahmadsalim - Thanks for the PR! It will be great if you could walk me through an example (maybe a toy model), and help me understand how this change helps the particular use case you have. While this looks mostly okay, I think I might be missing the larger picture of how you are using the `TracePredictive` class. We can sync offline too, if that helps.

ahmadsalim: @neerajprad OK, thanks for the response 😄 . I will see if I can make a good minimal example.

ahmadsalim:
@neerajprad I have tried making a small example below. Essentially I am training using `SVI` and then getting a trace over parameters. I then want to use the trace to do new predictions on test data. 

Since the current predictive distribution just resamples all parameters, I cannot get new useful predictions for `ys` using it. With the new version I can keep only predictions about the weights, without predictions about `ys` from training data, and use that to do predictions on `ys` for test data.

I hope this sounds reasonable 😄 , but I could be doing something wrong here.

```python
def model(xs, ys=None):
    inp = torch.stack([xs, torch.ones_like(xs)], dim=-1)
    ws = pyro.sample('ws', dist.Normal(0, 1).expand_by([2]).to_event())
    with pyro.plate('xs', xs.size(0), dim=-1):
        return pyro.sample('ys', dist.Bernoulli(logits=inp @ ws), obs=ys)


def main(_args):
    optim = Adam(dict(lr=1e-3))
    loss = Trace_ELBO()
    guide = AutoMultivariateNormal(model)
    svi = SVI(model, guide, optim, loss, num_samples=100)
    train_xs = torch.tensor([-2., -5., -1., 4., 3., 10., 11., 15., 100.])
    train_ys = torch.tensor([0., 0., 1., 1., 1., 1., 1., 1., 1.])

    test_xs = torch.tensor([-3.])
    test_ys = torch.tensor([0.])

    for i in range(5000):
        loss = svi.step(train_xs, ys=train_ys)
        if i % 100 == 0:
            print(f"Epoch [{i}]: {loss}")

    # Get samples from inferred posterior
    svi.run(train_xs)

    # Trace predictive over all variables
    predictive_cl = TracePredictive(model, svi, num_samples=100)

    # Trace predictive over only weights
    predictive_ws = TracePredictive(model, svi, num_samples=100, keep_sites=['ws'], keep_plates=[])

    # Get predictions for test data
    ys_p_c = predictive_cl.run(test_xs).marginal(sites=['ys']).empirical['ys'].mean
    ys_p_w = predictive_ws.run(test_xs).marginal(sites=['ys']).empirical['ys'].mean

    print(ys_p_c > 0.5)  # Expected: False , Actual: True -- This is because predictions from training are reused
    print(ys_p_w > 0.5)  # Expected: False , Actual: False -- Predictions are recalc'ed using only weights from training
```

neerajprad:
@ahmadsalim - Thanks a lot for this snippet, I think this helps me understand the issue much better. You will find that your example will work as expected if you remove this line:

```
    # Get samples from inferred posterior
    svi.run(train_xs)
```

As to why that is the case - that line is getting traces from the posterior predictive since `train_ys` isn't passed in. As a side effect, it is also marking `ys` as stochastic sites since the site. As such, any further resampling ends up sampling from this predictive distribution because TracePredictive mistakes this for a stochastic site. Note that observed sites will not be resampled by default from `TracePosterior`, i.e. if you do `trace_post()` you will get a trace that does not include any observed sites, and when you replay against this, you will get the desired behavior of getting useful predictions from the test data because the `train_ys` won't be resampled. 

I don't see an easy way to prevent errors like this except to discourage users from calling `svi.run` by themselves. Let me know if there is more to your use case that the above example does not capture. I think you also have another fix in this PR for the dtype issue that we should definitely merge.

ahmadsalim:
@neerajprad Thank you very much for the response! I didn't know that `TracePosterior` tried to handle this automatically 😄!

I have a few comments:

- It seems that if I don't call `run` before, the `_traces()` method tries to call run itself. If I only pass in test data without observations, and then call it with new test data, would I not experience the same problem? I would appreciate an example on how `TracePredictive`/`TracePosterior` is meant to be used for this. 😃 

- I also have some models where I add local variables, e.g. to impute missing data. I try to model them in a hierarchical fashion, so they have a common global parameter, but I would like to resample the local parameters for each new dataset. My plan is to add a data-dependent imputation mechanism in the future (to model MAR) and it would be nice to resample from the guide given the data, instead of reusing imputations from the training data.

- I know this sounds silly, but I spent multiple days debugging why I am getting such good training results, while my predictions where horrible when using `TracePredictive`. When I added `keep_sites` I could immediately see that the problem was because of random reuse of the output. Maybe having the option could aid debugging in the future?

neerajprad:
> It seems that if I don't call run before, the _traces() method tries to call run itself. If I only pass in test data without observations, and then call it with new test data, would I not experience the same problem? 

I see, you are right about that. I think we need to run `svi.run(train_xs, train_ys)` first before calling `TracePredictive`, exactly like for HMC. I think that should handle your use case? 

> I also have some models where I add local variables, e.g. to impute missing data. I try to model them in a hierarchical fashion, so they have a common global parameter, but I would like to resample the local parameters for each new dataset.

I think this sounds reasonable, and a good use case for the `keep_sites` argument. Let me take another look at your PR. :)

pyro/infer/abstract_infer.py
neerajprad: Since we are going to adjusting the plates based on the test data, what is the use case for `keep_plates`?

ahmadsalim:
Thanks for the comments and understanding!
I think you are right, that for my use case I should run with the labels as well! I will do that in my code.

pyro/infer/abstract_infer.py
ahmadsalim: It allows one to drop all sites under plates which are not kept, which is convenient if one has many local sites. Perhaps, I should drop it though, if you find it unimportant 😄

pyro/infer/abstract_infer.py
```diff
@@ -281,16 +281,18 @@ class TracePredictive(TracePosterior):
     achieved by constraining latent sites to randomly sampled parameter
     values from the model execution traces and running the model forward
     to generate traces with new response ("_RETURN") sites.
-
     :param model: arbitrary Python callable containing Pyro primitives.
-    :param TracePosterior posterior: trace posterior instance holding
-        samples from the model's approximate posterior.
+    :param TracePosterior posterior: trace posterior instance holding samples from the model's approximate posterior.
     :param int num_samples: number of samples to generate.
+    :param keep_sites: The sites which should be sampled from posterior predictive distribution (default: all)
```
neerajprad: nit: Let us change these to `sample from the posterior distribution`, since `predictive` is confusing in this context.

pyro/infer/abstract_infer.py
```diff
@@ -317,8 +333,12 @@ def _adjust_to_data(self, trace, data_trace):
                     # Select random sub-indices to replay values under conditionally independent stacks.
                     # Otherwise, we assume there is an dependence of indexes between training data
                     # and prediction data.
-                    subidxs = Categorical(logits=site["value"].new_ones(site["value"].size(cis.dim))).sample([cis.size])
-                    site["value"] = site["value"].index_select(cis.dim, subidxs)
+                    bdim = cis.dim - site["fn"].event_dim
```
neerajprad: nit: Can we use more explicit names like `batch_dim` or `cond_indep_dim`? That will make it easier to grok what is happening here. 

Also, I am not sure why we need to adjust by the `event_dim`. I think subsampling from `cis.dim` in the sample site should be enough? Let us add a test case, if that's not the case.

pyro/infer/abstract_infer.py
neerajprad: I see. I think this argument can seem confusing, so it is best to avoid it. When the API is more mature, and if it turns out to be a pain point, we can consider reinstating it back.

pyro/infer/abstract_infer.py
```diff
@@ -317,8 +333,12 @@ def _adjust_to_data(self, trace, data_trace):
...
+                    cat = Categorical(logits=torch.ones(site["value"].size(bdim),
+                                                        device=site["value"].device,
+                                                        dtype=torch.float))
+                    subidxs = cat.sample((cis.size,))
+                    site["value"] = site["value"].index_select(bdim, subidxs)
```
neerajprad: I think we discussed this the last time, and I am forgetting now, but can we not simply use:

```python
subidxs = torch.randint(0, site['value'].size(bdim), cis.size)
```

ahmadsalim:
Thanks for the comments!
I have fixed everything apart from the test, which I will add soon 😄.

pyro/infer/abstract_infer.py
ahmadsalim: I explicitly named the variable now.

The reason that I adjust by event_dim is for distributions like `MultiVariateNormal` or `OneHotCategorical` which have more complex event shapes than scalar outputs from other distributions. I will think of a test to add for this.

pyro/infer/abstract_infer.py
ahmadsalim: Thanks! I think it seems to work. I probably just thought in terms of distributions when I made the first PR 😄

pyro/infer/abstract_infer.py
neerajprad: That makes sense. I think for the test we can just change our existing model/guide from a uniform-binomial to a dirichlet-categorical instead which will exercise this code path. We can handle this in a separate PR, I'll be happy to help with this.

neerajprad:
@ahmadsalim - You'll need to merge the latest master. We can merge this right away, and modify our existing test to a more general model/guide with event dims in a separate PR.

ahmadsalim:
Sounds great! I will make a new PR with a test when I can.

Sorry for the wrong push before, I accidentally messed up the rebase 😄.
I have fixed it now, and also found that I forgot to readd the device fix when I changed to `randint`.

PR #1840, opened 2019-04-30
Fix unit test failures with torch==1.1.0

neerajprad:


pyro/optim/lr_scheduler.py
```diff
@@ -27,17 +27,11 @@ def __init__(self, scheduler_constructor, optim_args):
         # kwargs for the torch optimizer
         optim_kwargs = optim_args.pop('optim_args')
         self.kwargs = optim_args
-        # current epoch
-        self.epoch = None
```
neerajprad: @jpchen - Thanks for fixing the test. Do we not need to hold the current epoch in `self.epoch` anymore? What was the change on the PyTorch side?

jpchen: the way `get_lr()` was implemented for these schedulers. also the initialization of `last_epoch` attribute though that doesnt affect this.  every time `svi.step()` is the scheduler will take a step now.

neerajprad:
Currently the size of the CPU wheel seems to have increased by 10X so travis gets stuck at times.

jpchen:
> Currently the size of the CPU wheel seems to have increased by 10X so travis gets stuck at times.

yeah this is also why i couldnt build funsor docs. i spent half a day trying to work around this to no avail, so i would advise against updating the rtd config until this is resolved.

neerajprad:
> yeah this is also why i couldnt build funsor docs. i spent half a day trying to work around this to no avail, so i would advise against updating the rtd config until this is resolved.

Soumith just pointed out that the command to get the cpu wheels was directing to `python.filehosted` due to some pip quirk instead of actually grabbing the CPU wheels from pytorch's server. Look at `.travis.yml` and you can use the same command to get the pared down CPU wheels ~90 MB.

jpchen:
>do we need to update any rtd configs before merging this?

we should leave it for now until torch 1.1 builds. i will open a PR if/when it builds successfully

neerajprad:
I just noticed that the `baseball` example was throwing a leaked semaphore warning and needed https://github.com/tqdm/tqdm/pull/617 from tqdm; so I just updated the dependency version. This should be good to merge after this, and we can cut a minor release. Sorry for piling on another change in this PR.

neerajprad:
@fehiepsi - I had made some minor fixes to `gp` module to get the tests working with pytorch-1.1.0. Let me know if you have any further fixes that you would like to get into the next release.

fehiepsi:
@neerajprad Thanks! In previous version, JIT does not work with GP. I'll take a look now to see if it can be fixed in the mean time.

PR #1842, opened 2019-04-30
Avoid using legacy constructors Tensor.new_*()

fritzo:
The following constructors are deprecated and are [unsupported by the jit](https://github.com/pytorch/pytorch/pull/16770):
- `Tensor.new_tensor()`
- `Tensor.new_empty()`
- `Tensor.new_zeros()`
- `Tensor.new_ones()`
- `Tensor.new_full()`

## Updated
- [x] pyro core
- [x] most of pyro.contrib
- [x] examples/
- [x] tests/infer/test_jit.py

## Triaged (not updated)
- pyro.contrib.gp (I'll let @fehiepsi handle pyro.contrib.gp)
- tutorials
- tests

## Tested
- ran `CUDA_TEST=1 pytest -vx tests/test_examples.py::test_cuda` against torch==1.0.1

### DIFF
--- a/pyro/distributions/delta.py
+++ b/pyro/distributions/delta.py
@@ -34,7 +34,7 @@ class Delta(TorchDistribution):
         batch_shape = v.shape[:batch_dim]
         event_shape = v.shape[batch_dim:]
         if isinstance(log_density, numbers.Number):
-            log_density = v.new_empty(batch_shape).fill_(log_density)
+            log_density = torch.full(batch_shape, log_density, dtype=v.dtype, device=v.device)
         elif validate_args and log_density.shape != batch_shape:
             raise ValueError('Expected log_density.shape = {}, actual {}'.format(
                 log_density.shape, batch_shape))
--- a/examples/air/air.py
+++ b/examples/air/air.py
@@ -63,7 +63,8 @@ class AIR(nn.Module):
         self.baseline_scalar = baseline_scalar
         self.likelihood_sd = likelihood_sd
         self.use_cuda = use_cuda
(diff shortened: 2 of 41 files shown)


pyro/distributions/delta.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, v, log_density=0.0, event_dim=0, validate_args=None):
         batch_shape = v.shape[:batch_dim]
         event_shape = v.shape[batch_dim:]
         if isinstance(log_density, numbers.Number):
-            log_density = v.new_empty(batch_shape).fill_(log_density)
+            log_density = torch.full(batch_shape, log_density, dtype=v.dtype, device=v.device)
```
neerajprad: Not relevant to the cases that I see here, but I just discovered that `torch.full` doesn't generalize well in JIT when the fill value is scalar.

e.g.

```python
>>> a = torch.tensor([0, 0])

>>> def fn1(x):
        return torch.full(a.size(), x, dtype=a.dtype, device=a.device)

>>> torch.jit.trace(fn1, torch.tensor(1.))(torch.tensor(2.))
tensor([1, 1])

>>> def fn2(x):
        return torch.empty_like(a).fill_(x)

>>> torch.jit.trace(fn1, torch.tensor(1.))(torch.tensor(2.))
tensor([2, 2])
```

neerajprad:
Will merge when tests pass.

pyro/distributions/delta.py
fritzo: Interesting, it would be good to have this example in tests/infer/test_jit.py

fehiepsi:
>  (I'll let @fehiepsi handle pyro.contrib.gp)

Yes yes, I'll do it. :D

PR #1843, opened 2019-05-01
add inverse gamma distribution

martinjankowiak:
since there is no such thing in pytorch.dist.

@neerajprad do i need to do anything else to make this a fully functional pyro distribution?

### DIFF
--- a/pyro/distributions/inverse_gamma.py
+++ b/pyro/distributions/inverse_gamma.py
@@ -1,9 +1,8 @@
 from __future__ import absolute_import, division, print_function
 
 from torch.distributions import constraints
-from torch.distributions.gamma import Gamma
 from torch.distributions.transforms import PowerTransform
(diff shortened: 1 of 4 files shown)


pyro/distributions/inverse_gamma.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import absolute_import, division, print_function
+
+from torch.distributions import constraints
+from torch.distributions.gamma import Gamma
```
fritzo:
```py
from pyro.distributions.torch import Gamma, TransformedDistribution
```

neerajprad:
> @neerajprad do i need to do anything else to make this a fully functional pyro distribution?

I think it should work great.

martinjankowiak:
@neerajprad @fritzo should be good to merge

PR #1844, opened 2019-05-01
avoid using legacy constructors Tensor.new in contrib.gp

fehiepsi:
Follow up of #1842. This is ready to review if tests pass.

### DIFF
--- a/pyro/contrib/gp/kernels/brownian.py
+++ b/pyro/contrib/gp/kernels/brownian.py
@@ -45,4 +45,4 @@ class Brownian(Kernel):
         Zt = Z.t()
         return torch.where(X.sign() == Zt.sign(),
                            self.variance * torch.min(X.abs(), Zt.abs()),
-                           X.data.new_zeros(X.size(0), Z.size(0)))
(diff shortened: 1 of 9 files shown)


fehiepsi:
It seems that Travis is unstable...

fritzo:
@fehiepsi it appears there was some confusion on the PyTorch side and these are no longer considered legacy constructors. If pyro.contrib.gp does not currently support the jit, then there is no immediate reason to avoid these constructors.

I think it's safe to close this PR.

fehiepsi:
Actually, I just run the [benchmark test](https://github.com/pyro-ppl/pyro/blob/dev/tests/perf/test_benchmark.py#L102) with jit loss_fn and get the error `RuntimeError: Cannot insert a Tensor that requires grad as a constant. Consider making it a parameter or input, or detaching the gradient
Tensor:` (might be related to [this issue](https://github.com/pytorch/pytorch/issues/18569)). I am not sure what I should do to resolve it right now so jitted contrib.gp might not be supported in the near future. cc @neerajprad

neerajprad:
@fehiepsi - Please don't worry about it. This is just a minor release to make Pyro compatible with the latest PyTorch release.

Issue #1787, opened 2019-03-22
OOM error with NUTS on CUDA 


neerajprad:
### Issue Description

Issue mentioned in our [forum](https://forum.pyro.ai/t/reducing-memory-requirement-for-nuts/850):

When performing parameter inference with NUTS on CUDA I observed a linear increase in allocated memory that leads to the termination of the processes within a few iterations. This only happens after warmup is finished, and apparently also happens on CPU. The minimum program to reproduce this behavior is:

```
import torch
import pyro
from pyro.infer.mcmc import MCMC, NUTS
import pyro.distributions as dist

data = torch.ones((1000, 1000))
def model():
    x = pyro.sample('x', dist.Normal(0., 1.))
    pyro.sample('data', dist.Normal(x*data, 0.01), obs = data)
nuts_kernel = NUTS(model)
posterior = MCMC(nuts_kernel, 1000, warmup_steps = 0).run()
```

I suspect that the data is somehow copied and stored at each iteration. However, `del posterior` does not free any significant amount of memory (and it appear like only a reference and not copies to the original data object are stored). How can I avoid this excessive memory consumption?

### Environment

(torch 1.01, pyro 0.3.1, python 3.6.7)

neerajprad:
Thanks for the clear report! The memory would increase per iteration since we are storing `num_samples` execution traces, which could be more problematic on the GPU because of higher memory limitation. I think the only way around this would be to store the bare minimal information from the trace as mentioned in this [issue](https://github.com/pyro-ppl/pyro/issues/1725) (yet again!), so I'll definitely make sure that we address this as part of the refactoring.

cweniger:
I used poutine.block to remove the data nodes in `_get_trace` in the HMC and NUTS sampler. This solved the problem for now. See https://github.com/cweniger/pyro for details if you are interested.

neerajprad:
> I used poutine.block to remove the data nodes in _get_trace in the HMC and NUTS sampler. 

That's great! You can certainly use block to not record certain sites in your model. Were you cloning the data in your model? Because otherwise the data nodes should point to the same memory reference and shouldn't consume memory that scales with the number of samples (unless we have some bug in our code!).

cweniger:
I did not clone the data, and indeed the data nodes were all pointing to the same memory reference (inplace modifications modified all at once). Still, blocking these nodes solved the problem. I'm also confused by what is going on, but for now it seems to work.

fehiepsi:
[...]

Issue #1816, opened 2019-04-09
Improvements to the HMC interface


neerajprad:
For the next major release, we can incorporate some of the lessons from numpyro and issues from our forum, to improve the HMC interface as follows:
 - [ ] Provide a `potential_fn` like in numpyro so that users do not necessarily need to pass in their model, but can generate samples from any callable whose `log_density` can be evaluated. This will be useful for integrating with `funsors`, and eventually getting rid of our `TraceTreeEvaluator` wrappers. https://github.com/pyro-ppl/funsor/issues/123
 - [ ] If an initial trace is provided, we should make sure that we do not ever run the model to set up our initial state. This is important because many a time we want to sample from distributions which do not have a `sample` method defined. We should still be able to run HMC on such models.
 - [ ] Related to the points above, it will be nice if (just like in numpyro) once given a `potential_fn` and `initial_sample` we can generate subsequent samples, without having to make any assumptions about the container data structure. Right now, we assume that this container data structure is a Pyro trace object, but this is only needed to interface with the `TracePosterior` class. This will make it really simple to integrate NUTS/HMC into other libraries. I think a middle ground might be to move the trace wrapping/unwrapping logic into the MCMC class, but more discussion is needed.

Related issue: Refactoring the `TracePosterior` interface. #1725

neerajprad:
cc. @fritzo, @eb8680, @fehiepsi

fehiepsi:
I think that following numpyro approach is a good idea. It is also helpful for the future when things like `lax.cond, lax.while_loop` is available in pytorch (disclaim: I don't know what is the state of pytorch jit right now) so we can jit the whole trajectory as in numpyro to improve the speed.

fehiepsi:
@neerajprad Can I work on this issue too? It is required to support GPyTorch when only `potential_fn` and `params` are needed.

neerajprad:
Absolutely! Please feel free to create / mark off any sub-issue that you are working on above. I was planning to look at changes to the TracePosterior interface, so please feel free to take up the `potential_fn` refactor, which we also need to interface with funsors.

fehiepsi:
I guess we can close this issue now. Further work should go under TracePosterior refactoring.

Issue #1730, opened 2019-02-01
MCMC does not work well with multi-chains in CPU due to Memory Error


fehiepsi:
The following script is copied exactly from Bayesian regression tutorial (except `num_samples=3000` and `num_chains=2`). While running it, I get MemoryError. I have tried to debug this memory error by changing various settings (ulimit, shared memory segment) in my system but no hope. I get this issue in two systems which I have so I guess this will not happen for only me. If so, then this is a serious problem which can come from how we use torch.multiprocessing in Pyro.

```
import numpy as np
import pandas as pd
import torch

import pyro
import pyro.distributions as dist
from pyro.infer.mcmc import MCMC, NUTS

pyro.enable_validation(True)
pyro.set_rng_seed(1)
DATA_URL = "https://d2fefpcigoriu7.cloudfront.net/datasets/rugged_data.csv"
rugged_data = pd.read_csv(DATA_URL, encoding="ISO-8859-1")

#torch.multiprocessing.set_sharing_strategy("file_system")


def model(is_cont_africa, ruggedness, log_gdp):
    a = pyro.sample("a", dist.Normal(8., 1000.))
    b_a = pyro.sample("bA", dist.Normal(0., 1.))
    b_r = pyro.sample("bR", dist.Normal(0., 1.))
    b_ar = pyro.sample("bAR", dist.Normal(0., 1.))
    sigma = pyro.sample("sigma", dist.Uniform(0., 10.))
    mean = a + b_a * is_cont_africa + b_r * ruggedness + b_ar * is_cont_africa * ruggedness
    with pyro.iarange("data", len(ruggedness)):
        pyro.sample("obs", dist.Normal(mean, sigma), obs=log_gdp)

        
df = rugged_data[["cont_africa", "rugged", "rgdppc_2000"]]
df = df[np.isfinite(df.rgdppc_2000)]
df["rgdppc_2000"] = np.log(df["rgdppc_2000"])
train = torch.tensor(df.values, dtype=torch.float)
is_cont_africa, ruggedness, log_gdp = train[:, 0], train[:, 1], train[:, 2]

nuts_kernel = NUTS(model, adapt_step_size=True)
hmc_posterior = MCMC(nuts_kernel, num_samples=4000, warmup_steps=1000, num_chains=4).run(is_cont_africa, ruggedness, log_gdp)
```

Here is its [backtrack](https://gist.github.com/fehiepsi/03a28b713e90aa4b13cd644b68b27782)

[...]

PR #1845, opened 2019-05-02
Revise HMC to support general potential_fn

fehiepsi:
As part of #1816, this PR tries to revise HMC class to support general `potential_fn`. This feature is requested by some users in Pyro forum (e.g. to run HMC with custom or unnormalized log_prob) and GPyTorch folks.

The plan of this PR is to have a working version first, then we can discuss about what'd be needed to do with the interface later.

TODO:
- [x] Implements `initialize_model` function to create `intial_params`, `potential_fn`, and `transform_fn` (to generate `trace` of the current params)
- [x] Change the input of HMC/NUTS to `potential_fn` instead of `model`. And `initial_trace` is replaced by `initial_params`.
- [x] Think about a compatible layer for the current API. (need discussions about this when the above tasks are done).
- [x] Update tests if necessary (or maybe not)

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -12,14 +12,14 @@ import pyro.poutine as poutine
 from pyro.distributions.util import eye_like, scalar_like
 from pyro.infer import config_enumerate
 from pyro.infer.mcmc.adaptation import WarmupAdapter
-from pyro.infer.mcmc.trace_kernel import TraceKernel
+from pyro.infer.mcmc.mcmc_kernel import MCMCKernel
 from pyro.infer.mcmc.util import TraceEinsumEvaluator
 from pyro.ops.integrator import velocity_verlet
 from pyro.poutine.subsample_messenger import _Subsample
 from pyro.util import ignore_jit_warnings, optional, torch_isinf, torch_isnan
 
 
-class HMC(TraceKernel):
+class HMC(MCMCKernel):
     r"""
     Simple Hamiltonian Monte Carlo kernel, where ``step_size`` and ``num_steps``
     need to be explicitly specified by the user.
@@ -118,8 +118,6 @@ class HMC(TraceKernel):
         # In NUTS paper, this threshold is set to a fixed log(0.5).
         # After https://github.com/stan-dev/stan/pull/356, it is set to a fixed log(0.8).
         self._direction_threshold = math.log(0.8)  # from Stan
-        # number of tries to get a valid initial trace
-        self._max_tries_initial_trace = 100
         self._reset()
         self._adapter = WarmupAdapter(step_size,
                                       adapt_step_size=adapt_step_size,
@@ -170,12 +168,9 @@ class HMC(TraceKernel):
         self._r_shapes = {}
         self._r_numels = {}
         self._args = None
-        self._compiled_potential_fn = None
         self._kwargs = None
         self._prototype_trace = None
-        self._initial_trace = None
-        self._has_enumerable_sites = False
-        self._trace_prob_evaluator = None
+        self._initial_params = None
         self._z_last = None
         self._potential_energy_last = None
         self._z_grads_last = None
@@ -245,14 +240,12 @@ class HMC(TraceKernel):
         return max(1, int(self.trajectory_length / self.step_size))
 
     @property
-    def initial_trace(self):
-        return self._initial_trace
+    def initial_params(self):
+        return self._initial_params
 
     @initial_trace.setter
-    def initial_trace(self, trace):
-        self._initial_trace = trace
-        init_samples = {name: trace[name]["value"] for name in self._init_params}
-        self._init_params = self._transform_fn(init_samples, invert=False)
+    def initial_params(self, params):
+        self._init_params = params
 
     def _initialize_model_properties(self, model_args, model_kwargs):
         init_params, potential_fn, transform_fn, inital_trace = initialize_model(
@@ -270,6 +263,41 @@ class HMC(TraceKernel):
         self.transform_fn = transform_fn
         self._initial_trace = initial_trace
         # make mass_matrix based on init_params
+
+    @initial_params.setter
+    def initial_params(self, params):
+        self._initial_params = params
+        if self._warmup_steps is not None:  # if setup is already called
+            self._initialize_step_size()
+
+    def _initialize_sampler(self):
+        if self.max_plate_nesting is None:
+            self._guess_max_plate_nesting()
+        # Wrap model in `poutine.enum` to enumerate over discrete latent sites.
+        # No-op if model does not have any discrete latents.
+        self.model = poutine.enum(config_enumerate(self.model),
+                                  first_available_dim=-1 - self.max_plate_nesting)
+        if self._automatic_transform_enabled:
+            self.transforms = {}
+        trace = poutine.trace(self.model).get_trace(*self._args, **self._kwargs)
+        self._prototype_trace = trace
+        site_value = None
+        for name, node in trace.iter_stochastic_nodes():
+            if isinstance(node["fn"], _Subsample):
+                continue
+            if node["fn"].has_enumerate_support:
+                self._has_enumerable_sites = True
+                continue
+            site_value = node["value"]
+            if node["fn"].support is not constraints.real and self._automatic_transform_enabled:
+                self.transforms[name] = biject_to(node["fn"].support).inv
+                site_value = self.transforms[name](node["value"])
+            self._r_shapes[name] = site_value.shape
+            self._r_numels[name] = site_value.numel()
+        self._trace_prob_evaluator = TraceEinsumEvaluator(trace,
+                                                          self._has_enumerable_sites,
+                                                          self.max_plate_nesting)
+
         if site_value is not None:
             mass_matrix_size = sum(self._r_numels.values())
             if self._adapter.is_diag_mass:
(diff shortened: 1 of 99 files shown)


neerajprad:
@fehiepsi : As you correctly point out there are a few layers to this task. I would suggest refactoring as follows:
 1. Make the TraceKernel class take in dict instead of trace, any interconversion between trace/dict happens inside the MCMC class. - Straight refactor, no change to external interface, current tests work as expected.
 2. A new optional parameter called `potential_fn` is introduced. Current model kwarg remains but is deprecated. If model is provided, we call `initiallize_model` to get `potential_fn` and `init_params`.
 3. The above changes are still backwards compatible, except some deprecated arguments. Next we can have a light alternate API layer that doesn't subclass from `TracePosterior`.

These can all be separate PRs. I would be happy to chip in, if there is any part that you would like me to work on.

fehiepsi:
Wow, your suggestions are really helpful! Thanks @neerajprad !

I would be great if you can code on your first point (refactoring TraceKernel class). :)

neerajprad:
> I would be great if you can code on your first point (refactoring TraceKernel class)

Great, I'll get started on that!

neerajprad:
@fehiepsi - could you confirm that there is no regression in total test times for `test_hmc.py` and `test_nuts.py`?

fehiepsi:
Sure, I'll do it.

pyro/infer/mcmc/util.py
```diff
@@ -208,3 +212,135 @@ def log_prob(self, model_trace):
...
+
+    dims = [frame.dim
+            for site in sites
+            for frame in site["cond_indep_stack"]
+            if frame.vectorized]
+    max_plate_nesting = -min(dims) if dims else 0
+    return max_plate_nesting
+
+
+def _transform_fn(transforms, params, invert=True):
```
neerajprad: Great to see this refactored out!

pyro/infer/mcmc/util.py
```diff
@@ -208,3 +212,135 @@ def log_prob(self, model_trace):
...
+    return {k: transforms[k](v) if not invert else transforms[k].inv(v)
+            for k, v in params.items()}
+
+
+def _pe_maker(model, model_args, model_kwargs, trace_prob_evaluator, transforms):
+    def potential_energy(params):
+        params_constrained = {k: transforms[k].inv(v) for k, v in params.items()}
+        cond_model = poutine.condition(model, params_constrained)
```
neerajprad: Just a quick note here - `poutine.condition` will convert all stochastic sites into observed sites. I don't think that should really change anything at this point, but just in case you see any weird results, this is worth checking.

pyro/infer/mcmc/hmc.py
```diff
@@ -30,6 +27,8 @@ class HMC(TraceKernel):
     Radford M. Neal
 
     :param model: Python callable containing Pyro primitives.
+    :param potential_fn: Python callable calculating potential energy with input
```
neerajprad: maybe also specify that the input parameters have real support.

pyro/infer/mcmc/util.py
fehiepsi: Hmm, I thought that I have deleted this function because it is not used elsewhere. ^^!

neerajprad: I meant `initialize_model` and its contents. I think this method can be deleted.

fehiepsi:
@neerajprad 

The following benchmark is for pre-this PR vs this PR

HMC:
+ logistic regression: 19.98s vs 20.52s
+ dirichlet categorical: 6.62s vs 6.37s
+ beta bernoulli: 8.09s vs 7.34s

NUTS:
+ gamma normal: 5.50s vs 5.23s
+ gamma beta: 13.90s vs 12.24s
+ gaussian mixture model: 26.54s vs 21.71s

Except HMC logistic regression, there is a gain with this PR. All diagnostic infos (acc rate, step size) are identical.

neerajprad:
> Except HMC logistic regression, there is a gain with this PR. All diagnostic infos (acc rate, step size) are identical.

Sounds great! I generally look at the total test time. For individual tests, it is worth checking whether the chosen step size changed due to a different initial trace for instance. That can lead to different times, but that's acceptable.

pyro/infer/mcmc/util.py
```diff
@@ -208,3 +212,131 @@ def log_prob(self, model_trace):
...
+
+def _get_init_params(model, model_args, model_kwargs, transforms, potential_fn, prototype_params,
+                     max_tries_initial_params=100):
+    params = prototype_params
+    for i in range(max_tries_initial_params):
+        potential_energy = potential_fn(params)
+        if not torch_isnan(potential_energy) and not torch_isinf(potential_energy):
+            return params
+        trace = poutine.trace(model).get_trace(*model_args, **model_kwargs)
+        samples = {name: trace.nodes[name]["value"].detach() for name in params}
```
neerajprad: Do we need to `.detach()` here since the only place where we set/unset `requires_grad` is the verlet integrator?

pyro/infer/mcmc/util.py
```diff
@@ -208,3 +212,131 @@ def log_prob(self, model_trace):
...
+    model_trace = poutine.trace(model).get_trace(*model_args, **model_kwargs)
+    has_enumerable_sites = False
+    prototype_samples = {}
+    for name, node in model_trace.iter_stochastic_nodes():
+        if isinstance(node["fn"], _Subsample):
+            continue
+        if node["fn"].has_enumerate_support:
+            has_enumerable_sites = True
+            continue
+        prototype_samples[name] = node["value"].detach()
```
neerajprad: Same here, do we need `.detach()` here?

neerajprad:
@fehiepsi - Since travis seems to be taking forever, I'm happy to merge this PR as is and do follow-up PRs if needed.

pyro/infer/mcmc/util.py
fehiepsi: It seems that without it, there is an error happen. I'll check it against.

fehiepsi: Yes, it is required when our model has `pyro.param(...)` statements (so sample sites will be leaf variables, which make us unable to change `requires_grad` flag in velocity_verlet)

pyro/infer/mcmc/util.py
```diff
@@ -208,3 +212,131 @@ def log_prob(self, model_trace):
...
+        optimized executable trace in the integrator.
+    :param dict jit_options: A dictionary contains optional arguments for
+        :func:`torch.jit.trace` function.
+    :param bool ignore_jit_warnings: Flag to ignore warnings from the JIT
+        tracer when ``jit_compile=True``. Default is False.
+    :returns: a tuple of (`initial_params`, `potential_fn`, `transforms`, `prototype_trace`)
+    """
+    # XXX `transforms` domains are sites' supports
+    if transforms is None:
+        automatic_transform_enabled = True
```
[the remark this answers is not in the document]
fehiepsi: I still can't find a good pattern here so I keep the current behaviour and would like to defer the change to another PR. The simplest change is to drop `transforms` arg. The better one (in my opinion), which covers a large number of usage cases, is to have `transforms={}` as default args and add default transforms for each latent site which does not appear in user-provided `transforms` arg. There we will need to check if codomain for each user-provided transform is `real` and raise an error when users provide wrong incompatible transforms.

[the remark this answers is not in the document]
neerajprad: I think your proposed change would be the best.

pyro/infer/mcmc/util.py
neerajprad: I see. That because `pyro.param` also sets the `requires_grad` flag to True, I guess.

fehiepsi:
@neerajprad We can resolve Memory error at https://github.com/pyro-ppl/pyro/issues/1730 with the strategy in your last PR (which is only yield params in workers):

```
class MCMC(TracePosterior):
    def __init__(...):
        self._model = kernel.model

    def _trace_wrap(self, z, *args, **kwargs):
        #if self.kernel.model is None:
        #    return z
        for name, transform in self._transforms.items():
            z[name] = transform.inv(z[name])
        z_trace = self._prototype_trace
        for name, value in z.items():
            z_trace.nodes[name]["value"] = value
        trace_poutine = poutine.trace(poutine.replay(self._model, z_trace))
        trace_poutine(*args, **kwargs)
        return trace_poutine.trace

        from pyro.infer.mcmc.util import initialize_model
        _, _, self._transforms, self._prototype_trace = initialize_model(self._model, args, kwargs)

        for sample in self.sampler._traces(*args, **kwargs):
            trace = self._trace_wrap(sample[0], *args, **kwargs)
            yield (trace, *sample[1:])
```
However this way will affect the performance because the main process will do a bunch of replaying job to get trace (at least this is the case for `baseball` example). How about we consume only samples (after transforming from params) instead of traces? And these samples will be populated when users call `posterior.exec_traces` if they want to do something with traces. At least, to get marginal posterior, we don't need these traces, only params and transforms are enough. This way also resolves https://github.com/pyro-ppl/pyro/issues/1787, where traces whole a bunch of big and non-used info such as `nodes['data']['fn'].mean`.

I guess you have thought about all of these issues with your big refactoring intention. I just want to share some of my observations and thoughts which might be helpful for you. :)

pyro/infer/mcmc/util.py
fehiepsi: I think so too.

neerajprad:
> How about we consume only samples (after transforming from params) instead of traces? And these samples will be populated when users call posterior.exec_traces

That's a good observation. I think the problem with the current interface is that MCMC inherits from `TracePosterior` and methods like `_traces` that return an iterator over traces. I think rather than catering to this interface at all, it might be better to build new posterior / predictive wrappers and address these issues there. Hopefully, we can then go and deprecate `TracePosterior` if the newly designed interface works well.

jiamings:
Thanks for the implementation! But I am not sure if there is a bug in the HMC code, it seems that the potential_fn argument in __init__() is never used? The self.potential_fn is always overridden by the one derived from the model.

PR #1847, opened 2019-05-03
Bump version to 0.3.3

neerajprad:
This bumps up the version of Pyro from `0.3.2` to `0.3.3`. Since `0.3.2` was released quite recently, this mostly has some minor code changes due to functions that got deprecated in `torch==1.1.0`.

### DIFF
--- a/examples/air/main.py
+++ b/examples/air/main.py
@@ -246,7 +246,7 @@ def main(**kwargs):
 
 
 if __name__ == '__main__':
-    assert pyro.__version__.startswith('0.3.2')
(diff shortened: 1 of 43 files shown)


fritzo:
> i feel like checking major and minor versions are enough for the examples

Historically we have kept examples updated with syntax that becomes available in patch releases, and these minor changes can lead to difficult-to-diagnose errors for users with mismatched versions.

neerajprad:
@fritzo, @jpchen - Can we merge this? I'll cut a release once its in dev.

PR #1848, opened 2019-05-03
 Change MCMC kernels to not depend on trace

neerajprad:
As discussed in #1845, this renames `TraceKernel` to `MCMCKernel` and changes the interface so that the kernels take in dicts instead of model traces. 

I have tried to make minimal changes. ~but since the current interface is very coupled with the `model`, I had to make slightly more changes than I had planned~. Existing tests work as expected.

### DIFF
--- a/pyro/infer/mcmc/mcmc.py
+++ b/pyro/infer/mcmc/mcmc.py
@@ -266,8 +266,9 @@ class MCMC(TracePosterior):
             self.sampler = _SingleSampler(kernel, num_samples, self.warmup_steps, disable_progbar)
         super(MCMC, self).__init__(num_chains=num_chains)
 
-    # TODO: Refactor so that the MCMC class directly has access to the
-    # trace generator and transforms needed to do this wrapping.
+    # TODO: Refactor so that the MCMC class directly has access to the trace generator
+    # and transforms needed to do this wrapping. Note that only unconstrained parameters
+    # are passed to `MCMCKernel` classes.
     def _trace_wrap(self, z, *args, **kwargs):
         for name, transform in self.kernel.transforms.items():
             z[name] = transform.inv(z[name])
--- a/pyro/infer/mcmc/mcmc.py
+++ b/pyro/infer/mcmc/mcmc.py
@@ -195,6 +195,19 @@ class _SingleSampler(TracePosterior):
             self.logger.info(diagnostics, extra={"msg_type": DIAGNOSTIC_MSG})
             yield params
 
+    # TODO: Refactor so that this class directly has access to the trace generator
+    # and transforms needed to do this wrapping. Note that only unconstrained parameters
+    # are passed to `MCMCKernel` classes.
+    def _trace_wrap(self, z, *args, **kwargs):
+        for name, transform in self.kernel.transforms.items():
+            z[name] = transform.inv(z[name])
+        z_trace = self.kernel._prototype_trace
+        for name, value in z.items():
+            z_trace.nodes[name]["value"] = value
+        trace_poutine = poutine.trace(poutine.replay(self.kernel.model, z_trace))
+        trace_poutine(*args, **kwargs)
+        return trace_poutine.trace
+
     def _traces(self, *args, **kwargs):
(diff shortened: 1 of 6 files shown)


neerajprad:
@fehiepsi - Feel free to commit to this PR directly, if there is anything you would like to add. I think this should be independent of your changes.

neerajprad:
Hmm..this change is not compatible with `pyro.param` statements in the model captured by the failing test. While we don't need `pyro.param` for MCMC, our existing implementation that uses `prototype_trace` handles this without any issues, whereas this proposed change doesn't. I'll think about a fix for this.

pyro/infer/mcmc/mcmc_kernel.py
```diff
@@ -42,34 +42,34 @@ def end_warmup(self):
...
+        Sets the parameters to initiate the MCMC run.
         """
         raise NotImplementedError
 
     @abstractmethod
-    def sample(self, trace):
+    def sample(self, params):
         """
-        Samples a trace from the approximate posterior distribution, when given an existing trace.
+        Samples parameters from the posterior distribution, when given an existing trace.
```
fehiepsi: typo: when given existing *parameters*.

pyro/infer/mcmc/mcmc_kernel.py
```diff
@@ -42,34 +42,34 @@ def end_warmup(self):
...
 
-        :param trace: Current execution trace.
+        :param trace: Current parameter values.
```
fehiepsi: typo: `:param params:`

pyro/infer/mcmc/mcmc.py
```diff
@@ -247,6 +248,7 @@ class MCMC(TracePosterior):
     """
     def __init__(self, kernel, num_samples, warmup_steps=None,
                  num_chains=1, mp_context=None, disable_progbar=False):
+        self.kernel = kernel
```
fehiepsi: We need to revise docs for `kernel` too.

pyro/infer/mcmc/mcmc.py
```diff
@@ -264,9 +266,21 @@ def __init__(self, kernel, num_samples, warmup_steps=None,
             self.sampler = _SingleSampler(kernel, num_samples, self.warmup_steps, disable_progbar)
         super(MCMC, self).__init__(num_chains=num_chains)
 
+    # TODO: Refactor so that the MCMC class directly has access to the
+    # trace generator and transforms needed to do this wrapping.
```
fehiepsi: Could you also add a TODO for mentioning about the params from kernel lie in unconstrained space?

neerajprad:
@fehiepsi - I have addressed your comments, but I realized that we will need to address the TODO to make the multiprocessing tests pass. The issue is that we cannot access `prototype_trace` or `transforms` from `self.kernel` in `MCMC` since they are set in a different process with python multiprocessing. 

This will require a refactoring so that we can pass model to `MCMC` directly and call `initialize_model` to get `transforms`, `prototype_trace`, `initial_params` within the `MCMC` class. That makes it much harder to have a backwards compatible change though. I am thinking that it might be worth breaking backwards compatibility and clean up all the interface issues for once. 

For the time being, please feel free to pull in these changes into your PR, or we can merge your PR first in case that's ready.

fehiepsi:
Thanks @neerajprad ! I'll pull these changes to my PR and think about what we can do to address multiprocessing issue.

fehiepsi:
@neerajprad How about move `trace_wrap` logic to workers? There it is safe to get access to prototype trace and transform_fn.

neerajprad:
> @neerajprad How about move trace_wrap logic to workers? There it is safe to get access to prototype trace and transform_fn.

That's a good idea, let me change this and report back.

neerajprad:
@fehiepsi - your suggestion worked great. I think this should be good to merge once tests pass, pending further comments.

fehiepsi:
Yeah, I also have an almost working version in my branch. Currently, tests only fail for some edge cases such as empty param or adapt_step_size=False.

PR #1849, opened 2019-05-04
Add different initialization methods for autoguides

### LINKED ISSUE #1850: [bug] Continuous autoguides do not take device into account
### Issue Description
Many of the `AutoContinuous` guides do not take into account that the model might be on the GPU instead of the CPU. For example, see the code snippet below, where it fails if I pass `device='cuda'` to the tensors I create.

It would of course be nice if it instead relied on the device given from the model trace.
Do you have any suggestions?

Thanks!

### Environment
 - OS and python version: Windows 10, Python 3.7
 - PyTorch version: 1.1
 - Pyro version: 0.3.3

### Code Snippet
```python
import pyro
import torch
from pyro.contrib.autoguide import AutoDiagonalNormal
from pyro.optim import Adam


def model(xs, prior_loc, prior_scale):
    mu = pyro.sample('mu', dist.Normal(prior_loc, prior_scale))
    return pyro.sample('obs', dist.Normal(mu, prior_scale), obs=xs)


def train():
    xs = torch.tensor([5, 5.5, 5.6, 5.8, 6.3], device='cuda')
    prior_loc = torch.tensor(0.0, device='cuda')
    prior_scale = torch.tensor(1.0, device='cuda')
    optim = Adam({'lr': 1e-3})
    loss = Trace_ELBO()
    svi = SVI(model, guide, optim, loss)
    svi.step(xs, prior_loc, prior_scale)

if __name__ == "__main__":
  train()
```

Error:
```
  File "C:\tools\miniconda3\envs\scanner-pyro\lib\site-packages\IPython\core\interactiveshell.py", line 3265, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-6-68d49e4377f9>", line 26, in <module>
    train()
  File "<ipython-input-6-68d49e4377f9>", line 23, in train
    svi.step(xs, prior_loc, prior_scale)
  File "C:\tools\miniconda3\envs\scanner-pyro\lib\site-packages\pyro\infer\svi.py", line 99, in step
  File "C:\tools\miniconda3\envs\scanner-pyro\lib\site-packages\pyro\infer\trace_elbo.py", line 136, in loss_and_grads
    surrogate_loss_particle.backward(retain_graph=self.retain_graph)
  File "C:\tools\miniconda3\envs\scanner-pyro\lib\site-packages\torch\tensor.py", line 107, in backward
    torch.autograd.backward(self, gradient, retain_graph, create_graph)
  File "C:\tools\miniconda3\envs\scanner-pyro\lib\site-packages\torch\autograd\__init__.py", line 93, in backward
    allow_unreachable=True)  # allow_unreachable flag
RuntimeError: Function AddBackward0 returned an invalid gradient at index 1 - expected type torch.cuda.FloatTensor but got torch.FloatTensor
```

fritzo:
This PR
- enables customization of initialization strategy in pyro.contrib.autoguide
- adds two new strategies: mean and median
- changes the default strategy to `init_to_median` (previously `Delta` used a single sample and `Auto*Normal` used an arbitrary feasible point).

This was motivated by a project I'm working on where I realized that the `AutoDiagonalNormal` initialization completely ignored the prior. In my project, `init_to_mean` converges faster (by orders of magnitude).

Resolves #1850
I've also added fixes to initialization code so as to preserve dtype and device info.

## Tested
- added a smoke test
- added a convergence test with `test_median`

### DIFF
--- a/pyro/contrib/autoguide/initialization.py
+++ b/pyro/contrib/autoguide/initialization.py
@@ -1,35 +1,89 @@
+r"""
+The pyro.contrib.autoguide module contains initialization functions for
+automatic guides.
+
+The standard interface for initialization is a function that inputs a Pyro
+trace ``site`` dict and returns an appropriately sized ``value`` to serve
+as an initial constrained value for a guide estimate.
+"""
 from __future__ import absolute_import, division, print_function
 
 import torch
 from torch.distributions import transform_to
 
+from pyro.poutine.messenger import Messenger
+from pyro.util import torch_isnan
 
-@torch.no_grad()
+
 def init_to_feasible(site):
-    value = site["value"].detach()
+    """
+    Initialize to an arbitrary feasible point, ignoring distribution
+    parameters.
+    """
+    value = site["fn"].sample().detach()
     t = transform_to(site["fn"].support)
     return t(torch.zeros_like(t.inv(value)))
 
 
-@torch.no_grad()
 def init_to_sample(site):
-    return site["value"].detach()
+    """
+    Initialize to a random sample from the prior.
+    """
+    return site["fn"].sample().detach()
 
 
-@torch.no_grad()
-def init_to_mean(site):
+def init_to_median(site, num_samples=15):
+    """
+    Initialize to the prior median; fallback to a feasible point if median is
+    undefined.
+    """
     try:
-        value = site["fn"].mean.detach()
+        # Try to compute empirical median.
+        samples = site["fn"].sample(sample_shape=(num_samples,))
+        value = samples.median(dim=0)[0]
+        if torch_isnan(value):
+            raise ValueError
         if hasattr(site["fn"], "_validate_sample"):
-            # Fall back to a feasible point for distributions with
-            # infinite variance, e.g. Cauchy.
             site["fn"]._validate_sample(value)
         return value
-    except (NotImplementedError, ValueError):
+    except (RuntimeError, ValueError):
+        # Fall back to feasible point.
         return init_to_feasible(site)
 
 
-@torch.no_grad()
-def init_to_median(site, num_samples=32):
-    samples = site["fn"].sample(sample_shape=(num_samples,))
-    return samples.median(dim=0)[0]
+def init_to_mean(site):
+    """
+    Initialize to the prior mean; fallback to median if mean is undefined.
+    """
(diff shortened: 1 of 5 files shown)


fritzo:
cc @stefanwebb

fritzo:
Thanks for reviewing @martinjankowiak I've updated the docs and added some links.

> how might a `init_to_map` method look like?

Accounting for data is quite a bit more complicated. Whereas initializing to some-aspect-of-the-prior is simple and requires only a single pass over the model, implementing an optimization step would require also examining data in observe statements. I think the best path forward is to slowly refactor the autoguide library to make it more composable, so we can run `AutoDelta` to initialize a later `AutoDiagonalNormal`, and then use that to initialize an `AutoLowRankMultivariateNormal`.

pyro/contrib/autoguide/initialization.py
```diff
@@ -1,35 +1,89 @@
...
-@torch.no_grad()
 def init_to_feasible(site):
-    value = site["value"].detach()
+    """
+    Initialize to an arbitrary feasible point, ignoring distribution
+    parameters.
+    """
+    value = site["fn"].sample().detach()
     t = transform_to(site["fn"].support)
     return t(torch.zeros_like(t.inv(value)))
```
fehiepsi: This seems like you are taking inverse transform of `0` instead of `arbitrary`. What do you think about initializing uniformly in (-2, 2) in unconstrained space (which is Stan approach)?

fritzo: I think Stan's approach will not work well in Pyro. This zero=arbitrary function is added for backwards compatibility only. After this PR, all autoguides will init to the prior median.

fehiepsi: From my experience (with MAP inference for a bunch of small models), using median still does not work well if priors are not good enough. To resolve that, I did evaluate model's log_probs for a bunch of random init values and select the init value with largest log_prob. If you find that `median` does not work then you can give that method a try. :)

fritzo: > evaluate model's log_probs for a bunch of random init values

Sounds reasonable. If I end up having trouble with autoguide init, maybe we can share some machinery with HMC.

PR #1854, opened 2019-05-07
Add test for trace predictive with complex distributions and fix correlations

### LINKED ISSUE #1858: [discussion] Keeping subsample correlations in TracePredictive
While developing #1854, I noticed that we do new random subsampling for each stochastic site in `TracePredictive` distribution. After reflecting about this, I have figured out that this might drop correlations between the sampled parameters under a subsampling statement.

I think what one needs to do is to fix a subset of random indices for each subsample site for all the sites. Any ideas on how one could this in a nice fashion?

My current idea is to create a dictionary when fixing the subsample sites, that map each subsample site to the set of random indices based on the provided subsample site size. Would this work?

Thank you in advance! 😄

ahmadsalim:
In connection with #1837 .
CC: @neerajprad 

*EDIT:* Fixes #1858

### DIFF
--- a/pyro/infer/abstract_infer.py
+++ b/pyro/infer/abstract_infer.py
@@ -326,7 +326,10 @@ class TracePredictive(TracePosterior):
                     # Select random sub-indices to replay values under conditionally independent stacks.
                     # Otherwise, we assume there is an dependence of indexes between training data
                     # and prediction data.
-                    batch_dim = cis.dim - site["fn"].event_dim
+                    if site_is_subsample(site):
+                        batch_dim = cis.dim
+                    else:
+                        batch_dim = cis.dim - site["fn"].event_dim
                     subidxs = torch.randint(0, site['value'].size(batch_dim), (cis.size,),
                                             device=site["value"].device)
                     site["value"] = site["value"].index_select(batch_dim, subidxs)
--- a/pyro/infer/abstract_infer.py
+++ b/pyro/infer/abstract_infer.py
@@ -313,28 +313,23 @@ class TracePredictive(TracePosterior):
                 continue
 
     def _adjust_to_data(self, trace, data_trace):
-        for name, site in list(trace.nodes.items()):
+        for name, site in trace.iter_stochastic_nodes():
             # Adjust subsample sites
             if site_is_subsample(site):
                 site["fn"] = data_trace.nodes[name]["fn"]
                 site["value"] = data_trace.nodes[name]["value"]
             # Adjust sites under conditionally independent stacks
-            try:
-                site["cond_indep_stack"] = data_trace.nodes[name]["cond_indep_stack"]
-                site["fn"] = data_trace.nodes[name]["fn"]
-                for cis in site["cond_indep_stack"]:
-                    # Select random sub-indices to replay values under conditionally independent stacks.
-                    # Otherwise, we assume there is an dependence of indexes between training data
-                    # and prediction data.
-                    if site_is_subsample(site):
-                        batch_dim = cis.dim
-                    else:
-                        batch_dim = cis.dim - site["fn"].event_dim
+            site["cond_indep_stack"] = data_trace.nodes[name]["cond_indep_stack"]
+            site["fn"] = data_trace.nodes[name]["fn"]
+            for cis in site["cond_indep_stack"]:
+                # Select random sub-indices to replay values under conditionally independent stacks.
+                # Otherwise, we assume there is an dependence of indexes between training data
+                # and prediction data.
+                if not site_is_subsample(site):
+                    batch_dim = cis.dim - site["fn"].event_dim
                     subidxs = torch.randint(0, site['value'].size(batch_dim), (cis.size,),
                                             device=site["value"].device)
                     site["value"] = site["value"].index_select(batch_dim, subidxs)
-            except KeyError:
-                pass
 
     def marginal(self, sites=None):
         """
--- a/pyro/infer/abstract_infer.py
+++ b/pyro/infer/abstract_infer.py
@@ -313,23 +313,28 @@ class TracePredictive(TracePosterior):
                 continue
 
     def _adjust_to_data(self, trace, data_trace):
+        subsampled_idxs = dict()
         for name, site in trace.iter_stochastic_nodes():
             # Adjust subsample sites
             if site_is_subsample(site):
                 site["fn"] = data_trace.nodes[name]["fn"]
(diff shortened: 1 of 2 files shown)


pyro/infer/abstract_infer.py
```diff
@@ -326,7 +326,10 @@ def _adjust_to_data(self, trace, data_trace):
                     # Select random sub-indices to replay values under conditionally independent stacks.
                     # Otherwise, we assume there is an dependence of indexes between training data
                     # and prediction data.
-                    batch_dim = cis.dim - site["fn"].event_dim
+                    if site_is_subsample(site):
+                        batch_dim = cis.dim
```
neerajprad: I don't think we should need this because subsample sites shouldn't have non-empty `cond_indep_stack`. I'm also noticing that we can remove the try-catch block by only iterating over `sample` sites to begin with, something like this:

```diff
     def _adjust_to_data(self, trace, data_trace):
-        for name, site in list(trace.nodes.items()):
+        for name, site in trace.iter_stochastic_nodes():
             # Adjust subsample sites
             if site_is_subsample(site):
                 site["fn"] = data_trace.nodes[name]["fn"]
                 site["value"] = data_trace.nodes[name]["value"]
             # Adjust sites under conditionally independent stacks
-            try:
-                site["cond_indep_stack"] = data_trace.nodes[name]["cond_indep_stack"]
-                site["fn"] = data_trace.nodes[name]["fn"]
-                for cis in site["cond_indep_stack"]:
-                    # Select random sub-indices to replay values under conditionally independent stacks.
-                    # Otherwise, we assume there is an dependence of indexes between training data
-                    # and prediction data.
-                    if site_is_subsample(site):
-                        batch_dim = cis.dim
-                    else:
-                        batch_dim = cis.dim - site["fn"].event_dim
-                    subidxs = torch.randint(0, site['value'].size(batch_dim), (cis.size,),
-                                            device=site["value"].device)
-                    site["value"] = site["value"].index_select(batch_dim, subidxs)
-            except KeyError:
-                pass
+            site["cond_indep_stack"] = data_trace.nodes[name]["cond_indep_stack"]
+            site["fn"] = data_trace.nodes[name]["fn"]
+            for cis in site["cond_indep_stack"]:
+                # Select random sub-indices to replay values under conditionally independent stacks.
+                # Otherwise, we assume there is an dependence of indexes between training data
+                # and prediction data.
+                batch_dim = cis.dim - site["fn"].event_dim
+                subidxs = torch.randint(0, site['value'].size(batch_dim), (cis.size,),
+                                        device=site["value"].device)
+                site["value"] = site["value"].index_select(batch_dim, subidxs)
```

Do you think this will work for all your use cases?

ahmadsalim:
Thanks for the comments!
I added the suggested improvements.

I however, have experienced some nested plates where the inner one had a non-empty `cond_indep_stack`, so I kept the condition for now. Do you want me to add another test for this?

pyro/infer/abstract_infer.py
ahmadsalim: Yes, that will cover my use cases! 😄

neerajprad:
Let us discuss #1858 before merging this.

pyro/infer/abstract_infer.py
```diff
@@ -313,25 +313,23 @@ def _remove_dropped_nodes(self, trace):
...
-                    # Select random sub-indices to replay values under conditionally independent stacks.
-                    # Otherwise, we assume there is an dependence of indexes between training data
-                    # and prediction data.
+            site["cond_indep_stack"] = data_trace.nodes[name]["cond_indep_stack"]
+            site["fn"] = data_trace.nodes[name]["fn"]
+            for cis in site["cond_indep_stack"]:
+                # Select random sub-indices to replay values under conditionally independent stacks.
+                # Otherwise, we assume there is an dependence of indexes between training data
+                # and prediction data.
+                if not site_is_subsample(site):
```
neerajprad: Correct me if I'm mistaken, but I don't think you should ever get here if `site_is_subsample` is True.

ahmadsalim: You are probably right, so I have changed the `if` to an assertion. 😄 
Maybe I should add a test with nested plates as well.

neerajprad:
LGTM. I think you might need to reduce the precision for the one hot categorical test, as it's failing the threshold on travis by a small margin. Also, if possible could you run the tests on CUDA, and confirm that they pass?

```sh
CUDA_TEST=1 PYRO_TENSOR_TYPE=torch.cuda.DoubleTensor pytest -vx tests/infer/test_abstract_infer.py
```

ahmadsalim:
It seems to work if I try to set the environment variables and run it 😄

tests/infer/test_abstract_infer.py
```diff
@@ -81,6 +88,19 @@ def test_posterior_predictive_svi_auto_diag_normal_guide():
...
 
 
+def test_posterior_predictive_svi_one_hot():
+    pseudocounts = torch.ones(3) * 0.1
+    true_probs = torch.tensor([0.15, 0.6, 0.25])
+    classes = dist.OneHotCategorical(true_probs).sample((1000,))
+    opt = optim.Adam(dict(lr=0.1))
+    loss = Trace_ELBO()
+    guide = AutoDiagonalNormal(one_hot_model)
+    svi_run = SVI(one_hot_model, guide, opt, loss, num_steps=1000, num_samples=100).run(pseudocounts, classes=classes)
```
neerajprad: This test still seems to fail - I think you might want to increase the `num_samples` to `1000`.

ahmadsalim: OK, I will try again 😄 . It's a bit weird it kept failing, since I ran it 10 times locally without any issue

ahmadsalim:
I have tried making it more stable now, but it seems it's failing for another reason?
Any ideas? Thanks!

neerajprad:
@ahmadsalim - This should be fixed by #1864. Feel free to pull in that change in the meantime.

ahmadsalim:
Awesome, thanks for merging 😄

deoxyribose:
Hi, the assertion `assert not site_is_subsample(site)` in line 332 still fails on models where the inner plate has a non-empty cond_indep_stack. Here's a test and fix (just what @ahmadsalim had in his previous version, but with the changes discussed in [#1858](https://github.com/pyro-ppl/pyro/issues/1858#issuecomment-490593182): https://github.com/deoxyribose/compositions_mixtures_factors/blob/master/TracePredictive%20nested%20plates%20test.ipynb

PR #1855, opened 2019-05-07
Fix examples/lda.py and support --jit

### LINKED ISSUE #1841: LDA tutorial: incorrect tensor shape when document size is larger than the vocabulary size
https://github.com/pyro-ppl/pyro/blob/56c06176763f1676bd586c89b2dc503c64315f0a/examples/lda.py#L101

The filler array of ones should have expanded into the size of the sliced data. This is problematic when the number of words per document is larger than the vocabulary size
Try 
`counts.scatter_add_(0, data[:, ind], torch.tensor(1.).expand(data[:,ind].shape))`
instead

fritzo:
Fixes #1841 

## Tested
- ran locally with words smaller than doc size
- removed xfail marker from jit test

### DIFF
--- a/examples/lda.py
+++ b/examples/lda.py
@@ -92,13 +92,11 @@ def parametrized_guide(predictor, data, args, batch_size=None):
     # Use an amortized guide for local variables.
     pyro.module("predictor", predictor)
     with pyro.plate("documents", args.num_docs, batch_size) as ind:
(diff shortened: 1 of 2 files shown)


fritzo:
@neerajprad can you please merge this before our extra cores disappear :smile:

PR #1859, opened 2019-05-09
Sylvester Flow

stefanwebb:
Implements (non-amortized) Sylvester Flow of the Householder type


robsalomone:
Thanks for this @stefanwebb , I was going to suggest implementing this! If it is not too much trouble, could you also implement the Triangular flow? It is simple by comparison to the Householder one. 

Might I suggest also implementing a non-amortised version of [Hamiltonian VAE](https://arxiv.org/pdf/1805.11328.pdf)?

stefanwebb:
Glad you found it useful, @robsalomone!

I thought about implementing triangular and orthogonal Sylvester flow but decided against it. They require the user to know the details of how they work and would likely lead to confusion... E.g. for triangular Sylvester flow, the layers of flow have to have alternating upper and lower triangular Q, and this contract would have to be enforced by the user rather than Pyro. Also, the results from the paper show a negligible difference between the types of Sylvester flow.

Do go ahead and submit a PR though if you feel strongly about this :)

stefanwebb:
I've had some issues with the commit history, so I'll close this and try again, starting from a fresh `dev` upstream branch

PR #1861, opened 2019-05-09
Respect transforms for new hmc api

fehiepsi:
This makes a quick fix so that users using dev branch can get expected behaviour. Currently, we can use the new `initialize_model` utility to feed `potential_fn` and `transforms` into HMC kernel, but MCMC only returns params in unconstrained space. This PR makes it return samples in constrained space instead.

### DIFF
--- a/pyro/infer/mcmc/mcmc.py
+++ b/pyro/infer/mcmc/mcmc.py
@@ -190,10 +190,11 @@ class _SingleSampler(TracePosterior):
     # and transforms needed to do this wrapping. Note that only unconstrained parameters
     # are passed to `MCMCKernel` classes.
     def _trace_wrap(self, z, *args, **kwargs):
+        if self.kernel.transforms:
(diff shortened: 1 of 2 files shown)


pyro/infer/mcmc/mcmc.py
```diff
@@ -190,10 +190,11 @@ def _gen_samples(self, num_samples, init_params):
     # and transforms needed to do this wrapping. Note that only unconstrained parameters
     # are passed to `MCMCKernel` classes.
     def _trace_wrap(self, z, *args, **kwargs):
+        if self.kernel.transforms:
```
neerajprad: Should we also address the other issue of `init_params` being respected with `models` in this PR?

fehiepsi: Yup, I just add a quick fix for it.

neerajprad:
Thanks for fixing this in dev, @fehiepsi.

PR #1864, opened 2019-05-13
Register pytest marker to avoid warning on latest pytest version

neerajprad:
On the latest pytest version, any custom markers from plugins need to be explicitly registered to avoid [PytestUnknownMarkWarning](https://docs.pytest.org/en/latest/mark.html#raising-errors-on-unknown-marks).

Our dev branch won't build until this is merged.


neerajprad:
@jpchen - could you merge this to unblock dev?

neerajprad:
Thanks for merging, @fehiepsi!

PR #1865, opened 2019-05-14
Sylvester Flow

stefanwebb:
(Non-amortized) Sylvester flow with Householder transforms.

This time the commit history is clean!


stefanwebb:
@martinjankowiak this is ready to be looked at! I've made SylvesterFlow inherit from HouseholderFlow and cleaned up some of the logic

pyro/distributions/sylvester.py
```diff
@@ -0,0 +1,136 @@
...
+        return self.R_dense * self.triangular_mask + torch.diag(self.tanh(self.R_diag))
+
+    # Construct upper diagonal S matrix
+    def S(self):
+        return self.S_dense * self.triangular_mask + torch.diag(self.tanh(self.S_diag))
+
+    # Construct orthonomal matrix using Householder flow
+    def Q(self):
+        u = self.u()
+        partial_Q = torch.eye(self.input_dim) - 2. * torch.ger(u[0], u[0])
```
martinjankowiak: `torch.eye(self.input_dim)` this will lead to problems on cuda

pyro/distributions/sylvester.py
```diff
@@ -0,0 +1,136 @@
...
+
+        for idx in range(1, self.count_transforms):
+            partial_Q = torch.matmul(partial_Q, torch.eye(self.input_dim) - 2. * torch.ger(u[idx], u[idx]))
+
+        return partial_Q
+
+    # NOTE: self.u_unnormed is initialized in parent class
+    def reset_parameters2(self):
```
martinjankowiak: is there a way to do this by overriding`reset_parameters` so you don't have 2 of them floating around?

pyro/distributions/householder.py
```diff
@@ -57,6 +57,7 @@ class HouseholderFlow(TransformModule):
     codomain = constraints.real
     bijective = True
     event_dim = 1
+    volume_preserving = True
```
martinjankowiak: is `volume_preserving` used anywhere?

pyro/distributions/householder.py
```diff
@@ -68,12 +69,17 @@ def __init__(self, input_dim, count_transforms=1):
                 "Number of Householder transforms, {}, is greater than input dimension {}, which is an \
 over-parametrization!".format(count_transforms, input_dim))
         self.count_transforms = count_transforms
-        self.u = nn.Parameter(torch.Tensor(count_transforms, input_dim))
+        self.u_unnormed = nn.Parameter(torch.Tensor(count_transforms, input_dim))
```
martinjankowiak: can you make `assert count_transforms > 0` raise a `ValueError` instead?

pyro/distributions/sylvester.py
stefanwebb: Overriding `reset_parameters` would make things difficult. If I overrode it, I would need to change the position of `super(SylvesterFlow, self).__init__(input_dim, count_transforms)`  so that it didn't try to initialize `self.R_dense` etc. before they were created. But creating `self.R_dense` before the initializer is called would introduce a bug because the initializer for `torch.nn.Module` won't have been called yet, and those parameters won't be added to `self._modules`. Any solution to this would be much more complicated than just calling the method `reset_parameters2`

pyro/distributions/sylvester.py
stefanwebb: Done!

pyro/distributions/householder.py
stefanwebb: Not yet, but I think it's useful in case we want to incorporate it into the tests in the future. For instance, I use a similar `autoregressive = True` flag in other flows so I know how to run the Jacobian test

pyro/distributions/householder.py
stefanwebb: Is raising an exception not too extreme? It will still work when `assert count_transforms > 0`?

stefanwebb:
I was actually thinking about a more extreme reorganization and wanted to do it in a separate PR so we can discuss. Basically, I want to put transforms into `pyro.transforms` by default, and copy them into `pyro.distributions` for backwards compatibility

pyro/distributions/householder.py
martinjankowiak: i mean raise the error if that's false, e.g. if it's equal to 0 or -7

martinjankowiak:
@stefanwebb before pushing an actual `pyro.transforms` PR you might just want to quickly outline what you want to do and why. then people like @fritzo who have a finer feeling for codebase organization etc. can weigh in

pyro/distributions/householder.py
stefanwebb: Done!

fritzo:
> reorganization

We've historically thought of `pyro.distributions` as a component that is independent of the rest of Pyro and that extends `torch.distributions`. From that view, it seems like a nice organization would be `pyro.distributions.transforms`, extending `torch.distributions.transforms`.

stefanwebb:
It should be ready to go once the tests pass. I'll reorganize to `pyro.distributions.transforms` in a separate PR

martinjankowiak:
@stefanwebb great thanks!

PR #1867, opened 2019-05-15
get float acc instead of scalar tensor in svdkl

fehiepsi:
This fixes a typo in reporting accuracy of the model: if `correct` is a scalar long tensor `tensor(9943)`, then `100. * correct` is also a long tensor, hence the reporting result of `100. * tensor(9943) / 10000` will be `99` instead of 99.43.

### DIFF
--- a/examples/contrib/gp/sv-dkl.py
+++ b/examples/contrib/gp/sv-dkl.py
@@ -70,6 +70,7 @@ def train(args, train_loader, gpmodule, optimizer, loss_fn, epoch):
         loss = loss_fn(gpmodule.model, gpmodule.guide)
         loss.backward()
         optimizer.step()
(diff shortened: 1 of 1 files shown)


examples/contrib/gp/sv-dkl.py
```diff
@@ -70,6 +70,7 @@ def train(args, train_loader, gpmodule, optimizer, loss_fn, epoch):
         loss = loss_fn(gpmodule.model, gpmodule.guide)
         loss.backward()
         optimizer.step()
+        batch_idx = batch_idx + 1
```
fehiepsi: We report the number of training data has been trained in the epoch at `batch_idx`, so for `batch_idx = 0`, we should report `len(batch_data)` instead of `0`.

PR #1868, opened 2019-05-16
Docs for existing flows

stefanwebb:
I forgot to add entries in the docs for some flows that were previously merged into Pyro, so here they are...


stefanwebb:
I'm not sure why the examples fail in the Unit Tests... It seems like the error you get when the `pillow` package isn't installed

stefanwebb:
Does the configuration need to be changed?

neerajprad:
Restarted the build, should be good to merge if tests pass.

PR #1870, opened 2019-05-19
Fix device placement for autocorrelation tensors

activatedgeek:
The function `effective_sample_size` was failing with CUDA tensors because of the newly created tensors not being created on the same device.


fehiepsi:
Thanks @activatedgeek !

activatedgeek:
@fehiepsi There appears to be a `ImportError: cannot import name 'imresize'`. I'm guessing this does not stem from my changes.

fehiepsi:
Yup, I am creating an issue for it. Looking like scipy deprecates that function.

Issue #1871, opened 2019-05-20
scipy imresize is deprecated


fehiepsi:
`imresize` is used in `air` example, but it is deprecated now. We need to replace it by something else to make tests pass.

More context about this issue: `imresize` is used in [observations](https://github.com/edwardlib/observations) but this library is deprecated now in favour of tensorflow [datasets](https://github.com/tensorflow/datasets). But multi-mnist is not available in tensorflow's datasets yet. So I'm not sure what is a good solution for this issue.

neerajprad:
The problem is with the `observations` library (that uses this function) which is not being maintained anymore. I think we need to remove dependency on `observations`.

null-a:
> We need to replace it by something else

The [docs](https://docs.scipy.org/doc/scipy-1.2.0/reference/generated/scipy.misc.imresize.html) suggest using `numpy.array(Image.fromarray(arr).resize())` inplace of `imresize`. This requires `PIL` which AIR already has as a dependency.

PR #1872, opened 2019-05-20
Downgrade scipy dependency for travis tests to pass

neerajprad:
This enables the travis tests to pass again by using a lower version of scipy that has `imresize` function used by the AIR tutorial. This is needed until https://github.com/pyro-ppl/pyro/issues/1871 is fixed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -97,7 +97,9 @@ setup(
             'nbval',
             'pytest>=4.1',
             'pytest-cov',
-            'scipy>=1.1',
+            # TODO: remove once https://github.com/pyro-ppl/pyro/issues/1871
+            # is fixed.
+            'scipy>=1.1, <1.3',
         ],
         'profile': ['prettytable', 'pytest-benchmark', 'snakeviz'],
         'dev': EXTRAS_REQUIRE + [


mattwescott:
Built a docker image at 3ac5a02e0e6b0a11ae796707413c11df2c14ff6b, and it seems to have `scipy==1.3` still

`make test-examples` is giving me `cannot import name 'imresize'` (#1871) until I manually downgrade.

neerajprad:
@mattwescott - You will need to manually remove `scipy==1.3` first. This change is for our CI, and for users who have not already updated scipy to 1.3. If you already have the latest scipy version, you will need to remove it, and run `make install` which should pick up the right scipy version that has the `imresize` function. A permanent fix for this is to remove dependency on the `observations` library.

mattwescott:
@neerajprad That makes sense for my native environment, but I'd expect the docker build to emerge with the right version. What am I missing?

neerajprad:
> but I'd expect the docker build to emerge with the right version. What am I missing?

I would expect the same. Could you point me to how you are building the docker image?

mattwescott:
@neerajprad Rebuilt with `--no-cache` and got `scipy==1.3` again

Looks like it is coming from the other scipy dependency statement in `setup.py`: https://github.com/pyro-ppl/pyro/pull/1872/commits/b06e5b5872764fab4da7484d9c5105e129656761#diff-2eeaed663bd0d25b7e608891384b7298R115

neerajprad:
> Looks like it is coming from the other scipy dependency statement in setup.py:

That's my bad, I forgot that we had a dev dependency too. Will rectify it asap. Thanks for pointing out!

PR #1879, opened 2019-05-22
Support outside contributions

mattwescott:
### Description

Support configuring the location of the pyro repo. 

I'm planning to use this to run containerized tests on my fork before submitting another small PR. I've left the image naming alone. My thinking is that development will typically be based on a single repo, so name collisions will be infrequent. 

Note that the default repo location now lives in both `docker/Makefile` and `docker/install.sh`. It might be cleaner to gate on the existence of a `PYRO_REPO_URL` environment variable?

### Testing

* Built a docker container locally using `make build pyro_branch=support-outside-contrib pyro_git_url=https://github.com/mattwescott/pyro.git` and the container contained the target repo and branch

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -9,6 +9,7 @@ ARG pytorch_branch=release
 ARG uid=1000
 ARG gid=1000
 ARG ostype=Linux
+ARG pyro_git_url=https://github.com/pyro-ppl/pyro.git
 
 # Configurable settings
 ENV USER_NAME pyromancer
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -11,6 +11,7 @@ GID=$(shell id -g)
 OSTYPE=$(shell uname)
 USER=pyromancer
 DOCKER_WORK_DIR=/home/${USER}/workspace/shared
+pyro_git_url=https://github.com/pyro-ppl/pyro.git
 
 # Optional args
 python_version?=2.7
@@ -54,6 +55,7 @@ build: ##
 	--build-arg ostype=${OSTYPE} \
 	--build-arg python_version=${python_version} \
 	--build-arg pytorch_branch=${pytorch_branch} \
+	--build-arg pyro_git_url=${pyro_git_url} \
 	--build-arg pyro_branch=${pyro_branch} -f ${DOCKER_FILE} .
 
 build-gpu: ##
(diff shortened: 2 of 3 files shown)


fritzo:
Thanks for these improvements!

I'd prefer keeping the `Makefile` using `-vx` rather than `-v`. The reason we use `-vx` is that some tests take a very long time to run and pytest doesn't print error information until all tests complete. Note you can always manually run `pytest -v` if you want to collect all failing tests (I do this in the less-frequent circumstance of collecting failing tests). Does anyone else have a preference?

I'll defer to @neerajprad for the docker changes.

docker/Dockerfile
```diff
@@ -9,6 +9,7 @@ ARG pytorch_branch=release
 ARG uid=1000
 ARG gid=1000
 ARG ostype=Linux
+ARG pyro_git_url=https://github.com/uber/pyro.git
```
neerajprad: Can we change this to the updated url - https://github.com/pyro-ppl/pyro.git?

Makefile
```diff
@@ -36,26 +36,26 @@ profile: FORCE
...
+	pytest -v -n auto --stage unit
 
 test-examples: lint FORCE
-	pytest -vx -n auto --stage test_examples
+	pytest -v -n auto --stage test_examples
 
 test-tutorials: lint FORCE
 	grep -l smoke_test tutorial/source/*.ipynb | xargs grep -L 'smoke_test = False' \
-		| CI=1 xargs pytest -vx --nbval-lax --current-env
+		| CI=1 xargs pytest -v --nbval-lax --current-env
```
neerajprad: As @fritzo mentioned, all of these test stages are quite expensive with a fair amount of generated log, which motivates the current default to stop at the first failure. We have documented some of the ways to run tests in our [contributing](https://github.com/pyro-ppl/pyro/blob/dev/CONTRIBUTING.md) guide. Please feel free to add to that instead.

mattwescott:
Thanks for the comments, all makes sense. I've updated the PR. 

For future reference, do you prefer cleaning up branch history via rebase and force-push, or adding reverting commits?

fritzo:
> cleaning up branch history via rebase and force-push, or adding reverting commits?

I prefer to avoid force-push. Note that we always squash-and-merge PRs, so history is not maintained within a single PR. This policy is important to ensure we can git bisect and guarantee that all commits have passed CI.

fritzo:
@neerajprad could you review the docker changes?

PR #1880, opened 2019-05-23
Fix torchvision dependency in .travis to enable tests

neerajprad:
Currently pip installing torchvision results in an ImportError:

```
ImportError: libcudart.so.9.0: cannot open shared object file: No such file or directory
```

, which is fixed by https://github.com/pytorch/vision/issues/946 by providing separate CPU only wheels.


neerajprad:
Other PRs are blocked until this is merged.

Issue #1824, opened 2019-04-16
[FR] Contrib module for analyzing tabular data


fritzo:
This issue proposes a pyro.contrib.tabular module for analyzing heterogeneous tabular data with missing cells.

The particular latent variable model I'm interested in is [TreeCat](https://github.com/posterior/treecat), which supports pluggable models for different feature types (e.g. real, boolean, categorical, count, date, lat_long). However, once these feature models are implemented, it should be easy to implement additional latent variable models including [CrossCat](http://jmlr.org/papers/v17/11-392.html), various [latent tree models](https://arxiv.org/abs/1402.0577), [TreeGauss](https://arxiv.org/abs/1901.08770) (using the [Funsor](https://github.com/pyro-ppl/funsor) backend), and various models using normalizing flows. Ideally we can set up enough models to compare performance across a variety of datasets.

## Tasks
- [x] #1821 Add a `SpanningTree` distribution.
- [ ] #1370 Implement a `TreeCat` model and inference.
- [ ] #1370 Add example: [Boston housing](http://edwardlib.org/api/observations/boston_housing) (506 rows x 14 cols, real+binary)
- [ ] Add example: [US census 1990](https://archive.ics.uci.edu/ml/datasets/US+Census+Data+%281990%29) (2.5M rows x 68 cols, categorical)
- [ ] Support masked observe statement #1106
- [ ] Fix `infer_discrete` log_prob #1722 so `TreeCat` can be used as a guide

PR #1883, opened 2019-05-23
Support combination of auto and manual init in autoguides

fritzo:
Addresses #1824 

This makes #1849 more useful by allowing users to interleave manually-initialized autoguide params (via `pyro.param("auto_...", ...)`) and automatic initialization strategies (e.g. `init_from_median`). This is often useful when manually initializing top-level parameters followed by automatically initializing lower level parameters based on top-level parameters.

Before this PR, manual param initialization would be ignored when initializing subsequent parameters via strategies. After this PR, manually initialized params will override strategies.

## Tested
- manually confirmed behavior in the contrib-treecat branch

### DIFF
--- a/pyro/contrib/autoguide/__init__.py
+++ b/pyro/contrib/autoguide/__init__.py
@@ -265,7 +265,7 @@ class AutoDelta(AutoGuide):
         guide = AutoDelta(model)
         svi = SVI(model, guide, ...)
 
-    By default latent variables are randomly initialized by the model. To
+    By default latent variables are initialized using ``init_loc_fn()``. To
     change this default behavior the user should call :func:`pyro.param` before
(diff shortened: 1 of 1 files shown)


pyro/contrib/autoguide/__init__.py
```diff
@@ -265,7 +265,7 @@ class AutoDelta(AutoGuide):
         guide = AutoDelta(model)
         svi = SVI(model, guide, ...)
 
-    By default latent variables are randomly initialized by the model. To
+    By default latent variables are initialized using ``init_loc_fn()``. To
     change this default behavior the user should call :func:`pyro.param` before
     beginning inference, with ``"auto_"`` prefixed to the targetd sample site
```
martinjankowiak: (pre-existing) typo "targetd"

PR #1886, opened 2019-05-27
Clarify `tracegraph_elbo`

mattwescott:
### Description

Refactors several functions in `tracegraph_elbo`, adds comments on how the implementation relates to [the paper](https://arxiv.org/abs/1506.05254), and possibly fixes a small bug by including a baseline loss in the jit version.

Two of the changes are substantial enough to highlight:

#### Downstream cost accumulation

The downstream costs are now accumulated in a single pass instead of including a second stage to add missing nodes from the model. This may be less efficient for some models/guides where the new order of accumulation results in more collisions. 

From here it may be easier to modify the accumulation. If a particular subset of nodes are a source of frequent collisions, their costs could be accumulated separately. Or the cost tree could be walked in the case of a collision instead of jumping to the leaves. 

#### Sharing code with jit version

The surrogate loss computation is now shared between `TraceGraph_ELBO` and `JitTraceGraph_ELBO`. I may have neglected some subtleties of jit in torch. 

### Testing

* `make test` passes
* have not assessed test coverage
* have not done any profiling

### DIFF
--- a/pyro/infer/tracegraph_elbo.py
+++ b/pyro/infer/tracegraph_elbo.py
@@ -122,50 +122,69 @@ def _compute_elbo_reparam(model_trace, guide_trace):
     return elbo, surrogate_elbo
 
 
+def _construct_baseline(node, guide_site, downstream_cost):
+
+    # XXX should the average baseline be in the param store as below?
+
+    baseline = 0.0
+    baseline_loss = 0.0
+
+    (nn_baseline, nn_baseline_input, use_decaying_avg_baseline, baseline_beta,
+        baseline_value) = _get_baseline_options(guide_site)
+
+    use_nn_baseline = nn_baseline is not None
+    use_baseline_value = baseline_value is not None
+
+    use_baseline = use_nn_baseline or use_decaying_avg_baseline or use_baseline_value
+
+    assert(not (use_nn_baseline and use_baseline_value)), \
+        "cannot use baseline_value and nn_baseline simultaneously"
+    if use_decaying_avg_baseline:
+        dc_shape = downstream_cost.shape
+        param_name = "__baseline_avg_downstream_cost_" + node
+        with torch.no_grad():
+            avg_downstream_cost_old = pyro.param(param_name,
+                                                 torch.zeros(dc_shape, device=guide_site['value'].device))
+            avg_downstream_cost_new = (1 - baseline_beta) * downstream_cost + \
+                baseline_beta * avg_downstream_cost_old
+        pyro.get_param_store()[param_name] = avg_downstream_cost_new
+        baseline += avg_downstream_cost_old
+    if use_nn_baseline:
+        # block nn_baseline_input gradients except in baseline loss
+        baseline += nn_baseline(detach_iterable(nn_baseline_input))
+    elif use_baseline_value:
+        # it's on the user to make sure baseline_value tape only points to baseline params
+        baseline += baseline_value
+    if use_nn_baseline or use_baseline_value:
+        # accumulate baseline loss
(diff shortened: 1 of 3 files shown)


pyro/infer/tracegraph_elbo.py
```diff
@@ -270,43 +355,24 @@ class JitTraceGraph_ELBO(TraceGraph_ELBO):
...
         kwargs['_pyro_guide_id'] = id(guide)
-        if getattr(self, '_loss_and_surrogate_loss', None) is None:
+        if getattr(self, '_jit_lsl', None) is None:
             # build a closure for loss_and_surrogate_loss
             weakself = weakref.ref(self)
 
             @pyro.ops.jit.trace(ignore_warnings=self.ignore_jit_warnings,
                                 jit_options=self.jit_options)
-            def loss_and_surrogate_loss(*args, **kwargs):
+            def jit_lsl(*args, **kwargs):
```
fritzo: nit: it's easier to read `jit_loss_and_surrogate_loss`

pyro/infer/tracegraph_elbo.py
```diff
@@ -31,135 +31,208 @@ def _get_baseline_options(site):
...
+    # motivates a recursive computation. Note, while only the downstream costs
+    # for nonreparameterized guide nodes will ultimately be used, the reparam
+    # node costs count toward those downstream totals.
 
-    downstream_guide_cost_nodes = {}
     downstream_costs = {}
+
+    # While accumulating costs, track which terms have been included for each
+    # node to support detecting and handling redundant paths.
+    included_model_terms = defaultdict(lambda: set())
```
fritzo: nit: You can simply write `defaultdict(set)` here and elsewhere in this PR

pyro/infer/tracegraph_elbo.py
```diff
@@ -31,135 +31,208 @@ def _get_baseline_options(site):
...
-    # 1. downstream costs used for rao-blackwellization
-    # 2. model observe sites (as well as terms that arise from the model and guide having different
-    # dependency structures) are taken care of via 'children_in_model' below
-    topo_sort_guide_nodes = guide_trace.topological_sort(reverse=True)
-    topo_sort_guide_nodes = [x for x in topo_sort_guide_nodes
-                             if guide_trace.nodes[x]["type"] == "sample"]
-    ordered_guide_nodes_dict = {n: i for i, n in enumerate(topo_sort_guide_nodes)}
+def _construct_baseline(node, guide_site, downstream_cost):
+
+    # XXX should the average baseline be in the param store as below?
```
fritzo: @martinjankowiak can you comment?

mattwescott:
Thanks for the feedback @fritzo, I've added those changes. 

If it would be easier on your end, I can split this into three PRs. Maybe bugfix, refactor, and one-pass cost accumulation?

martinjankowiak:
@mattwescott splitting the PR would be very helpful.

can you explain your rationale for one-pass cost accumulation? is it just perceived simplicity? as you say, this may be slower for some (common) model/guide patterns--even if this overhead may not be a huge deal in most cases of practical interest..

mattwescott:
> @mattwescott splitting the PR would be very helpful.

Alright, I'll close this and add smaller PRs sequentially.

> can you explain your rationale for one-pass cost accumulation? is it just perceived simplicity? as you say, this may be slower for some (common) model/guide patterns--even if this overhead may not be a huge deal in most cases of practical interest.. 

Both perceived simplicity and as a first step toward other modifications, namely walking the cost tree instead of jumping to the leaves, and optional separate accumulation for high-conflict subsets of nodes. I'll plan on holding off on the one-pass PR for now.

PR #1887, opened 2019-05-27
Sum of squares

stefanwebb:
[Sum-of-squares polynomial flow](https://arxiv.org/abs/1905.02325)


stefanwebb:
This is ready to be looked at

pyro/distributions/polynomial.py
```diff
@@ -0,0 +1,139 @@
...
+    Priyank Jaini, Kira A. Shelby, Yaoliang Yu
+    """
+
+    domain = constraints.real
+    codomain = constraints.real
+    bijective = True
+    event_dim = 1
+    autoregressive = True
+
+    def __init__(self, autoregressive_nn, input_dim, count_degree, count_sum):
```
martinjankowiak: @stefanwebb could you please clarify how `count_degree` etc. correspond to the notation in the paper (r, k, etc.)? this will help me review the polynomial construction

stefanwebb: `r = count_degree`
`k = count_sum`

Another important detail: how the author's proposed to calculate the forward operation/gradient wasn't made explicit in the paper. I chose to expand the squared term in Eq (8) and integrate each power of `u`, which I think is the most numerically stable.

stefanwebb: And the Fundamental Theorem of Calculus for the Jacobian of course

pyro/distributions/polynomial.py
```diff
@@ -0,0 +1,139 @@
...
+        self.c = nn.Parameter(torch.Tensor(input_dim))
+        self.reset_parameters()
+
+        # Vector of powers of input dimension
+        powers = torch.arange(1, count_degree + 2, dtype=torch.get_default_dtype())
+        self.register_buffer('powers', powers)
+
+        # Build mask of constants
+        mask = torch.zeros(count_degree + 1, count_degree + 1)
+        for jdx in range(count_degree + 1):
```
martinjankowiak: can't this be replaced by something like
`mask = powers + torch.arange(count_degree + 1).unsqueeze(-1).type_as(powers)`

pyro/distributions/polynomial.py
```diff
@@ -0,0 +1,139 @@
...
+        # Take cross product of coefficients across degree dim
+        # ~ (batch_size, count_sum, count_degree+1, count_degree+1, input_dim)
+        coefs = A.unsqueeze(-2) * A.unsqueeze(-3)
+
+        # Calculate output as sum-of-squares polynomial
+        x_view = x.view(-1, 1, 1, self.input_dim)
+        x_pow_matrix = x_view.pow(self.power_mask.unsqueeze(-1)).unsqueeze(-4)
+
+        # NOTE: The view_as is necessary because the batch dimensions were collapsed previously
+        y = self.c + (coefs * x_pow_matrix * self.mask.unsqueeze(-1)).sum((1, 2, 3)).view_as(x)
```
martinjankowiak: btw is there any way to make this more performant? this might be a bit slow since you're realizing these large 4-dimensional tensors before summing out 3 of the dimensions. e.g. can it be done more greedily or with matrix multiplies or einsums?

pyro/distributions/polynomial.py
stefanwebb: Done!

pyro/distributions/polynomial.py
stefanwebb: I don't think it's that slow relative to the alternatives as they're all vectorized operations. This wouldn't be a good use of time to look into, the more important question is how can you actually produce a flexible and stable normalizing flow?

stefanwebb:
Made the requested changes, this should be good to go after #1908

stefanwebb:
@martinjankowiak okay, I've moved to `transforms` so this is ready to be looked at again!

PR #1888, opened 2019-05-28
Add an EasyGuide module, more flexible than AutoGuide

fritzo:
This adds a `pyro.contrib.easyguide` module that aims to be simpler and more flexible than our original `pyro.contrib.autoguide`. The main differences are:
1. `EasyGuide` supports data subsampling, relying on param subsampling #1796.
2. `EasyGuide` is a framework in which users write a `guide` method using standard Pyro syntax. Whereas `AutoGuide`s are composed using `AutoGuideList`, `EasyGuide` achieves composition by creating multiple `group.sample()` statements in the `guide` method.
3. `EasyGuide` exposes machinery to make it simpler to define amortized guides, specifically machinery to amortize multiple sample sites from a single nn and to batch those nn evaluations over a plate.

One major design decision was to avoid `AutoGuide`'s pattern of hiding `pyro.param` statements. In practice I've found initialization to be crucial for convergence in SVI. By exposing `pyro.param` statements, we force the user to create initial param values.

## Tested
- added unit tests
- added a tutorial


pyro/contrib/easyguide/easyguide.py
```diff
@@ -0,0 +1,267 @@
...
+        param_name = "auto_{}".format(name)
+        init_value = None
+        if param_name not in pyro.get_param_store():
+            init_value = site["value"].detach()
+        with ExitStack() as stack:
+            for frame in site["cond_indep_stack"]:
+                plate = self.plate(frame.name)
+                if plate not in runtime._PYRO_STACK:
+                    stack.enter_context(plate)
+                elif init_value is not None and plate.subsample_size < plate.size:
```
martinjankowiak: `elif init_value is not None and plate.subsample_size < plate.size` when does this happen?

pyro/contrib/easyguide/easyguide.py
```diff
@@ -0,0 +1,267 @@
...
+            value = pyro.param(param_name, init_value,
+                               constraint=fn.support, event_dim=fn.event_dim)
+            return pyro.sample(name, dist.Delta(value, event_dim=fn.event_dim))
+
+
+class Group(object):
+    """
+    An autoguide helper to match a group of model sites.
+
+    :ivar list sites: A list of all matching sample sites in the model.
```
martinjankowiak: docstring needs fixing?

pyro/contrib/easyguide/easyguide.py
```diff
@@ -0,0 +1,267 @@
...
+    :param EasyGuide guide: An easyguide instance.
+    :param list sites: A list of model sites.
+    """
+    def __init__(self, guide, sites):
+        assert isinstance(sites, list)
+        assert sites
+        self._guide = weakref.ref(guide)
+        self.sites = sites
+
+        # A group is in a frame only if all its sample sites are in that frame.
```
martinjankowiak: what happens in 176ff if a group of sites crosses a frame boundary?

tutorial/source/easyguide.ipynb
```diff
@@ -0,0 +1,325 @@
...
+   "metadata": {},
+   "source": [
+    "## Writing a guide without EasyGuide"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Consider a possible guide for this model where we point-estimate the `drift` parameter using a `Delta` distribution, and then model local time series using shared uncertainty but local means, using a `LowRankMultivariateNormal` distribution. There is a single global sample site which we can model with a `param` and `sample` statement. Then we sample a global pair of uncertainty parameters `cov_diag` and `cov_factor`. Next we sample a local `loc` parameter using `pyro.param(..., event_dim=...)` and an auxiliary sample site. Finally we unpack that site into one element per time series."
```
martinjankowiak: add a parenthetic remark about what "an auxiliary sample site" means?

pyro/contrib/easyguide/easyguide.py
fritzo: Looks correct to me.

pyro/contrib/easyguide/easyguide.py
fritzo: - `init_value not None` happens on the first guide call.
- `plate.subsample_size < plate.size` happens whenever the plate is subsampled.
If both happen, then we need to adjust the `init_value` from subsampled shape to full shape; `pyro.plate` will then subsample back down.

pyro/contrib/easyguide/easyguide.py
fritzo: Then `self.frames` will be set to only those frames populated by all `sites`. If a frame is occupied by a strict subset of `sites` (and hence not in `self.frames`) then than frame cannot be subsampled (if it is subsampled, an error will be raised).

martinjankowiak: where exactly does that error get triggered?

fritzo: Hmm good point. I've added an informative error message, as well as two new tests:
1. test that it's ok to mix shared and non-shared plates if all shared plates are left of non-shared plates.
2. test that the error is triggered if a shared plate is right of a non-shared plate.

martinjankowiak: thanks for the additional tests! i'm still a bit confused about this. would like to discuss offline tomorrow before merging

PR #1890, opened 2019-05-29
Block Neural Autoregressive Flow

stefanwebb:
[https://arxiv.org/abs/1904.04676](https://arxiv.org/abs/1904.04676)


pyro/distributions/block_naf.py
```diff
@@ -0,0 +1,229 @@
+# This implementation uses some code from https://github.com/nicola-decao/BNAF under the MIT license.
```
martinjankowiak: can you please clarify what this means? we are only able to merge code with certain sorts of licenses

stefanwebb: I used the details from a few lines of equations from the author's implementation but not any whole blocks of code verbatim: [https://github.com/nicola-decao/BNAF/blob/master/bnaf.py](https://github.com/nicola-decao/BNAF/blob/master/bnaf.py)

fritzo: If you aren't pasting code verbatim, could you change the comment to

> This implementation is adapted from ...

stefanwebb: Done!

stefanwebb:
I've made the requested change above. I discovered a bug too, where an in-place operation stopped gradients on the flow parameters.

This should be good to go after #1908, or alternative I can move this one into `dist.transforms` later

stefanwebb:
I'm refactoring this right now and fixing a few bugs, so it's not quite ready yet

stefanwebb:
@martinjankowiak This is ready to be looked at! I'd like to apply the same refactoring the the NAF classes that applies a different activation function depending on a initializer argument. I think it was the wrong design decision to have three separate classes for these (`DeepELUFlow`, `DeepSigmoidalFlow`, `DeepLeakyReLUFlow`)

pyro/distributions/transforms/block_naf.py
```diff
@@ -0,0 +1,234 @@
...
+    def get_weights(self):
+        """
+        Computes the weight matrix using masks and weight normalization.
+        It also compute the log diagonal blocks of it.
+        """
+
+        # Form block weight matrix, making sure it's positive on diagonal!
+        w = torch.exp(self._weight) * self.mask_d + self._weight * self.mask_o
+
+        # NOTE: Commented out weight normalization for now!
```
martinjankowiak: is this comment still correct?

pyro/distributions/transforms/block_naf.py
```diff
@@ -0,0 +1,234 @@
...
+        # Sum is taken over columns, i.e. one norm per row
+        w_squared_norm = (w ** 2).sum(-1, keepdim=True)
+
+        # Effect of multiplication and division is that each row is normalized and rescaled
+        w = self._diag_weight.exp() * w / (w_squared_norm.sqrt() + eps)
+
+        # Taking the effect of weight normalization into account in calculating the log-gradient is straightforward!
+        # Instead of differenting, e.g. d(W_1x)/dx, we have d(g_1W_1/(W_1^TW_1)^0.5x)/dx, roughly speaking, and taking
```
martinjankowiak: typo 'differenting'

tests/distributions/test_flows.py
```diff
@@ -125,11 +129,22 @@ def test_batchnorm_jacobians(self):
         for input_dim in [2, 3, 5, 7, 9, 11]:
             self._test_jacobian(input_dim, self._make_batchnorm)
 
+    def test_blocknaf_jacobians(self):
+        for activation in ['ELU', 'LeakyReLU', 'sigmoid', 'tanh']:
+            for residual in [None, 'normal', 'gated']:
+                for input_dim in [2, 3, 5, 7, 9, 11]:
```
martinjankowiak: do we need quite so many `input_dim`s...? unfortuantely, we always have somewhat limited travis cores

more generally, we probably don't need every last cross product

pyro/distributions/transforms/block_naf.py
```diff
@@ -0,0 +1,234 @@
...
+            raise ValueError('Hidden factors, {}, must all be >= 1'.format(hidden_factors))
+
+        if residual not in [None, 'normal', 'gated']:
+            raise ValueError('Invalid value {} for keyword argument "residual"'.format(residual))
+
+        # Mix in activation function methods
+        name_to_mixin = {'ELU': ELUMixin, 'LeakyReLU': LeakyReLUMixin, 'sigmoid': SigmoidalMixin, 'tanh': TanhMixin}
+        if activation not in name_to_mixin:
+            raise ValueError('Invalid activation function "{}"'.format(activation))
+        self.f = types.MethodType(name_to_mixin[activation].f, self)
```
martinjankowiak: @neerajprad can you please take a look at use of `type` and mixins here?

pyro/distributions/transforms/block_naf.py
stefanwebb: Removed!

pyro/distributions/transforms/block_naf.py
stefanwebb: Fixed!

tests/distributions/test_flows.py
stefanwebb: Fixed!

pyro/distributions/transforms/block_naf.py
stefanwebb: Fixed!

tests/distributions/test_flows.py
stefanwebb: I've reduced the number of tests

pyro/distributions/transforms/block_naf.py
stefanwebb: Removed

pyro/distributions/transforms/block_naf.py
stefanwebb: I've removed the use of `types.MethodType` in favour of static methods as I couldn't get it to work on Python 2.7

stefanwebb:
@martinjankowiak Ready for another look!

pyro/distributions/transforms/block_naf.py
```diff
@@ -0,0 +1,232 @@
...
+    :type residual: string
+
+    References:
+
+    Block Neural Autoregressive Flow [arXiv:1904.04676]
+    Nicola De Cao, Ivan Titov, Wilker Aziz
+    """
+    domain = constraints.real
+    codomain = constraints.real
+    bijective = True
```
neerajprad: Should this be false given that `inverse` isn't implemented?

pyro/distributions/transforms/block_naf.py
```diff
@@ -0,0 +1,232 @@
...
+        """
+        :param y: the output of the bijection
+        :type y: torch.Tensor
+
+        Inverts y => x. As noted above, this implementation is incapable of inverting arbitrary values
+        `y`; rather it assumes `y` is the result of a previously computed application of the bijector
+        to some `x` (which was cached on the forward call)
+        """
+
+        raise KeyError("BlockNAFFlow expected to find key in intermediates cache but didn't")
```
neerajprad: Is this intended? It seems like this will always throw a KeyError without even looking into the previously cached values.

pyro/distributions/transforms/block_naf.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def log_abs_det_jacobian(self, x, y):
```
neerajprad: This seems a bit misleading since this always returns the last cached log det term regardless of x and y. Consider asserting that x and y are the same as in the cache, and throwing an exception otherwise.

pyro/distributions/transforms/naf.py
```diff
@@ -13,12 +13,123 @@
 eps = 1e-8
 
 
+class ELUMixin(object):
```
neerajprad: nit: Instead of using a constant eps, would suggest adapting eps depending on tensor dtype as follows:

```python
def _eps(x):
    return torch.finfo(x.dtype).tiny
```

pyro/distributions/transforms/naf.py
```diff
@@ -13,12 +13,123 @@
...
+        """
+        return torch.sigmoid(x) * (1. - eps) + 0.5 * eps
+
+    @staticmethod
+    def f_inv(x):
+        """
+        Implements the inverse scaled sigmoid nonlinearity
+        """
+        y = (x - 0.5 * eps) / (1. - eps)
+        return safe_log(y) - safe_log(1. - y)
```
neerajprad: Instead of `safe_log(1. - y)`, could you instead use `log1p(-y)`?

pyro/distributions/transforms/naf.py
```diff
@@ -13,12 +13,123 @@
...
+
+    @staticmethod
+    def log_df_inv_dx(x):
+        """
+        Implements the log derivative of inverse NAF nonlinearity
+        """
+        return torch.where(x >= 0., torch.zeros_like(x), torch.ones_like(x) * math.log(100.0))
+
+
+def safe_log(x):
```
neerajprad: Just curious - in which cases do you find `safe_log` to be useful?

pyro/distributions/transforms/block_naf.py
stefanwebb: The transform is bijective even though an analytic inverse doesn't exist. I think this should be true based on the comment: https://github.com/pytorch/pytorch/blob/master/torch/distributions/transforms.py#L63

pyro/distributions/transforms/block_naf.py
stefanwebb: Yes! The caching happens in the parent class `torch.distributions.transforms.Transform`

pyro/distributions/transforms/block_naf.py
stefanwebb: I agree, although this function should only be called from `torch.distributions.transform.TransformedDistribution`. I'll add the check!

pyro/distributions/transforms/naf.py
stefanwebb: `.tiny` is the machine epsilon for a type, which I think may be too small to prevent numerical instabilities (although I'm not entirely certain). What do you think?

pyro/distributions/transforms/naf.py
stefanwebb: `safe_log` is a way to get extra precision at the lower end of the input. I copied it from the author's implementation, but I'm not sure whether it's entirely necessary. `log1p(-y)` would work instead of `safe_log(1. - y)` and I'll change that, but I don't think `log1p(y-1)` would adequately replace `safe_log(y)`. 

It's perhaps better to keep it in just in case?

pyro/distributions/transforms/naf.py
stefanwebb: I think the intention for for very small `x`

pyro/distributions/transforms/block_naf.py
neerajprad: Got it, thanks for explaining.

pyro/distributions/transforms/naf.py
neerajprad: > log(x) where x is almost 0

I think log(x) should have good precision when x is quite small without resorting to this, and log1p can be used for values close to 1. We are using this function at quite a few places, so I'll leave it to you to decide whether to keep this! :)

pyro/distributions/transforms/naf.py
neerajprad: We have found it to work quite well e.g. for SigmoidTransform and many other transforms in PyTorch. If you don't need high precision and the have found 1e-8 to work well in practice, you don't have to worry about it though.

pyro/distributions/transforms/block_naf.py
neerajprad: There are other parts of the code that call `log_abs_det_jacobian` directly, e.g. in HMC, though we probably wouldn't use this transform directly.

stefanwebb: I've made this change!

pyro/distributions/transforms/naf.py
stefanwebb: I've replaced it with your suggestion

pyro/distributions/transforms/naf.py
stefanwebb: I've removed `safe_log`

stefanwebb:
@neerajprad @martinjankowiak I've made those changes after the last review

pyro/distributions/transforms/block_naf.py
```diff
@@ -12,7 +12,9 @@
 from pyro.distributions.util import copy_docs_from
 from pyro.distributions.transforms.naf import ELUMixin, LeakyReLUMixin, TanhMixin, SigmoidalMixin
 
-eps = 1e-8
+
+def _eps(x):
```
martinjankowiak: are we confident this is large enough?

pyro/distributions/transforms/block_naf.py
```diff
@@ -152,7 +155,12 @@ def log_abs_det_jacobian(self, x, y):
         """
         Calculates the elementwise determinant of the log jacobian
         """
-        return self._cached_logDetJ.sum(-1)
+        x_old, y_old = self._cached_x_y
```
martinjankowiak: isn't it a bit strange to put this check here but not in other transformed distributions that have caching? either all should be fixed or none? and if all, maybe it can be done in some more uniform way?

neerajprad: I agree, it makes sense to uniformly handle this for all transforms that cache the log det term. I'm guessing all of them are in Pyro since torch transforms don't cache this term. We could add a method to `TransformModule` to handle this instead, if that's the case. Please feel free to defer this to a later PR.

stefanwebb: I'll leave this as is for now

pyro/distributions/transforms/block_naf.py
stefanwebb: Not very, I've rolled back this change

stefanwebb:
I've rolled back that change to `eps` and renamed the class based on what I think is a better API

pyro/distributions/transforms/block_autoregressive.py
```diff
@@ -0,0 +1,237 @@
...
+        to some `x` (which was cached on the forward call)
+        """
+
+        raise KeyError("BlockAutoregressive expected to find key in intermediates cache but didn't")
+
+    def log_abs_det_jacobian(self, x, y):
+        """
+        Calculates the elementwise determinant of the log jacobian
+        """
+        x_old, y_old = self._cached_x_y
```
neerajprad: @stefanwebb - Can we remove this check for now? This seems like a pattern that we are following in other transforms too, and is probably a more general interface issue that can be separately discussed and addressed.

stefanwebb: Okay I've rolled back this change

PR #1891, opened 2019-05-31
Share surrogate loss in `tracegraph_elbo`

### LINKED ISSUE #1894: Graph retention in JitTraceGraph_ELBO
In reviewing #1891, @martinjankowiak  noted that `JitTraceGraph_ELBO` isn't using `retain_graph`. It could make sense to simply switch to `backward(retain_graph=self.retain_graph)`. I'm reluctant to do that without some understanding of how graph retention interacts with jit.

mattwescott:
Shares the surrogate loss method between `TraceGraph_ELBO` and `JitTraceGraph_ELBO`. This may fix an unreported bug as previously the gradients were not computed for the baseline loss in `JitTraceGraph_ELBO`. Also closes #1894.

This is the first of three PRs from #1886

### DIFF
--- a/pyro/infer/tracegraph_elbo.py
+++ b/pyro/infer/tracegraph_elbo.py
@@ -219,32 +219,49 @@ class TraceGraph_ELBO(ELBO):
         Performs backward on the latter. Num_particle many samples are used to form the estimators.
         If baselines are present, a baseline loss is also constructed and differentiated.
         """
-        loss = 0.0
-        weight = 1./self.num_particles
-        for model_trace, guide_trace in self._get_traces(model, guide, *args, **kwargs):
-            loss += self._loss_and_grads_particle(weight, model_trace, guide_trace)
+
+        loss, surrogate_loss = self._loss_and_surrogate_loss(model, guide, *args, **kwargs)
+
+        torch_backward(surrogate_loss)
+
+        loss = torch_item(loss)
+        warn_if_nan(loss, "loss")
         return loss
 
-    def _loss_and_grads_particle(self, weight, model_trace, guide_trace):
+    def _loss_and_surrogate_loss(self, model, guide, *args, **kwargs):
+
+        loss = 0.0
+        surrogate_loss = 0.0
+
+        for model_trace, guide_trace in self._get_traces(model, guide, *args, **kwargs):
+
+            lp, slp = self._loss_and_surrogate_loss_particle(model_trace, guide_trace, *args, **kwargs)
+            loss += lp
+            surrogate_loss += slp
+
+        loss /= self.num_particles
+        surrogate_loss /= self.num_particles
+
+        return loss, surrogate_loss
+
+    def _loss_and_surrogate_loss_particle(self, model_trace, guide_trace, *args, **kwargs):
+
         # compute elbo for reparameterized nodes
         elbo, surrogate_elbo = _compute_elbo_reparam(model_trace, guide_trace)
         baseline_loss = 0.0
 
         # the following computations are only necessary if we have non-reparameterizable nodes
         non_reparam_nodes = set(guide_trace.nonreparam_stochastic_nodes)
-
         if non_reparam_nodes:
             downstream_costs, _ = _compute_downstream_costs(model_trace, guide_trace, non_reparam_nodes)
             surrogate_elbo_term, baseline_loss = _compute_elbo_non_reparam(guide_trace,
-                                                                           non_reparam_nodes, downstream_costs)
-
+                                                                           non_reparam_nodes,
+                                                                           downstream_costs)
             surrogate_elbo += surrogate_elbo_term
 
-        torch_backward(weight * (-surrogate_elbo + baseline_loss))
+        surrogate_loss = -surrogate_elbo + baseline_loss
 
-        loss = torch_item(- weight * elbo)
-        warn_if_nan(loss, "loss")
-        return loss
+        return elbo, surrogate_loss
 
 
 class JitTraceGraph_ELBO(TraceGraph_ELBO):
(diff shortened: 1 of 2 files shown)


pyro/infer/tracegraph_elbo.py
```diff
@@ -219,37 +219,49 @@ def loss_and_grads(self, model, guide, *args, **kwargs):
...
 
-        # collect parameters to train from model and guide
-        trainable_params = any(site["type"] == "param"
-                               for trace in (model_trace, guide_trace)
-                               for site in trace.nodes.values())
-
-        if trainable_params:
-            surrogate_loss = -surrogate_elbo
-            torch_backward(weight * (surrogate_loss + baseline_loss), retain_graph=self.retain_graph)
+        surrogate_loss = -surrogate_elbo + baseline_loss
```
fritzo: Shouldn't `baseline_loss` be scaled by `1 / self.num_particles`?

pyro/infer/tracegraph_elbo.py
```diff
@@ -219,37 +219,49 @@ def loss_and_grads(self, model, guide, *args, **kwargs):
         Performs backward on the latter. Num_particle many samples are used to form the estimators.
         If baselines are present, a baseline loss is also constructed and differentiated.
         """
+
+        loss, surrogate_loss = self._loss_and_surrogate_loss(model, guide, *args, **kwargs)
+
+        torch_backward(surrogate_loss)
```
fritzo: You need to pass in `retain=self.retain_graph`.

pyro/infer/tracegraph_elbo.py
mattwescott: Might be missing something, but I believe the needed scaling is done here: https://github.com/pyro-ppl/pyro/blob/2ca18fdfe3fb7cada2540a3cc5820a39eb43123e/pyro/infer/tracegraph_elbo.py#L243

fritzo: Thanks for explaining!

pyro/infer/tracegraph_elbo.py
```diff
@@ -219,37 +219,49 @@ def loss_and_grads(self, model, guide, *args, **kwargs):
...
-            loss += self._loss_and_grads_particle(weight, model_trace, guide_trace)
-        return loss
 
-    def _loss_and_grads_particle(self, weight, model_trace, guide_trace):
+            lp, slp = self._loss_and_surrogate_loss_particle(model_trace, guide_trace, *args, **kwargs)
+            loss += lp
+            surrogate_loss += slp
+
+        loss /= self.num_particles
+        surrogate_loss /= self.num_particles
```
fritzo: Can you avoid in-place operators, since this is now being used by `JitTraceGraph_ELBO`?

pyro/infer/tracegraph_elbo.py
```diff
@@ -219,37 +219,49 @@ def loss_and_grads(self, model, guide, *args, **kwargs):
...
         # the following computations are only necessary if we have non-reparameterizable nodes
-        baseline_loss = 0.0
+        non_reparam_nodes = set(guide_trace.nonreparam_stochastic_nodes)
         if non_reparam_nodes:
             downstream_costs, _ = _compute_downstream_costs(model_trace, guide_trace, non_reparam_nodes)
             surrogate_elbo_term, baseline_loss = _compute_elbo_non_reparam(guide_trace,
-                                                                           non_reparam_nodes, downstream_costs)
+                                                                           non_reparam_nodes,
+                                                                           downstream_costs)
             surrogate_elbo += surrogate_elbo_term
```
fritzo: Avoid in-place ops.

pyro/infer/tracegraph_elbo.py
```diff
@@ -91,7 +91,7 @@ def _compute_downstream_costs(model_trace, guide_trace,  #
     return downstream_costs, downstream_guide_cost_nodes
 
 
-def _compute_elbo_reparam(model_trace, guide_trace, non_reparam_nodes):
+def _compute_elbo_reparam(model_trace, guide_trace):
```
fritzo: Avoid in-place ops in this function.

mattwescott:
@fritzo Thanks for surfacing these. My understanding from [the torch docs](https://pytorch.org/docs/stable/jit.html#tracing-edge-cases) is that in-place operations on views can be problematic, and otherwise [in-place operation is a shorthand](https://pytorch.org/docs/stable/jit.html#statements). What am I missing?

Also looks like some of these in-place ops were around before my changes. To reduce conflicts, would it make sense to leave at least some of them for after the rest of #1886?

fritzo:
> ... otherwise tracing can handle in-place operations.

OK, do you think these in-place ops are all safe then? I have only vague understanding of torch.jit's current limitations, and my aim with the review comments is to avoid breakage due to the refactoring now using in-place ops. Note we don't have strong test coverage of convergence of jit elbos. If you have a clear understanding of torch.jit's current limitations and you believe the in-place ops are ok, then I'm happy leaving them in this PR.

mattwescott:
I'd be surprised if they aren't equivalent to the in-place ops that were already present. Added #1893 to try to surface best practices.

mattwescott:
On a related note, I'm not sure how baselines interact with jit, and haven't come across any tests. Looks like there are some baseline tests in `integration_tests/test_tracegraph_elbo.py`. But I'm not seeing any for the jit version. 

Wouldn't expect adding the baseline loss to `backward` to introduce any new issues, but I could easily be overlooking something, and in any case I'm open to testing ideas.

pyro/infer/tracegraph_elbo.py
```diff
@@ -270,43 +282,24 @@ class JitTraceGraph_ELBO(TraceGraph_ELBO):
...
-        loss, surrogate_loss = self._loss_and_surrogate_loss(*args, **kwargs)
-        surrogate_loss.backward()
-        loss = loss.item()
+                return self._loss_and_surrogate_loss(model, guide, *args, **kwargs)
+
+            self._jit_loss_and_surrogate_loss = jit_loss_and_surrogate_loss
 
+        loss, surrogate_loss = self._jit_loss_and_surrogate_loss(*args, **kwargs)
+
+        surrogate_loss.backward()  # triggers jit compilation
```
martinjankowiak: does this mean `retain_graph` only works for non-jit mode?

martinjankowiak: (if so should probably update relevant docstring?)

PR #1892, opened 2019-05-31
add easyguide example to sparse gamma def

martinjankowiak:
the easy guide is functionally equivalent (afaict) to the custom guide but performs quite a bit worse. i think this is because analytic KLs are not being used in the former case, whereas they are in the latter. 

@fritzo are there any easy fixes to support this? it seems like no, since the whole point of these `group` constructions is that we're breaking mean-fieldedness (if not actually, in this case)

EDIT: i made the easy guide look more like the autoguide because the previous gamma construction performed rather poorly. interestingly the easy guide performs the best.


examples/sparse_gamma_def.py
```diff
@@ -126,13 +132,48 @@ def sample_ws(name, width):
...
+                pyro.param(param + wz + layer).data.clamp_(min=clip)
+
+
+# Define a custom guide using the EasyGuide class.
+# Unlike the 'auto' guide, this guide supports data subsampling.
+# This is the best performing of the three guides.
+# (This guide is functionally similar to the auto guide, but performs
+# somewhat better. The reason seems to be some combination of: i) the better
+# numerical stability of the softplus; and ii) the custom initialization.
+# Though note that in the easy guide case KL divergences are not computed
```
fritzo: I believe the cause of `TraceMeanField_ELBO`'s missed opportunity to use KL is due to `EasyGuide`'s pattern of using `Delta` distributions to split up a single `Normal` distribution; this is equally a problem for `AutoGuide`. Could you update the comment to reflect this?
> Note that in the easy guide and auto guide cases, KL divergences are not computed analytically in the ELBO because the guide and model sample sites are separated by a Delta distribution for reshaping.

martinjankowiak: done. lmk if not sufficient

Issue #1893, opened 2019-05-31
Evaluate risk from in-place operations in `tracegraph_elbo`


mattwescott:
In discussing #1891 @fritzo raised the possibility of in-place operations being an issue for jit tracing. 

My impression from [the torch docs](https://pytorch.org/docs/stable/jit.html#tracing-edge-cases) is that in-place operations on views are to be avoided (and will result in a warning).

It looks like these tracer warnings are being ignored in most of the tests in `test_jit`. Is that important? If it would be easy to miss introducing a problematic in-place operation on a view, would it reduce risk to avoid in-place operations entirely, even where it is currently safe?

fritzo:
In general all new code in Pyro should avoid in-place ops. I believe `TraceGraph_ELBO` was one of the earlier elbos.

fritzo:
Thanks for opening this @mattwescott . I think I'll close it now, since the PyTorch jit has been increasing support for in-place ops.

PR #1895, opened 2019-06-04
Clarify surrogate elbos in `tracegraph_elbo`

mattwescott:
Adds comments on how `_compute_elbo_reparam` relates to the reference, and factors baseline construction out of `_compute_elbo_nonreparam`.

This is the second of three PRs from #1886

### DIFF
--- a/pyro/infer/tracegraph_elbo.py
+++ b/pyro/infer/tracegraph_elbo.py
@@ -92,20 +92,30 @@ def _compute_downstream_costs(model_trace, guide_trace,  #
 
 
 def _compute_elbo_reparam(model_trace, guide_trace):
+
+    # Compute a surrogate ELBO. In [1], this is simply the ELBO because
+    # (non)reparameterization is captured in the structure of the computation
+    # graph, and the required behavior under parameter differentiation emerges
+    # from that structure. Here the behavior of E_p[log(p)] under parameter
+    # differentiation is instead represented in site["score_parts"].
+
     elbo = 0.0
     surrogate_elbo = 0.0
 
-    # deal with log p(z|...) terms
+    # Bring log p(x, z|...) terms into both the ELBO and the surrogate
     for name, site in model_trace.nodes.items():
         if site["type"] == "sample":
             elbo += site["log_prob_sum"]
             surrogate_elbo += site["log_prob_sum"]
 
-    # deal with log q(z|...) terms
+    # Bring log q(z|...) terms into the ELBO, and effective terms into the
+    # surrogate
     for name, site in guide_trace.nodes.items():
         if site["type"] == "sample":
             elbo -= site["log_prob_sum"]
             entropy_term = site["score_parts"].entropy_term
+            # For fully reparameterized terms, this entropy_term is log q(z|...)
+            # For fully non-reparameterized terms, it is zero
             if not is_identically_zero(entropy_term):
                 surrogate_elbo -= entropy_term.sum()
 
--- a/pyro/infer/tracegraph_elbo.py
+++ b/pyro/infer/tracegraph_elbo.py
@@ -138,11 +138,11 @@ def _compute_downstream_costs(model_trace, guide_trace,  #
 
 def _compute_elbo_reparam(model_trace, guide_trace):
 
-    # Compute a surrogate ELBO. In [1], this is simply the ELBO because
-    # (non)reparameterization is captured in the structure of the computation
-    # graph, and the required behavior under parameter differentiation emerges
-    # from that structure. Here the behavior of E_p[log(p)] under parameter
-    # differentiation is instead represented in site["score_parts"].
+    # In ref [1], section 3.2, the part of the surrogate loss computed here is
+    # \sum{cost}, which in this case is the ELBO. Instead of using the ELBO,
+    # this implementation uses a surrogate ELBO which modifies some entropy
+    # terms depending on the parameterization. This reduces the variance of the
+    # gradient under some conditions.
 
     elbo = 0.0
     surrogate_elbo = 0.0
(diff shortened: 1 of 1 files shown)


pyro/infer/tracegraph_elbo.py
```diff
@@ -116,46 +171,20 @@ def _compute_elbo_non_reparam(guide_trace, non_reparam_nodes, downstream_costs):
...
-                raise ValueError("Expected baseline at site {} to be {} instead got {}".format(
-                    node, downstream_cost.shape, baseline.shape))
-            downstream_cost = downstream_cost - baseline
-        surrogate_elbo += (score_function_term * downstream_cost.detach()).sum()
+        score_function = guide_site["score_parts"].score_function
+
+        use_baseline, baseline_loss_term, baseline = _construct_baseline(node, guide_site, downstream_cost)
+
+        if use_baseline:
+            downstream_cost -= baseline
```
fritzo: Please avoid introducing any new in-place ops.

mattwescott: Fixed. After this batch of PRs, might also swap out the old in-place ops for #1893. Might as well while I'm neck deep in here?

pyro/infer/tracegraph_elbo.py
```diff
@@ -92,20 +137,30 @@ def _compute_downstream_costs(model_trace, guide_trace,  #
 
 
 def _compute_elbo_reparam(model_trace, guide_trace):
+
+    # Compute a surrogate ELBO. In [1], this is simply the ELBO because
```
martinjankowiak: i don't really understand this comment. in ref [1] they also make use of the language of a surrogate loss function (section 3.2)

pyro/infer/tracegraph_elbo.py
```diff
@@ -92,20 +137,30 @@ def _compute_downstream_costs(model_trace, guide_trace,  #
...
 
-    # deal with log p(z|...) terms
+    # Bring log p(x, z|...) terms into both the ELBO and the surrogate
     for name, site in model_trace.nodes.items():
         if site["type"] == "sample":
             elbo += site["log_prob_sum"]
             surrogate_elbo += site["log_prob_sum"]
 
-    # deal with log q(z|...) terms
+    # Bring log q(z|...) terms into the ELBO, and effective terms into the
```
martinjankowiak: what's an "effective term"?

pyro/infer/tracegraph_elbo.py
mattwescott: A more precise start to my comment might be: 

In ref [1], the portion of the surrogate loss handled by this function is `\sum{c}`, which in this context would simply be the ELBO...

pyro/infer/tracegraph_elbo.py
mattwescott: If reparameterizing a stochastic node inserted the transformation into the computation graph, say `z=t(epsilon, theta)`, the resulting cost terms `log q(epsilon|...)` naturally wouldn't contribute to the gradients. Here reparamerization is not explicit in the graph, instead by default it seems to be inferred from the attribute `has_rsample`. Then given that (non)reparameterization, the behavior under differentiation is stored in `score_parts`.

Coming from the paper, `site['score_parts'].entropy_term` felt like an effective term for `log q` given that the parameterization is implicit. 

Would it be useful to make these statements more precise?

pyro/infer/tracegraph_elbo.py
mattwescott: Thanks for engaging with this! Realized that if my image were right, the implementation would be wrong, and that has been a very useful signal to revisit my understanding. Would this be closer to the truth?

    # In ref [1], the part of the surrogate loss computed here is `\sum{c}`, 
    # which in this case is simply the ELBO. Depending on the parameterization, 
    # some of the entropy terms may not contribute (in expectation) to the 
    # gradient estimate. To reduce the variance of the gradient estimates under some
    # conditions, the log q(z|...) terms are instead taken from `site['score_parts']`

mattwescott: @martinjankowiak marking these as (hopefully) resolved to simplify reviewing the new commit

fritzo:
@martinjankowiak can you also approve? I'm happy to merge now.

martinjankowiak:
great, thanks @mattwescott 

re: the 3rd PR you were considering, it'd be great if you outlined your rationale before getting started with any code.

some things that i think are worth keeping in mind:

- this is a complex piece of code, and so it can be a bit onerous to change. consequently i think it's best to leave it unchanged unless there are compelling reasons
- at some point in the new few months (?) PyTorch will have named tensor dimensions, and that will likely require more-or-less significant changes to this part of the codebase 
- at some point we plan to migrate to [funsor](https://github.com/pyro-ppl/funsor/), which will likely [deal](https://github.com/pyro-ppl/funsor/blob/416cff6dff9cec6f5a21d769edf90917e7c66bc0/funsor/torch.py#L283) with score function gradients rather differently

@fritzo thoughts?

mattwescott:
Thanks @martinjankowiak. I'm going to try to bother you with one more commit from #1886, and I'll follow up on these leads and have some discussion before going further in that direction. 

> this is a complex piece of code, and so it can be a bit onerous to change. consequently i think it's best to leave it unchanged unless there are compelling reasons

For now my primary motivation is to understand this better, both from mucking around in the code and from your review. My hope is that the changes warrant the effort to review, even if they wouldn't by any means be a priority for the project.

> at some point in the new few months (?) PyTorch will have named tensor dimensions, and that will likely require more-or-less significant changes to this part of the codebase

Thanks for the heads up, was puzzling over some dimension logic in `TraceEnum_ELBO` just yesterday. 

> at some point we plan to migrate to funsor, which will likely deal with score function gradients rather differently

Very interesting!

PR #1904, opened 2019-06-06
Refactor downstream costs in `tracegraph_elbo`

mattwescott:
Adds abstraction as a step toward model-specific optimizations: single-pass accumulation, finding cost subtrees instead of jumping to the leaves, and/or independent accumulation for high-conflict nodes.

From discussion in #1895, sounds like `tracegraph_elbo` may soon require major surgery? Naturally I'd like to think this refactoring is easier to understand and work with, but can understand if it doesn't make sense to invest in and risk this change if Pyro is heading in another direction. 

This is the last of three PRs from #1886

### DIFF
--- a/pyro/infer/tracegraph_elbo.py
+++ b/pyro/infer/tracegraph_elbo.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 import weakref
-from operator import itemgetter
+from collections import defaultdict
 
 import torch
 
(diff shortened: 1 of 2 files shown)


martinjankowiak:
hi @mattwescott, thanks for this carefully constructed PR. unfortunately, at this time i don't think it makes much sense for us to accept these changes, for the following reasons:

- as previously mentioned, this part of the codebase will likely be changing in the future anyway (because of named dimensions etc.)
- while the test coverage here is pretty good, it's probably not perfect, and so it's hard for me to check for correctness unless i spend hours thinking through exact equivalence of the two algorithms
- it'll be easier for us to maintain this code if it's written in the way that we found natural and we ourselves wrote

i'd be less hesitant about accepting major changes here if, e.g., it was demonstrated that this approach was significantly more performant, or enabled RL applications that were otherwise too slow, or the like

PR #1908, opened 2019-06-09
Reorganization of transforms

stefanwebb:
Moves all normalizing flows into `pyro.distributions.transforms`

Could we please merge this in before #1887 and #1890? I realized I should do this PR before I work on `ConditionalTransform` and `ConditionalTransformedDistribution`, and the amortizing version of the flows. These are super high priority for making a useful NF library!


pyro/distributions/transforms/__init__.py
```diff
@@ -0,0 +1,22 @@
+from .batch_norm import BatchNormTransform
```
neerajprad: nit: Would suggest adding `from __future__ import absolute_import, division, print_function` at the top of file and use absolute imports below `from transforms.householder import HouseholderFlow` for consistency with the rest of the codebase.

stefanwebb: Done!

PR #1910, opened 2019-06-11
Support ReduceLROnPlateau scheduler

jpchen:
This PR:
1) changes the scheduler interface to be more similar to PyTorch's (and changes the order of the calls as per 1.1)
2) supports the remaining schedulers (esp ReduceLROnPlateau [which has been requested](https://forum.pyro.ai/t/properly-using-a-scheduler-with-an-optimizer/730/4))

there are still some outstanding bugs that are being fixed on the pytorch end.

Tested:
- schedulers that have different interfaces
- fixed and variable epochs

### DIFF
--- a/pyro/optim/lr_scheduler.py
+++ b/pyro/optim/lr_scheduler.py
@@ -5,20 +5,22 @@ from pyro.optim.optim import PyroOptim
 
 class PyroLRScheduler(PyroOptim):
     """
-    A wrapper for torch.optim.lr_scheduler objects that adjust learning rates
+    A wrapper for :class:`~torch.optim.lr_scheduler` objects that adjusts learning rates
     for dynamically generated parameters.
 
-    :param optim_constructor: a torch.optim.lr_scheduler
+    :param scheduler_constructor: a :class:`~torch.optim.lr_scheduler`
     :param optim_args: a dictionary of learning arguments for the optimizer or a callable that returns
         such dictionaries. must contain the key 'optimizer' with pytorch optimizer value
 
     Example::
 
         optimizer = torch.optim.SGD
-        pyro_scheduler = pyro.optim.ExponentialLR({'optimizer': optimizer, 'optim_args': {'lr': 0.01}, 'gamma': 0.1})
+        scheduler = pyro.optim.ExponentialLR({'optimizer': optimizer, 'optim_args': {'lr': 0.01}, 'gamma': 0.1})
         svi = SVI(model, guide, pyro_scheduler, loss=TraceGraph_ELBO())
-        svi.step()
-        pyro_scheduler.step(epoch=epoch)
+        for i in range(epochs):
+            for minibatch in DataLoader(dataset, batch_size):
+                svi.step(minibatch)
+            scheduler.step(epoch=epoch)
     """
     def __init__(self, scheduler_constructor, optim_args):
         # pytorch scheduler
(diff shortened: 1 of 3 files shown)


pyro/optim/lr_scheduler.py
```diff
@@ -18,6 +18,7 @@ class PyroLRScheduler(PyroOptim):
         pyro_scheduler = pyro.optim.ExponentialLR({'optimizer': optimizer, 'optim_args': {'lr': 0.01}, 'gamma': 0.1})
         svi = SVI(model, guide, pyro_scheduler, loss=TraceGraph_ELBO())
         svi.step()
+        pyro_scheduler.step(epoch=epoch)
```
martinjankowiak: can we please update this to be a full and complete snippet with comments, something like
```
for epoch in range(num_epochs):
    blah() # this happens once an epoch
    for minibatch in loader:
        svi.step() # this takes single gradient step and generally happens many times per epoch
```

martinjankowiak: also please double-check the docstring for correctness/completeness in light of changes

pyro/optim/optim.py
```diff
@@ -51,12 +51,12 @@ def __call__(self, params,  *args, **kwargs):
                     state = self._state_waiting_to_be_consumed.pop(param_name)
                     self.optim_objs[p].load_state_dict(state)
 
-            # actually perform the step for the optim object
-            self.optim_objs[p].step(*args, **kwargs)
-
-            # if optim object was a scheduler, perform an actual optim step
-            if isinstance(self.optim_objs[p], torch.optim.lr_scheduler._LRScheduler):
+            if isinstance(self.optim_objs[p], torch.optim.lr_scheduler._LRScheduler) or \
```
martinjankowiak: why is this branch necessary

jpchen: we only want to step the optimizer, not the scheduler here. this branch is if `self.optim_objs[p]` is a scheduler.

pyro/optim/lr_scheduler.py
```diff
@@ -5,20 +5,22 @@
...
 
     Example::
 
         optimizer = torch.optim.SGD
-        pyro_scheduler = pyro.optim.ExponentialLR({'optimizer': optimizer, 'optim_args': {'lr': 0.01}, 'gamma': 0.1})
+        scheduler = pyro.optim.ExponentialLR({'optimizer': optimizer, 'optim_args': {'lr': 0.01}, 'gamma': 0.1})
         svi = SVI(model, guide, pyro_scheduler, loss=TraceGraph_ELBO())
-        svi.step()
-        pyro_scheduler.step(epoch=epoch)
+        for i in range(epochs):
```
martinjankowiak: i => epoch?

tests/optim/test_optim.py
```diff
@@ -97,20 +99,32 @@ def guide():
 
     svi = SVI(model, guide, scheduler, loss=TraceGraph_ELBO())
     for epoch in range(4):
+        svi.step()
         svi.step()
         loc = pyro.param('loc').unconstrained()
         opt_loc = scheduler.optim_objs[loc].optimizer
         opt_scale = scheduler.optim_objs[loc].optimizer
+        if scheduler.pt_scheduler_constructor is torch.optim.lr_scheduler.ReduceLROnPlateau:
```
fritzo: `issubclass`

pyro/optim/optim.py
fritzo: smells bad to me.

jpchen: i'm not sure how to make it cleaner given the design choice. we just want to step the optimizers here.  `optim_objs` is a dict of optimizers _or_ a dict of schedulers each of which is wrapping an optimizer.  in the `lr_scheduler` initialization, we could store the schedulers and optimizers separately, which would simplify this logic but essentially have two references to the optimizer.  WDYT?

fritzo: ok, I guess we've inherited this bad smell from PyTorch 😞

PR #1913, opened 2019-06-18
Refactoring of MCMC interface to unify with NumPyro

### LINKED ISSUE #1787: OOM error with NUTS on CUDA

neerajprad:
Fixes #1787. 

This refactors the MCMC API to remove dependency on TracePosterior, and thereby not expose the trace data-structure in the API. This is motivated by the following reasons:
 - Integration with other PyTorch libraries.
 - bringing the interface closer to NumPyro to make it easier to write backend agnostic code.
 - lowers memory requirement of running MCMC specially on the GPU. Traces require up to 5x more memory due to all the auxiliary (repeated) information like sample args/kwargs, function return values etc. 

**TODOs:**
 - [x] Add deprecation warning to old API. ~Additionally, deprecate passing in model directly to the HMC/NUTS kernel with the old API.~ (Update - this can be separately discussed, as it will add a bit of boiler plate code in terms of calling `initialize_model`).
 - [x] Make unit tests pass.
 - [ ] ~Remove setting event in multiprocessing code; test that it works on CUDA.~ Will handle this in a separate PR with changes for the baseball example.

**Future work:**
 - [ ] Changes to tutorials and examples.
 - [ ] Add functions for diagnostics and predictions to remove dependency on TracePredictive.


pyro/infer/mcmc/api.py
```diff
@@ -0,0 +1,322 @@
...
+                break
+            metadata, msg = record.getMessage().split("]", 1)
+            _, msg_type, logger_id = metadata[1:].split()
+            if msg_type == DIAGNOSTIC_MSG:
+                pbar_pos = int(logger_id.split(":")[-1])
+                num_samples[pbar_pos] += 1
+                if num_samples[pbar_pos] == warmup_steps:
+                    progress_bars.set_description("Sample [{}]".format(pbar_pos + 1), pos=pbar_pos)
+                diagnostics = json.loads(msg, object_pairs_hook=OrderedDict)
+                progress_bars.set_postfix(diagnostics, pos=pbar_pos)
```
fehiepsi: Maybe we can add the arg `refresh=False` to avoid the flakiness in the speed when using tqdm as in [forum](https://forum.pyro.ai/t/logistic-regression-nuts-on-titanic-dataset-extremely-slow-but-the-time-it-takes-also-wildly-varies/1061/13) (I observed it too).

pyro/infer/mcmc/api.py
```diff
@@ -0,0 +1,322 @@
...
+    def __init__(self, kernel, num_samples, warmup_steps=None, initial_params=None,
+                 num_chains=1, hook_fn=None, mp_context=None, disable_progbar=False):
+        self.warmup_steps = num_samples if warmup_steps is None else warmup_steps  # Stan
+        self.num_samples = num_samples
+        self.kernel = kernel
+        if num_chains > 1:
+            # old API does not work since we need reference to transforms
+            # in the main process.
+            if kernel.transforms is None:
+                raise ValueError("`pyro.infer.mcmc.api` needs explicit reference"
```
fehiepsi: nit: `pyro.infer.mcmc.api` -> `pyro.infer.mcmc.api.MCMC`

pyro/infer/mcmc/api.py
```diff
@@ -0,0 +1,322 @@
...
+    :param bool disable_progbar: Disable progress bar and diagnostics update.
+    """
+    def __init__(self, kernel, num_samples, warmup_steps=None, initial_params=None,
+                 num_chains=1, hook_fn=None, mp_context=None, disable_progbar=False):
+        self.warmup_steps = num_samples if warmup_steps is None else warmup_steps  # Stan
+        self.num_samples = num_samples
+        self.kernel = kernel
+        if num_chains > 1:
+            # old API does not work since we need reference to transforms
+            # in the main process.
```
fehiepsi: is this issue fixed in this PR?

pyro/infer/mcmc/api.py
```diff
@@ -0,0 +1,322 @@
...
+        self.num_chains = num_chains
+
+        if num_chains > 1:
+            self.sampler = _MultiSampler(kernel, num_samples, self.warmup_steps, num_chains, mp_context,
+                                         disable_progbar, initial_params=initial_params, hook=hook_fn)
+        else:
+            self.sampler = _UnarySampler(kernel, num_samples, self.warmup_steps, disable_progbar,
+                                         initial_params=initial_params, hook=hook_fn)
+
+    def run(self, *args, **kwargs):
```
fehiepsi: the implementation logic LGTM +1

pyro/infer/mcmc/mcmc.py
```diff
@@ -272,6 +272,9 @@ def __init__(self, kernel, num_samples, warmup_steps=None,
         else:
             self.sampler = _SingleSampler(kernel, num_samples, self.warmup_steps, disable_progbar)
         super(MCMC, self).__init__(num_chains=num_chains)
+        warnings.warn("This interface to MCMC is deprecated and will be removed in the "
+                      "next version of Pyro. Please use `pyro.infer.mcmc.api` instead.",
+                      DeprecationWarning)
```
fehiepsi: So if I understand correctly, then when we remove this interface, we can use `infer.mcmc.MCMC` with the new api?

fehiepsi:
> this can be separately discussed, as it will add a bit of boiler plate code in terms of calling `initialize_model`

Yeah, agree, the old interface is already pretty nice to use
```
kernel = HMC(model)
samples = MCMC(kernel).run(*args, **kwargs)
```
and should work for most of usage case.

pyro/infer/mcmc/mcmc.py
neerajprad: We can add this to `infer.mcmc.__init__.py` so that this will be the default interface.

pyro/infer/mcmc/api.py
neerajprad: Done.

pyro/infer/mcmc/api.py
neerajprad: This used to create an issue with the `notebook` env where without the refresh arg, the progress bar wouldn't update (with num_chains > 1). Let me check first if that issue is resolved.

pyro/infer/mcmc/api.py
neerajprad: This now assumes that we are passing the transforms directly to the `MCMC` class itself. But this wouldn't work if we just passed the model arg to `kernel` since with the spawn multiprocessing context, we wouldn't have access to `transforms` in the main process.

neerajprad:
> Yeah, agree, the old interface is already pretty nice to use

> kernel = HMC(model)
> samples = MCMC(kernel).run(*args, **kwargs)
> and should work for most of usage case.

The only exception is that it wouldn't work with multiprocessing (using "spawn") where we need access to the transforms. So we will need to pass the transforms to `MCMC` by calling `initialize_model` (or call initialize model from inside `MCMC.run`).

pyro/infer/mcmc/api.py
```diff
@@ -0,0 +1,322 @@
...
+    :param int warmup_steps: Number of warmup iterations. The samples generated
+        during the warmup phase are discarded. If not provided, default is
+        half of `num_samples`.
+    :param int num_chains: Number of MCMC chains to run in parallel. Depending on
+        whether `num_chains` is 1 or more than 1, this class internally dispatches
+        to either `_SingleSampler` or `_ParallelSampler`.
+    :param dict initial_params: dict containing initial tensors to initiate
+        the markov chain. The leading dimension's size must match that of
+        `num_chains`. If not specified, parameter values will be sampled from
+        the prior.
```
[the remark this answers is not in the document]
neerajprad: Sounds good. Let me add that.

pyro/infer/mcmc/api.py
```diff
@@ -0,0 +1,322 @@
...
+    :param hook_fn: Python callable that takes in `(kernel, samples, stage, i)`
+        as arguments. stage is either `sample` or `warmup` and i refers to the
+        i'th sample for the given stage. This can be
+    :param str mp_context: Multiprocessing context to use when `num_chains > 1`.
+        Only applicable for Python 3.5 and above. Use `mp_context="spawn"` for
+        CUDA.
+    :param bool disable_progbar: Disable progress bar and diagnostics update.
+    """
+    def __init__(self, kernel, num_samples, warmup_steps=None, initial_params=None,
+                 num_chains=1, hook_fn=None, mp_context=None, disable_progbar=False, transforms=None):
```
fehiepsi: nit: we need to add docs for `transforms`. But let's discuss a bit more about this point. I get it now that `mcmc` need to access `transforms`, which is not possible if they are created in child processes. I think that we can just call `initialize_model` here, rather than requires `transforms` to both HMC and MCMC, as you suggested. WDYT?

pyro/infer/mcmc/api.py
```diff
@@ -0,0 +1,322 @@
...
+        given an execution trace returns another sample trace from the target
+        (posterior) distribution.
+    :param int num_samples: The number of samples that need to be generated,
+        excluding the samples discarded during the warmup phase.
+    :param int warmup_steps: Number of warmup iterations. The samples generated
+        during the warmup phase are discarded. If not provided, default is
+        half of `num_samples`.
+    :param int num_chains: Number of MCMC chains to run in parallel. Depending on
+        whether `num_chains` is 1 or more than 1, this class internally dispatches
+        to either `_SingleSampler` or `_ParallelSampler`.
```
fehiepsi: nit: `_SingleSampler` -> `_UnarySampler` :D

pyro/infer/mcmc/api.py
neerajprad: If we use `potential_fn`, we shouldn't need to provide `transforms` to the HMC or NUTS kernel, and I think it makes more sense to provide transforms directly to the MCMC class instead. 

I will try to make it so that we do not need to provide transforms to both the classes though, that would be quite needless.

pyro/infer/mcmc/api.py
```diff
@@ -0,0 +1,332 @@
...
+                                           args=(self.log_queue, self.warmup_steps, self.num_samples,
+                                                 self.num_chains, disable_progbar))
+        self.log_thread.daemon = True
+        self.log_thread.start()
+        self.events = [self.ctx.Event() for _ in range(num_chains)]
+
+    def init_workers(self, *args, **kwargs):
+        self.workers = []
+        for i in range(self.num_chains):
+            init_params = {k: v[i] for k, v in self.initial_params.items()}
```
fehiepsi: I think that this won't work if `initial_params=None`. So for my understanding, here are what is working and what not, which we can address in follow up PRs:
+ with `potential_fn`:
  - need an explicit kwargs `transforms` either in HMC or MCMC.
  - need an explicit `inital_params`, in both 1 chain and multi-chain
+ with `model`:
  - need an explicit `initial_params` for multi-chain

I think we can resolve part of them with
+ with `potential_fn`, use `transforms={}` if it is not provided
+ with `potential_fn`, fix HMC setup so that if  `initial_params` is None, then sample `initial_params` (maybe in (-2, 2)) interval
+ for both case `potential_fn` or `model`, change this code to
```
init_params = {k: v[i] for k, v in self.initial_params.items()} if self.initial_params is not None else None
```

neerajprad: All excellent points. I was planning to address this with the next PR that gets the baseball example working, but I think these are small changes that we might as well fix right now. I'll address them shortly.

neerajprad: Actually, (2) isn't possible. We need to explicitly provide `initial_params` with `potential_fn` since we don't know the data structure / site shapes etc. that the `potential_fn` takes which is additional information that the user needs to provide.

fehiepsi: Oppss, you are right. It is good to add warning as you did. +1

fehiepsi:
@neerajprad is this good to merge if tests pass?

neerajprad:
> @neerajprad is this good to merge if tests pass?

This should be good to merge. I will address one last remaining comment about changing `refresh=False` in a follow-up PR after verifying that the notebook example works.

neerajprad:
Thanks so much for a thorough review, @fehiepsi!

fehiepsi:
Nicely done @neerajprad ! Now, I can start to address some hmc FR and enhancements.

PR #1914, opened 2019-06-18
NUTS: detach `potential_energy`'s from the computation graph

jburroni:
`velocity_verlet` is not detaching the `potential_energy` from the computation graph, and so it keeps track of unnecessary computations.
This might not be the right place to add the `detach` action

### DIFF
--- a/pyro/ops/integrator.py
+++ b/pyro/ops/integrator.py
@@ -34,7 +34,7 @@ def velocity_verlet(z, r, potential_fn, inverse_mass_matrix, step_size, num_step
                                                                         inverse_mass_matrix,
                                                                         step_size,
                                                                         z_grads)
-    return z_next, r_next, z_grads, potential_energy
+    return z_next, r_next, z_grads, potential_energy.detach()
 
 
 def _single_step_verlet(z, r, potential_fn, inverse_mass_matrix, step_size, z_grads=None):
--- a/pyro/ops/integrator.py
+++ b/pyro/ops/integrator.py
@@ -34,7 +34,8 @@ def velocity_verlet(z, r, potential_fn, inverse_mass_matrix, step_size, num_step
                                                                         inverse_mass_matrix,
                                                                         step_size,
                                                                         z_grads)
-    return z_next, r_next, z_grads, potential_energy.detach()
+        
+    return z_next, r_next, z_grads, potential_energy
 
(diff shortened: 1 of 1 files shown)


pyro/ops/integrator.py
```diff
@@ -34,7 +34,7 @@ def velocity_verlet(z, r, potential_fn, inverse_mass_matrix, step_size, num_step
                                                                         inverse_mass_matrix,
                                                                         step_size,
                                                                         z_grads)
-    return z_next, r_next, z_grads, potential_energy
```
neerajprad: Could you share your model where this happens? At no point in HMC do we set `requires_grad = True`, except for when we take the gradient `_potential_grad` below, right after which this is set to False again. Are you by any chance using `pyro.param` in the model?

jburroni:
```python
import torch
from torch import tensor, rand
import pyro
import torch.distributions.constraints as constraints
from pyro.infer import mcmc


def model(x: 'int[10]'=None):
    theta: 'real' = pyro.sample('theta', dist.Uniform(0.0, 1.0))
    pyro.sample('theta' + '__1', dist.Beta(1, 1), obs=theta)
    for i in range(1, 10 + 1):
        pyro.sample('x' + '__{}'.format(i - 1) + '__2', dist.Bernoulli(
            theta), obs=x[i - 1])

x = torch.Tensor([0, 0, 0, 0, 0, 0, 1, 0, 0, 1])
def nuts(model, **kwargs):
    nuts_kernel = mcmc.NUTS(model)
    return mcmc.MCMC(nuts_kernel, **kwargs)

nuts(model, num_samples=10000, warmup_steps=100).run(x=x)
```

pyro/ops/integrator.py
neerajprad: Okay I see the issue. We are also returning `potential_energy` from `_potential_grad`. Could you instead make this change in `_potential_grad` below?

```python
return dict(zip(z_keys, grads)), potential_energy.detach()
```

fehiepsi:
Thanks for detecting this! It seems like an issue to me. I just have a minor comment: Could you detach at _potential_grad function instead? :)

pyro/ops/integrator.py
jburroni: done

Issue #1912, opened 2019-06-17
System Usage with NUTS


jburroni:
While running NUTS on cpu, a big part of the CPU Load is devoted to System.
 This is a consequence of the CPU backend trying to split the work between several threads, even when using a single chain. In particular, this happens inside the calls to `logsumexp`

This is the performance monitor's visualization running NUTS with  a single chain:
![cpu-pyro](https://user-images.githubusercontent.com/2650090/59634665-505f1b00-911d-11e9-9612-37834293b82d.jpg)

 

The following code solves the issue:
```diff
--- a/pyro/infer/mcmc/nuts.py
+++ b/pyro/infer/mcmc/nuts.py
@@ -11,6 +11,14 @@ from pyro.infer.mcmc.hmc import HMC
 from pyro.ops.integrator import velocity_verlet
 from pyro.util import optional, torch_isnan

+
+def _logsumexp(vals, dim=0):
+    assert dim == 0
+    assert vals.shape == (2,)
+    M = vals[0] if vals[0] > vals[1] else vals[1]
+    p = (vals-M).exp().sum().log() + M
+    return p
+
 # sum_accept_probs and num_proposals are used to calculate
 # the statistic accept_prob for Dual Averaging scheme;
 # z_left_grads and z_right_grads are kept to avoid recalculating
@@ -219,7 +227,7 @@ class NUTS(HMC):
                                            direction, tree_depth-1, energy_current)

         if self.use_multinomial_sampling:
-            tree_weight = logsumexp(torch.stack([half_tree.weight, other_half_tree.weight]), dim=0)
+            tree_weight = _logsumexp(torch.stack([half_tree.weight, other_half_tree.weight]), dim=0)
         else:
             tree_weight = half_tree.weight + other_half_tree.weight
         sum_accept_probs = half_tree.sum_accept_probs + other_half_tree.sum_accept_probs
@@ -363,7 +371,7 @@ class NUTS(HMC):
                     break
                 else:  # update tree_weight
                     if self.use_multinomial_sampling:
-                        tree_weight = logsumexp(torch.stack([tree_weight, new_tree.weight]), dim=0)
+                        tree_weight = _logsumexp(torch.stack([tree_weight, new_tree.weight]), dim=0)
                     else:
                         tree_weight = tree_weight + new_tree.weight
```
Notice that I'm not proposing using this code, but only showing where the root of the problem is. 
### Environment
OS: OSX 10.13.6
PyTorch: 1.1.0
[...]

PR #1915, opened 2019-06-18
NUTS: reduce _System usage_. Simpler `logsumexp`

jburroni:
For NUTS, a `logsumexp` with only two arguments is called many times.
Empirically this solves #1912, but as shown by the discussion, it's not clear exactly why this is the case.

### DIFF
--- a/pyro/infer/mcmc/nuts.py
+++ b/pyro/infer/mcmc/nuts.py
@@ -19,6 +19,7 @@ def _logsumexp(vals, dim=0):
     p = (vals-M).exp().sum().log() + M
     return p
 
+
 # sum_accept_probs and num_proposals are used to calculate
 # the statistic accept_prob for Dual Averaging scheme;
 # z_left_grads and z_right_grads are kept to avoid recalculating
--- a/pyro/infer/mcmc/nuts.py
+++ b/pyro/infer/mcmc/nuts.py
@@ -12,12 +12,9 @@ from pyro.ops.integrator import velocity_verlet
 from pyro.util import optional, torch_isnan
 
 
-def _logsumexp(vals, dim=0):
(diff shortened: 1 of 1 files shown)


pyro/infer/mcmc/nuts.py
```diff
@@ -6,11 +6,20 @@
...
 import pyro
 import pyro.distributions as dist
-from pyro.distributions.util import logsumexp, scalar_like
+from pyro.distributions.util import scalar_like
 from pyro.infer.mcmc.hmc import HMC
 from pyro.ops.integrator import velocity_verlet
 from pyro.util import optional, torch_isnan
 
+
+def _logsumexp(vals, dim=0):
```
neerajprad: Would suggest simplifying this to take in the values directly so that we can avoid `stack` below. Also, since we know that the values are scalars, we can remove the dim arg and the asserts as this is only internally used.

```python
def _logaddexp(x, y):
    minval, maxval = (x, y) if x < y else (y, x)
    return (maxval - minval).exp().log() + maxval    
```

neerajprad: ^ should be log1p above; corrected.

fehiepsi: Yeah, I think this version is better. Should it be min - max instead of max - min?

jburroni:
```python
def _logaddexp(x, y):
    minval, maxval = (x, y) if x < y else (y, x)
    return (minval - maxval).exp().log1p() + maxval
```

neerajprad: You're right, should be min - max.

fehiepsi:
Happy to merge this when test passes. :)

PR #1916, opened 2019-06-19
Conditional transform

stefanwebb:
**Status:** Author plans to delete all code in this PR and start over?

Adds functionality to write conditional transformed distributions and gives an example of a normalizing flow for p(x | y)

- [ ] design doc
- [ ] reimplementation along lines of doc
- [ ] ready for re-review


martinjankowiak:
thanks for this PR. this will be great to have.

i'm a little bit unclear about some of the high level design choices here. will each flow need a custom conditional implementation? if so, can't we just have all transforms be conditional such that they effectively become non-conditional when `input=None` (or similar)?

stefanwebb:
> i'm a little bit unclear about some of the high level design choices here. will each flow need a custom conditional implementation? if so, can't we just have all transforms be conditional such that they effectively become non-conditional when `input=None` (or similar)?

Whether a flow is conditional or not needs to be decided when the class is initialized. Conditional transforms differ in that they drop the `nn.parameters` that are set in the initializer and save a hypernetwork that is given to the initializer. They also typically require modifications to the forward, reverse, and logDetJ operations due to broadcasting issues.

The forward/backward/logDetJ methods also need to take an `obs` variable to condition on, and this requires modifying the PyTorch abstractions for implementing transformations. If you'll see I implemented new classes: `ConditionalTransform`, `ConditionalTransformModule`, `ConditionalTransformedDistribution`.

So each flow needs a custom conditional implementation. The question is, should this be contained in the one class with a keyword arg given to the initializer, or should there be two separate classes for this? The former would require overwriting `Transform`, `TransformModule`, and all the PyTorch transforms, etc. similarly to how I implemented `ConditionalTransform` etc. to make it work.

stefanwebb:
@fritzo @martinjankowiak What are your thoughts on how to proceed with the new interface?

stefanwebb:
After thinking about it for a bit, the best solution to me seems to have the conditional and non-conditional transforms in the same class with a `cond=True/False` arg in the initializer.

There needs to be a separate `Transform` and `ConditionalTransform` because forwarding an `obs=...` will break a PyTorch transform. But there doesn't need to be a separate `TransformedDistribution` and `ConditionalTransformedDistribution` if I use optional keyword arguments. The convention will be that the optional keyword argument is `obs=...`, although this won't be enforced.

By the way, the current `ConditionalTransformedDistribution` allows you to mix conditional and non-conditional transforms. E.g. you might want to do an amortized flow, and then pass it through a sigmoid at the end to restrict the support. This would be maintained in a single `TransformedDistribution`

martinjankowiak:
@stefanwebb can you please give some guidance as to the best order in which to review/merge these various PRs?

stefanwebb:
Could we please do #1918 and #1919 first? They are pretty non-controversial and won't require discussion on APIs etc.

#1916 has a high priority from a usefulness perspective, but I think it will take some back and forth to get an interface/design that we can all agree on. Perhaps we could discuss this one while the first two are getting merged?

#1920 is much lower priority. This one is about improving the interface to the normalizing flow library rather than adding any new features.

docs/source/distributions.rst
```diff
@@ -190,7 +197,7 @@ BatchNormTransform
 .. autoclass:: pyro.distributions.BatchNormTransform
     :members:
     :undoc-members:
-    :show-inheritance:
+    :show-inheritance:
```
fritzo: revert

pyro/distributions/conditional_transformed_distribution.py
```diff
@@ -0,0 +1,157 @@
...
+    Note that the ``.event_shape`` of a :class:`ConditionalTransformedDistribution` is the
+    maximum shape of its base distribution and its transforms, since transforms
+    can introduce correlations among events.
+
+    """
+    arg_constraints = {}
+
+    def __init__(self, base_distribution, transforms, validate_args=None):
+        self.base_dist = base_distribution
+        if isinstance(transforms, Transform):
```
fritzo: Why can't this be a `ConditionalTransform`?

Also this logic is gross, please simplify.

fritzo:
I'd like to contribute to the design discussion, but I have no background in conditional flows. Can you  please populate the PR description with background material or point me at a design doc.

fritzo:
@stefanwebb can you also please add some "status: ..." indicator text or something in the PR description? As I understand from the long extended discussion, the code in this PR is all planned for deletion. I've added a WIP tag until you're ready again for review.

pyro/distributions/conditional_transformed_distribution.py
stefanwebb: Yes, you're right it can be a `ConditionalTransform` and the logic ought to be changed. I am hoping to eliminate having separate `Transform` and `ConditionalTransform` in a revised proposal so this issue will be eliminated

stefanwebb:
> @stefanwebb can you also please add some "status: ..." indicator text or something in the PR description? As I understand from the long extended discussion, the code in this PR is all planned for deletion. I've added a WIP tag until you're ready again for review.

What the best way to do this?

fritzo:
@stefanwebb I added some status text to your PR description.

PR #1918, opened 2019-06-24
RealNVP

stefanwebb:
The affine coupling transform from RealNVP


stefanwebb:
@martinjankowiak this is ready to be looked at

pyro/distributions/transforms/affine_coupling.py
```diff
@@ -0,0 +1,130 @@
+from __future__ import absolute_import, division, print_function
+
+import torch
+from torch.distributions import constraints
+
+from pyro.distributions.torch_transform import TransformModule
+from pyro.distributions.util import copy_docs_from
+
+
+def clamp_preserve_gradients(x, min, max):
```
martinjankowiak: if this is being reused in several places it should be put in a utils file somewhere

stefanwebb: Done!

martinjankowiak:
i guess if you rebase off of #1933 tests will pass? or do you want to certify that the failing test is definitively spurious?

stefanwebb:
I think it's probably not worth checking since it passes on my machine and we can clear the backlog by merging this one in

PR #1919, opened 2019-06-24
Conditional MADE Autoregressive Network

stefanwebb:
A MADE that can conditional on an additional input. This will be required for the conditional versions of IAF/NAF/polynomial flow


stefanwebb:
@martinjankowiak this is ready to be looked at

pyro/nn/auto_reg_nn.py
```diff
@@ -98,22 +98,25 @@ def forward(self, _input):
         return F.linear(_input, masked_weight, self.bias)
 
 
-class AutoRegressiveNN(nn.Module):
+class ConditionalAutoRegressiveNN(nn.Module):
     """
-    An implementation of a MADE-like auto-regressive neural network.
+    An implementation of a MADE-like auto-regressive neural network that can conditional on an additional input.
```
martinjankowiak: typo

pyro/nn/auto_reg_nn.py
```diff
@@ -140,15 +143,17 @@ class AutoRegressiveNN(nn.Module):
...
             permutation=None,
             skip_connections=False,
             nonlinearity=nn.ReLU()):
-        super(AutoRegressiveNN, self).__init__()
+        super(ConditionalAutoRegressiveNN, self).__init__()
         if input_dim == 1:
-            warnings.warn('AutoRegressiveNN input_dim = 1. Consider using an affine transformation instead.')
+            warnings.warn('ConditionalAutoRegressiveNN input_dim = 1. Consider using an affine transformation instead.')
         self.input_dim = input_dim
+        self.observed_dim = observed_dim
```
martinjankowiak: should we check that `observed_dim>0`?

fehiepsi:
@stefanwebb While you are at this, I was wondering a bit about the following non-related issue at `create_mask` functionality. Could you help me clarify it?

Basically,
```
import torch
from pyro.nn.auto_reg_nn import create_mask

create_mask(2, 0, [10, 10], torch.tensor([0, 1]), 2)
```
which returns
```
([tensor([[0., 0.],
          [0., 0.],
          [0., 0.],
          [0., 0.],
          [0., 0.],
          [1., 0.],
          [1., 0.],
          [1., 0.],
          [1., 0.],
          [1., 0.]]), tensor([[1., 1., 1., 1., 1., 0., 0., 0., 0., 0.],
          [1., 1., 1., 1., 1., 0., 0., 0., 0., 0.],
          [1., 1., 1., 1., 1., 0., 0., 0., 0., 0.],
          [1., 1., 1., 1., 1., 0., 0., 0., 0., 0.],
          [1., 1., 1., 1., 1., 0., 0., 0., 0., 0.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]]), tensor([[1., 1., 1., 1., 1., 0., 0., 0., 0., 0.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [1., 1., 1., 1., 1., 0., 0., 0., 0., 0.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]])],
 tensor([[0., 0.],
         [1., 0.],
         [0., 0.],
         [1., 0.]]))
```
I can observe that it is over-masking the weights: e.g. the first input only connects to 5 hidden nodes. This suggests that: despite we are setting `hidden_dim=10`, we only use 5 hidden nodes, so it is somehow equivalent to set `hidden_dim=5` from MADE paper. If I change [this line](https://github.com/pyro-ppl/pyro/blob/ab5e5ed1493321027265155c90fcaf2133f6bf2d/pyro/nn/auto_reg_nn.py#L24)  to `torch.arange(1, hidden_dim + 1, dtype=torch.float32) % (input_dim - 1) + 1`, I get
```
([tensor([[1., 0.],
          [1., 0.],
          [1., 0.],
          [1., 0.],
          [1., 0.],
          [1., 0.],
          [1., 0.],
          [1., 0.],
          [1., 0.],
          [1., 0.]]), tensor([[1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]]), tensor([[0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.],
          [0., 0., 0., 0., 0., 0., 0., 0., 0., 0.],
          [1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]])],
 tensor([[0., 0.],
         [1., 0.],
         [0., 0.],
         [1., 0.]]))
```
which is the result I expect following MADE paper.

Is this difference expected? or are we using some kinds of dropout here?

stefanwebb:
@fehiepsi The reason I designed it that way is so that when you have observed inputs, there are hidden units at every layer that are responsible for transforming only the observed variables. I agree it is a bit wasteful when `observed_dim=0`. I've added a fix for that case!

pyro/nn/auto_reg_nn.py
stefanwebb: Fixed!

pyro/nn/auto_reg_nn.py
stefanwebb: If I added `if observed_dim < 1: ...` into `ConditionalAutoRegressiveNN` then `AutoRegressiveNN` would throw the warning/error. There isn't a neat way to test the type in the initializer. The best I can think of is: `if type(self) == "<class 'pyro.nn.auto_reg_nn.ConditionalAutoRegressiveNN'>" and observed_dim < 1: ...`, but it seems like it isn't worth the trouble for a fragile solution (e.g. if you changed the module it was in it would break!)

stefanwebb:
@martinjankowiak I've made those changes!

pyro/nn/auto_reg_nn.py
```diff
@@ -52,16 +52,17 @@ def create_mask(input_dim, observed_dim, hidden_dims, permutation, output_dim_mu
 
     # Create the indices that are assigned to the neurons
     input_indices = torch.cat((torch.zeros(observed_dim), 1 + var_index))
-    hidden_indices = [sample_mask_indices(input_dim, h) for h in hidden_dims]
+    if observed_dim > 0:
+        hidden_indices = [sample_mask_indices(input_dim, h) for h in hidden_dims]
+    else:
+        hidden_indices = [sample_mask_indices(input_dim - 1, h) + 1 for h in hidden_dims]
```
fehiepsi: This change works perfectly! Thanks @stefanwebb !

stefanwebb: No worries!

fehiepsi: Can we reduce the offset of hidden_indices by `1` and change the inequality at lines 66 and 71 to `>=` and `>` instead? There is no harm with it, I just want to make it clear that they are indices as in the figure in MADE paper. :)

stefanwebb: Okay, I've made this change!

stefanwebb:
I think there is something in the config that is stopping all of my PRs from passing CI

fehiepsi:
@stefanwebb could open an issue for it? I'll take a look later. I think that the issue is due to the `jit` tutorial is not compatible to new package.

martinjankowiak:
@stefanwebb i should look at this next, right?

stefanwebb:
> @stefanwebb i should look at this next, right?

yes please

pyro/nn/auto_reg_nn.py
```diff
@@ -52,22 +52,23 @@ def create_mask(input_dim, observed_dim, hidden_dims, permutation, output_dim_mu
 
     # Create the indices that are assigned to the neurons
     input_indices = torch.cat((torch.zeros(observed_dim), 1 + var_index))
-    hidden_indices = [sample_mask_indices(input_dim, h) for h in hidden_dims]
+    if observed_dim > 0:
+        hidden_indices = [sample_mask_indices(input_dim, h)-1 for h in hidden_dims]
```
martinjankowiak: what's going on in this branch? `mask-1`?

pyro/nn/auto_reg_nn.py
```diff
@@ -98,22 +99,25 @@ def forward(self, _input):
...
+    >>> p = arn(x, obs=y)  # 1 parameters of size (100, 10)
+    >>> arn = ConditionalAutoRegressiveNN(10, 5, [50], param_dims=[1, 1])
+    >>> m, s = arn(x, obs=y) # 2 parameters of size (100, 10)
+    >>> arn = ConditionalAutoRegressiveNN(10, 5, [50], param_dims=[1, 5, 3])
+    >>> a, b, c = arn(x, obs=y) # 3 parameters of sizes, (100, 1, 10), (100, 5, 10), (100, 3, 10)
 
-    :param input_dim: the dimensionality of the input
+    :param input_dim: the dimensionality of the input variable
     :type input_dim: int
+    :param observed_dim: the dimensionality of the observed variable
```
martinjankowiak: i think we should rename this `context_dim` etc. throughout unless there are better name suggestions?

pyro/nn/auto_reg_nn.py
stefanwebb: If there are context variables, the mask will have indices from 0 to (input_dim - 1). If there are no context variables, the mask will go from 1 to (input_dim - 1). The 0 index corresponds to the context variables, so this lets us build a representation of the context variables at each layer of the MADE in the conditional case

pyro/nn/auto_reg_nn.py
stefanwebb: I've changed `observed_dim` to `context_dim` and `obs` to `context` elsewhere

stefanwebb:
@martinjankowiak this is ready for you to look at again

fehiepsi:
@stefanwebb I am not familiar with conditional flow. Could you provide me some references so I can grab the background first? ^^

stefanwebb:
Thanks, @fehiepsi! This paper adds a tiny extension so you can represent conditional distributions with MADE: https://arxiv.org/abs/1602.06701

Essentially it introduces a 0 index that all the conditioned variables are connected to

fehiepsi:
Thanks @stefanwebb ! Now, I understand why you use `sample_mask_indices(input_dim, h) - 1` for hidden indices.

pyro/nn/auto_reg_nn.py
```diff
@@ -96,22 +97,25 @@ def forward(self, _input):
...
+    >>> arn = ConditionalAutoRegressiveNN(10, 5, [50], param_dims=[1, 1])
+    >>> m, s = arn(x, context=y) # 2 parameters of size (100, 10)
+    >>> arn = ConditionalAutoRegressiveNN(10, 5, [50], param_dims=[1, 5, 3])
+    >>> a, b, c = arn(x, context=y) # 3 parameters of sizes, (100, 1, 10), (100, 5, 10), (100, 3, 10)
 
-    :param input_dim: the dimensionality of the input
+    :param input_dim: the dimensionality of the input variable
     :type input_dim: int
+    :param context_dim: the dimensionality of the context variable
+    :type context_dim: int
```
fehiepsi: nit: it would be clearer if you add a reference to https://arxiv.org/abs/1602.06701.

pyro/nn/auto_reg_nn.py
fehiepsi: nit: I think you can add a comment with the above explanation or link to Section 3.3 of https://arxiv.org/abs/1602.06701

stefanwebb:
Great, I'll make that small change!

@fehiepsi The next PR I'll be doing after this and #1937 gets merged is a conditional BNAF, which will show how a conditional MADE is useful. Also, I'll then have the details fresh in my mind to check out your NumPyro BNAF implementation

stefanwebb:
Okay, made the adjustments

stefanwebb:
I don't think the Travis error has to do with my code...

PR #1920, opened 2019-06-24
Reorganization of Neural Autoregressive Flow

stefanwebb:
Reorganization of NAF. From three separate classes into a single class `transforms.NeuralAutoregressive` that names what the transformation actually does


stefanwebb:
@martinjankowiak this is ready to be looked at too

stefanwebb:
@fehiepsi Yes, that's a cracking idea! I've added 3 new element-wise transforms and rewritten those two flows

stefanwebb:
This is ready for re-review

stefanwebb:
What's causing the CI to fail?

pyro/distributions/transforms/block_autoregressive.py
```diff
@@ -109,17 +112,22 @@ def _call(self, x):
         y = x
         for idx in range(len(self.layers)):
             pre_activation, dy_dx = self.layers[idx](y.unsqueeze(-1))
-            J_act = self.log_df_dx(pre_activation).view(*(list(x.size()) + [-1, 1]))
 
             if idx == 0:
+                y = self.T._call(pre_activation)
+                J_act = self.T.log_abs_det_jacobian((pre_activation).view(
+                    *(list(x.size()) + [-1, 1])), y.view(*(list(x.size()) + [-1, 1])))
```
fehiepsi: I think that it is better to avoid computing `y` and simply use `self.T.log_abs_det_jacobian(x, None)`. But the current version of `sigmoid` activation has logdet depending on `y` (it is fine with the master version of PyTorch), so it is up to you to decide what is the best solution. Maybe we can leave a TODO comment here to address it in the future.

pyro/distributions/transforms/neural_autoregressive.py
```diff
@@ -0,0 +1,191 @@
...
+        """
+
+        A = self._cached_A
+        W_pre = self._cached_W_pre
+        C = self._cached_C
+        T = self.T
+
+        log_dydD = self._cached_log_df_inv_dx
+        log_dDdx = torch.logsumexp(torch.log(A + eps) + self.logsoftmax(W_pre) +
+                                   T.log_abs_det_jacobian(C, T._call(C)), dim=-2)
```
fehiepsi: nit: we can use `T(C)` or `None` if `Sigmoid` issue is resolved.

pyro/distributions/transforms/neural_autoregressive.py
```diff
@@ -0,0 +1,191 @@
...
+        T = self.T
+
+        # Divide the autoregressive output into the component activations
+        A = F.softplus(A)
+        C = A * x.unsqueeze(-2) + b
+        W = F.softmax(W_pre, dim=-2)
+        D = (W * T._call(C)).sum(dim=-2)
+        y = T._inverse(D)
+
+        self._cached_log_df_inv_dx = -T.log_abs_det_jacobian(y, D)
```
fehiepsi: nit: we can use `T.inv.log_abs_det_jacobian(D, y)`

pyro/distributions/transforms/neural_autoregressive.py
```diff
@@ -0,0 +1,191 @@
...
+        # A, W, b ~ batch_shape x hidden_units x event_shape
+        A, W_pre, b = self.arn(x)
+        T = self.T
+
+        # Divide the autoregressive output into the component activations
+        A = F.softplus(A)
+        C = A * x.unsqueeze(-2) + b
+        W = F.softmax(W_pre, dim=-2)
+        D = (W * T._call(C)).sum(dim=-2)
+        y = T._inverse(D)
```
fehiepsi: nit: use `y = T.inv(D)`

pyro/distributions/transforms/neural_autoregressive.py
```diff
@@ -0,0 +1,191 @@
...
+        """
+        # A, W, b ~ batch_shape x hidden_units x event_shape
+        A, W_pre, b = self.arn(x)
+        T = self.T
+
+        # Divide the autoregressive output into the component activations
+        A = F.softplus(A)
+        C = A * x.unsqueeze(-2) + b
+        W = F.softmax(W_pre, dim=-2)
+        D = (W * T._call(C)).sum(dim=-2)
```
fehiepsi: nit: use `T(C)`

pyro/distributions/transforms/block_autoregressive.py
```diff
@@ -109,17 +112,22 @@ def _call(self, x):
...
                 logDetJ = dy_dx + J_act
-                y = self.f(pre_activation)
+
             elif idx < len(self.layers) - 1:
+                y = self.T._call(pre_activation)
+                J_act = self.T.log_abs_det_jacobian((pre_activation).view(
+                    *(list(x.size()) + [-1, 1])), y.view(*(list(x.size()) + [-1, 1])))
```
fehiepsi: nit: similar to above

pyro/distributions/transforms/block_autoregressive.py
stefanwebb: Computing `y` is necessary for the next iteration of the loop

pyro/distributions/transforms/neural_autoregressive.py
stefanwebb: I've decided to cache T(C) from the forward method

pyro/distributions/transforms/block_autoregressive.py
fehiepsi: yup, you are right. :D

stefanwebb:
Okay, I've made the changes!

fehiepsi:
@stefanwebb you will need to resolve conflict before we can merge.

stefanwebb:
> @stefanwebb you will need to resolve conflict before we can merge.

Done!

martinjankowiak:
@fehiepsi please don't merge any flow PRs until i've had a chance to look at them and approve them 👍

fehiepsi:
Sure! :D

stefanwebb:
I've pulled in the latest changes so this can merge.

Just to refresh your memory, the purpose of this and #1932 is to get a consistent, readable naming of the flow transforms. E.g., we will have: AffineAutoregressive, NeuralAutoregressive, BlockAutoregressive, AffineCouple, SplineCoupling etc. The names will say what the transform does, rather than the name of the original paper they were in.

This PR also improves the interface for `transforms.NeuralAutoregressive`, combining three separate NAF transforms that should have been a single one...

fehiepsi:
@martinjankowiak I have reviewed the changes in this PR previously, so would like to defer another pass for you. I don't have much opinion on the names of these flows.

martinjankowiak:
@stefanwebb i'm a bit confused by this. what is the relationship between `ELUTransform` and `NeuralAutoregressive`?

fehiepsi:
@stefanwebb There are two enhancements which I used while implementing NumPyro version
+ For gate, using `a * f(x) + (1 - a) * x` is simpler than `a * x + (1 - a) * f(x)` when computing logdet
+ For `log_matrix_product`, using @fritzo's [version](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/hmm.py#L10) will be faster.

Those are not related to this PR so you might consider updating them in a future PR.

stefanwebb:
> @stefanwebb There are two enhancements which I used while implementing NumPyro version
> 
> * For gate, using `a * f(x) + (1 - a) * x` is simpler than `a * x + (1 - a) * f(x)` when computing logdet
> * For `log_matrix_product`, using @fritzo's [version](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/hmm.py#L10) will be faster.
> 
> Those are not related to this PR so you might consider updating them in a future PR.

Thanks for the tip!

stefanwebb:
> @stefanwebb i'm a bit confused by this. what is the relationship between `ELUTransform` and `NeuralAutoregressive`?

`ELUTransform`, `LeakyRELUTransform`, and `TanhTransform` are standard transforms.

`NeuralAutoregressive` now takes an initializer argument that selects the transformation to use in the formula. So there is a separation now of the NAF transform and the exact activation function it uses! 

Previously there was a separate version of `NeuralAutoregressive` for each activation function

martinjankowiak:
should ELUTransform etc. be set off separately in the documentation? they haven no learnable params etc

stefanwebb:
> should ELUTransform etc. be set off separately in the documentation? they haven no learnable params etc

I think they should, because they can be used separately, just like `torch.distributions.transforms.SigmoidTransform`, which is used by `NeuralAutoregressive` in the same way

martinjankowiak:
in that case i think you want to add something like

Parameter Free Transforms
~~~~~~~~~~~~~~~~~~~~~~~~~

to the relevant .rst file

stefanwebb:
> in that case i think you want to add something like
> 
> Parameter Free Transforms
> 
> ```
> 
> to the relevant .rst file
> ```

How about this (see last commit)?

separating transforms and transformmodules in docs

stefanwebb:
It looks like Travis died...

Issue #1926, opened 2019-06-28
Support HMC/NUTS within Gibbs, online HMC


fritzo:
It would be nice to support HMC/NUTS-within-Gibbs / online HMC (e.g. [HMC with subsampling](https://arxiv.org/pdf/1708.00955.pdf)), but this may require some changes to the HMC and NUTS interfaces. This issue tracks those changes.

## Required changes
- [x] Support changing `*args,**kwargs` across HMC steps.
    Currently these are bound in the `potential_fn()` closure created by `_pe_maker()`. It's easy enough to pass references rather than tensors (e.g. `[x]` rather than a tensor `x`), but that breaks jit compilation.
    - **Workaround:** define a custom `potential_fn`, `transforms`, and set `.initial_params`.
- [x] #1927 Support recomputing `potential_energy` in `.sample()`.
    This may be triggered due to state change other than in `params` (e.g. when new data is added via a side channel into `potential_fn()`). Maybe a `.clear_cache()` arg and logic to recompute the results of `._fetch_from_cache()`?
- [x] Warmup indefinitely.
    Currently there are 5 stages of mass matrix adaptation. For online problems we will want to continually update the mass matrix, maybe every `adapt_window` steps when provided, and when `warmup_steps=float('inf')`.
    - **Workaround:** Use HMC or NUTS directly rather than the MCMC interface. Call `kernel.setup(warmup_steps=9999999999)`, forcing effectively infinite adaptation phase. To set a non-exponential schedule for mass matrix update, set the `kernel._adapter._adaptation_schedule` property.

fritzo:
@neerajprad @fehiepsi maybe we can incorporate these into your current refactoring work.

neerajprad:
> Workaround: define a custom potential_fn, transforms, and set .initial_params.

Are you able to get this to work? I think we may need some changes for this since by default HMC assumes that we need to run MCMC on all the parameters passed to sample. It makes similar assumptions for potential_fn, which should take a dict of unconstrained samples as arguments. While it can be bound in a closure with changing args, kwargs, we'll not be able to JIT it.

fritzo:
> Are you able to get [the workaround] to work?

Yes, I believe this should work but I haven't verified end-to-end. My plan is to:
1. bind extra (non-MCMC) params in a `potential_fn()` closure.
2. JIT the `potential_fn()` by hand before passing it to `HMC`, taking care to handle both MCMC params and non-MCMC dynamic params.

[...]

PR #1927, opened 2019-06-28
Ability to clear cached potential energy computation in HMC

neerajprad:
Addresses #1926 - particularly, ability to clear the cache and recompute PE. 

I noticed that clearing the cache in warmup gives different results due to a different step size being learnt in adaptation (between steps 80 to 100), which is odd because this shouldn't affect the adaptation in any way. 

TODO:
 - [x] Debug why clearing the cache results in different adaptation results (step size changes after step 80 or so in warmup).

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -149,14 +149,18 @@ class HMC(MCMCKernel):
         self._z_grads_last = None
         self._warmup_steps = None
 
-    def _find_reasonable_step_size(self):
+    def _find_reasonable_step_size(self, params):
         step_size = self.step_size
 
         # We are going to find a step_size which make accept_prob (Metropolis correction)
         # near the target_accept_prob. If accept_prob:=exp(-delta_energy) is small,
         # then we have to decrease step_size; otherwise, increase step_size.
         z, potential_energy, z_grads = self._fetch_from_cache()
-        if not z:
+        # recompute PE when cache is cleared
+        if z is None:
+            z = params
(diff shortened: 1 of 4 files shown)


pyro/infer/mcmc/hmc.py
```diff
@@ -270,13 +270,22 @@ def _cache(self, z, potential_energy, z_grads):
...
+
     def _fetch_from_cache(self):
         return self._z_last, self._potential_energy_last, self._z_grads_last
 
     def sample(self, params):
         z, potential_energy, z_grads = self._fetch_from_cache()
+        # recompute PE when cache is cleared
+        if z is None:
+            z = params
+            potential_energy = self.potential_fn(z)
```
fritzo: IIUC `z_grads` will still be `None` when passed into `velicity_verlet()` below. Is this intended?

fritzo: Ah I see this is ok :+1:

neerajprad: Yes, the integrator will compute it if it's None.

fehiepsi:
I think that somehow MCMC just modifies `params` for the next step. Let me check where is the cause.

neerajprad:
> the test does not use MCMC, so that might not be related. Let me check where is the cause.

@fehiepsi - the issue was with `find_reasonable_step_size` which uses the cache. I am currently recomputing here as well, though it might be simpler to just not use cache in this function since it is probably called only a select number of times.

fehiepsi:
Yeah, that's right!

PR #1932, opened 2019-06-30
Reorganization of IAF + helper functions to create transforms

stefanwebb:
Renamed IAF transforms to match other transform names: `pyro.distributions.transforms.AffineAutoregressive`. Combined two very similar versions of IAF into one class with a switch keyword argument


fehiepsi:
Maybe it is better to keep Inverse name. Affine is used for MaskedAutoregressiveTransform too.

stefanwebb:
> Maybe it is better to keep Inverse name. Affine is used for MaskedAutoregressiveTransform too.

I've got a few arguments for changing the name to `AffineAutoregressive`:

* Parallelism with the other transform names: `AffineCoupling`, `NeuralAutoregressive`, `BlockAutoregressive` (and the forthcoming`SplineCoupling`/`SplineAutoregressive` https://arxiv.org/abs/1906.04032)
* Acknowledging that this is a bijective transform, not a complete normalizing flow method. You could form IAF by, e.g.: `flow = [AffineAutoregressive(...), PermuteTransform(...), AffineAutoregressive(...)]` and MAF by e.g.: `flow = [AffineAutoregressive(...).inv(), BatchNormTransform(...), PermuteTransform(...), AffineAutoregressive(...).inv()]`
* "Inverse Autoregressive Flow" and "Masked Autoregressive Flow" are both essentially the same method, swapping the forward and inverse operation, and IAF came first in the literature. But unless you are familiar with the papers, the parts of the names "inverse" and "masked" are confusing. The name of the class should say what it does!
* On the note of an affine transformation is used for MAF too, there is a general convention in the transforms library that the forward operation is tractable and the inverse not so. So `AffineAutoregressive` would be like `NeuralAutoregressive` in that the forward operation is the tractable one. You can then get the density estimator versions (for MAF & MAF-NAF) by e.g. `AffineAutoregressive(...).inv()` and `NeuralAutoregressive(...).inv()`

stefanwebb:
@fehiepsi @martinjankowiak I've pulled in the latest commits so this can be automatically merged!

martinjankowiak:
@stefanwebb fix merge conflicts?

stefanwebb:
I was fixing the merge conflicts and realized that:

1. It's quite hard to produce a hypernetwork you give as an argument to some of the flows that has the correct output dimensions unless you know what you're doing.

2. Much of the transform test code is made up of helpers to create the transform objects correctly.

The solution I came up with is having helper functions in each transform module that creates the transform with sensible default arguments, correctly specifying the input/output dimensions of the hypernetwork if it is used.

They are really useful for transforms that take a hypernetwork as an argument, and non-essential for ones that don't but are included for the latter for consistency.

This makes it easier to get started with using flows without errors, plus the logic of the test files is vastly simplified!

I have taken this opportunity to simplify the names of the other existing transforms as well. It seems superfluous to end a transform with `Transform` - they're already in the `pyro.distributions.transform` module.

Also, I discovered and fixed a bug where the Householder inverse wasn't being calculated correctly when there was more than 1 transform!

@martinjankowiak apologies that this PR has morphed!

stefanwebb:
@martinjankowiak would it be helpful if I broke this up into a number of PRs, one for each transform that I've changed? I can do the conditional versions in each one as well at the same time

martinjankowiak:
some sort of breaking up might be helpful (though too many prs is also bad). in any case i have no time to look at this until after aistats (oct8)

martinjankowiak:
some sort of breaking up might be helpful (though too many prs is also bad). in any case i have no time to look at this until after aistats (oct8)

stefanwebb:
> some sort of breaking up might be helpful (though too many prs is also bad). in any case i have no time to look at this until after aistats (oct8)

Okay, let's keep this one as it is and I'll keep working on some other features like the conditional versions of the transforms in separate PRs. Good luck for AISTATS!

pyro/distributions/transforms/batchnorm.py
```diff
@@ -127,3 +127,19 @@ def log_abs_det_jacobian(self, x, y):
             # NOTE: You wouldn't typically run this function in eval mode, but included for gradient tests
             var = self.moving_variance
         return (-self.constrained_gamma.log() + 0.5 * torch.log(var + self.epsilon))
+
+
+def batchnorm(input_dim, **kwargs):
```
martinjankowiak: do we really want trivial helpers like this one?

pyro/distributions/transforms/block_autoregressive.py
```diff
@@ -237,3 +237,25 @@ def forward(self, x):
 
         w, wpl = self.get_weights()
         return (torch.matmul(w, x) + self.bias.unsqueeze(-1)).squeeze(-1), wpl
+
+
+def block_autoregressive(input_dim, **kwargs):
```
martinjankowiak: ditto

pyro/distributions/transforms/householder.py
```diff
@@ -105,7 +105,14 @@ def _inverse(self, y):
         point around a plane, then the same operation will reflect it back
         """
 
-        return self._call(y)
+        x = y
+        u = self.u()
+        for idx in range(self.count_transforms):
+            # NOTE: Need to apply transforms in reverse order as forward operation!
+            jdx = self.count_transforms - idx - 1
```
martinjankowiak: use `reversed()` instead of explicit indexing?

martinjankowiak: was this not exercised by previous tests?

pyro/distributions/transforms/householder.py
```diff
@@ -114,3 +121,19 @@ def log_abs_det_jacobian(self, x, y):
         """
 
         return torch.zeros(x.size()[:-1], dtype=x.dtype, layout=x.layout, device=x.device)
+
+
+def householder(input_dim, count_transforms=None):
```
martinjankowiak: ditto (trivial?)

pyro/distributions/transforms/neural_autoregressive.py
```diff
@@ -35,9 +37,16 @@ def log_abs_det_jacobian(self, x, y):
         return -F.relu(-x)
 
 
-class LeakyReLUTransform(torch.distributions.transforms.Transform):
+def elu():
```
martinjankowiak: ditto

pyro/distributions/transforms/permute.py
```diff
@@ -83,3 +83,20 @@ def log_abs_det_jacobian(self, x, y):
         """
 
         return torch.zeros(x.size()[:-1], dtype=x.dtype, layout=x.layout, device=x.device)
+
+
+def permute(input_dim, permutation=None):
```
martinjankowiak: ditto

stefanwebb:
Yikes, I better get on this! Thanks for the heads up

pyro/distributions/transforms/batchnorm.py
stefanwebb: I think yes! The point of the helpers is that they give you a default which doesn't require you to know any of the details of how the transformation is parametrized. If you had to know that the batch-norm doesn't take a hypernet then that would defeat the point. Also, if there were conditional versions of batch-norm, householder, etc. they would take a hypernet and have a non-trivial helper function

pyro/distributions/transforms/affine_autoregressive.py
```diff
@@ -222,13 +195,29 @@ def _inverse(self, y):
...
-        return log_scale.sum(-1)
+
+def affine_autoregressive(input_dim, hidden_dims=None, **kwargs):
+    """
+    A helper function to create an AffineAutoregressive object that takes care of constructing
+    an autoregressive network with the correct input/output dimensions.
+
+    :param input_dim: Dimension of input variable
+    :type input_dim: int
+    :param hidden_dims: The desired hidden dimensions of of the autoregressive network. Defaults
```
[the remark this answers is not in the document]
stefanwebb: Fixed

pyro/distributions/transforms/householder.py
stefanwebb: No it wasn't, but it is now. And I've made that change!

pyro/distributions/transforms/affine_autoregressive.py
```diff
@@ -222,13 +195,29 @@ def _inverse(self, y):
...
-        Calculates the elementwise determinant of the log jacobian
-        """
-        if self._cached_log_scale is not None:
-            log_scale = self._cached_log_scale
-        else:
-            _, logit_scale = self.arn(x)
-            log_scale = self.logsigmoid(logit_scale + self.sigmoid_bias)
-        return log_scale.sum(-1)
+
+def affine_autoregressive(input_dim, hidden_dims=None, **kwargs):
```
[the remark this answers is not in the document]
stefanwebb: I've made the switch to `affine_autoregressive` for the DMM example

pyro/distributions/transforms/batchnorm.py
fritzo: I think it's fine to add trivial helpers, but only if you add some documentation section somewhere with a menu of all available helpers with links to their docs.

docs/source/distributions.rst
```diff
@@ -377,10 +377,68 @@ TransformModule
...
     :show-inheritance:
 
-Constraints
-~~~~~~~~~~~
-.. automodule:: pyro.distributions.constraints
-    :members:
-    :member-order: bysource
-    :undoc-members:
-    :show-inheritance:
+Transform Factories
```
fritzo: As commented below, let's try to ensure these appear in some docs sidebar somewhere. No need to do so in this PR, we can tweak docs in follow-up PRs.

fritzo:
@stefanwebb can you
- fix failing tests
- ensure you're up to date with `dev` branch
- `make docs` locally and verify thinks look ok

examples/dmm/dmm.py
```diff
@@ -156,7 +156,7 @@ def __init__(self, input_dim=88, z_dim=100, emission_dim=100,
                           dropout=rnn_dropout_rate)
 
         # if we're using normalizing flows, instantiate those too
-        self.iafs = [AffineAutoregressive(AutoRegressiveNN(z_dim, [iaf_dim])) for _ in range(num_iafs)]
+        self.iafs = [affine_autoregressive(z_dim, hidden_dims=[iaf_dim]) for _ in range(num_iafs)]
```
martinjankowiak: this change should probably be mirrored in dmm.ipynb

docs/source/distributions.rst
```diff
@@ -262,9 +269,9 @@ LeakyReLUTransform
     :undoc-members:
     :show-inheritance:
 
-PermuteTransform
-----------------
-.. autoclass:: pyro.distributions.transforms.PermuteTransform
+Permute
```
martinjankowiak: some of these changes probably need deprecation warnings? @fritzo can you please point stefan to a good example pattern to follow?

pyro/distributions/transforms/neural_autoregressive.py
```diff
@@ -84,21 +102,29 @@ def log_abs_det_jacobian(self, x, y):
         return - 2. * (x - math.log(2.) + F.softplus(- 2. * x))
 
 
+def tanh():
```
martinjankowiak: name overloaded?

pyro/distributions/transforms/permute.py
```diff
@@ -83,3 +84,21 @@ def log_abs_det_jacobian(self, x, y):
...
+    :param input_dim: Dimension of input variable
+    :type input_dim: int
+    :param permutation: Torch tensor of integer indices representing permutation. Defaults
+        to a random permutation.
+    :type permutation: torch.LongTensor
+
+    """
+
+    if permutation is None:
+        permutation = torch.randperm(input_dim, device='cpu').to(torch.Tensor().device)
```
martinjankowiak: optimal device usage? @fritzo ?

martinjankowiak:
yes let's push this through!

are either of these issues relevant?

https://github.com/pyro-ppl/pyro/issues/2076
https://github.com/pyro-ppl/pyro/issues/2147

Am Do., 14. Nov. 2019 um 05:30 Uhr schrieb Fritz Obermeyer <
notifications@github.com>:

stefanwebb:
Yes those two issues are relevant, and also #2084. Will fix this evening

pyro/distributions/transforms/permute.py
fritzo: `torch.randperm` now has cuda support. Please change to simply:
```py
permutation = torch.randperm(input_dim)
```

docs/source/distributions.rst
fritzo: IIUC think these have been renamed without making redirects, and I am ok with that. If you believe any of these flows have important existing users, you could add a redirect
```py
def PermuteTransform(*args, **kwargs):
    warnings.warn("PermuteTransform has been renamed to Permute",
                  DeprecationWarning)
    return Permute(*args, **kwargs)
```
However I think its fine to omit these redirects.

pyro/distributions/transforms/neural_autoregressive.py
fritzo: I think this is fine as long as it only lives in this file and `pyro.distributions.transforms`.

fritzo:
@stefanwebb If possible we'd prefer merging this big PR asap, and moving bugfixes to small easy-to-review follow-up PRs. It is a significant burden to review large PRs 😉 Indeed I'm ok leaving bugfixes until after 1.0 release (1.0 is about interfaces more than correctness). But if those fixes are needed for tests to pass, do add them here.

docs/source/distributions.rst
stefanwebb: Okay, I'll omit these!

pyro/distributions/transforms/batchnorm.py
stefanwebb: I've added a sidebar for helpers in docs

pyro/distributions/transforms/permute.py
stefanwebb: Done!

examples/dmm/dmm.py
stefanwebb: Could we leave this for a separate PR after the 1.0 release?

stefanwebb:
Hmm, not sure why docs are failing to build

fritzo:
> why docs are failing to build

could be a cyclic dependency

stefanwebb:
> > why docs are failing to build
> 
> could be a cyclic dependency

Does that mean two of my docstrings refer to each other?

fritzo:
Probably two of your modules import each other. My hypothesis is that the docs failing is actually a deeper problem.

fritzo:
@stefanwebb the problem is that `pyro.nn` now depends on `pyro.distributions`, and you are adding a reverse dependency. To fix change all imports `from pyro.nn import ___` in pyro.distributions to more specific imports like `from pyro.nn.auto_reg_nn import ___`. For example
```diff
- from pyro.nn import AutoRegressiveNN
+ from pyro.nn.auto_reg_nn import AutoRegressiveNN
```

fritzo:
@stefanwebb Could you instead change these two lines in pyro.nn to avoid it depending on pyro.distributions:
```diff
diff --git a/pyro/nn/module.py b/pyro/nn/module.py
index 7073cc1b..bcdc7e04 100644
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -15,7 +15,6 @@ import torch
 from torch.distributions import constraints, transform_to

 import pyro
-import pyro.distributions as dist
 from pyro.poutine.runtime import _PYRO_PARAM_STORE

 PyroParam = namedtuple("PyroParam", ("init_value", "constraint", "event_dim"))
@@ -202,7 +201,7 @@ class PyroModule(torch.nn.Module):
             _pyro_samples = self.__dict__['_pyro_samples']
             if name in _pyro_samples:
                 prior = _pyro_samples[name]
-                if not isinstance(prior, (dist.Distribution, torch.distributions.Distribution)):
+                if not hasattr(prior, "sample"):  # if not a distribution
                     prior = prior(self)
                 context = self._pyro_context
                 if context.active:
```

stefanwebb:
@fritzo thanks, I'm seeing if the merge has fixed the docs now! I'll make your suggested change if not

stefanwebb:
The docs build now, but there was a problem with my merge that meant an AutoGuide test was failing. I think I've fixed it now!

stefanwebb:
It's only the examples that fail with error message: `No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`

Are you able to restart Travis?

martinjankowiak:
@stefanwebb glad this got merged, thanks stefan and @fritzo!  can we please make a github issue to keep track of remaining bug fixes etc that still need to be fixed?

PR #1937, opened 2019-06-30
Transform and TransformedDistribution that can forward optional keyword args

stefanwebb:
I've redesigned my proposal for conditional transformed distributions and will break it up into several parts (@fritzo, I will post a design spec here soon).

The first part redefines the `Transform` and `TransformedDistribution` classes of PyTorch in Pyro so they can optionally take keyword arguments (for an input to condition on) in the `sample()`, `log_prob()` etc. methods.

This is transparent to the existing transforms- all the normalizing flow transforms use these modified classes without any change in functionality, but we are now able to use keyword args in conditional flows.


fritzo:
Hmm `**kwargs` is a cumbersome mechanism. For example your `kwargs` cannot use the names `x`, `y`, `value` due to conflict in various methods.

What do you think of these alternatives:
1. A single standard arg `cond_data` or `condition_data` or `conditioning_data` that defaults to `None` but can hold your various data
2. Keep the existing interfaces for `Transform` and `TransformedDistribution` methods.Create a new method `.conditional(-)` that returns a standard `Transform` or `TransformedDistribution` object, basically a `functools.partial(transform_builder, condition_data=my_condition_data)`.

Can you also point me to some example code snippets demonstrating your typical usage patterns? It would be nice to design the interface so that we could automate some of those usage patterns (and `**kwargs` is difficult to automate).

martinjankowiak:
btw a better name for the thing we're conditioning on might be `context`  (short, and isn't potentially confusing like `observed`)

fritzo:
Hi @stefanwebb I started a [design doc](https://docs.google.com/document/d/15hxDQtkupIUfTWDeg5--95p9xRi7JAl6MvDkMy1WAP8/edit?usp=sharing) to discuss a new interface for Distributions and TransformedDistributions, since I'm facing similar need in other projects. Let's collaborate on that doc and try to find a flexible interface :smile:

stefanwebb:
Hi @fritzo, thanks for the nudge to get started on this and yes, let's collaborate to work out the best solution!

stefanwebb:
Both of those idioms seem possible. I've added a code snippet to the design doc.

stefanwebb:
@fritzo see some preliminary notes in the design doc

fritzo:
@stefanwebb thanks for adding details to the design doc!

It seems to me that a `ConditionalTransform` is not a transform at all but rather an indexed transform or a transform factory. From a type theoretic perspective you're lifting `Distribution` and `Transform` to a context. From that perspective it seems parsimonious to me to create separate classes. We would start with types for conditional things
```py
class ConditionalDistribution(object):
        @abstractmethod
        def in_context(self, context):
        """:rtype: torch.distributions.Distribution"""
        raise NotImplementedError

class ConditionalTransform(object):
    @abstractmethod
    def in_context(self, context):
        """:rtype: torch.distributions.Transform"""
        raise NotImplementedError
```
then add a way to treat unconditional things as trivially conditional things
```py
class ConstantConditionalDistribution(ConditionalDistribution):
    def __init__(self, base_dist):
        assert isinstance(base_dist, torch.distributions.Distribution)
        self.base_dist = base_dist
    def in_context(self, context):
        return self.base_dist

class ConstantConditionalTransform(ConditionalTransform):
    def __init__(self, transform):
        assert isinstance(transform, torch.distributions.Transform)
        self.transform
    def in_context(self, context):
        return self.transform
```
and finally lift `TransformedDistribution` using those pieces
```py
class ConditionalTransformedDistribution(ConditionalDistribution):
    def __init__(self, base_dist, transforms):
        assert isinstance(base_dist, ConditionalDistribution)
        assert all(isinstance(t, ConditionalTransform) for t in transforms)
        self.base_dist = base_dist
        self.transforms = transforms
    def in_context(self, context):
        base_dist = self.base_dist.in_context(context)
        transforms = [t.in_context(context) for t in self.transforms]
        return TransformedDistribution(base_dist, transforms)
```
Here's how you'd use this
```py
base_dist = dist.Normal(...)
assert isinstance(base_dist, Distribution)
flow = ConditionalPlanarFlow(...)
assert isinstance(flow, ConditionalTransform)
d = ConditionalTransformedDistribution(
    ConstantConditionalDistribution(base_dist), [flow])
...
pyro.sample("foo", d.in_context(x), obs=y)
```

stefanwebb:
@fritzo that's a brilliant design! I've had a go at implementing it and give a demonstration for the planar transformation. I was able to refactor it so that the conditional and non-conditional versions share most of the code.

What do you think of the changes? If you're happy with it, I'll go ahead and make the docstrings for the new classes

pyro/distributions/transforms/planar.py
```diff
@@ -7,12 +7,67 @@
...
 import torch.nn.functional as F
 
+from pyro.distributions.conditional import ConditionalTransform
 from pyro.distributions.torch_transform import TransformModule
 from pyro.distributions.util import copy_docs_from
 
 
 @copy_docs_from(TransformModule)
-class PlanarFlow(TransformModule):
+class ConditionedPlanarFlow(TransformModule):
```
fritzo: I think you'll the `ConditionalPlanarFlow` class to be an `nn.Module` and this class to be a simple `Transform`, right? I guess that means adding a `ConditionalTransformModule` class?

stefanwebb: I think you're right about this. I'll give it a go making that change. This would mean that `ConditionalTransformedDistribution` would need to be an `nn.Module` as well, just like `TransformedDistribution` is, so you can get the parameters out:

`>>> plf_dist = dist.ConditionalTransformedDistribution(base_dist, [plf])`
`>>> plf_dist.parameters()`

Otherwise there wouldn't be any way to get the parameters out as after making this change the following wouldn't work:

`>>> plf_dist = dist.ConditionalTransformedDistribution(base_dist, [plf]).condition(z)`
`>>> plf_dist.parameters()`

fritzo: Let's separate the two classes, as we do with `Teansform` and `TransformModule`:
```py
class ConditionalTransform(object):
    ...
class ConditionalTransformModule(ConditionalTransform, torch.nn.Module):
```
This avoids forcing a stateful style, and gives user freedom to choose purely functional style.

stefanwebb: @fritzo I think I've addressed this comment so it's ready for you to look at again! Apologies for the delay

stefanwebb:
When this gets approved I will start doing the conditional versions of the existing transformations

stefanwebb:
With the conditional transforms plus a simple tutorial the first version of this normalizing flow sub-library will be complete

pyro/distributions/conditional.py
```diff
@@ -0,0 +1,63 @@
+from abc import abstractmethod
+
+import torch
+import torch.nn
+
+
+class ConditionalDistribution(object):
```
fritzo: I think you'll want to inherit from `ABC` rather than `object`, here and in all other abstract classes
```py
from abc import ABC, abstractmethod

class ConditionalDistribution(ABC):
    ...
```

pyro/distributions/conditional.py
```diff
@@ -0,0 +1,63 @@
...
+    @abstractmethod
+    def condition(self, context):
+        """:rtype: torch.distributions.Transform"""
+        raise NotImplementedError
+
+
+class ConditionalTransformModule(ConditionalTransform, torch.nn.Module):
+    """
+    Conditional transforms with learnable parameters such as normalizing flows should inherit from this class rather
+    than `ConditionalTransform` so they are also a subclass of `nn.Module` and inherit all the useful methods of that
```
fritzo: nit: add a sphinx link via
```
than :class:`ConditionalTransform` so they are ...
```

pyro/distributions/conditional.py
stefanwebb: Fixed!

pyro/distributions/conditional.py
fritzo: I believe you need to fully qualify references to classes in different modules, but you can tell sphinx to hide the qualification with a tilde:
```
blah blah :class:`~pyro.distributions.conditional.ConditionalTransform` blah blah
```

pyro/distributions/transforms/planar.py
```diff
@@ -53,55 +109,73 @@ class PlanarFlow(TransformModule):
...
-        sample from the base distribution (or the output of a previous flow)
-        """
+        :math:`\\mathbf{y} = \\mathbf{x} + \\mathbf{u}\\tanh(\\mathbf{w}^T\\mathbf{z}+b)`
 
-        y = x + self.u_hat() * torch.tanh(self.lin(x))
-        return y
+    where :math:`\\mathbf{x}` are the inputs with dimension :math:`D`, :math:`\\mathbf{y}` are the outputs,
+    and the pseudo-parameters :math:`b\\in\\mathbb{R}`, :math:`\\mathbf{u}\\in\\mathbb{R}^D`, and
+    :math:`\\mathbf{w}\\in\\mathbb{R}^D` are the output of a function, e.g. a NN, with input
+    :math:`z\\in\\mathbb{R}^{M}` representing the observed variable to condition on. For this to be an
```
martinjankowiak: ditto

pyro/distributions/transforms/planar.py
```diff
@@ -53,55 +109,73 @@ class PlanarFlow(TransformModule):
...
-        Inverts y => x. As noted above, this implementation is incapable of inverting arbitrary values
-        `y`; rather it assumes `y` is the result of a previously computed application of the bijector
-        to some `x` (which was cached on the forward call)
-        """
+    Example usage:
 
-        raise KeyError("PlanarFlow expected to find key in intermediates cache but didn't")
+    >>> from pyro.nn.dense_nn import DenseNN
+    >>> input_dim = 10
+    >>> observed_dim = 5
```
martinjankowiak: ditto

pyro/distributions/transforms/planar.py
```diff
@@ -53,55 +109,73 @@ class PlanarFlow(TransformModule):
...
+    >>> batch_size = 3
+    >>> base_dist = dist.Normal(torch.zeros(input_dim), torch.ones(input_dim))
+    >>> hypernet = DenseNN(observed_dim, [50, 50], param_dims=[1, input_dim, input_dim])
+    >>> plf = ConditionalPlanarFlow(hypernet)
+    >>> z = torch.rand(batch_size, observed_dim)
```
martinjankowiak: ditto

pyro/distributions/conditional.py
stefanwebb: Okay, made that change!

stefanwebb:
Okay, I've made the requested change to the docstrings :)

PR #1938, opened 2019-07-01
Add @weakmethod decorator to avoid memory leaks

fritzo:
This adds a `@weakmethod` decorator for classes that store a bound copy of one of their methods.

The issue happens when an object `foo` is initialized with some state `bar` that needs a function, and that function is passed as a bound method `foo.method`. This creates a reference cycle: `foo` -> `bar` -> `foo.method` -> `foo`. The `@weakmethod` decorator breaks the final reference.

The only existing leaks I've found are in testing distributions. I'll also need this decorator in #1370 (see NUTS snippet below).

I'd happily replace this PR with a more standard solution if anyone knows one 🤔 

## Tested

- added a regression test that fails without `@weakmethod`
- tested in an example using the pattern:
    ```py
    class ABigInferenceAlgorithm(object):
        def __init__(self, ...):
            self.nuts = NUTS(potential_fn=self.potential_fn, ...)
            ...

        # I can't use @staticmethod here since this depends on dynamic state in self
        @weakmethod
        def potential_fn(self, params):
            ...
    ```

### DIFF
--- a/pyro/distributions/testing/rejection_exponential.py
+++ b/pyro/distributions/testing/rejection_exponential.py
@@ -5,7 +5,7 @@ from torch.distributions.utils import broadcast_all
 
 from pyro.distributions.rejector import Rejector
 from pyro.distributions.torch import Exponential
-from pyro.distributions.util import copy_docs_from
+from pyro.distributions.util import copy_docs_from, weakmethod
 
 
 @copy_docs_from(Exponential)
@@ -17,6 +17,7 @@ class RejectionExponential(Rejector):
         log_scale = self.factor.log()
         super(RejectionExponential, self).__init__(propose, self.log_prob_accept, log_scale)
 
+    @weakmethod
     def log_prob_accept(self, x):
         result = (self.factor - 1) * self.rate * x
         assert result.max() <= 0, result.max()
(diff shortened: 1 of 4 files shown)


fritzo:
Let me see if I can simplify this using `__get__` and `__set__` rather than `@property`...

fritzo:
Ok, I give up with `__get__` and `__set__` :smile:

neerajprad:
Wow..this is a tricky issue. The `weakmethod` decorator looks great to me. I was trying to play around with `__get__` and the following works for your test case, though I'm not sure if that's general enough for your use case:

```python
class _Method(object):
    def __get__(self, instance, owner):
        return instance._method


def test_weakmethod():
    class Foo(object):
        method = _Method()

        def __init__(self, state):
            self.state = state

        def _method(self, *args, **kwargs):
            return self.state, args, kwargs

    foo = Foo(42)

    foo_ref = weakref.ref(foo)
    assert foo_ref() is foo
    del foo
    assert foo_ref() is None

```

fritzo:
I also tried, but needed to handle `__set__` to handle the Rector use cases where the base class tries to overwrite the bound method with a bound copy (my implementation simply checks that it is a no-op and does nothing). Do you want me to continue playing with a `__get__`, `__set__` version, or are you ok with the `property` version?

neerajprad:
> Do you want me to continue playing with a __get__, __set__ version, or are you ok with the property version?

The existing version looks great to me actually. I was only trying to fiddle around and learn more about `__get__` / `__set__`. We can always refactor the logic inside the decorator later, but it seems very clean to me and I'm not sure if any alternatives would be better.

fritzo:
Thanks for reviewing this tricky PR @neerajprad ! I didn't even know about `__get__` and `__set__` until reviewing your #1940 . Let's try to learn more.

PR #1940, opened 2019-07-01
Allow for pickling of poutine wrapped models

### LINKED ISSUE #1929: Make potential energy function picklable
Related: #1930. To get multiprocessing to work correctly with the new API, we will need to make the potential function returned by `_pe_maker` to be picklable.

neerajprad:
Fixes #1929. 

This sketches one possible way to serialize poutine wrapped models via pickle. Suggestions on alternative schemes are welcome.

Summary of changes:
 - [x] Replaces `Messenger.__call__._wraps` with a picklable version. This required defining `_bound_partial` in addition to using `functools.partial` so as to allow for pickling of models that are defined as bound methods. This was needed to have the `test_poutines::test_method_decorator_interface_condition` pass.
 - [x] Changes other common handlers to use `functools.partial` instead of closures.
 - [x] Our transforms aren't picklable due to the use of `weakref.ref` when caching inverse. I have currently patched `transforms` to ignore these when pickling.

HMC related change:
 - [x] Changing _pemaker to a class, and ensuring that compilation is not exercised during call to `initialize_model` otherwise pickle will not be able to serialize the compiled `torch._C.Function` object. In HMC, the compilation will happen inside the child processes.

Tests added for poutine serialization as well as potential energy serialization.


fritzo:
Nice @neerajprad !

> otherwise pickle will not be able to serialize the compiled torch._C.Function object

Does it work to use `torch.save()` instead of `pickle.dump()`?

pyro/poutine/messenger.py
```diff
@@ -22,11 +40,8 @@ def __init__(self):
         pass
 
     def __call__(self, fn):
-        def _wraps(*args, **kwargs):
-            with self:
-                return fn(*args, **kwargs)
-        _wraps.msngr = self
-        return _wraps
+        wraps = _bound_partial(partial(_context_wrap, self, fn))
```
fritzo: Why not simply `_bound_partial(_context_wrap, self, fn)`?

pyro/infer/mcmc/util.py
```diff
@@ -233,18 +233,47 @@ def _guess_max_plate_nesting(model, args, kwargs):
...
+            names, vals = zip(*sorted(params.items()))
+
+            def _pe_jit(*zi):
+                params = dict(zip(names, zi))
+                return self._potential_fn(params)
+
+            self._compiled_fn = torch.jit.trace(_pe_jit, vals, **self.jit_options)
+            return self._compiled_fn(*vals)
+
+    def get_potential_fn(self):
```
fritzo: nit: It seems a tad simpler to pass `jit_compile` and `jit_options` to this method, then below call
```py
pe_maker = _PEMaker(...)
init_params = _get_init_params(..., pe_maker.get_potential_fn(jit_compile=False))
potential_fn = pe_maker.get_potential_fn(jit_compile=jit_compile, jit_options=jit_options)
return init_params, potential_fn, transforms, model_trace
```

pyro/poutine/messenger.py
neerajprad: For this to work, I will need to change `bound_partial` to 

```
return partial(partial(self.func, *self.args, **self.keywords), instance)
```

from

```
return partial(self.func, instance)
```

The reason is that instance should be bound in that case not to `self.func` but rather to the `fn` argument of `_context_wrap`. Maybe there is a simpler way to do this that I am missing out. Let me know if you want me to make this change.

fritzo: ok, thanks for explaining, this is subtler than I thought!

neerajprad:
Thanks for reviewing!

PR #1941, opened 2019-07-03
Using new MCMC API in existing examples

neerajprad:
**Code changes:**

 - Added a predictive and diagnostics function that can work with MCMC samples dictionary and doesn't rely on bag of traces, allowing for deprecating `TracePosterior` and `TracePredictive` in the future.
 - Changed docs, tests and tutorials to use the new API.

**TODOs:**

 - [x] Implement `predictive` function for MCMC which is similar to TracePredictive except that it takes in posterior samples directly instead of traces and is vectorized.
 - [x] Implement diagnostics so that the MCMCMarginals class is deprecated.
 - [x] Migrate all tutorials / examples to the new API. 
 - [x] Add `PendingDeprecationWarning` to TracePredictive. This should be deprecated as soon as we have a viable predictive utility for the SVI class.  
 - [x] Add `DeprecationWarning` to `MCMCMarginals`.


pyro/util.py
```diff
@@ -380,6 +380,17 @@ def optional(context_manager, condition):
         yield
 
 
+class ExperimentalWarning(UserWarning):
+    pass
```
neerajprad: @fritzo, @fehiepsi - I have introduced this warning for experimental interfaces, in this case the `predictive` function. Does this pattern seem okay?

fehiepsi: It is ok to me.

examples/baseball.py
```diff
@@ -235,78 +257,95 @@ def main(args):
...
+                                                                num_chains=args.num_chains)
+    nuts_kernel = NUTS(potential_fn=potential_fn)
+    mcmc = MCMC(nuts_kernel,
+                num_samples=args.num_samples,
+                warmup_steps=args.warmup_steps,
+                num_chains=args.num_chains,
+                initial_params=init_params,
+                transforms=transforms)
+    num_chains = mcmc.num_chains  # Note that this might be less than args.num_chains  when CPU cores are limited.
+    samples_fully_pooled = mcmc.run(at_bats, hits)
```
neerajprad: @fehiepsi - We will need to change all these examples / tests if we don't want `.run(..)` to return the dict of samples as discussed in #1942. I'm happy to do that in a separate PR.

examples/baseball.py
```diff
@@ -235,78 +257,95 @@ def main(args):
...
+    logging.info(summary(samples_partially_pooled_logit,
                          sites=["alpha"],
                          player_names=player_names,
-                         transforms={"alpha": lambda x: 1. / (1 + (-x).exp())})["alpha"])
-    posterior_predictive = TracePredictive(partially_pooled_with_logit,
-                                           posterior_partially_pooled_with_logit,
-                                           num_samples=num_predictive_samples)
-    sample_posterior_predictive(posterior_predictive, baseball_dataset)
-    evaluate_log_predictive_density(posterior_predictive, baseball_dataset)
+                         transforms={"alpha": lambda x: 1. / (1 + (-x).exp())},
```
fehiepsi: nit: I think we can use `lambda x: torch.sigmoid(x)` or `torch.sigmoid`

pyro/infer/mcmc/util.py
```diff
@@ -373,3 +377,132 @@ def initialize_model(model, model_args=(), model_kwargs={}, transforms=None, max
...
+        """
+
+        def wrapped_fn(*args, **kwargs):
+            with pyro.plate("_num_predictive_samples", num_samples, dim=-max_plate_nesting-1):
+                return fn(*args, **kwargs)
+
+        return wrapped_fn
+
+    trace = poutine.trace(poutine.condition(_vectorized_fn(model), posterior_samples))\
+        .get_trace(*args, **kwargs)
```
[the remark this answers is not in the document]
fehiepsi: Btw, this trick is pretty neat!

tests/infer/mcmc/test_mcmc_util.py
```diff
@@ -0,0 +1,39 @@
...
+                                                                model_args=(data,))
+    nuts_kernel = NUTS(potential_fn=potential_fn, transforms=transforms)
+    samples = MCMC(nuts_kernel,
+                   100,
+                   initial_params=init_params,
+                   warmup_steps=100).run(data)
+    with ignore_experimental_warning():
+        predictive_samples = predictive(model, samples,
+                                        return_sites=["beta", "obs"],
+                                        num_samples=200)
```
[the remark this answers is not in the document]
neerajprad: Done.

tests/infer/mcmc/test_mcmc_util.py
```diff
@@ -0,0 +1,47 @@
...
+    def model(data=None):
+        with pyro.plate("num_components", 5):
+            beta = pyro.sample("beta", dist.Beta(1., 1.))
+            with pyro.plate("data", N):
+                pyro.sample("obs", dist.Bernoulli(beta), obs=data)
+
+    return model, data, true_probs
+
+
+@pytest.mark.parametrize("num_samples", [100, 200, None])
```
fehiepsi: It seems that `num_samples` does not play any role here. Could you please revise this test?

neerajprad: Thanks for catching this; fixed!

pyro/distributions/torch_patch.py
```diff
@@ -53,7 +53,9 @@ def _Transform__init__(self, cache_size=0):
 def _Transform__getstate__(self):
     attrs = {}
     for k, v in self.__dict__.items():
-        if not isinstance(v, weakref.ref):
+        if isinstance(v, weakref.ref):
```
neerajprad: I realized that it is important to serialize this when I ran the baseball example on CUDA using multiprocessing, so that `transform._inv` is None and recomputed but doesn't raise an AttributeError.

Issue #1931, opened 2019-06-29
[FR] support divergence diagnostics for NUTS


fehiepsi:
Based on @riversdark [request](https://github.com/pyro-ppl/pyro/issues/1093#issuecomment-506943294), it would be convenient for users to be able to access to divergence information. Let's support that feature.

robsalomone:
@fehiepsi Straightforward divergence counts should be simple (you just keep count of how often the energy is not conserved within a given tolerance, and ideally flag those samples to identify problem regions).  However, there is an  additional diagnostic (and more discussion about divergences) here
https://arxiv.org/abs/1604.00695

These are also in Stan (and only know to more advanced users). It could be worth implementing this as well.

fehiepsi:
Thanks @robsalomone ! We might try to have the simple case first. To be honest, I don't know what `diverging` means in Stan and PyMC3 result. I thought that we just need to calculate the number of [diverging at this line](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/mcmc/nuts.py#L34) for NUTS algorithm, that is the number of trajectories which end up with `delta_energy = new_energy - old_energy > 1000`. Do you know if that number matches `Stan/PyMC3` divergence diagnostic? In addition, is there also a `diverging` notion in HMC?

robsalomone:
That’s correct @fehiepsi, it’s just used to flag areas of problematic curvature. Divergences usually indicate the chain is unlikely to be mixing well.  The precise number is in the Stan documentation somewhere I believe, but more or less it just needs to be some large number that would indicate the HMC integrator is failing (flies off due to overly steep gradient).

fehiepsi:
@robsalomone Thanks for explaining! It is very helpful to me. I just find the number in the footnote of the [manual](https://mc-stan.org/docs/2_18/reference-manual/divergent-transitions.html). :)

neerajprad:
Handled by @fehiepsi in #1942.

Issue #1934, opened 2019-06-30
Print summary table in MCMC


fehiepsi:
It would be nice to print summary table after running MCMC. E.g.

```
                           mean         sd       5.5%      94.5%      n_eff       Rhat
  emission_prob[0,0]       0.00       0.01       0.00       0.01     599.79       1.00
  emission_prob[0,1]       0.00       0.00       0.00       0.00     962.09       1.00
  emission_prob[0,2]       0.01       0.02       0.00       0.01     605.00       1.00
  emission_prob[0,3]       0.49       0.09       0.33       0.61     699.05       1.00
  emission_prob[0,4]       0.00       0.01       0.00       0.01     776.91       1.00
  emission_prob[0,5]       0.41       0.10       0.25       0.56     686.47       1.00
  emission_prob[0,6]       0.01       0.02       0.00       0.02     595.12       1.00
  emission_prob[0,7]       0.01       0.03       0.00 
```

PR #1942, opened 2019-07-04
Add diverging diagnostics for HMC/NUTS

fehiepsi:
Addresses #1931, this PR adds diverging diagnostics to HMC/NUTS. The progbar will display the number of diverging trajectories when that number is greater than 0.

For illustration, please see the [gist](https://gist.github.com/fehiepsi/3827fbaeafc25a920ee98337e0057888).

cc @robsalomone

TODO:
- [x] add and fix tests


fehiepsi:
@neerajprad with this, should we change
```
with optional(pyro.validation_enabled(False), self._t < self._warmup_steps):
```
to always disable validation even in sampling phase, and users can rely on diverging diagnostics to detect if something is wrong? (instead of raising ValueError if `nan` happens)

robsalomone:
@fehiepsi Nice work but there is a subtle issue so I don't recommend merging yet. You should not halt construction of the tree once a divergence is observed, the chain will no longer satisfy detailed balance that way. You need to let it continue until it hits the maximum tree depth. Divergence is just something you watch for to inform the user, it shouldn't affect the MCMC. 

Now, when you diverge the probability of all states from the start until the maximum tree depth in the resampling step are likely to be very small (unlikely to be chosen) due to the divergence, but they technically still matter, and there is a always chance it could diverge the energy back upwards in some cases. 

So, for these reasons, I recommend you let NUTS build the tree to the max depth.

robsalomone:
Of course, if you get NaNs due to overflow etc. there's not much you can do, but in general you want to avoid deviating from the NUTS procedure. I would suggest flagging NaNs separate to divergences. I suspect most divergences will not be numerical issues.

fehiepsi:
Hi @robsalomone, AFAIK diverging and turning are two conditions (formula [8] of NUTS paper) for stop doubling. And detailed balance is satisfied with the biased transition kernel (formula [10] of NUTS paper), which does not require a full tree to calculate the probability of a proposal. Those stopping conditions make NUTS efficiently in that: it won't waste computation resource to build full tree. Unlike HMC, where we need to specify a trajectory length, NUTS does not require max_tree_depth. Typically, we want to keep doubling the tree until those stopping conditions happen (indeed, there is no thing like `max_tree_depth` in the algorithms in NUTS paper). But as you know, a tree with depth 10, which requires ~1000 leap frog steps, is already expensive to build. So in practice, we need to set a threshold (max_tree_depth) to prevent the tree grow too large.

pyro/infer/mcmc/hmc.py
```diff
@@ -304,12 +306,13 @@ def sample(self, params):
...
-        # Set accept prob to 0.0 if delta_energy is `NaN` which may be
-        # the case for a diverging trajectory when using a large step size.
-        if torch_isnan(delta_energy):
-            accept_prob = scalar_like(delta_energy, 0.)
-        else:
-            accept_prob = (-delta_energy).exp().clamp(max=1.)
+        # handle the NaN case which may be the case for a diverging trajectory
+        # when using a large step size.
+        delta_energy = scalar_like(delta_energy, float("inf")) if torch_isnan(delta_energy) else delta_energy
+        if delta_energy > self._max_sliced_energy and self._t >= self._warmup_steps:
```
neerajprad: Just want to confirm - the 1000 factor is being applied in log space, so the actual difference in energy is much higher?

pyro/infer/mcmc/nuts.py
```diff
@@ -352,7 +352,13 @@ def sample(self, params):
                 sum_accept_probs = sum_accept_probs + new_tree.sum_accept_probs
                 num_proposals = num_proposals + new_tree.num_proposals
 
-                if new_tree.turning or new_tree.diverging:  # stop doubling
+                # stop doubling
+                if new_tree.diverging:
+                    if self._t >= self._warmup_steps:
+                        self._num_diverging += 1
```
neerajprad: This diagnostic information will be really useful. I think users also like to plot their posterior samples and see where the divergences lie. To support that, we will also need to annotate the samples somehow for later analysis. This is much easier to do in NumPyro by adding to the named tuple data structure, but I wonder what will be the best way to do this in Pyro. Maybe we can store it in a separate list inside the HMC/NUTS kernels, called `kernel._divergences`. What do you think?

fehiepsi: Sound reasonable for me! For now, I'll go with your proposal and will think how to let users access it with multi chains.

pyro/infer/mcmc/hmc.py
fehiepsi: I guess you mean actual difference is accept prob? It's pretty small `exp(-1000) ~ 0`.

pyro/infer/mcmc/api.py
```diff
@@ -310,9 +316,19 @@ def __init__(self, kernel, num_samples, warmup_steps=None, initial_params=None,
...
+            if isinstance(sample, dict):
+                for k, v in sample.items():
+                    z_acc[k][chain_id].append(v)
+            else:  # diverging info
+                divergences[chain_id] = sample
+
+        diverging_mask = torch.zeros((self.num_chains, self.num_samples), dtype=torch.bool)
+        for i in range(self.num_chains):
+            diverging_mask[i][divergences[i]] = True
+        self._diverging_mask = diverging_mask
```
fehiepsi: @neerajprad Users can get the mask for diverging samples here. Could you suggest me a pattern to expose this information for user? I think we can make this a property or create a method returning this value. I would like to propose the following pattern:

Make this MCMC class have some methods as follows
+ `run` returns `self` as in the old API
+ `marginal` returns samples as in the old API
+ `diagnostics` returns diagnostics as in the old API with additional `diverging` (or better name) field
+ `summary` prints summary table (which I'll address in #1934)

robsalomone:
@fehiepsi that’s good I just wanted to make sure you considered this. I was a bit hazy on the specifics of NUTS. That said, if you do the trajectory sampled version of regular HMC (as in Michael Betancourts conceptual introduction paper), you can not stop once a divergence is observed. Of course, that is a different case entirely.

fehiepsi:
@robsalomone You are right. In HMC we won't stop when a divergence is observed. Indeed, we don't have a mechanism to catch those divergences (diverging samples at the middle of a trajectory) so in HMC, I only report a trajectory is diverged when energy of the last sample diverges from energy of the start sample (of that trajectory).

pyro/infer/mcmc/hmc.py
neerajprad: Sorry, I was mistaken. The energy isn't in the log space - this is fine!

pyro/infer/mcmc/api.py
neerajprad: I largely agree with your suggestions. My thoughts are as follows:
 - Let us not have `.run` return anything. I think if a function / method both sets states and returns something, it usually causes confusion.
 - `.marginal(flatten=True)` can return samples where the chain dim is flattened out by default.
 -  I think `diagnostics`, and `summary` will be great to add as well. 
 - Also, let us consider renaming `kernel._diagnostics` to `kernel._logging` or something, and `kernel.post_sampling()` to `kernel.diagnostics()` since diagnostics usually refers to post sampling convergence related diagnostics.

pyro/infer/mcmc/api.py
```diff
@@ -310,9 +313,19 @@ def __init__(self, kernel, num_samples, warmup_steps=None, initial_params=None,
 
     def run(self, *args, **kwargs):
         z_acc = defaultdict(lambda: [[] for _ in range(self.num_chains)])
+        divergences = [None for _ in range(self.num_chains)]
         for sample, chain_id in self.sampler.run(*args, **kwargs):
-            for k, v in sample.items():
-                z_acc[k][chain_id].append(v)
+            if isinstance(sample, dict):
```
neerajprad: It will be nice to keep the interface as decoupled from the HMC kernels as possible (I know that we have some HMC specific checks in init, but better to avoid this as much as possible). My suggestion would be to instead do something like:

```python
num_samples = [0] * args.num_chains
for x, chain_id in self.sampler.run(*args, **kwargs):
    if num_samples[chain_id] = self.num_samples:
        self._diagnostics[chain_id] = x
    else:
        for k, v in x.items():
            z_acc[k][chain_id].append(v)
             ...
```

Here, diagnostics itself can be a dict and for HMC/NUTS it might look like `{"divergence_ids" = torch.Tensor([2, 10, 14, ...])}`. 

Then as you suggested `MCMC.diagnostics` property can combine these chain level information stored in `._diagnostics` and return some dict of diagnostic information per sample site.

fehiepsi: Yup, I expected that you will have a better idea for this issue. And I was right. :D Thanks!

pyro/infer/mcmc/api.py
```diff
@@ -310,9 +313,19 @@ def __init__(self, kernel, num_samples, warmup_steps=None, initial_params=None,
...
+                for k, v in sample.items():
+                    z_acc[k][chain_id].append(v)
+            else:  # diverging info
+                divergences[chain_id] = sample
+
+        diverging_mask = torch.zeros((self.num_chains, self.num_samples), dtype=torch.bool)
```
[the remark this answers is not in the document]
fehiepsi: ~The indices do not play nicely with multiple chains unless we store these indices in list.~ Returning the mask for users also makes it easier to work with flatten (no chain info) samples. Some other frameworks such as PySan and ArviZ also work with the diverging mask. *Edit*: got your point now. I'll think which way is more convenient.

pyro/infer/mcmc/hmc.py
```diff
@@ -325,7 +328,13 @@ def sample(self, params):
         return z.copy()
 
     def diagnostics(self):
-        return OrderedDict([
+        diagnostics = OrderedDict([
             ("step size", "{:.2e}".format(self.step_size)),
-            ("acc. rate", "{:.3f}".format(self._accept_cnt / self._t))
+            ("acc. rate", "{:.3f}".format(self._accept_cnt / self._t)),
         ])
+        if self._divergences:
```
neerajprad: My suggestion would be to not add divergence to the progress bar, since this will be available as diagnostic information later. Right now, our progress bar becomes garbled if the terminal width is small and this will further make this limit even larger, forcing the users to have to maximize their terminal to get reasonable looking output. If there is some way we can cram this information within the same number of columns, I'm all for it! :)

fehiepsi: yup, since we have a diagnostics method, there is no need to expose this in progbar. Let me change it together with a few other points when #1941 is ready.

tests/infer/mcmc/test_mcmc.py
```diff
@@ -104,8 +102,7 @@ def _empty_model():
 @pytest.mark.parametrize("jit", [False, True])
 @pytest.mark.parametrize("num_chains", [
     1,
-    skipif_param(2, condition="CI" in os.environ or "CUDA_TEST" in os.environ,
-                 reason="CI only provides 1 CPU; also see https://github.com/pytorch/pytorch/issues/2517")
```
fehiepsi: CUDA_TEST pass in my system

pyro/infer/mcmc/hmc.py
```diff
@@ -323,8 +326,11 @@ def sample(self, params):
...
 
-    def diagnostics(self):
+    def logging(self):
         return OrderedDict([
             ("step size", "{:.2e}".format(self.step_size)),
             ("acc. rate", "{:.3f}".format(self._accept_cnt / self._t))
         ])
+
+    def diagnostics(self):
+        return {"divergences": self._divergences}
```
neerajprad: Just to confirm - this will return a python list of ints that can be used for indexing into the samples?

fehiepsi: Yes. It seems to me that returning a mask makes it harder to see directly how many and which samples come from diverging transition. So I follow your suggestion to only returning indices. There might be other useful diagnostic infos so I don't aggregate chain_idx with sample_idx.

fehiepsi:
> Do you think we should change the interface and add methods for summary, marginal etc. (in a separate PR)?

Yes, I think that would be useful. Typically, researchers would like to extract (resampled) samples from these `exec_samples` for prediction. Despite that we have `ops.stats.resample` method, that function only works for 1 site. So having a method here would be useful. About summary, maybe having an arg `print_summary=True` at `diagnostics` method is enough. WDYT?

fehiepsi:
Btw, should we address `run` method to not return `samples`? We can make `exec_samples` a property (exposed in docs) if we do that way, or having a `marginal` method would also be good to me. What I think would be very useful is to have an `extract_samples` (or `sample_posterior`) method to get random samples from a MCMC posterior, a fitted autoguide, or a fitted SVI. That would also make `mcmc.util.predict` function pretty handy. :)

neerajprad:
> Btw, should we address run method to not return samples? 

I think it makes for users to hold a reference to MCMC directly since there are multiple methods that they might call on it like `.diagnostics()` so it definitely makes sense to not have `.run` return samples. 

> What I think would be very useful is to have an extract_samples (or sample_posterior) method to get random samples from a MCMC posterior, a fitted autoguide, or a fitted SVI.

Could you elaborate on what you mean by random samples - do you mean resampling with replacement from the set of generated samples? I think this will be useful. I think it makes sense to have a `.marginal` method (maybe with `flatten=True`) as well as `.sample_posterior` method.

tests/infer/mcmc/test_mcmc.py
neerajprad: I am still seeing CUDA initialization error for this test with `CUDA_TEST=1 PYRO_TENSOR_TYPE=torch.cuda.DoubleTensor pytest -vs -k test_empty_sample_sites` with the latest pytorch.

fehiepsi: Sorry, you are right. That is my mistake. I did use only `CUDA_TEST=1`. We'd need `PYRO_TENSOR_TYPE=torch.cuda.DoubleTensor` too. >"<

fehiepsi: I'll revert it back in the last PR.

Issue #1304, opened 2018-08-13
Implement Sequential Monte Carlo


fritzo:
This issue tracks progress towards implementing Sequential Monte Carlo aka Particle Filtering.

For general control flow this appears to be difficult due to Python's lack of forkable coroutines.

For static computation graphs we may be able to implement tensorized forking via sparse tensors.

jpchen:
may be useful but extremely outdated: @OptimusLime 's [implementation](https://github.com/uber/pyro/tree/nightmare-poutine) using OS forking iirc

ndepalma:
I'd love this feature. Maybe you could start small with a limited example (non-language support). I'm currently evaluating some of these probabilistic languages and it looks like only Gen has support for SMC. 

I'd be glad to help you all out if you're currently working on an implementation.

PR #1947, opened 2019-07-08
Add SMCFilter class; Add simple model example that uses SMC Filter

chanjed:
Addresses #1304 
pair coded with @fritzo and @karalets 

This PR:
1. Implements SMCFilter class
2. Simple State Space model example to run SMCFiltering on.

## Tasks
- [x] add examples/smc.py
- [x] add a file tutorial/source/smc.rst following the pattern of tutorial/source/hmm.py, and register tutorial/source/smc.rst in tutorial/source/index.rst
- [x] add a docstring to `SMCFilter` as in other inference algorithms
- [x] add an `SMCFilter` section to docs/source/inference_algos.rst
- [x] add smoke tests with different numbers of sample sites and different `event_shape` and `batch_shape`. Tests should live in tests/infer/test_smcfilter.py
- [x] add a simple statistical test that evaluates correctness

## Traiged
- Implement Conjugate SMC Class (this should be in a separate PR)

### DIFF
--- a/docs/source/inference_algos.rst
+++ b/docs/source/inference_algos.rst
@@ -59,6 +59,14 @@ Importance
     :undoc-members:
     :show-inheritance:
 
+Sequential Monte Carlo
+----------------------
+
+.. automodule:: pyro.infer.smcfilter
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
 Discrete Inference
 ------------------
 
--- a/examples/smcfilter.py
+++ b/examples/smcfilter.py
@@ -1,0 +1,111 @@
+import argparse
+import logging
+
+import torch
+
+import pyro
+import pyro.distributions as dist
+from pyro.infer import SMCFilter
+
+logging.basicConfig(format='%(relativeCreated) 9d %(message)s', level=logging.INFO)
+
+"""
+This file demonstrates how to use the SMCFilter algorithm with
+a simple model of a noisy harmonic oscillator of the form:
+
+    z[t] ~ N(A*z[t-1], B*sigma_z)
+    y[t] ~ N(z[t][0], sigma_y)
+
+"""
+
+
+class SimpleHarmonicModel:
+
+    def __init__(self, process_noise, measurement_noise):
+        self.A = torch.tensor([[0., 1.],
+                               [-1., 0.]])
+        self.B = torch.tensor([3., 3.])
+        self.sigma_z = torch.tensor(process_noise)
+        self.sigma_y = torch.tensor(measurement_noise)
+
+    def init(self, initial):
+        self.t = 0
+        self.z = initial
+        self.y = None
+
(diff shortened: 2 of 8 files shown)


examples/smc.py
```diff
@@ -0,0 +1,97 @@
+from __future__ import absolute_import, division, print_function
+
+import argparse
+
+import torch
+
+import pyro
+import pyro.distributions as dist
+import pyro.infer.smcfilter as smcfilter
```
fritzo: nit: can you `from pyro.infer import SMCFilter` and use `SMCFilter` below?

examples/smc.py
```diff
@@ -0,0 +1,97 @@
+from __future__ import absolute_import, division, print_function
```
fritzo: Can you add a little docstring at the top of this file, similar to those in other examples. Something like
```py
"""
This file demonstrates how to use the SMCFilter algorithm with a simple model of a noisy harmonic oscillator.
"""
```

tuananhle7:
Hi @chanjed, @fritzo and @karalets,

Thank you for implementing this feature.

I was looking at [this example](https://github.com/pyro-ppl/pyro/blob/dev/examples/smcfilter.py) and the [source code](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/smcfilter.py) in order to understand the implementation.

I don't understand how resampling works:

Assume that at time `t - 1`, we have particles `{z_(t - 1)k}` and weights `{w_(t - 1)k}` (`k = 1, ..., K`) which I think (?) are stored in [`_values`](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/smcfilter.py#L42) and [`_log_weights`](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/smcfilter.py#L43).

The [resampling step](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/smcfilter.py#L73) produces ancestral indices `a_(t - 1)k` (`k = 1, ..., K`) which are marginally distributed according to a Categorical distribution parameterized by the normalized weights ([here](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/smcfilter.py#L123) I think). Then `z_(t - 1)k` which is stored in `_values` is updated to be `z_(t - 1)(a_(t - 1)k)` [here](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/smcfilter.py#L124).

At time `t`, we should propose `z_tk` from `guide(z_t(k - 1))` where `z_t(k - 1)` is already resampled. However, [here](https://github.com/pyro-ppl/pyro/blob/dev/examples/smcfilter.py#L60-L61), `z_tk` is proposed from a guide which is conditioned on `self.z` (attribute of guide) which doesn't seem to interact with `_values` which stores the resampled values.

Thank you,
Tuan Anh

PR #1948, opened 2019-07-08
Use refresh=False for tqdm

fehiepsi:
This PR uses `refresh=False` for progressbar. The hanging issue does not only happen in my system but also a user in forum.

#### Tested:
+ one chain + terminal/notebook + CPU/GPU + old/new api
+ multi chain + CPU/GPU + terminal + old api
+ multi chain + CPU + notebook + old api

#### What is not tested:
+ multi chain + GPU + notebook: It is a bit tricky to run multi chains in GPU in notebook. New version of pytorch only supports `spawn`/`forkserver` (see [the docs](https://pytorch.org/docs/master/notes/multiprocessing.html#cuda-in-multiprocessing)). On the other hand, [it is tricky to use spawn method](https://github.com/pytorch/pytorch/issues/20375) in jupyter notebook. MCMC does not run when I use those methods with CUDA tensors in notebooks.

Anyway, the above issue does not related to this PR because progress bars work well in notebook with CPU tensors.

### DIFF
--- a/examples/eight_schools/mcmc.py
+++ b/examples/eight_schools/mcmc.py
@@ -46,7 +46,7 @@ def main(args):
 
 
 if __name__ == '__main__':
-    #assert pyro.__version__.startswith('0.3.3')
+    assert pyro.__version__.startswith('0.3.3')
     parser = argparse.ArgumentParser(description='Eight Schools MCMC')
     parser.add_argument('--num-samples', type=int, default=1000,
                         help='number of MCMC samples (default: 1000)')
--- a/examples/eight_schools/mcmc.py
+++ b/examples/eight_schools/mcmc.py
@@ -46,7 +46,7 @@ def main(args):
 
 
 if __name__ == '__main__':
-    assert pyro.__version__.startswith('0.3.3')
+    #assert pyro.__version__.startswith('0.3.3')
     parser = argparse.ArgumentParser(description='Eight Schools MCMC')
     parser.add_argument('--num-samples', type=int, default=1000,
                         help='number of MCMC samples (default: 1000)')
--- a/pyro/infer/mcmc/api.py
+++ b/pyro/infer/mcmc/api.py
@@ -122,7 +122,7 @@ def _add_logging_hook(logger, progress_bar=None, hook=None):
         diagnostics = json.dumps(kernel.diagnostics())
         logger.info(diagnostics, extra={"msg_type": DIAGNOSTIC_MSG})
         if progress_bar:
-            progress_bar.set_description(stage)
+            progress_bar.set_description(stage, refresh=False)
         if hook:
             hook(kernel, params, stage, i)
 
--- a/pyro/infer/mcmc/api.py
+++ b/pyro/infer/mcmc/api.py
@@ -55,7 +55,7 @@ def logger_thread(log_queue, warmup_steps, num_samples, num_chains, disable_prog
                 if num_samples[pbar_pos] == warmup_steps:
                     progress_bars.set_description("Sample [{}]".format(pbar_pos + 1), pos=pbar_pos)
                 diagnostics = json.loads(msg, object_pairs_hook=OrderedDict)
-                progress_bars.set_postfix(diagnostics, pos=pbar_pos)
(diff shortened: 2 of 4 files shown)


neerajprad:
Just to confirm - multi-chain with the new API as well works for both notebook (CPU) and terminal?

fehiepsi:
[tqdm docs](https://tqdm.github.io/docs/tqdm/) said that "if your progress is erratic with both fast and slow iterations (network, skipping items, etc) you should set miniters=1". This problem is [observed in tqdm](https://github.com/tqdm/tqdm/issues/249) with the solution is to tweak `miniters` and `mininterval`. I have played with those settings. They help somehow but the problem still happens.

Here is an example script which hangs when running in notebook
```
import torch
import pyro
from pyro.infer.mcmc.api import MCMC, NUTS

def model(sigma, y):
    eta = pyro.sample('eta', dist.Normal(torch.zeros(J), torch.ones(J)))
    mu = pyro.sample('mu', dist.Normal(torch.zeros(1), 10 * torch.ones(1)))
    tau = pyro.sample('tau', dist.HalfCauchy(scale=25 * torch.ones(1)))
    theta = mu + tau * eta
    return pyro.sample("obs", dist.Normal(theta, sigma), obs=y)

J = 8
y = torch.tensor([28,  8, -3,  7, -1,  1, 18, 12]).type(torch.Tensor)
sigma = torch.tensor([15, 10, 16, 11, 9, 11, 10, 18]).type(torch.Tensor)
pyro.set_rng_seed(0)
nuts_kernel = NUTS(model, jit_compile=True)
mcmc = MCMC(nuts_kernel, num_samples=1000, warmup_steps=1000, num_chains=1)
posterior = mcmc.run(sigma, y)
```

fehiepsi:
> Just to confirm - multi-chain with the new API as well works for both notebook (CPU) and terminal?

I just test the above example with new API. multi-chain in notebook (CPU) works well. So I think it will work well in terminal too.

neerajprad:
> Here is an example script which hangs when running in notebook

This still seems to work fine for me both for `num_chains = 1, 2`. My guess is that this is highly dependent on platform / random seed etc. What is the version of tqdm that you are using?

The issue that you linked relates to how frequently the progress bar should be updated. `tqdm` uses `miniters` and `mininterval` to decide this and when this is set dynamically (which is the default), the first few iterations will be used to learn this value. However, if the first few iterations are really slow but the remaining iterations are really fast, then the progress bar might not be updated until much later and the program may have to wait for this update if the loop has already terminated. With `refresh=True` we are practically forcing a refresh per sample, i.e. setting `miniter=1`, so this should actually fix such freezes. If setting `refresh=False` fixes the freezing issue instead, it probably points to another source of bug. 😕

fehiepsi:
For an independent of Pyro example, non-random repro script, could you take a look at
```
import time
import tqdm

def f(n):
    if n % 20 == 1 or n % 20 == 3 or n % 20 == 10:
        time.sleep(0.1)
        return "slow"
    else:
        time.sleep(0.01)
        return "fast"

t = tqdm.tqdm(total=2000)

for i in range(2000):
    postfix = OrderedDict(stat=f(i))
    t.set_postfix(postfix, refresh=True)
    t.update()
t.close()
```

This will hang my notebook after running weirdly from iteration 300. There is no way to stop it unless I use `kill -9 [pid]`. I use `tqdm=4.32.2`. Maybe this only happens in my linux system or a few other platforms. I don't think that `refresh=True` plays any role here when we already have `update()`, which will update the progress bar regardless whatever info we have updated.

I agree that this is a expected/unexpected tqdm issue. But rather than following up with this issue and waiting for a (possibly) fix in future version of tqdm, we can use `refresh=False`. At least it has no harm because we have `update()`.

neerajprad:
I tried using versions `4.32.1` and `4.32.2` and they both ran fine. It is probably a platform specific issue. 

> But rather than following up with this issue and waiting for a (possibly) fix in future version of tqdm, we can use refresh=False. 

Agreed, I just wanted to point out that this is likely some bug in tqdm that should be fixed upstream. I'm happy to merge this fix once the build passes!

neerajprad:
I'm not sure why this has triggered two travis builds, and one of them has been queued for quite a while now.

PR #1949, opened 2019-07-09
Use refresh=False for tqdm

fehiepsi:
Resubmit #1948. For some reasons, that PR triggers 2 Travis builds and one of them was stuck.


neerajprad:
Hmm..this has triggered the build twice again. I'll merge this since the first one passes.

PR #1950, opened 2019-07-10
Performance tweaks: Trace.__contains__ and ignore_jit_warnings

fritzo:
This PR makes two tweaks to remove overhead of small Python ops:
1. Adds a `Trace.__contains__()` method. Previously `site_name in self` was executed using `Trace.__iter__` hence linearly scanning through the trace. This quadratic cost is apparent in large models with ~1000 sample sites. After this PR, `Trace.add_node()` takes negligible time.
2. Changes the `ignore_jit_warnings()` context manager to only install warnings *inside* jitted code. Installing warnings filters is surprisingly expensive, and was happening at each sample site. After this PR, `ignore_jit_warnings()` must be used inside jitted functions, hence the changes to pyro.ops.jit and pyro.infer.mcmc.util.

## Performance
1. 7% speedup of a TreeCat inference run
2. 5% speedup of a TreeCat inference run

These are independent and total to a 12% speedup.
## Tested
- exercised by existing tests
- ran a TreeCat inference script to verify speedup

### DIFF
--- a/pyro/poutine/trace_struct.py
+++ b/pyro/poutine/trace_struct.py
@@ -78,9 +78,11 @@ class Trace(object):
         self._succ = OrderedDict()
         self._pred = OrderedDict()
 
+    def __contains__(self, name):
+        return name in self.nodes
+
     def __iter__(self):
-        for node in self.nodes:
-            yield node
+        return iter(self.nodes.keys())
 
     def __len__(self):
         return len(self.nodes)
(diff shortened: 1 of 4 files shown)


pyro/poutine/trace_struct.py
```diff
@@ -78,9 +78,11 @@ def __init__(self, graph_type="flat"):
         self._succ = OrderedDict()
         self._pred = OrderedDict()
 
+    def __contains__(self, name):
+        return name in self.nodes
```
neerajprad: Can't believe we were iterating through all the nodes. 😄 Missing this was probably my handiwork while refactoring to remove `networkx`.

PR #1952, opened 2019-07-11
Minor change to the MCMC interface

neerajprad:
Based on discussions in #1942, this makes a change to the mcmc interface:

**Previous:**
```python
samples = MCMC(model, ..).run(x, y)
```

**Proposed:**
```python
mcmc = MCMC(model, ..)
mcmc.run(x, y)
samples = mcmc.get_samples()
diagnostics = mcmc.diagnostics()
```

The motivation for this change is that the `MCMC` class supports more methods now like `.diagnostics` and it therefore makes sense to have users hold a reference to the MCMC object itself, so that we can later call `mcmc.diagnostics()` etc. This also makes it convenient to get samples where the chain dim is collapsed/uncollapsed so that users don't need to do this for each sample site.

Other minor change is the `disable_validation=True` flag. Since we record divergent transitions, it makes sense to disable distribution validation checks, otherwise divergent transitions would exit with an exception which will be quite common with complex models.

### DIFF
--- a/pyro/infer/mcmc/api.py
+++ b/pyro/infer/mcmc/api.py
@@ -25,7 +25,6 @@ import pyro
 from pyro.infer.mcmc import HMC, NUTS
 from pyro.infer.mcmc.logger import initialize_logger, DIAGNOSTIC_MSG, TqdmHandler, ProgressBar
 from pyro.infer.mcmc.util import diagnostics, initialize_model
-from pyro.util import optional
 
 MAX_SEED = 2**32 - 1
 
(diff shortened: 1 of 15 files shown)


pyro/infer/mcmc/api.py
```diff
@@ -319,13 +324,14 @@ def __init__(self, kernel, num_samples, warmup_steps=None, initial_params=None,
...
         num_samples = [0] * self.num_chains
         z_acc = defaultdict(lambda: [[] for _ in range(self.num_chains)])
-        for x, chain_id in self.sampler.run(*args, **kwargs):
-            if num_samples[chain_id] == self.num_samples:
-                self._diagnostics[chain_id] = x
-            else:
-                num_samples[chain_id] += 1
-                for k, v in x.items():
-                    z_acc[k][chain_id].append(v)
+        with pyro.validation_enabled(not self.disable_validation):
```
fehiepsi: Thanks for catching this! I intended to change [this line](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/mcmc/hmc.py#L299) in the last PR but somehow I forgot to do so.

PR #1954, opened 2019-07-12
CUDA fixes for MCMC

neerajprad:
This PR has some last remaining fixes for CUDA:

 - [x] I was seeing a segmentation fault towards the very end of the baseball example, but all reported results were correct. gdb showed the error coming from this [file](https://github.com/pytorch/pytorch/blob/master/torch/csrc/CudaIPCTypes.cpp). It seems that this is related to a recent PyTorch change that implements reference counting for shared CUDA tensors (https://github.com/pytorch/pytorch/pull/16854). Since we were terminating our workers before the tensors were stacked, it resulted in a segmentation fault during deallocation. This is now fixed by only terminating workers once the tensors are stacked in the main process.
 - [x] ~With https://github.com/pytorch/pytorch/pull/16854, we also don't need to use `mp.Event` anymore and this has been removed. This makes the code slightly faster for small models.~ This isn't true - see discussion below.
 - [x] Moved to using the default initialization as in Stan and NumPyro, in `initialize_model`. This isn't exposed to the user currently since it might be worth refactoring this part to expose multiple initialization strategies as in NumPyro (https://github.com/pyro-ppl/numpyro/pull/237).
 - [x] Checking for both the finiteness of potential energy and its gradient when computing valid initial parameters.
 - [x] Resolving other minor CUDA test failures by setting `mp_context` to `spawn` for CUDA tensors as per PyTorch guidelines.
 - [x] Check that all remaining MCMC tests pass on CUDA.

### DIFF
--- a/pyro/infer/mcmc/api.py
+++ b/pyro/infer/mcmc/api.py
@@ -64,8 +64,8 @@ def logger_thread(log_queue, warmup_steps, num_samples, num_chains, disable_prog
 
 
 class _Worker(object):
-    def __init__(self, chain_id, result_queue, log_queue, event,
-                 kernel, num_samples, warmup_steps, initial_params=None, hook=None):
+    def __init__(self, chain_id, result_queue, log_queue, kernel, num_samples, warmup_steps,
+                 initial_params=None, hook=None):
         self.chain_id = chain_id
         self.kernel = kernel
         if initial_params is not None:
@@ -77,7 +77,6 @@ class _Worker(object):
         self.result_queue = result_queue
         self.default_tensor_type = torch.Tensor().type()
         self.hook = hook
-        self.event = event
 
     def run(self, *args, **kwargs):
         pyro.set_rng_seed(self.rng_seed)
@@ -96,8 +95,6 @@ class _Worker(object):
             for sample in _gen_samples(self.kernel, self.warmup_steps, self.num_samples, logging_hook,
                                        *args, **kwargs):
                 self.result_queue.put_nowait((self.chain_id, sample))
-                self.event.wait()
-                self.event.clear()
             self.result_queue.put_nowait((self.chain_id, None))
         except Exception as e:
             logger.exception(e)
@@ -146,6 +143,9 @@ class _UnarySampler(object):
         self.hook = hook
         super(_UnarySampler, self).__init__()
 
+    def terminate(self, *args, **kwargs):
+        pass
+
     def run(self, *args, **kwargs):
         logger = logging.getLogger("pyro.infer.mcmc")
         progress_bar = ProgressBar(self.warmup_steps, self.num_samples, disable=self.disable_progbar)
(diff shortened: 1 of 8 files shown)


pyro/infer/mcmc/api.py
```diff
@@ -231,8 +234,9 @@ def run(self, *args, **kwargs):
                     yield val, chain_id
                 else:
                     active_workers -= 1
+            exc_raised = False
         finally:
-            self.terminate()
+            self.terminate(terminate_workers=exc_raised)
 
 
 class MCMC(object):
```
fehiepsi: could you update the docs for `hook_fn`?

pyro/infer/mcmc/util.py
```diff
@@ -281,21 +285,32 @@ def get_potential_fn(self, jit_compile=False, skip_jit_warnings=True, jit_option
...
+            if strategy == "uniform":
+                params = {k: dist.Uniform(v.new_full(v.shape, -2), v.new_full(v.shape, 2)).sample()
+                          for k, v in params.items()}
+            elif strategy == "prior":
+                trace = poutine.trace(model).get_trace(*model_args, **model_kwargs)
+                samples = {name: trace.nodes[name]["value"].detach() for name in params}
+                params = {k: transforms[k](v) for k, v in samples.items()}
+            pe_grad, pe = potential_grad(potential_fn, params)
+
+            if torch.isfinite(pe) and all(map(torch.all, map(torch.isfinite, pe_grad.values()))):
```
fehiepsi: +1

pyro/infer/mcmc/api.py
neerajprad: Oops..thanks for catching this.

neerajprad:
> I'll need to double check CUDA issue in my system because the current PyTorch docs suggests to use Event.

I think the docs are just using event to ensure that the worker process doesn't quit until the producer is done with the tensors! But please do check so that we don't end up facing the same issues as earlier regarding receiving garbage values.

fehiepsi:
@neerajprad The example runs smoothly in my system. Nicely clean up!

neerajprad:
> I think the docs are just using event to ensure that the worker process doesn't quit until the producer is done with the tensors

@fehiepsi - I think we need to still use `events` actually. It does seem that the tensors can be gc'd while they are in the queue if the producer process doesn't hold a reference to it. I ran your example from https://github.com/pytorch/pytorch/issues/16141#issuecomment-455893253 and it fails without using events, so I think we'll need to revert that change. I think the reason why it seems to work in our case is that the consumer is faster than the producers.

tests/infer/mcmc/test_nuts.py
```diff
@@ -184,7 +184,9 @@ def model(data):
     assert_equal(samples["p_latent"].mean(0), true_probs, prec=0.02)
 
 
-@pytest.mark.parametrize("jit", [False, mark_jit(True)], ids=jit_idfn)
+@pytest.mark.parametrize("jit", [False, mark_jit(True, marks=[
+    pytest.mark.skipif("CUDA_TEST" in os.environ, reason="FIXME: Too small step size with CUDA on JIT.")
+])], ids=jit_idfn)
```
fehiepsi: It is tricky to debug the root of this. All I can tell is potential_fn gives wrong derivatives under CUDA+JIT for this type of model. Gamma prior does not cause problem because when I replace `Gamma` prior by `HalfNormal(rate)`, I still get the grad issue. This looks weird to me because this issue does not happen for some other models. I think we need to create an issue to track this problem.

pyro/infer/mcmc/api.py
```diff
@@ -371,11 +388,13 @@ def get_samples(self, num_samples=None, group_by_chain=False):
...
-                batch_dim = 0
-            else:
-                batch_dim = 1
+            batch_dim = 0
+            if self.num_chains > 1:
+                if group_by_chain:
+                    batch_dim = 1
+                else:
+                    samples = {k: v.reshape((-1,) + v.shape[2:]) for k, v in samples.items()}
+                    batch_dim = 0
```
fehiepsi: haha, the logic here is tricky to detect :D

tests/infer/mcmc/test_nuts.py
neerajprad: Thanks for digging into this!

Issue #1737, opened 2019-02-07
Pyro minor release 0.3.1


neerajprad:
Task-list:
 - [x] Remove networks x from dependency. #1736 
 - [x] Update ~~setup.py and~~ .travis.yml to use PyTorch 1.0.1
 - [x] Fix CUDA tests.
 - [ ] Address issues for next release:
   - [x] Progress bar fixes in mcmc: https://github.com/pyro-ppl/pyro/issues/1728
   - [ ] ~Abstract infer classes lazily consume traces: https://github.com/pyro-ppl/pyro/issues/1725. (not a blocker)~
   - [ ] ~Sample from collapsed latent variables in conjugate distributions. (not a blocker)~
   - Any other?

Please feel free to add to the list if we think there are certain fixes (or features, though prioritizing on fixes) that should make it into this release.

eb8680:
Are there any other easy performance wins along the lines of #1736 we could include?

fritzo:
I suggest we update travis to use PyTorch 1.0.1 but keep setup.py permissive via `torch>=1.0`, since Pyro doesn't require anything in the new PyTorch release.

neerajprad:
> Are there any other easy performance wins along the lines of #1736 we could include?

I can't think of much at the moment. There's the issue of TracePosterior being lazy for MCMC that I mentioned above, but that's a narrow use case that will help multiprocessing.

neerajprad:
Is everyone good with releasing a minor version with the current dev branch?

PR #1956, opened 2019-07-14
Fix memory issue of MCMC for multiple chains

fehiepsi:
Addresses #1737 

This PR fixes shared memory issue when running mcmc with multiple chains. Previously, we hold all references of tensors in subprocesses and main process, which might cause memory issue when the number of samples is large. This PR clones and deletes shared tensors in the main process.

Because `.clone()` operator adds overhead, especially when the model has many latent sites, I revise `gen_samples` a bit to make it return only 1 flatten tensor. After collecting all flatten samples, we unpack them into the original `dict` format. As for benchmark, I don't observe any regression with this approach:
+ test for `logistic_regression` example, this PR took 15.05s while `dev` branch took 15.39s
+ for `beta_bernoulli` example, this PR took 18.72s while `dev` branch took 18.86s
+ for bayesian regression model, with 3000 samples, 1000 warmup steps, 4 chains, this PR took 1.28s while `dev` branch took 1.29s. In GPU, with 200 warmup steps and 200 samples, both took 45s.
+ similarly, the same performance for baseball example

### DIFF
--- a/examples/baseball.py
+++ b/examples/baseball.py
@@ -373,6 +373,9 @@ if __name__ == "__main__":
                         help="run this example in GPU")
     args = parser.parse_args()
 
+    # work around the error "CUDA error: initialization error" when arg.cuda is False
+    # see https://github.com/pytorch/pytorch/issues/2517
+    torch.multiprocessing.set_start_method("spawn")
     pyro.set_rng_seed(args.rng_seed)
     # Enable validation checks
     pyro.enable_validation(__debug__)
--- a/examples/baseball.py
+++ b/examples/baseball.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function
 import argparse
 import logging
 import math
+import six
 
 import pandas as pd
 import torch
@@ -375,7 +376,8 @@ if __name__ == "__main__":
 
     # work around the error "CUDA error: initialization error" when arg.cuda is False
     # see https://github.com/pytorch/pytorch/issues/2517
-    torch.multiprocessing.set_start_method("spawn")
+    if six.PY3:
+        torch.multiprocessing.set_start_method("spawn")
     pyro.set_rng_seed(args.rng_seed)
     # Enable validation checks
     pyro.enable_validation(__debug__)
--- a/pyro/infer/mcmc/api.py
+++ b/pyro/infer/mcmc/api.py
@@ -14,7 +14,7 @@ import logging
 import signal
 import threading
 import warnings
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict
 
 import six
 import torch
@@ -107,13 +107,15 @@ class _Worker(object):
 def _gen_samples(kernel, warmup_steps, num_samples, hook, *args, **kwargs):
     kernel.setup(warmup_steps, *args, **kwargs)
     params = kernel.initial_params
+    # yield structure (key, value.shape) of params
+    yield {k: v.shape for k, v in params.items()}
     for i in range(warmup_steps):
         params = kernel.sample(params)
         hook(kernel, params, 'warmup', i)
     for i in range(num_samples):
         params = kernel.sample(params)
         hook(kernel, params, 'sample', i)
-        yield params
+        yield torch.cat([params[site].reshape(-1) for site in sorted(params)])
     yield kernel.diagnostics()
     kernel.cleanup()
 
@@ -337,18 +339,36 @@ class MCMC(object):
 
     def run(self, *args, **kwargs):
         num_samples = [0] * self.num_chains
-        z_acc = defaultdict(lambda: [[] for _ in range(self.num_chains)])
+        z_flat_acc = [[] for _ in range(self.num_chains)]
         with pyro.validation_enabled(not self.disable_validation):
             for x, chain_id in self.sampler.run(*args, **kwargs):
-                if num_samples[chain_id] == self.num_samples:
+                if num_samples[chain_id] == 0:
+                    num_samples[chain_id] += 1
+                    z_structure = x
+                elif num_samples[chain_id] == self.num_samples + 1:
                     self._diagnostics[chain_id] = x
                 else:
                     num_samples[chain_id] += 1
-                    for k, v in x.items():
-                        z_acc[k][chain_id].append(v)
+                    if self.num_chains > 1:
+                        x_cloned = x.clone()
+                        del x
+                    else:
+                        x_cloned = x
+                    z_flat_acc[chain_id].append(x_cloned)
 
-        z_acc = {k: [torch.stack(l) for l in v] for k, v in z_acc.items()}
-        z_acc = {k: v[0] if self.num_chains == 1 else torch.stack(v) for k, v in z_acc.items()}
+        z_flat_acc = torch.stack([torch.stack(l) for l in z_flat_acc])
+
+        # unpack latent
+        pos = 0
+        z_acc = z_structure.copy()
+        for k in sorted(z_structure):
+            shape = z_structure[k]
+            next_pos = pos + shape.numel()
+            # NB: squeeze in case num_chains=1
+            z_acc[k] = z_flat_acc[:, :, pos:next_pos].reshape(
+                (self.num_chains, self.num_samples) + shape).squeeze(dim=0)
+            pos = next_pos
+        assert pos == z_flat_acc.shape[-1]
 
         # If transforms is not explicitly provided, infer automatically using
         # model args, kwargs.
--- a/pyro/infer/mcmc/api.py
+++ b/pyro/infer/mcmc/api.py
@@ -115,7 +115,7 @@ def _gen_samples(kernel, warmup_steps, num_samples, hook, *args, **kwargs):
     for i in range(num_samples):
         params = kernel.sample(params)
         hook(kernel, params, 'sample', i)
-        yield torch.cat([params[site].reshape(-1) for site in sorted(params)])
+        yield torch.cat([params[site].reshape(-1) for site in sorted(params)]) if params else torch.tensor([])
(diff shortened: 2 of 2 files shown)


examples/baseball.py
```diff
@@ -373,6 +373,9 @@ def main(args):
                         help="run this example in GPU")
     args = parser.parse_args()
 
+    # work around the error "CUDA error: initialization error" when arg.cuda is False
+    # see https://github.com/pytorch/pytorch/issues/2517
+    torch.multiprocessing.set_start_method("spawn")
```
fehiepsi: When using multiprocessing in CPU, if pytorch is installed with cuda support, we should "avoid any call to cuda prior to starting subprocesses". In this case, one of the call is `torch.manual_seed(...)` statement.

pyro/infer/mcmc/api.py
```diff
@@ -337,18 +339,36 @@ def __init__(self, kernel, num_samples, warmup_steps=None, initial_params=None,
 
     def run(self, *args, **kwargs):
         num_samples = [0] * self.num_chains
-        z_acc = defaultdict(lambda: [[] for _ in range(self.num_chains)])
+        z_flat_acc = [[] for _ in range(self.num_chains)]
         with pyro.validation_enabled(not self.disable_validation):
             for x, chain_id in self.sampler.run(*args, **kwargs):
-                if num_samples[chain_id] == self.num_samples:
+                if num_samples[chain_id] == 0:
```
neerajprad: Thanks for detecting and fixing this issue! My preliminary thoughts on this are as follows:

 - Based on your careful profiling results, I think our existing examples are all small and `.clone` does not incur much of an overhead on these examples. My guess is (correct me if I am mistaken) that it will also not matter for examples with larger sample sites because the rate determining step there would be the worker processes and not the main process that is accumulating the samples. In that scenario, I'm not sure if its worth optimizing for 1 .clone over 10 .clone on smaller tensors. If it is not, I think we might as well keep our code simpler and slightly easier to debug by not doing this flattening / unflattening. Also we can just do all the `.clone`s in the `_Multisampler` and avoid it for the the `chain=1` case. That way, even if there is some minor regression it is confined to the multi chain case.
 - If it is the case that there is a significant difference, is it possible to just make changes to do this flattening / unflattening only within the workers / _MultiSampler so that we don't need to add complexity / any additional overhead to the single chain case?

fehiepsi: About the first point, I thought that you would worry about the overhead issue because you did mention that there are applications which requires large number of parameters or large tensors, so I tried to think how to optimize it to avoid regression. If you don't worry about it, I will make the change to use only `.clone()` operator. My little background told me that when the sampler (generator) generates a sample, it have to wait for the consumer triggers some `next(...)` signal, then it'll execute the code after `yield ...` statement. In our case, the code after `yield` statement is to activate the subprocess which generates that sample (through `event.set()`). So if the consumer overhead is large (in case of it has to perform many `clone` operators), that subprocess has to wait a while to run again. That might not be the end of the story, subprocesses put samples to a queue. So a sample from a subprocess has to wait for current samples in the queue consumed before it is yielded to consumer.

About the second point, it would be more complicated (and harder to debug) if we maintain two separate codes for MultiSampler and SingleSampler. Consider the single chain case, I don't think that there is overhead involved here. Could you suggest an example so I can test? I did a benchmark to compare two ways
+ populate samples to a defaultdict, for each site, apply a stack operator over num_samples
+ apply num_samples flatten operator + 1 stack operator over num_samples  + split the result into num_sites
and see that they are quite comparable

```
from collections import defaultdict
import torch
params = [{'x_{}'.format(i): torch.tensor(i) for i in range(100)} for j in range(10000)]

%%time
z = defaultdict(list)
for j in range(10000):
    for i in range(100):
        z['x_{}'.format(i)].append(params[j]['x_{}'.format(i)])
stacked_params = {}
for i in range(100):
    stacked_params['x_{}'.format(i)] = torch.stack(z['x_{}'.format(i)])

%%time
flatten_params = [torch.cat([s.reshape(-1) for s in p.values()]) for p in params]
stacked_flatten_params = torch.stack(flatten_params)
stacked_params = {}
for i in range(100):
    stacked_params['x_{}'.format(i)] = stacked_flatten_params[:, i]
```
which in my system shows that the first approach took 2s and the second approach took 2.5s.

That is with num_sites=100. If I change 100 -> 10, I observe that they are 250ms vs 320ms. So I think that these numbers are quite small for us to worry about.

neerajprad: >  So if the consumer overhead is large (in case of it has to perform many clone operators), that subprocess has to wait a while to run again. That might not be the end of the story, subprocesses put samples to a queue. So a sample from a subprocess has to wait for current samples in the queue consumed before it is yielded to consumer.

While this is true currently, I believe it is only an artificial constraint. In the workers, we can always put the samples into the queue and also append the sample to a list and only use `mp.Event` to periodically clear that list (which we discussed at one point). That way, the workers can generate samples more or less asynchronously without waiting for the main process to read off the same sample. I'm not sure if its worth making this optimization just yet, but we can do so if reading from the main process and cloning is expensive for a model.

> If you don't worry about it, I will make the change to use only .clone() operator.

I only prefer the simpler solution because doing otherwise seems like a case of premature optimization at this point. It is good to know that we can do this if this really turned out to be a significant problem though. That said, this is purely an internal change and does not affect the user interface, so I'll defer to your judgment on this.

pyro/infer/mcmc/api.py
```diff
@@ -107,13 +107,15 @@ def run(self, *args, **kwargs):
...
+    # yield structure (key, value.shape) of params
+    yield {k: v.shape for k, v in params.items()}
     for i in range(warmup_steps):
         params = kernel.sample(params)
         hook(kernel, params, 'warmup', i)
     for i in range(num_samples):
         params = kernel.sample(params)
         hook(kernel, params, 'sample', i)
-        yield params
+        yield torch.cat([params[site].reshape(-1) for site in sorted(params)]) if params else torch.tensor([])
```
neerajprad: What do you think about having generic pack, unpack utilities?

```python
> structure, tensor = pack({"a": torch.ones(2, 2), "b": torch.tensor(2.)})
> tensors
torch.Tensor([1., 1., 1., 1., 2.])

> structure
{"a": torch.Size([2, 2]), "b": torch.Size([])}

> unpack(tensor, structure)
{"a": torch.ones(2, 2), "b": torch.tensor(2.)}
```

We can also do a `stacked_unpack` which would involve stacking the tensors and unpacking them so that the leading dim has size `len(tensors)`:

```
> stacked_unpack([torch.tensor([1., 1., 1., 1., 2.]), torch.tensor(0., 0., 0., 0., 4.)], structure)
{"a": torch.Tensor([[[1., 1.], [1., 1.]], [[0., 0.], [0., 0]]]), "b": torch.Tensor([[2.], [4.]])}
```

pyro/infer/mcmc/api.py
neerajprad: I'm also warming up to this idea of using packed/unpacked tensors. See my comment above, though you don't need to worry about building any generic utilities in this PR; I'm happy to merge this as is, and refactor this later if needed.

fehiepsi: I think that we can test our assumption with a simple kernel. Will do it soon to see the effect of these changes.

pyro/infer/mcmc/api.py
fehiepsi: I love these utilities. :D

pyro/infer/mcmc/api.py
neerajprad: > I think that we can test our assumption with a simple kernel.

Please don't spend too much time profiling. I don't think there are any perf repercussions to this change, and my concerns around code complexity can be handled by building generic utilities later.

fehiepsi: I just test on a simple PriorKernel
```
class PriorKernel(MCMCKernel):
    @property
    def initial_params(self):
        return self._initial_params

    @initial_params.setter
    def initial_params(self, params):
        self._initial_params = params

    def sample(self, params):
        return params

kernel = PriorKernel()
initial_params = {'x_{}'.format(i): torch.tensor(i).repeat(4) for i in range(100)}
mcmc = MCMC(kernel, 1000, 0, initial_params=initial_params, num_chains=4, transforms={})
mcmc.run()
```
and see that the dev version took 1m24s with the default sharing strategy and 4s with `torch.multiprocessing.set_sharing_strategy('file_system')` (which is default in MacOS) while this PR took 1s. Beyond the memory error, it is also easy to trigger `RuntimeError: received 0 items of ancdata.` with default sharing strategy (for example, baseball example). So we can see other benefits of using this pack/unpack job:
+ users don't have to worry about the error `RuntimeError: received 0 items of ancdata.` with default sharing strategy
+ it is a bit faster when model has many sites :)

neerajprad: I think with the PriorKernel, the workers will be super fast, so it may not be surprising that with large number of sites, we are bottlenecked in the main process but the perf benefit might not be noticeable for HMC/NUTS. :)

> torch.multiprocessing.set_sharing_strategy('file_system')

Should we make this the default?

fehiepsi: > Should we make this the default?

It just happens when we don't use `clone` & `del`. Because this PR is merged, I think we don't have to worry about it.

I forgot the test for using `.clone()` for each site. It took 18s (comparing to 4s without clone and 1s when using pack/unpack) to get 1000 samples. Delay the sampling speed (by using `time.sleep(0.01)`) does not affect pack/unpack pattern (took 11s to get 1000 samples) but it helps for the case we use `clone` for each site (took 19s to get 1000 samples). So somehow the overhead of the main process does not affect much the sampling speed.

PR #1958, opened 2019-07-15
Implement parallelized DiscreteHMM distribution

fritzo:
This PR follows an idea of @eb8680 to implement a parallelized HMM evaluator from [this paper](https://arxiv.org/pdf/1905.13002.pdf). This PR implements that evaluator as a new distribution, currently with only a `.log_prob()` method.

## Performance

The following table compares two equivalent models: one using `TraceEnum_ELBO` for enumeration and one using `DiscreteHMM`.

| Time (sec) | Method | `python -O examples/hmm.py --num-steps 100 ...` |
| -: | - | - |
| 72.9 | `TraceEnum_ELBO` | `--model 5` |
| 63.4 | `TraceEnum_ELBO` | `--model 5 --cuda` |
| 6.6 | `DiscreteHMM` | `--model 7` |
| 5.4 | `DiscreteHMM` | `--model 7 --cuda` |

## Tasks
- [x] add generic `DiscreteHMM` distribution
- [x] add tests for underlying math ops
- [x] add tests for `.log_prob()`
- [x] add docs and citation
- [x] add an example model in examples/hmm.py
- [x] add tests for shapes and broadcasting


pyro/distributions/hmm.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def _logmatmulexp(x, y):
+    """
+    Numerically stable version of ``(x.log() @ y.log()).exp()``.
+    """
+    x_shift = x.max(-1, keepdim=True)[0]
+    y_shift = y.max(-2, keepdim=True)[0]
+    xy = torch.matmul((x - x_shift).exp(), (y - y_shift).exp()).log()
+    return xy + x_shift + y_shift
```
fehiepsi: Nice trick! I think you can also use [this implementation](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/transforms/block_autoregressive.py#L18) for simplicity.

pyro/distributions/hmm.py
```diff
@@ -0,0 +1,95 @@
...
+        x[..., 0, :, :] @ x[..., 1, :, :] @ ... @ x[..., T-1, :, :]
+
+    but does so numerically stably in log space.
+    """
+    batch_shape = logits.shape[:-3]
+    state_dim = logits.size(-1)
+    while logits.size(-3) > 1:
+        time = logits.size(-3)
+        even_time = time // 2 * 2
+        even_part = logits[..., :even_time, :, :]
```
martinjankowiak: yeah, i think this'll be a bit faster than a for loop.... 😮

pyro/distributions/hmm.py
```diff
@@ -0,0 +1,95 @@
...
+        x_y = even_part.reshape(batch_shape + (even_time // 2, 2, state_dim, state_dim))
+        x, y = x_y.unbind(-3)
+        contracted = _logmatmulexp(x, y)
```
eb8680: Note if `x`, `y` were `funsor.Gaussian`s, we would be able to reuse this code for Gaussian HMMs simply by changing the definition of this operation

pyro/distributions/hmm.py
fritzo: `torch.matmul()` is faster and uses less memory than broadcast-and-contract.

fehiepsi: That makes sense, thanks for explaining!

pyro/distributions/hmm.py
fritzo: Yeah I'd like to gradually make more abstract versions of this code, including multiple `opt_einsum` backends, more complex model structure (e.g. passing in an einsum eqution), and eventually Funsor support. In this first PR I've tried to customize so we can measure speedup with minimal abstraction overhead.

pyro/distributions/hmm.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+    arg_constraints = {"initial_logits": constraints.real,
+                       "transition_logits": constraints.real}
+
+    def __init__(self, initial_logits, transition_logits, observation_dist, validate_args=None):
+        if initial_logits.dim() < 1:
+            raise ValueError("expected initial_logits to have at least one dim, "
+                             "actual shape = {}".format(initial_logits.shape))
+        if transition_logits.dim() < 2:
+            raise ValueError("expected transition_logits to have at least two dims, "
```
martinjankowiak: according to docstring `transition_logits.dim()>2`

pyro/distributions/hmm.py
```diff
@@ -0,0 +1,120 @@
...
+        categorical distribution over latent states. Should have rightmost size
+        ``state_dim`` and be broadcastable to ``batch_shape + (state_dim,)``.
+    :param torch.Tensor transition_logits: A logits tensor for transition
+        conditional distributions between latent states. Should have rightmost
+        shape ``(num_steps, state_dim, state_dim)`` (time, old, new), and be
+        broadcastable to ``batch_shape + (num_steps, state_dim, state_dim)``.
+    :param torch.distriburtions.Distribution observation_dist: A conditional
+        distribution of observed data conditioned on latent state. The
+        ``.batch_shape`` should have rightmost size ``state_dim`` and be
+        broadcastable to ``batch_shape + (num_steps + state_dim)``. The
```
martinjankowiak: num_steps + state_dim => num_steps , state_dim

pyro/distributions/hmm.py
```diff
@@ -0,0 +1,120 @@
...
+        https://arxiv.org/pdf/1905.13002.pdf
+
+    :param torch.Tensor initial_logits: A logits tensor for an initial
+        categorical distribution over latent states. Should have rightmost size
+        ``state_dim`` and be broadcastable to ``batch_shape + (state_dim,)``.
+    :param torch.Tensor transition_logits: A logits tensor for transition
+        conditional distributions between latent states. Should have rightmost
+        shape ``(num_steps, state_dim, state_dim)`` (time, old, new), and be
+        broadcastable to ``batch_shape + (num_steps, state_dim, state_dim)``.
+    :param torch.distriburtions.Distribution observation_dist: A conditional
```
martinjankowiak: typo: distriburtions

pyro/distributions/hmm.py
```diff
@@ -0,0 +1,120 @@
...
+        distribution of observed data conditioned on latent state. The
+        ``.batch_shape`` should have rightmost size ``state_dim`` and be
+        broadcastable to ``batch_shape + (num_steps + state_dim)``. The
+        ``.event_shape`` may be arbitrary.
+    """
+    arg_constraints = {"initial_logits": constraints.real,
+                       "transition_logits": constraints.real}
+
+    def __init__(self, initial_logits, transition_logits, observation_dist, validate_args=None):
```
martinjankowiak: is `validate_args=None` more confusing than `validate_args=False`?

pyro/distributions/hmm.py
```diff
@@ -0,0 +1,120 @@
...
+        self.transition_logits = transition_logits - transition_logits.logsumexp(-1, True)
+        self.observation_dist = observation_dist
+        super(DiscreteHMM, self).__init__(batch_shape, event_shape, validate_args=validate_args)
+
+    def expand(self, batch_shape, _instance=None):
+        new = self._get_checked_instance(DiscreteHMM, _instance)
+        batch_shape = torch.Size(broadcast_shape(self.batch_shape, batch_shape))
+        new.initial_logits = self.initial_logits.expand(batch_shape + (-1,))  # cheapest to expand
+        new.transition_logits = self.transition_logits
+        new.observation_dist = self.observation_dist
```
martinjankowiak: why do `transition_logits` and `observation_dist` not get expanded?

fritzo: I've added the following comment in the code:

> We only need to expand one of the inputs, since batch_shape is determined
  by broadcasting all three. To save computation in _sequential_logmatmulexp(),
  we expand only initial_logits, which is applied only after the logmatmulexp.
  This is similar to the `._unbroadcasted_*` pattern used elsewhere in distributions.

pyro/distributions/hmm.py
fritzo: `validate_args=None` has the special meaning "consult the global validation flag", which is controlled by `enable_validation()`. This behavior can be overridden in a single distribution by passing either `True` or `False`.

Issue #1960, opened 2019-07-16
Bug (?) in the shape of samples from auto-guided LKJCorr


null-a:
### Issue Description

In the following example, the shape of the value sampled at the `LKJCorrCholesky` distributed `sample` statement differs between the model and the auto-generated guide. Is this expected?

```python
import torch
import pyro
import pyro.distributions as dist
from pyro.contrib.autoguide import AutoDiagonalNormal

d = dist.LKJCorrCholesky(2, torch.tensor(1.))

def model():
    return {'x': pyro.sample('x', d)}

guide = AutoDiagonalNormal(model)

print(model()['x'].shape) # == [2,2]
print(guide()['x'].shape) # == [1,2,2]
```

### Environment

macos, python 3.7.3, pyro 0.3.3+bf2f9542

<details>

I dug into this a little while trying to understand what was happening. If this *is* a bug, it may be that the problem stems from the shape of the value returned by `_unpack_latent`:

```python
import torch
import pyro
import pyro.distributions as dist
from torch.distributions import biject_to
from pyro.contrib.autoguide import AutoDiagonalNormal

d = dist.LKJCorrCholesky(2, torch.tensor(1.))
support = d.support

def model():
    return {'x': pyro.sample('x', d)}

guide = AutoDiagonalNormal(model)

latent = guide.sample_latent()
unconstrained_value = [val for site, val in guide._unpack_latent(latent)
                       if site['name'] == 'x'][0]
# The shape of the unconstrained value is [1,1]
print(unconstrained_value.shape) # == [1,1]
# Which gives a [1,2,2] tensor once transformed:
print(biject_to(support)(unconstrained_value).shape) # == [1,2,2]

# Reshaping the unconstrained value from [1,1] to [1] gives the result
# of the transform the expected shape.
print(biject_to(support)(unconstrained_value.reshape(-1)).shape) # == [2,2]
```

</details>

fehiepsi:
Yes, I think it is a bug at [this line](https://github.com/pyro-ppl/pyro/blob/dev/pyro/contrib/autoguide/__init__.py#L409). Probably we should replace it by
```
unconstrained_shape = broadcast_shape(unconstrained_shape, batch_shape + unconstrained_shape)
```
The reason is for LKJ transform, unconstrained value is a vector, while constrained value is a matrix.

fritzo:
[...]

PR #1961, opened 2019-07-16
Fix autoguide latent shape mismatch

fehiepsi:
Addresses #1960. The issue is for some transform such as LKJCholeskyTransform, shapes of domain and codomain are different. This PR corrects the event_dim of unconstrained value.

Solution is from @fritzo, thanks!

TODO:
- [x] add test to exercise this bug


fehiepsi:
@fritzo done!

PR #1962, opened 2019-07-16
Drop Python2.7 tests from ci grid

fritzo:
Addresses #1963

This is the first step to dropping Python 2 support.
This removes Python 2 tests from ci and merges two of our test stages.
This should lead to lower-latency ci tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,10 +35,10 @@ branches:
         - dev
 
 jobs:
-    fast_finish: true
     include:
         - stage: lint
           python: 3.6
+          name: lint
           before_install: pip install flake8 nbstripout nbformat
           install:
           script:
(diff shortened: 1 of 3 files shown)


fritzo:
@neerajprad @jpchen are there any other changes that need to be made in this first PR, e.g. docs or tutorial configs?

fritzo:
@neerajprad what do you think about combining the "auxiliary", "perf", and "profiler" stages?

neerajprad:
> what do you think about combining the "auxiliary", "perf", and "profiler" stages, so they run in parallel?

Just did this. So currently, we have the following schedule:
 - lint
 - docs, benchmark, profiling
 - unit tests, examples
 - integration tests - two batches

fritzo:
@neerajprad I prefer merging the unit test and integration test stages, since we have enough cores to run them all in parallel. Do you have a good reason for separating them?

neerajprad:
> Do you have a good reason for separating them?

Not really. We'll have max usage of 4 parallel jobs, same as earlier. So I'll combine unit, examples and integration tests together, in that case.

fritzo:
38 minutes is pretty good :smile:

Issue #1963, opened 2019-07-16
Drop support for Python 2 (a checklist)


fritzo:
Follows #1935 

I've attempted to prioritize these:

- [x] #1962 Remove Python 2 from ci test grid.
- [x] #1965 Remove dependency on `six`:
    - replace`six.moves.reduce` with `functools.reduce`.
    - ...
- [x] #1967 Replace contextlib2 with contextlib
- [x] #1967 Remove `from __future__ import ...` headers.
- [x] #1972 Remove py2 installation from [docs](http://docs.pyro.ai/en/stable/installation.html) and readme
- [x] #1972 Change dockerfile python versions.
- [x] #1962 Remove Python 2 from setup.py.
- ...others?

### Triaged

- Replace `super(MyClass, self)` -> `super()` in most cases?
- Modernize `six.reraise` logic to clean up error messages.
     Currently error messages look good in Python2 but give duplicate stack traces in Python3.

neerajprad:
> six.moves.reduce is used often

I think we can just use `functools.reduce` in python 3.

jpchen:
> Currently error messages look good in Python2 but give duplicate stack traces in Python3.

this seems to be a consequence of pytest exception handling, not `six`.  will investigate further but this may be separate from the migration.

PR #1965, opened 2019-07-17
drop six dependency

jpchen:
addresses #1963 

- [x] replace methods with their py3 libraries
- [x] replace `six.reraise()` with `raise exception(..).with_traceback(...)`
- [x] replace `@add_metaclass()` with `class foo(object, metaclass=)`
- [x] fix examples

only ran the test suite, assuming we have coverage for all these functions and exception handling.

### DIFF
--- a/pyro/contrib/easyguide/easyguide.py
+++ b/pyro/contrib/easyguide/easyguide.py
@@ -6,7 +6,6 @@ from abc import ABCMeta, abstractmethod
 
 import torch
 from contextlib2 import ExitStack
-from six import add_metaclass
 from torch.distributions import biject_to
 
 import pyro
@@ -18,8 +17,7 @@ from pyro.distributions.util import broadcast_shape, sum_rightmost
 from pyro.poutine.util import prune_subsample_sites
 
 
-@add_metaclass(ABCMeta)
-class EasyGuide(object):
+class EasyGuide(object, metaclass=ABCMeta):
     """
     Base class for "easy guides".
 
--- a/pyro/poutine/trace_messenger.py
+++ b/pyro/poutine/trace_messenger.py
@@ -2,8 +2,6 @@ from __future__ import absolute_import, division, print_function
 
 import sys
 
-import six
-
 from .messenger import Messenger
 from .trace_struct import Trace
 from .util import site_is_subsample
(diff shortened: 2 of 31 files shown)


jpchen:
@neerajprad  i get a precision error in
```
tests/infer/mcmc/test_hmc.py::test_unnormalized_normal[JIT=False-HMC]
AssertionError: 5.104769527853391 vs 5.0
```

related?

neerajprad:
> related?

Not sure how your change affects that test. Please feel free to increase the threshold.

pyro/poutine/trace_messenger.py
```diff
@@ -148,9 +146,7 @@ def __call__(self, *args, **kwargs):
             except (ValueError, RuntimeError):
                 exc_type, exc_value, traceback = sys.exc_info()
                 shapes = self.msngr.trace.format_shapes()
-                six.reraise(exc_type,
-                            exc_type(u"{}\n{}".format(exc_value, shapes)),
-                            traceback)
+                raise exc_type(u"{}\n{}".format(exc_value, shapes)).with_traceback(traceback)
```
neerajprad: Does this get rid of the problem of duplicate stack trace that @fritzo mentioned?

pyro/contrib/easyguide/easyguide.py
```diff
@@ -6,7 +6,6 @@
 
 import torch
 from contextlib2 import ExitStack
```
neerajprad: I think its probably time we can remove `contextlib2` from setup.py and just use the standard `contextlib` library in python 3?

fritzo: Let's add that as a task for a separate PR.

fritzo:
I'll defer to @neerajprad for a second review.

pyro/contrib/easyguide/easyguide.py
```diff
@@ -18,8 +17,7 @@
 from pyro.poutine.util import prune_subsample_sites
 
 
-@add_metaclass(ABCMeta)
-class EasyGuide(object):
+class EasyGuide(object, metaclass=ABCMeta):
```
neerajprad: We can also use [ABC](https://docs.python.org/3/library/abc.html#abc.ABC) as the base class directly with python 3.4+.

jpchen: Will address in follow up, I copied this pattern from the six documentation

jpchen: this syntax seems to be for python 3.6+ but i was thinking that we would support python 3.4+ WDYT?

Issue #1969, opened 2019-07-18
DMM giving NaN


sameerkhurana10:
Hi,

while running DMM example with PyTorch: 1.1 and pyro 0.3.4+6704741 i get nans

training epoch 0003]  nan                              (dt = 10.456 sec)
[training epoch 0004]  nan                              (dt = 10.910 sec)
[training epoch 0005]  nan                              (dt = 10.591 sec)
[training epoch 0006]  nan                              (dt = 11.308 sec)
[training epoch 0007]  nan                              (dt = 11.300 sec)
[training epoch 0008]  nan                              (dt = 10.955 sec)
[training epoch 0009]  nan                              (dt = 10.482 sec)

jpchen:
related: #1272 #711
cc: @martinjankowiak

sameerkhurana10:
it sometimes works and sometimes does not work.

It is really quite unstable.

python dmm.py --cuda
Namespace(annealing_epochs=1000, beta1=0.96, beta2=0.999, checkpoint_freq=0, clip_norm=10.0, cuda=True, iaf_dim=100, jit=False, learning_rate=0.0003, load_model='', load_opt='', log='dmm.log', lr_decay=0.99996, mini_batch_size=20, minimum_annealing_factor=0.2, num_epochs=5000, num_iafs=0, rnn_dropout_rate=0.1, save_model='', save_opt='', weight_decay=2.0)
N_train_data: 229     avg. training seq. length: 60.29    N_mini_batches: 12
[training epoch 0000]  62.9712                          (dt = 11.384 sec)
/data/sls/temp/sameerk/tools/anaconda3/envs/distsup/lib/python3.7/site-packages/pyro/infer/trace_elbo.py:136: UserWarning: Encountered NaN: loss
  warn_if_nan(loss, "loss")
[training epoch 0001]  nan                              (dt = 10.777 sec)
[training epoch 0002]  nan                              (dt = 10.652 sec)

martinjankowiak:
these kinds of models tend to be inherently unstable during training. and the trouble is that you need to stay close to the regime in which it unstable because it's precisely in that more aggressive regime that you're more likely to avoid bad local minima. but it's not such a big problem in practice:
- you can put in try/catch statements, try ~10 different seeds for 20 epochs, say, and then continue training with the one that is performing the best. you can also use the checkpointing mechanism to recover from any nans
[...]

PR #1971, opened 2019-07-18
use more conservative default hyperparameters in dmm.py

martinjankowiak:
for me this works fine on torch 1.1. addresses #1969


martinjankowiak:
uh works means it produces elbo curves that look like what i remember but i only look at a few dozen epochs i don't wait until convergence. if anyone wants to get top-notch log likelihoods they will need to do some tuning that's machine learning

PR #1976, opened 2019-07-23
Added Summary class for conjugate updates

chanjed:
class designed with @fritzo and @karalets

This PR:
   1. Creates an abstract Summary class to aggregate sufficient statistics
   2. Creates two subclasses BetaBernoulliSummary and NIGNormalRegressionSummary

## TODO:
- [ ] I allowed the optional inclusion of input features. Not sure if this should be separated into two classes.
- [ ] Also unclear how much flexibility I should allow in terms of input dims (scalar, 1-D, 2-D, etc)
- [x] Add tests
- [ ] Add a new section at the end of docs/source/distributions.rst, following the pattern of the "Transformed Distributions" section


fritzo:
Closing in favor of https://github.com/pyro-ppl/funsor/issues/367

Issue #1305, opened 2018-08-13
Tensor SMC via local num_particles + resampling


fritzo:
Addresses #1304 Sequential Monte Carlo
Requires #915 dependent enumeration, #1302 local `num_particles`

Currently `TraceEnum_ELBO` would have exponential cost in the number of sampled variables (via local `num_particles` #1302). To mitigate this growth, we could introduce an SMC resampling step in the computation, effectively implementing a COO sparse tensor format whose metadata is recorded in Pyro's `CondIndepStack`. Roughly:
1. Allocate a new tensor dim to index over COO entries of an existing set of tensor dims, say of combined shape `resample_shape`.
2. Resample the existing tensor dims using `num_particles` sparse entries. This will transform a tensor between shapes w.l.o.g.
    ```
    resample_shape + keep_shape -> (num_particles,) + (1,)*len(resample_shape) + keep_shape
    ```
2. Perform special handling of sparse tensor dims in `Dice` and `EnumTraceProbEvaluator`, using metadata stored in the `CondIndepStack`. The special handling should adjust sample weights based on the resampling probabilities.

fritzo:
Closing this in favor of a Funsor implementation https://github.com/pyro-ppl/funsor/issues/12 We may reopen if the Funsor implementation doesn't work out.

PR #1977, opened 2019-07-23
Implement Tensor Monte Carlo in TraceEnum_ELBO

eb8680:
Addresses #1305 

This PR implements [Tensor Monte Carlo](https://arxiv.org/abs/1806.08593) using our enumeration machinery and adds an option to use it in the DMM example.  `TraceEnum_ELBO` now performs Tensor Monte Carlo automatically for sample sites with `infer["enumerate"] == "parallel" and infer["num_samples"] > 1 and infer["expand"] == False`.


pyro/poutine/enumerate_messenger.py
```diff
@@ -5,13 +5,63 @@
...
+    # find batch dims that aren't plate dims
+    batch_shape = [1] * len(dist.batch_shape)
+    for f in msg["cond_indep_stack"]:
+        if f.vectorized:
+            batch_shape[f.dim] = f.size
+    batch_shape = tuple(batch_shape)
+
+    # sample a batch
+    sample_shape = (num_samples,) if batch_shape == dist.batch_shape else ()
+    fat_sample = dist(sample_shape=sample_shape)  # TODO thin before sampling
```
eb8680: Sampling like this is inefficient - we should really collapse the batch dimensions in the distribution parameters before sampling rather than taking polynomially too many samples and dropping most of them, but that will require quite a bit of distribution plumbing.

pyro/poutine/enumerate_messenger.py
```diff
@@ -5,13 +5,63 @@
...
+
+        thin_sample = thin_sample.transpose(0, squashed_dim1)
+
+        if squashed_dim2 >= len(target_shape) - len(dist.event_shape):
+            continue
+
+        # permute dims to left if necessary
+        thin_sample = thin_sample.transpose(1, squashed_dim2)
+
+        thin_sample = thin_sample.diagonal(dim1=0, dim2=1).unsqueeze(-1)
```
eb8680: "Forgetting" actually happens here

pyro/poutine/enumerate_messenger.py
```diff
@@ -5,13 +5,63 @@
...
+    sample_shape = (num_samples,) if batch_shape == dist.batch_shape else ()
+    fat_sample = dist(sample_shape=sample_shape)  # TODO thin before sampling
+    assert fat_sample.shape == sample_shape + dist.batch_shape + dist.event_shape
+
+    fat_sample = fat_sample.unsqueeze(0) if not sample_shape else fat_sample
+    target_shape = (num_samples,) + batch_shape + dist.event_shape
+
+    # forget particle IDs for all sample_shape dims
+    thin_sample = fat_sample
+    while thin_sample.shape != target_shape:
```
eb8680: Some of the operations in this loop may be triggering unnecessary tensor copies

examples/dmm/dmm.py
```diff
@@ -432,7 +437,7 @@ def do_evaluation():
 
 # parse command-line arguments and execute the main method
 if __name__ == '__main__':
-    assert pyro.__version__.startswith('0.3.4')
+    # assert pyro.__version__.startswith('0.3.4')
```
fritzo: nit: can you please revert before merging

eb8680:
Having thought about this a bit more and looked at memory usage in the DMM and HMM examples, I think this implementation strategy is dubious because the the way it represents the samples and factors uses something like `O(num_variables * num_samples ** max_factor_rank)` memory when it should in theory only use `O(num_variables * num_samples)`.  

Even though the variable elimination computation in `TraceEnum_ELBO` has higher asymptotic spatial complexity (something like `O(num_variables * num_samples ** max_clique_size)`), the constant factor in the sample representation, bounded by the number of PyTorch ops per factor, is high enough to make large numbers of samples too expensive.

I think this could be rescued since the loss computation in `TraceEnum_ELBO` is still basically optimal in memory usage, but we would need to be smarter about the use of `EnumerateMessenger` for dependency tracking.

PR #1979, opened 2019-07-25
add sparse regression example

martinjankowiak:
we port the sparse regression [example](https://github.com/pyro-ppl/numpyro/blob/master/examples/sparse_regression.py) from numpyro to pyro. in contrast to numpyro, where we do HMC inference for the kernel hyperparameters, here we do a MAP estimate. this works better than you might expect. along the way, we show how to use autodeltas to do MAP.

### DIFF
--- a/examples/sparse_regression.py
+++ b/examples/sparse_regression.py
@@ -305,7 +305,6 @@ def main(args):
 if __name__ == '__main__':
     assert pyro.__version__.startswith('0.3.4')
     parser = argparse.ArgumentParser(description='Krylov KIT')
-    parser.add_argument('--inference', type=str, default='map')
     parser.add_argument('--num-data', type=int, default=750)
     parser.add_argument('--num-steps', type=int, default=1000)
     parser.add_argument('--num-dimensions', type=int, default=100)
--- a/examples/sparse_regression.py
+++ b/examples/sparse_regression.py
@@ -8,7 +8,6 @@ import math
 import pyro
 import pyro.distributions as dist
 from pyro import poutine
-from pyro import sample, param
 from pyro.contrib.autoguide import AutoDelta
 from pyro.infer import Trace_ELBO
 from pyro.contrib.autoguide.initialization import init_to_median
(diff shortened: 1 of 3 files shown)


examples/sparse_regression.py
```diff
@@ -0,0 +1,318 @@
...
+            init_losses.append(loss_fn(model, guide, X, Y, hypers).item())
+
+    pyro.set_rng_seed(np.argmin(init_losses))
+    pyro.clear_param_store()
+    guide = AutoDelta(model, init_loc_fn=functools.partial(init_to_median, num_samples=50))
+    with torch.no_grad():
+        loss_fn(model, guide, X, Y, hypers)
+    # we also make sure the initial observation noise is not too large
+    # (otherwise we run the danger of getting stuck in bad local optima during optimization).
+    pyro.param('auto_var_obs').unconstrained().data -= 2.0
```
jpchen: is 2.0 empirically selected?

examples/sparse_regression.py
```diff
@@ -0,0 +1,318 @@
...
+
+    report_frequency = 50
+    print("Beginning MAP optimization...")
+
+    # the optimization loop
+    for step in range(args.num_steps):
+        loss = loss_fn(model, guide, X, Y, hypers) / args.num_data
+        loss.backward()
+        adam.step()
+        adam.zero_grad()
```
jpchen: this should happen before `step()`

examples/sparse_regression.py
```diff
@@ -0,0 +1,318 @@
...
+
+        # we manually reduce the learning rate according to this schedule
+        if step in [100, 300, 700, 900]:
+            adam.param_groups[0]['lr'] *= 0.2
```
jpchen: there's a [scheduler](https://pytorch.org/docs/stable/optim.html?highlight=scheduler#torch.optim.lr_scheduler.MultiStepLR) for this

examples/sparse_regression.py
```diff
@@ -0,0 +1,318 @@
...
+import functools
+
+import numpy as np
+import torch
+import math
+
+import pyro
+import pyro.distributions as dist
+from pyro import poutine
+from pyro import sample, param
```
jpchen: nit: i know this is ported from numpyro but it'd be nice to use `pyro.sample()` as in all the other tutorials

martinjankowiak: ok

examples/sparse_regression.py
martinjankowiak: how else

examples/sparse_regression.py
martinjankowiak: the two are equivalent

examples/sparse_regression.py
martinjankowiak: I wanted to control everything myself and show that's done

martinjankowiak:
I originally had code support but removed it because this is already 300 linez

martinjankowiak:
I originally had code support but removed it because this is already 300 lines

examples/sparse_regression.py
jpchen: hm yeah, in 0.3 or something the tensor `grad`s werent by default initialized to zero so i've always followed the pattern in their tutorials.

fehiepsi:
> because this is already 300 lines

:)) np, I'll review this PR today.

examples/sparse_regression.py
```diff
@@ -0,0 +1,320 @@
...
+Compare to theorem 5.1 in reference [1].
+"""
+@torch.no_grad()
+def compute_posterior_stats(X, Y, msq, lam, eta1, xisq, c, var_obs, jitter=1.0e-4):
+    N, P = X.shape
+
+    # prepare for computation of posterior statistics for singleton weights
+    probe = torch.zeros((2, P, P), dtype=X.dtype, device=X.device)
+    probe[0, ...] = torch.eye(P, dtype=X.dtype, device=X.device)
+    probe[1, ...] = -torch.eye(P, dtype=X.dtype, device=X.device)
```
fehiepsi: +1 another cool observation! It would be nicer if we can make `probe` have shape `(P, 2, P)` to match the pairwise part (it is also easier to think that we are create P probes of shapes `(2, P)`).

examples/sparse_regression.py
```diff
@@ -0,0 +1,320 @@
...
+
+    # generate artificial dataset
+    X, Y, expected_thetas, expected_quad_dims = get_data(N=args.num_data, P=P, S=S,
+                                                         Q=Q, sigma_obs=args.sigma)
+
+    loss_fn = Trace_ELBO().differentiable_loss
+
+    # We initialize the AutoDelta guide (for MAP estimation) with args.num_trials many
+    # initial parameters sampled from the vicinity of the median of the prior distribution
+    # and then continue optimizing with the best performing initialization.
```
fehiepsi: +1

examples/sparse_regression.py
martinjankowiak: done

Issue #1970, opened 2019-07-18
[FR] Gaussian ops for GaussianHMM


fritzo:
It would be nice to generalize the `DiscreteHMM` distribution #1958 to a `GaussianHMM` distribution, as a structured conjugate model #1723 with many applications. I think we can follow [Sarrka & Garcia-Fernandez](https://arxiv.org/pdf/1905.13002.pdf) or simply develop a `torch.matmul()`-like operation for non-normalized rank-deficient Gaussians. We can already perform this op in [funsor](https://github.com/pyro-ppl/funsor/blob/master/funsor/gaussian.py) using syntax like
```py
from funsor import Gaussian
x = Gaussian(...)
y = Gaussian(...)
z = (x + y).reduce(ops.logaddexp, ...)
```
Indeed this could be used to test an initial Pyro version, however the new Pyro op should be both faster and more numerically stable, so that we can later use this op in funsor. The relevant math for `funsor.Gaussian` is in:
- [eager_add_gaussian_gaussian](https://github.com/pyro-ppl/funsor/blob/ce5cdeb/funsor/gaussian.py#L531) to add two gaussians in log-space, including aligning them so that two gaussians over different sets of random variables can fuse to a single gaussian over the union of those sets; and
- [Gaussian.eager_reduce(ops.add, -)](https://github.com/pyro-ppl/funsor/blob/ce5cdeb/funsor/gaussian.py#L432) to marginalize out a subset of the variables.

Note this funsor version is inefficient, equivalent to implementing matrix multiplcation via broadcast-and-reduce
```py
def naive_matmul(x, y):
    return (x.unsqueeze(-1) * y.unsqueeze(-3)).sum(-2)
```
We should use a smarter algorithm for `gaussian_contract()`, maybe exploiting the Woodbury identity.

Here is branch sketching a module pyro.ops.gaussian with an example `Gaussian` class and a `gaussian_contract()` op
https://github.com/pyro-ppl/pyro/compare/gaussian-ops?expand=1

Also not that once `gaussian_contract()` is defined, it [should be easy](https://github.com/dgasmith/opt_einsum/issues/90#issuecomment-508214059) to integrate it with `opt_einsum` using a custom cost_fn.

fritzo:
@fehiepsi do you have some time to try to implement this? Feel free to take over the gaussian-ops branch.

fehiepsi:
Yup, I'll look into this FR. This sounds like an interesting exercise.

[...]

PR #1980, opened 2019-07-26
Implement Gaussian ops

fehiepsi:
This PR implements Gaussian ops as suggested in https://github.com/pyro-ppl/pyro/issues/1970.

The trickiest part is to deal with rank-deficient precision matrix. I have tried considering various representatives of Gaussian but most of them involve taking inverse of precision matrices. My solution is to use canonical representation as in [information filter](https://en.wikipedia.org/wiki/Kalman_filter#Information_filter). To integrate out joined variables, the condition is the precision matrix part restricted on those variables has to be positive definite.

Alternative, we can follow [Sarrka & Garcia-Fernandez](https://arxiv.org/pdf/1905.13002.pdf) but their methods do not keep track of `log_normalizer` term. Instead, they use an additional parallel scan pass, which seems do not applicable outside bayesian filtering context.

TODO:
- [x] implement gaussian tensordot
- ~[ ] generate gaussian tensordot for general contraction~ defer to later PRs
- [x] test gaussian tensordot
- ~[ ] test gaussian contract~ defer to later PRs

### DIFF
--- a/pyro/ops/gaussian.py
+++ b/pyro/ops/gaussian.py
@@ -63,6 +63,7 @@ def gaussian_contract_base(yz, xy, nx):
     #          inv(Px.bb) + inv(Py.bb))
     # Cp(a)q(c) = ...
     # It seems complicated to represent in terms of N((a, c); mz.a, mz.c, ...)
+    # We might follow Proposition 8, https://arxiv.org/abs/1905.13002 here.
     ny = xy.size(-1) - nx
     Ayz, byz, Cyz, etayz, Jyz = mean_precision_to_filtering_parameters(yz.mean, yz.precision, ny)
     Axy, bxy, Cxy, etaxy, Jxy = mean_precision_to_filtering_parameters(xy.mean, xy.precision, nx)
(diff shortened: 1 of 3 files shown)


pyro/ops/gaussian.py
```diff
@@ -0,0 +1,98 @@
...
+
+class Gaussian(object):
+    """
+    Non-normalized Gaussian distribution.
+
+    This represents an arbitrary semidefinite quadratic function, which can be
+    interpreted as a rank-deficient scaled Gaussian distribution. The precision
+    matrix may have zero eigenvalues, thus it may be impossible to work
+    directly with the covariance matrix.
+    """
```
fritzo: Can you add a little `:param:` section, especially defining what `info_vec` is?

pyro/ops/gaussian.py
```diff
@@ -0,0 +1,108 @@
...
+        assert precision.dim() >= 2
+        assert precision.shape[-2:] == info_vec.shape[-1:] * 2
+        self.log_normalizer = log_normalizer
+        self.info_vec = info_vec
+        self.precision = precision
+
+    def dim(self):
+        return self.info_vec.size(-1)
+
+    @property
```
fritzo: We could make this a `@lazy_property` in a follow-up PR.

PR #1981, opened 2019-07-26
Implement GaussianMRF distribution

fritzo:
Addresses #1970 

This implements a `GaussianMRF` that uses time-parallel filtering to compute a differentiable `.log_prob()`.

Note this Markov Random Field model involves an underlying factor graph, whereas the more standard Hidden Markov Model involves conditional normal distributions. This MRF implementation is simpler to implement than an HMM and is closer in code to the `DiscreteHMM`. In practice provides similar modeling capabilities, but requires slightly different interpretation of factors. We may add an additional `GaussianHMM` distribution in the future.

The main parts to review are: new methods of the `Gaussian` class, the `_sequential_gaussian_tensordot()` function, and the `GaussianMRF` class.

## Tested
- [x] unit tests for new `Gaussian` methods
- [x] math test for the recursive `_sequential_gaussian_tensordot()`
- [x] shape tests for `GaussianMRF`
- [x] density tests for `GaussianMRF`

### DIFF
--- a/pyro/distributions/hmm.py
+++ b/pyro/distributions/hmm.py
@@ -1,8 +1,10 @@
 import torch
 from torch.distributions import constraints
 
+from pyro.distributions.torch import MultivariateNormal
 from pyro.distributions.torch_distribution import TorchDistribution
 from pyro.distributions.util import broadcast_shape
+from pyro.ops.gaussian import Gaussian, gaussian_tensordot
 
 
 def _logmatmulexp(x, y):
@@ -36,6 +38,39 @@ def _sequential_logmatmulexp(logits):
             contracted = torch.cat((contracted, logits[..., -1:, :, :]), dim=-3)
         logits = contracted
     return logits.squeeze(-3)
+
+
+def _mvn_to_gaussian(mvn):
+    """
+    Convert a MultivaiateNormal distribution to a Gaussian.
+    """
+    assert isinstance(mvn, torch.distributions.MultivariateNormal)
+    precision = mvn.precision
+    info_vec = precision.matmul(mvn.loc.unsqueeze(-1)).squeeze(-1)
+    log_normalizer = -2 * torch.cholesky(precision).diagonal(dim1=-2, dim2=-1).log().sum(-1)
+    return Gaussian(log_normalizer, info_vec, precision)
+
+
+def _sequential_gaussian_tensordot(gaussian):
+    """
+    Integrates a Gaussian ``x`` whose rightmost batch dimension is time, computes::
+
+        x[..., 0] @ x[..., 1] @ ... @ x[..., T-1]
+    """
+    assert isinstance(gaussian, Gaussian)
+    batch_shape = gaussian.batch_shape[:-1]
+    state_dim = gaussian.precision.size(-1) // 2
+    while gaussian.batch_shape[-1] > 1:
+        time = gaussian.batch_shape[-1]
+        even_time = time // 2 * 2
+        even_part = gaussian[..., :even_time]
+        x_y = even_part.reshape(batch_shape + (even_time // 2, 2))
+        x, y = x_y[..., 0], x_y[..., 1]
+        contracted = gaussian_tensordot(x, y, state_dim)
+        if time > even_time:
+            contracted = Gaussian.cat((contracted, gaussian[..., -1:]), dim=-1)
+        gaussian = contracted
+    return gaussian[..., 0]
 
 
 class DiscreteHMM(TorchDistribution):
@@ -129,4 +164,91 @@ class DiscreteHMM(TorchDistribution):
 
         # Marginalize out final state.
         result = result.logsumexp(-1)
+        return result
+
+
+class GaussianMRF(TorchDistribution):
+    """
+    Temporal Markov Random Field with Gaussian factors for initial, transition,
+    and observation distributions. This adapts [1] to parallelize over time,
+    achieving O(log(time)) parallel complexity. This differs from [1] in that
+    it tracks the log normalizer term to support loss gradients.
+
+    The event_shape of this distribution includes time on the left::
+
+        event_shape = (num_steps,) + observation_dist.event_shape
+
+    This distribution supports any combination of homogeneous/heterogeneous
+    time dependency of ``transition_dist`` and ``observation_dist``. However,
+    because time is included in this distribution's event_shape, the
+    homogeneous+homogeneous case will have a broadcastable event_shape with
+    ``num_steps = 1``, allowing :meth:`log_prob` to work with arbitrary length
+    data::
+
+        event_shape = (1, obs_dim)  # homogeneous + homogeneous case
+
+    **References:**
+
+    [1] Simo Sarkka, Angel F. Garcia-Fernandez (2019)
+        "Temporal Parallelization of Bayesian Filters and Smoothers"
+        https://arxiv.org/pdf/1905.13002.pdf
+
+    :param torch.distributions.MultivariateNormal initial_dist: A distribution
+        over initial states. This should have batch shape broadcastable to
+        ``self.batch_shape``.  This should have event shape ``(hidden_dim,)``.
+    :param torch.distributions.MultivariateNormal transition_dist: A joint
+        distribution factor over a pair of successive time steps. This should
+        have batch shape broadcastable to ``self.batch_shape + (num_steps,)``.
+        This should have event shape ``(2 * hidden_dim,)``.
+    :param torch.distributions.Distribution observation_dist: A joint
+        distribution factor over a hidden and an observed state. This should
+        have batch shape broadcastable to ``self.batch_shape + (num_steps,)``.
+        This should have event shape ``(hidden_dim + obs_dim,)``.
+    """
+    arg_constraints = {}
+
+    def __init__(self, initial_dist, transition_dist, observation_dist, validate_args=None):
+        assert isinstance(initial_dist, MultivariateNormal)
+        assert isinstance(transition_dist, MultivariateNormal)
+        assert isinstance(observation_dist, MultivariateNormal)
+        hidden_dim = initial_dist.event_shape[0]
+        assert transition_dist.event_shape[0] == (hidden_dim, hidden_dim)
+        obs_dim = observation_dist.event_shape[0] - hidden_dim
+
+        shape = broadcast_shape(initial_dist.batch_shape + (1,),
+                                transition_dist.batch_shape,
+                                observation_dist.batch_shape)
+        batch_shape, time_shape = shape[:-1], shape[-1:]
+        event_shape = time_shape + (obs_dim,)
+        super(GaussianMRF, self).__init__(batch_shape, event_shape, validate_args=validate_args)
+        self._init = _mvn_to_gaussian(initial_dist)
+        self._trans = _mvn_to_gaussian(transition_dist)
+        self._obs = _mvn_to_gaussian(observation_dist)
+
+    def expand(self, batch_shape, _instance=None):
+        new = self._get_checked_instance(GaussianMRF, _instance)
+        batch_shape = torch.Size(broadcast_shape(self.batch_shape, batch_shape))
+        # We only need to expand one of the inputs, since batch_shape is determined
+        # by broadcasting all three. To save computation in _sequential_gaussian_tensordot(),
+        # we expand only _init, which is applied only after _sequential_gaussian_tensordot().
+        new._init = self._init.expand(batch_shape)
+        new._trans = self._trans
+        new._obs = self._obs
+        super(GaussianMRF, new).__init__(batch_shape, self.event_shape, validate_args=False)
+        new.validate_args = self.__dict__.get('_validate_args')
+        return new
+
+    def log_prob(self, value):
+        # Combine observation and transition factors.
+        obs = self._obs.condition(value)
+        result = gaussian_tensordot(self._trans, obs, dims=0)
+
+        # Eliminate time dimension.
+        result = _sequential_gaussian_tensordot(result)
+
+        # Combine initial factor.
+        result = gaussian_tensordot(self._init, result, dims=0)
+
+        # Marginalize out final state.
+        result = result.logsumexp()
         return result
--- a/pyro/distributions/hmm.py
+++ b/pyro/distributions/hmm.py
@@ -46,8 +46,9 @@ def _sequential_gaussian_tensordot(gaussian):
         x[..., 0] @ x[..., 1] @ ... @ x[..., T-1]
     """
     assert isinstance(gaussian, Gaussian)
+    assert gaussian.dim() % 2 == 0, "dim is not even"
     batch_shape = gaussian.batch_shape[:-1]
-    state_dim = gaussian.precision.size(-1) // 2
+    state_dim = gaussian.dim() // 2
     while gaussian.batch_shape[-1] > 1:
         time = gaussian.batch_shape[-1]
         even_time = time // 2 * 2
@@ -199,7 +200,7 @@ class GaussianMRF(TorchDistribution):
         assert isinstance(transition_dist, torch.distributions.MultivariateNormal)
         assert isinstance(observation_dist, torch.distributions.MultivariateNormal)
         hidden_dim = initial_dist.event_shape[0]
-        assert transition_dist.event_shape[0] == (hidden_dim, hidden_dim)
+        assert transition_dist.event_shape[0] == hidden_dim + hidden_dim
         obs_dim = observation_dist.event_shape[0] - hidden_dim
 
         shape = broadcast_shape(initial_dist.batch_shape + (1,),
--- a/pyro/distributions/hmm.py
+++ b/pyro/distributions/hmm.py
@@ -231,17 +231,26 @@ class GaussianGaussianMRF(TorchDistribution):
         return new
 
     def log_prob(self, value):
+        # We compute a normalized distribution as p(obs,hidden) / p(hidden).
+        logp_oh = self._trans
+        logp_h = self._trans
+
         # Combine observation and transition factors.
-        result = self._trans
-        result += self._obs.condition(value).event_pad(left=self.hidden_dim)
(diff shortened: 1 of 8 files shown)


pyro/ops/gaussian.py
```diff
@@ -105,4 +179,6 @@ def gaussian_tensordot(x, y, dims=0):
         diff = 0.5 * nb * math.log(2 * math.pi) + 0.5 * Linvb.squeeze(-1).pow(2).sum(-1) - logdet
         log_normalizer = log_normalizer + diff
 
-    return Gaussian(log_normalizer, info_vec, precision)
+    result = Gaussian(log_normalizer, info_vec, precision)
+    assert result.dim() == x.dim() + y.dim() - dims
```
fritzo: @fehiepsi It looks like we missed a bug in the previous review, leading to a failure of this new assertion (which should be the guarantee provided by `tensordot`). Could you take a look?

fehiepsi: @fritzo I think that this should be `x.dim() + y.dim() - 2 * dims`. It is tested in the test.

fritzo: See [tensordot documentation](https://docs.scipy.org/doc/numpy/reference/generated/numpy.tensordot.html). The dims argument specifies how many dims of each of x and y are eliminted, i.e. the amount of overlap.

pyro/ops/gaussian.py
```diff
@@ -34,17 +35,72 @@ def __init__(self, log_normalizer, info_vec, precision):
...
+        """
+        assert self.dim() == other.dim()
+        return Gaussian(self.log_normalizer + other.log_normalizer,
+                        self.info_vec + other.info_vec,
+                        self.precision + other.precision)
+
     def log_density(self, value):
         """
-        Evaluate the log density of this Gaussian at a point value.
+        Evaluate the log density of this Gaussian at a point value::
```
fritzo: @fehiepsi Can you take a look at this change when you get a chance? I believe I have added a missing term, but I haven't been able to get all tests passing.

pyro/ops/gaussian.py
```diff
@@ -54,7 +110,53 @@ def log_density(self, value):
...
+        This should satisfy::
+
+            x.condition(y).dim() == x.dim() - y.size(-1)
+        """
+        assert isinstance(value, torch.Tensor)
+        assert value.size(-1) <= self.info_vec.size(-1)
+
+        # FIXME this is bogus but has the right type:
+        n = self.info_vec.size(-1) - value.size(-1)
+        info_vec = self.info_vec[..., :n]
```
fehiepsi: I guess for conditioning, `info_vec` should be `self.info_vec[..., :n] - self.precision[..., :n, n:].matmul(value.unsqueeze(-1)).squeeze(-1)`

pyro/ops/gaussian.py
```diff
@@ -54,7 +110,53 @@ def log_density(self, value):
...
+        precision = self.precision[..., :n, :n]
+        return Gaussian(self.log_normalizer, info_vec, precision)
+
+    def logsumexp(self):
+        """
+        Integrates out all latent state.
+        """
+        n = self.info_vec.size(-1)
+        return (self.log_normalizer + 0.5 * n * math.log(2 * math.pi) -
+                self.precision.cholesky().diagonal(dim1=-2, dim2=-1).log().sum(-1))
```
fehiepsi: If we integrate out, we'll need to add `0.5 * info_vec @ inv(precision) @ info_vec` term. Basically, we need to return [the diff term here](https://github.com/pyro-ppl/pyro/blob/dev/pyro/ops/gaussian.py#L105) (you can consider `na = nc = 0` in that implementation.

pyro/ops/gaussian.py
```diff
@@ -54,7 +110,53 @@ def log_density(self, value):
...
+    :param ~torch.distributions.MultivariateNormal mvn: A multivariate normal distribution.
+    :return: An equivalent Gaussian object.
+    :rtype: ~pyro.ops.gaussian.Gaussian
+    """
+    assert isinstance(mvn, torch.distributions.MultivariateNormal)
+    n = mvn.loc.size(-1)
+    precision = mvn.precision_matrix
+    info_vec = precision.matmul(mvn.loc.unsqueeze(-1)).squeeze(-1)
+    log_normalizer = (-0.5 * n * math.log(2 * math.pi) +
+                      precision.cholesky().diagonal(dim1=-2, dim2=-1).log().sum(-1))
```
fehiepsi: As above, `0.5 * info_vec @ inv(precision) @ info_vec` is missing here. We might just need to add `-0.5 * (info_vec * mvn.loc).sum(-1)`

pyro/ops/gaussian.py
fehiepsi: I think that the additional term which you added does not depend on `value`, so would be absorbed in `log_normalizer` term.

tests/ops/test_gaussian.py
```diff
@@ -4,23 +4,127 @@
...
+    y = random_gaussian(shape, dim)
+    value = torch.randn(dim)
+    assert_close((x + y).log_density(value), x.log_density(value) + y.log_density(value))
+
+
+@pytest.mark.parametrize("batch_shape", [(), (4,), (3, 2)], ids=str)
+@pytest.mark.parametrize("dim", [1, 2, 3])
+def test_logsumexp(batch_shape, dim):
+    gaussian = random_gaussian(batch_shape, dim)
+    gaussian.info_vec.fill_(0)  # centered
```
fehiepsi: I think we need to set info_vec to nonzero values, so we can detect the missing term in `logsumexp`.

pyro/ops/gaussian.py
fehiepsi: The math would look like
```
-0.5 * (a, b)^T @ precision @ (a, b) + (a, b)^T @ info
= -0.5 * a^T @ P_aa @ a - a^T @ P_ab @ b + a^T @ info_a + ...
= -0.5 * a^T @ P_aa @ a + a^T @ (info_a - P_ab @ b) + ...
```
So I think we'll need to adjust log_normalize according to
```
log_normalize += -0.5 * b^T @ P_bb @ b + b^T @ info_b
```

pyro/distributions/hmm.py
```diff
@@ -130,3 +153,94 @@ def log_prob(self, value):
...
+    def log_prob(self, value):
+        # Combine observation and transition factors.
+        result = self._trans
+        result += self._obs.condition(value).pad(left=self.hidden_dim)
+
+        # Eliminate time dimension.
+        result = _sequential_gaussian_tensordot(result)
+
+        # Combine initial factor.
+        result += self._init.pad(right=self.hidden_dim)
```
fehiepsi: nit: it'll be a bit faster if we combine using tensordot: `gaussian_tensordot(self._init, result, dims=self.hidden_dim)` and then we can marginalize out the final state. The reason is we'll only need to compute cholesky of 2 `hidden_dim x hidden_dim` matrices, instead of one `2hidden_dim x 2hidden_dim` matrix. This way also matches DiscreteHMM version.

pyro/ops/gaussian.py
fritzo: Thanks so much @fehiepsi , this worked perfecty!

tests/ops/test_gaussian.py
fritzo: Hmm, I'm worried that the distribution will lie outside of the `[-5,5]` uniform sampling scheme below. I've weakened this line but not removed it entirely.

fritzo:
@fehiepsi I'm pretty happy with this PR now. The last `test_gaussian_mrf_log_prob` did catch one transpose error. Let me know if you can think of any stronger tests.

In follow-up PRs I'd like to add an example, test jit support, test cuda support, etc.

pyro/distributions/hmm.py
```diff
@@ -38,6 +39,29 @@ def _sequential_logmatmulexp(logits):
     return logits.squeeze(-3)
 
 
+def _sequential_gaussian_tensordot(gaussian):
```
fritzo: This is a line-by-line translation of `_sequential_logmatmulexp()` above.

pyro/distributions/hmm.py
```diff
@@ -90,12 +114,11 @@ def __init__(self, initial_logits, transition_logits, observation_dist, validate
         if len(observation_dist.batch_shape) < 1:
             raise ValueError("expected observation_dist to have at least one batch dim, "
                              "actual .batch_shape = {}".format(observation_dist.batch_shape))
-        time_shape = broadcast_shape((1,), transition_logits.shape[-3:-2],
-                                     observation_dist.batch_shape[-2:-1])
+        shape = broadcast_shape(initial_logits.shape[:-1] + (1,),
```
fritzo: I've simplified this following the pattern used below in `GaussianMRF`.

tests/ops/gaussian.py
```diff
@@ -0,0 +1,42 @@
+import torch
+
+import pyro.distributions as dist
+from pyro.ops.gaussian import Gaussian
+from tests.common import assert_close
+
+
+def random_gaussian(batch_shape, dim, rank=None):
```
fritzo: This test helper is moved from tests/ops/test_gaussian.py. I've changed rank logic so that the default is to use high-rank components, thereby improving the conditioning of `precision`.

pyro/ops/gaussian.py
```diff
@@ -34,28 +35,147 @@ def __init__(self, log_normalizer, info_vec, precision):
...
     def batch_shape(self):
         return broadcast_shape(self.log_normalizer.shape,
                                self.info_vec.shape[:-1],
                                self.precision.shape[:-2])
 
+    def expand(self, batch_shape):
+        n = self.info_vec.size(-1)
+        log_normalizer = self.log_normalizer.expand(batch_shape)
+        info_vec = self.info_vec.expand(batch_shape + (n,))
+        precision = self.precision.expand(batch_shape + (n, n))
```
fehiepsi: I was wondering if we need to expand `info_vec` and `precision` here. I think we can manipulate `reshape` of `info_vec` manually. But I can't find a good solution for this. Something likes
```
def find_unbroadcasted_new_shape(shape, unbroadcasted_shape, new_shape):
    # return expand_shape & unbroadcasted_new_shape
    # where we will expand `info_vec` to `expand_shape`
    # and then reshape it to `unbroadcasted_new_shape`
    # We'll do this work at at `Gaussian.reshape` operator
```
Btw, I would like to defer this to another PR, when we have an example to test out the performance. I just raise this to see what you think?

pyro/ops/gaussian.py
```diff
@@ -34,28 +35,147 @@ def __init__(self, log_normalizer, info_vec, precision):
...
+    :return: An equivalent Gaussian object.
+    :rtype: ~pyro.ops.gaussian.Gaussian
+    """
+    assert isinstance(mvn, torch.distributions.MultivariateNormal)
+    n = mvn.loc.size(-1)
+    precision = mvn.precision_matrix
+    info_vec = precision.matmul(mvn.loc.unsqueeze(-1)).squeeze(-1)
+    log_normalizer = (-0.5 * n * math.log(2 * math.pi) +
+                      -0.5 * (info_vec * mvn.loc).sum(-1) +
+                      precision.cholesky().diagonal(dim1=-2, dim2=-1).log().sum(-1))
```
fehiepsi: nit: because scale_tril is always available in MVN, it is faster to use `- mvn.scale_tril.diagonal(dim1=-2, dim2=-1).log().sum(-1))`

tests/ops/test_gaussian.py
fehiepsi: Thanks! This is also related to the unstable problem of tensordot density test. I'll adopt this to make that test more stable (currently there is a single test case failing in Travis but there might be more for the next PyTorch version).

tests/distributions/test_hmm.py
```diff
@@ -164,3 +185,81 @@ def model(data):
...
+    unrolled_obs = unrolled_obs.event_permute(perm)
+    unrolled_data = data.reshape(data.shape[:-2] + (T * obs_dim,))
+
+    assert init.dim() == hidden_dim
+    assert unrolled_trans.dim() == (1 + T) * hidden_dim
+    assert unrolled_obs.dim() == T * (hidden_dim + obs_dim)
+    joint = gaussian_tensordot(init, unrolled_trans, hidden_dim)
+    joint = gaussian_tensordot(joint, unrolled_obs, T * hidden_dim)
+    expected_log_prob = joint.log_density(unrolled_data)
+    assert_close(actual_log_prob, expected_log_prob)
```
fehiepsi: This test assumes that we trust the math of tensordot. How about we add one more test for the case `obs_dist` is a diagonal block MVN? That implies obs is independent of hidden, so I think that in that case, `expected_log_prob = obs_dist.log_prob(value).sum(-1)`. This helps us detect if there are missing constants.

fritzo: I've added a test like that, but it is a little more complex due to the MRF factors combining to a non-normalized distribution (which wouldn't have been needed if this had been an HMM).

pyro/ops/gaussian.py
fritzo: For cost purposes I've intentionally separated `.reshape()` and `.expand()`, so `.reshape()` may fail if `.expand()` has not been performed first. Similarly, `__getitem__` assumes `.expand()` has been called. It might be possible to keep tensors smaller (as in `._unbroadcasted_*`), but doing so is complex and requires lots of Python manipulation. I'd prefer to defer this optimization until profiling shows that it is worth the added complexity.

tests/distributions/test_hmm.py
fritzo: On second thought, I guess we'd like the `GaussianGaussianMRF` distribution to be normalized (it is not currently normalized). I think we could do this by performing two contractions: the current contraction, and another contraction marginalizing over observations, i.e. `p(obs) = p(obs, hidden) / p(hidden)` WDYT?

fritzo: I think we'll want to normalize. Here's a potential implementation (I'll work on tests):
```py
def log_prob(self, value):
    # We compute a normalized distribution as p(obs,hidden) / p(hidden).
    # Combine observation and transition factors.
    logp_h = self._trans
    logp_oh = self._trans
    logp_oh += self._obs.condition(value).event_pad(left=self.hidden_dim)

    # Concatenate p(hidden) and p(observed,hidden) into a single Gaussian.
    batch_shape = logp_oh.batch_shape
    logp = Gaussian.cat([logp_h.expand(batch_shape),
                         logp_oh.expand(batch_shape)])

    # Eliminate time dimension.
    logp = _sequential_gaussian_tensordot(logp)

    # Combine initial factor.
    logp = gaussian_tensordot(self._init, logp, dims=self.hidden_dim)

    # Marginalize out final state.
    logp = logp.event_logsumexp()
    logp_h, logp_oh = logp[0], logp[1]
    return logp_oh - logp_h  # = log( p(obs,hidden) / p(hidden) )
```

pyro/ops/gaussian.py
```diff
@@ -34,28 +35,176 @@ def __init__(self, log_normalizer, info_vec, precision):
...
+        b = value
+
+        info_vec = info_a - P_ab.matmul(b.unsqueeze(-1)).squeeze(-1)
+        precision = P_aa
+        log_normalizer = (self.log_normalizer +
+                          -0.5 * P_bb.matmul(b.unsqueeze(-1)).squeeze(-1).mul(b).sum(-1) +
+                          b.mul(info_b).sum(-1))
+        return Gaussian(log_normalizer, info_vec, precision)
+
+    def marginalize(self, slice_):
```
fritzo: @fehiepsi I believe this is the final op we need to normalize `GaussianGaussianMRF`. I'm working on the linear algebra now. I've added the two equations below as unit tests. Any help appreciated 🙂

pyro/ops/gaussian.py
fehiepsi: yup, I tried to implement this down this noon but it is a bit complicated to handle various patterns of expand and reshape. Will play more with it to see if I can come up with a simple solution.

pyro/ops/gaussian.py
```diff
@@ -155,25 +155,37 @@ def condition(self, value):
...
-        n = info_vec.size(-1)
+        if left == 0 and right == 0:
+            return self
+        if left > 0 and right > 0:
+            raise NotImplementedError
+        n = self.dim()
+        n_b = left + right
+        a = slice(left, n - right)  # preserved
+        b = slice(None, left) if left else slice(n - right, None)
+        info_vec = self.info_vec[..., a]  # FIXME
```
fehiepsi: I think
```
P_at = P_a.transpose(-1, -2)
b_tmp = b.unsqueeze(-1).triangular_solve(P_b, upper=False).solution
info_vec = self.info_vec[..., a] - torch.matmul(P_at, b_tmp).squeeze(-1)
```

tests/distributions/test_hmm.py
fehiepsi: > p(obs) = p(obs, hidden) / p(hidden)

Agree, this would account for normalization.

pyro/ops/gaussian.py
fritzo: Yes, I implemented this and re-derived your math to verify. However tests still fail 😕 
- Tests pass if we ignore `info_vec` computations using `gaussian.info_vec.fill_(0)` in  the test examples. This suggests that the `precision` and `log_normalier` math is correct.
- Your snippet provides equivalent results to the brute force
  ```
  info_vec = precision.matmul(
      self.precision.inverse().matmul(self.info_vec.unsqueeze(-1))[..., a, :]).squeeze(-1)
  ```
  This suggests that your snippet is correct.

Any ideas what could be wrong?

fehiepsi: Just recognize that you'll need to add `0.5 * b_tmp.squeeze(-1).pow(2).sum(-1)` term to `log_normalizer` too. When `info_vec=0`, this term is zero, so you didn't detect this. :D

fritzo:
Thanks for the detailed review and all of your math help, @fehiepsi !

PR #1985, opened 2019-07-31
basic SVGD implementation

martinjankowiak:
pair-coded with @fritzo.

turned out to be pretty simple thanks to existing `AutoDelta` machinery.

[paper reference](https://arxiv.org/abs/1608.04471)

PR #1989, opened 2019-08-01
move autoguide to core

martinjankowiak:
autoguide is now mature enough to be in core.

this move will also enable me to subclass `AutoDelta` in #1985


docs/source/index.rst
```diff
@@ -30,7 +30,6 @@ Pyro Documentation
    :maxdepth: 2
    :caption: Contributed Code:
 
-   contrib.autoguide
```
martinjankowiak: someone who understands docs please check this stuff

pyro/infer/autoguide/guides.py
```diff
@@ -23,43 +23,14 @@ def model():
...
-    'AutoMultivariateNormal',
-    'init_to_feasible',
-    'init_to_mean',
-    'init_to_median',
-    'init_to_sample',
-    'mean_field_entropy',
-]
-
-
-def _product(shape):
```
martinjankowiak: moved to utils

pyro/infer/autoguide/guides.py
```diff
@@ -779,24 +750,3 @@ def __call__(self, *args, **kwargs):
                 result[name] = pyro.sample(name, discrete_dist, infer={"enumerate": "parallel"})
 
         return result
-
-
-def mean_field_entropy(model, args, whitelist=None):
```
martinjankowiak: moved to utils

fritzo:
I think it would be kind to users if for the 0.4.x releases we retained a file pyro/contrib/autoguide.py with contents:
```py
from pyro.infer.autoguide import *
```
We could then remove this file in the 0.5.0 release. @neerajprad WDYT?

tutorial/source/bayesian_regression.ipynb
```diff
@@ -315,7 +315,7 @@
...
+    "In order to do inference we're going to need a guide, i.e. a variational family of distributions.  We will use Pyro's [autoguide library](http://docs.pyro.ai/en/dev/infer.autoguide.html) to automatically place Gaussians with diagonal covariance on all of the distributions in the model.  Under the hood, this defines a `guide` function with `Normal` distributions with learnable parameters corresponding to each `sample()` in the model.  Autoguide also supports learning MAP estimates with `AutoDelta` or composing guides with `AutoGuideList` (see the [docs](http://docs.pyro.ai/en/dev/infer.autoguide.html) for more information).   In [Part II](bayesian_regression_ii.ipynb) we will explore how to write guides by hand."
```
martinjankowiak: are my doc links ok?

fritzo: yes, these appear to be ok.

pyro/infer/autoguide/guides.py
```diff
@@ -23,43 +23,14 @@ def model():
 import pyro
```
fritzo: It's now a bit dangerous to `import pyro` from within Pyro. A safer alternative would be to import `primitives`, maybe via something like:
```python
from pyro.primitives import sample as pyro_sample
from pyro.primitives import plate as pyro_plate
```
@eb8680 do you have an opinion?

martinjankowiak: can you please elaborate on where exactly the potential danger lies?

martinjankowiak:
> ```python
> # DEPRECATED pyro.contrib.autoguide has moved to pyro.infer.autoguide
> from pyro.infer.autoguide import *
> ```

seems reasonable to me. done

fritzo:
This may pacify flake8:
```
# DEPRECATED pyro.contrib.autoguide has moved to pyro.infer.autoguide
from pyro.infer.autoguide import *  # noqa F403
```

docs/source/index.rst
fritzo: this is correct.

pyro/infer/autoguide/guides.py
fritzo: Hmm, I guess we can `import pyro` in pyro.infer.

What I was worried about was a cycle like pyro -> pyro.infer -> pyro. But I recall that we decided to restrict `pyro` dependencies to: `poutine`, `logger`, `primitives`, and `util`. Thus it should be fine for `infer` to depend on `pyro`.

fritzo: Resolving, since this seems fine.

pyro/contrib/util.py
```diff
@@ -66,17 +66,6 @@ def rtril(M, diagonal=0, upper=False):
     return M*torch.tril(torch.ones(M.shape[-2], M.shape[-1]), diagonal=diagonal)
 
 
-def hessian(y, xs):
```
martinjankowiak: moved to pyro.ops

pyro/ops/hessian.py
```diff
@@ -0,0 +1,17 @@
+import torch
+
+
+def hessian(y, xs):
+    """
```
martinjankowiak: anyone want to add to this docstring? who wrote this? @eb8680

tests/contrib/test_hessian.py
```diff
@@ -0,0 +1,30 @@
+from collections import OrderedDict
+import pytest
+import torch
+
+import pyro.distributions as dist
+from pyro.ops.hessian import hessian
+from tests.common import assert_equal
+
+
+def test_hessian_mvn():
```
martinjankowiak: moved from contrib's test_utils

pyro/ops/hessian.py
fritzo: `git blame` says @fehiepsi wrote `hessian()`

martinjankowiak:
@fritzo looks like tests are going to pass. anything else i should address?

Issue #1257, opened 2018-07-23
Stein variational gradient descent, Stein message passing


fritzo:
1. How can Stein Variational Gradient Descent ([Liu16](https://arxiv.org/pdf/1608.04471.pdf)) be implemented in Pyro? Is this simply a mixture-of-deltas guide, e.g. `AutoMixtureOfDeltas`?
2. Can Pyro's enumeration framework be used to implement Stein Message Passing ([Wan18](http://proceedings.mlr.press/v80/wang18l/wang18l.pdf), [Shou18](http://proceedings.mlr.press/v80/zhuo18a/zhuo18a.pdf))?

fritzo:
cc @OptimusLime

robsalomone:
Hi @fritzo, 

The guide will be a mixture of deltas, however the gradient is not ELBO-based and relies on the computation of Stein Kernels. These kernels, and the resulting gradient of the Kernelized Stein Discrepency which gives the GD direction are reasonably straightforward to compute provided one has access to the score function (grad log joint) over latents. 

I could give this a go myself, but I'm not well versed in how the finer points of the back-end works. Have you considered making a method that creates a function to the log-joint for a model as well as its gradient? 

Perhaps easier said than done, but it would make Pyro a lot more appealing for researchers who develop new inference techniques to prototype and test approaches.

eb8680:
@robsalomone thanks for your input.  Here's one way to get a log-joint function that takes a dictionary `vars` mapping sample site names to sample values as inputs along with the usual inputs to the corresponding model and returns a torch scalar containing the log-joint probability that can be differentiated with `torch.autograd.grad` or `backward`:
```python
def make_log_joint(model):
    def _log_joint(vars, *args, **kwargs):
        tr = poutine.trace(poutine.condition(model, data=vars)).get_trace(*args, **kwargs)
        # make sure all variables in the model are accounted for
        assert all(node["is_observed"] for node in tr.nodes.values() if node["type"] == "sample")
        return tr.log_prob_sum()
    return _log_joint

log_joint_fn = make_log_joint(model)
logp = log_joint_fn({"x": torch.tensor(0.123, requires_grad=True), ...}, ...)
```

We completely agree with the broader point about better support and documentation of internals for research users, and we're planning to prioritize that much more going forward.

robsalomone:
[...]

PR #1991, opened 2019-08-02
upgrade SVGD to allow use of Kernelized Complete Conditional Stein Discrepancies

martinjankowiak:
builds on #1985 

addresses #1257 

[paper reference](https://arxiv.org/abs/1904.04478)


pyro/infer/svgd.py
```diff
@@ -0,0 +1,220 @@
...
+        index = torch.arange(num_particles)
+        norm_sq = norm_sq[index > index.unsqueeze(-1), ...]
+        median = norm_sq.median(dim=0)[0]
+        if self.bandwidth_factor is not None:
+            median = self.bandwidth_factor * median
+        assert median.shape == norm_sq.shape[-1:]
+        return median / math.log(num_particles + 1)
+
+    @torch.no_grad()
+    @copy_docs_from(SteinKernel.log_kernel_and_grad)
```
martinjankowiak: @neerajprad is this invocation ok?

neerajprad: Hmm..I don't think that works at the level of class methods. Why not just use this to annotate the class itself? It will only override for methods that don't have a docstring of their own.

martinjankowiak: thanks, done (i think)

pyro/infer/svgd.py
```diff
@@ -138,78 +130,91 @@ class SVGD(object):
...
         self.max_plate_nesting = max_plate_nesting
+        self.mode = mode
 
         self.loss = Trace_ELBO().differentiable_loss
         self.guide = _SVGDGuide(self.model)
 
     def get_named_particles(self):
         """
-        Get a dictionary of named particles of the form {name: particle}
+        Get a dictionary of named particles of the form {name: particle}. The leading dimension
```
fritzo: nit: the plural in this name is confusing I would suggest replacing this comment with
```
Create a dictionary mapping name to vectorized value, of the form ``{name: tensor}``. The leading dimension of each tensor corresponds to particles. I.e. this creates a struct of arrays.
```

martinjankowiak: done

tests/infer/test_svgd.py
```diff
@@ -12,38 +10,47 @@
 from tests.common import assert_equal
 
 
-@pytest.mark.parametrize("latent_dist", [dist.Normal(torch.zeros(1), torch.ones(1)).to_event(1),
+@pytest.mark.parametrize("latent_dist", [dist.Normal(torch.zeros(2), torch.ones(2)).to_event(1),
                                          dist.LogNormal(torch.tensor([-1.0]), torch.tensor([0.7])).to_event(1),
-                                         dist.Beta(torch.tensor([0.4]), torch.tensor([0.6])).to_event(1)])
-def test_mean_variance(latent_dist):
+                                         dist.LogNormal(torch.tensor(-1.0), torch.tensor(0.7)),
```
fritzo: TODO before next release we should fix this to support empty event shape.

pyro/infer/svgd.py
```diff
@@ -81,51 +77,47 @@ def _bandwidth(self, norm_sq):
...
         """
-        log_kernels, grads = {}, {}
-        for name, param in params.items():
-            log_kernel, grad = self._log_kernel_and_grad(param, bandwidth_factor=bandwidth_factor)
-            log_kernels[name] = log_kernel
-            grads[name] = grad
-        kernel = torch.exp(sum(log_kernels.values()))
-        grads = {name: torch.einsum("ji,ji...->i...", kernel, grad) for name, grad in grads.items()}
-        return kernel, grads
+        :param float bandwidth_factor: Optional factor by which to scale the bandwidth
```
fritzo: I'm not sure this shows up in docs. If not, you could add the following the class documentation:
```
:ivar float bandwidth_factor: Optional factor ...
```

martinjankowiak: thanks done

PR #1994, opened 2019-08-05
Warn if importing pyro.contrib.autoguide rather than pyro.infer.autoguide

fritzo:
Following #1989 , this adds a `DeprecationWarning`

### DIFF
--- a/pyro/contrib/autoguide.py
+++ b/pyro/contrib/autoguide.py
@@ -1,4 +1,5 @@
 import warnings
+
 from pyro.infer.autoguide import *  # noqa F403
 
 warnings.warn("pyro.contrib.autoguide has moved to pyro.infer.autoguide. "
--- a/pyro/contrib/autoguide.py
+++ b/pyro/contrib/autoguide.py
@@ -1,2 +1,6 @@
-# DEPRECATED pyro.contrib.autoguide has moved to pyro.infer.autoguide
+import warnings
 from pyro.infer.autoguide import *  # noqa F403
+
+warnings.warn("pyro.contrib.autoguide has moved to pyro.infer.autoguide. "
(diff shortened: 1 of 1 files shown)


fritzo:
@neerajprad @fehiepsi I was going to suggest making the same change in NumPyro, but it looks like NumPyro no longer supports usage as a Pyro backend via pyro.generic. Is there a plan to re-enable pyro.generic support in NumPyro?

fehiepsi:
Hi @fritzo, I think we can focus on it after the next numpyro release. Right now, the autoguide functionality in numpyro is a bit unstable. We are also going to simplify svi interface, support latent sites with dynamic support, make hmc work with tricky posterior or with large datasets (using autoguide).

neerajprad:
> I was going to suggest making the same change in NumPyro, but it looks like NumPyro no longer supports usage as a Pyro backend via pyro.generic. Is there a plan to re-enable pyro.generic support in NumPyro?

@fritzo - we didn't build out `numpyro.compatibility` module in the first place. It was more of a placeholder to begin with. As @fehiepsi mentioned, the autoguide functionality in NumPyro is quite limited and needs a bit of work before it is mature to get out of `contrib`. We do want to build out a compatibility layer in `pyro.generic`, but my initial thoughts are that it will require some non-trivial code to go from Pyro's stateful classes to NumPyro (and JAX's) functional style, supporting things like `pyro.module`, `plate`, optimizers etc. Some of these things like `plate` support and `module` handler have not yet been developed in NumPyro.

It is nice to be able to refactor and evolve the API for SVI and MCMC, without having to maintain a separate compatibility layer at this point. Once some of these APIs are more mature (by next release) and we have all the relevant functionality, we will prioritize building out `pyro.generic` support.

PR #1997, opened 2019-08-06
Add sequential=True option to mcmc.predictive

### LINKED ISSUE #1995: Add option for sequential prediction to `mcmc.predictive`
Based on this [forum](https://forum.pyro.ai/t/predictive-sampling-with-logistic-regression/1189/5) post, I think a lot of new users of MCMC struggle with using plate handlers to correctly annotate their batch dimensions (which isn't strictly needed for doing inference). 

While this isn't necessarily a requirement for MCMC, it becomes important if users would like to use the `predictive` utility to do predictions after running MCMC. This utility wraps the model in an outermost plate to vectorize predictions. I think we should support a `parallel=False` argument to this utility which by default will run the model forward on a per trace basis and only use `plate` if parallel is set to True. This will make it easier for new users to play around with MCMC who may not be well acquainted with our auto batching semantics.

neerajprad:
Fixes #1995.

### DIFF
--- a/pyro/infer/mcmc/util.py
+++ b/pyro/infer/mcmc/util.py
@@ -470,17 +470,18 @@ def summary(samples, prob=0.9, num_chains=1):
 
 
 def _predictive_sequential(model, posterior_samples, model_args, model_kwargs,
-                           num_samples, return_trace=False, sample_sites=None):
-    traces = []
+                           num_samples, sample_sites, return_trace=False):
+    collected = []
     samples = [{k: v[i] for k, v in posterior_samples.items()} for i in range(num_samples)]
     for i in range(num_samples):
-        traces.append(poutine.trace(poutine.condition(model, samples[i])).get_trace(*model_args, **model_kwargs))
+        trace = poutine.trace(poutine.condition(model, samples[i])).get_trace(*model_args, **model_kwargs)
+        if return_trace:
+            collected.append(trace)
+        else:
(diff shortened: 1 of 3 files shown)


pyro/infer/mcmc/util.py
```diff
@@ -469,6 +469,20 @@ def summary(samples, prob=0.9, num_chains=1):
     print('\n')
 
 
+def _predictive_sequential(model, posterior_samples, model_args, model_kwargs,
+                           num_samples, return_trace=False, sample_sites=None):
+    traces = []
+    samples = [{k: v[i] for k, v in posterior_samples.items()} for i in range(num_samples)]
+    for i in range(num_samples):
+        traces.append(poutine.trace(poutine.condition(model, samples[i])).get_trace(*model_args, **model_kwargs))
```
fehiepsi: I think that in case of `return_trace=False`, we don't want to hold all the traces in a list. For example, consider a model
```
def model(X, y):
    coef = pyro.sample('coef', dist.Normal(0, 1))
    pyro.sample('y', dist.Normal(coef * X, 1), obs=y)
```
We have a new tensor `coef * X` in each trace, which will cause memory problem if X is high dimensional and num_samples is large.

neerajprad: Good point. Let me change this so that we start stacking samples right away if `return_trace=False.`

PR #2001, opened 2019-08-08
Fixes to enable dev branch to move to pytorch-1.2

neerajprad:


tests/infer/test_enum.py
```diff
@@ -2853,6 +2853,7 @@ def guide():
 
 @pytest.mark.parametrize("gate", [0.1, 0.25, 0.5, 0.75, 0.9])
 @pytest.mark.parametrize("rate", [0.1, 1., 3.])
+@pytest.mark.skip(reason="FIXME: regression in pytorch-1.2")
```
fritzo: Let's try to fix this in 0.4, since Paul wanted to use this for rate modeling.

neerajprad: Yes, still looking into this and some jit failures.

neerajprad:
> can you also change the wheel here

Done.

> i assume you tested on cuda too?

Not yet, will do that separately.

pyro/distributions/zero_inflated_poisson.py
```diff
@@ -27,13 +27,13 @@ def log_prob(self, value):
         gate, rate, value = broadcast_all(self.gate, self.rate, value)
         log_prob = (-gate).log1p() + (rate.log() * value) - rate - (value + 1).lgamma()
         zeros = value == 0
-        log_prob[zeros] = (gate[zeros] + log_prob[zeros].exp()).log()
+        log_prob = torch.where(zeros, (gate + log_prob.exp()).log(), log_prob)
```
neerajprad: This is the place where autograd was complaining about in place mutation of variable that requires grad.

neerajprad:
This should be good to review now.

pyro/distributions/zero_inflated_poisson.py
fehiepsi: Your change looks more reasonable to me.

PR #2003, opened 2019-08-09
Remove obsolete torch patches

fritzo:
##Tested

- [x] Test `Transform.__init__()` patch
- [x] Test `Categorical.sample()` patch
- [x] Test dirichlet grad patch on CUDA
- [x] Test `linspace` patch on CUDA

### DIFF
--- a/tests/distributions/test_torch_patch.py
+++ b/tests/distributions/test_torch_patch.py
@@ -13,4 +13,4 @@ def test_dirichlet_grad_cuda():
 @requires_cuda
 def test_linspace():
     x = torch.linspace(-1., 1., 100, device="cuda")
-    assert x.device == "cuda"
+    assert x.device.type == "cuda"
--- a/tests/distributions/test_torch_patch.py
+++ b/tests/distributions/test_torch_patch.py
@@ -1,0 +1,16 @@
+import torch
+
+import pyro.distributions as dist
+from tests.common import requires_cuda
+
+
+@requires_cuda
+def test_dirichlet_grad_cuda():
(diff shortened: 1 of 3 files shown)


fritzo:
@neerajprad could you check that the dirichlet and linspaces patches work on CUDA (my drivers are too old for PyTorch 1.2, and it will take me a while to upgrade)? I think the linspace test requires a test that does not sloppily `torch.set_default_tensor_type("torch.cuda.FloatTensor")`.

fritzo:
@neerajprad I've tried to add some regression tests. On a machine with GPU, you can simply
```sh
pytest -v tests/distributions/test_torch_patch.py
```

neerajprad:
Awesome, thanks Fritz! I'll just run the tests and confirm shortly.

tests/distributions/test_torch_patch.py
```diff
@@ -0,0 +1,16 @@
...
+@requires_cuda
+def test_dirichlet_grad_cuda():
+    concentration = torch.ones(3, requires_grad=True)
+    dist.Dirichlet(concentration).rsample().sum().backward()
+
+
+@requires_cuda
+def test_linspace():
+    x = torch.linspace(-1., 1., 100, device="cuda")
+    assert x.device == "cuda"
```
neerajprad: This works correctly on the GPU, but we will need to test for `x.device.type == "cuda"` since device is a python object now.

fritzo: Done, let me know if there is anything else, or feel free to push to this branch.

PR #2004, opened 2019-08-09
Add option for lax testing that ignores AssertionError

neerajprad:
I have found it hard to keep our test suite up to date for CUDA, but have noticed that almost all our failures are assertion failures, where the same test on CUDA violates the set precision thresholds. This adds a `--lax` flag to pytest so that assertion failures are `xfailed` dynamically. This makes it easier for me to run all the tests on CUDA and examine the most serious failures (e.g. due to mixed device types, unsupported operation on CUDA, etc.). 

A further enhancement to this would be to not add xfail for tests that are marked with `@requires_cuda`. Another possibility might be to dynamically relax the precision in `assert_equal` / `assert_close` when some environment variable is set.


neerajprad:
Please ignore the `codecov` report failure; I have removed the plugin but it is still showing for this PR.

PR #2005, opened 2019-08-09
Support Independent(Normal) observations in GaussianHMM

fritzo:
This adds support for diagonal observation covariances in `GaussianHMM`.

My use case is an HMM with very large observation space, say latent dimension of M=32 but observed dimension of N=2000. This PR reduces complexity from `M^3+N^3` to `M^3+MN`.

## Tested
- unit test for `matrix_and_mvn_to_gaussian()` with `AffineNormal`
- added diag versions of shape tests to `GaussianHMM` tests

### DIFF
--- a/pyro/ops/gaussian.py
+++ b/pyro/ops/gaussian.py
@@ -7,7 +7,7 @@ from torch.nn.functional import pad
 from pyro.distributions.util import broadcast_shape
 
 
-class Gaussian(object):
+class Gaussian:
     """
     Non-normalized Gaussian distribution.
 
@@ -213,7 +213,7 @@ class AffineNormal:
     variable ``Y`` whose mean is an affine function of a random variable ``X``.
     The likelihood of ``X`` is thus::
 
-        LinearNormal(matrix, loc, scale).condition(y).log_density(x)
+        AffineNormal(matrix, loc, scale).condition(y).log_density(x)
 
     which is equivalent to::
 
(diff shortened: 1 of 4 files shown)


pyro/ops/gaussian.py
```diff
@@ -207,6 +207,49 @@ def event_logsumexp(self):
                 chol_P.diagonal(dim1=-2, dim2=-1).log().sum(-1))
 
 
+class AffineNormal:
```
fehiepsi: nit: do we need to subclass `object` here?

pyro/ops/gaussian.py
```diff
@@ -207,6 +207,49 @@ def event_logsumexp(self):
...
+    """
+    Represents a conditional diagonal normal distribution over a random
+    variable ``Y`` whose mean is an affine function of a random variable ``X``.
+    The likelihood of ``X`` is thus::
+
+        LinearNormal(matrix, loc, scale).condition(y).log_density(x)
```
fehiepsi: nit: LinearNormal -> AffineNormal
You might want to use the name `MultivariateNormalDiag` as in tensorflow (though I don't like this idea much).

pyro/ops/gaussian.py
fritzo: Python 3 no longer requires `(object)`. I'll also remove `(object)` from `Gaussian` in this file, but let's do general cleanup in a separate PR.

pyro/ops/gaussian.py
fritzo: Done.

I agree that `AffineNormal` is a weird name, but I don't think `MultivariateNormalDiag` is the right name either. This class really serves a very narrow role as a likelihood of a diag normal observation of an affine transformation. I'm open to renaming / refactoring once we see whether and how this pattern generalizes.

fritzo:
Thanks for reviewing, @fehiepsi !

PR #2008, opened 2019-08-09
Bump up Pyro version to 0.4.0

neerajprad:


neerajprad:
> is the torchvision version for the examples still correct?

You mean the minimum version in setup.py?

Issue #1802, opened 2019-04-03
Support Viterbi Algorithm for Markov Models


createmomo:
The tutorial of DMM (Deep Markov Model) is great!
https://pyro.ai/examples/dmm.html#

I think there may be much more people, not only me, would like to see how to use the model to infer the most likely latent state sequence that has the maximum likelihood by using the Viterbi decoder. In addition, the decoder is implemented in the Pyro style.

I can only found the infer_discrete in the document. But it looks not straight forward how to use this function to do the Viterbi algorithm on a Markov model, given a transition and an emission distribution. 

I could use Pytorch to implement one for myself. But I believe, there could be one much better way to implement it using Pyro. Thus, it would be great if you could add an official implementation of this algorithm for infer sequential labels, or add one more code snippet at the end of this tutorial to show how to use the infer_discrete function to do the algorithm.

Many thanks,

fritzo:
You can access the `Viterbi` algorithm by setting `temperature=0` in infer-discrete. See the [example](http://docs.pyro.ai/en/0.3.0-release/inference_algos.html#pyro.infer.discrete.infer_discrete) in the `infer_discrete` documentation. You're welcome to turn that little example into a more full-fledged tutorial, especially if you have good demo problem in mind.

bcjonescbt:
Upvote here. For those of us without a lot of Pyro experience, there are a lot of bridges to cross to infer the state sequence. Something showing a more substantial example that correctly uses `infer_discrete` would be very helpful, particularly with enumeration involved. The example in the linked block is pretty light.

fritzo:
I'll try to get to this soon. In particular I'd like to show how to train with `DiscreteHMM` then predict with `infer_discrete`.

adamcweiner:
Are there any updates on this? I've built my own HMM but haven't been able to get state assignments (or state assignment probabilities) with `infer_discrete` or a `full_guide` which are recommended [here](https://pyro.ai/examples/gmm.html#Serving-the-model:-predicting-membership).

fritzo:
@adamcweiner As [pointed out by dave on the forum](https://forum.pyro.ai/t/access-latent-discrete-variables/2555/2?u=fritzo) you can check out the [enumeration tutorial](https://pyro.ai/examples/enumeration.html#Examining-discrete-latent-states).

PR #2010, opened 2019-08-10
Implement *HMM.filter() methods for forecasting

fritzo:
This adds `Discrete-` and `GaussianHMM.filter()` methods for for forecasting.

Currently the `DiscreteHMM` and `GaussianHMM` methods are fast for training but cannot be used for forecasting, i.e. for sampling posterior observations `x[t],...,x[t+w-1] | x[0],...,x[t-1]`. The solution proposed in this PR is to:
1. use `*HMM.log_prob()` to train parameters; then
2. use `*HMM.filter()` to compute a posterior `z[t-1] | x[0],...,x[t-1]`; and finally
3. use a standard sequential Pyro model to sample `z[t],...,z[t+w-1]` and `x[t],...,x[t+w-1]` from the posterior.

This solution benefits from parallel scan in the filter (typically long-window `[0,...,t-1]`), but does not use parallel scan for the forecast (typically short-window `[t,...,t+w-1]`) forecast. However that forecast phase usually benefits from parallelization over samples rather than time.

@martinjankowiak `funsor.pyro.SwitchingLinearHMM` is implemented in https://github.com/pyro-ppl/funsor/pull/191

## Documentation

- added docstrings
- I plan to add examples to both the viterbi tutorial #1802 and Pyro [workshop tutorials](https://github.com/pyro-ppl/sandbox/pull/3)

## Tested
- added shape tests (math is nearly identical to `.log_prob()`)

### DIFF
--- a/pyro/distributions/hmm.py
+++ b/pyro/distributions/hmm.py
@@ -1,6 +1,7 @@
 import torch
 from torch.distributions import constraints
 
+from pyro.distributions.torch import Categorical, MultivariateNormal
 from pyro.distributions.torch_distribution import TorchDistribution
 from pyro.distributions.util import broadcast_shape
 from pyro.ops.gaussian import Gaussian, gaussian_tensordot, matrix_and_mvn_to_gaussian, mvn_to_gaussian
@@ -135,7 +136,7 @@ class DiscreteHMM(TorchDistribution):
         new.transition_logits = self.transition_logits
         new.observation_dist = self.observation_dist
         super(DiscreteHMM, new).__init__(batch_shape, self.event_shape, validate_args=False)
-        new.validate_args = self.__dict__.get('_validate_args')
+        new._validate_args = self.__dict__.get('_validate_args')
         return new
 
     def log_prob(self, value):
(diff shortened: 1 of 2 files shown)


pyro/distributions/hmm.py
```diff
@@ -135,7 +136,7 @@ def expand(self, batch_shape, _instance=None):
         new.transition_logits = self.transition_logits
         new.observation_dist = self.observation_dist
         super(DiscreteHMM, new).__init__(batch_shape, self.event_shape, validate_args=False)
-        new.validate_args = self.__dict__.get('_validate_args')
+        new._validate_args = self.__dict__.get('_validate_args')
```
fritzo: fixes a typo

tests/distributions/test_hmm.py
```diff
@@ -95,15 +95,22 @@ def test_discrete_hmm_shape(ok, init_shape, trans_shape, obs_shape, event_shape,
...
+    d = dist.DiscreteHMM(init_logits, trans_logits, obs_dist)
+
+    actual = d.log_prob(data)
+    expected_shape = broadcast_shape(init_shape, trans_shape[:-1], obs_shape[:-1])
+    assert actual.shape == expected_shape
+    check_expand(d, data)
+
+    final = d.filter(data)
+    assert isinstance(final, dist.Categorical)
+    assert final.batch_shape == d.batch_shape
```
martinjankowiak: add assert for `support` to make sure hidden_dim is right?

fritzo: done.

PR #2014, opened 2019-08-14
Address cholesky issue for potential function

### LINKED ISSUE #2009: Cholesky issue in HMC
In PyTorch, if Cholesky issue happens, a Runtime Error will be raised (instead of returning `nan`) regarding singular matrix. So it will be difficult to run MCMC with gaussian process kernels (which involves Cholesky decomposition of covariance matrices). This issue is raised in https://github.com/pyro-ppl/pyro/issues/1863 and #1973. Because `nan` is a valid potential energy in HMC, we should add a mechanism to catch this error.

My proposed solution is to add a try, catch exception to potential function so that when Cholesky issue happens, it will return `nan`.

fehiepsi:
This PR fixes #2009 and related issues.

TODO
- [x] ~test with jit_compile=True in stochastic volatility model~ that model's potential_fn is not jittable yet.
- [x] test the catch in GPU
- [x] add a simple test to make sure that we catch the error in future versions of pytorch. The simplest one is a MVN model with a fixed large step_size.

### DIFF
--- a/pyro/ops/integrator.py
+++ b/pyro/ops/integrator.py
@@ -70,7 +70,16 @@ def potential_grad(potential_fn, z):
     z_keys, z_nodes = zip(*z.items())
     for node in z_nodes:
         node.requires_grad_(True)
-    potential_energy = potential_fn(z)
+    try:
+        potential_energy = potential_fn(z)
+    # deal with singular matrices
+    except RuntimeError as e:
+        if "singular U" in str(e):
+            grads = {k: v.new_zeros(v.shape) for k, v in z.items()}
+            return grads, z_nodes[0].new_tensor(float('nan'))
+        else:
+            raise e
+
     grads = grad(potential_energy, z_nodes)
     for node in z_nodes:
         node.requires_grad_(False)
--- a/tests/infer/mcmc/test_hmc.py
+++ b/tests/infer/mcmc/test_hmc.py
@@ -297,3 +297,19 @@ def test_unnormalized_normal(kernel, jit):
     posterior = torch.stack([sample["z"] for sample in posterior])
     assert_close(torch.mean(posterior), true_mean, rtol=0.05)
     assert_close(torch.std(posterior), true_std, rtol=0.05)
+
+
+@pytest.mark.parametrize("jit", [False, mark_jit(True)], ids=jit_idfn)
+def test_singular_matrix_catch(jit):
+    def potential_energy(z):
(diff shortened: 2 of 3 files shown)


tests/infer/mcmc/test_hmc.py
```diff
@@ -297,3 +297,23 @@ def potential_energy(params):
...
+        return op(z['cov']).sum()
+
+    init_params = {'cov': torch.eye(3)}
+    potential_fn = potential_energy if not jit else torch.jit.trace(potential_energy, init_params)
+    hmc_kernel = HMC(potential_fn=potential_fn, adapt_step_size=False,
+                     num_steps=10, step_size=1e-20)
+    hmc_kernel.initial_params = init_params
+    hmc_kernel.setup(warmup_steps=0)
+    # setup an invalid cache to trigger singular error for torch.inverse
+    hmc_kernel._cache({'cov': torch.ones(3, 3)}, torch.tensor(0.), {'cov': torch.zeros(3, 3)})
```
fehiepsi: I don't know how to construct a potential_energy which triggers singular error for torch.inverse with an valid initial param. So I use this hack to setup an invalid initial param.

pyro/ops/integrator.py
```diff
@@ -70,7 +70,16 @@ def potential_grad(potential_fn, z):
     z_keys, z_nodes = zip(*z.items())
     for node in z_nodes:
         node.requires_grad_(True)
-    potential_energy = potential_fn(z)
+    try:
+        potential_energy = potential_fn(z)
+    # deal with singular matrices
+    except RuntimeError as e:
+        if "singular U" in str(e):
```
neerajprad: Is this the only way to detect if the matrix was singular? We'll need to change this check if the error message changes, though I think that's unlikely to.

fehiepsi: The error messages are:
```
inverse_cpu, U(4, 4) is zero, singular U
```
where `inverse` can be replaced by `cholesky`, `cpu` can be replaced by `gpu`. The general format is [at this line](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/LinearAlgebraUtils.h#L170). So I pick the common one: singular U

> We'll need to change this check if the error message changes

I think the test will detect the new error message and we can update this check then.

neerajprad: > I think the test will detect the new error message and we can update this check then.

Sounds good!

Issue #2017, opened 2019-08-16
Cholesky error in GaussianHMM, _sequetial_gaussian_tensordot()


fritzo:
I'm seeing cholesky errors in `_sequential_gaussian_tensordot()` when computing `GaussianHMM.log_prob()`, including precision matrices with very negative eigenvalues. I'd like to figure out how to numerically stabilize `gaussian_tensordot()`.

## To reproduce
```py
import torch
from pyro.distributions.hmm import _sequential_gaussian_tensordot
gaussian = torch.load("gaussian.pkl")  # file is attached to this issue
_sequential_gaussian_tensordot(gaussian)
```

[gaussian.pkl.zip](https://github.com/pyro-ppl/pyro/files/3507641/gaussian.pkl.zip)

fritzo:
So far this appears to be a deficiency in `torch.distributions.constraints.lower_cholesky` whereby the optimized parameters become singular. My immediate workaround was to switch from full-covariance parameters to diagonal-covariance parameters. However we should look into more stable ways of optimizing positive definite matrices. I'll leave this issue open for discussion.

fehiepsi:
@fritzo Your point is interesting! I tried to see how that issue comes
```python
import torch
torch.manual_seed(2)
x = torch.randn(40, 40).tril(-1) + torch.diag_embed(torch.randn(40).exp())
print("tril diag\n", x.diag())
print("cov eigen values\n", x.matmul(x.T).eig()[0][:, 0])
x_inv_stable = torch.eye(40).triangular_solve(x, upper=False).solution
print("tril stable inverse\n", x_inv_stable.diagonal(dim1=-2, dim2=-1))
x_inv = x.inverse()
print("tril inverse\n", x_inv.diagonal(dim1=-2, dim2=-1))
precision = x_inv.transpose(-2, -1).matmul(x_inv)
precision_stable = x_inv_stable.transpose(-2, -1).matmul(x_inv_stable)
print(precision_stable.cholesky()) . # singular error
print(precision.cholesky())  # singular error
```
which returns
```
tril diag
 tensor([ 1.8164,  0.9008,  0.5388,  2.4168,  0.4780,  0.2259,  0.3260,  8.2935,
         0.7697,  2.7750,  1.0188,  0.1855,  0.4467,  0.6841,  1.8005,  0.8780,
         0.3297,  2.1109,  0.1622,  0.0466,  0.1453,  4.5444, 11.1870,  0.3835,
[...]

PR #2019, opened 2019-08-19
Represent Gaussian using square root of precision

fehiepsi:
This is an effort to resolve Cholesky issue as reported in #2017.

There is some work during the transition to this representation but I hope that it will be smooth.

TODO:
- [ ] Convert `marginalize` method to use `prec_sqrt`
- [ ] Convert `matrix_and_mvn_to_gaussian`, `mvn_to_gaussian` to this new representation. It is best to have a MVN distribution parameterized by `prec_sqrt` instead of using `torch.inverse(scale_tril)`, but we can defer this to future improvements.
- [ ] Implement gaussian_tensordot (instead of using `pad + add + marginalize`).
- [ ] Add tests for this new representation.

### DIFF
--- a/pyro/ops/gaussian_sqrt.py
+++ b/pyro/ops/gaussian_sqrt.py
@@ -1,0 +1,370 @@
+import math
+
+import torch
+from torch.distributions.utils import lazy_property
+from torch.nn.functional import pad
+
+from pyro.distributions.util import broadcast_shape
+
+
+class GaussianS:
+    """
+    Non-normalized Gaussian distribution.
+
+    This represents an arbitrary semidefinite quadratic function, which can be
+    interpreted as a rank-deficient scaled Gaussian distribution. The precision
+    matrix may have zero eigenvalues, thus it may be impossible to work
+    directly with the covariance matrix.
+
+    :param torch.Tensor log_normalizer: a normalization constant, which is mainly used to keep
+        track of normalization terms during contractions.
+    :param torch.Tensor info_vec: information vector, which is a scaled version of the mean
+        ``info_vec = precision @ mean``. We use this represention to make gaussian contraction
+        fast and stable.
+    :param torch.Tensor prec_sqrt: square root of precision matrix of this gaussian.
+    """
+    def __init__(self, log_normalizer, info_vec, prec_sqrt):
+        # NB: using info_vec instead of mean to deal with rank-deficient problem
+        assert info_vec.dim() >= 1
+        assert prec_sqrt.dim() >= 2
+        assert prec_sqrt.shape[-2:] == info_vec.shape[-1:] * 2
+        self.log_normalizer = log_normalizer
+        self.info_vec = info_vec
+        self.prec_sqrt = prec_sqrt
+
+    def dim(self):
+        return self.info_vec.size(-1)
+
+    @lazy_property
+    def batch_shape(self):
+        return broadcast_shape(self.log_normalizer.shape,
+                               self.info_vec.shape[:-1],
+                               self.prec_sqrt.shape[:-2])
+
+    @lazy_property
+    def precision(self):
+        return self.prec_sqrt.matmul(self.prec_sqrt.transpose(-2, -1))
+
+    def expand(self, batch_shape):
+        n = self.dim()
+        log_normalizer = self.log_normalizer.expand(batch_shape)
+        info_vec = self.info_vec.expand(batch_shape + (n,))
+        prec_sqrt = self.prec_sqrt.expand(batch_shape + (n, n))
+        return GaussianS(log_normalizer, info_vec, prec_sqrt)
+
+    def reshape(self, batch_shape):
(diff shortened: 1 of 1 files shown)


fehiepsi:
@fritzo The square root representation works smoothly and I believe that it will be more stable than the current representation. To make it easier to compare and test, I separated the implementation into a new file. In addition, I added the `S` notation to class and functions in this file to distinguish this implementation and the current implementation. Do you have other suggestion to name this new class and the corresponding functions.

Here are several properties of this representation:
+ event_permute will permute the row vectors of prec_sqrt matrix
+ add two gaussians will concatenate the column vectors of their prec_sqrt matrices
+ event_pad will add more zero rows to prec_sqrt matrix

I believe this version will be faster. It would be more stable (and faster) if use `pyro.param('prec_sqrt', constraint=lower_cholesky)` and use `prec_sqrt` instead of `scale_tril` in MVN implementation.

fehiepsi:
Btw, just realize that this is PR #2019 ! :D

pyro/ops/gaussian_sqrt.py
```diff
@@ -0,0 +1,385 @@
...
+        return broadcast_shape(self.log_normalizer.shape,
+                               self.info_vec.shape[:-1],
+                               self.prec_sqrt.shape[:-2])
+
+    @lazy_property
+    def precision(self):
+        if self.rank() == 0:
+            empty = self.prec_sqrt.new_tensor([])
+            return empty.reshape(self.prec_sqrt.shape[:-1] + (self.dim(),))
+        return self.prec_sqrt.matmul(self.prec_sqrt.transpose(-2, -1))
```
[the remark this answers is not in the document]
fehiepsi: Here we need to return empty tensor, not zeros tensor. I think it is used to pass tests for constant Gaussian.

pyro/ops/gaussian_sqrt.py
```diff
@@ -0,0 +1,385 @@
...
+def triangularize(A, method='cholesky'):
+    """
+    Transforms a matrix to a lower triangular matrix such that A @ A.T = f(A) @ f(A).T
+
+    :param torch.Tensor A: input matrix
+    :param str method: either 'cholesky' or 'qr'
+    """
+    # TODO: benchmark two versions in real application, using 'qr' will be more stable but slower
+    if method == 'cholesky':
+        return A.matmul(A.transpose(-1, -2)).cholesky()
```
fehiepsi: @fritzo The operator `A.matmul(A.T)` will remove all the expanded rank at each step. Assume that `A` has shape `n x 4n` (which is the case when using `tensordot`), then `A.matmul(A.T)` will have shape `n x n`. I will add a test to guarantee that the square root's shape does not expanded in tensordot op.

fehiepsi: Similarly, QR of A.T where `A.shape = n x 4n` has shape `n x n`.

fritzo: But you're not currently triangularizing `Psqrt_a` in `.marginalize()`. I think to achieve bounded rank, you'll need to manually triangularize at the end of `.marginalize()`:
```py
if prec_sqrt.size(-1) > prec_sqrt.size(-2):
    prec_sqrt = triangularize(prec_sqrt)
```

pyro/ops/gaussian_sqrt.py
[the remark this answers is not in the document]
fritzo: `.precision` should be the zeros tensor, not the empty tensor.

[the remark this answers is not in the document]
fehiepsi: You are right. I thought we are multiplying two empty tensor, which will return an empty tensor. :D

pyro/ops/gaussian_sqrt.py
fehiepsi: You are right! I missed this point. Let me see if there is an easy way to achieve this. :(

fehiepsi:
@fritzo Applying at the end of `.marginalize()` is correct but it seems that we have to use `qr` (which is more expensive than Cholesky) there because there is no requirement that the returned Gaussian has full rank. I want to consider other possibilities so I'll revisit this issue after numpyro 0.2 and GaussianGamma PR. Thanks for pointing out that `rank` issue!

fehiepsi:
@fritzo It seems that to marginalize out `x` in an (x,y) Gaussian pair, we can just simply triangularize the `prec_sqrt` part and keep the bottom right block. This way, no need for using cholesky during marginalization and the rank is also bounded. More precisely, given a square root matrix,
```
[[A 0]
 [B C]]
```
the marginalized precision is (B@Bt+C@Ct) - B@At@(A@At)^-1@A@Bt, which is the same as C@Ct if we assume that A is invertible (which is required to be true if we want to marginalize `x` out).

This is the main trick used in square root information filters, there only QR is needed.

I can complete this work if this is needed by someone.

PR #2022, opened 2019-08-21
Add a pyro.factor primitive

### LINKED ISSUE #1454: [FR] Make it easier to add an ad hoc log prob term?
Current solutions to adding a custom log prob term include (1) subclassing `ELBO` #1059 #958, and (2) using a bogus statement
```py
pyro.sample(name, dist.Bernoulli(logits=custom_term), obs=torch.tensor(1.))
```
By contrast STAN  makes it easy to add custom loss terms.

Should we add a helper that adds a custom loss term, effectively adding a non-normalized log-prob via a hidden observe statement?

fritzo:
Resolves #1454 

This implements a `pyro.factor(name, log_factor)` primitive, possibly the most-requested feature we've avoided adding. Previous workarounds include `Bernoulli` observations or `Delta` observations with a log_density argument (as suggested by @fehiepsi).

## Design

This PR implements a more transparent cheaper alternative:
- Native `pyro.factor` statements, and
- a non-normalized `dist.Unit` distribution with empty data.

Alternatives include a `Delta` or `Bernoulli` observation, but both incur nontrivial overhead due to tensor ops.

## Tested

- added unit tests for `Unit`
- added autoguide tests for `pyro.factor`
- examples/rsa refactoring is covered by existing tests
- pyro.contrib.gp refactoring is covered by existing tests
- manually verified RSA examples and tutorials produce identical results before and after this PR


fritzo:
> do the RSA examples and tutorials produce the same results?

Yes (after repairing bitrot in the tutorials)

jpchen:
is the `log_density` kwarg in Delta still needed?

fritzo:
@jpchen yes the `log_density` kwarg in `Delta` still needed. You could have answered this by grepping.

jpchen:
ah woops i missed easy_guide 😅, i only looked in our examples tutorials and tests. there are a handful of pyro_models that need to be fixed after this PR lands as well.

PR #2025, opened 2019-08-25
Implement GaussianGamma op

fehiepsi:
This is a step towards having StudentT filters (see https://github.com/pyro-ppl/funsor/issues/190).

The main reference which I use is [1], which constructs an exact filter for scaled mixtures of normals and elliptically contoured distributions (which are more general than StudentT). I am not sure if this approach will work for us but it is the most promising one which I know. The idea is to introduce an auxiliary Gamma variable `s`, which we will keep updating the posteriors `p(x | s, y)` and `p(s | y)` during filtering process.

My intuition is to mimic how StudentT filter works
```
1. p(x_t-1 | y1,...,y_t-1, s) = Normal(m, sC), where p(s|y1,...y_t-1) = Gamma(a, b)
2. p(xt | y1,...,y_t-1, s) = Normal(..., sD)
3. p(yt | y1,...,y_t-1, s) = Normal(..., sE)
4. p(xt | y1,...,yt, s) = Normal(..., sF)
5. p(s | y1,...,yt) ~ p(s | y1,...,yt-1).p(yt=yt | y1...,y_t-1, s) = Gamma(a', b')
```
where steps 1-4 follow Kalman filter predict and update.

### Reference
1. [Bayesian Kalman Filtering with Elliptically Contoured Errors](https://www.jstor.org/stable/2336969?seq=1#page_scan_tab_contents)

### Done
- Implement trivial event methods: pad, permute, add

### Tasks (ordered according to its complexity)
- [x] Implement event_logsumexp
- [x] Implement condition
- [x] Implement marginalize
- [x] Add tests for the above methods
- [x] Implement matrix_and_mvt_to_GaussianGamma. An auxiliary Gamma variable `s` will be introduced so that p(x, s) is GaussianGamma and the marginal p(x) is StudentT.

Follow-up PR (this PR is already large)
- ~Test against the formula in the paper
[Student’s t-Filters for Noise Scale Estimation](https://users.aalto.fi/~ssarkka/pub/SPL2019.pdf)~ Edit: use `skip_multiplier_prior` flag for that model, no need to test because we can implement the same model

Hopefully all pieces will stick together. So we can use temporal parallelization for StudentT filter.


fehiepsi:
All tests pass now. It is time to test against the formula in the paper
[Student’s t-Filters for Noise Scale Estimation](https://users.aalto.fi/~ssarkka/pub/SPL2019.pdf)

fehiepsi:
I just added an option to make it easier to implement Student-t filter following the model (2) of the above paper. Now, an implementation for that model works as follows
```
# model p(x_0, s)
self._init = mvt_to_gaussian_gamma(initial_dist)
# model p(x_n+1 | x_n, s)
self._trans = matrix_and_mvt_to_gaussian_gamma(
    transition_matrix, transition_dist, skip_multiplier_prior=True)
# model p(y_n | x_n, s)
self._obs = matrix_and_mvt_to_gaussian_gamma(
    observation_matrix, observation_dist, skip_multiplier_prior=True)
```
then process exactly the same as GaussianHMM (but use `sequential_gaussian_gamma_tensordot`).

The formula is exact, and the filter computation is scanned parallelly. whoo-hoo...

fritzo:
@fehiepsi can we chat offline about this some time? I'd like us to have a heavy-tailed alternative to `GaussianHMM`. As I understand, one approach is to use `StudentT` filters, with dof either local or global. Do you know which this PR implements?

Another approach might be to try to do filtering with the [Stable distribution](https://en.wikipedia.org/wiki/Stable_distribution), a heavy-tailed generalization of the `Normal` distribution. I do not know the state of the art of Stable filtering, and I do not know whether it generalizes to multiple dimensions.

fehiepsi:
@fritzo Sure, please ping me anytime tomorrow! This PR implements the global ddof version (model (2) of [this paper](https://users.aalto.fi/~ssarkka/pub/SPL2019.pdf)) which assumes the joined distribution of (all_latents, all_observations) follows MVT distribution. StudentT version of GaussianHMM would be straightforward to implement from this PR. For the local ddof version, the filter would be non-exact and we need some sort of approximation (e.g. moment matching) for tensordot op. So it is less interesting than the global version.

Re stable distribution: Me neither. :)

pyro/ops/studentt.py
```diff
@@ -0,0 +1,387 @@
...
+        return GaussianGamma(*args)
+
+    def event_pad(self, left=0, right=0):
+        """
+        Pad along event dimension.
+        """
+        lr = (left, right)
+        info_vec = pad(self.info_vec, lr)
+        precision = pad(self.precision, lr + lr)
+        # no change for alpha, beta because we are working with reparameterized version
```
fritzo: Would it be more precise to say

> no change for log_normalizer, alpha, and beta because they are shared across dimensions

fritzo:
Do I understand correctly that this can be viewed as a scale-mixture of multivariate `Gaussian`s, and that there exists in theory a more general `InverseWishartGaussian` that extends `GammaGaussian` to be a "covariance-scaled" mixture? I'd like to better understand the family of distributions in containing `Gaussian` and `GammaGaussian`.

karalets:
Hey Fritz, maybe you are also interested in this paper here on generalized
wishart processes:

https://arxiv.org/abs/1101.0240

fehiepsi:
They are different afaik. WishartGaussian can be seen as covariance-scaled
of diagonal gaussians with constant diag. But I’ll read more to see if
there is a common theory for both.

pyro/ops/studentt.py
fehiepsi: I just add a comment to clarify this point.

fehiepsi:
> Is there any math you'd like me to check in detail?

Thanks! I think that the tests cover enough math for this PR. :)

fritzo:
I am a little confused with the tacit assumptions behind the conversions `mvt_to_gamma_gaussian()` and `matrix_and_mvt_to_gamma_gaussian()`. After this PR merges we will use these to convert `trans_dist` and `obs_dist` in `GammaGaussianHMM` #2213 . In that context, we will have e.g. a batch of elliptical `MultivariateStudentT` distributions for, say, `obs_dist`. Now as I understand there is a difference between (1) a batch of elliptical `MultivariateStudentT` distributions which corresponds to a batch of scale-mixtures of gaussians whose marginalized out scale is a vector of batch size, and (2) a single `batch_size x event_size`-sized elliptical `MutlivariateStudentT` which corresponds to single scale mixture of `MultivariateGaussian`s with a single shared scale.

That is, whereas the a `GaussianHMM` is specified by a pair
```
trans_dist : MultivariateNormal
obs_dist : MultivariateNormal
```
the `GammaGaussian` should be specified by a triple
```
scale_dist : Gamma
trans_dist : MultivariateNormal
obs_dist : MultivariateNormal
```
in which case we will need conversions
```
gamma_and_mvn_to_gamma_gaussian()
gamma_and_matrix_and_mvn_to_gamma_gaussian()
```
It does note make sense to convert multiple `StudentT` distributions to `GammaGaussian` distributions and then fuse them together, since the former assumes the `scale` latent variable is independent across factors, whereas the latter assumes a single global `scale`.

fehiepsi:
You are absolutely right about the purpose of those utilities!!!

fritzo:
@fehiepsi I am happy to merge this now or later, but I think we should change the `*_to_gamma_gaussian()` helpers before merging #2213 .

fritzo:
@fehiepsi I can't review: either github is buggy or you reset --hard dev into this branch.

fehiepsi:
@fritzo Sorry, I rename those files so they are not available in my last commit. :(

I have made a couple of changes based on your suggestions:
+ use the name `alpha`, `beta` as the reparameterized versions of `concentration`, `rate` (previously I used alpha' and beta', which have caused confusions)
+ rename `mvt_to_gamma_gaussian(..., return_conditional=False)` to `gamma_and_mvn_to_gamma_gaussian`
+ rename `matrix_and_mvt_to_gamma_gaussian(..., return_conditional=True)` to `matrix_and_mvn_to_gamma_gaussian`
+ rewrite tests for the above utilities; now the tests are exact (rather than invoking monte carlo for marginalization)
+ rename `ops.studentt` to `ops.gamma_gaussian`. I intended to use GammaGaussian for robust student-t (with approximations) but it will be better to distinguish the dependent-error-terms case and the independent-error-terms case. So I would like to reserve `ops.studentt` for the robust one.

PR #2028, opened 2019-09-05
Working memory tutorial

ae-foster:
This is the first real tutorial for `pyro.contrib.oed`.

In this notebook, we use OED to learn working memory capacity through a sequence of adaptive experiments. The only EIG estimator introduced is `marginal_eig`


martinjankowiak:
this is great! some initial comments/suggestions

- even if you want to save the main discussion of oed to later, i think it'd be good if you have a bit more context/intro at the top (instead of directly jumping into memory capacity)
- please explicate 'number of rounds'
- note in a comment that you initialize the guide at the prior
- 'SVI' returns the negative ELBO not the ELBO
- after your first use of SVI inference note that the uncertainty has decreased?
- the first time you use "KL" in the text use the full form (kullback....)
- "following formula for EIG" => "following variational estimator for the EIG"
- explain some of the args like `num_samples`
- "This framework applies equally well to more sophisticated models and experiments." =>
(something like) "Note that while this may seem like an obvious design choice, this framework applies equally well to more sophisticated models and experiments where optimal designs are much more difficult to reason about."
- explain in words what 'synthetic_person' is doing
- typo 'paricipant'
- 'Now let's visualize the evolution of the posterior' + 'over theta'

martinjankowiak:
awesome, this is looking ever more polished! here are a few more suggestions:

- 'we can obtain data' => 'we can try to obtain data'
- 'length of a sequence that we ask a participant to remember' => + 'sequence of digits'
- 'where the participants' => 'where the participant' (no s)
- make the xlabel in the sigmoid plot 'length of sequence l'

once we've stabilized the code snippets it will also make sense to add a few more comments about tensor shapes and some of the other technical details.

@fritzo @eb8680 this intro tutorial is now in pretty good shape. could you please take a look and see if there are places for improvement, in particular with respect to good software practices? these could either be changes in the notebook itself or in `contrib.oed`. some things to focus on include: i) the way `make_model` is being used; and ii) the way `ExitStack()` and `iter_plates_to_shape` are being used.

fritzo:
I would recommend replacing `iter_plates_to_shape` with inlined code, since it's just as many lines and since `iter_plates_to_shape` is an undocumented internal helper function. Here's a try:
```diff
  with ExitStack() as stack:
-     for plate in iter_plates_to_shape(l.shape[:-1]):
-         stack.enter_context(plate)
+     for dim in range(-l.dim(), -1):
+         stack.enter_context(pyro.plate(f"plate_{dim}", l.size(dim), dim=dim))
```

ae-foster:
@martinjankowiak @fritzo that latest commit should address your comments. Can't figure out what is up with the CI though

eb8680:
> could you please take a look and see if there are places for improvement, in particular with respect to good software practices?

Looks mostly good to me, `contrib.glmm` is where the really gross code lives.  My only comments are 
1. you might want a sentence or two explaining in words how the guide used by `marginal_eig` is different from the usual guides used in variational inference
2. the use of `marginal_eig` to pretrain without using the return value and to compute the final EIG is somewhat confusing, we might want to either explain what `marginal_eig` does under the hood more explicitly or refactor `marginal_eig` into a class with separate methods for pretraining and estimation, or at least give it a single method with a name that indicates what it's doing (`run` or `train`, say).

ae-foster:
Thanks @eb8680 luckily the new brmp will replace `contrib.glmm` and we will no longer have to look at that.

With regards the separate pretrain/estimate calls to `marginal_eig` I agree this is unsightly. What I could do is use a learning rate scheduler in place of the optimiser with a fixed learning rate. This should roughly accomplish what I want in just one call. (The two calls are basically just a way to very naively adapt the learning rate)

tutorial/source/working_memory.ipynb
```diff
@@ -127,16 +127,21 @@
...
+    "from pyro.contrib.util import rexpand\n",
     "\n",
     "def multiround_model(l):\n",
-    "    # The shape of l is (batch dims..., number of rounds)\n",
+    "    # The shape of l is (*batch dims, number of rounds)\n",
+    "    batch_dims = l.shape[:-1]\n",
     "    with ExitStack() as stack:\n",
-    "        for plate in iter_plates_to_shape(l.shape[:-1]):\n",
-    "            stack.enter_context(plate)\n",
+    "        for i, dim in enumerate(batch_dims[::-1]):\n",
```
fritzo: nit: could you call this variable `size` rather than `dim`? (`pyro.plate` has a different arg named `dim` and it is confusing to pass `dim` to the `size` arg)

tutorial/source/working_memory.ipynb
```diff
@@ -440,13 +446,18 @@
    "source": [
     "def make_model(mean, sd):\n",
     "    def model(l):\n",
-    "        # The shape of l is (batch dims..., number of rounds)\n",
+    "        # The shape of l is (*batch dims, number of rounds)\n",
+    "        batch_dims = l.shape[:-1]\n",
     "        with ExitStack() as stack:\n",
-    "            for plate in iter_plates_to_shape(l.shape[:-1]):\n",
-    "                stack.enter_context(plate)\n",
+    "            for i, dim in enumerate(batch_dims[::-1]):\n",
```
fritzo: ditto: `dim` -> `size`

eb8680:
> What I could do is use a learning rate scheduler in place of the optimiser with a fixed learning rate.

That seems fine, I would also explain that `marginal_eig` is side-effectful in that it's updating the guide parameters.

tutorial/source/working_memory.ipynb
ae-foster: sure thing

martinjankowiak:
is there no way to move all the `ExitStack` business into eig code?

ae-foster:
I thought about it and here is the issue. The difference between batch design dims and "event" design dims is model specific. E.g. typically a regression model has two event dimensions (n, p) whereas this example has one (l) and others have zero. We would have to replace the exit stack code with some ugly decorator or something, which is less explicit than what we have now.

In short

    batch_dims = design.shape[:-k]

how do we know `k` from eig code?

martinjankowiak:
pass as an argument to eig code?

Am Di., 10. Sept. 2019 um 17:44 Uhr schrieb ae-foster <
notifications@github.com>:

ae-foster:
Yeah you could, or you could `model = add_exit_stack(model, n_design_event_dims)`. The downside is that it makes it less explicit what's going on.

fritzo:
The `ExitStack` doesn't bother me fwiw

martinjankowiak:
well the trouble is that it's difficult for a beginning user to reason
about. "do i need this?" "where does it go?" "do i also need it when i want
predictions from my model?" etc.

Am Di., 10. Sept. 2019 um 18:11 Uhr schrieb Fritz Obermeyer <
notifications@github.com>:

ae-foster:
I agree it could be confusing. What solution would you prefer? We could have

    marginal_eig(model, design, ..., design_event_dims=2)

or 

    model = parallelize_by_input(model, design_event_dims=2)
    marginal_eig as usual

fritzo:
What do you think of this syntax:
```py
sizes = (size1, size2, ..., sizeN)
with pyro.plate_stack("prefix", sizes, rightmost_dim=-2):
    ...
```
This is easy to implement:
```py
# in pyro/primitives.py perhaps?
@contextmanager
def plate_stack(prefix, sizes, rightmost_dim=-1):
    assert rightmost_dim < 0
    with ExitStack() as stack:
        for i, size in enumerate(reversed(sizes)):
            plate = pyro.plate(f"{prefix}_{i}", size, dim=rightmost_dim - i)
            stack.enter_context(plate)
        yield
```

pyro/primitives.py
```diff
@@ -243,6 +243,18 @@ def __init__(self, *args, **kwargs):
         super(irange, self).__init__(*args, **kwargs)
 
 
+@contextmanager
+def plate_stack(prefix, sizes, rightmost_dim=None):
```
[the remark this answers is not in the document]
fritzo: Please use the logic I pasted above. Dim should be negative. For reference:
```
@contextmanager
def plate_stack(prefix, sizes, rightmost_dim=-1):
    """
    Create a contiguous stack of :class:`plate` s with dimensions::

        rightmost_dim - len(sizes), ..., rightmost_dim

    :param str prefix: Name prefix for plates.
    :param iterable sizes: An iterable of plate sizes.
    :param int rightmost_dim: The rightmost dim, counting from the right.
    """
    assert rightmost_dim < 0
    with ExitStack() as stack:
        for i, size in enumerate(reversed(sizes)):
            plate = pyro.plate(f"{prefix}_{i}", size, dim=rightmost_dim - i)
            stack.enter_context(plate)
        yield
```
> tests?

You could add a couple tests to tests/infer/test_valid_models.py, say with sizes 1,2,3.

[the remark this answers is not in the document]
ae-foster: Sure thing- I will update the `plate_stack` and add a test

martinjankowiak:
@ae-foster is this ready for another review?

ae-foster:
Yes please!

PR #2030, opened 2019-09-07
Update cloudfront link for new S3 bucket

neerajprad:
Our S3 bucket was deleted accidentally by EngSec. I have tried to recreate another bucket with copies of the datasets that I found on my local system / online. As an example / tutorial writer, if you did any post-processing on these datasets, that might be missing. Hopefully our test suite should catch that. If not, let me know.


neerajprad:
This should be good to merge. Other PRs will continue to fail until this is merged.

Issue #2031, opened 2019-09-07
torch_patch should use functools.wraps 


SsnL:
`torch_patch` is a bad bad hack. It may be necessary. But at least `wraps` should be used so that the docs and things are preserved.

fritzo:
@SsnL thanks for the suggestion. I see `torch_patch` as mechanism to hide latency / match impedance between two teams in separate orgs whose release processes are decoupled. I often need a patch today and can't wait for a PyTorch bugfix to land and release :smile:

SsnL:
@fritzo That makes sense. Thanks for the quick reply. If possible, may I suggest instead of changing the attributes of other libraries, writing and calling new custom wrappers within pyro? The current method has the problem of having severe side effect at importing and hurting people who use source build pytorch and a release version pyro.

fritzo:
@SsnL We can make some effort to use custom wrappers, however we continue to need to patch PyTorch functions that are used internally within PyTorch, e.g. `Transform.__getstate__()`.

Historically `torch_patch.py` has contained lots of patches for bugs that would have been caught if PyTorch devs had stricter testing practices; these patches are generally short-lived, for the duration of one PyTorch release. If there is anything you can do on the FB side to encourage stricter testing practices, we'd sure appreciate it on the Pyro side :smile:. E.g. https://github.com/pytorch/pytorch/issues/11578 would help.

SsnL:
Ah I see. That is a good point. :) I am not at FB anymore but I'll bring this up next time I see the FB guys.

fritzo:
This is now in the dev branch and will be available in the next release.

PR #2032, opened 2019-09-07
Use functools.update_wrapper in torch_patch

fritzo:
Addresses #2031


fritzo:
@SsnL I'll look into using Pyro custom wrappers as well as this PR.

fritzo:
Note test failures are due to Uber EngSec #2030

PR #2033, opened 2019-09-08
Fix inconsistency with tqdm requirements

svenrdz:
Requirements within [setup.py](https://github.com/pyro-ppl/pyro/blob/dev/setup.py#L88) and [docs/requirements.txt](https://github.com/pyro-ppl/pyro/blob/dev/docs/requirements.txt#L8) do not currently match for tqdm version.


neerajprad:
@svenrdz - do you mind signing the CLA? I'm unable to merge without that.

svenrdz:
> @svenrdz - do you mind signing the CLA? I'm unable to merge without that.

I signed it right after submitting the PR.

fritzo:
@svenrdz My guess is that the email on your git commit message is not registered to your Github profile. If that's the case, I believe you can ammend that commit with one of your github registered emails then force push to the branch. I sometimes have to  do this on a new machine when I forget to configure the default email for git commits.

(thanks for taking the time to deal with legal stuff, that's what allows us to be open source :smile:)

svenrdz:
I believe it's now good on the CLA side of things, thanks for the help!

PR #2036, opened 2019-09-13
Add ops module to pyro.generic interface

fritzo:
This aims to make Pyro model code more generic.

Funsor already has a `funsor.ops` module that mimics torch but dispatches to torch/funsor/numpy depending on argument type. I've set the NumPyro backend ops to point to `numpy` in this PR, but @neerajprad may want to switch to a more-torch-compatible `numpyro.ops` library in a follow-up PR.

I have not replaced `torch.randn(...)` and `torch.tensor(...)` in this PR; I think we can do this in follow-up PRs once numpyro and funsor add wrappers. After that.

## Tested
- added a couple unit tests
- updated examples/minipyro.py
- locally ran `examples/minipyro.py --backend=funsor`

### DIFF
--- a/examples/minipyro.py
+++ b/examples/minipyro.py
@@ -11,7 +11,7 @@ import torch
 # We use the pyro.generic interface to support dynamic choice of backend.
 from pyro.generic import pyro_backend
 from pyro.generic import distributions as dist
-from pyro.generic import infer, optim, pyro
+from pyro.generic import infer, optim, pyro, ops
 
 
 def main(args):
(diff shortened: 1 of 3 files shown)


neerajprad:
> Funsor already has a funsor.ops module that mimics torch but dispatches to torch/funsor/numpy depending on argument type.

Looks good! Based on https://github.com/pytorch/pytorch/issues/22402, do you think it makes more sense to have ops following the numpy API or the torch API? I am hoping that at one point we can remove ops once PyTorch implements the `__array_ufunc__` and `__array_function__` interface, so it will be nice to have our `ops` also support the same API.

fritzo:
> numpy API or the torch API?

I don't know? Maybe the union of the two so users can use whichever they like? NumPy is more stable, but PyTorch has some ops that aren't available in NumPy, e.g. `torch.sigmoid`.

PR #2039, opened 2019-09-14
Use torch.jit.script in DiscreteHMM

fritzo:
This converts `_sequential_logmatmulexp()` to a torch script, enabling dynamic size computations even when jitted. Thus this allows the jitted code to be run on time series of different length. Note that `_sequential_gaussian_tensordot()` is much more difficult to convert to a script.

This also removes the deprecated `optimize` setting from examples/hmm.py.

## Performance

`examples/hmm.py -m 7 --jit` (best of three runs)

Before this PR: 4.7 sec

After this: PR 3.7 sec

## Tested
- exercised by existing tests


fritzo:
bump

PR #2041, opened 2019-09-17
Election tutorial

ae-foster:
Here is the second tutorial for pyro.contrib.oed.

In this tutorial, we introduce the posterior estimator and show how it could be used to design a polling strategy that can lead to accurate predictions of the outcome of a US presidential election.


neerajprad:
> @neerajprad should we move this data files (*.pickle) somewhere else?

Even though these are small binaries, we should upload it to aws like other datasets and read it from there, rather than committing it in this PR. 

> @neerajprad what's the optimal way to link to the other OED tutorial?

For the time being, how about just directing to the github [link](https://github.com/pyro-ppl/pyro/blob/dev/tutorial/source/working_memory.ipynb)? Once this is published, we can change those links to direct to the website.

jpchen:
>  what's the optimal way to link to the other OED tutorial?

look at the links in the other intro tutorials. iirc a normal md link should work `[first tutorial](first_tutorial.ipynb)` and then nbconvert will automatically create the correct html extensions.  alternatively, you could hardcode the url `pyro.ai/examples/tutorial_name.html`

martinjankowiak:
@ae-foster when you get to this let's also fix the formatting errors in http://pyro.ai/examples/working_memory.html

ae-foster:
I've gone through and applied @martinjankowiak 's suggestions.

> why eig=False and use of ape?
Since `w` is defined only implicitly by the model we need a large number of samples to compute the prior entropy, better to estimate it once to avoid extra noise comparing designs. I can add a sentence to this effect, though it may confuse some people.

> i'm confused by what's going on in the second-to-last section
I rewrote this section- have a look and see if it is better

> Even though these are small binaries, we should upload it to aws like other datasets and read it from there, rather than committing it in this PR.
I take it you guys have to do that as you have the necessary AWS keys

> a normal md link should work
I'm sticking with this then

> let's also fix the formatting errors in http://pyro.ai/examples/working_memory.html
The only one I could see was a malformed numbered list near the beginning- I tried to fix this by throwing in some extra new lines. Hard to tell how it'll render though. Were there other formatting errors we need to pick up?

martinjankowiak:
@neerajprad can you please help get the pickle files in this PR onto aws?

martinjankowiak:
@ae-foster looks good.
- are you able to `make docs` locally? i think there's still some weird formatting
- i still find the second-to-last section somewhat confusing. what is the message you're trying to get across?
- re: "I can add a sentence to this effect, though it may confuse some people." i think you should explain this a bit in the text (although i guess you should keep the explanation short)

ae-foster:
Just pushed some more updates.

> are you able to make docs locally? i think there's still some weird formatting
I went through the existing tutorial online and picked up a few other changes which are fixed now

> i still find the second-to-last section somewhat confusing
I had another go at rewriting and simplifying this section. The general point is: having done the experimental design, let's simulate actually running the experiment and conducting the data analysis: have a look at the posterior compared to the prior. Nothing really intellectual here, just showing the complete experiment loop. If you think it's still confusing it's no problem to drop this section altogether.

> i think you should explain this a bit in the text
I've added explanation as comments near where I put `eig=False` so it's clear what I'm referring to.

Overall, I think the tutorial is looking close to ready now. Did you get the files up on AWS?

martinjankowiak:
looking great. a few last nits:
- the 'previous tutorial' link and the 'working memory tutorial' link should point to http://pyro.ai/examples/working_memory.html
- can you call the intermediates in `compute_dem_probability` something other than `y`
- typos: 'If we have the option of first design the polling strategy using our prior'

i'll let you know once we have the AWS pickles sorted

neerajprad:
@ae-foster, @martinjankowiak - The dataset files should be in https://d2hg8soec8ck9v.cloudfront.net/datasets/us_elections/ directory on aws.

e.g. https://d2hg8soec8ck9v.cloudfront.net/datasets/us_elections/us_presidential_election_data_test.pickle

ae-foster:
Thanks- I've sorted those comments from @martinjankowiak and I have removed the data from the repo, instead we now have

    !curl -sO "https://d2hg8soec8ck9v.cloudfront.net/datasets/us_elections/electoral_college_votes.pickle"
    etc.

which should download the data to the right directory at the start

ae-foster:
Now using `urlopen` :+1:

ae-foster:
It's inherently stochastic because it depends on the simulated outcome of the experiment

PR #2043, opened 2019-09-23
Implement KL divergence for delta and independent distributions

fritzo:
This implements `kl_divergence()` for some new pairs, useful in `TraceMeanField_ELBO`:
- (Delta, *)
- (Independent, Independent) in a more general way
- (Independent(Delta), MultivariateNormal)
- (Independent(Normal), MultivariateNormal)

I've also moved our custom kl registrations into a new file. Some of these can eventually be moved to PyTorch.

## Future work

I'd next like to implement `kl_divergence(Independent(Normal(...), 2), GaussianHMM(...))`. I believe this requires computing the diagonal of the posterior precision matrix, i.e. marginalizing out all hidden variables. In theory this can be done cheaply via parallel-scan, but would require both a forward and backward pass.

@fehiepsi any ideas how to simplify the `GaussianHMM` version?

## Tested
- added unit tests

### DIFF
--- a/pyro/distributions/__init__.py
+++ b/pyro/distributions/__init__.py
@@ -27,6 +27,8 @@ from pyro.distributions.zero_inflated_poisson import ZeroInflatedPoisson
 from pyro.distributions.lkj import CorrLCholeskyTransform, LKJCorrCholesky, corr_cholesky_constraint
 from pyro.distributions.transforms import *  # noqa F403
 
+import pyro.distributions.kl  # noqa F403 isort:skip
+
 __all__ = [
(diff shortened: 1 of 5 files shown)


fritzo:
@eb8680 and I are implementing a funsor version and I'll cc you on that PR.

fehiepsi:
> computing the diagonal of the posterior precision matrix, i.e. marginalizing out all hidden variables. In theory this can be done cheaply via parallel-scan

@fritzo I am still unable to figure out how to compute this term cheaply. I'll wait for your PR to see how it works. The solution would be neat! :)

PR #2044, opened 2019-09-25
Avoid in-place ops in a few places

fritzo:
This avoids in-place ops in a few places, e.g. replacing `.masked_fill_()` with `torch.where()`.

The goal is to reduce `torch.jit.trace` compilation time in a future version of the PyTorch jit, where purely functional blocks of code can be compiled more quickly by neglecting aliasing information.

### DIFF
--- a/examples/hmm.py
+++ b/examples/hmm.py
@@ -555,7 +555,7 @@ def main(args):
     sequences = sequences[..., present_notes]
 
     if args.truncate:
-        lengths.clamp_(max=args.truncate)
+        lengths = lengths.clamp(max=args.truncate)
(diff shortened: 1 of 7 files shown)


neerajprad:
> PyTorch jit, where purely functional blocks of code can be compiled more quickly by neglecting aliasing information.

Is this under consideration by the PyTorch team?

fritzo:
Yes, during our hacking session, @suo suggested that alias analysis could be avoided for purely functional blocks of code. I believe that after this PR examples/hmm.py will be purely functional, so @suo and jit team can use `examples/hmm.py --jit` to test speed of `torch.jit.trace` for large compute graphs.

PR #2045, opened 2019-09-26
Implement CRPS for empirical distributions

fritzo:
This implements an `n log(n)`-time algorithm for the (negative) [Continuous Ranked Probability Score](https://www.stat.washington.edu/raftery/Research/PDF/Gneiting2007jasa.pdf) between an unweighted set of samples and truth. This is a nice "strictly proper" evaluation metric that favors uncertainty but reduces to absolute error for degenerate distributions. This is worth adding to `pyro.ops.stats` because the naive version is quadratic-time and it's nice to have the slightly more subtle `n log(n)` version implemented with unit tests.

## Tested
- added unit tests
- verified fast running time on a largish BART forecasting dataset ([code](https://github.com/pyro-ppl/sandbox/blob/df7813d/2019-08-time-series/bart/evaluate.py#L79))

### DIFF
--- a/pyro/ops/stats.py
+++ b/pyro/ops/stats.py
@@ -397,3 +397,42 @@ def fit_generalized_pareto(X):
     k = k * N / (N + 10.0) + 5.0 / (N + 10.0)
 
     return k, sigma
+
+
+def crps_empirical(pred, truth):
+    """
+    Computes negative Continuous Ranked Probability Score CRPS* [1] between a
+    set of samples ``pred`` and true data ``truth``. This uses an ``n log(n)``
+    time algorithm to compute a quantity equal that would naively have
+    complexity quadratic in the number of samples ``n``::
+
+        CRPS* = E|pred - truth| - 1/2 E|pred - pred'|
+              = (pred - truth).abs().mean(0)
+              - (pred - pred.unsqueeze(1)).abs().mean([0, 1]) / 2
+
+    Note that for a single sample this reduces to absolute error.
+
+    References
+    [1] `Strictly Proper Scoring Rules, Prediction, and Estimation`
+    Tilmann Gneiting, Adrian E. Raftery (2007)
+    https://www.stat.washington.edu/raftery/Research/PDF/Gneiting2007jasa.pdf
+
+    :param torch.Tensor pred: A set of sample predictions batched on rightmost dim.
+        This should have shape ``(num_samples,) + truth.shape``.
+    :param torch.Tensor truth: A true tensor.
+    """
+    if pred.dim() != 1 + truth.dim() or pred.shape[1:] != truth.shape:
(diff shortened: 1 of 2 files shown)


pyro/ops/stats.py
```diff
@@ -397,3 +397,44 @@ def fit_generalized_pareto(X):
...
+    Note that for a single sample this reduces to absolute error.
+
+    References
+    [1] `Strictly Proper Scoring Rules, Prediction, and Estimation`
+    Tilmann Gneiting, Adrian E. Raftery (2007)
+    https://www.stat.washington.edu/raftery/Research/PDF/Gneiting2007jasa.pdf
+
+    :param torch.Tensor pred: A set of sample predictions batched on rightmost dim.
+        This should have shape ``(num_samples,) + truth.shape``.
+    :param torch.Tensor truth: A true tensor.
```
martinjankowiak: is there a better description than 'true tensor'?  tensor of observations or the like?

pyro/ops/stats.py
```diff
@@ -397,3 +397,44 @@ def fit_generalized_pareto(X):
...
+    """
+    if pred.dim() != 1 + truth.dim() or pred.shape[1:] != truth.shape:
+        raise ValueError(f"Expected pred to have one extra sample dim on left. "
+                         f"Actual shapes: {pred.shape} versus {truth.shape}")
+    opts = dict(device=pred.device, dtype=pred.dtype)
+    num_samples = pred.size(0)
+    if num_samples == 1:
+        return (pred[0] - truth).abs()
+
+    pred = pred.sort(dim=0).values
```
martinjankowiak: is the n log n algorithm from [hersbach](https://journals.ametsoc.org/doi/full/10.1175/1520-0434%282000%29015%3C0559%3ADOTCRP%3E2.0.CO%3B2)?

fritzo: Yeah, it looks like eqn. 20 in that reference contains the same trick. They sure spend lots of words to describe 10 lines of code 😓

pyro/ops/stats.py
fritzo: done

YRChen1998:
The naive version "(pred - truth).abs().mean(0) - (pred - pred.unsqueeze(1)).abs().mean([0, 1]) / 2" doesn't need to sort, why is it slower than this version "(pred - truth).abs().mean(0) - (diff * weight).sum(0) / num_samples**2" ?

fritzo:
@YRChen1998 (this is far from fresh in my mind but...) `(pred - pred.unsqueeze(1)).abs().mean([0, 1])` is an all-pairs comparison and has complexity `O(n^2)`, whereas computing `weight` has complexity only `O(n log n)`.

YRChen1998:
Great idea! Your version also has O(n) space complexity, thanks！

PR #2049, opened 2019-10-01
Add some models for generic testing of MCMC

neerajprad:
Some of the changes needed to enable this in NumPyro are not merged - https://github.com/pyro-ppl/numpyro/pull/374. 

Simple models can use generic dispatch to run MCMC. This has the following changes:

 - Moves `generic.py` into a separate `generic` module.
 - Creates `generic.distributions` module which basically imports distributions from the Pyro namespace along with torch constraints and transforms. We can also create separate dispatch modules for either of these, but it seemed a little excess.
 - Exposes `handlers` in generic. Creates a `seed` contextmanager in Pyro which simply calls `pyro.set_rng_seed` in Pyro but calls the `seed` handler in NumPyro.
 - A simple testing module is added, but it can be modified to add separate model, guide pairs to test SVI. This is important for funsor, and might at one point be useful to have for NumPyro.

I have not formalized the generic API. Doing so will simply involve creating abstract base classes in generic, and ensuring that the dispatched classes inherit from this generic API. This doesn't preclude such a formalization in the future.


pyro/generic/distributions.py
```diff
@@ -0,0 +1,5 @@
+from pyro.distributions import *  # noqa: F401, F403
+
+# Expose constraints and transforms
+from torch.distributions.constraints import *  # noqa: F401, F403
```
fritzo: I think it's a bit safer to access constraints and transforms in a partially qualified way, e.g. `transforms.Planar` and `constraints.positive`. Could we allow this somehow, say either `import torch.distributions.constraints as constraints` or treat these as a top-level `pyro_generic`-managed thing?

Note this is becoming more important as #1932 is dropping `Transform` suffix on transforms and assumes they will be accessed partially qualified.

pyro/generic/testing.py
```diff
@@ -0,0 +1,106 @@
...
+    logits = ops.sum(true_coefs * data, axis=-1)
+    labels = pyro.sample('labels', dist.Bernoulli(logits=logits))
+
+    def model(x, y):
+        coefs = pyro.sample('coefs', dist.Normal(ops.zeros(dim), ops.ones(dim)))
+        intercept = pyro.sample('intercept', dist.Normal(0., 1.))
+        logits = ops.sum(coefs * x, axis=-1) + intercept
+        return pyro.sample('obs', dist.Bernoulli(logits=logits), obs=y)
+
+    return model, (data, labels), {}
```
fritzo: Could we make these dicts so as to be forwards compatible with SVI and special forms of testing? E.g.
```py
return {'model': model, 'model_args': (data, labels)}
```

fritzo: That way we can add things like `'guide': guide` in the future.

pyro/generic/testing.py
```diff
@@ -0,0 +1,106 @@
...
+            with pyro.plate("plate_1", data.shape[-2]):
+                probs = pyro.sample("probs", dist.Beta(alpha, beta))
+                with pyro.plate("data", data.shape[0]):
+                    pyro.sample("binomial", dist.Binomial(probs=probs, total_count=total_count), obs=data)
+
+    return model, (data,), {}
+
+
+def main(args):
+    with pyro_backend(args.backend):
```
fritzo: It would be nice to factor this for easy use via pytest. Maybe something like
```py
# in pyro.generic.testing
def check_model(backend, name):
    get_model = MODELS[name]
    with pyro_backend(backend):
        ...

def main(args):
    for name in sorted(MODEL):
        check_model(args.backend, name)
```
and then
```py
# in tests/pyro/generic/test_testing.py
from pyro.generic.testing import MODELS, check_model
import pytest

@pytest.mark.parametrize('backend', ['pyro', 'minipyro'])
@pytest.mark.parametrize('name', sorted(MODELS))
def test_model(backend, name):
    check_model(backend, name)
```
and later elsewhere
```py
# in funsor/test/test_minipyro.py
from pyro.generic.testing import MODELS, check_model
import pytest

@pytest.mark.parametrize('name', sorted(MODELS))
def test_model(name):
    check_model('funsor', name)
```

pyro/generic/generic.py
```diff
@@ -27,6 +27,12 @@ def __getattribute__(self, name):
         return getattr(module, name)
 
 
+@contextmanager
+def model(rng):
```
fritzo: 1. nit: the arg is a seed, not an rng, right? rng usually means an object with full rng state.
2. is it possible to define this in a way where a seed need not be specified by default, so as to exactly reproduce current vanilla pyro behavior?

pyro/generic/generic.py
```diff
@@ -55,36 +61,42 @@ def pyro_backend(*aliases, **new_backends):
 _ALIASES = {
     'pyro': {
         'pyro': 'pyro',
-        'distributions': 'pyro.distributions',
+        'distributions': 'pyro.generic.distributions',
```
fritzo: Can you explain the advantage of adding a new `pyro.generic.distributions` module? I don't see why this is necessary. I'd even prefer to add a pyro/distributions/constraints.py and pyro/distributions/transforms.py that merely import torch stuff, rather than create an extra `pyro/generic/distributions`

neerajprad: > Can you explain the advantage of adding a new pyro.generic.distributions module? I don't see why this is necessary. I'd even prefer to add a pyro/distributions/constraints.py and pyro/distributions/transforms.py that merely import torch stuff

There's no advantage at all. I think we can instead create `pyro.distributions.contstraints` and `pyro.distributions.transforms` or even import them as separate modules.

pyro/generic/generic.py
neerajprad: 1. That's bad naming, it should just be named `seed` or `rng_seed`. Will change. 
2. Sure. I think we can even remove this contextmanager now, since we can just use `handlers.seed(...)` instead, which will not have an effect for Pyro.

pyro/generic/testing.py
neerajprad: Sounds good!

pyro/generic/distributions.py
neerajprad: Hmm..either option sounds good to me. I'll go with the first one right now, but let me know if you would prefer it to be specified at the generic level.

pyro/generic/testing.py
neerajprad: That's a great suggestion, will refactor.

pyro/generic/distributions.py
neerajprad: The first option unfortunately doesn't work because we have a local `pyro.distributions.transforms` module, so I cannot just import `torch.distributions.transforms` into `pyro.distributions.__init__`. I thought it will be best to go with the second option instead which is creating separate generic modules for constraints and transforms.

neerajprad:
Thanks for reviewing, @fritzo! I will address the remaining comments and update when this PR is ready for a second review.

neerajprad:
I think this is good for a second round of reviewing.

neerajprad:
Some changes based on prior comments:
 - I have exposed `transforms` and `constraints` in the generic module itself that dispatch to torch constraints/transforms. This is because we have a local `pyro.distributions.transforms` module, and given how important these are (for transformed distributions, etc), it makes sense to have them directly in generic.
 - Also added the `generic` module itself. This was needed so that we don't have a mock `handlers.seed` in pyro and can use the `generic.seed` decorator / contextmanager instead. This however will only have any effect when under the `pyro_backend` context manager which requires this module be available for generic dispatching. An alternative would be to build a `handlers.seed` mock handler in other backends.

pyro/infer/__init__.py
```diff
@@ -23,13 +26,16 @@
...
+    "HMC",
     "Importance",
     "IMQSteinKernel",
     "infer_discrete",
     "JitTraceEnum_ELBO",
     "JitTraceGraph_ELBO",
     "JitTraceMeanField_ELBO",
     "JitTrace_ELBO",
+    "MCMC",
+    "NUTS",
```
fehiepsi: probably this is a long-awaited feature. :D I am not a fan of having different inference engines in different namespaces.

pyro/generic/generic.py
```diff
@@ -36,55 +73,83 @@ def pyro_backend(*aliases, **new_backends):
     or by providing a dict mapping module name to backend module name.
     Standard backends include: pyro, minipyro, funsor, and numpy.
     """
+    global _BACKEND
+    old_backend_name = _BACKEND
     if aliases:
```
fehiepsi: nit: not related to this PR, but I think we can simplify the logic by using the signature: `def pyro_backend(alias=None, **new_backends):`, then `if alias is not None` here.

tests/test_generic.py
```diff
@@ -0,0 +1,19 @@
...
+
+from pyro.generic import infer, pyro_backend, seed
+from pyro.generic.testing import MODELS
+
+
+pytestmark = pytest.mark.stage('unit')
+
+
+@pytest.mark.parametrize('model', MODELS)
+@pytest.mark.parametrize('backend', ['pyro'])
```
fehiepsi: I think we can add a check if `numpyro` is available here. It is helpful to test locally.

neerajprad: The idea is that each backend should be able to import these set of models and run tests as they see fit, so the tests for the numpy backend can be in numpyro. Of course, right now the issue is that we will need to install pyro first to be able to use this module - I have a suggestion in the design doc for this.

pyro/generic/distributions.py
fritzo: I think it might make most sense to import torch transforms into pyro transforms, so that just like pyro.distributions is a strict extension of torch.distributions, pyro.distributions.transforms would be a strict extension of torch.distributions.transforms:
```py
# in pyro/distributions/transforms.py
from torch.distributions.transforms import *
```
@stefanwebb WDYT?

docs/source/pyro.generic.rst
```diff
@@ -4,7 +4,7 @@ Generic Interface
 The ``pyro.generic`` module provides an interface to dynamically dispatch Pyro code
 to custom backends.
 
-.. automodule:: pyro.generic
+.. automodule:: pyro.generic.generic
```
fritzo: nit: it would be a little more accurate to name this file `pyro.generic.dispatch`

pyro/generic/generic.py
```diff
@@ -27,6 +31,39 @@ def __getattribute__(self, name):
...
+    :param fn: Python callable. Not required when used as a context manager.
+    :param int rng_key: random number generator seed.
+    """
+    seed_ctx = _SeedCtx(rng_key=rng_key)
+    return seed_ctx if fn is None else seed_ctx(fn)
+
+
+class _SeedCtx(object):
+    def __init__(self, rng_key=None):
+        self.rng_key = rng_key if rng_key is not None else random.randint()
```
fritzo: Could you explain usage to me? Wouldn't this mean that repeated calls to a seeded model would all return the same result? That does not match the current behavior of vanilla Pyro, but maybe I am missing typical usage patterns in NumPyro.

pyro/generic/generic.py
```diff
@@ -27,6 +31,39 @@ def __getattribute__(self, name):
         return getattr(module, name)
 
 
+def seed(fn=None, rng_key=None):
```
fritzo: This should be provided by each backend, otherwise there is no way for backends to override this. I believe it should live either in poutine or primitives (and be a top-level `pyro.seed` decorator).

pyro/generic/generic.py
```diff
@@ -27,6 +31,39 @@ def __getattribute__(self, name):
...
+    return seed_ctx if fn is None else seed_ctx(fn)
+
+
+class _SeedCtx(object):
+    def __init__(self, rng_key=None):
+        self.rng_key = rng_key if rng_key is not None else random.randint()
+        self.seed_ctx = None
+
+    def __enter__(self):
+        if _BACKEND == 'numpy':
```
fritzo: if-statements are not extensible.

pyro/generic/generic.py
```diff
@@ -36,55 +73,83 @@ def pyro_backend(*aliases, **new_backends):
     or by providing a dict mapping module name to backend module name.
     Standard backends include: pyro, minipyro, funsor, and numpy.
     """
+    global _BACKEND
```
fritzo: I'd rather avoid duplicate state if possible. We should not be if-else branch on a name in this file, because that pattern is not extensible.

pyro/generic/generic.py
neerajprad: This currently only affects the NumPyro backend and should be a NoOp for other backends (that's the reason for the if block below), which are free to instead call their own mechanism to set the random seed. This was motivated by your previous suggestion to redesign keeping the flow for Pyro the same (maybe I misunderstood your earlier comment?). I also think it will be better for this to be in `pyro.poutine`. 

Do you have any suggestions as to the behavior of that poutine? Should it just call `pyro.set_rng_seed` in `__init__` and have no other effect? Note that this needs to be a poutine in NumPyro since it has to intercept every `sample` statement. With that design, we can just remove this function and also `generic` from dispatch and let `handlers.seed` determine how each backend likes to set the rng seed.

pyro/generic/generic.py
neerajprad: > Wouldn't this mean that repeated calls to a seeded model would all return the same result?

That's right. To generate multiple samples from the same model, we'll need to do this seeding in the outer scope.

```
with seed(rng_key=1):
  samples = [model() for _ in range(100)]
```

In NumPyro, we typically do this in a vectorized fashion, but note that a single seeded instance of a model is deterministic.

```
samples = vmap(lambda key: seed(model, key)())(arange(100))
```
.

docs/source/pyro.generic.rst
neerajprad: I'm fine with removing this once we finalize the behavior of the `seed` handler in Pyro.

pyro/generic/generic.py
neerajprad: +1, I think this is ugly. We should instead figure out what the behavior of the seed handler be in Pyro and remove this code.

pyro/generic/generic.py
fritzo: I believe you and @eb8680 and I agreed on a new handler that has no effect in Pyro:
```py
#in pyro/poutine/handlers.py
def model(fn=None, **kwargs):
    """
    Decorator to declare a Pyro model. This is optional and has no effect in the
    standard Pyro backend, but this may be required for alternative backends, e.g.
    in the NumPyro backend this may install default handlers.
    """
    if fn is None:
        return model
    return fn
```
I would also be happy for this to live in primitives.py, so we can wrap with `@pyro.model`. @eb8680 do you have an opinion?

pyro/generic/generic.py
fritzo: Thanks for explaining, that makes sense!

pyro/generic/generic.py
neerajprad: Ahh..I see what you meant now, I had this model decorator in generic earlier, but it will be better if this resides in `primitives`. I would prefer if this was simply a `seed` context manager in `pyro.primitives`, so that there is a way to set the rng seed in a backend agnostic way. 

```
with seed(rng_key=1):
  ...
```

For Pyro this will just call `pyro.set_rng_seed(..)`. Otherwise, we will need to separately call `pyro.set_rng_seed` for the Pyro backend and use the pattern below for NumPyro, which requires detecting what backend we are on. From the examples and tests, I think the contextmanager use is more important rather than the decorator pattern, since the latter will give the same samples from the model, but both can be supported if this is a poutine.

fritzo: A seed context manager in primitives sounds great, thanks for explaining.

pyro/generic/generic.py
fritzo: The `*args` pattern avoids any possible masking of `**kwargs`. While it is a little more complex, it safely handles the edge case of a backend providing a library named 'alias'.

pyro/generic/distributions.py
fritzo: I like the idea of adding `constraints`  and `transforms` to `pyro.generic`. I think we'll also need a bit of reorg #2052 so that both pyro and torch transforms are accessible.

neerajprad: Done. I have also imported `torch.constraints` into the namespace which can be accessed as `pyro.distributions.constraints`.

neerajprad: Sorry, I missed your last comment. The latest commit assumes that `transforms` and `constraints` are available from the distributions namespace. I am not sure if that conflicts with any plans from #2052. I am happy to revert that to the previous choice of exposing these directly from `pyro.generic`.

pyro/generic/generic.py
neerajprad: This is added to `handlers` now. Could you take a look and see if that looks okay? One advantage to that is that we can use it both as a context manager and decorator and the effect is the same across all backends.

pyro/generic/distributions.py
neerajprad: Just looking at #2052 - lets get that merged first.

neerajprad: Now that #2052 already exposes `transforms` and `constraints`, do you have any opinion on whether we should also separately expose them in `generic`

```
'transforms': 'pyro.distributions.transforms'
```

 or just assume that they will be available from the `distributions` namespace for generic dispatch?

fritzo: I think it's a little cleaner to expose `constraints` and `transforms` from generic.

pyro/generic/generic.py
```diff
@@ -1,12 +1,15 @@
 import importlib
-
 from contextlib import contextmanager
 
 
+_BACKEND = 'pyro'
```
fritzo: Can we remove this now?

pyro/generic/distributions.py
fritzo: On second thought, I have no opinion. I believe this PR is mostly orthogonal from #2052 : in either one we need to be able to access both torch and pyro transforms in a single place.

pyro/distributions/__init__.py
```diff
@@ -1,3 +1,5 @@
+import torch.distributions.constraints as constraints
```
fritzo: I think this is the only line that conflicts with #2052 . I have added the following to #2052:
```py
from . import constraints, kl, transforms
```

pyro/distributions/__init__.py
```diff
@@ -28,6 +30,7 @@
 from pyro.distributions.transforms import *  # noqa F403
 
 __all__ = [
+    "constraints",
```
fritzo: you can remove this line, it was updated in #2052

pyro/generic/distributions.py
neerajprad: Sounds good. I'll just merge this against master but keep things as they are, for now.

docs/source/pyro.generic.rst
neerajprad: Sorry, I misread your earlier comment. My comment above referred to the presence of `generic` module in `_ALIASES` which has since been removed. I'll address this now - I agree `dispatch` is a more accurate name.

pyro/distributions/__init__.py
neerajprad: done.

neerajprad:
Thanks a lot for a thorough review, @fritzo, @fehiepsi. I think I have addressed all existing comments, so this should be ready for another round of review!

pyro/generic/dispatch.py
```diff
@@ -54,37 +53,43 @@ def pyro_backend(*aliases, **new_backends):
...
         'ops': 'torch',
+        'optim': 'pyro.contrib.minipyro',
+        'pyro': 'pyro.contrib.minipyro',
     },
     'funsor': {
-        'pyro': 'funsor.minipyro',
-        'infer': 'funsor.minipyro',
-        'optim': 'funsor.minipyro',
         'distributions': 'funsor.distributions',
+        'handlers': 'pyro.poutine',
```
fritzo: this should be
```
'handlers': 'funsor.minipyro',
```

tests/test_generic.py
```diff
@@ -0,0 +1,19 @@
...
+
+from pyro.generic import infer, pyro_backend, handlers
+from pyro.generic.testing import MODELS
+
+
+pytestmark = pytest.mark.stage('unit')
+
+
+@pytest.mark.parametrize('model', MODELS)
+@pytest.mark.parametrize('backend', ['pyro'])
```
fritzo: Can we add the `minipyro` backend here, or does that not support MCMC?

neerajprad: `minipyro` doesn't support MCMC. Maybe we can add a vanilla HMC to minipyro that uses the integrator in `pyro.ops`, but I'm not sure if it will be as concise as SVI. It could be a nice little hackathon project.

pyro/generic/dispatch.py
neerajprad: Done. I forgot that the simplified effect handling abstraction in minipyro is one of the reasons why we have it.

tests/test_generic.py
```diff
@@ -0,0 +1,19 @@
...
+def test_mcmc_interface(model, backend):
+    with pyro_backend(backend), handlers.seed(rng_seed=20):
+        f = MODELS[model]()
+        model, args, kwargs = f['model'], f.get('model_args', ()), f.get('model_kwargs', {})
+        nuts_kernel = infer.NUTS(model=model)
```
fritzo: Well we should later factor this out into `check_model()` so as to minimize the effort of adding tests to each backend, but we can do that in a follow-up PR when we decide how to add SVI tests.

neerajprad: > Well we should later factor this out into check_model() so as to minimize the effort of adding tests to each backend

Could you elaborate on that? There is a very basic `testing.check_model` that runs the model forward and traces it. I have tried to follow a convention that args represent model inputs and kwargs are used to represent observed data (like Stan's data block). The idea was that each backend will just import `MODELS` from testing and use it as they see fit. We can continue to enrich the metadata from each model, currently its just the model callable, inputs and observed data (maybe we can add ML estimate for the latent posteriors etc).

fritzo: Ideally I'd like to implement a very narrow testing interface with two methods that do not mention pytest and do not mention pyro.generic. Their use is equivalent to the following:
```py
import pytest
from pyro.generic.testing import get_test_cases, check_test_case
# n.b. do not import any pyro stuff, including pyro.generic

@pytest.mark.parametrize('test_case', get_test_cases())
def test_generic(test_case):
    try:
        check_test_case(test_case)
    except NotImplementedError as e:
        pytest.xfail(f"Not implemented:\n{e}")
```
I don't care what these are named or whether they are lists or functions or whatever, but I do think we separate concerns such that:
1. all pyro.generic interface is hidden inside these test helpers, and
2. all testing infra (pytest) is applied outside these test helpers.

I see these helpers as kind of an autoconf mechanism for pyro implementations: running them automatically shows what features are available in a given backend.

fritzo: I've moved this discussion to a new issue https://github.com/pyro-ppl/pyro/issues/2053

Issue #2026, opened 2019-08-25
[FR] Named Normal Autoguide


patrickeganfoley:
Hello!

I find the [AutoDiagnoalNormal](http://docs.pyro.ai/en/0.3.0-release/contrib.autoguide.html#autodiagonalnormal) guide a bit hard to work with, since it gives you a single vec of mean values, rather than a set of means with names that match your model's parameter names.  The [AutoDelta](http://docs.pyro.ai/en/0.3.0-release/contrib.autoguide.html#pyro.contrib.autoguide.AutoDelta) guide on the other hand is great!  For example, if you have a model with a `mu` and `beta`, AutoDelta gives you `auto_mu_loc` and `auto_beta_loc`, while AutoDiagonalNormal just gives you a length two `auto_loc` param.

Would it be a welcome contribution to submit an AutoDiagonalNormal guide that works more like the AutoDelta?  I have one (copied below) that works for my models, but am not sure if I've missed some important things.  

Thanks!  And thanks for all the work on this, it's really neat to use!

```py
import pyro
import pyro.distributions as dist
import pyro.contrib.autoguide as ag

import torch
from torch.distributions import constraints
#  Note the default prefix is '_auto'

from contextlib import ExitStack


class AutoNormal(ag.AutoGuide):
    """
    This implementation of :class:`AutoGuide` uses Normal(0, 1) distributions
    to construct a guide over the entire latent space. The guide does not
    depend on the model's ``*args, **kwargs``.

    It should be equivalent to pyro.contrib.autoguide.AutoDiagonalNormal, but
    with more convenient names.  In AutoDiagonalNormal, if your model has N
    named parameters with dimensions k_i and sum k_i = D, you get a single
    vector of length D for your mean, and a single vec of length D for sigmas.
    This guide gives you N distinct normals that you can call by name.

    Usage::

        guide = AutoNormal(model)
        svi = SVI(model, guide, ...)
    """
    def __call__(self, *args, **kwargs):
        """
        An automatic guide with the same ``*args, **kwargs`` as the base
        ``model``.

        :return: A dict mapping sample site name to sampled value.
        :rtype: dict
        """
        # if we've never run the model before, do so now so we can inspect the
        # model structure
        if self.prototype_trace is None:
            self._setup_prototype(*args, **kwargs)

        plates = self._create_plates()
        result = {}
        for name, site in self.prototype_trace.iter_stochastic_nodes():
            print(f'param {name} is {site}')
[...]

PR #2050, opened 2019-10-01
WIP -  AutoNormal Guide

patrickeganfoley:
Addresses [this issue](https://github.com/pyro-ppl/pyro/issues/2026).

Tests do not pass yet, I am trying to add it to a few of the tests in tests/infer/test_autoguide.py.

I think I have some mistakes in my logic for [generating the shape](https://github.com/patrickeganfoley/pyro/commit/a283d46b5fe95faf9af3cdf4293effa1ff436f08#diff-90aa7b71b70738437a3db6bc6e6d52d8R334) of new RVs.  I'll keep tinkering and trying to better understand the batch shape / event shape / sample shape distinction until I can get these tests to pass, but if I'm making super obvious mistakes I would appreciate any pointers!

Example test error here:

```
E               ValueError: at site "z2", invalid log_prob shape
E                 Expected [], actual [2]
E                 Try one of the following fixes:
E                 - enclose the batched tensor in a with plate(...): context
E                 - .to_event(...) the distribution being sampled
E                 - .permute() data dimensions
```

## Tested

The following tests pass in tests/infer/test_autoguide.py:

- [ ]  test_scores - there's a name assertion that might need to change
    @fritzo: It's ok to add a guard  to that portion of the test:
    ```py
    assert model_trace.nodes['z']['log_prob_sum'].item() != 0.0
    assert guide_trace.nodes['z']['log_prob_sum'].item() == 0.0
    if issubclass(auto_class, AutoContinuous):
        assert '_auto_latent' not in model_trace.nodes
        assert guide_trace.nodes['_auto_latent']['log_prob_sum'].item() != 0.0
    ```
- [ ] test_factor
- [x] test_shapes - __init__ needs to be able to take init_loc_fn as an arg
- [x] test_iplate_smoke - .. everything fails?
- [ ] test_median - fails (I don't think autonormal implements it)
    @fritzo: This is implemented by `AutoContinuous`, the base class of `AutoNormal`. You can either adapt that code or add an `with xfail_if_not_implemented(): ...` to that test and we can implement in a follow-up PR.
- [ ] test_quantiles - also fails (didn't implement)
    @fritzo: ditto
- [ ] test_discrete_parallel - also fails
- [x] test_guide_list
- [x] test_callable
- [x] test_callable_return_dict


pyro/infer/autoguide/guides.py
```diff
@@ -290,6 +290,65 @@ def median(self, *args, **kwargs):
...
+        for name, site in self.prototype_trace.iter_stochastic_nodes():
+            with ExitStack() as stack:
+                for frame in site["cond_indep_stack"]:
+                    if frame.vectorized:
+                        stack.enter_context(plates[frame.name])
+                loc_name = "{}_{}_{}".format(self.prefix, name, 'loc')
+                scale_name = "{}_{}_{}".format(self.prefix, name, 'scale')
+                loc_value = pyro.param(
+                    loc_name,
+                    lambda: torch.zeros(site["fn"]._batch_shape + site["fn"]._event_shape),
```
fritzo: To ensure proper device placement you can instead
```py
lambda: site["value"].new_zeros(site["fn"].shape())
```

pyro/infer/autoguide/guides.py
```diff
@@ -290,6 +290,65 @@ def median(self, *args, **kwargs):
...
+                    constraint=site["fn"].support
+                )
+                scale_value = pyro.param(
+                    scale_name,
+                    lambda: torch.ones(site["fn"]._batch_shape + site["fn"]._event_shape),
```
fritzo: ditto:
```py
lambda: site["value"].new_ones(site["fn"].shape())
```

pyro/infer/autoguide/guides.py
```diff
@@ -290,6 +290,65 @@ def median(self, *args, **kwargs):
...
+                    constraint=constraints.positive
+                )
+
+                result[name] = pyro.sample(
+                    name,
+                    dist.Normal(
+                        loc_value, scale_value
+                    )
```
fritzo: You'll need to match event dim:
```py
dist.Normal(
    loc_value, scale_value
).to_event(site["fn"].event_dim)
```

pyro/infer/autoguide/guides.py
```diff
@@ -290,6 +290,65 @@ def median(self, *args, **kwargs):
         return self(*args, **kwargs)
 
 
+class AutoNormal(AutoGuide):
+    """
+    This implementation of :class:`AutoGuide` uses Normal(0, 1) distributions
+    to construct a guide over the entire latent space. The guide does not
+    depend on the model's ``*args, **kwargs``.
+
+    It should be equivalent to pyro.contrib.autoguide.AutoDiagonalNormal, but
```
fritzo: I'd advertise the other benefits too :smile:
```
It should be equivalent to :class:`AutoDiagonalNormal` , but
with more convenient site names and with better support for
:class:`~pyro.infer.trace_mean_field_elbo.TraceMeanField_ELBO` . In
:class:`AutoDiagonalNormal` , if your model has ...
```

fritzo:
Hi @patrickeganfoley, do you plan to continue working on this PR? We plan to release Pyro 0.5 in the next day, and 1.0 in the next month, and I'd like to merge this PR since it's a nice idea. Let me know if you need any help. If you don't have time to work on this I could take over the PR and try to get it in to 1.0, but you'll need to sign the CLA. Cheers!

patrickeganfoley:
Hi @fritzo !  Thanks for the heads up!  Let me make another attempt on this tonight - I still have a mistake or two related to shapes.  

Do you want this by end of day?

pyro/infer/autoguide/guides.py
patrickeganfoley: oh thanks so much for this!  Will run the tests w/ these changes!

fritzo:
> Do you want this by end of day?

oh no hurry for today. We're planning a 1.0 release in a couple weeks.

patrickeganfoley:
Just incorporated your changes and am checking out the tests.  



Will try to fix a few of these, but so far, I'm getting
  test_scores - there's a name assertion that might need to change
  test_factor PASSES
  test_shapes - __init__ needs to be able to take init_loc_fn as an arg
  test_iplate_smoke - .. everything fails?
  test_median - fails (I don't think autonormal implements it)
  test_quantiles - also fails (didn't implement)
  test_discrete_parallel - also fails
  test_guide_list - PASSES
  test_callable - PASSES
  test_callable_return_dict - PASSES

fritzo:
@patrickeganfoley great, I've added your test results and some comments to a `Tested` section of the PR description.

patrickeganfoley:
Great!  Should have some time today to get the other tests working 👍

fritzo:
Hi @patrickeganfoley, over the last week we've made major changes to our autoguides, using a new `PyroModule` class (see the changes to `AutoDelta` for details). Given all these changes, I think it's better to hold off on this PR until after the 1.0 release.

patrickeganfoley:
Thanks for the heads up 👍 .

abelstam12:
@fritzo @patrickeganfoley This is an awesome feature since you do not have to worry about the order and position of the parameters that are in the store. In [this](https://forum.pyro.ai/t/order-of-latent-variables-in-autoguide/1451) topic I suggested exposing methods like
`
guide.means(), guide.scales() or guide.variances() or guide.standard_devs()
`
**because I like to get confidence intervals for my parameters**. Since the values in the parameter store are learned in the unconstrained parameter space, the parameters should first be mapped to constrained parameter space in the methods above. If this seems useful it would be really awesome **to have the same methods available in the other autoguides**. What is the reason only .median() and .quantiles() is implemented at the moment? **Maybe there is a good reason for not implementing the proposed methods.**

Lastly, IMO the named parameters and mean/scale methods would be a very **nice feature for the AutoMultivariateNormal and other guides aswell**.

 Although I have just been working with pyro for little over a few months I would love to contribute some of the mentioned features if they are feasible/useful.

abelstam12:
> @fritzo @patrickeganfoley This is an awesome feature since you do not have to worry about the order and position of the parameters that are in the store. In [this](https://forum.pyro.ai/t/order-of-latent-variables-in-autoguide/1451) topic I suggested exposing methods like
> `guide.means(), guide.scales() or guide.variances() or guide.standard_devs()`
> **because I like to get confidence intervals for my parameters**. Since the values in the parameter store are learned in the unconstrained parameter space, the parameters should first be mapped to constrained parameter space in the methods above. If this seems useful it would be really awesome **to have the same methods available in the other autoguides**. What is the reason only .median() and .quantiles() is implemented at the moment? **Maybe there is a good reason for not implementing the proposed methods.**
> 
> Lastly, IMO the named parameters and mean/scale methods would be a very **nice feature for the AutoMultivariateNormal and other guides aswell**.
> 
> Although I have just been working with pyro for little over a few months I would love to contribute some of the mentioned features if they are feasible/useful.

I think the methods I proposed need a specific implementation for different autoguides, because there is no universal relationship between the parameters of the distributions and the variance/mean of a distribution. Maybe this implementation should be left to the end-user to implement. Still, to implement getting the mean of a distribution or the variance, one would need access to the parameter representation in the constrained space. I think it would be nice to at least implement this on the auto guides:

`
>>guide = AutoNormal(model)
>>print(guide.get_constrained_parameters())
{
    "auto_theta_locs":torch.tensor(), # all containing the constrained param representations
    "auto_theta_scales":torch.tensor(),
    "auto_alpha_locs":torch.tensor(),
    "auto_alpha_scales":torch.tensor(),
    ....
}
`
Then one could use these results to define certain confidence intervals for the wanted parameters, plot the corresponding distributions etc.

patrickeganfoley:
BTW I closed this inadvertently while rebasing to get the new-ish guides refactor.  Just started trying to get things working yesterday.  Right now things work for me on my simple example, but fail when trying to sample the Dirichlet in the model defined in `test_autoguides.py::test_shapes`.

patrickeganfoley:
and @abelstam12 generic .mean() methods sound terrific to me!  For myself I'm just trying to get this AutoNormal working.

patrickeganfoley:
BTW I am now using the same shape constraint logic that the AutoContinuous guides use, and am finally getting `test_scores` and `test_shapes` to pass.  In the previous version, they were all failing due to support or shape errors.

It's now failing on median() and quantile(), but passing everything else.  I'm going to try to get median() working and then will reopen the PR.

patrickeganfoley:
Hello!  Just got median() and quantiles() working! 😁 

The following tests now work: 

☑️ test_scores
☑️ test_shapes
☑️ test_factor
☑️ test_callable
☑️ test_callable_return_dict
☑️ test_median
☑️ test_quantiles
☑️ test_discrete_parallel
☑️ test_guide_list
☑️ test_linear_regression_smoke

But these do not:
🔲 test_autoguide_serialization
🔲 test_predictive
🔲 test_init_loc_fn
🔲 test_init_scale
🔲 test_median_module

test_iplate_smoke also does not work - but for me every guide fails with `XFAIL`.  

The predictive and serialization tests both fail with this runtime error:

```
E       RuntimeError: Cannot insert a Tensor that requires grad as a constant. Consider making it a parameter or input, or detaching the gradient
E       Tensor:
E       0
E       [ torch.DoubleTensor{} ]

../.pyenv/versions/3.7.3/envs/pyro37/lib/python3.7/site-packages/torch/distributions/transforms.py:310: RuntimeError
```

Any tips on the runtime error?  And which of the remaining tests should I try to get working v.s. not run for this guide?

fritzo:
@patrickeganfoley can you merge the latest Pyro dev into this branch, then I'll take a look?

patrickeganfoley:
Just merged latest dev 👍

pyro/infer/autoguide/guides.py
```diff
@@ -347,6 +347,145 @@ def median(self, *args, **kwargs):
...
+
+            pyro.param(
+                loc_name,
+                lambda: site["value"].new_zeros(unconstrained_shape_for_params)
+            )
+            pyro.param(
+                scale_name,
+                lambda: site["value"].new_ones(unconstrained_shape_for_params),
+                constraint=constraints.positive
+            )
```
fritzo: The major change in 1.0 was to switch from `pyro.param` to `PyroModule`s to support serialization and serving. I think you can update this block as follows:
```py
self.locs = PyroModule()
self.scales = PyroModule()
    ...
    if hasattr(self.locs, name):
        continue
    init_loc = site["value"].new_zeros(unconstrained_shape_for_params)
    init_scale = site["value"].new_full(unconstrained_shape_for_params, self.init_scale)
    setattr(self.locs, name, nn.Parameter(init_loc))
    setattr(self.scales, name, PyroParam(init_scale, constraints.positive))
```

pyro/infer/autoguide/guides.py
```diff
@@ -347,6 +347,145 @@ def median(self, *args, **kwargs):
...
+
+    def _get_loc_and_scale(self, name):
+        loc_name = "{}_{}_{}".format(self.prefix, name, 'loc')
+        scale_name = "{}_{}_{}".format(self.prefix, name, 'scale')
+        site_loc = pyro.get_param_store().get_param(loc_name)
+        site_scale = pyro.get_param_store().get_param(scale_name)
```
fritzo: ...then you can update this block to
```py
site_loc = getattr(self.locs, name)
site_scale = getattr(self.scales, name)
```

pyro/infer/autoguide/guides.py
```diff
@@ -347,6 +347,145 @@ def median(self, *args, **kwargs):
         return self(*args, **kwargs)
 
 
+class AutoNormal(AutoGuide):
+    """This is AutoNormal, but it uses the same shape constraint logic found
+    in AutoMultivariateNormal and AutoDiagonalNormal.
+    """
+    def __init__(self, model, init_loc_fn=init_to_median, prefix='auto'):
+        #  Does this accept a 'prefix' arg?  It doesn't look like it.
+        self.init_loc_fn = init_loc_fn
```
fritzo: It looks like you're not yet using `init_loc_fn`. I would recommend adding
```py
if init_loc_fn is not init_to_feasible:
    raise NotImplementedError("TODO")
```
and we can work on that in a follow-up PR (I'm happy to add support if you want).

pyro/infer/autoguide/guides.py
```diff
@@ -347,6 +347,145 @@ def median(self, *args, **kwargs):
         return self(*args, **kwargs)
 
 
+class AutoNormal(AutoGuide):
+    """This is AutoNormal, but it uses the same shape constraint logic found
+    in AutoMultivariateNormal and AutoDiagonalNormal.
+    """
+    def __init__(self, model, init_loc_fn=init_to_median, prefix='auto'):
```
fritzo: We removed the `prefix` arg in 1.0. You shouldn't need it after you make the `PyroModule` changes I suggest below.

Also, could you add an `init_scale` so this has the same signature as `AutoDiagonalNormal`? That way it will be super easy for users to swap between the two :smile:

patrickeganfoley: Got it - I removed `prefix` and added `init_scale`, which looks like it works 👍

pyro/infer/autoguide/guides.py
patrickeganfoley: So I made those changes - and now the serialization test passes (🎉 ) - but I'm getting `E           KeyError: 'parameter name can\'t contain "."'` errors from the register_parameter method in the test_predictive() test, which used to work.

The names in get_param_store() are now `AutoNormal.locs.<site_name>`.  Is there a way to control these?  Using `self.locs = PyroModule(name='locs')` appears not to change anything.

patrickeganfoley: Ah - I commented too soon.  Looks like the PyroModule() edits also make the median() and quantiles() tests fail (`AssertionError: cannot compare <class 'torch.nn.parameter.Parameter'> and <class 'torch.Tensor'>).  I try to get them working again tomorrow 👍 .

fritzo: > Is there a way to control [names in the param store]?

Pyro 1.0 gives up some naming freedom in return for serializability and servability (Pyro models can now be served even in C++!). That said, we could change design a bit to allow a different standard naming scheme. I suggested `AutoNormal.locs.<site_name>` because it appeared to minimize risk of name conflict with `dir(AutoNormal)` (which could happen if e.g. a user named a site `median` or `init_scale`). Did you have a different naming convention in mind?

fritzo:
Looks like tests pass with just a couple changes:

<details>

```diff
diff --git a/pyro/infer/autoguide/guides.py b/pyro/infer/autoguide/guides.py
index 0e629d48..8572219b 100644
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -58,6 +58,12 @@ def _deep_setattr(obj, key, val):
     setattr(obj, rpart, val)


+def _deep_getattr(obj, key):
+    for part in key.split("."):
+        obj = getattr(obj, part)
+    return obj
+
+
 def prototype_hide_fn(msg):
     # Record only stochastic sites in the prototype_trace.
     return msg["type"] != "sample" or msg["is_observed"] or site_is_subsample(msg)
@@ -393,12 +399,12 @@ class AutoNormal(AutoGuide):

             init_loc = site["value"].new_zeros(unconstrained_shape_for_params)
             init_scale = site["value"].new_full(unconstrained_shape_for_params, self._init_scale)
-            setattr(self.locs, name, nn.Parameter(init_loc))
-            setattr(self.scales, name, PyroParam(init_scale, constraints.positive))
+            _deep_setattr(self.locs, name, nn.Parameter(init_loc))
+            _deep_setattr(self.scales, name, PyroParam(init_scale, constraints.positive))

     def _get_loc_and_scale(self, name):
-        site_loc = getattr(self.locs, name)
-        site_scale = getattr(self.scales, name)
+        site_loc = _deep_getattr(self.locs, name)
+        site_scale = _deep_getattr(self.scales, name)
         return site_loc, site_scale

     def forward(self, *args, **kwargs):
@@ -454,6 +460,8 @@ class AutoNormal(AutoGuide):
         for name, site in self.prototype_trace.iter_stochastic_nodes():
             site_loc, _ = self._get_loc_and_scale(name)
             median = biject_to(site["fn"].support)(site_loc)
+            if median is site_loc:
+                median = median.clone()
             medians[name] = median

         return medians
```

</details>

patrickeganfoley:
@fritzo  thanks for the help!  All tests now pass except for

* test_init_loc_fn (not implemented)
* test_init_scale (this is written for AutoContinuous guides only (i.e. expects .loc_scale to exist) so I did not add it.
* test_median_module - this is the one I can't figure out.

I initially thought it was due to AutoNormal initializing locs as zeros, while the AutoContinuous used _init_loc().  However, I used the function below to try to do the same, and still got errors on test_median_module(), so I'm not sure that's it.

```    def _init_loc(self):
        """
        Creates initial latent vectors using a per-site init function.
        """
        init_locs = {}
            constrained_value = site["value"].detach()
            unconstrained_value = biject_to(site["fn"].support).inv(constrained_value)
            init_locs[name] = unconstrained_value.reshape(-1)
        return init_locs
```

(I then was using those locs in the `_setup_prototype()`).  

Should we try to get test_median_module() working as well?

pyro/infer/autoguide/guides.py
```diff
@@ -347,6 +353,157 @@ def median(self, *args, **kwargs):
...
+                site_batch_shape + (1,) * event_dim
+            )
+
+            # Collect independence contexts.
+            self._cond_indep_stacks[name] = site["cond_indep_stack"]
+
+            init_loc = site["value"].new_zeros(unconstrained_shape_for_params)
+            init_scale = site["value"].new_full(unconstrained_shape_for_params, self._init_scale)
+            _deep_setattr(self.locs, name, nn.Parameter(init_loc))
+            _deep_setattr(self.scales, name, PyroParam(init_scale, constraints.positive))
```
fritzo: On second thought I think it's easiest to support `init_loc_fn` in this PR. I think this should work:
```py
    # Collect the shapes of unconstrained values.
    # These may differ from the shapes of constrained values.
    constrained_shape = site["value"].shape
    init_loc = biject_to(site["fn"].support).inv(site["value"].detach())
    self._unconstrained_shapes[name] = init_loc.shape

    # Collect independence contexts.

    _deep_setattr(self.locs, name, nn.Parameter(init_loc))
    _deep_setattr(self.scales, name, PyroParam(init_scale, constraints.positive))
```
Then you can remove that `NotImplementedError` from `__init__` and I think even get rid of those `xfail_if_not_implemented`s in the tests.

patrickeganfoley: Awesome!  this works!  I removed the `xfail` 👍 .

patrickeganfoley:
Awesome!  Thanks so much for all your help!

PR #2051, opened 2019-10-02
Separate requirements.txt file

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,9 @@
+
+# if you add any additional libraries, please also
+# add them to `docs/requirements.txt`
+graphviz>=0.8
+# numpy is necessary for some functionality of PyTorch
+numpy>=1.7
+opt_einsum>=2.3.2
+torch>=1.2.0
+tqdm>=4.31
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,11 @@ except Exception as e:
 # This assumes the first 10 lines contain badge info.
 long_description = '\n'.join([str(line) for line in long_description.split('\n')[10:]])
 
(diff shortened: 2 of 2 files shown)


neerajprad:
Could you elaborate on the motivation for having a `requirements.txt` file; in particular, are there any tasks / workflows which this facilitates? My understanding is that it is more useful for applications to have `requirements.txt` to pin dependencies to specific versions, but since Pyro is a library, we have typically left it to our users to define their concrete dependencies as they see fit in their own `requirements.txt` for as long is it meets our minimum requirements. Just looking to understand, the proposed change looks good otherwise.

gliptak:
@neerajprad a ```requirements.txt``` can be used by your contributors (and maybe build system) to ```pip``` install dependencies

neerajprad:
> @neerajprad a requirements.txt can be used by your contributors (and maybe build system) to pip install dependencies

How about just `pip install -e .` for locally installing? Even just having `-e .` in the `requirements.txt` file should work. If there are build tools that require an explicit set of dependencies, then I can see the need for this. I can see that another use might just be to keep our `docs/requirements.txt` in sync by symlinking it to this. What do you think @jpchen, @fritzo?

fritzo:
Unless there are compelling use cases, I think it's cleanest to avoid duplication of logic in both setup.py and requrements.txt. As @neerajprad mentioned, `pip install -e .` satisfies the only use case I can think of. @gliptak can you be more specific about use cases you would like to support?

gliptak:
Thanks for the pointers. Closing at this time

PR #2052, opened 2019-10-02
Reorganize constraints and transforms modules

fritzo:
Addresses issues in #2049
 
This reorganizes the `constraints` and `transforms` modules such that:
- pyro.distributions strictly extends torch.distributions
- pyro.distributions.constraints strictly extends torch.distributions.constaints
- pyro.distributions.transforms strictly extends torch.distributions.transforms

This is important for `pyro.generic` so all constraints and transforms are accessible, both those from PyTorch and those from Pyro.

After this PR we can idiomatically:
```py
from pyro.distributions import constraints, transforms
```
or more generically, following #2049 ,
```py
from pyro.generic import distributions as dist
from pyro.generic import constraints, transforms
```

cc @stefanwebb

### DIFF
--- a/pyro/distributions/__init__.py
+++ b/pyro/distributions/__init__.py
@@ -22,8 +22,6 @@ from pyro.distributions.torch import *  # noqa F403
 from pyro.distributions.torch import __all__ as torch_dists
 from pyro.distributions.torch_distribution import TorchDistribution
 from pyro.distributions.torch_transform import TransformModule
-from pyro.distributions.transforms import *  # noqa F403
-from pyro.distributions.transforms import __all__ as pyro_transforms
 from pyro.distributions.unit import Unit
 from pyro.distributions.util import enable_validation, is_validation_enabled, validation_enabled
 from pyro.distributions.von_mises import VonMises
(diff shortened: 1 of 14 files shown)


pyro/distributions/__init__.py
```diff
@@ -19,15 +22,13 @@
 from pyro.distributions.torch import __all__ as torch_dists
 from pyro.distributions.torch_distribution import TorchDistribution
 from pyro.distributions.torch_transform import TransformModule
-from pyro.distributions.conditional import (ConditionalDistribution, ConditionalTransform, ConditionalTransformModule,
-                                            ConditionalTransformedDistribution)
-from pyro.distributions.util import enable_validation, is_validation_enabled, validation_enabled
+from pyro.distributions.transforms import *  # noqa F403
```
neerajprad: Won't this also import all torch transforms into the distributions namespace? Are there any reasons for doing so?

fritzo: Yes, you're right. We already do this, and I had avoided making this change in this PR, but I guess we can do this now.

Issue #2053, opened 2019-10-03
Make pyro.generic.testing usable as a todo list


fritzo:
Ideally I'd like to implement a very narrow testing interface with two methods that do not mention pytest and do not require testing code to mention pyro.generic. Their use is equivalent to the following:
```py
# in my_pyro_implementation/test/test_generic.py
import pytest
from pyro.generic.testing import get_test_cases, check_test_case
# n.b. do not import any pyro stuff, including pyro.generic

@pytest.mark.parametrize('test_case', get_test_cases())
def test_generic(test_case):
    try:
        check_test_case(test_case)
    except NotImplementedError as e:
        pytest.xfail(f"Not implemented:\n{e}")
```
I don't care what these are named or whether they are lists or functions or whatever, but I do think we should separate concerns such that:
1. all pyro.generic interface is hidden inside these test helpers, and
2. all testing infra (pytest) is applied outside these test helpers.

I see these helpers as kind of an autoconf mechanism for pyro implementations: running them automatically shows what features are available in a given backend.

(see original discussion in https://github.com/pyro-ppl/pyro/pull/2049#discussion_r331176695)

## Tasks
- [x] #2055 raise `NotImplementedError` in pyro.generic if a backend module is missing an attribute
- [ ] refactor pyro.generic.testing to free users from mentioning pyro
- [ ] add SVI examples to pyro.generic.testing
- [ ] exercise these tests with the pyro backend
- [ ] exercise these tests with the minipyro backend
- [ ] exercise these tests with the numpyro backend
- [ ] exercise these tests with the funsor backend

fritzo:
Moved to https://github.com/pyro-ppl/pyro-api/issues/2

PR #2055, opened 2019-10-04
Raise NotImplementedError on missing backend functionality

fritzo:
Addresses #2053 

~~Note I've been having mkl errors when using `.summary()`. It would be nice to find a non-mkl workaround `if not torch.backends.mkl.is_available()`.~~

## Tested
- added a unit test

### DIFF
--- a/tests/test_generic.py
+++ b/tests/test_generic.py
@@ -1,5 +1,4 @@
 import pytest
-import torch
 
 from pyro.generic import handlers, infer, pyro, pyro_backend
 from pyro.generic.testing import MODELS
@@ -7,6 +6,7 @@ from pyro.generic.testing import MODELS
 pytestmark = pytest.mark.stage('unit')
 
 
+@pytest.mark.filterwarnings("ignore", category=UserWarning)
 @pytest.mark.parametrize('model', MODELS)
 @pytest.mark.parametrize('backend', ['pyro'])
 def test_mcmc_interface(model, backend):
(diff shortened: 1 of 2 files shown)


neerajprad:
> if not torch.backends.mkl.is_available()

I think the diagnostics should just have `nan` if mkl isn't available (that's probably required by the fft function), and IIRC we made a change to do that sometime back. This looks like a regression to me. cc. @fehiepsi

tests/test_generic.py
```diff
@@ -16,4 +16,14 @@ def test_mcmc_interface(model, backend):
...
+            mcmc.summary()
+
+
+@pytest.mark.parametrize('backend', ['pyro', 'minipyro'])
+def test_not_implemented(backend):
+    with pyro_backend(backend):
+        pyro.sample  # should be implemented
+        pyro.param  # should be implemented
+        with pytest.raises(NotImplementedError):
+            pyro.nonexistant_primitive
```
neerajprad: nit: `existent` misspelt.

neerajprad:
LGTM! I'll fix the mkl issue in another PR.

neerajprad:
@fritzo - could you paste the error trace? I think the only diagnostic function that is likely to throw this is `effective_sample_size` but that is wrapped in a _safe block. I think the issue is that perhaps pytest is treating that warning as an error? If that's the case, we could just add `@pytest.mark.filterwarnings("ignore:..")` to the test. 

```
def _safe(fn):
    """
    Safe version of utilities in the :mod:`pyro.ops.stats` module. Wrapped
    functions return `NaN` tensors instead of throwing exceptions.

    :param fn: stats function from :mod:`pyro.ops.stats` module.
    """
    @functools.wraps(fn)
    def wrapped(sample, *args, **kwargs):
        try:
            val = fn(sample, *args, **kwargs)
        except Exception:
            warnings.warn(tb.format_exc())
            val = torch.full(sample.shape[2:], float("nan"),
                             dtype=sample.dtype, device=sample.device)
        return val

    return wrapped
```

fritzo:
Thanks, @neerajprad your suggestion of `@pytest.mark.filterwarnings` fixed my issue.

PR #2056, opened 2019-10-07
Remove deprecated MCMC code for next release

neerajprad:
Changes:

 - [x] Remove deprecated code from `mcmc.mcmc.py`, and redirect to the new interface.
 - [x] Change behavior of `.get_sample()` so that when `group_by_chain=True`, we have an empty leftmost dim as expected. Discussion in https://github.com/pyro-ppl/brmp/pull/23. 
 - [x] Also change behavior of `diagnostics`, `summary` to have a `group_by_chain` arg as in NumPyro.


pyro/infer/mcmc/api.py
```diff
@@ -361,7 +362,7 @@ def run(self, *args, **kwargs):
             next_pos = pos + shape.numel()
             # NB: squeeze in case num_chains=1
```
fehiepsi: nit: this comment is not necessary anymore

pyro/infer/mcmc/api.py
```diff
@@ -333,6 +334,7 @@ def __init__(self, kernel, num_samples, warmup_steps=None, initial_params=None,
             self.sampler = _UnarySampler(kernel, num_samples, self.warmup_steps, disable_progbar,
                                          initial_params=initial_params, hook=hook_fn)
 
+    @poutine.block
```
fehiepsi: @neerajprad why we need `block` here?

neerajprad: This is so that if we are tracing a model using the generated samples from MCMC, we can run inference within a `poutine.trace` contextmanager without capturing the traces from the inference method itself. See `test_abstract_infer.test_nesting`. That test will fail without this. This was also the behavior in `TracePosterior` class which is now deprecated.

fehiepsi: Understood, thanks for explaining!

PR #2059, opened 2019-10-11
Improve U-turn condition

fehiepsi:
This PR uses a better approximation for rho (r_sum) term, which has been shown in https://github.com/pyro-ppl/numpyro/pull/395 that the change greatly improves effective sample size for some models.

### DIFF
--- a/tests/infer/mcmc/test_nuts.py
+++ b/tests/infer/mcmc/test_nuts.py
@@ -182,9 +182,7 @@ def test_beta_bernoulli(step_size, adapt_step_size, adapt_mass_matrix, full_mass
     assert_equal(samples["p_latent"].mean(0), true_probs, prec=0.02)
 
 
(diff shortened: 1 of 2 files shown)


tests/infer/mcmc/test_nuts.py
```diff
@@ -182,9 +182,7 @@ def model(data):
     assert_equal(samples["p_latent"].mean(0), true_probs, prec=0.02)
 
 
-@pytest.mark.parametrize("jit", [False, mark_jit(True, marks=[
-    pytest.mark.skipif("CUDA_TEST" in os.environ, reason="FIXME: Too small step size with CUDA on JIT.")
-])], ids=jit_idfn)
```
fehiepsi: this error is fixed in PyTorch before the last release.

neerajprad: Thanks for cleaning this up!

Issue #1996, opened 2019-08-06
Improve binary/multi_class likelihood precision


fehiepsi:
Currently, we use `link_function` as a map from `logits` to `probs`, and feed `probs` into Binary or Categorical distributions. We can improve the performance of these likelihoods by use `logits` directly in these distributions.

PR #2060, opened 2019-10-11
Make some gp likelihoods more stable

fehiepsi:
Addresses #1996. When `response_function` takes its default value, what it does is to convert `logits` to `probs` in Binary/Categorical distributions. This PR skips applying `response_function` for the default case to take advantage of the numerical stability when using these distributions with `logits` arg.

### DIFF
--- a/examples/contrib/gp/sv-dkl.py
+++ b/examples/contrib/gp/sv-dkl.py
@@ -143,7 +143,7 @@ def main(args):
     # Turns on "whiten" flag will help optimization for variational models.
     gpmodule = gp.models.VariationalSparseGP(X=Xu, y=None, kernel=deep_kernel, Xu=Xu,
                                              likelihood=likelihood, latent_shape=latent_shape,
-                                             num_data=60000, whiten=True)
+                                             num_data=60000, whiten=True, jitter=2e-6)
     if args.cuda:
         gpmodule.cuda()
 
--- a/pyro/contrib/gp/likelihoods/binary.py
+++ b/pyro/contrib/gp/likelihoods/binary.py
@@ -40,9 +40,11 @@ class Binary(Likelihood):
         """
         # calculates Monte Carlo estimate for E_q(f) [logp(y | f)]
         f = dist.Normal(f_loc, f_var.sqrt())()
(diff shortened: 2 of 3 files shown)


examples/contrib/gp/sv-dkl.py
```diff
@@ -143,7 +143,7 @@ def main(args):
     # Turns on "whiten" flag will help optimization for variational models.
     gpmodule = gp.models.VariationalSparseGP(X=Xu, y=None, kernel=deep_kernel, Xu=Xu,
                                              likelihood=likelihood, latent_shape=latent_shape,
-                                             num_data=60000, whiten=True)
+                                             num_data=60000, whiten=True, jitter=2e-6)
```
fehiepsi: When I run this example again, I get Cholesky error so I increase jitter a little bit. The reason might come from the use of a different system or the changes in PyTorch linear algebra module.

pyro/contrib/gp/likelihoods/binary.py
```diff
@@ -40,9 +40,11 @@ def forward(self, f_loc, f_var, y=None):
...
-
-        y_dist = dist.Bernoulli(f_res)
+        if self.response_function is torch.sigmoid:
+            y_dist = dist.Bernoulli(logits=f)
+        else:
+            f_res = self.response_function(f)
+            y_dist = dist.Bernoulli(f_res)
         if y is not None:
-            y_dist = y_dist.expand_by(y.shape[:-f_res.dim()]).to_event(y.dim())
+            y_dist = y_dist.expand_by(y.shape[:-f.dim()]).to_event(y.dim())
```
fritzo: Can you confirm that `f.dim()` is never 0? Because if `f.dim()` is ever zero, you'll need the more complex expression `y.shape[:y.dim() - f.dim()]`.

fehiepsi: Yes. Kernels only allow [inputs to be 1D or 2D](https://github.com/pyro-ppl/pyro/blob/dev/pyro/contrib/gp/kernels/kernel.py#L52), so `f` will be at least 1D, even when we are applying GP on 1 data point.

PR #2061, opened 2019-10-11
Exclude seed handler from the PYRO_STACK

fritzo:
The seed handler is special in Pyro and Minipyro in that it affects state in the global rng seed, rather than in Pyro. Moreover we would like to use this handler globally, as is done in Numpyro. One thing that makes this difficult is that much of our code tests `if not PYRO_STACK: ...do shortcut...`. This PR omits the `Seed` handler from the `PYRO_STACK`, thus making it possible to use that handler globally while preserving old shortcuts.

This PR also makes the seed handler restore previous state on exit, thereby providing compatibility with the numpyro backend.

For motivation, see https://github.com/pyro-ppl/numpyro/pull/396/files#r334110296

### DIFF
--- a/pyro/poutine/handlers.py
+++ b/pyro/poutine/handlers.py
@@ -509,7 +509,9 @@ class _SeedMessenger(Messenger):
 
     def __enter__(self):
         set_rng_seed(self.rng_seed)
-        super(_SeedMessenger, self).__enter__()
+
+    def __exit__(self, type, value, traceback):
+        pass
 
 
 def seed(fn=None, rng_seed=None):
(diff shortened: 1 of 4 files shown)


pyro/poutine/handlers.py
```diff
@@ -508,8 +508,11 @@ def __init__(self, rng_seed):
         super(_SeedMessenger, self).__init__()
 
     def __enter__(self):
+        self.old_state = get_rng_state()
         set_rng_seed(self.rng_seed)
-        super(_SeedMessenger, self).__enter__()
+
+    def __exit__(self, type, value, traceback):
+        set_rng_state(self.old_state)
```
neerajprad: +1 on restoring the previous state - this makes the behavior identical across backends.

PR #2062, opened 2019-10-11
Deprecate pyro.generic in favor of pyroapi

fritzo:
~~Blocked by #2061 ~~

We have moved `pyro.generic` to a new package [pyroapi](https://github.com/pyro-ppl/pyro-api).

@neerajprad is there anything else we need to do before merging this?

## Tested

- added standard pyroapi tests
- refactored existing test_generic.py to use pyroapi
  (eventually this test should be subsumed by standard pyroapi tests)


tests/test_generic.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 
 from pyro.generic import handlers, infer, pyro, pyro_backend
-from pyro.generic.testing import MODELS
```
neerajprad: These tests are also covered in `pyro-api` and can be removed from here.

neerajprad:
> @neerajprad is there anything else we need to do before merging this?

I don't think so. Since funsor is using `pyro-api` now, we can safely remove this. You'll need to merge the changes from master though.

tests/test_generic.py
fritzo: I'd like to keep these tests in both repos so as to prevent both breakage of api and breakage of implementation. Does that make sense?

I think we can eventually move those tests into the more automatic `pyroapi.tests`, so at least code won't be duplicated.

neerajprad: Sure, that makes sense!

Issue #1930, opened 2019-06-29
Remove dependency on TracePosterior and deprecate utilities in AbstractInfer


neerajprad:
While attempting the MCMC class refactoring (as originally suggested in #1725), I realized that it will be best to not subclass off of `TracePosterior`. 

 - For MCMC, storing traces leads to inefficient memory usage where the full trace can often consume 5X more memory. This also leads to our internal trace structure leaking into the API, and makes it harder to integrate with other libraries. e.g. NUTS integration with GPyTorch.
 - For SVI, once we have the trained parameters in the param store, we can just use these each time we need to generate traces, do predictions etc. rather than maintaining a bag of traces. This has often led to confusion that users have surfaced through the forum. [example]( https://forum.pyro.ai/t/tracepredictive-worse-than-sampling-guides/715).

**TODOs:**

Next Release:
 - [ ] Refactor MCMC module to not subclass off of TracePosterior, and bring the interface closer to NumPyro, deprecate existing API. #1913 
 - [ ] Replacement for TracePredictive for MCMC.

Next Release:
 - [ ] Deprecate `.run` method in SVI module to plan for removal
 - [ ] Replacement for TracePredictive for SVI (this additionally needs to handle data subsampling).

Following Release:
 - [ ] Removing these code paths completely.

fritzo:
@neerajprad is any part of this issue blocking the 0.4 release?

neerajprad:
Deprecating `TracePosterior`/`TracePredictive` for SVI should be punted for later, since it will involve changes in other tutorials, OED module, and may need some discussion.

In the meantime, I will try to see if we can support SVI by making some minor changes to mcmc's predictive utility. As this utility becomes mature, that will make it easier for us to deprecate the old interface.

neerajprad:
@fehiepsi - Since you volunteered helping with the predictive utility, I have added you to this task. Note that we already have a `predictive` utility in `mcmc.util`. We need to figure out a more general API so that it works with `SVI` too, and which will allow us to remove `TracePredictive` completely.

fehiepsi:
Sure, I am happy to work on this. I'll try to make a PR which covers usage cases which I know, then we can discuss there how to support more general cases, which I believe that you have some experience with.

neerajprad:
[...]

PR #2063, opened 2019-10-12
Move mcmc predictive to infer

fehiepsi:
This PR attempts to make `predictive` utility work for SVI, which is part of issue #1930.

There are some issues, which are not addressed yet:

#### "_RETURN" site

+ Former behavior: "_RETURN" site is skipped in `guide` predictive and collected in `model` predictive

+ This PR behavior: "_RETURN" site is skipped for both model and guide.

This difference can be resolved by allowing collecting the "_RETURN" site through the `return_sites` keyword. But "_RETURN" is not visible in model/guide, so supporting this or not will depend on reviewers' opinion.

#### num_samples

+ Former behavior: num_samples is different for collecting samples from the guide and collecting samples from the model. Trace predictive will resample samples from guide to predict model.

I think that it is not necessary to have this distinction (maybe it is useful for important sampling). Posterior samples with important weights can be obtained separately. Then for posterior predictive, we just use those "resampled" samples for prediction.

#### param site

+ Former behavior: we can train guides with param_shape = num_data and get posterior samples with shape = num_samples x num_data. Then for prediction with data_new, we can get predictions with shape `num_samples x num_data_new` (by resampling indices?).

I don't understand the theoretical aspect of this pattern. Would it be more reasonable to have `param_shape` that does not depend on `num_data`? I think I need more explanation from reviewers.

cc @ahmadsalim

### DIFF
--- a/pyro/infer/predictive.py
+++ b/pyro/infer/predictive.py
@@ -1,0 +1,152 @@
+from functools import reduce
+import warnings
+
+import torch
+
+import pyro
+import pyro.poutine as poutine
+from pyro.poutine.util import prune_subsample_sites
+
+
+def _guess_max_plate_nesting(model, args, kwargs):
+    """
+    Guesses max_plate_nesting by running the model once
+    without enumeration. This optimistically assumes static model
+    structure.
+    """
+    with poutine.block():
+        model_trace = poutine.trace(model).get_trace(*args, **kwargs)
+    sites = [site for site in model_trace.nodes.values()
+             if site["type"] == "sample"]
+
+    dims = [frame.dim
+            for site in sites
+            for frame in site["cond_indep_stack"]
+            if frame.vectorized]
+    max_plate_nesting = -min(dims) if dims else 0
+    return max_plate_nesting
+
+
+def _predictive_sequential(model, posterior_samples, model_args, model_kwargs,
+                           num_samples, sample_sites, return_trace=False):
+    collected = []
+    samples = [{k: v[i] for k, v in posterior_samples.items()} for i in range(num_samples)]
+    for i in range(num_samples):
+        trace = poutine.trace(poutine.condition(model, samples[i])).get_trace(*model_args, **model_kwargs)
+        if return_trace:
+            collected.append(trace)
+        else:
+            collected.append({site: trace.nodes[site]['value'] for site in sample_sites})
+
+    return collected if return_trace else {site: torch.stack([s[site] for s in collected])
+                                           for site in sample_sites}
+
+
+def predictive(model, posterior_samples, *args, num_samples=None, return_sites=None,
+               return_trace=False, parallel=False, **kwargs):
+    """
+    Run model by sampling latent parameters from `posterior_samples`, and return
+    values at sample sites from the forward run. By default, only sites not contained in
+    `posterior_samples` are returned. This can be modified by changing the `return_sites`
+    keyword argument.
+
+    .. note:: Using empty dict `posterior_samples={}` for prior predictive or getting posterior
+        samples from `guide` in SVI. In those predictive, the number of samples `num_samples`
+        should be specified.
+
+    .. warning::
+        The interface for the `predictive` class is experimental, and
+        might change in the future.
+
+    :param model: Python callable containing Pyro primitives.
+    :param dict posterior_samples: dictionary of samples from the posterior.
+    :param args: model arguments.
+    :param kwargs: model kwargs; and other keyword arguments (see below).
+
+    :Keyword Arguments:
+        * **num_samples** (``int``) - number of samples to draw from the predictive distribution.
+          This argument has no effect if ``posterior_samples`` is non-empty, in which case, the
+          leading dimension size of samples in ``posterior_samples`` is used.
+        * **return_sites** (``list``) - sites to return; by default only sample sites not present
+          in `posterior_samples` are returned.
+        * **return_trace** (``bool``) - whether to return the full trace. Note that this is
+          vectorized over `num_samples`.
+        * **parallel** (``bool``) - predict in parallel by wrapping the existing model
+          in an outermost `plate` messenger. Note that this requires that the model has
+          all batch dims correctly annotated via :class:`~pyro.plate`. Default is `False`.
+
+    :return: dict of samples from the predictive distribution, or a single vectorized
+        `trace` (if `return_trace=True`).
+    """
+    # num_samples = kwargs.pop('num_samples', None)
+    # return_sites = kwargs.pop('return_sites', None)
+    # return_trace = kwargs.pop('return_trace', False)
+    # parallel = kwargs.pop('parallel', False)
+
+    max_plate_nesting = _guess_max_plate_nesting(model, args, kwargs)
+    model_trace = prune_subsample_sites(poutine.trace(model).get_trace(*args, **kwargs))
+    reshaped_samples = {}
+
+    for name, sample in posterior_samples.items():
+
+        batch_size, sample_shape = sample.shape[0], sample.shape[1:]
+
+        if num_samples is None:
+            num_samples = batch_size
+
+        elif num_samples != batch_size:
+            warnings.warn("Sample's leading dimension size {} is different from the "
+                          "provided {} num_samples argument. Defaulting to {}."
+                          .format(batch_size, num_samples, batch_size), UserWarning)
+            num_samples = batch_size
+
+        sample = sample.reshape((num_samples,) + (1,) * (max_plate_nesting - len(sample_shape)) + sample_shape)
+        reshaped_samples[name] = sample
+
+    if num_samples is None:
+        raise ValueError("No sample sites in model to infer `num_samples`.")
+
+    return_site_shapes = {}
+    for site in model_trace.stochastic_nodes + model_trace.observation_nodes:
+        site_shape = (num_samples,) + model_trace.nodes[site]['value'].shape
+        if return_sites:
+            if site in return_sites:
+                return_site_shapes[site] = site_shape
+        else:
+            if site not in reshaped_samples:
+                return_site_shapes[site] = site_shape
+
+    if not parallel:
+        return _predictive_sequential(model, posterior_samples, args, kwargs, num_samples,
+                                      return_site_shapes.keys(), return_trace)
+
+    def _vectorized_fn(fn):
+        """
+        Wraps a callable inside an outermost :class:`~pyro.plate` to parallelize
+        sampling from the posterior predictive.
+
+        :param fn: arbitrary callable containing Pyro primitives.
+        :return: wrapped callable.
+        """
+
+        def wrapped_fn(*args, **kwargs):
+            with pyro.plate("_num_predictive_samples", num_samples, dim=-max_plate_nesting-1):
+                return fn(*args, **kwargs)
+
+        return wrapped_fn
+
+    trace = poutine.trace(poutine.condition(_vectorized_fn(model), reshaped_samples))\
+        .get_trace(*args, **kwargs)
+
+    if return_trace:
+        return trace
+
+    predictions = {}
+    for site, shape in return_site_shapes.items():
+        value = trace.nodes[site]['value']
+        if value.numel() < reduce((lambda x, y: x * y), shape):
+            predictions[site] = value.expand(shape)
+        else:
+            predictions[site] = value.reshape(shape)
+
+    return predictions
--- a/pyro/infer/predictive.py
+++ b/pyro/infer/predictive.py
@@ -42,34 +42,36 @@ def _predictive_sequential(model, posterior_samples, model_args, model_kwargs,
                                            for site in sample_sites}
 
 
-def _predictive(model, posterior_samples, *args, **kwargs):
-    num_samples = kwargs.pop('num_samples', None)
-    return_sites = kwargs.pop('return_sites', None)
-    return_trace = kwargs.pop('return_trace', False)
-    parallel = kwargs.pop('parallel', False)
-
-    max_plate_nesting = _guess_max_plate_nesting(model, args, kwargs)
-    model_trace = prune_subsample_sites(poutine.trace(model).get_trace(*args, **kwargs))
+def _predictive(model, posterior_samples, num_samples, return_sites=None,
+                return_trace=False, parallel=False, model_args=(), model_kwargs={}):
+    max_plate_nesting = _guess_max_plate_nesting(model, model_args, model_kwargs)
+    model_trace = prune_subsample_sites(poutine.trace(model).get_trace(*model_args, **model_kwargs))
     reshaped_samples = {}
 
     for name, sample in posterior_samples.items():
-
-        batch_size, sample_shape = sample.shape[0], sample.shape[1:]
-
-        if num_samples is None:
-            num_samples = batch_size
-
-        elif num_samples != batch_size:
-            warnings.warn("Sample's leading dimension size {} is different from the "
-                          "provided {} num_samples argument. Defaulting to {}."
-                          .format(batch_size, num_samples, batch_size), UserWarning)
-            num_samples = batch_size
-
+        sample_shape = sample.shape[1:]
         sample = sample.reshape((num_samples,) + (1,) * (max_plate_nesting - len(sample_shape)) + sample_shape)
         reshaped_samples[name] = sample
 
-    if num_samples is None:
-        raise ValueError("No sample sites in model to infer `num_samples`.")
+    def _vectorized_fn(fn):
+        """
+        Wraps a callable inside an outermost :class:`~pyro.plate` to parallelize
+        sampling from the posterior predictive.
+
+        :param fn: arbitrary callable containing Pyro primitives.
+        :return: wrapped callable.
+        """
+
+        def wrapped_fn(*args, **kwargs):
+            with pyro.plate("_num_predictive_samples", num_samples, dim=-max_plate_nesting-1):
+                return fn(*args, **kwargs)
+
+        return wrapped_fn
+
+    if return_trace:
+        trace = poutine.trace(poutine.condition(_vectorized_fn(model), reshaped_samples))\
+            .get_trace(*model_args, **model_kwargs)
+        return trace
 
     return_site_shapes = {}
     for site in model_trace.stochastic_nodes + model_trace.observation_nodes:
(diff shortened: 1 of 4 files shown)


tests/infer/test_predictive.py
```diff
@@ -0,0 +1,92 @@
...
+    conditioned_model = poutine.condition(model, data={"obs": num_success})
+    svi = SVI(conditioned_model, beta_guide, optim.Adam(dict(lr=1.0)), Trace_ELBO())
+    for i in range(1000):
+        svi.step(num_trials)
+    posterior_samples = predictive(beta_guide, {}, num_trials[:3], num_samples=100)
+    posterior_predictive = predictive(model, posterior_samples, num_trials[:3], num_samples=10000)
+    #     svi.step(num_trials[:3])
+    # posterior_samples = predictive(beta_guide, {}, num_trials, num_samples=100)
+    # posterior_predictive = predictive(model, posterior_samples, num_trials, num_samples=10000)
+    marginal_return_vals = posterior_predictive["_RETURN"]
```
fehiepsi: I group all of the issues in this test.

ahmadsalim:
I think the code overall looks good, and I don't have major comments there 😄.

I have some clarifications on the questions/remarks on the comments. I am not sure everything I mention is strictly necessary, but it mostly reflects how I used the tool 🙂.

- In general the reason I use `_RETURN` in the model before was that, during parameter inference values would be observed (so used for likelihood calculation and not present in the trace), while for prediction I would often need the outcome to do posterior predictive checks. 
   * Perhaps, allowing deterministic nodes in general which can be recomputed from context could be useful as well. A bit like `pm.Deterministic` from PyMC.
- The reason that I often used `num_samples` higher for prediction than for collecting samples is that it was possible to resample some values locally depending on the provided input data (by excluding them from `keep_sites` in the old interface), and getting a good distribution might require more than one sample for each posterior sample value. I think this relates well to the importance sample remark.
- The reason we resample indices from the posterior, is that often one would need to run prediction of different batches (usually online) than during training (which would be mini-batched for efficiency). 
   * I think we would like the interface to allow posterior predictions, even though we cannot do ancestral resampling of the data-dependent parameters (whose size have changed).
   * Index resampling was the only reasonable way I could come up with which could provide values from the trained posterior distribution. I am not sure there is a good theoretical justification for the approach, and there can certainly be hidden biases if the training set is small while prediction is done on a larger set. I would love to discuss better alternatives.

I hope the comments are useful.

fehiepsi:
Hi @ahmadsalim , thanks so much for your feedback! About the first point, I also used `RETURN` to get those deterministic sites. But @neerajprad has suggested me a clever solution, that is to put Delta prior on those deterministic sites:
```
pyro.sample("deterministic", dist.Delta(value), obs=value)
```
The convenience of this solution is: the result of `predictive` will be consistent: a dict with keys are site names and values are torch tensors (we can't achieve the same thing if we use "_RETURN" with multi deterministic values). :D This is similar to `pm.Deterministic` in PyMC.

About resampling posterior for posterior predictive (apply `predictive` for the model to get predictions for `obs` node), would it be enough to provide a utility for it (because I think it is not directly related to `predictive`)? I guess something similar to [stats.resample](http://docs.pyro.ai/en/stable/ops.html#pyro.ops.stats.resample) with `input` is a dict (instead of torch.tensor) is enough for your purpose?

> we cannot do ancestral resampling of the data-dependent parameters (whose size have changed)

I don't quite understand in which situation, we want to optimize parameters with shapes depending on data size. I think we should construct a specific model for that purpose: e.g. the guide
```
def guide(X):
    b = pyro.param("b", torch.zeros(X.shape[0]))
    pyro.sample("f", dist.Normal(b * X, 1))
```
should be
```
def guide(X):
    b = pyro.param("b", torch.tensor(0.))
    pyro.sample("f", dist.Normal(b * X, 1))
```

Otherwise, we will run into the situation: First, we trained perfect parameters: x1, p1 predicts y1;  x2, p2 predicts y2; x3, p3 predicts y3. Then given x1 as the testing input, if we choose randomly p1, p2, p3, then it is unlikely that we will get the prediction `y1`.

For models with parameters depending on batch_size, I would consider a batch of data is a single "joined" data point. E.g. this is the situation where data has weekly information, we would like to construct a model with "weekly" parameters (to capture the pattern during the whole week). In that case, we would reshape data from `num_days x features` into ` num_weeks x 7 x features` and consider the problem as we have `num_weeks` data points with each data point having shape `7 x features`.

> run prediction of different batches (usually online) than during training (which would be mini-batched for efficiency)

I think that as long as parameter shapes are independent of batch_size, this should be not a problem. Given a new batch of inputs X_new, we use `predictive` on `guide` to get a batch of posterior samples (each with shape: `num_samples x batch_size x ...`). Then use those posterior_samples for posterior predictive (on `model`) to get `num_samples` predictions for `y_new` (hence the output would be `num_samples x batch_size x ...`).

The above thoughts are just based on my limited experience on just some patterns of modelling. Could you provide a concrete example where having parameter size depend on data size is useful? The current tests might not be good examples because we have constant `true_probs = torch.ones(5) * 0.7`, hence the assertation `assert_close(marginal_return_vals.mean, torch.ones(3) * 700, rtol=0.05)` would pass. When `true_probs` is not constant (e.g. torch.tensor([0.4, 0.5, 0.6, 0.7, 0.8]), I don't know which value we should compare `marginal_return_vals.mean` to.

ahmadsalim:
@fehiepsi Thanks for the response.

* I like the idea of using `Delta` to get deterministic values. 😄 I have two concerns however:
   - Would a delta prior not enforce a fixed-value for observations?
   - Does it work well with MCMC sampling?
* I would personally assume that `predictive` could also be used to get samples for (non-provided) obs-nodes. Of course, if you find it more relevant to separate the feature into another method, I guess that is useful too 🙂 .
* Often values under `plate` depend on the size of the observed data.
   - I personally always use a hierarchical model and do ancestral resampling of the local parameters. However, this requires that the user specifies a hierarchical model and explicitly states that they want to resample. 
- The reason I reused indices from training data was that it was an alternative that works when the parameters are kept, but I agree with you that this is not a good design pattern.
- One could alternatively give a warning that fixed local parameters (`pyro.param`) under plates depending on data size are unsupported, but I am not sure how to nicely detect that.

fehiepsi:
> Would a delta prior not enforce a fixed-value for observations?

The site `deterministic` of `pyro.sample("deterministic", dist.Delta(value), obs=value)` will hold the value `value`. If `value = a + 1` where `a = pyro.sample("a", dist.Normal())`, then the site `deterministic` will hold the values depending on the prior/posterior samples of `a` through the map: x -> x + 1 (so for a fixed `a`, we get a fixed `deterministic`).

> Does it work well with MCMC sampling?

It does not affect MCMC because log_prob of `deterministic` site is `0`.

> predictive could also be used to get samples for (non-provided) obs-nodes

Yes, it will. `predictive` can be used to get prior samples. To do prior/posterior predictive (get samples of obs nodes), we need to set `obs=None`.

> fixed local parameters (pyro.param) under plates depending on data size are unsupported

I guess you mean something likes [this test](https://github.com/pyro-ppl/pyro/blob/dev/tests/infer/test_valid_models.py#L238)? I haven't thought about it. I don't know how to do prediction properly for that model/guide pair. :( @neerajprad Could you give your opinion?

ahmadsalim:
- I understand the `Delta` prior now 😄. 
   * Side note: It would be cool if the pattern with the delta distribution you suggested was aliased to `pyro.deterministic` or similar, to make the purpose more clear. It took me some time to understand what was happening.
- I think it would be really nice to also allow posterior predictive checks (with `obs=None` or just replaying the guide trace on top of the model with user provided values) from the same interface if it is not too hard. It seems like it would make the function interface more complete.
   * A possible suggestion is to have an optional `guide=None` parameter to `predictive` which if specified samples from the guide and replays on the model (to get posterior predictive values)
- Yes, exactly the test you suggested! It would certainly be nice to have a third opinion on what is a useful behaviour in that case.

neerajprad:
Just some quick comments based on the issues you highlighted above. I'll look at the tests and provide more detailed feedback:

 - Regarding `_RETURN`, I think it is fine to skip that. The only thing is that if users want to use the same model for inference and prediction and collect certain deterministic quantities, they might need to use a pattern like below. This isn't quite as nice as `pm.Deterministic`, but putting all of that into `_RETURN` also doesn't seem like a very elegant solution. We can document this somewhere, and maybe improve on this pattern in the future, if needed.

```python
def model(..):
  .. run inference code.
  if mode == 'predict':
    .. pyro.sample('collect_node', dist.Delta(..))
```

- Regarding `num_samples`, that has been a source of confusion for users and they have brought it up on the forum multiple times, so let us just get rid of it, and provide other utilities for resampling posterior samples if needed.

- `param site`: If we have a `pyro.param` site inside a plate and the size of the plate changes is different between training and prediction, we should throw an error. I don't think it makes sense to learn parameters that are changing per mini-batch.

neerajprad:
Regarding the failing test, I think we were planning to remove support for `subsample_size` from `plate`, so this doesn't seem important to support. Let us confirm that separately.

I don't think this problem would arise if we were passing in the subsamples directly using the `subsample` argument? @ahmadsalim - are there models where that would be a serious inconvenience? i.e. instead of subsampling using:

```python
with pyro.plate(N, 5):
   pyro.sample(..)
```
, you will instead just pass in your mini-batch of size 5:

```python
with pyro.plate(N, subsample=data):
  pyro.sample(..)
```

Let me know if I'm missing something.

tests/infer/test_predictive.py
```diff
@@ -0,0 +1,92 @@
...
+def test_posterior_predictive_svi_auto_diag_normal_guide():
+    true_probs = torch.ones(5) * 0.7
+    num_trials = torch.ones(5) * 1000
+    num_success = dist.Binomial(num_trials, true_probs).sample()
+    conditioned_model = poutine.condition(model, data={"obs": num_success})
+    guide = AutoDiagonalNormal(conditioned_model)
+    svi = SVI(conditioned_model, guide, optim.Adam(dict(lr=1.0)), Trace_ELBO())
+    for i in range(1000):
+        svi.step(num_trials)
+    posterior_samples = predictive(guide, {}, num_trials, num_samples=10000)
```
neerajprad: I think we should have `predictive` take in an optional `guide` keyword argument and do this behind the scenes, as @ahmadsalim suggested.

neerajprad: I think that will also let us return `_RETURN` sites from the model trace.

fehiepsi: Yes, the behavior of having `guide` keyword should cover usage cases I can think about. I agree that we can support it for user convenience. I'll go for a final pass before requesting your review. :)

fehiepsi:
> new stddevs of possibly different size

There are some possible ways to get posterior predictive for such case:
+ Resample `theta` (or just simply take the mean accross schools of those 8 thetas and broadcast it to the new size). This way will neglect the posterior of `mu` and `tau` (because `score` only depends on `theta`), hence I guess the predicted values will not be good.
+ Reparameterize the model: sample `theta_std ~ Normal(0, 1, shape=(8,))` and put `theta = mu + tau * theta_std`. Now `score` depends on all latent variables `mu`, `tau`, and `theta_std`. We then resample `theta_std` or just simply do average+broadcast for theta_std as above (I believe both way gives similar predictive distribution).

As we discussed earlier, I think this resampling step should be decided by users. An error should be raised if the shapes are mismatched. Combining two steps getting posterior and doing posterior predictive through `guide` arg will fail for such cases. Users should do three steps separately:
```
posterior = predictive(guide)
posterior["theta"] = resample(posterior["theta"], new_shape)  # or avarage+broadcast
posterior_predictive = predictive(model, posterior)
```

> automatically erase inferred theta

`score` depends on `theta`, so I think that we should not "erase" the inferred theta. Doing it automatically is even more ambiguous to me. Users should replace
```
     theta = pyro.sample('theta', dist.Normal(mu, tau))
     return pyro.sample('scores', dist.Normal(theta, stddevs), obs=scores)
```
by
```
     return pyro.sample('scores', dist.Normal(mu,  tau + stddevs), obs=scores)
```
if they don't want to use `theta`. WDYT?

neerajprad:
@ahmadsalim - Thanks for adding that example, it will really help focus the discussion on that. 

I had some more basic question that perhaps you and @fehiepsi could help me understand:

> Here, we would like to resample theta locally using the hierarchical parameters mu and tau for new stddevs of possibly different size. 

 - In this example, I don't understand why we would like to change the size of `stddevs` or `data` during prediction. If we learned some local `theta` for each of the 8 schools, generating predictions on 10 different schools doesn't seem like a meaningful thing to do. 
 - Also, what happens if we don't resample `theta` for such hierarchical models? It seems to me that we should simply be able to reuse the guide generated values (the guide should probably have some local parameter per school) or posterior samples from MCMC. 

I don't have much experience with using `predictive` on complex models, so I may be missing something really basic, which I am hoping to better understand. It will definitely help us iterate to a good solution if we have some  concrete examples. We can develop this example further by adding a guide, doing SVI, etc. if that helps clarify the discussion.

tests/infer/test_predictive.py
```diff
@@ -0,0 +1,92 @@
...
+def one_hot_model(pseudocounts, classes=None):
+    probs_prior = dist.Dirichlet(pseudocounts)
+    probs = pyro.sample("probs", probs_prior)
+    with pyro.plate("classes", classes.size(0) if classes is not None else 1, dim=-1):
+        return pyro.sample("obs", dist.OneHotCategorical(probs), obs=classes)
+
+
+def beta_guide(num_trials):
+    phi_c0 = pyro.param("phi_c0", num_trials.new_tensor(5.0).expand([num_trials.size(0)]))
+    phi_c1 = pyro.param("phi_c1", num_trials.new_tensor(5.0).expand([num_trials.size(0)]))
```
neerajprad: @ahmadsalim: I think this is probably similar to the 8 schools example we discussed? Once we have learnt `phi_c0` and `phi_c1`, it seems to me that changing the size of `num_trials` during prediction isn't a valid thing to do without adding further information in the guide - i.e. which local parameter to use for the new_trials? It is possible that we want to predict on only a subset of originally learned data-points. e.g. for the 8 schools case, say we want to do predictions only for schools - 1, 3, and 4. In that case. I think we will need to pass those indices to the guide explicitly, and not just randomly subsample from the learned data points. 

```python
def beta_guide(num_trials, indices=None):
    phi_c0 = pyro.param("phi_c0", num_trials.new_tensor(5.0).expand([num_trials.size(0)]))
    phi_c1 = pyro.param("phi_c1", num_trials.new_tensor(5.0).expand([num_trials.size(0)]))
    if indices is not None:
        phi_c0 = phi_c0[indices]
        phi_c1 = phi_c1[indices]
```

I think most likely you were using this as a simplifying example for a more complex use case that you have. Let me know if that's the case, and if I missed something.

fehiepsi:
> why we would like to change the size of stddevs or data during prediction

I think it depends on what users want (I haven't used that model to do such prediction). But I assumed that users wanted to do so (that is they assume that the posterior `theta_std`, which is inferred from the training eight schools, gives us some information for the new testing school - again, I think users should construct a different model but here I assumed that they don't want to do so!).

> If we learned some local theta for each of the 8 schools, generating predictions on 10 different schools doesn't seem like a meaningful thing to do.

They use different thetas for each school just to fit the training set well similar to pooling and no pooling. And I think they want to take the average value of those thetas and give a prediction based on that averaged value.

In my opinion, the correct usage of such no pooling model is to give predictions for those **8** schools, given 8 new stddevs (for example, stddevs obtained from the later years), not for the new **10** schools. But... you know, there is nothing preventing them from using that model for the new **10** schools. In that case, posterior predictive depends on what users want from the posterior. And that "middle" step (e.g. resampling or avarage+broadcast,...) should be handled by users, not by some utilities such as `predictive`. To me, it is their job to make things meaningful.

neerajprad:
> They use different thetas for each school just to fit the training set well similar to pooling and no pooling. And I think they want to take the average value of those thetas and give a prediction based on that averaged value.

The example above is partially pooled. The no-pooling case isn't hierarchical and will simply be as follows:

Case 1:
```python
def no_pooling(scores, stddevs):
  with pyro.plate('data', len(scores)):
    theta = pyro.sample('theta', dist.Normal(0., 1.))
    return pyro.sample('scores', dist.Normal(theta, stddevs), obs=scores)
```

Case 2:
The fully pooled case will be:

```python
def fully_pooled(scores, stddevs):
  theta = pyro.sample('theta', dist.Normal(0., 1.))
  with pyro.plate('data', len(scores)):
    return pyro.sample('scores', dist.Normal(theta, stddevs), obs=scores)
```

For case 1, changing the size of the `data` plate does not seem justified. Doing predictions for a subset of those data-points is fine, and in that case we should probably pass in the indices to those data points, as in my comment for the test. For case 2, we should be able to pass in different sized data / stddevs without issues.  

In any case, I agree with your assessment that this isn't something that should be a responsibility of the `predictive` class. To me, it actually seems like a problem that the user should fix in their model, unless we have better motivating examples to discuss.

ahmadsalim:
@neerajprad and @fehiepsi Thanks for the comments.

I think my core point was that case 2 in @neerajprad's example should work without issues. I would not expect case 1 to work for the comments you have made (it does not make sense to reuse samples that are local to the data). 

I think the 8-schools example could potentially work if the parameters the user provide are not under a plate (data-size dependent). I do not think that one should reuse the local samples of `theta` from training, but rather use the learned `tau` and `mu` to sample again for prediction.

It does not have to be in `predictive`, if general utilities like `poutine.block` work for this use-case it would be cool too!

pyro/infer/predictive.py
```diff
@@ -0,0 +1,170 @@
...
+            continue
+        if value.numel() < reduce((lambda x, y: x * y), shape):
+            predictions[site] = value.expand(shape)
+        else:
+            predictions[site] = value.reshape(shape)
+
+    return predictions
+
+
+def predictive(model, posterior_samples, *args, **kwargs):
```
neerajprad: Since we are moving this, I think we can also clean up the interface as follows. WDYT?

```
def predictive(model, posterior_samples=None, guide=None, return_trace=False, return_sites=(), 
               parallel=False, model_args=(), model_kwargs={})
```

Then, we don't need to pass in `posterior_samples=None` for SVI and can just pass the guide. I think this is also one of those cases where its just cleaner to pass the `model_args` and `model_kwargs` separately rather than using varargs (code editors / notebooks will display all the arguments available).

pyro/infer/predictive.py
```diff
@@ -0,0 +1,170 @@
...
+import warnings
+
+import torch
+
+import pyro
+import pyro.poutine as poutine
+from pyro.poutine.util import prune_subsample_sites
+
+
+def _guess_max_plate_nesting(model, args, kwargs):
```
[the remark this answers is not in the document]
fritzo: I'm fine moving this helper if the logic is indeed duplicated. However beware falsely unifying subtly different functions. There was a refactoring effort a while ago to "deduplicate" logic among ELBO implementations, and the result was a morass of pyro/infer/enum.py that now breaks all dependency ordering and has no natural boundaries. I'd prefer three slightly different implementations than one weird conglomerate of things as disparate as, say probabilistic programming, deep reinforcement learning, and evolutionary algorithms.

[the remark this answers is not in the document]
neerajprad: I think we can use the same implementation as the one in `elbo` and just pass the guide or an empty guide if no guide is provided. I will leave it to Du to make that determination. If it requires modifications to the utility function, then we should just replicate it as we are doing here.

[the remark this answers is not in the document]
neerajprad: @fehiepsi : On second thoughts, this will require some changes inside HMC utilities, so its best to leave it as is right now to not complicate this PR.

pyro/infer/predictive.py
fehiepsi: I have gone back and forth to find an easy-to-interpret args/kwargs for this utility, but none of them satisfied me. :( Regarding `model_args`, `model_kwargs`, I think we should get more opinions about it. The problem is most keywords are not necessary, but model `args/kwargs` are required to have. I think it is also good to make `predictive` a class with a method `get_samples` to return samples given particular args, kwargs; and another method `get_trace` to return traces. WDYT?

neerajprad: > The problem is most keywords are not necessary, but model args/kwargs are required to have. 

kwargs won't be required - by default it will be `{}` or `None`. Same with args - by default it will be an empty tuple - `()`.

> I think it is also good to make predictive a class with a method get_samples to return samples given particular args, kwargs; and another method get_trace to return traces. WDYT?

I think this is a better option. Lets do this instead.

fehiepsi: Yeah, making it a class looks good. I don't feel it messy anymore. The tests look cleaner than before. :)

neerajprad:
@ahmadsalim - If you notice anything that can be further improved with the interface, please don't hesitate to open an issue!

ahmadsalim:
Sounds great, thanks! 😄

Issue #2054, opened 2019-10-04
FR Add tutorial on how to serve trained models from C++


fritzo:
Originated [as a forum post](https://forum.pyro.ai/t/how-to-integrate-pyro-model-to-c/1330)

Let's add a tutorial for how to serve trained models in C++ using [torch.jit.save()](https://pytorch.org/docs/stable/jit.html#torch.jit.save) and `torch::jit::load()`. I believe there is some subtlety requiring a `torch.jit.trace()` to wrap `poutine.trace` and `poutine.replay` logic for SVI models and guides.

PR #2073, opened 2019-10-22
Add example for model serving using ScriptModule

neerajprad:
Addresses #2054. This adds a simple example for model serving using ModuleScript. Note that we cannot use scripting due to limitations like lack of support for context manager, but I was able to use `torch.jit.trace_module` for this purpose. 

TODOs:
 - [x] Verify SVI model serving.
 - [ ] Verify NUTS model serving.
 - [ ] Debug why the test example fails on CI.

We are storing the learned parameter values as constants in the trace for SVI. I had to write a `no_grad` context manager for the param store without which JIT complains when we trace the predictor about storing constants with `requires_grad=True`. For NUTS, I think we will need to store the posterior samples as `nn.Parameter` in the Predictor class.

It will be nice if we can do this as part of the `Predictive` utility class eventually, once we see more patterns emerge for different kinds of models.


examples/model_serving.py
```diff
@@ -0,0 +1,145 @@
+"""
+This example demonstrates using torch.jit.ModuleScript for model deployment.
```
fritzo: ModuleScript -> ScriptModule and add a link https://pytorch.org/docs/stable/jit.html#torch.jit.ScriptModule

examples/model_serving.py
```diff
@@ -0,0 +1,145 @@
...
+The Bayesian Regression model is taken from http://pyro.ai/examples/bayesian_regression.html.
+
+In particular, the prediction function can be saved as a `torch.jit.ModuleScript`, and
```
fritzo: ditto

examples/model_serving.py
```diff
@@ -0,0 +1,145 @@
...
+be run separately as a C++ program without a Python runtime. To do so, we recommend the
+following approach:
+
+ - Limit state initialization in model / guide callable passed to `SVI`.
+   e.g. initializing NNs. An easy way to ensure this is to inherit from
+   `nn.Module`, and do this initialization in `__init__`.
+ - Likewise, have a separate prediction function that derives from `nn.Module`.
+   This allows us to store parameters as `nn.Parameter` attributes that can be
+   serialized by `torch.jit.save`. For SVI, we can also simply store the learned
```
fritzo: nit: add link: https://pytorch.org/docs/stable/jit.html#torch.jit.save

pyro/params/param_store.py
```diff
@@ -96,6 +119,8 @@ def __getitem__(self, name):
         Get the constrained value of a named parameter.
         """
         unconstrained_value = self._params[name]
+        if self._no_grad:
+            unconstrained_value = self._fetch_no_grad(unconstrained_value)
```
fritzo: What is the advantage of mutating `unconstrained_value.requires_grad` rather than `.detach()`ing?

examples/model_serving.py
```diff
@@ -0,0 +1,145 @@
...
+class Predictor(nn.Module):
+    def __init__(self, model, guide):
+        super(Predictor, self).__init__()
+        self.model = model
+        self.guide = guide
+        self.params = {}
+
+    def forward(self, x, y):
+        # Replay using saved guide parameters
+        guide_trace = poutine.trace(poutine.replay(self.guide, params=self.params)).get_trace(x, y=y)
```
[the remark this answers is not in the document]
neerajprad: Sorry, that's left over from my previous commit where I was storing `self.params` as `nn.Parameters`. Its not needed now since we are simply storing these parameters as constants in the trace. 

> Will it be simpler if we get params from param stores, store non-grad version in Predictor, then jit

I was doing exactly that in my previous commit, but its not needed for SVI. It might still be better to do it explicitly so that we can reuse the same pattern for NUTS. I'll try to keep that in mind as I get the NUTS predictive version working. 

> but I can't figure out why param_no_grad is needed. 

I think param initialization results in creating tensors that have `requires_grad=True` and JIT tracing complains about not having constants that require grad, even if these constants are not eventually used.

pyro/params/param_store.py
neerajprad: I was planning to use `.detach` initially, but it still gives the problem of `RuntimeError: Cannot insert a Tensor that requires grad as a constant. Consider making it a parameter or input, or detaching the gradient`. My guess is that `detach`ing creates a new wrapper but the old tensor still exists in the param store and still has requires_grad set to True. As such, `jit.trace` complains about tensor constants that require grad. Whereas if we set `.requires_grad_`, it sets the flag on the existing tensor itself and the tracer is happy when all constant tensors don't require grad.

examples/model_serving.py
```diff
@@ -0,0 +1,145 @@
...
+        loss = svi.step(x_data, y_data)
+        if j % 100 == 0:
+            logging.info("[iter %04d] loss: %.4f" % (j, loss / len(data)))
+
+    print_mse(predict_fn, x_data, y_data, "after training")
+
+    # Turn requires_grad=False on parameter values so that these are
+    # stored as constants during tracing.
+    with param_no_grad():
+        predict_module = torch.jit.trace_module(predict_fn, {"forward": (x_data, y_data)}, check_trace=False)
```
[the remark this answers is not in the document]
neerajprad: This was only meant to save time; let me see how much the overhead is and if we can remove this now.

neerajprad:
> Are ci failures due to unwritable temp dir? Maybe cd into /tmp ?

This happens when we try to save the model. We are able to load other datasets / read from the `.data` directory in other examples without issues. I'm not sure why this fails. The exception message is also not very clear:

```
  File "/home/travis/build/pyro-ppl/pyro/examples/model_serving.py", line 140, in main
    torch.jit.save(predict_module, os.path.join(DATA_DIR, 'reg_predict.pt'))
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/torch/jit/__init__.py", line 223, in save
    m.save(f, _extra_files=_extra_files)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/torch/jit/__init__.py", line 1550, in save
    return self._c.save(*args, **kwargs)
RuntimeError: [enforce fail at inline_container.cc:319] . PytorchStreamWriter failed opening archive.
frame #0: c10::ThrowEnforceNotMet(char const*, int, char const*, std::string const&, void const*) + 0x47 (0x7f53cc87ce17 in /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/torch/lib/libc10.so)
frame #1: caffe2::serialize::PyTorchStreamWriter::valid(char const*) + 0xe6 (0x7f53ce6d1e96 in /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/torch/lib/libtorch.so)
```

examples/model_serving.py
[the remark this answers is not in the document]
fehiepsi: Understood! Thanks, Neeraj!

neerajprad:
Closing in favor of #2162.

PR #2074, opened 2019-10-22
typo fix: reparametrize -> reparameterize

bpshaver:


fritzo:
both spellings are valid english

PR #2077, opened 2019-10-22
seed ops.ssm_gp with some matern kernel utilities

martinjankowiak:
this is basically done but i still want to add support for nu=2.5

will have more demanding tests in the context of contrib.timeseries

### DIFF
--- a/pyro/ops/ssm_gp.py
+++ b/pyro/ops/ssm_gp.py
@@ -16,8 +16,10 @@ class MaternKernel(nn.Module):
 
     :param float nu: The order of the Matern kernel (1.5 or 2.5)
     :param int num_gps: the number of GPs
-    :param torch.tensor log_length_scale_init: optional `num_gps`-dimensional vector of initializers for the length scale
-    :param torch.tensor log_kernel_scale_init: optional `num_gps`-dimensional vector of initializers for the kernel scale
+    :param torch.tensor log_length_scale_init: optional `num_gps`-dimensional vector of initializers
+        for the length scale
+    :param torch.tensor log_kernel_scale_init: optional `num_gps`-dimensional vector of initializers
+        for the kernel scale
 
     **References**
 
@@ -35,12 +37,12 @@ class MaternKernel(nn.Module):
         if log_length_scale_init is not None:
             assert log_length_scale_init.shape == (num_gps,)
         else:
-            log_length_scale_init = 0.01 * torch.randn(num_gps)
+            log_length_scale_init = torch.zeros(num_gps)
 
         if log_kernel_scale_init is not None:
             assert log_kernel_scale_init.shape == (num_gps,)
         else:
-            log_kernel_scale_init = 0.01 * torch.randn(num_gps)
+            log_kernel_scale_init = torch.zeros(num_gps)
 
         super(MaternKernel, self).__init__()
 
@@ -73,7 +75,7 @@ class MaternKernel(nn.Module):
                     dt * self.mask10 + \
                     (1.0 - root_three * dt_rho) * self.mask11
             return torch.exp(-root_three * dt_rho) * trans
-        else:
+        elif self.nu == 2.5:
             rho = self.log_length_scale.exp().unsqueeze(-1).unsqueeze(-1)
             dt_rho = root_five * dt / rho
             dt_rho_sq = dt_rho.pow(2.0)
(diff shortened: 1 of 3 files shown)


pyro/ops/ssm_gp.py
```diff
@@ -0,0 +1,113 @@
...
+    def process_covariance(self, A):
+        """
+        Given a transition matrix `A` computed with `transition_matrix` compute the
+        the process covariance as described in Eqn. 3.11 in reference [2].
+
+        :returns torch.Tensor: a batched covariance matrix of shape (num_gps, state_dim, state_dim)
+        """
+        assert A.shape == (self.num_gps, self.state_dim, self.state_dim)
+        p = self.stationary_covariance()
+        q = p - torch.matmul(A.transpose(-1, -2), torch.matmul(p, A))
```
fritzo: I believe you can abbreviate
```py
q = p - A.transpose(-1, -2) @ p @ A
```

martinjankowiak: thanks but i really don't like using `@` 😬

martinjankowiak:
i was going to include the log_prob tests in contrib.timeseries since otherwise i have to replicate a bunch of code to compute log_probs

martinjankowiak:
@fritzo i fixed nu=2.5 so this is now ready for review. 

tests for exact correspondence between this and T^3 ops to follow elsewhere

(but actually the existing tests are already pretty demanding; a typo would probably cause something to be non PSD)

pyro/ops/ssm_gp.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class MaternKernel(nn.Module):
+    """
+    Provides the building blocks for representing univariate Gaussian Processes (GPs)
+    with Matern kernels as state space models.
+
+    :param float nu: The order of the Matern kernel (1.5 or 2.5)
+    :param int num_gps: the number of GPs
+    :param torch.tensor log_length_scale_init: optional `num_gps`-dimensional vector of initializers
```
fritzo: `torch.tensor` -> `torch.Tensor`

pyro/ops/ssm_gp.py
```diff
@@ -0,0 +1,139 @@
...
+        for the length scale
+    :param torch.tensor log_kernel_scale_init: optional `num_gps`-dimensional vector of initializers
```
fritzo: `torch.tensor` -> `torch.Tensor`

pyro/ops/ssm_gp.py
```diff
@@ -0,0 +1,139 @@
...
+    [2] `Stochastic Differential Equation Methods for Spatio-Temporal Gaussian Process Regression`,
+        Arno Solin.
+    """
+    def __init__(self, nu=1.5, num_gps=1, log_length_scale_init=None, log_kernel_scale_init=None):
+        assert nu in [1.5, 2.5], "The only supported values of nu are 1.5 and 2.5"
+        self.nu = nu
+        self.state_dim = {1.5: 2, 2.5: 3}[nu]
+        self.num_gps = num_gps
+
+        if log_length_scale_init is not None:
```
fritzo: nit: in general it is cleaner to locate assertions after control flow so it is obvious that both code paths follow the same invariant. While you can indeed figure that out by reading the code, `sed` isn't as smart and often makes mistakes when editing code.
```py
if log_length_scale is None:
    log_length_scale_init = torch.zeros(num_gps)
assert log_length_scale_init.shape == (num_gps,)

if log_kernel_scale_init is None:
    log_kernel_scale_init = torch.zeros(num_gps)
assert log_kernel_scale_init.shape == (num_gps,)
```

pyro/ops/ssm_gp.py
```diff
@@ -0,0 +1,139 @@
...
+
+    **References**
+
+    [1] `Kalman Filtering and Smoothing Solutions to Temporal Gaussian Process Regression Models`,
+        Jouni Hartikainen and Simo Sarkka.
+    [2] `Stochastic Differential Equation Methods for Spatio-Temporal Gaussian Process Regression`,
+        Arno Solin.
+    """
+    def __init__(self, nu=1.5, num_gps=1, log_length_scale_init=None, log_kernel_scale_init=None):
+        assert nu in [1.5, 2.5], "The only supported values of nu are 1.5 and 2.5"
```
fritzo: nit:
```py
if nu not in [1.5, 2.5]:
    raise NotImplementedError("The only supported values of nu are 1.5 and 2.5")
```

martinjankowiak: done

pyro/ops/ssm_gp.py
martinjankowiak: done

martinjankowiak: done

pyro/ops/ssm_gp.py
martinjankowiak: done

pyro/ops/ssm_gp.py
martinjankowiak: done

PR #2079, opened 2019-10-22
Fix initialization in bayesian regression tutorial

### LINKED ISSUE #2047: Cannot reproduce "Bayesian Regression - Introduction (Part 1)"
Hello, 

I'm getting non-sense predictions from the example in the Pyro site: <http://pyro.ai/examples/bayesian_regression.html?highlight=africa>

Here's what I'm running:

```python
import os
from functools import partial
import numpy as np
import pandas as pd
import seaborn as sns
import torch
import torch.nn as nn

import pyro
from pyro.distributions import Normal, Uniform, Delta
from pyro.optim import Adam
from pyro.distributions.util import logsumexp
from pyro.infer import EmpiricalMarginal, SVI, Trace_ELBO, TracePredictive
import pyro.optim as optim
from pyro.contrib.autoguide import AutoDiagonalNormal, AutoMultivariateNormal

# for CI testing
smoke_test = ('CI' in os.environ)
assert pyro.__version__.startswith('0.4.1')
pyro.enable_validation(True)
pyro.set_rng_seed(1)
pyro.enable_validation(True)


%matplotlib inline
#logging.basicConfig(format='%(message)s', level=logging.INFO)
# Enable validation checks
pyro.enable_validation(True)
smoke_test = ('CI' in os.environ)
pyro.set_rng_seed(1)
DATA_URL = "https://d2hg8soec8ck9v.cloudfront.net/datasets/rugged_data.csv"
data = pd.read_csv(DATA_URL, encoding="ISO-8859-1")
df = data[["cont_africa", "rugged", "rgdppc_2000"]]
#rugged_data = rugged_data[rugged_data["rgdppc_2000"].notnull()]
df = df[np.isfinite(df.rgdppc_2000)]
df["rgdppc_2000"] = np.log(df["rgdppc_2000"])

class RegressionModel(nn.Module):
    def __init__(self, p):
        # p = number of features
        super(RegressionModel, self).__init__()
        self.linear = nn.Linear(p, 1)
        self.factor = nn.Parameter(torch.tensor(1.))

    def forward(self, x):
        return self.linear(x) + (self.factor * x[:, 0] * x[:, 1]).unsqueeze(1)

p = 2  # number of features
regression_model = RegressionModel(p)

loc = torch.zeros(1, 1)
scale = torch.ones(1, 1)
# define a unit normal prior
prior = Normal(loc, scale)
# overload the parameters in the regression module with samples from the prior
lifted_module = pyro.random_module("regression_module", regression_model, prior)
# sample a nn from the prior
sampled_reg_model = lifted_module()

def model(x_data, y_data):
    # weight and bias priors
    w_prior = Normal(torch.zeros(1, 2), torch.ones(1, 2)).to_event(1)
    b_prior = Normal(torch.tensor([[8.]]), torch.tensor([[1000.]])).to_event(1)
    f_prior = Normal(0., 1.)
    priors = {'linear.weight': w_prior, 'linear.bias': b_prior, 'factor': f_prior}
    scale = pyro.sample("sigma", Uniform(0., 10.))
    # lift module parameters to random variables sampled from the priors
    lifted_module = pyro.random_module("module", regression_model, priors)
    # sample a nn (which also samples w and b)
    lifted_reg_model = lifted_module()
    with pyro.plate("map", len(x_data)):
        # run the nn forward on data
        prediction_mean = lifted_reg_model(x_data).squeeze(-1)
        # condition on the observed data
        pyro.sample("obs",
                    Normal(prediction_mean, scale),
                    obs=y_data)
        return prediction_mean

optim = Adam({"lr": 0.03})
svi = SVI(model, guide, optim, loss=Trace_ELBO(), num_samples=1000)

def train():
    pyro.clear_param_store()
    for j in range(num_iterations):
        # calculate the loss and take a gradient step
        loss = svi.step(x_data, y_data)
        if j % 100 == 0:
            print("[iteration %04d] loss: %.4f" % (j + 1, loss))

train()

get_marginal = lambda traces, sites:EmpiricalMarginal(traces, sites)._get_samples_and_weights()[0].detach().cpu().numpy()

def summary(traces, sites):
    marginal = get_marginal(traces, sites)
    site_stats = {}
    for i in range(marginal.shape[1]):
        site_name = sites[i]
        marginal_site = pd.DataFrame(marginal[:, i]).transpose()
        describe = partial(pd.Series.describe, percentiles=[.05, 0.25, 0.5, 0.75, 0.95])
        site_stats[site_name] = marginal_site.apply(describe, axis=1) \
            [["mean", "std", "5%", "25%", "50%", "75%", "95%"]]
    return site_stats

def wrapped_model(x_data, y_data):
    pyro.sample("prediction", Delta(model(x_data, y_data)))

posterior = svi.run(x_data, y_data)

trace_pred = TracePredictive(wrapped_model,
                             posterior,
                             num_samples=1000)

post_pred = trace_pred.run(x_data, None)

post_summary['prediction']['mean'].mean()
-362.3633981143727
```

I should be getting a positive number.

neerajprad:
~This is a temporary fix for #2047. We do not fully understand why `init_to_median` performs worse than `init_to_feasible` or `init_to_mean` on this example.~ 

Fixes #2047. We had on overly wide prior on the NN's bias (std of 1000). We could have addressed that by increasing the number of samples used to estimate the median, but I chose to decrease the prior's width instead.


neerajprad:
I am seeing an unrelated test failure:

```
tests/test_examples.py::test_cpu[sparse_regression.py --num-steps=2 --num-data=50 --num-dimensions 30] I 	 Running:
python examples/sparse_regression.py --num-steps=2 --num-data=50 --num-dimensions 30
Beginning MAP optimization...
[step 0000]  loss: 8.67906
[step 0001]  loss: 6.10070
Expected singleton thetas:
 [ 0.29434243  0.358192   -0.02293254 -0.14172214  0.01926203  0.06307288
 -0.35329664 -0.02380897 -0.13799758  0.25468335]
Identified the following active dimensions: [0 1]
Mean estimate for active singleton weights:
 [0.1048793  0.13842413]
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/numpy/lib/shape_base.py", line 843, in split
    len(indices_or_sections)
TypeError: object of type 'int' has no len()
  File "/home/travis/build/pyro-ppl/pyro/examples/sparse_regression.py", line 321, in <module>
    main(args)
  File "/home/travis/build/pyro-ppl/pyro/examples/sparse_regression.py", line 285, in main
    pyro.param('auto_var_obs').double())
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/torch/autograd/grad_mode.py", line 49, in decorate_no_grad
    return func(*args, **kwargs)
  File "/home/travis/build/pyro-ppl/pyro/examples/sparse_regression.py", line 172, in compute_posterior_stats
    active_quadratic_dims = np.split(active_quadratic_dims, active_quadratic_dims.shape[0])
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/numpy/lib/shape_base.py", line 847, in split
    if N % sections:
ZeroDivisionError: integer division or modulo by zero
FAILED
```

fritzo:
> I am seeing an unrelated test failure:

I believe it is related: that file does use `init_to_median()`. Maybe we can replace with `init_to_mean`?

fritzo:
Alternatively, we can unwrap `Independent` distributions:
```py
fn = site["fn"]
while isinstance(fn, Independent):
    fn = fn.base_dist
...
```

fritzo:
From offline discussion, it appears the problem is due to the `init_to_median` change. @fritzo suggested making the following change, either here or before 1.0:
```py
from pyro.distributions import Independent, MaskedDistribution

def is_multivariate(d):
    while isinstance(d, (Independent, MaskedDistribution)):
        d = d.base_dist
    return all(size <= 1 for size in d.event_shape)
```
```diff
 def init_to_median(site, num_samples=15):
     """
     Initialize to the prior median; fallback to a feasible point if median is
     undefined.
     """
+    # The median undefined for multivariate distributions.
+    if is_multivariate(site["fn"]):
+        return init_to_feasible(site)
     try:
         # Try to compute empirical median.
         samples = site["fn"].sample(sample_shape=(num_samples,))
     ...
```

neerajprad:
@fritzo - I think we should make the change above that you mentioned where we dispatch to `init_to_feasible` for multivariate distributions. 

In this case, what I observed is that, the prior on the bias parameter for the NN is `dist.Normal(8. * torch.ones(bias_shape), 1000.)` and `init_to_feasible` gives values like `-149` and it takes a long time for the NN to converge from then on (`init_to_feasible` or `mean` just gives 0). So we should just fix the prior. 😄

neerajprad:
Marking the sparse regression example as xfail since it seems to flakily fail on travis.

neerajprad:
Good to merge, finally.

PR #2086, opened 2019-10-23
add block_diag() tensor helper

martinjankowiak:


pyro/ops/tensor_utils.py
```diff
@@ -0,0 +1,18 @@
...
+
+    :param torch.Tensor m: 3-dimensional input tensor
+    :returns torch.Tensor: a 2-dimensional block diagonal tensor
+    """
+    assert m.dim() == 3, "Input to block_diag() must be a 3-dimensional tensor"
+    batch_size = m.size(0)
+    block_shape = m.shape[-2:]
+    eye = torch.eye(batch_size).type_as(m).unsqueeze(-2)
+    eye = eye.reshape(eye.shape + (1,))
+    m = (m.unsqueeze(-2) * eye).reshape(torch.Size(torch.tensor(block_shape) * batch_size))
```
fritzo: This seems sketchy, e.g. if the default device is cuda. What do you think of this version?
```py
B, M, N = m.shape
eye = torch.eye(B, dtype=m.dtype, device=m.device).reshape(B, 1, B, 1)
return (m.unsqueeze(-2) * eye).reshape(B * M, B * N)
```

martinjankowiak: sure, that's cleaner. done

PR #2087, opened 2019-10-23
Remove f string.

collincchoy:
It looks like an f-string snuck into 0.5.0 breaking compatibility with python <3.6. I imagine this was an accident as it looks like there's only this single occurrence and I don't see the docs updated to require 3.6+

### DIFF
--- a/pyro/ops/stats.py
+++ b/pyro/ops/stats.py
@@ -424,8 +424,8 @@ def crps_empirical(pred, truth):
     :rtype: torch.Tensor
     """
     if pred.dim() != 1 + truth.dim() or pred.shape[1:] != truth.shape:
-        raise ValueError(f"Expected pred to have one extra sample dim on left. "
-                         f"Actual shapes: {pred.shape} versus {truth.shape}")
+        raise ValueError("Expected pred to have one extra sample dim on left. "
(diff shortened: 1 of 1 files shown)


fritzo:
Thanks for the fix! Is this causing failure to `pip install pyro` in Python <3.6? If this is a major issue, we can push another quick 0.5.1 release. cc @neerajprad

fritzo:
@neerajprad do you think we should test with Python 3.4 on travis?

collincchoy:
I noticed the failure on _import_ and _not install_.

neerajprad:
> @neerajprad do you think we should test with Python 3.4 on travis?

Maybe we can explicitly set [python_requires](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires) in setup.py, and use the minimum version for running tests on travis? I think we can probably use Python 3.5; from whatever I have seen, Python 3.4 and less is a very small sliver of users.

fritzo:
I agree we should set `python_requires`; I've started doing so in funsor. Maybe we can simply manually test on Python 3.4 before each release.

eb8680:
> Maybe we can simply manually test on Python 3.4 before each release.

Does PyTorch support Python 3.4?  Their website only shows releases for 2.7 and 3.5+.

PR #2089, opened 2019-10-23
Bump version to 0.5.1 that is python 3.5 compatible

neerajprad:
Preparing to cut a 0.5.1 release that is compatible with python 3.5:
 - Moving travis to using python 3.5 instead. If this is really slow, we can still use python 3.6 for testing and separately test release candidates locally.
 - Updated readme to remove conversion to .rst, since twine can publish markdown based long description. Also updated `python_requires`.
 - Updating PyTorch to 1.3 in setup.py. I had missed doing that in the earlier version and while the last released changes are compatible with PyTorch 1.2 as well, I think it is better to bump up the PyTorch version to what has been tested. 

Also, fix included for #2091.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,7 +82,7 @@ setup(
         # numpy is necessary for some functionality of PyTorch
         'numpy>=1.7',
         'opt_einsum>=2.3.2',
-        'pyro-api>=0.1.0',
+        'pyro-api>=0.1.1',
         'torch>=1.3.0',
         'tqdm>=4.36',
     ],
--- a/setup.py
+++ b/setup.py
@@ -108,7 +108,7 @@ setup(
             'yapf',
         ],
     },
-    python_requires='>=3.4',
+    python_requires='>=3.5',
     keywords='machine learning statistics probabilistic programming bayesian modeling pytorch',
     license='MIT License',
     classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -35,17 +35,18 @@ if commit_sha:
 with open(os.path.join(PROJECT_PATH, 'pyro', '_version.py'), 'w') as f:
     f.write(VERSION.format(version))
 
-# Populate long description from README.md
+
+# READ README.md for long description on PyPi.
+# This requires uploading via twine, e.g.:
+# $ python setup.py sdist bdist_wheel
+# $ twine upload --repository-url https://test.pypi.org/legacy/ dist/*  # test version
+# $ twine upload dist/*
 try:
     long_description = open('README.md', encoding='utf-8').read()
 except Exception as e:
     sys.stderr.write('Failed to read README.md\n'.format(e))
     sys.stderr.flush()
     long_description = ''
-
-# Remove badges since they will always be obsolete.
-# This assumes the first 10 lines contain badge info.
-long_description = '\n'.join([str(line) for line in long_description.split('\n')[10:]])
 
 # examples/tutorials
 EXTRAS_REQUIRE = [
@@ -58,9 +59,6 @@ EXTRAS_REQUIRE = [
     'seaborn',
     'wget',
 ]
-
-if sys.version_info[0] == 2:
-    EXTRAS_REQUIRE.append('functools32')
 
 setup(
     name='pyro-ppl',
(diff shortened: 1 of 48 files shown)


setup.py
```diff
@@ -35,22 +35,18 @@
...
-    long_description = pypandoc.convert('README.md', 'rst')
-    print(long_description)
+    long_description = open('README.md', encoding='utf-8').read()
 except Exception as e:
-    sys.stderr.write('Failed to convert README.md to rst:\n  {}\n'.format(e))
+    sys.stderr.write('Failed to read README.md\n'.format(e))
     sys.stderr.flush()
-    long_description = open('README.md').read()
-
-# Remove badges since they will always be obsolete.
```
fritzo: You'll want to keep the badge removal bit. But I think we need to update to "first 12 lines"

neerajprad: We found that badges seem to work fine, at least with markdown. e.g. here's the numpyro pypi description - https://pypi.org/project/numpyro/ with the badges intact.

setup.py
```diff
@@ -122,7 +117,10 @@
         'Intended Audience :: Science/Research',
         'Operating System :: POSIX :: Linux',
         'Operating System :: MacOS :: MacOS X',
+        'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.8',
```
fritzo: PyTorch does not claim to support 3.8:
```
            'Programming Language :: Python :: 2',
            'Programming Language :: Python :: 2.7',
            'Programming Language :: Python :: 3',
            'Programming Language :: Python :: 3.5',
            'Programming Language :: Python :: 3.6',
            'Programming Language :: Python :: 3.7',
```

neerajprad:
@fritzo - I think we'll need to do a pyro-api release too that is python 3.5 compatible! 😄

setup.py
```diff
@@ -114,6 +108,7 @@
             'yapf',
         ],
     },
+    python_requires='>=3.4',
```
fritzo: `'>=3.5'`?

fehiepsi:
@neerajprad Let me try to fix #2091 before we release 0.5.2. I observed that issue in the latest tqdm version.

setup.py
fritzo: The problem is not that badges don't show up, the problem is that "they will always be obsolete". Pypi packages are static historical artifacts, whereas those badges are dynamic. If your build happens to be failing at the time of Pypi upload, then that artifact will forever print "build failing" (this has actually happened to me).

fritzo: but I don't care so much. oh well.

neerajprad: I see. Thanks for explaining, let me fix that.

neerajprad:
> @neerajprad Let me try to fix #2091 before we release 0.5.2. I observed that issue in the latest tqdm version.

Do we know what the issue is? Lets try to get it out in `0.5.1` then, i.e. before this PR is merged.

fehiepsi:
As explained by @NarayanSchuetz, we should set [bar_format](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/mcmc/logger.py#L61) to the new default value, which is `'{l_bar}{bar}{r_bar}'` instead of None in the new version of tqdm. To make the change compatible with the old version, I would suggest adding the following line:
```
        bar_format_kwarg = {} if bar_format is None else {"bar_format": bar_format}
```
then later, replace
```
pbar = pbar_cls(total=total_steps, desc=description, bar_format=bar_format,
                            position=i, file=sys.stderr, disable=disable)
```
by
```
pbar = pbar_cls(total=total_steps, desc=description, position=i, file=sys.stderr,
                disable=disable, **bar_format)
```
Do you want me to create a new PR for this or it is better to make the fix in this PR because Travis is slow?

neerajprad:
> Do you want me to create a new PR for this or it is better to make the fix in this PR because Travis is slow?

Done. I simply removed the two lines that were needed earlier because `bar_format` threw an error on older tqdm versions. Just removing that fixes the issue.

neerajprad:
> LGTM, confirmed that the change fixes the issue in my system. Probably you will want to fix the f string issue at this build.

I think @fritzo is working on a fix for that in pyro-api. I'll bump up the pyro-api version once that is released.

neerajprad:
Good to merge.

PR #2093, opened 2019-10-24
Implement 1d convolution via fft

fritzo:
This adds a helper to do long time convolution, as common in signal processing.
The interface matches [np.convolve](https://docs.scipy.org/doc/numpy/reference/generated/numpy.convolve.html) but supports batching and gradients.

## Tested
- added unit tests

### DIFF
--- a/pyro/ops/tensor_utils.py
+++ b/pyro/ops/tensor_utils.py
@@ -53,7 +53,7 @@ def convolve(signal, kernel, mode='full'):
         raise ValueError('Unknown mode: {}'.format(mode))
 
     # Compute convolution using fft.
-    padded_size = m + n
+    padded_size = m + n - 1
     # Round up to next power of 2 for cheaper fft.
     padded_size = 2 ** int(math.ceil(math.log2(padded_size)))
     f_signal = torch.rfft(torch.nn.functional.pad(signal, (0, padded_size - m)), 1)
--- a/pyro/ops/tensor_utils.py
+++ b/pyro/ops/tensor_utils.py
@@ -1,3 +1,5 @@
+import math
+
 import torch
 
 
@@ -23,7 +25,7 @@ def _complex_mul(a, b):
     return torch.stack([ar * br - ai * bi, ar * bi + ai * br], dim=-1)
 
 
-def conv1d_fft(signal, kernel):
+def convolve(signal, kernel, mode='full'):
     """
     Computes the 1-d convolution of signal by kernel using FFTs.
     The two arguments should have the same rightmost dim, but may otherwise be
@@ -31,12 +33,32 @@ def conv1d_fft(signal, kernel):
 
     :param torch.Tensor signal: A signal to convolve.
     :param torch.Tensor kernel: A convolution kernel.
-    :return: A tensor of shape
-        ``broadcast_shape(convolve, kernel.shape)``
+    :param str mode: One of: 'full', 'valid', 'same'.
+    :return: A tensor with broadcasted shape. Letting ``m = signal.size(-1)``
+        and ``n = kernel.size(-1)``, the rightmost size of the result will be:
+        ``m + n - 1`` if mode is 'full';
+        ``max(m, n) - min(m, n) + 1`` if mode is 'valid'; or
+        ``m`` if mode is 'same'.
     :rtype torch.Tensor:
     """
-    assert signal.size(-1) == kernel.size(-1)
-    f_signal = torch.rfft(torch.cat([signal, torch.zeros_like(signal)], dim=-1), 1)
-    f_kernel = torch.rfft(torch.cat([kernel, torch.zeros_like(kernel)], dim=-1), 1)
+    m = signal.size(-1)
+    n = kernel.size(-1)
+    if mode == 'full':
+        truncate = m + n - 1
+    elif mode == 'valid':
+        truncate = max(m, n) - min(m, n) + 1
+    elif mode == 'same':
+        truncate = m
+    else:
+        raise ValueError('Unknown mode: {}'.format(mode))
+
+    # Compute convolution using fft.
+    padded_size = m + n
+    # Round up to next power of 2 for cheaper fft.
+    padded_size = 2 ** int(math.ceil(math.log2(padded_size)))
+    f_signal = torch.rfft(torch.nn.functional.pad(signal, (0, padded_size - m)), 1)
+    f_kernel = torch.rfft(torch.nn.functional.pad(kernel, (0, padded_size - n)), 1)
     f_result = _complex_mul(f_signal, f_kernel)
-    return torch.irfft(f_result, 1)[..., :signal.size(-1)]
+    result = torch.irfft(f_result, 1)
+
+    return result[..., :truncate]
--- a/pyro/ops/tensor_utils.py
+++ b/pyro/ops/tensor_utils.py
@@ -53,12 +53,11 @@ def convolve(signal, kernel, mode='full'):
     # Compute convolution using fft.
     padded_size = m + n - 1
     # Round up to next power of 2 for cheaper fft.
-    fast_ftt_size = 2 ** int(math.ceil(math.log2(padded_size)))
+    fast_ftt_size = 2 ** math.ceil(math.log2(padded_size))
     f_signal = torch.rfft(torch.nn.functional.pad(signal, (0, fast_ftt_size - m)), 1, onesided=False)
     f_kernel = torch.rfft(torch.nn.functional.pad(kernel, (0, fast_ftt_size - n)), 1, onesided=False)
     f_result = _complex_mul(f_signal, f_kernel)
     result = torch.irfft(f_result, 1, onesided=False)
 
-    result = result[..., :padded_size]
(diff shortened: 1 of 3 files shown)


fritzo:
@fehiepsi could you help me out on this PR? I need this helper for a model I'm working on and it's too late at night for me to figure out what's wrong. For reference see [wikipedia](https://en.wikipedia.org/wiki/Convolution_theorem) and the implementation in https://github.com/pytorch/pytorch/issues/21462 (more than I need here).

pyro/ops/tensor_utils.py
```diff
@@ -13,3 +13,30 @@ def block_diag(m):
     B, M, N = m.shape
     eye = torch.eye(B, dtype=m.dtype, device=m.device).reshape(B, 1, B, 1)
     return (m.unsqueeze(-2) * eye).reshape(B * M, B * N)
+
+
+def _complex_mul(a, b):
+    ar, ai = a.unbind(-1)
+    br, bi = b.unbind(-1)
+    # FIXME should b be conjugated?
+    # bi = -bi  # Conjugates b.
```
fehiepsi: I think that no conjugate here.

pyro/ops/tensor_utils.py
```diff
@@ -13,3 +13,30 @@ def block_diag(m):
...
+    arbitrarily broadcastable.
+
+    :param torch.Tensor signal: A signal to convolve.
+    :param torch.Tensor kernel: A convolution kernel.
+    :return: A tensor of shape
+        ``broadcast_shape(convolve, kernel.shape)``
+    :rtype torch.Tensor:
+    """
+    assert signal.size(-1) == kernel.size(-1)
+    f_signal = torch.rfft(torch.cat([signal, torch.zeros_like(signal)], dim=-1), 1)
```
fehiepsi: This is relevant to [autocorrelation](https://github.com/pyro-ppl/pyro/blob/dev/pyro/ops/stats.py#L95). I think [next_fast_len](https://docs.scipy.org/doc/scipy/reference/generated/scipy.fftpack.next_fast_len.html) might be helpful for performance. Let me verify tomorrow if we need `onesided` arg  in `rfft` and `irfft`.

pyro/ops/tensor_utils.py
```diff
@@ -13,3 +13,30 @@ def block_diag(m):
...
+    f_kernel = torch.rfft(torch.cat([kernel, torch.zeros_like(kernel)], dim=-1), 1)
+    f_result = _complex_mul(f_signal, f_kernel)
+    return torch.irfft(f_result, 1)[..., :signal.size(-1)]
```
fehiepsi: I guess you need the output to have the same last dimension as the signal. Probably you will want to take the center of the f_result?

tests/ops/test_tensor_utils.py
```diff
@@ -22,3 +22,18 @@ def test_block_diag(batch_size, block_size):
...
+
+
+@pytest.mark.parametrize('T', [2, 3, 4, 5, 10])
+@pytest.mark.parametrize('batch_shape', [(), (4,), (2, 3)], ids=str)
+def test_conv1d_fft(batch_shape, T):
+    signal = torch.randn(T)  # not batched
+    kernel = torch.randn(*batch_shape, T).exp()  # batched
+    actual = conv1d_fft(signal, kernel)
+
+    # FIXME is this correct?
```
fehiepsi: I think it is better to compare against https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.fftconvolve.html#scipy.signal.fftconvolve.

pyro/ops/tensor_utils.py
fehiepsi: I see, so I think you don't need to take conjugate. :)

tests/ops/test_tensor_utils.py
fritzo: The scipy.signal.convolve should agree with scipy.signal.fftconvolve.

pyro/ops/tensor_utils.py
```diff
@@ -13,3 +15,50 @@ def block_diag(m):
...
+        truncate = m + n - 1
+    elif mode == 'valid':
+        truncate = max(m, n) - min(m, n) + 1
+    elif mode == 'same':
+        truncate = m
+    else:
+        raise ValueError('Unknown mode: {}'.format(mode))
+
+    # Compute convolution using fft.
+    padded_size = m + n
```
fehiepsi: nit: I think `padded_size` would be `m + n - 1` but adding more zeros might have no harm.

pyro/ops/tensor_utils.py
fritzo: done, I've updated padding logic.

pyro/ops/tensor_utils.py
```diff
@@ -13,3 +15,50 @@ def block_diag(m):
...
+    :rtype torch.Tensor:
+    """
+    m = signal.size(-1)
+    n = kernel.size(-1)
+    if mode == 'full':
+        truncate = m + n - 1
+    elif mode == 'valid':
+        truncate = max(m, n) - min(m, n) + 1
+    elif mode == 'same':
+        truncate = m
```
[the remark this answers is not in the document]
fritzo: I thing it's best to match scipy.signal.convolve, which defines 'same' as size of the first arg.

pyro/ops/tensor_utils.py
```diff
@@ -13,3 +15,50 @@ def block_diag(m):
...
+    # Compute convolution using fft.
+    padded_size = m + n
+    # Round up to next power of 2 for cheaper fft.
+    padded_size = 2 ** int(math.ceil(math.log2(padded_size)))
+    f_signal = torch.rfft(torch.nn.functional.pad(signal, (0, padded_size - m)), 1)
+    f_kernel = torch.rfft(torch.nn.functional.pad(kernel, (0, padded_size - n)), 1)
+    f_result = _complex_mul(f_signal, f_kernel)
+    result = torch.irfft(f_result, 1)
+
+    return result[..., :truncate]
```
fehiepsi: I think it is better (here I use `padded_size` as m + n - 1, rather than fast_len size) to do
```
result = result[..., :padded_size]
start_idx = (padded_size - truncate) // 2
return result[..., start_idx: start_idx + truncate]

File: pyro/ops/tensor_utils.py (hunk 3 above)
Comment (fritzo) [reply pyro/ops/tensor_utils.py]: The updated truncation logic now matches scipy.signal.convolve in size.

File: tests/ops/test_tensor_utils.py (hunk 4 above)
Comment (fritzo) [reply tests/ops/test_tensor_utils.py]: Actually, np.convolve seems like an even better gold standard.

File: pyro/ops/tensor_utils.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (fritzo) [reply pyro/ops/tensor_utils.py]: Oh I see [np.convolve](https://docs.scipy.org/doc/numpy/reference/generated/numpy.convolve.html) follows `truncate = max(m,n)`, so I agree with you. I'll change.

File: pyro/ops/tensor_utils.py [hunk 8]
```diff
@@ -13,3 +15,48 @@ def block_diag(m):
...
+        truncate = max(m, n)
+    else:
+        raise ValueError('Unknown mode: {}'.format(mode))
+
+    # Compute convolution using fft.
+    padded_size = m + n - 1
+    # Round up to next power of 2 for cheaper fft.
+    padded_size = 2 ** int(math.ceil(math.log2(padded_size)))
+    f_signal = torch.rfft(torch.nn.functional.pad(signal, (0, padded_size - m)), 1)
+    f_kernel = torch.rfft(torch.nn.functional.pad(kernel, (0, padded_size - n)), 1)
```
fehiepsi: ditto: `onesided=False`

pyro/ops/tensor_utils.py
```diff
@@ -13,3 +15,48 @@ def block_diag(m):
...
+    f_result = _complex_mul(f_signal, f_kernel)
+    result = torch.irfft(f_result, 1)
```
fehiepsi: ditto: `onesided=False`

pyro/ops/tensor_utils.py
```diff
@@ -13,3 +15,48 @@ def block_diag(m):
...
+        truncate = max(m, n) - min(m, n) + 1
+    elif mode == 'same':
+        truncate = max(m, n)
+    else:
+        raise ValueError('Unknown mode: {}'.format(mode))
+
+    # Compute convolution using fft.
+    padded_size = m + n - 1
+    # Round up to next power of 2 for cheaper fft.
+    padded_size = 2 ** int(math.ceil(math.log2(padded_size)))
```
fehiepsi: nit: no need for `int`.

fritzo:
Thanks for all the help @fehiepsi !

ctr26:
Any thoughts on this being implemented directly into pyTorch?

fritzo:
@ctr26 you're welcome to move this upstream to PyTorch. I believe they would welcome any contribution that closely matches the NumPy interface. When we're busy we often simply implement things in Pyro and move on.

PR #2097, opened 2019-10-25
add parallel scan repeated matmul tensor utility

martinjankowiak:


fritzo:
This is not parallel scan, it is merely optimal binary reduction. Can you instead implement standard generic optimal reduction, something like this or find it on some job application site on the web:
```
def reduce_power(op, value, power):
    assert isinstance(power, int) and power >= 0
    if power == 0:
        raise NotImpleementedError
    bits = []
    while power:
        if power & 1:
            bits.append(power)
        power <<= 1
    result = value
    for bit in reversed(bits[:-1]):
        result = op(result, result)
        if bit:
            result = op(result, value)
    return result
```

martinjankowiak:
i've refactored this for simplicity. please lmk if the name of the method or something else can be improved

pyro/ops/tensor_utils.py
```diff
@@ -61,3 +61,27 @@ def convolve(signal, kernel, mode='full'):
 
     start_idx = (padded_size - truncate) // 2
     return result[..., start_idx: start_idx + truncate]
+
+
+def parallel_scan_repeated_matmul(M, n):
```
fritzo: Can you rename this to just `repeated_matmul`? the "parallel scan" part is an implementation detail.

pyro/ops/tensor_utils.py
```diff
@@ -61,3 +61,27 @@ def convolve(signal, kernel, mode='full'):
...
+
+    doubling_rounds = 0 if n <= 2 else math.ceil(math.log(n, 2)) - 1
+
+    result = torch.stack([M, torch.matmul(M, M)])
+
+    for i in range(doubling_rounds):
+        doubled = torch.matmul(result[-1, ...].unsqueeze(0), result)
+        result = torch.stack([result, doubled]).reshape(-1, *result.shape[1:])
+
+    return result[0:n, ...]
```
fritzo: nit: this just means `result[:n]`

pyro/ops/tensor_utils.py
```diff
@@ -61,3 +61,27 @@ def convolve(signal, kernel, mode='full'):
...
+    """
+    assert M.size(-1) == M.size(-2), "Input tensors must satisfy M.size(-1) == M.size(-2)."
+    assert n > 0, "argument n to parallel_scan_repeated_matmul must be 1 or larger"
+
+    doubling_rounds = 0 if n <= 2 else math.ceil(math.log(n, 2)) - 1
+
+    result = torch.stack([M, torch.matmul(M, M)])
+
+    for i in range(doubling_rounds):
+        doubled = torch.matmul(result[-1, ...].unsqueeze(0), result)
```
fritzo: nit: just `torch.matmul(result[-1], result)` should work fine

martinjankowiak:
addressed comments

PR #2098, opened 2019-10-25
Implement ReflectedDistribution

fritzo:
This implements a `ReflectedDistribution` which is like `TransformedDistribution(-, AbsTransform())` but also supports `.log_prob()`. Unlike truncating or censoring, reflecting leads to generic tractable `.rasample()` and `.log_prob()` methods. Unlike the [similar implementation](https://github.com/tensorflow/probability/blob/r0.8/tensorflow_probability/python/distributions/transformed_distribution.py#L407) in tensorflow distributions, I have not attempted to generalize Bijectors to multivalued functions, since this one case gets us quite far.

The motivation is in count-valued time series forecasting, where @slaweku recommends modeling positive data as power-transformed with a robust noise distribution. We'd like to use something like
```py
TransformedDistribution(StudentT(...),
                        PowerTransform(...))
```
which works in Stan, but fails in PyTorch because `PowerTransform` assumes positive data. In the past I've tried to replace the `PowerTransform` with an `ExpTransform`, but this ends up being less robust to zeros. Instead we can wrap with `ReflectedDistribution` to get a working Pyro version.
```py
TransformedDistribution(ReflectedDistribution(StudentT(...)),
                        PowerTransform(...))
```
I have tested this version for both SVI training and data generation.

## Tested
- added unit tests
- tested in a real time series application

### DIFF
--- a/pyro/distributions/reflected.py
+++ b/pyro/distributions/reflected.py
@@ -24,6 +24,8 @@ class ReflectedDistribution(TransformedDistribution):
         return super().expand(batch_shape, _instance=new)
 
     def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
         dim = max(len(self.batch_shape), value.dim())
         plus_minus = value.new_tensor([1., -1.]).reshape((2,) + (1,) * dim)
         return self.base_dist.log_prob(plus_minus * value).logsumexp(0)
--- a/pyro/distributions/reflected.py
+++ b/pyro/distributions/reflected.py
@@ -1,0 +1,29 @@
+from torch.distributions import constraints
+from torch.distributions.transforms import AbsTransform
+
+from pyro.distributions.torch import TransformedDistribution
+
+
+class ReflectedDistribution(TransformedDistribution):
+    """
+    Equivalent to ``TransformedDistribution(base_dist, AbsTransform())``,
+    but additionally supports :meth:`log_prob` .
+
+    :param ~torch.distributions.Distribution base_dist: The distribution to
+        reflect.
+    """
+    support = constraints.positive
+
+    def __init__(self, base_dist):
+        if base_dist.event_shape:
+            raise ValueError("Only univariate distributions can be reflected.")
+        super().__init__(base_dist, AbsTransform())
+
+    def expand(self, batch_shape, _instance=None):
+        new = self._get_checked_instance(type(self), _instance)
+        return super().expand(batch_shape, _instance=new)
+
+    def log_prob(self, value):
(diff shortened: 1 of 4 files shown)


fritzo:
cc @alicanb should we additionally move this upstream to torch.distributions?

neerajprad:
LGTM. Regarding:

> but fails in PyTorch because PowerTransform assumes positive data. 

Do you know why we make that assumption in PyTorch? It looks to me that it should also work for negative data, but it states that the domain / codomains should both be positive.

fritzo:
> Do you know why we make [the positivity] assumption in PyTorch?

Because non-integer powers of negative numbers are not real. This seems like a fine assumption.

Note Stan avoids this issue by allowing implicitly truncated distributions (rejecting samples out of the domain of the funtion). I find it cleaner to explicitly either truncate, censor, or reflect.

neerajprad:
> Because non-integer powers of negative numbers are not real. This seems like a fine assumption.

😄Of course, makes sense! 

> Note Stan avoids this issue by allowing implicitly truncated distributions (rejecting samples out of the range of the transform). I find it cleaner to explicitly either truncate, censor, or reflect; that way models are normalized.

Sounds good. Explicitly rejecting samples would be a quite a significant deviation from the design anyways.

pyro/distributions/reflected.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self, base_dist, validate_args=None):
+        if base_dist.event_shape:
+            raise ValueError("Only univariate distributions can be reflected.")
+        super().__init__(base_dist, AbsTransform(), validate_args)
+
+    def expand(self, batch_shape, _instance=None):
+        new = self._get_checked_instance(type(self), _instance)
+        return super().expand(batch_shape, _instance=new)
+
+    def log_prob(self, value):
```
neerajprad: I think we'll also need to add a check here:

```
if self._validate_args:
    self._validate_sample(value)
```

alicanb:
> cc @alicanb should we additionally move this upstream to torch.distributions?

Hey @fritzo , sorry I am late replying. This is cool, but maybe it would create confusion in the namespace for the user. Do you think there's a way to implement this in `TransformedDistributions`

fritzo:
@alicanb the only way to implement this as a `TransformedDistribution` would be to follow the route of tfp.distributions and add logic to sum over the branches of multi-functions. I'd prefer to avoid that complexity if possible, since this one `ReflectedDistribution` covers most use cases. I guess we'll just keep this in Pyro for now.

Issue #2090, opened 2019-10-23
Add a contrib.timeseries module


martinjankowiak:
- [x] matern kernels as state space models #2077 
- [x] basic univariate GP model and base class for `TimeSeriesModel`s #2088
- [x] add example usage for a contrib.timeseries model #2092 
- [x] add `LinearlyCoupledMaternGP' #2094 
- [x] add 'GenericLGSSM` #2099 
- [x] add 'GenericLGSSMWithGPNoiseModel' #2100 
- [x] add `DependentMaternGP` #2107 
- [x] decorate with `torch.no_grad()` where appropriate #2106 
- [x] add support for nu=0.5 #2106
- [x] add `LinearlyCoupledDependentMaternGP` #2112 
- [x] refactor to use `PyroModule` #2142 
- ~~add support for observation (noise) models~~
- ~~do an optimization pass (use `scale_tril` where appropriate, etc.)~~
- ~~make use of recently added constraints technology as in #2104~~

PR #2099, opened 2019-10-26
add GenericLGSSM to contrib.timeseries

martinjankowiak:
addresses #2090


pyro/contrib/timeseries/lgssm.py
```diff
@@ -0,0 +1,118 @@
...
+        return predicted_mean, predicted_covar
+
+    def forecast(self, targets, N_timesteps):
+        """
+        :param torch.Tensor targets: A 2-dimensional tensor of real-valued targets
+            of shape `(T, obs_dim)`, where `T` is the length of the time series and `obs_dim`
+            is the dimension of the real-valued targets at each time step. These
+            represent the training data that are conditioned on for the purpose of making
+            forecasts.
+        :param int N_timesteps: The number of timesteps to forecast into the future from
```
martinjankowiak: changed the API here as compared to the (continuous-time) GP models. what is the best way to do this? just expand the docstring in the base class?

pyro/contrib/timeseries/lgssm.py
```diff
@@ -0,0 +1,118 @@
...
+                                        N_trans_obs))  # N O O
+
+        # next compute the contribution from process noise that is injected at each timestep.
+        # (we need to do a cumulative sum to integrate across time)
+        process_covar = self._get_trans_dist().covariance_matrix
+        N_trans_obs_shift = torch.cat([self.obs_matrix.unsqueeze(0), N_trans_obs[0:-1]])
+        predicted_covar2 = torch.matmul(N_trans_obs_shift.transpose(-1, -2),
+                                        torch.matmul(process_covar, N_trans_obs_shift))  # N O O
+
+        predicted_covar = predicted_covar1 + torch.cumsum(predicted_covar2, dim=0)
```
martinjankowiak: computing `predicted_covar2` is a bit tricky. the monotonicity test suggests this is correct but one could imagine adding more tests?

martinjankowiak: in particular we get terms like AAAQAAA + AAQAA + AQA + Q

fritzo: How about the following test with a single datapoint and two different HMMs with dt different by a factor of 2. You should be able to get the predictions of one to match the even predictions of the other, by carefully setting parameters.

martinjankowiak: i can try something like that...

pyro/contrib/timeseries/lgssm.py
```diff
@@ -0,0 +1,118 @@
...
+
+        super(GenericLGSSM, self).__init__()
+
+        self.log_obs_noise_scale = nn.Parameter(log_obs_noise_scale_init)
+        self.log_trans_noise_scale_sq = nn.Parameter(torch.zeros(state_dim))
+        self.trans_matrix = nn.Parameter(torch.eye(state_dim) + 0.03 * torch.randn(state_dim, state_dim))
+        self.obs_matrix = nn.Parameter(0.3 * torch.randn(state_dim, obs_dim))
+        self.log_init_noise_scale_sq = nn.Parameter(torch.zeros(state_dim))
+
+    def _get_obs_noise_scale(self):
```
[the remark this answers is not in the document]
martinjankowiak: well, i'm actually going to generalize all these models to make use of a shared `ObservationModel` that supports diagonal/full-rank/rank-k covariance matrices. so i'd prefer to do that in the follow-up PR

pyro/contrib/timeseries/lgssm.py
```diff
@@ -0,0 +1,118 @@
...
+        return self.log_obs_noise_scale.exp()
+
+    def _get_init_dist(self):
+        loc = self.obs_matrix.new_zeros(self.state_dim)
+        eye = torch.eye(self.state_dim, device=self.obs_matrix.device, dtype=self.obs_matrix.dtype)
+        return MultivariateNormal(loc, self.log_init_noise_scale_sq.exp() * eye)
+
+    def _get_obs_dist(self):
+        loc = self.obs_matrix.new_zeros(self.obs_dim)
```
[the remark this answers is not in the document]
martinjankowiak: ok

pyro/contrib/timeseries/lgssm.py
```diff
@@ -0,0 +1,118 @@
...
+        :returns torch.Tensor: A (scalar) log probability.
+        """
+        assert targets.dim() == 2 and targets.size(-1) == self.obs_dim
+        return self._get_dist().log_prob(targets)
+
+    def _filter(self, targets):
+        """
+        Return the filtering state for the associated state space model.
+        """
+        assert targets.dim() == 2 and targets.size(-1) == self.obs_dim
```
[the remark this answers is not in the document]
martinjankowiak: supporting log prob with `targets.dim()>2` is easy enough but it'll be more work if the forecasting supports it....

pyro/contrib/timeseries/lgssm.py
martinjankowiak: wasn't really sure how to do this but added a `test_generic_lgssm_forecast` test that unrolls the tricky part of the forecast computation for 2 timesteps and verifies correctness

tests/contrib/timeseries/test_lgssm.py
```diff
@@ -0,0 +1,48 @@
...
+
+from tests.common import assert_equal
+from pyro.contrib.timeseries import GenericLGSSM
+import pytest
+
+
+@pytest.mark.parametrize('state_dim', [2, 3])
+@pytest.mark.parametrize('obs_dim', [2, 4])
+@pytest.mark.parametrize('T', [11, 17])
+def test_generic_lgssm_forecast(state_dim, obs_dim, T):
```
martinjankowiak: explicitly calculate terms like AQA+ Q

PR #2101, opened 2019-10-28
add block_diagonal tensor utility and rename block_diag => block_diag_embed

martinjankowiak:


martinjankowiak:
thanks to @fritzo for the implementation : )

Issue #2078, opened 2019-10-22
Make the ParamStore an nn.Module?


fritzo:
Three recent issues #2058, #2054, #1839 suggest that serialization might be easier if the ParamStore were an nn.Module, e.g. so it could be torch::jit::saved and loaded.

### Questions 2019-10-22
- Would this make it easier to jit-and-serve Pyro models from C++?
- Could we replace `param_store.load()` / `.save()` with `torch.load()`/`.save()`?
- Are there conflicts in the module-vs-dictlike interfaces?
- For example, can we `.clear()` a module?

### Update 2019-10-29

Now that #2102 is merged, it appears a possible strategy may be to recommend against the `pyro.param()` statement, and instead rely entirely on `pyro.module`. By following this strategy, users can avoid serialization issues with Pyro's param store, and rely entirely on native PyTorch `nn.Module` serialization.

- Is the "avoid `pyro.param`" idiom compatible with NumPyro?

## Tasks
- [x] provide autoguide methods to init autoguide parameters @fritzo 
- [x] refactor autoguides to be `nn.Module`s @neerajprad

neerajprad:
> Now that #2102 is merged, it appears a possible strategy may be to recommend against the pyro.param() statement, and instead rely entirely on pyro.module.

For autoguides, I think this solution may be a bit clunky since we'll need to write a separate wrapper that sets all the autoguide parameters like `init_loc` as nn.Parameter attributes on a wrapper class. Maybe the `ParamStoreModule` solution will work better in that case?

fritzo:
> For autoguides

I was just facing that issues :smile: I am also looking into making the autoguides `nn.Modules`.

fehiepsi:
Though a bit different, the [Parameterized](https://github.com/pyro-ppl/pyro/blob/dev/pyro/contrib/gp/parameterized.py#L21) class has some logic to handle parameters and buffers, which can be helpful for you guys. :)

Let's see how this evolves, so I might be able to make Parameterized a subclass of ParamStoreModule.

fritzo:
@neerajprad What do you think of a mixed autoguide solution?
1. Use `@constrained` in the static autoguides 
```diff
- class AutoContinuous(AutoGuide):
+ class AutoContinuous(AutoGuide, nn.Module):
  ...
  class AutoDiagonalNormal(AutoContinuous):
+ @constrained
+ def scale(self):
+     return constraints.positive
  ...
```
[...]

PR #2102, opened 2019-10-29
Implement constraints for nn.Parameters of nn.Modules

fritzo:
Addresses #2078 

This implements a surprisingly succinct syntax for constraints of module parameters, e.g.
```py
from torch import nn
from pyro.params import constraint

class Normal(nn.Module):
    def __init__(self, loc, scale):
        super().__init__()
        self.loc = loc                 # this will be unconstrained
        self.scale = scale             # this will be constrained

    @constraint
    def scale(self):                   # must match name of attribute
        return constraints.positive    # a constraint object,
                                       # can even depend on other params
```
Hopefully this moves us towards making the Pyro param store a nn.Module. Currently this is not used by the ParamStore, but this PR will unblock later refactoring.

In the nearer term, @martinjankowiak might want to use this for constrained parameters in pyro.contrib.timeseries, e.g. this could simplify #2099.

## How does this work?

For each constrained parameter `x`, this creates an unconstrained parameter with name `x_unconstrained`, which is accessible in the usual way. The name `x` then points to a Python descriptor (like a property) that transforms to/from `unconstrained_x` to a user-defined value. Thus all the usual `nn.Module` magic should continue to work correctly, with only the addition of this `transform_to()` layer.

## Tested
- [x] tested with a variety of constraints
- [x] tested chaining, where one param's constraint might depend on another param (e.g. for lower bound)
- [x] tested serialization

### DIFF
--- a/pyro/params/constrained_parameter.py
+++ b/pyro/params/constrained_parameter.py
@@ -1,0 +1,106 @@
+import weakref
+
+import torch
+from torch.distributions import transform_to
+from torch.distributions.constraints import Constraint
+
+
+class ConstrainedParameter:
+    """
+    Constrained wrapper around a :class:`~torch.nn.Parameter` that obeys a
+    :class:`~torch.distributions.constraints.Constraint` .
+
+    Like :class:`~torch.nn.Parameter` , this can be accessed directly as an
+    attribute of an enclosing :class:`~torch.nn.Module` . Unlike a
+    :class:`~torch.nn.Parameter` , the ``.data`` attribute cannot be set
+    directly; instead set data via the :meth:`unconstrained` method::
+
+        my_module.scale = ConstrainedParameter(torch.ones(2,3,4),
+                                               constraint=constraints.positive)
+        assert isinstance(my_module.scale, torch.Tensor)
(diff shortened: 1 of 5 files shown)


martinjankowiak:
to clarify, if such an object is an attribute of some `module` this will support methods like `module.cuda()` right?

fritzo:
Yes, this will support methods like `module.cuda()`.

pyro/params/constrained_parameter.py
```diff
@@ -0,0 +1,84 @@
+import torch
+from torch.distributions import transform_to
+
+
+class ConstrainedParameter:
+    """
+    Descriptor to add a constraint to a
```
neerajprad: nit: additional `to a` above.

fritzo:
> will this require us to start storing parameters as attributes of ParamStore class?

Good question. I think that, independently of this PR, if we make the ParamStore an nn.Module, then yes we would need to store pyro parameters as attributes of the ParamStore. While that would be easy to do for unconstrained parameters, I do not know how to dynamically set constraints (this PR only allows static constraints). We should think more on this.

PR #2103, opened 2019-10-29
Sketch ParamStoreModule object

fritzo:
Addresses #2078 

This sketches a possible `ParamStoreModule` that might eventually replace our existing `ParamStoreDict`. This would likely incur major breaking changes (e.g. restriction of param names to valid python identifiers), so we might want to delay this refactoring until 2.0 or gate it by an environment variable (e.g. `pyro.configure(use_param_store_module=True)`).

## Tests
- [ ] test inference
- [ ] test serialization

### DIFF
--- a/pyro/params/constrained_parameter.py
+++ b/pyro/params/constrained_parameter.py
@@ -1,3 +1,5 @@
+import weakref
+
 import torch
 from torch.distributions import transform_to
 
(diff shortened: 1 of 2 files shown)


fritzo:
closing in favor of #2102 and #2104 combined with `pyro.module()`

PR #2104, opened 2019-10-29
Implement a ConstrainedModule for dynamic constraints

fritzo:
Addresses #2078 

This implements a `ConstrainedModule` subclass of `torch.nn.Module`. Whereas the `@constraint` decorator of #2102 which stores constraints statically in the module definition, this version stores constraints dynamically in the module instance. (the two have different advantages).

This is one step in making the `ParamStore` and dynamic autoguides like `AutoDelta` and `AutoNormal` into `nn.Module`s (static autoguides like `AutoDiagonalNormal` could simply use `@constraint`).

## Tested
- [x] unit test for `__getattr__`, `__setattr__`, `__delattr__`
- [x] unit test for `torch.save()`/`torch.load()`

### DIFF
--- a/docs/source/parameters.rst
+++ b/docs/source/parameters.rst
@@ -16,8 +16,8 @@ ParamStore
     :show-inheritance:
     :member-order: bysource
 
-ConstrainedParameter
---------------------
+Constraints
+-----------
 
 .. automodule:: pyro.params.constrained_parameter
     :members:
--- a/pyro/params/__init__.py
+++ b/pyro/params/__init__.py
@@ -1,7 +1,8 @@
-from .constrained_parameter import ConstrainedParameter, constraint
+from .constrained_parameter import ConstrainedModule, ConstrainedParameter, constraint
 from .param_store import module_from_param_with_module_name, param_with_module_name, user_param_name
 
 __all__ = [
+    "ConstrainedModule",
     "ConstrainedParameter",
     "constraint",
     "module_from_param_with_module_name",
--- a/pyro/params/constrained_parameter.py
+++ b/pyro/params/constrained_parameter.py
@@ -1,10 +1,101 @@
+from collections import OrderedDict, namedtuple
+
 import torch
 from torch.distributions import transform_to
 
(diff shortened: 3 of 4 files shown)


fritzo:
@fehiepsi indeed it looks like this is similar to your [Parametrized](https://github.com/pyro-ppl/pyro/blob/dev/pyro/contrib/gp/parameterized.py#L21) class. It appears the interfaces differ in that:
- this version tries to use `my_module.x = ConstrainedParameter(data, constraint)`, where as the other version uses `my_module.set_constraint()`.
- this version does't handle `._buffers` like the `Parametrized` version does.

Should we combine these? Maybe make `Parameterized` inherit from this, and use `pyro.module()` rather than `pyro.param()` in `pyro.contrib.gp`?

fehiepsi:
> this version tries to use my_module.x = ConstrainedParameter(data, constraint)

I like this better. In GP, I have to use two statements for this purpose
```
        my_module.variance = Parameter(variance)
        my_module.set_constraint("variance", constraints.positive)
```
and the fact that `variance` is no longer a `Parameter` after `set_constraint("variance"...)` can cause confusion.

> Should we combine these (in a follow-up PR)?

Definitely! It is also a good chance to test edge cases of the ConstrainedModule interface (many GP examples do not involve parameter's priors, so it is good to go). Let me try to do it.

The only missing feature is the ability to set priors for those parameters (and we can use AutoDelta and AutoNormal on them), but I think I can keep the current behavior of `Parameterized` for it.

> use pyro.module() rather than pyro.param()

GP models are trained using PyTorch optimizers directly on models' parameters, so the module does not need to use `pyro.param` or `pyro.module`.

fritzo:
> using nn.Module with @constraint decorator unless we need to set constraints dynamically ... ?

You are correct that `@constraint` incurs less overhead than `ConstrainedModule`. Another advantage of `@constraint` is that its constraints are computed dynamically from function calls, and hence constraints can depend on other (upstream) parameters. I suspect there are also portability issues with both versions, and we'll need to see which idiom supports easier serialization.

fritzo:
Thanks for reviewing!

PR #2107, opened 2019-11-01
add DependentMaternGP to contrib.timeseries

martinjankowiak:
addresses #2090

### DIFF
--- a/pyro/contrib/timeseries/gp.py
+++ b/pyro/contrib/timeseries/gp.py
@@ -314,7 +314,7 @@ class DependentMaternGP(TimeSeriesModel):
         self.kernel.register_buffer("log_kernel_scale", torch.zeros(obs_dim))
 
         self.log_obs_noise_scale = nn.Parameter(log_obs_noise_scale_init)
-        self.log_diag_wiener_noise = nn.Parameter(-2.0 * torch.ones(obs_dim))
+        self.log_diag_wiener_noise = nn.Parameter(torch.zeros(obs_dim))
         self.off_diag_wiener_noise = nn.Parameter(0.03 * torch.randn(obs_dim, obs_dim))
 
         obs_matrix = torch.zeros(self.full_state_dim, obs_dim)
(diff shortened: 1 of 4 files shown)


pyro/contrib/timeseries/gp.py
```diff
@@ -18,6 +20,12 @@ class IndependentMaternGP(TimeSeriesModel):
     :param float nu: The order of the Matern kernel; one of 0.5, 1.5 or 2.5.
     :param float dt: The time spacing between neighboring observations of the time series.
     :param int obs_dim: The dimension of the targets at each time step.
+    :param torch.Tensor log_length_scale_init: optional initial values for the kernel length scale
```
martinjankowiak: adding missing docstrings

pyro/contrib/timeseries/gp.py
```diff
@@ -30,7 +38,7 @@ def __init__(self, nu=1.5, dt=1.0, obs_dim=1,
             log_obs_noise_scale_init = -2.0 * torch.ones(obs_dim)
         assert log_obs_noise_scale_init.shape == (obs_dim,)
 
-        super(IndependentMaternGP, self).__init__()
+        super().__init__()
```
martinjankowiak: simplify super() calls

pyro/contrib/timeseries/gp.py
```diff
@@ -253,3 +267,159 @@ def forecast(self, targets, dts):
...
+
+        self.kernel = MaternKernel(nu=nu, num_gps=obs_dim,
+                                   log_length_scale_init=log_length_scale_init)
+        self.full_state_dim = self.kernel.state_dim * obs_dim
+
+        # we demote self.kernel.log_kernel_scale from being a nn.Parameter
+        # since the relevant scales are now encoded in the wiener noise matrix
+        del self.kernel.log_kernel_scale
+        self.kernel.register_buffer("log_kernel_scale", torch.zeros(obs_dim))
```
martinjankowiak: perhaps there's a better solution here?

fritzo: solution is good, but nit: might as well store `kernel_scale` if it is constant.

fritzo: do you want to fix this now, or should I merge?

martinjankowiak:
I'll try to address that later,  thanks

PR #2110, opened 2019-11-01
Support python bool in TorchDistribution.mask() w/ short circuit for log_prob

fritzo:
This implements a feature requested by @martinjankowiak and @gpleiss, allowing distribution `.log_prob()` values to be masked with zero computation cost, i.e. to avoid computation in `Trace.compute_log_prob()` or `Trace.compute_score_parts()`. For example the following are nearly* equivalent but the second is cheaper:
```py
# Version 1. triggers a call to .log_prob() inside Trace_ELBO()
with pyro.mask(False):
    pyro.sample("x", MultivariateNormal(loc, scale_tril=scale_tril))

# Version 2. short circuits
pyro.sample("x", MultivariateNormal(loc, scale_tril=scale_tril).mask(False))
```
*Since we don't have full access to `.dtype` and `.device` information for values that have not been computed, we implement best-effort results as follows:
1. Return `torch.zeros((), device=value.device)` from`.log_prob()`, which may sometimes return a tensor of incorrect precision (e.g. double vs float)
2. Return a Python `0.` from `kl_divergence` which may break some downstream logic, however we have no generic access to device, and this was the simplest device-agnostic solution I could find.

Note we chose to implement this short-circuit in `MaskedDistribution` rather than in `poutine.mask` or `Trace.compute_*()` because `MaskedDistribution` has fewer users and because this solution touches fewer pieces of tricky logic.

## Tested
- tested short circuit logic
- tests for mask types
- tests for kl_divergence types

### DIFF
--- a/pyro/distributions/torch_distribution.py
+++ b/pyro/distributions/torch_distribution.py
@@ -211,13 +211,13 @@ class MaskedDistribution(TorchDistribution):
     arg_constraints = {}
 
     def __init__(self, base_dist, mask):
+        self.base_dist = base_dist
         if isinstance(mask, bool):
             self._mask = mask
         else:
             if broadcast_shape(mask.shape, base_dist.batch_shape) != base_dist.batch_shape:
                 raise ValueError("Expected mask.shape to be broadcastable to base_dist.batch_shape, "
                                  "actual {} vs {}".format(mask.shape, base_dist.batch_shape))
-            self.base_dist = base_dist
             self._mask = mask.bool()
         super(MaskedDistribution, self).__init__(base_dist.batch_shape, base_dist.event_shape)
 
@@ -261,7 +261,7 @@ class MaskedDistribution(TorchDistribution):
 
     def score_parts(self, value):
         if isinstance(self._mask, bool):
-            return super().score_parts(self, value)  # calls self.log_prob()
+            return super().score_parts(value)  # calls self.log_prob(value)
         return self.base_dist.score_parts(value).scale_and_mask(mask=self._mask)
 
     def enumerate_support(self, expand=True):
@@ -290,7 +290,7 @@ def _kl_masked_masked(p, q):
         mask = p._mask & q._mask
 
     if mask is False:
-        return 0.  # We cannot generically determine device, so we return a float.
+        return 0.  # Return a float, since we cannot determine device.
     if mask is True:
         return kl_divergence(p.base_dist, q.base_dist)
     kl = kl_divergence(p.base_dist, q.base_dist)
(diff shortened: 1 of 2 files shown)


pyro/distributions/torch_distribution.py
```diff
@@ -200,26 +202,32 @@ class TorchDistribution(torch.distributions.Distribution, TorchDistributionMixin
 
 class MaskedDistribution(TorchDistribution):
     """
-    Masks a distribution by a zero-one tensor that is broadcastable to the
+    Masks a distribution by a boolean tensor that is broadcastable to the
     distribution's :attr:`~torch.distributions.distribution.Distribution.batch_shape`.
 
-    :param torch.Tensor mask: A zero-one valued tensor.
+    :param mask: A boolean or boolean-valued tensor.
+    :type mask: torch.Tensor or book
```
martinjankowiak: typo

pyro/distributions/torch_distribution.py
```diff
@@ -243,9 +255,17 @@ def rsample(self, sample_shape=torch.Size()):
         return self.base_dist.rsample(sample_shape)
 
     def log_prob(self, value):
+        if self._mask is False:
+            shape = broadcast_shape(self.base_dist.batch_shape,
+                                    value.shape[:value.dim() - self.event_dim])
+            return torch.zeros((), device=value.device).expand(shape)
```
neerajprad: You alluded to this in the description I think (but I didn't quite follow) - why can't we use the same `dtype` as `value` in the call to `torch.zeros`?

fritzo: Value could be long, but log_prob must be float of various precision.

neerajprad: I suppose we could still disambiguate float32/64 and use the default tensor type for long, but not sure if its worth it.

gpleiss:
Thanks for the feature! One tiny issue that I'm noticing: `.mask(False)` doesn't work in conjunction with `TraceMeanField_ELBO`. This is because the `kl_divergence` for the masked distributions is the float 0.0, which conflicts with this line:

https://github.com/pyro-ppl/pyro/blob/e5073c19e6ea7370c609771f222fb8bac8117745/pyro/infer/trace_mean_field_elbo.py#L101

Would you be open to changing that line to
```python
if torch.is_tensor(kl_qp):
    assert kl_qp.shape == guide_site["fn"].batch_shape
```

fritzo:
@gpleiss yes, thanks for finding the fix. We'll quickly review a PR if you send one, or we can make the change later today.

fritzo:
Maybe this would be a bit safer:
```py
if isinstance(kl_qp, torch.Tenosor):
    assert kl_qp.shape == guide_site["fn"].batch_shape
    kl_qp_sum = kl_qp.sum()
else:
    kl_qp_sum = kl_qp * torch.Size(guide_site["fn"].batch_shape).numel()
elbo_particle = elbo_particle - kl_qp_sum
```

gpleiss:
Cool I'll put up a PR with that.

Issue #2109, opened 2019-11-01
[feature request] [discussion] generalized gradient clipping


jamestwebber:
As @martinjankowiak said in the discussion of #1293, it'd be nice to have a generic way to clip gradients in Pyro. I _think_ I understand how to implement this and I'm interested in contributing the PR if I can get a little guidance. The architecture of Pyro is complex and it's not obvious to me where this feature should live.

The current way to clip gradients in PyTorch is with [`clip_grad_norm_`](https://pytorch.org/docs/stable/nn.html#clip-grad-norm) and [`clip_grad_value_`](https://pytorch.org/docs/stable/nn.html#clip-grad-value). This is different from how Pyro has implemented ClippedAdam, as the clipping happens outside of the optimizer step (see [this SO discussion](https://stackoverflow.com/questions/54716377/how-to-do-gradient-clipping-in-pytorch)):

```python
optimizer.zero_grad()
loss, hidden = model(data, hidden, targets)
loss.backward()

torch.nn.utils.clip_grad_norm_(model.parameters(), args.clip)
optimizer.step()
```

Because Pyro encapsulates those steps into e.g. the `SVI.step` method, it's not totally clear where to add this ability, but maybe the most natural place is in the `PyroOptim` class, allowing any wrapped optimizer to clip gradients if desired. It could be a global attribute of the optimizer:

```
def __call__(self, params,  *args, **kwargs):
        ...
        for p in params:
            ...

            if isinstance(self.optim_objs[p], torch.optim.lr_scheduler._LRScheduler) or \
                    isinstance(self.optim_objs[p], torch.optim.lr_scheduler.ReduceLROnPlateau):
                # if optim object was a scheduler, perform an optimizer step
                if self.clip_norm > 0:
                    clip_grad_norm_(self.optim_objs[p].optimizer.params(), self.clip_norm)
                self.optim_objs[p].optimizer.step(*args, **kwargs)
            else:
                if self.clip_norm > 0:
                    clip_grad_norm_(self.optim_objs[p].params(), self.clip_norm)
                self.optim_objs[p].step(*args, **kwargs)
```

This would entail adding `clip_norm` (or more generically `**kwargs`) to the PyroOptim init method and exposing it in e.g. `pytorch_optimizers.py` (I haven't pored over enough of the codebase to know everything that would be affected).

[...]

PR #2111, opened 2019-11-02
First pass at adding `clip_args` to PyroOptim

jamestwebber:
This is a WIP to spur discussion on #2109. This code runs but I didn't test it very well–I will use one of the existing tutorials I know will train quickly, so that I can verify the gradients are being clipped as expected.

In terms of the API I'm not in love with how I did this but I'm not sure what would be better. The main trade-off to me is a) adding an argument to the `PyroOptim` initializer versus b) adding keywords to the existing `optim_args` argument. a) feels a little clunky and requires more code changes (in `pytorch_optimizers.py` and `lr_scheduler.py`, maybe in other places I didn't catch). b) feels cleaner (and for people who don't care about this feature, is invisible) but might be confusing, as the user is passing keywords to the optimizer that don't really exist (I would pop them out in the `PyroOptim` class). 

Another question is what to do with the existing `ClippedAdam` optimizer. It's not exactly the same as "`Adam` + `clip_norm`" because it also has learning-rate decay built into it. That should be done by an `LRScheduler`, though. If the powers that be agree I will add a deprecation warning to `ClippedAdam` that tells people to use `Adam` with a clipping parameter and/or a scheduler.

### DIFF
--- a/pyro/optim/lr_scheduler.py
+++ b/pyro/optim/lr_scheduler.py
@@ -9,6 +9,8 @@ class PyroLRScheduler(PyroOptim):
     :param scheduler_constructor: a :class:`~torch.optim.lr_scheduler`
     :param optim_args: a dictionary of learning arguments for the optimizer or a callable that returns
         such dictionaries. must contain the key 'optimizer' with pytorch optimizer value
+    :param clip_args: a dictionary of clip_norm and/or clip_value args or a callable that returns
+        such dictionaries.
 
     Example::
 
--- a/pyro/optim/lr_scheduler.py
+++ b/pyro/optim/lr_scheduler.py
@@ -20,7 +20,7 @@ class PyroLRScheduler(PyroOptim):
                 svi.step(minibatch)
             scheduler.step(epoch=i)
     """
-    def __init__(self, scheduler_constructor, optim_args):
+    def __init__(self, scheduler_constructor, optim_args, clip_args=None):
         # pytorch scheduler
         self.pt_scheduler_constructor = scheduler_constructor
         # torch optimizer
@@ -28,7 +28,7 @@ class PyroLRScheduler(PyroOptim):
         # kwargs for the torch optimizer
         optim_kwargs = optim_args.pop('optim_args')
         self.kwargs = optim_args
-        super(PyroLRScheduler, self).__init__(pt_optim_constructor, optim_kwargs)
+        super(PyroLRScheduler, self).__init__(pt_optim_constructor, optim_kwargs, clip_args)
 
     def __call__(self, params, *args, **kwargs):
         super(PyroLRScheduler, self).__call__(params, *args, **kwargs)
--- a/pyro/optim/optim.py
+++ b/pyro/optim/optim.py
@@ -14,7 +14,7 @@ class PyroOptim(object):
     :param optim_constructor: a torch.optim.Optimizer
     :param optim_args: a dictionary of learning arguments for the optimizer or a callable that returns
         such dictionaries
-    :param clip_args: a dictionary of clip_norm and/or clip_value args or a callable that return
+    :param clip_args: a dictionary of clip_norm and/or clip_value args or a callable that returns
         such dictionaries
     """
     def __init__(self, optim_constructor, optim_args, clip_args=None):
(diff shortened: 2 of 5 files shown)


jamestwebber:
@martinjankowiak I will try to write one. It seems a little trickier than the existing `clipped_adam` test because I need to use a `PyroOptim` object, which means I can't just call `optim.backward()` with a known gradient. If there are any examples of the right way to do this that'd be helpful.

martinjankowiak:
i believe you could model your test on a pared down version of the many tests [here](https://github.com/pyro-ppl/pyro/blob/d03084358f6500b71cc037a1440bb9fe7bb99e65/tests/infer/test_gradient.py#L88)

jamestwebber:
I realized that I could call PyroOptim with the relevant parameter and it'd do the right thing, just pushed the following addition to `tests/optim/test_optim.py`:

```

@pytest.mark.parametrize('pyro_optim', [optim.Adam, optim.SGD])
@pytest.mark.parametrize('clip', ['clip_norm', 'clip_value'])
@pytest.mark.parametrize('value', [1., 3., 5.])
def test_clip_norm(pyro_optim, clip, value):
    x1 = torch.tensor(0., requires_grad=True)
    x2 = torch.tensor(0., requires_grad=True)
    opt_c = pyro_optim({"lr": 1.}, {clip: value})
    opt = pyro_optim({"lr": 1.})
    for step in range(3):
        x1.backward(Uniform(value, value + 3.).sample())
        x2.backward(torch.tensor(value))
        opt_c([x1])
        opt([x2])
        assert_equal(x1, x2)
        opt_c.optim_objs[x1].zero_grad()
        opt.optim_objs[x2].zero_grad()
```

The parameterization might be a little excessive, but I thought it'd be good to test multiple optimizers. I had to call `zero_grad` at the end of the loop because the optimizer doesn't exist before the call. The tests all passed for me, let me know if anything needs to be changed.

pyro/optim/lr_scheduler.py
```diff
@@ -20,15 +20,15 @@ class PyroLRScheduler(PyroOptim):
...
+    def __init__(self, scheduler_constructor, optim_args, clip_args=None):
         # pytorch scheduler
         self.pt_scheduler_constructor = scheduler_constructor
         # torch optimizer
         pt_optim_constructor = optim_args.pop('optimizer')
         # kwargs for the torch optimizer
         optim_kwargs = optim_args.pop('optim_args')
         self.kwargs = optim_args
-        super(PyroLRScheduler, self).__init__(pt_optim_constructor, optim_kwargs)
+        super(PyroLRScheduler, self).__init__(pt_optim_constructor, optim_kwargs, clip_args)
```
martinjankowiak: the docstring above needs to include `clip_args`

pyro/optim/optim.py
```diff
@@ -13,19 +14,30 @@ class PyroOptim(object):
     :param optim_constructor: a torch.optim.Optimizer
     :param optim_args: a dictionary of learning arguments for the optimizer or a callable that returns
         such dictionaries
+    :param clip_args: a dictionary of clip_norm and/or clip_value args or a callable that return
```
martinjankowiak: nit: typo 'returns'

pyro/optim/optim.py
```diff
@@ -52,8 +66,10 @@ def __call__(self, params,  *args, **kwargs):
             if isinstance(self.optim_objs[p], torch.optim.lr_scheduler._LRScheduler) or \
                     isinstance(self.optim_objs[p], torch.optim.lr_scheduler.ReduceLROnPlateau):
                 # if optim object was a scheduler, perform an optimizer step
+                self.grad_clip[p](p)
```
martinjankowiak: should we short circuit these calls if they are no-ops?

tests/optim/test_optim.py
```diff
@@ -131,6 +131,24 @@ def test_autowrap(factory):
...
+    x1 = torch.tensor(0., requires_grad=True)
+    x2 = torch.tensor(0., requires_grad=True)
+    opt_c = pyro_optim({"lr": 1.}, {clip: value})
+    opt = pyro_optim({"lr": 1.})
+    for step in range(3):
+        x1.backward(Uniform(value, value + 3.).sample())
+        x2.backward(torch.tensor(value))
+        opt_c([x1])
+        opt([x2])
+        assert_equal(x1, x2)
```
martinjankowiak: this assert is fine but can you please add explicit tests checking `x_i.grad`?

martinjankowiak:
ideally we would also have some example usage somewhere, e.g. here: http://pyro.ai/examples/custom_objectives.html

pyro/optim/optim.py
jamestwebber: I was trying to think of a good way to do that earlier and came up blank, but I will add one possible approach.

pyro/optim/optim.py
```diff
@@ -52,8 +66,12 @@ def __call__(self, params,  *args, **kwargs):
             if isinstance(self.optim_objs[p], torch.optim.lr_scheduler._LRScheduler) or \
                     isinstance(self.optim_objs[p], torch.optim.lr_scheduler.ReduceLROnPlateau):
                 # if optim object was a scheduler, perform an optimizer step
+                if self.grad_clip[p] is not None:
```
fritzo: nit: Can you move this outside of the `if isinstance(...optim...)` statement, since the two branches are identical?

jamestwebber: Oh duh good point! My original version needed to be inside the `if` statement but then later rewrite made it unnecessary

PR #2112, opened 2019-11-02
add LinearlyCoupledDependentMaternGP to contrib.timeseries

martinjankowiak:
addresses #2090

### DIFF
--- a/pyro/contrib/timeseries/gp.py
+++ b/pyro/contrib/timeseries/gp.py
@@ -321,6 +321,9 @@ class DependentMaternGP(TimeSeriesModel):
             obs_matrix[self.kernel.state_dim * i, i] = 1.0
         self.register_buffer("obs_matrix", obs_matrix)
 
+    def _get_obs_matrix(self):
+        return self.obs_matrix
+
     def _get_obs_noise_scale(self):
         return self.log_obs_noise_scale.exp()
 
@@ -367,7 +370,7 @@ class DependentMaternGP(TimeSeriesModel):
         """
         trans_matrix, trans_dist, stat_covar = self._trans_matrix_distribution_stat_covar(self.dt)
         return dist.GaussianHMM(self._get_init_dist(stat_covar), trans_matrix,
-                                trans_dist, self.obs_matrix, self._get_obs_dist())
+                                trans_dist, self._get_obs_matrix(), self._get_obs_dist())
 
     def log_prob(self, targets):
         """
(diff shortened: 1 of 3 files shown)


pyro/contrib/timeseries/gp.py
```diff
@@ -321,6 +321,9 @@ def __init__(self, nu=1.5, dt=1.0, obs_dim=1,
             obs_matrix[self.kernel.state_dim * i, i] = 1.0
         self.register_buffer("obs_matrix", obs_matrix)
 
+    def _get_obs_matrix(self):
```
martinjankowiak: refactor a bit to support straightforward subclassing below

pyro/contrib/timeseries/gp.py
```diff
@@ -422,3 +426,58 @@ def forecast(self, targets, dts):
...
+        del self.kernel.log_kernel_scale
+        self.kernel.register_buffer("log_kernel_scale", torch.zeros(obs_dim))
+
+        self.log_obs_noise_scale = nn.Parameter(log_obs_noise_scale_init)
+        self.log_diag_wiener_noise = nn.Parameter(torch.zeros(obs_dim))
+        self.off_diag_wiener_noise = nn.Parameter(0.03 * torch.randn(obs_dim, obs_dim))
+
+        self.obs_matrix = nn.Parameter(0.3 * torch.randn(self.obs_dim, self.obs_dim))
+
+    def _get_obs_matrix(self):
```
martinjankowiak: basically the only thing different about this model is that the obs_matrix is learned (instead of just being a particular matrix made of 0s and 1s)

pyro/contrib/timeseries/gp.py
```diff
@@ -422,3 +426,58 @@ def forecast(self, targets, dts):
...
         return MultivariateNormal(predicted_mean, predicted_covar)
+
+
+class LinearlyCoupledDependentMaternGP(DependentMaternGP):
+    """
+    A time series model in which each output dimension is modeled as a linear combination
+    of univariate Gaussian Processes with Matern kernels. The different output dimensions
+    are additionally correlated because the Gaussian Processes are driven by a correlated
+    Wiener process. That is, this time series model combines the structure of the
+    `LinearlyCoupledMaternGP` (with `num_gps` equal to `obs_dim`) with that of the
```
fritzo: nit: You should either use sphinx annotations for linkable objects
```
:class:`LinearlyCoupledMaternGP`
```
or double ticks to trigger code font
```
``num_gps`` equal to ``obs_dim``
```

pyro/contrib/timeseries/gp.py
fritzo: Thanks for explaining.

PR #2115, opened 2019-11-04
Rename ReflectedDistribution to FoldedDistribution

fritzo:
Amends #2098 to use the standard term in stats literature: "folded". See for example
https://en.wikipedia.org/wiki/Folded_normal_distribution
https://en.wikipedia.org/wiki/Folded-t_and_half-t_distributions

(sorry for the churn)


pyro/distributions/folded.py
```diff
@@ -16,7 +16,7 @@ class ReflectedDistribution(TransformedDistribution):
 
     def __init__(self, base_dist, validate_args=None):
         if base_dist.event_shape:
-            raise ValueError("Only univariate distributions can be reflected.")
+            raise ValueError("Only univariate distributions can be foldeed.")
```
martinjankowiak: typo

PR #2119, opened 2019-11-05
Add interface for custom autoguide initialization

fritzo:
Addresses #2120, #2078 

This refactors `AutoDelta` and some `AutoContinuous` classes to make it less hacky to provide custom initialization. Previously the recommended path was to either
1. call `pyro.param('auto_scale', torch.ones(100) * 0.01)` before initialization, or
2. call `pyro.param('auto_scale').unconstained().fill(-4)` after initialization;

After this PR you can instead pass
```py
guide = AutoDiagonalNormal(model, init_scale=0.01)
```

This also documents how to create custom `init_loc_fn()`s and uses custom `init_loc_fn()` in two examples.

## Tested
- [x] added tests for interface
- [x] updated notebooks and tutorials
- [x] ran sparse_regression.py locally and verified convergence
- [x] ran gmm.ipynb locally and verified convergence


examples/sparse_regression.py
```diff
@@ -278,11 +280,12 @@ def main(args):
     print("Expected singleton thetas:\n", expected_thetas.data.numpy())
 
     # we do the final computation using double precision
+    median = guide.median()  # == mode for MAP inference
```
fritzo: The next few lines are unrelated to initialization, but will help with #2078 .

pyro/infer/autoguide/guides.py
```diff
@@ -463,23 +467,24 @@ class AutoMultivariateNormal(AutoContinuous):
...
-        pyro.param("auto_loc", torch.randn(latent_dim))
-        pyro.param("auto_scale_tril", torch.tril(torch.rand(latent_dim)),
-                   constraint=constraints.lower_cholesky)
+    By default the mean vector is initialized by ``init_loc_fn()`` and the
+    Cholesky factor is initialized to the identity.
     """
 
+    def __init__(self, model, prefix="auto", init_loc_fn=init_to_median,
+                 init_scale=1.0):
+        if not isinstance(init_scale, float) or not (init_scale > 0):
```
neerajprad: This is going to be super convenient. Can we also add `init_scale` to the docstring?

pyro/infer/autoguide/guides.py
```diff
@@ -501,22 +506,23 @@ class AutoDiagonalNormal(AutoContinuous):
...
-        pyro.param("auto_loc", torch.randn(latent_dim))
-        pyro.param("auto_scale", torch.ones(latent_dim),
-                   constraint=constraints.positive)
+    initialized to the identity.
     """
 
+    def __init__(self, model, prefix="auto", init_loc_fn=init_to_median,
+                 init_scale=1.0):
+        if not isinstance(init_scale, float) or not (init_scale > 0):
+            raise ValueError("Expected init_scale > 0. but got {}".format(init_scale))
```
neerajprad: Is there a use case where we might want `init_scale` to be differently initialized per parameter? I think in that case we could also allow for a tensor of `latent_size` (though this can also be easily relaxed later depending on use)?

fritzo: Yes, we might want to do that. I think we can handle that in a subsequent PR, where we allow say `init_scale` to be a function like `init_loc_fn` and evaluate it on a per-site basis.

pyro/infer/autoguide/guides.py
fritzo: done.

examples/sparse_regression.py
```diff
@@ -237,24 +245,18 @@ def main(args):
...
-    guide = AutoDelta(model, init_loc_fn=functools.partial(init_to_median, num_samples=50))
-    with torch.no_grad():
-        loss_fn(model, guide, X, Y, hypers)
-    # we also make sure the initial observation noise is not too large
-    # (otherwise we run the danger of getting stuck in bad local optima during optimization).
-    pyro.param('auto_var_obs').unconstrained().data -= 2.0
-
-    with poutine.block(), poutine.trace(param_only=True) as param_capture:
-        guide(X, Y, hypers)
+    guide = AutoDelta(model, init_loc_fn=init_loc_fn)
```
martinjankowiak: nice this much cleaner!

pyro/infer/autoguide/guides.py
```diff
@@ -642,11 +663,12 @@ class AutoLaplaceApproximation(AutoContinuous):
         # ...then train the delta_guide...
         guide = delta_guide.laplace_approximation()
 
-    By default the mean vector is initialized to zero. To change this default behavior
-    the user should call :func:`pyro.param` before beginning inference, e.g.::
+    By default the mean vector is initialized to an empirical median.
```
martinjankowiak: + over the prior?

pyro/infer/autoguide/guides.py
```diff
@@ -501,22 +514,31 @@ class AutoDiagonalNormal(AutoContinuous):
...
+    initialized to the identity.
 
-        latent_dim = 10
-        pyro.param("auto_loc", torch.randn(latent_dim))
-        pyro.param("auto_scale", torch.ones(latent_dim),
-                   constraint=constraints.positive)
+    :param callable model: A generative model.
+    :param callable init_loc_fn: A per-site initialization function.
+        See :ref:`autoguide-initialization` section for available functions.
+    :param float init_scale: Approximate initial scale for the standard
```
martinjankowiak: shouldn't the word 'approximate' be removed here?

pyro/infer/autoguide/guides.py
neerajprad: Since these are parameters it may also be easier for the users to use the default initialization and rescale it later on a component-wise basis. I'm not fully sure if they will be able to do things like `self.scale[0] = ..` though.

pyro/infer/autoguide/guides.py
fritzo: I have removed 'approximate' from `AutoNormal` and `AutoMultivariateNormal`. In the `AutoLowRankMultivariateNormal`, `init_scale` determines the rms standard deviation, since the `cov_factor` must be initialized randomly.

pyro/infer/autoguide/guides.py
fritzo: done.

fritzo:
Thanks for reviewing!

PR #2123, opened 2019-11-06
Fix preditive shapes

### LINKED ISSUE #2122: Predictive class fails for non-scalar sites
I usually use raw poutines to for prediction:
```py
vectorize = pyro.plate("_vectorize", num_samples, dim=-2)
trace = poutine.trace(vectorize(guide)).get_trace()
samples = poutine.replay(vectorize(model), trace)()
```
When instead I try the new `Predictive` class, this fails on models with nontrivial plate structure:
```py
predictive = Predictive(model, guide=guide, return_sites=["x", "y"],
                        num_samples=num_samples, parallel=True)
samples = predictive.get_samples()
```
I believe this is due to reshaping of posterior samples without regard to their event shape. I don't know why `_predictive()` even needs to reshape samples.

I have added a failing test in this branch. Feel free to take over the branch and create a PR:
https://github.com/pyro-ppl/pyro/pull/new/fix-preditive-shapes

fehiepsi:
Resolves #2122

Consider the model
```py
    def model():
        x = pyro.sample("x", dist.Normal(0, 1).expand([2]).to_event())
        with pyro.plate("plate", 5):
            y = pyro.sample("y", dist.Normal(0, 1))
```
, the latent variable `x` has shape `(2,)`. However, in "vectorized" semantic, its shape actually is `(1, 2)`, where the appended dim is to match the plate `"plate"`. So
```py
vectorize = pyro.plate("_vectorize", 10, dim=-2)
trace = poutine.trace(vectorize(model)).get_trace()
```
will give us a trace with `x.shape = (10, 1, 2)`, `y.shape = (10, 5)`.

This PR makes Predictive class follow that convention: that is 10 samples at `x` site will have shape `(10, 1, 2)` instead of `(10, 2)`. This change makes the behavior of `get_vectorized_trace` and `get_samples` with `parallel=True/False` consistent. The only downside I can think about is users might wonder why their samples have additional singleton dimensions. WDYT?

### DIFF
--- a/pyro/infer/predictive.py
+++ b/pyro/infer/predictive.py
@@ -28,7 +28,7 @@ def _guess_max_plate_nesting(model, args, kwargs):
 
 
 def _predictive_sequential(model, posterior_samples, model_args, model_kwargs,
-                           num_samples, sample_sites, return_trace=False):
+                           num_samples, return_site_shapes, return_trace=False):
     collected = []
     samples = [{k: v[i] for k, v in posterior_samples.items()} for i in range(num_samples)]
     for i in range(num_samples):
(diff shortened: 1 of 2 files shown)


pyro/infer/predictive.py
```diff
@@ -54,14 +56,30 @@ def _predictive(model, posterior_samples, num_samples, return_sites=None,
         sample = sample.reshape((num_samples,) + (1,) * (max_plate_nesting - len(sample_shape)) + sample_shape)
         reshaped_samples[name] = sample
 
+    def _vectorized_fn(fn):
```
fritzo: Why did you revert this?

fehiepsi: Oppss... I didn't merge dev branch yet.

fehiepsi: Hmm, this is an issue of IDE I guess... I will change it back

fehiepsi: I see the reason now. I make the change in another branch, then copy that file into this branch, then this kind of thing happens...

fehiepsi: Thanks, I'll try it next time.

fritzo:
> samples have additional singleton dimensions. WDYT?

I think this is a reasonable tradeoff. In an ideal future world, we could use named dimensions rather than positional dimensions. In the short term, it benefits us to be consistent so we can chain inference algorithms. If users have difficulty with these interstitial dimensions of size 1, they can easily call `sample.squeeze()`; we might even document that option for them.

fritzo:
@fehiepsi ready to merge?

neerajprad:
Thanks for fixing this, @fehiepsi. Just fyi, this makes the `vmap` based implementation in numpyro differ from the Pyro version in that it will not insert any interstitial singleton dims. I think this is a fine tradeoff though.

fehiepsi:
I see. In NumPyro, to vectorize, it is better to use `vmap` rather than `plate` primitive, so we might not need this change.

Issue #2095, opened 2019-10-24
Make MCMC behavior identical to NumPyro when num_cpu < num_chains


neerajprad:
In NumPyro, if the number of chains exceed the number of CPU's we draw `num_samples * num_chains` samples sequentially, whereas in Pyro we throw a warning and set `num_chains = available_cpu`, without adjusting `num_samples`. We should try to aim for a uniform behavior in both the cases - NumPyro's behavior seems more sound.

PR #2125, opened 2019-11-06
Draw mcmc chains sequentially when num_cpus < num_chains

fehiepsi:
Addresses #2095.

TODO:
- [x] Add tests to verify that parallel and sequential are consistent, modulo random_seed.

### DIFF
--- a/pyro/infer/mcmc/adaptation.py
+++ b/pyro/infer/mcmc/adaptation.py
@@ -28,7 +28,8 @@ class WarmupAdapter(object):
         self.adapt_mass_matrix = adapt_mass_matrix
         self.target_accept_prob = target_accept_prob
         self.is_diag_mass = is_diag_mass
-        self._init_step_size = 1 if step_size is None else step_size
+        self.step_size = 1 if step_size is None else step_size
+        self._init_step_size = self.step_size
         self._adaptation_disabled = not (adapt_step_size or adapt_mass_matrix)
         if adapt_step_size:
             self._step_size_adapt_scheme = DualAveraging()
--- a/pyro/infer/mcmc/adaptation.py
+++ b/pyro/infer/mcmc/adaptation.py
@@ -28,7 +28,7 @@ class WarmupAdapter(object):
         self.adapt_mass_matrix = adapt_mass_matrix
         self.target_accept_prob = target_accept_prob
         self.is_diag_mass = is_diag_mass
-        self.step_size = 1 if step_size is None else step_size
+        self._init_step_size = 1 if step_size is None else step_size
         self._adaptation_disabled = not (adapt_step_size or adapt_mass_matrix)
         if adapt_step_size:
             self._step_size_adapt_scheme = DualAveraging()
@@ -129,12 +129,10 @@ class WarmupAdapter(object):
             mass matrix is changed.
         """
         self._warmup_steps = warmup_steps
-        if initial_step_size is not None:
-            self.step_size = initial_step_size
+        self.step_size = initial_step_size if initial_step_size is not None else self._init_step_size
         if find_reasonable_step_size_fn is not None:
             self._find_reasonable_step_size = find_reasonable_step_size_fn
-        if inv_mass_matrix is not None:
-            self.inverse_mass_matrix = inv_mass_matrix
+        self.inverse_mass_matrix = inv_mass_matrix
         if self.inverse_mass_matrix is None or self.step_size is None:
             raise ValueError("Incomplete configuration - step size and inverse mass matrix "
                              "need to be initialized.")
--- a/pyro/infer/mcmc/adaptation.py
+++ b/pyro/infer/mcmc/adaptation.py
@@ -45,7 +45,6 @@ class WarmupAdapter(object):
         self._adapt_start_buffer = 75  # from Stan
         self._adapt_end_buffer = 50  # from Stan
         self._adapt_initial_window = 25  # from Stan
-        self._current_window = 0  # starting window index
 
         # configured later on setup
         self._warmup_steps = None
@@ -141,6 +140,11 @@ class WarmupAdapter(object):
                              "need to be initialized.")
         if not self._adaptation_disabled:
             self._adaptation_schedule = self._build_adaptation_schedule()
+        self._current_window = 0  # starting window index
+        if self.adapt_step_size:
+            self._step_size_adapt_scheme.reset()
+        if self.adapt_mass_matrix:
+            self._mass_matrix_adapt_scheme.reset()
 
     def step(self, t, z, accept_prob):
         r"""
--- a/pyro/infer/mcmc/api.py
+++ b/pyro/infer/mcmc/api.py
@@ -103,17 +103,17 @@ class _Worker(object):
             self.result_queue.put_nowait((self.chain_id, e))
 
 
-def _gen_samples(kernel, warmup_steps, num_samples, hook, *args, **kwargs):
+def _gen_samples(kernel, warmup_steps, num_samples, hook, chain_id, *args, **kwargs):
     kernel.setup(warmup_steps, *args, **kwargs)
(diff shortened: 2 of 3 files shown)


pyro/infer/mcmc/adaptation.py
```diff
@@ -28,7 +28,7 @@ def __init__(self,
         self.adapt_mass_matrix = adapt_mass_matrix
         self.target_accept_prob = target_accept_prob
         self.is_diag_mass = is_diag_mass
-        self.step_size = 1 if step_size is None else step_size
+        self._init_step_size = 1 if step_size is None else step_size
```
neerajprad: Was this additional adaptation related refactoring required for sequential chaining, or does this pertain to some other issue?

pyro/infer/mcmc/api.py
```diff
@@ -103,17 +103,17 @@ def run(self, *args, **kwargs):
...
-def _gen_samples(kernel, warmup_steps, num_samples, hook, *args, **kwargs):
+def _gen_samples(kernel, warmup_steps, num_samples, hook, chain_id, *args, **kwargs):
     kernel.setup(warmup_steps, *args, **kwargs)
     params = kernel.initial_params
     # yield structure (key, value.shape) of params
     yield {k: v.shape for k, v in params.items()}
     for i in range(warmup_steps):
         params = kernel.sample(params)
-        hook(kernel, params, 'warmup', i)
+        hook(kernel, params, 'Warmup [{}]'.format(chain_id) if chain_id is not None else 'Warmup', i)
```
neerajprad: I think it is a good idea to pass in chain information for additional logging, but is it needed for sequential chaining?

pyro/infer/mcmc/api.py
```diff
@@ -152,13 +152,20 @@ def terminate(self, *args, **kwargs):
...
+                self.kernel.initial_params = initial_params
+
+            progress_bar = ProgressBar(self.warmup_steps, self.num_samples, disable=self.disable_progbar)
+            logger = initialize_logger(logger, "", progress_bar)
+            hook_w_logging = _add_logging_hook(logger, progress_bar, self.hook)
+            for sample in _gen_samples(self.kernel, self.warmup_steps, self.num_samples, hook_w_logging,
+                                       i if self.num_chains > 1 else None,
+                                       *args, **kwargs):
+                yield sample, i  # sample, chain_id
+            self.kernel.cleanup()
```
neerajprad: We do a lot of state setting in our kernel classes, and `kernel.cleanup()` has not been very thoroughly tested because we never needed to do this for either single / parallel multi-chaining. We will need that for sequential chaining though. If you find that to be an issue, it might be worth handling that separately. Happy to take a look too.

pyro/infer/mcmc/api.py
```diff
@@ -312,26 +319,25 @@ def __init__(self, kernel, num_samples, warmup_steps=None, initial_params=None,
...
+                sequential = True
+        else:
+            if initial_params:
+                initial_params = {k: v.unsqueeze(0) for k, v in initial_params.items()}
 
         self.num_chains = num_chains
         self._diagnostics = [None] * num_chains
 
-        if num_chains > 1:
+        sequential = True
```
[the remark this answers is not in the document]
fehiepsi: Thanks, it forgot to remove it (I use it to debug before knowing your trick of using monkey patch :D).

pyro/infer/mcmc/api.py
fehiepsi: I am also worried about it so I added `test_sequential_consistence` to verify `cleanup` is done properly. Some changes in adapter class clean up adapter too.

pyro/infer/mcmc/api.py
fehiepsi: Oh, I am fine removing this. I do this to make it consistent with MultiSampler progbar. Probably this is just helpful for visualization if num_chains=10 or so. WDYT?

pyro/infer/mcmc/adaptation.py
fehiepsi: yup, this is to reset step_size in adapter when we configure it again.

pyro/infer/mcmc/api.py
```diff
@@ -313,6 +313,9 @@ def __init__(self, kernel, num_samples, warmup_steps=None, initial_params=None,
                     if v.shape[0] != num_chains:
                         raise ValueError("The leading dimension of tensors in `initial_params` "
                                          "must match the number of chains.")
+                # FIXME: probably we want to use "spawn" method by default to avoid the error
+                # CUDA initialization error https://github.com/pytorch/pytorch/issues/2517
+                # even that we run MCMC in CPU.
```
fehiepsi: This error won't be detected in Travis or Mac. This happens when we use Linux (with default mp_context is "fork"), install PyTorch with GPU support, and run MCMC with multi-chains. The solution
> to work around, you need to make sure there are no CUDA calls prior to starting your subprocesses; torch.cuda.is_available() and torch.manual_seed both count as a CUDA calls
is hard to meet in Pyro (e.g. users usually run `pyro.set_rng_seed()` or run some PyTorch code involving checking if a tensor is CUDA tensor).

tests/infer/mcmc/test_mcmc_api.py
```diff
@@ -73,19 +73,24 @@ def normal_normal_model(data):
...
+                mp_context="spawn", initial_params=initial_params, transforms=transforms)
     mcmc.run(data)
     samples = mcmc.get_samples(num_draws, group_by_chain=group_by_chain)
     # test sample shape
     expected_samples = num_draws if num_draws is not None else num_samples
     if group_by_chain:
         expected_shape = (mcmc.num_chains, expected_samples, 1)
+    elif num_draws is not None:
+        # FIXME: what is the expected behavior of num_draw is not None and group_by_chain=False?
+        expected_shape = (expected_samples, 1)
```
fehiepsi: I think this is an issue. We should address it in a separate PR to make the behavior consistent. Probably we can remove `num_samples` arg in `get_samples` and defer the resampling job to a utility.

pyro/infer/mcmc/api.py
neerajprad: Is "spawn" much slower as compared to "fork" or is the speed comparable? One of the reasons why we didn't have "spawn" is that it wasn't available in python 2, but since we have moved to python 3, we can make this change easily if there aren't any performance reasons blocking us. We can do it in a separate PR.

pyro/infer/mcmc/api.py
neerajprad: That sounds reasonable, thanks for explaining.

tests/infer/mcmc/test_mcmc_api.py
neerajprad: I think it is okay to draw `num_samples` from the combined pool if the samples are not grouped by chain. I think it will be even more surprising to return twice or thrice the number of samples unless the chain_dim is preserved. Alternatively we could rename this argument to imply number of samples per chain instead. Deprecation is a possibility too, but this shouldn't affect anything else so I'm fine keeping it.

fehiepsi: Yeah, I'm fine with it too.

pyro/infer/mcmc/api.py
fehiepsi: I remember that fork is faster in Linux. I'll verify it again. If so, I'll ask PyTorch devs if they have a better solution for this issue.

fehiepsi: There is a better solution to this issue: that is to disable CUDA at the beginning of the notebook/script.
```
import os
os.environ["CUDA_VISIBLE_DEVICES"] = ""
```
"spawn" method [does not play well with Jupyter notebook](https://github.com/pytorch/pytorch/issues/20375).

PR #2126, opened 2019-11-06
Make autoguides derive from nn.Module

neerajprad:
Addresses #2078, using the machinery from @fritzo's #2102, #2104. 

This makes AutoGuide derive from `nn.Module` instances, where optimizable parameters are stored as `nn.Parameter` attributes of the instance.
 - The `_setup_prototyp` method is used to initialize autoguide parameters. 
 - We need to call `pyro.module(self.prefix, self)` inside `__call__` so that guide parameters can be captured as `pyro.param` values in the guide trace for SVI. 

**Additional notes:**
 - I had to create a new `call` method that returns the result from `__call__` in the form of a tuple of tensors. This is because functions with the `dict` return type are not currently supported by the JIT tracer. This method can be removed once this restriction is lifted.
 - Note that this requires setting the `constraint` for all static guide parameters even if they are unconstrained. e.g. `self.loc`. Otherwise, these will not be registered as `nn.Parameter`s and won't be registered into the param store by `pyro.module`. **Update** Changed to using `nn.Parameter` for unconstrained parameters.
 - We are using `self.prefix` as the module name so the parameters in the param store will have names like `guide.auto$$$loc_unconstrained`. I think it might be better to use the autoguide name as the module name and append the prefix to the parameter name instead, which will have fewer clashes even with the default prefix?
 - It is the responsibility of the user to ensure that they are storing all autoguide parameters as `nn.Parameter` in their custom autoguides that they pass in to `AutoguideList`. Even otherwise, it should work fine, but those parameters will likely cause issues during serialization. 

**Tests:**
 - Existing tests pass without any modification.
 - New test added for testing guide serialization.

**TODOs:**
 - [x] Fix failing guide serialization tests.

### DIFF
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -178,6 +178,10 @@ class AutoGuideList(AutoGuide):
         if self.prototype_trace is None:
             self._setup_prototype(*args, **kwargs)
 
+        # Needed to capture autoguide parameters as pyro.param statements in the
+        # guide trace
+        if self.master is None:
+            pyro.module(self.prefix, self)
         # create all plates
         self.plates = {frame.name: pyro.plate(frame.name, frame.size, dim=frame.dim)
                        for frame in sorted(self._plates.values())}
@@ -303,7 +307,8 @@ class AutoDelta(AutoGuide):
 
         # Needed to capture autoguide parameters as pyro.param statements in the
         # guide trace
-        pyro.module(self.prefix, self)
+        if self.master is None:
+            pyro.module(self.prefix, self)
 
         plates = self._create_plates()
         result = {}
@@ -434,7 +439,8 @@ class AutoContinuous(AutoGuide):
 
         # Needed to capture autoguide parameters as pyro.param statements in the
         # guide trace
-        pyro.module(self.prefix, self)
+        if self.master is None:
+            pyro.module(self.prefix, self)
 
         latent = self.sample_latent(*args, **kwargs)
         plates = self._create_plates()
@@ -804,7 +810,8 @@ class AutoDiscreteParallel(AutoGuide):
 
         # Needed to capture autoguide parameters as pyro.param statements in the
         # guide trace
-        pyro.module(self.prefix, self)
+        if self.master is None:
+            pyro.module(self.prefix, self)
 
         plates = self._create_plates()
 
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -29,15 +29,16 @@ from pyro.infer.autoguide.utils import _product
 from pyro.infer.enum import config_enumerate
 from pyro.nn import AutoRegressiveNN
 from pyro.ops.hessian import hessian
-from pyro.params import constraint, ConstrainedModule, ConstrainedParameter
+from pyro.params import ConstrainedModule, ConstrainedParameter
 from pyro.poutine.util import prune_subsample_sites
 
 
-class AutoGuide(object):
+class AutoGuide(ConstrainedModule):
     """
     Base class for automatic guides.
 
-    Derived classes must implement the :meth:`__call__` method.
+    Derived classes must implement the :meth:`forward` method, with the
+    same ``*args, **kwargs`` as the base ``model``.
 
     Auto guides can be used individually or combined in an
     :class:`AutoGuideList` object.
@@ -55,20 +56,12 @@ class AutoGuide(object):
         self.prototype_trace = None
         self._plates = {}
 
-    def __call__(self, *args, **kwargs):
-        """
-        A guide with the same ``*args, **kwargs`` as the base ``model``.
-
-        :return: A dict mapping sample site name to sampled value.
-        :rtype: dict
-        """
-        raise NotImplementedError
-
     def call(self, *args, **kwargs):
         """
-        Method that returns parameter values of the guide as a `tuple` instead
-        of a `dict`, which is a requirement for JIT tracing. Unlike `__call__`,
-        this method can be traced by :func:`torch.jit.trace_module`.
+        Method that calls :meth:`forward` and returns parameter values of the
+        guide as a `tuple` instead of a `dict`, which is a requirement for
+        JIT tracing. Unlike :meth:`forward`, this method can be traced by
+        :func:`torch.jit.trace_module`.
 
         .. warning::
             This method may be removed once PyTorch JIT tracer starts accepting
@@ -116,7 +109,7 @@ class AutoGuide(object):
         raise NotImplementedError
 
 
-class AutoGuideList(AutoGuide, nn.Module):
+class AutoGuideList(AutoGuide):
     """
     Container class to combine multiple automatic guides.
 
@@ -166,11 +159,10 @@ class AutoGuideList(AutoGuide, nn.Module):
         # Initialize all module parameters
         for part in self.parts:
             part._setup_prototype(*args, **kwargs)
-            if isinstance(part, nn.Module):
-                for param, value in part.named_parameters():
-                    setattr(self, param, value)
+            for param, value in part.named_parameters():
+                setattr(self, param, value)
 
-    def __call__(self, *args, **kwargs):
+    def forward(self, *args, **kwargs):
         """
         A composite guide with the same ``*args, **kwargs`` as the base ``model``.
 
@@ -239,12 +231,12 @@ class AutoCallable(AutoGuide):
         self._guide = guide
         self.median = median
 
-    def __call__(self, *args, **kwargs):
+    def forward(self, *args, **kwargs):
         result = self._guide(*args, **kwargs)
         return {} if result is None else result
 
 
-class AutoDelta(AutoGuide, ConstrainedModule):
+class AutoDelta(AutoGuide):
     """
     This implementation of :class:`AutoGuide` uses Delta distributions to
     construct a MAP guide over the entire latent space. The guide does not
@@ -293,7 +285,7 @@ class AutoDelta(AutoGuide, ConstrainedModule):
             return store[name]
         return self.init_loc_fn(site)
 
-    def __call__(self, *args, **kwargs):
+    def forward(self, *args, **kwargs):
         """
         An automatic guide with the same ``*args, **kwargs`` as the base ``model``.
 
@@ -330,7 +322,7 @@ class AutoDelta(AutoGuide, ConstrainedModule):
         return self(*args, **kwargs)
 
 
-class AutoContinuous(AutoGuide, nn.Module):
+class AutoContinuous(AutoGuide):
     """
     Base class for implementations of continuous-valued Automatic
     Differentiation Variational Inference [1].
@@ -424,7 +416,7 @@ class AutoContinuous(AutoGuide, nn.Module):
         if not torch._C._get_tracing_state():
             assert pos == latent.size(-1)
 
-    def __call__(self, *args, **kwargs):
+    def forward(self, *args, **kwargs):
         """
         An automatic guide with the same ``*args, **kwargs`` as the base ``model``.
 
@@ -533,11 +525,8 @@ class AutoMultivariateNormal(AutoContinuous):
         super()._setup_prototype(*args, **kwargs)
         # Initialize guide params
         self.loc = nn.Parameter(self._init_loc())
-        self.scale_tril = eye_like(self.loc, self.latent_dim) * self._init_scale
-
-    @constraint
-    def scale_tril(self):
-        return constraints.lower_cholesky
+        self.scale_tril = ConstrainedParameter(eye_like(self.loc, self.latent_dim) * self._init_scale,
+                                               constraints.lower_cholesky)
 
     def get_posterior(self, *args, **kwargs):
         """
@@ -583,11 +572,8 @@ class AutoDiagonalNormal(AutoContinuous):
         super()._setup_prototype(*args, **kwargs)
         # Initialize guide params
         self.loc = nn.Parameter(self._init_loc())
-        self.scale = self.loc.new_full((self.latent_dim,), self._init_scale)
-
-    @constraint
-    def scale(self):
-        return constraints.positive
+        self.scale = ConstrainedParameter(self.loc.new_full((self.latent_dim,), self._init_scale),
+                                          constraints.positive)
 
     def get_posterior(self, *args, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


pyro/infer/autoguide/guides.py
```diff
@@ -217,7 +236,7 @@ def __call__(self, *args, **kwargs):
         return {} if result is None else result
 
 
-class AutoDelta(AutoGuide):
+class AutoDelta(AutoGuide, ConstrainedModule):
```
fritzo: It's is probably safer to avoid multiple inheritance if possible. One way is to avoid letting `AutoGuide` be an `nn.Module`:
```py
class AutoGuide: ...
class AutoGuideList(AutoGuide, nn.Module): ...
class AutoDelta(AutoGuide, ConstrainedModule): ...
class AutoContinuous(AutoGuide, nn.Module): ...
class Auto*Normal(AutoContinuous): ...
```

neerajprad: The AutoDelta's `__mro__` looks fine, i.e. `ConstrainedModule` precedes `nn.Module`. Are there other considerations? 

```
<class 'pyro.infer.autoguide.guides.AutoDelta'>, 
<class 'pyro.infer.autoguide.guides.AutoGuide'>, 
<class 'pyro.params.constrained_parameter.ConstrainedModule'>, 
<class 'torch.nn.modules.module.Module'>, 
<class 'object'>)
```

We can also just use the hierarchy above - the only downside that I can see is that we cannot call module methods like `named_parameters` unless we know which autoguide it is. Doesn't seem like a big deal.

fritzo: OK, if you're happy with `AutoDelta`'s `__mro__`, I'm fine with `AutoGuide(nn.Module)`.

neerajprad: I just followed your suggestion which would keep the parent classes intact for other custom autoguides, and was a simple fix. In any case, those classes like the ones passed to `autocallable` will need to modified for them to be practically usable as `nn.Module`s.

pyro/infer/autoguide/guides.py
```diff
@@ -250,6 +269,12 @@ def __init__(self, model, prefix="auto", init_loc_fn=init_to_median):
         model = InitMessenger(self._init_loc_fn)(model)
         super().__init__(model, prefix=prefix)
 
+    def _init_params(self):
```
fritzo: Should we use the pattern
```py
def _setup_prototype(self, *args, **kwars):
    super()._setup_prototype(*args, **kwargs)
    for name, site in ...:
        # set up parameters
```
as is used in `AutoContinuous` below?

pyro/infer/autoguide/guides.py
```diff
@@ -267,6 +292,11 @@ def __call__(self, *args, **kwargs):
         # if we've never run the model before, do so now so we can inspect the model structure
         if self.prototype_trace is None:
             self._setup_prototype(*args, **kwargs)
+            self._init_params()
```
fritzo: (following the above suggestion, this like could be deleted)

pyro/infer/autoguide/guides.py
```diff
@@ -321,6 +350,13 @@ def __init__(self, model, prefix="auto", init_loc_fn=init_to_median):
         model = InitMessenger(init_loc_fn)(model)
         super().__init__(model, prefix=prefix)
 
+    def _init_params(self):
+        self.loc = self._init_loc()
+
+    @constraint
```
fritzo: We shouldn't need a constraint for loc, it can be unconstrained.

pyro/infer/autoguide/guides.py
```diff
@@ -583,25 +627,29 @@ def __init__(self, model, prefix="auto", init_loc_fn=init_to_median, init_scale=
         self.rank = rank
         super().__init__(model, prefix=prefix, init_loc_fn=init_loc_fn)
 
+    def _init_params(self):
```
fritzo: FYI merge conflict with #2127

pyro/infer/autoguide/guides.py
neerajprad: Yes, I think that sounds reasonable. Let me address this.

pyro/infer/autoguide/guides.py
```diff
@@ -583,25 +627,29 @@ def __init__(self, model, prefix="auto", init_loc_fn=init_to_median, init_scale=
...
+        super()._init_params()
+        self.cov_factor = self.loc.new_empty(self.latent_dim, self.rank).normal_(
+            0, self._init_scale * (0.5 / self.rank) ** 0.5)
+        self.cov_diagonal = self.loc.new_full((self.latent_dim,), 0.5 * self._init_scale ** 2)
+
+    @constraint
+    def cov_factor(self):
```
fritzo: No `@constraint` needed.

pyro/infer/autoguide/guides.py
neerajprad: This is to register them as `nn.Parameters`, otherwise these won't be captured by `pyro.module`. Alternatively, we can have do `nn.Parameter(self._init_loc())`?

tests/infer/test_autoguide.py
```diff
@@ -195,6 +196,44 @@ def model():
...
+    auto_guide_list_x,
+    auto_guide_callable,
+    functools.partial(AutoDiagonalNormal, init_loc_fn=init_to_feasible),
+    functools.partial(AutoDiagonalNormal, init_loc_fn=init_to_mean),
+    functools.partial(AutoDiagonalNormal, init_loc_fn=init_to_median),
+    functools.partial(AutoDiagonalNormal, init_loc_fn=init_to_sample),
+])
+@pytest.mark.parametrize("Elbo", [Trace_ELBO, TraceGraph_ELBO, TraceEnum_ELBO])
+@pytest.mark.xfail(reason="FIXME: auto_laplace, auto_guide_list and auto_guide_callable fail.")
+def test_autoguide_serialization(auto_class, Elbo):
```
fritzo: 💯

pyro/infer/autoguide/guides.py
neerajprad: Changed to `nn.Parameter` - it seems a little odd to have some parameters as `nn.Parameter` and others as regular tensors which are converted into `nn.Parameter` by the constraint descriptor later, but it works.

neerajprad:
Tests for `autoguidelist` and custom autoguides now pass. The only failure seems to be that the serialization of `AutoLaplace` doesn't pass some checks if `check_trace=True`.

pyro/infer/autoguide/guides.py
fritzo: Hmm, good point. Maybe it's cleanest to always set `self.param = nn.Parameter(...)` even if we know that param is constrained.

neerajprad: I tried that. 😄 The issue with that is that if we set `self.scale = nn.Parameter(..)`, it does not go through `ConstraintDescriptor.__set__` and the constraint isn't registered. My guess is that it gets intercepted by `__set__` from `nn.Module`, but I need to look into that.

pyro/infer/autoguide/guides.py
```diff
@@ -337,6 +377,9 @@ def _setup_prototype(self, *args, **kwargs):
         if self.latent_dim == 0:
             raise RuntimeError('{} found no latent variables; Use an empty guide instead'.format(type(self).__name__))
 
+        # Initialize guide params
+        self.loc = nn.Parameter(self._init_loc())
```
fritzo: `.loc` is an implementation detail of a few derived classes, but is not common to all. E.g. `AutoIAFNormal` has no `.loc`. Could you move this attribute down to derived classes: `Auto*Normal`, `AutoLaplaceApproximation`?

pyro/infer/autoguide/guides.py
neerajprad: If we would like to preserve the same overriding behavior for parameters that are initialized as tensors or `nn.Parameter` alike, one option is to create an intermediate class like `AutoGuideModule` (likewise, for `ConstrainedAutoGuideModule`), otherwise any `nn.Parameter` attributes will be set by `nn.Module.__setattr__` without triggering the `__set__` method of the ConstraintDescriptor (and hence no constraints will be set).

```python
class AutoGuideModule(AutoGuide, nn.Module):
    def __setattr__(self, key, value):
        if key in type(self).__dict__ and inspect.isdatadescriptor(getattr(type(self), key)):
            getattr(type(self), key).__set__(self, value)
        else:
            super().__setattr__(key, value)
```

If we do this, we can probably also just have `forward = __call__` for this class.

pyro/infer/autoguide/guides.py
```diff
@@ -688,29 +742,24 @@ def laplace_approximation(self, *args, **kwargs):
...
-        pyro.param(scale_tril_name, scale_tril,
-                   constraint=constraints.lower_cholesky)
-        # force an update to scale_tril even if it already exists
-        pyro.get_param_store()[scale_tril_name] = scale_tril
-
         gaussian_guide = AutoMultivariateNormal(self.model, prefix=self.prefix)
         gaussian_guide._setup_prototype(*args, **kwargs)
+        # Set loc, scale_tril parameters as computed above.
+        gaussian_guide.loc.data = loc
+        gaussian_guide.scale_tril_unconstrained.data = transform_to(constraints.lower_cholesky).inv(scale_tril)
```
neerajprad: @fritzo - Is this the best way to reinitialize `scale_tril`?

neerajprad:
> Is there anything else I should review in detail?

Not much - I have commented on the parts where I would like to get your thoughts. The most important being that currently the `__set__` method will not be triggered for ConstraintDescriptor objects if they are set to `nn.Parameter`. e.g. `self.scale = nn.Parameter(torch.tensor(1.))` will not obey any constraints since the `nn.Module.__setattr__` will intercept this and not call `__set__`. If we want to support this, I have suggested one option above. Otherwise, we'll just need to make sure that any attributes with constraints are initialized as simple tensors and not as `nn.Parameter`.

pyro/infer/autoguide/guides.py
fritzo: Hmm, I would hope we could either
```py
gaussian_guide.scale_tril = scale_tril
```
or if necessary
```py
del gaussian_guide.scale_tril
gaussian_guide.scale_tril = scale_tril
```
Does either of those work?

fritzo: (Note this PR is the most complex use of `ConstraintDescriptor` yet, and we may need to fix bugs)

pyro/infer/autoguide/guides.py
```diff
@@ -396,6 +435,10 @@ def __call__(self, *args, **kwargs):
         if self.prototype_trace is None:
             self._setup_prototype(*args, **kwargs)
 
+        # Needed to capture autoguide parameters as pyro.param statements in the
+        # guide trace
+        pyro.module(self.prefix, self)
```
fritzo: Can we move this line below `self.sample_latent`, to avoid out-of-order issues with #2130

fritzo:
> currently the `__set__` method will not be triggered for `ConstraintDescriptor` objects if they are set to `nn.Parameter`

Very subtle. What do you think of this solution: https://github.com/pyro-ppl/pyro/pull/2130 I have verified locally that we can use `nn.Parameter` with that fix.

fritzo:
@neerajprad on the other hand, #2130 is a little hacky, so I'd be ok using `ConstrainedModule` throughout in this PR, even `class AutoGuide(ConstrainedModule)`. Whatever you like. I'm glad we pushed these classes to find their limitations.

neerajprad:
> @neerajprad on the other hand, #2130 is a little hacky, so I'd be ok using ConstrainedModule throughout in this PR, even class AutoGuide(ConstrainedModule).

I am fine using `ConstrainedModule` throughout as well. Will try to use that first.

pyro/infer/autoguide/guides.py
fritzo: (obsolete if we instead use `ConstrainedModule`)

neerajprad:
I have updated the PR so that `AutoGuide` inherits from `ConstrainedModule` directly. This leads to a lot of code simplifications. e.g. 
 - We can simply call `.named_parameters` or `._parameters` for all guides without doing `isinstance(guide, nn.Module)` checks.
 - This allows us to easily override any `nn.Module` behavior in the future.
 - We can also just reset parameters easily. e.g. in the AutoLaplace case we can do `guide.scale_tril = ...`, without triggering runtime errors from the `nn.Module` class.
 - This lets us also use the `nn.Module`'s default `__call__` method by renaming our existing `__call__` to `forward` instead.

pyro/infer/autoguide/guides.py
neerajprad: This works now if we directly inherit from `ConstrainedModule` without the `del`s. :)

pyro/infer/autoguide/guides.py
```diff
@@ -29,37 +29,47 @@ def model():
...
 
     :param callable model: a pyro model
     :param str prefix: a prefix that will be prefixed to all param internal sites
+        as part of the module name.
     """
 
     def __init__(self, model, prefix="auto"):
+        super().__init__()
         self.master = None
         self.model = model
```
fritzo: I believe this will cause problems if `model` is an `nn.Module`, since e.g. `self.named_parameters` will iterate over model parameters. Do you think we should wrap it to bypass `nn.Module` behavior, e.g.
```py
def __init__(self, model, ...):
    self._model = (model,)
    ...

@property
def model(self):
    return self._model[0]
```

pyro/infer/autoguide/guides.py
```diff
@@ -29,37 +29,47 @@ def model():
...
-        :return: A dict mapping sample site name to sampled value.
-        :rtype: dict
+        .. warning::
+            This method may be removed once PyTorch JIT tracer starts accepting
+            `dict` as valid return types. See
+            `issue <https://github.com/pytorch/pytorch/issues/27743>_`.
         """
-        raise NotImplementedError
+        result = self(*args, **kwargs)
+        return tuple(result.values())
```
fritzo: This has undefined order. Instead
```py
return tuple(v for k, v in sorted(result.items()))
```

pyro/infer/autoguide/guides.py
neerajprad: Hmm..interesting. Let me add a test case and investigate further.

pyro/infer/autoguide/guides.py
neerajprad: I overlooked this working with `nn.Module` where everything is an OrderedDict. Will fix.

pyro/infer/autoguide/guides.py
neerajprad: It turned out that since the models are wrapped in an `InitMessenger` they fail the isinstance check for `nn.Module` and are therefore not registered as submodules for the guides. I am still using this for safety, and have added some cheap median tests for a model that is an `nn.Module` itself.

Issue #2120, opened 2019-11-06
Make autoguides more usable


fritzo:
This aims to make our most common autoguides easier to use: `AutoDelta`, `AutoDiagonalNormal`, `AutoMultivariateNormal`, `AutoLowRankMultivariateNormal`.

Feel free to add to this list (anyone :smile:) if you have usability requests.

I have added this to the 1.0 milestone because it involves interface changes that will enable later refactoring (e.g. #2078) while preserving backwards compatibility.

## Tasks
- [x] #2119 document use of custom `init_loc_fn`, and add examples in tutorials
- [x] #2119 add error checking logic in `init_loc_fn`
- [x] #2119 add easier `init_scale` args for `Auto*Normal` guides
- [x] #2126 make autoguides into `nn.Module`s so they are easier to `torch.load()`/`.save()`
    (this is an alternative to making the param store an `nn.Module` #2078 )
- [x] allow non-python-identifiers as site names in autoguides, ~~which was broken by #2126~~
- ~~make it easier to set optim configs per-guide-param (e.g. I often lower the learning rate `lr *= 0.1` for the `auto_cov_factor` param of `AutoLowRankMultivariateNormal`).~~
- [x] #2127 Stabilize parameterization of `AutoLowRankMultivariateNormal`
- [x] #2129 Support pickling of easyguides
- [x] #2137 Make `AutoGuideList` a bonafide `nn.ModuleList`
- [x] #2140 make easyguides into `nn.Modules`
- [x] #2148 Fix saving of `AutoGuides` in the param store
- [ ] #2156 Change default `init_scale` from 1.0 to 0.1 (@fritzo often uses this in practice)

fritzo:
@martinjankowiak @stefanwebb let me know if you have any other suggestions!

fehiepsi:
@fritzo Could we have a method in PyroModule to get constrained values (without having to use `pyro.get_param_store()`)? Right now, we can observe unconstrained parameters through `.named_parameters()` method but it would be nice to have support for constrained parameters too. :)

fritzo:
@fehiepsi sure, feel free to add a `.get_named_pyro_params()` or similar.

fehiepsi:
@fritzo Just discover a bug with the dev branch :D 
```
class Child(PyroModule):
    def __init__(self, x):
        super().__init__()
        self.a = PyroParam(torch.tensor(x))

class Family(PyroModule):
    def __init__(self):
        super().__init__()
[...]

PR #2113, opened 2019-11-02
Implement a staged Delta -> DiagNormal -> LowRankMVN autoguide

fritzo:
Addresses #2120 

This implements a new autoguide whose complexity can be increased through three stages of training. This captures an inference motif I often use in practice.
1. train an AutoDelta guide
2. train an AutoDiagonalNormal guide
3. train an AutoLowRankMultivariateNormal guide

Additionally this implementation uses previous models to warm-start subsequent stages.

## Tested
- [x] added unit tests
- [ ] validate heuristics on real data

PR #2127, opened 2019-11-07
Change parametrization of AutoLowRankMultivariateNormal

fritzo:
Addresses #2120 

This aims to improve convergence of `AutoLowRankMultivariateNormal` by separating "noise scale" parameters from "noise direction" parameters. Intuitively this allows the uncertainty to grow slowly early in learning, when the autoguide is still learning the loc parameter; thus this PR is a smooth way to accomplish the staged learning of #2113 .

The old parametrization was:
```py
cov = diagonal + cov_factor @ cov_factor.T
```
The new parametrization is:
```py
cov = scale.unsqueeze(-1) * (eye + cov_factor @ cov_factor.T) * scale
    = scale.pow(2).diag_embed() + scaled_cov_factor @ scaled_cov_factor
```
where
```py
scaled_cov_factor = scale.unsqueeze(-1) * cov_factor
```
### Experiment 1

I have tested this on a high-dimensional variational inference problem (dim=22407, rank=100, real world data) and found this PR to stabilize training:
<details>

old loss curve (which required custom per-parameter learning rates):
image
new loss curve (with higher learning rate):
image

</details>

## Experiment 2

This example demonstrates more stable convergence on a synthetic dataset:
https://github.com/pyro-ppl/sandbox/blob/auto-lowrank-mvn/2019-11-lowrank/experiment.ipynb

<details>

image

image

</details>

## Experiment 3

This example demonstrates identification of true low rank structure when the model matches the guide. https://github.com/pyro-ppl/sandbox/blob/auto-lowrank-mvn/2019-11-lowrank/exact.ipynb

<details>

image

</details>

## Experiment 4

This is like experiment 3, but when dimensions are randomly scaled by a scale in [exp(-1), exp(1)]. https://github.com/pyro-ppl/sandbox/blob/auto-lowrank-mvn/2019-11-lowrank/exact-scaled.ipynb

<details>

image

</details>

## Tested
- tested on experiments
- existing behavior is covered by unit tests

### DIFF
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -603,7 +603,7 @@ class AutoLowRankMultivariateNormal(AutoContinuous):
         scale = pyro.param("{}_scale".format(self.prefix))
         factor = pyro.param("{}_cov_factor".format(self.prefix))
         scale = scale * (factor.pow(2).sum(-1) + 1).sqrt()
-        return loc, scale * 2
+        return loc, scale
 
 
 class AutoIAFNormal(AutoContinuous):
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -588,20 +588,22 @@ class AutoLowRankMultivariateNormal(AutoContinuous):
         Returns a LowRankMultivariateNormal posterior distribution.
         """
         loc = pyro.param("{}_loc".format(self.prefix), self._init_loc)
+        scale = pyro.param("{}_scale".format(self.prefix),
+                           lambda: loc.new_full((self.latent_dim,), self._init_scale),
+                           constraint=constraints.positive)
         factor = pyro.param("{}_cov_factor".format(self.prefix),
                             lambda: loc.new_empty(self.latent_dim, self.rank).normal_(
-                                0, self._init_scale * (0.5 / self.rank) ** 0.5))
-        diagonal = pyro.param("{}_cov_diag".format(self.prefix),
-                              lambda: loc.new_full((self.latent_dim,), 0.5 * self._init_scale ** 2),
-                              constraint=constraints.positive)
+                                0, self._init_scale / self.rank ** 0.5))
+        factor = factor * scale.unsqueeze(-1)
+        diagonal = scale * scale
         return dist.LowRankMultivariateNormal(loc, factor, diagonal)
 
     def _loc_scale(self, *args, **kwargs):
         loc = pyro.param("{}_loc".format(self.prefix))
+        scale = pyro.param("{}_scale".format(self.prefix))
         factor = pyro.param("{}_cov_factor".format(self.prefix))
-        diagonal = pyro.param("{}_cov_diag".format(self.prefix))
-        scale = (factor.pow(2).sum(-1) + diagonal).sqrt()
-        return loc, scale
+        scale = scale * (factor.pow(2).sum(-1) + 1).sqrt()
+        return loc, scale * 2
 
 
 class AutoIAFNormal(AutoContinuous):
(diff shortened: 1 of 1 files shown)


pyro/infer/autoguide/guides.py
```diff
@@ -588,20 +588,22 @@ def get_posterior(self, *args, **kwargs):
...
 
     def _loc_scale(self, *args, **kwargs):
         loc = pyro.param("{}_loc".format(self.prefix))
+        scale = pyro.param("{}_scale".format(self.prefix))
         factor = pyro.param("{}_cov_factor".format(self.prefix))
-        diagonal = pyro.param("{}_cov_diag".format(self.prefix))
-        scale = (factor.pow(2).sum(-1) + diagonal).sqrt()
-        return loc, scale
+        scale = scale * (factor.pow(2).sum(-1) + 1).sqrt()
+        return loc, scale * 2
```
martinjankowiak: why 2?

fritzo: oops, fixed. The 2 was a mutation test: I added temporarily just to ensure tests were strong enough to catch that error. Then I forgot to revert the mutation.

fehiepsi:
Bring `diagonal` to `scale` make sense to me, but I am not persuaded that using scaled_cov_factor is better. Just to confirm (though I thought that you have tested it), could you plot the learning curve if we use `scale` and `cov_factor`?

fritzo:
@fehiepsi I'll add some more comprehensive studies tomorrow.

fritzo:
@fehiepsi I have added another experimental study, see notebook.

>  could you plot the learning curve if we use scale and cov_factor?

Can you explain? If we change only `cov_diagonal -> scale`, and leave `cov_factor` as-is, results should be identical to existing Pyro dev (proof: `cov_diagonal` is constrained by constraints.positive, which transformed via `ExpTransform`. Hence the `cov_diagonal` and `scale` parameters will differ by a shift of `log(2)`. Since they only differ by a shift, their gradients will be identical. After identical gradient updates, they will still differ by a shift; hence loss will be identical. []).

fritzo:
I'd like to add more tests to ensure this fits the true covariance of a true man...

fritzo:
OK, I have added another experiment. I am pretty convinced this parametrization is superior. Can reviewers think of any other experiments we might try to provide further evidence?

martinjankowiak:
experiments certainly convince me!

pyro/infer/autoguide/guides.py
```diff
@@ -588,19 +588,21 @@ def get_posterior(self, *args, **kwargs):
         Returns a LowRankMultivariateNormal posterior distribution.
         """
         loc = pyro.param("{}_loc".format(self.prefix), self._init_loc)
+        scale = pyro.param("{}_scale".format(self.prefix),
+                           lambda: loc.new_full((self.latent_dim,), self._init_scale),
```
fehiepsi: do you want to make the init values `0.5 ** 0.5 * self._init_scale` so that initial `self._loc_scale()[1] == self._init_scale`?

pyro/infer/autoguide/guides.py
```diff
@@ -588,19 +588,21 @@ def get_posterior(self, *args, **kwargs):
...
+                           constraint=constraints.positive)
         factor = pyro.param("{}_cov_factor".format(self.prefix),
                             lambda: loc.new_empty(self.latent_dim, self.rank).normal_(
-                                0, self._init_scale * (0.5 / self.rank) ** 0.5))
-        diagonal = pyro.param("{}_cov_diag".format(self.prefix),
-                              lambda: loc.new_full((self.latent_dim,), 0.5 * self._init_scale ** 2),
-                              constraint=constraints.positive)
+                                0, self._init_scale / self.rank ** 0.5))
```
fehiepsi: I think you want `.normal_(0, 1 / self.rank ** 0.5)` here.

pyro/infer/autoguide/guides.py
fritzo: done.

pyro/infer/autoguide/guides.py
fritzo: done.

fritzo:
Let's hold off merging this until after #2126, so as to avoid merge conflicts.

@fehiepsi thanks for the careful review!

fritzo:
I believe this is ready to merge.

PR #2130, opened 2019-11-08
Allow setting of @constraint descriptors to nn.Parameters

fritzo:
Addresses #2126, #2120 

In Python `__get__` takes [precedence](https://docs.python.org/3/howto/descriptor.html#invoking-descriptors) over `__getattr__`, but `__setattr__` takes precedence over `__set__`. This makes it error-prone to set `@constraint`ed parameters of an `nn.Module`, since `nn.Module.__setattr__` can steal control from `ConstraintDescriptor.__set__`.

This PR implements a workaround whereby, after a constrained parameter has been incorrectly set directly in `nn.Module._params`, at the first read of `ConstraintDescriptor.__get__` the parameter will be moved into the correct position.

This solution is a little dangerous, e.g. `pyro.module()` might record incorrect parameters in the first iteration, before the parameter is read. However in typical autoguide use, we can ensure parameters are read before `pyro.module()` is called.

## Tested
- added a unit test
- verified that this allows setting to `nn.Parameter` in #2126

### DIFF
--- a/pyro/params/constrained_parameter.py
+++ b/pyro/params/constrained_parameter.py
@@ -152,17 +152,17 @@ class ConstraintDescriptor:
         if obj is None:
             return self
 
+        # In case setattr(obj, self.name, constrained_value) accidentally
+        # bypassed self.__set__(obj, constrained_value), attempt to repair
+        # the situation.
         try:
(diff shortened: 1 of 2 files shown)


neerajprad:
Thanks for this fix, @fritzo! I'll merge this and update my PR.

Issue #2131, opened 2019-11-08
Register guides in AutoGuideList as submodules


neerajprad:
After #2126, autoguides subclass from `nn.Module` and support all their functionality. The design for `AutoGuideList` can also be made more module friendly. Currently, in `_setup_prototype` we iterate over all the guides and add set their parameters on to the main class so that these are detectable by `pyro.module`. 

Instead, we could add the logic from `AutoGuideList.add` to `__setattr__` in `ConstrainedModule`. Then, we will be able to directly set the subcomponent guides as `autoguidelist.next_guide = ...`. This will register `next_guide` as a submodule and make all its parameters available via `named_parameters()` without us having to do any additional setup in `_setup_prototype`. 

More importantly, currently any guide that itself contains a submodule (we don't have any currently) would throw an error with `AutoGuideList`. The result is that the submodule parameters will be named as `submodule_name.param_name` and as such `setattr(self, "submodule_name.param_name", value)` will fail in `_setup_prototype`. Registering these guides as submodules is much safer since there won't be name clashes and we can get rid of the hacky `setattr` logic in `_setup_prototype`.

fritzo:
Should we aim to make this change by 1.0 release?
Should we fix the "prefix" issue by 1.0 release?

neerajprad:
> Should we fix the "prefix" issue by 1.0 release?

I am hoping that this should be a relatively small fix, and we can do it before 1.0. I can take a look. It will be nice to have this by 1.0.

fritzo:
Closed by #2137

PR #2133, opened 2019-11-09
Remove prefixes from autoguides

neerajprad:
Attempts to address #2131. 
 - This removes prefixes from all autoguides.
 - For `AutoGuideList`, this allows us to set autoguides as below. In this case, `x` and `y` are registered as submodules of `autoguide_list` and their parameters are not immediately set as attributes of `autoguide_list`. However, these attributes are accessible in the same manner as for `nn.Modules`. e.g. `autoguide_list.x.x_unconstrained` gives us the AutoDelta params. The values in the param store will be stored as `AutoGuideList$$$x.x_unconstrained`, where the module name is given by the guide class name.

```python
autoguide_list = AutoGuideList(model)
autoguide_list.x = AutoDelta(poutine.block(model, expose='x'))
autoguide_list.y = AutoDiagonalNormal(poutine.block(model, expose='y'))
```
 - `AutoGuideList.add` is also supported and simply calls `setattr` under the hood. In this case, the names of the submodules are "guide_0", "guide_1", etc.
 - Also, fixes a bug in `AutoGuideList` (I think) where we need to use `poutine.block` when creating plates.

Tested the naming / serialization for a nested autoguide.

### DIFF
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -44,16 +44,14 @@ class AutoGuide(ConstrainedModule):
     :class:`AutoGuideList` object.
 
     :param callable model: a pyro model
-    :param str prefix: a prefix that will be prefixed to all param internal sites
-        as part of the module name.
     """
 
-    def __init__(self, model, prefix="auto"):
+    def __init__(self, model):
         super().__init__()
         self.master = None
         # Do not register model as submodule
         self._model = (model,)
-        self.prefix = prefix
+        self._prefix = "auto"
         self.prototype_trace = None
         self._plates = {}
 
@@ -126,12 +124,10 @@ class AutoGuideList(AutoGuide):
         svi = SVI(model, guide, optim, Trace_ELBO())
 
     :param callable model: a Pyro model
-    :param str prefix: a prefix that will be prefixed to all param internal sites
-        as part of the module name.
     """
 
-    def __init__(self, model, prefix="auto"):
-        super().__init__(model, prefix)
+    def __init__(self, model):
+        super().__init__(model)
         self.parts = []
         self.plates = {}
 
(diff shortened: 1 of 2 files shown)


pyro/infer/autoguide/guides.py
```diff
@@ -44,16 +44,14 @@ class AutoGuide(ConstrainedModule):
...
     """
 
-    def __init__(self, model, prefix="auto"):
+    def __init__(self, model):
         super().__init__()
         self.master = None
         # Do not register model as submodule
         self._model = (model,)
-        self.prefix = prefix
+        self._prefix = "auto"
```
neerajprad: This is to keep things backwards compatible, we can also expose this attribute but it will only be used for the master guide.

fritzo:
Nice! I've also removed some of the need for prefix in #2134.

It looks like the bulk of the remaining complexity is in `AutoGuideList`. What do you think of making it into a bonafide `nn.ModuleList`?
```py
class AutoGuideList(AutoGuide, nn.ModuleList):
    def append(self, part):
        assert isinstance(part, AutoGuide)
        super().append(part)
        part.master = weakref.ref(self)
        ...
    def add(self):
        """Deprecated alias for :meth:`append`."""
        warnings.warn("deprecated", DeprecationWarning)
        self.append(part)
    def forward(self, ...):
        ...
```

neerajprad:
> It looks like the bulk of the remaining complexity is in AutoGuideList. What do you think of making it into a bonafide nn.ModuleList, obviating the need for a __setattr__?

That would be ideal. Should we address it here, or do you think it is better to address that once #2134 is merged?

fritzo:
@neerajprad could we try to merge #2134 before this PR, then refactor `AutoGuideList` to be a bonafide `nn.ModuleList` in this PR? I think that will simplify this PR.

neerajprad:
> @neerajprad could we try to merge #2134 before this PR, then refactor AutoGuideList to be a bonafide nn.ModuleList in this PR? 

Yes, just taking a final look at #2134.

neerajprad:
Closing this in favor of #2137.

PR #2132, opened 2019-11-08
add example usage of hmc with temporal gp model

martinjankowiak:

PR #2134, opened 2019-11-09
Implement PyroModule with poutine aware __getattr__

fritzo:
This implements a `PyroModule` class whose `.__getattr__()` and `.__setattr__()` methods call `pyro.param()` and `pyro.sample()`, and hence which can be effected by poutines. The main new use case is `pyro.infer.autoguid`, which this PR refactors to use `PyroModule` throughout.

This aims to address some common issues as we make Pyro more `nn.Module`-compatible.
1. In #2126 we need to call `pyro.module` despite accessing parameters individually. In theory we could call `pyro.param()` when accessing an attribute. This would additionally allow passing of `event_dims` as is needed to make `EasyGuide` into an `nn.Module` (`event_dims` is needed for param subsampling #238 ).
2. Name logic in autoguides #2133 is arguably unrelated to their interesting statistical properties. This PR aims to factor out that naming logic separately.
3. Martin's https://github.com/pyro-ppl/pyro/pull/2132/files#r344404006 overwrites module attributes with sample statements. This PR aims to move that logic into `PyroModule`.
4. Du's [Parameterized](http://docs.pyro.ai/en/stable/contrib.gp.html#pyro.contrib.gp.parameterized.Parameterized) gp module contains logic to convert between constant, learnable, and sampled attributes.
5. Jake's https://github.com/cornellius-gp/gpytorch/pull/918 uses `.register_prior` to sample parameters.

Note this has very similar goals to `pyro.contrib.named` #540 from two years ago. Since then we've learned more about `nn.Module.__getattr__` implementation details.

After this merges, we can delete `pyro.param.constrained_module`.

## Tested
- [x] svi smoke test
- [x] `PyroParam` shape test
- [x] `PyroSample` test
- [x] serialization test
- [x] refactoring of autoguide is covered by existing tests


pyro/nn/module.py
```diff
@@ -0,0 +1,196 @@
...
+    :class:`PyroModule` s can contain normal :class:`torch.nn.Module` s, but not
+    vice versa. Accessing a normal :class:`torch.nn.Module` attribute of
+    a :class:`PyroModule` triggers a :func:`pyro.module` statement.
+
+    To create a Pyro-managed random attribute, set that attribute using the
+    :class:`PyroSample` helper, specifying a prior distribution. Reading that
+    attribute will then trigger a :func:`pyro.sample` statement. For example::
+
+        # Create Pyro-managed random attributes.
+        my_module.x = PyroSample(dist.Normal(0, 1))
```
neerajprad: I like the interface and the fact that `PyroParam` is poutine aware unlike `ConstrainedParam`. Could you talk about the use case for `PyroSample` as attributes? 

Unlike `params` that are useful as attributes that can be optimized and updated, `sample` statements in most models have a more limited scope and are mostly useful when aggregated. I suppose the biggest difference here is that we don't need to name them. Do you also foresee using these to collect posterior samples from a trained module?

pyro/nn/module.py
```diff
@@ -0,0 +1,196 @@
...
+                                     .to_event(2))
+    """
+    def __init__(self):
+        self._pyro_name = ""
+        self._pyro_params = OrderedDict()
+        self._pyro_samples = OrderedDict()
+        super().__init__()
+
+    def _pyro_set_name(self, name):
+        self._pyro_name = name
```
neerajprad: While not really necessary, it will be nice if the pyro names corresponded to the module / param names that we get from `module.named_parameters` or `module.named_modules`.

pyro/nn/module.py
neerajprad: Okay, I see Martin's PR as being the motivation for this. 🙂

fritzo:
@neerajprad any ideas on the failing test? I can't reproduce locally.

pyro/nn/module.py
fritzo: This should be the case, but I'll add a test.

pyro/nn/module.py
fritzo: As other motivation: consider using Pyro to train a 3rd-party `nn.Module`, say treating it as a Bayesian nn. We would point-estimate some `nn.Parameter`s by translating them to `PyroParam`s, and be fully Bayesian about other parameters, by translating them to `PyroSample` statements with given prior.

It remains to be seen how we can do this post-hoc, since we would need to insert `PyroModule` in the 3rd-party inheritance hierarchy. An intrusive partial approach is to replace `nn.Module` with `pyro.nn.PyroModule` in their code. But I'd prefer to find a way to dynamically change the type of objects in an existing data structure, similar to our fancy `Distribution.expand()` logic. E.g., we could temporarily override `nn.Module.__new__()` to instead return a `pyro.nn.PyroModule`.

pyro/nn/module.py
fritzo: done.

fritzo:
@fehiepsi The `PyroParam` part of this PR seems pretty stable, and we can immediately use it in `AutoGuide`. However the `PyroParam` part is speculative. After this PR merges, could we see whether your `Parametrized` class could use `PyroModule`, and whether `PyroSample` will need to be redesigned?

fehiepsi:
@fritzo It would be nice if PyroParam does not repeatedly transform unconstrained value to constrained value and a PyroSample property only calls `pyro.sample` primitive 1 time during `poutine.trace`. The first issue only affects the speed so we can ignore it for now. But the second issue is important (e.g. when I have a kernel, set LogNormal prior to its length-scale, and want to compute `kernel(X, X)` and `kernel(X, Xu)` in a GP model - this requires me access `kernel.lengthscale` at least two times). Can we have a special primitive that works like `pyro.sample` but `pyro.poutine.trace` will reuse its previous sample, kind of
```
x1 = pyro.module_sample('x', dist.Normal(...))
x2 = pyro.module_sample('x', dist.Normal(...))
assert x1 == x2
```
is a valid model for PyroModule?

pyro/nn/module.py
```diff
@@ -0,0 +1,196 @@
...
+
+        result = super().__getattr__(name)
+
+        # Regular nn.Parameters trigger pyro.param statements.
+        if isinstance(result, torch.nn.Parameter):
+            pyro.param(_make_name(self._pyro_name, name), result,
+                       event_dim=getattr(result, "_pyro_event_dim", None))
+
+        # Regular nn.Modules trigger pyro.module statements.
+        if isinstance(result, torch.nn.Module) and not isinstance(result, PyroModule):
```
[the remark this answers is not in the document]
fritzo: Hmm, while `poutine.block` won't work for this case, the following should work:
```py
my_module.requires_grad_(False)
my_module(*args)  # skips pyro.module calls
```

fritzo:
@fehiepsi I have added `pyro.sample` caching within invocations of `PyroModule.__call__()`.

pyro/nn/module.py
```diff
@@ -0,0 +1,196 @@
+"""
+Pyro includes a special class :class:`~pyro.nn.module.PyroModule` whose
+attributes can be modified by Pyro effects. To create one of these attributes,
+use either the :class:`PyroParam` struct or the :class:`PyroSample` struct::
+
+    my_module = PyroModule()
+    my_module.x = PyroParameter(torch.tensor(1.), constraint=constraints.positive)
```
fehiepsi: typo: PyroParameter -> PyroParam

pyro/nn/module.py
```diff
@@ -68,8 +93,10 @@ class PyroModule(torch.nn.Module):
         x = my_module.x  # Triggers a pyro.sample statement.
         y = my_module.y  # Triggers one pyro.sample + two pyro.param statements.
 
-    Note that, because sample statements can appear only once in a trace, you
-    can read the value of a random attribute only once per inference step.
+    Note that param and sample attribute access is cached within each
```
neerajprad: +1 on this change! This was also my only concern about `PyroSample`, I think this behavior is much more intuitive.

pyro/nn/module.py
fritzo: fixed.

PR #2135, opened 2019-11-11
Use PyroModule in GP

fehiepsi:
Follow-up #2134, this PR makes all GP classes derived from PyroModule.

### Phase 1: make sure that using PyroParam works

I intend to not use Parameterized class in this phase.

- [x] Make Kernel a subclass of PyroModule
- [x] Make Likelihood a subclass of PyroModule
- [x] Make GPModel a subclass of PyroModule
- [x] Make old tests pass

### Phase 2: make sure that using PyroSample works

- [x] Revise Parameterized class to use PyroSample. I expect that this class will only be useful to set AutoDelta/AutoNormal guides automatically. Probably we should rename it into GPModule.
- [x] Make GP tutorial work
- [x] Update all docs
- [x] Update all tutorials
- [x] Make sure that modules that rely on GP module work.


fehiepsi:
@fritzo So far, using `PyroModule` has simplified many internal GP codes. Naming logic for nested modules is expected (which is pretty cool - because I don't know how your last PR handles it correctly). Users' code, which is required to change change `set_prior` -> `PyroSample`, `set_constraints` -> `PyroParam`, is quite similar as before. I have been able to make tests for kernels and parameterized pass, so making the remaining gpmodel's tests pass should be straight forward.

I just wonder about your opinion about the behavior of `mode` property (which is used to trigger if parameters' sample statement should come from prior or autoguide). Can it be simplified using reason changes with AutoGuide/EasyGuide/PyroModule classes?

pyro/contrib/gp/parameterized.py
```diff
@@ -65,70 +63,15 @@ class Parameterized(nn.Module):
...
-            raise ValueError("There is no parameter with name: {}".format(name))
-
-        self._priors[name] = prior
-        # remove the constraint and its unconstrained parameter
-        self.set_constraint(name, constraints.real)
-
-        self.autoguide(name, dist.Delta)
+    def __setattr__(self, name, value):
+        super().__setattr__(name, value)
+        if isinstance(value, PyroSample):
```
fritzo: Regarding mode, could pass a state-dependent distribution, maybe like:
```py
def __setattr__(self, name, value):
    if isinstance(value, PyroSample):
        prior = value.prior
        self._priors[name] = prior
        self.autoguide(name, dist.Delta)
        value = PyroSample(lamba self: prior if self.mode == "model" else
                           self.self._get_guide(name.split(".")[-1], name))
    super().__setattr__(name, value)
```

pyro/nn/module.py
```diff
@@ -121,8 +121,8 @@ def _pyro_set_supermodule(self, name, cache):
         self._pyro_name = name
         self._pyro_cache = cache
         for key, value in self._modules.items():
-            assert isinstance(value, PyroModule)
-            value._pyro_set_supermodule(_make_name(name, key), cache)
+            if isinstance(value, PyroModule):
+                value._pyro_set_supermodule(_make_name(name, key), cache)
```
fehiepsi: @fritzo I think all the tests have passed with only minor changes in the codes. The API is almost the same as previously (you can take a look at GP tutorial, where we only need to change `set_prior` to `PyroSample`). I think that most of the changes would be at your expectations. There are two points which might surprise you so I would like to point out before working with docs:
+ Method `set_mode` is a context manager now (to cache values) without having to run `__call__`.
+ I have relaxed the assertation here. Actually, I don't quite understand the side effects of this change, but without this, [sv-dkl example](https://github.com/pyro-ppl/pyro/blob/dev/examples/contrib/gp/sv-dkl.py) raises an error here. I guess the reason is [CNN](https://github.com/pyro-ppl/pyro/blob/dev/examples/contrib/gp/sv-dkl.py#L42) is a nn.Module instead of `PyroModule`?

pyro/contrib/gp/parameterized.py
fritzo: nit: why do you call `super().__setattr__()` twice? I would have thought a single `super.__setattr__()` outside the if statement at the end of the method would suffice.

pyro/nn/module.py
fritzo: Thanks for fixing this, I am glad we are working out issues of `PyroModule` before the next release!

pyro/contrib/gp/parameterized.py
fehiepsi: My intention was to setup the initial value for autoguide's parameter. But it is not necessary. I'll change it back. Thanks for the trick! ;)

fehiepsi:
> Is there anything you'd like reviewed with extra scrutiny?

I just have a question to make the failing test of Trace_MMD passes. What is the best way to skip backward to PyroParam parameter `p`? The following ways do not work well in general:
+ module.p_unconstrained.requires_grad_(False) together with changing [this line](https://github.com/pyro-ppl/pyro/blob/dev/pyro/nn/module.py#L160) to `if isinstance(result, torch.nn.Parameter) and result.requires_grad_:` work for the SVI interface but does not work for modules using PyTorch optimizers directly.
+ Setting `p` to a specific tensor does not work due to [this logic](https://github.com/pyro-ppl/pyro/blob/dev/pyro/nn/module.py#L210).

The only solution which I can figure out is to `del module.p` then `module.p = value`.

fritzo:
> ... does not work when we train modules using PyTorch optimizers directly.

That surprises me. We should probably add a unit test in tests/nn/test_modules.py.

> The only solution ... is to `del module.p` then `module.p = value`.

Well, that's the recommended solution 🙂

fehiepsi:
> That surprises me.

Sorry, I thought that we need to use this trick: https://github.com/pytorch/pytorch/issues/679#issuecomment-277271892. But it turns out that recent PyTorch versions allow Parameter having `requires_grad=False`.

fehiepsi:
@fritzo Sorry for making confusion: using `p.requires_grad_(False)` actually works! All the tests should pass now. :)

pyro/infer/trace_mmd.py
```diff
@@ -86,18 +86,14 @@ def kernel(self, kernel):
             # fix kernel's parameters
             for k in kernel.values():
                 if isinstance(k, pyro.contrib.gp.kernels.kernel.Kernel):
-                    for p in list(k._pyro_params.keys()):
-                        v = getattr(k, p).detach()
-                        delattr(k, p)
-                        setattr(k, p, v)
+                    for p in k.parameters():
```
fritzo: I think you can even `k.requires_grad_(False)`, but this is good already.

fritzo:
I believe #2140 may contain some fixes to `.requires_grad_(False)`. Prior to that PR the module and param store could have different views of the same tensor, with different `.requires_grad`. After that PR, the param store should be updated to store the same `nn.Parameter` in a `PyroModule`.

pyro/infer/trace_mmd.py
fehiepsi: Thanks for the tip!

pyro/contrib/gp/likelihoods/binary.py
```diff
@@ -47,4 +48,4 @@ def forward(self, f_loc, f_var, y=None):
             y_dist = dist.Bernoulli(f_res)
         if y is not None:
             y_dist = y_dist.expand_by(y.shape[:-f.dim()]).to_event(y.dim())
-        return pyro.sample("y", y_dist, obs=y)
+        return pyro.sample(_make_name(self._pyro_name, "y"), y_dist, obs=y)
```
fritzo: Interesting usage pattern, let's try to later refactor this into `PyroModule`. Maybe we could add something like
```py
class Binary(Likelihood):
    def __init__(...):
        ...
        self.y = PyroConditionalSample()
    def forward(...):
        ....
        return self.y(prior=y_dist, obs=y)
```
Anyway this is something to think about...

pyro/contrib/gp/parameterized.py
```diff
@@ -152,47 +132,50 @@ def autoguide(self, name, dist_constructor):
...
 
-        if dist_constructor is not dist.Delta:
-            # each arg has a constraint, so we set constraints for them
-            for arg in dist_args:
-                self.set_constraint("{}_{}".format(name, arg),
-                                    dist_constructor.arg_constraints[arg])
         self._guides[name] = (dist_constructor, dist_args)
 
+    @pyro_method
+    def load_pyro_samples(self):
```
fritzo: Interesting, if this has other uses, we should consider moving it upstream, maybe to
```py
# in pyro/nn/module.py
@pyro_method  # _method is now a misnomer but behavior works :-)
def sample(module):
    assert isinstance(module, PyroModule)
    for submod in module.modules():
        ...
```
Can you make this private, `_load_pyro_samples()` to leave us room for future refactoring?

pyro/nn/module.py
```diff
@@ -160,10 +160,10 @@ def _pyro_set_supermodule(self, name, context):
         self._pyro_name = name
         self._pyro_context = context
         for key, value in self._modules.items():
-            assert isinstance(value, PyroModule)
-            assert not value._pyro_context.used, \
-                "submodule {} has executed outside of supermodule".format(name)
-            value._pyro_set_supermodule(_make_name(name, key), context)
+            if isinstance(value, PyroModule):
```
fritzo: :+1: thanks for fixing!

pyro/contrib/gp/parameterized.py
fehiepsi: done

pyro/contrib/gp/likelihoods/binary.py
fehiepsi: The current PyroModule is already pretty nice and clean. I really like it. I think that it is better to get feedback from users before adding this new pattern for it.

PR #2137, opened 2019-11-11
Subclass AutoguideList from nn.ModuleList

neerajprad:
Addresses #2120 
Replaces #2133. 

 - This removes the prefix argument from autoguides. The master autoguide will have their names set to their class names. The remaining naming convention just follows PyTorch's `nn.Module`.
 - `AutoGuideList` derives from `nn.ModuleList`. Submodules can can be added directly as attributes or by using `.append`.


pyro/infer/autoguide/guides.py
```diff
@@ -182,12 +178,17 @@ def forward(self, *args, **kwargs):
             self._setup_prototype(*args, **kwargs)
 
         # create all plates
-        self.plates = {frame.name: pyro.plate(frame.name, frame.size, dim=frame.dim)
-                       for frame in sorted(self._plates.values())}
+        with poutine.block():
```
fritzo: Can you explain the need for `poutine.block` here? As I understand, autoguides do not currently support subsampling, and hence the `_Subsample` sites can be ignored; however if we ever were to support subsampling I believe we would need to not block here.

pyro/infer/autoguide/guides.py
```diff
@@ -182,12 +178,17 @@ def forward(self, *args, **kwargs):
...
+            self.plates = {frame.name: pyro.plate(frame.name, frame.size, dim=frame.dim)
+                           for frame in sorted(self._plates.values())}
 
         # run slave guides
         result = {}
-        for part in self.parts:
+        for _, part in self.named_modules():
```
fritzo: Would `self.named_modules(recurse=False)` avoid the need for the `isinstance` check below?

neerajprad: I think we would want to recurse if we have one of the subcomponents that is itself an `AutoGuideList`? The check below is needed when for instance we have a neural network in one of the autoguides (AutoIAFNormal, for instance) that we do not want to call. Calling any submodule that isn't an `AutoGuide` instance isn't needed, but moreover since their return type will not be a dict, this will cause exceptions.

pyro/infer/autoguide/guides.py
neerajprad: I noticed this when working with AutoGuideList where the model has a `plate` statement. In that case, when we generate the guide trace there are two sample calls for the `plate` site - one from `create_plates` and the other in the `__call__`. I think the plate context should be correctly captured in `result.updatepart(*args, **kwargs)`.

fritzo: I think this should actually be
```py
# create all plates
self.plates = self._create_plates()
```

fritzo: Or maybe we should update `AutoGuide._create_plates()` as follows:
```py
class AutoGuide(PyroModule):
    def _update_plates(self):
        if self.master is None:
            self.plates = {frame.name: ...}
        else:
            self.plates = self.master().plates
        return self.plates
````

File: pyro/infer/autoguide/guides.py (hunk 2 above)
Comment (fritzo) [reply pyro/infer/autoguide/guides.py]: Thanks, that makes sense that we should treat `AutoGuide`s differently from other `nn.Module`s.

But I think we should set `recurse=False` and omit the `if isinstance(part, AutoGuideList)` below.

Comment (fritzo) [reply pyro/infer/autoguide/guides.py]: Actually, I think this should simply iterate over `self._modules`. There is no need to recurse.

Comment (fritzo) [reply pyro/infer/autoguide/guides.py]: How about simply:
```diff
- for part in self.parts:
+ for part in self:
```

pyro/infer/autoguide/guides.py
```diff
@@ -199,7 +200,9 @@ def median(self, *args, **kwargs):
         :rtype: dict
         """
         result = {}
-        for part in self.parts:
+        for _, part in self.named_modules():
```
fritzo: Simply
```diff
- for part in self.parts
+ for part in self
```

pyro/infer/autoguide/guides.py
```diff
@@ -43,8 +44,8 @@ class AutoGuide(PyroModule):
     :class:`AutoGuideList` object.
 
     :param callable model: a pyro model
-    :param str prefix: a prefix that will be prefixed to all param internal sites
-        as part of the module name.
+    :param str prefix: Name of module, which is used as a prefix
```
fritzo: I was looking forward to `prefix` being removed from the interface. Do you think we could remove it in this PR, or is that unrelated? IIUC it suffices to set `self._pyro_name = "auto"` in `AutoGuide.__init__()`.

neerajprad: We can remove it easily and set it to the class name or "auto". I'll push an update removing that.

fritzo: Or maybe even the name "guide"?

pyro/infer/autoguide/guides.py
```diff
@@ -272,22 +272,21 @@ def my_init_fn(site):
...
         # Initialize guide params
         for name, site in self.prototype_trace.iter_stochastic_nodes():
-            param_name = "{}_{}".format(self.prefix, name)
             value = PyroParam(site["value"].detach(), constraint=site["fn"].support)
-            setattr(self, param_name, value)
+            setattr(self, name, value)
 
     def _init_loc_fn(self, site):
-        name = "{}_{}".format(self.prefix, site["name"])
+        name = "{}.{}".format(self._pyro_name, site["name"])
```
neerajprad: Is it safe to remove this param store lookup code now? I'm guessing this was to allow users to set control the initialization?

pyro/infer/autoguide/guides.py
```diff
@@ -129,13 +129,10 @@ class AutoGuideList(AutoGuide):
...
 
     :param callable model: a Pyro model
-    :param str prefix: a prefix that will be prefixed to all param internal sites
-        as part of the module name.
     """
 
-    def __init__(self, model, prefix="auto"):
-        super().__init__(model, prefix)
-        self.parts = []
+    def __init__(self, model):
```
fritzo: Nice, can we even delete this `__init__` method now?

pyro/infer/autoguide/guides.py
```diff
@@ -182,12 +183,11 @@ def forward(self, *args, **kwargs):
             self._setup_prototype(*args, **kwargs)
 
         # create all plates
-        self.plates = {frame.name: pyro.plate(frame.name, frame.size, dim=frame.dim)
-                       for frame in sorted(self._plates.values())}
+        self.plates = self._create_plates()
```
fritzo: nit: simply `self._create_plates()` suffices.

pyro/infer/autoguide/guides.py
fritzo: :+1: Yes I believe so. Good catch.

pyro/infer/autoguide/guides.py
neerajprad: Done. I was considering converting into snakecase, but it will only add further complication to the test cases / assertions elsewhere, so I have kept it `type(self).__name__`

tests/infer/test_autoguide.py
```diff
@@ -140,8 +143,8 @@ def model():
 
 def auto_guide_list_x(model):
     guide = AutoGuideList(model)
-    guide.add(AutoDelta(poutine.block(model, expose=["x"])))
-    guide.add(AutoDiagonalNormal(poutine.block(model, hide=["x"])))
+    guide.x = AutoDelta(poutine.block(model, expose=["x"]))
```
fritzo: Should this be `.append(...)`?

pyro/infer/autoguide/guides.py
```diff
@@ -145,6 +145,12 @@ def _update_master(self, master_ref):
         for submodule in self:
             submodule._update_master(master_ref)
 
+    def __setattr__(self, name, value):
```
fritzo: Should this method be moved to the base class `AutoGuide`?

tests/infer/test_autoguide.py
neerajprad: I wanted to check that both directly setting the attribute as well as `.append` works correctly. So both the code-paths are exercised by the tests.

pyro/infer/autoguide/guides.py
neerajprad: Done.

tests/infer/test_autoguide.py
neerajprad: Actually, there are other guides the other code path; so I reverted this to `.append`.

PR #2140, opened 2019-11-12
Make EasyGuide derive from PyroModule

fritzo:
Addresses #2120 

This PR does two things:

**1.** Refactors `PyroModule` to make `event_dims` serializable and to be compatible with local `pyro.param()` statements. The main part of this refactoring was to rely more heavily on the Pyro param store for constraint mapping.

**2.** Makes `EasyGuide` derive from `PyroModule`, similar to #2126 for `AutoGuide`.

## Tested
- [x] fix autoguide serialization test
 - [x] added two unit tests suggested by @fehiepsi


tests/infer/test_autoguide.py
```diff
@@ -250,11 +252,11 @@ def model():
...
-    expected_names = {name for name, _ in guide.named_parameters()}
-    actual_names = {name for name, _ in guide_deser.named_parameters()}
-    assert actual_names == expected_names
-    for name in actual_names:
-        assert_equal(getattr(guide_deser, name), getattr(guide, name).data)
+
+    actual = guide_deser.call()
+    assert len(actual) == len(expected)
+    for name, a, e in zip(names, actual, expected):
+        assert_equal(a, e, msg="{}: {} vs {}".format(name, a, e))
```
fritzo: @neerajprad any ideas why this is failing?

neerajprad: Just taking a look.

neerajprad: Earlier, it seems like the parameters in `_parameters` and those stored as attributes were both unconstrained. With this PR, the parameters in `_parameters` are still unconstrained but those stored as attributes are constrained (which is probably the expected change?). When this autoguide is serialized and loaded back, both the attributes and the parameters in `_parameters` are unconstrained. In this test, we end up comparing the constrained original parameters with the unconstrained loaded ones. I think you have subsequently modified the test though for debugging.

fritzo: Thanks! I've fixed by 1. resurrecting the `_unconstrained` suffix, and 2. fixing rng seeds before calling `guide.call()` to ensure the randomized guides produce identical results.

tests/infer/test_autoguide.py
```diff
@@ -242,14 +243,28 @@ def model():
...
+    names = sorted(guide())
 
     # Ignore tracer warnings
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore", category=torch.jit.TracerWarning)
         # XXX: check_trace=True fails for AutoLaplaceApproximation
         traced_guide = torch.jit.trace_module(guide, {"call": ()}, check_trace=False)
-    torch.jit.save(traced_guide, "/tmp/test_guide_serialization.pt")
-    guide_deser = torch.jit.load("/tmp/test_guide_serialization.pt")
+    f = io.BytesIO()
```
fritzo: I've switched to from `/tmp/*` files to in-memory temp buffers to avoid spurious test errors with pytest-xdist.

pyro/nn/module.py
```diff
@@ -187,14 +201,14 @@ def __setattr__(self, name, value):
...
-            unconstrained_value = torch.nn.Parameter(unconstrained_value)
-            unconstrained_value._pyro_event_dim = event_dim
-            unconstrained_name = name + "_unconstrained"
-            setattr(self, unconstrained_name, unconstrained_value)
+            fullname = _make_name(self._pyro_name, name)
+            pyro.param(fullname, constrained_value, constraint=constraint, event_dim=event_dim)
+            unconstrained_param = _PYRO_PARAM_STORE._params[fullname]
+            if not isinstance(unconstrained_param, torch.nn.Parameter):
+                unconstrained_param = torch.nn.Parameter(unconstrained_param)
+                _PYRO_PARAM_STORE._params[fullname] = unconstrained_param
```
fritzo: @fehiepsi the following two lines should ensure `.requires_grad_(...)` on a module attribute is also respected by the param store.

fehiepsi: Thank Fritz! I'll merge this PR first to see if I need to do something else.

fehiepsi: @fritzo There is some leakage in this logic:
```
import torch
import pyro

m = pyro.nn.PyroModule()
m.a = pyro.nn.PyroParam(torch.tensor(1.))
m.a = pyro.nn.PyroParam(torch.tensor(0.1))
print(m.a)
```
should return 0.1 instead of 1.

fehiepsi: Another leakage is
```import torch
import pyro

class Child(pyro.nn.PyroModule):
    def __init__(self, a):
        super().__init__()
        self.a = pyro.nn.PyroParam(a, torch.distributions.constraints.positive)

class Family(pyro.nn.PyroModule):
    def __init__(self):
        super().__init__()
        self.child1 = Child(torch.tensor(1.))
        self.child2 = Child(torch.tensor(2.))

f = Family()
print(f.child2.a)
```
which returns 1 instead of 2

fritzo: Thanks, I'll add those as unit tests.

fritzo: OK, those tests now pass.

fehiepsi: Thanks! With your latest changes, my deep GP model also works as expected.

tests/infer/test_autoguide.py
```diff
@@ -256,14 +257,28 @@ def model():
...
+    torch.jit.save(traced_guide, f)
+    f.seek(0)
+    guide_deser = torch.jit.load(f)
+
+    # Check .call() result.
+    pyro.set_rng_seed(0)
+    actual = guide_deser.call()
```
neerajprad: For a `TorchScript` instance, we will completely bypass the getattr/setattr logic in our `PyroModule` class (it will probably just exercise some python glue in pytorch and call the underlying C++ object). As such, accessing params will not load their values into the param store, and we should be using `pickle` if we would like to reuse serialized guides for training purposes. Is that correct? Just looking to check my understanding.

fritzo: Correct. Actually I wanted to chat with you about that: `pyro.load_param_store()` will no longer work as a method to save `AutoGuide` or `EasyGuide`s. I'm not sure whether we should simply document this or add new options (similar to `pyro.module(..., update_module_params=True)`. Adding as a task to #2120 .

PR #2143, opened 2019-11-12
Add multi-sample ELBO computation via Tensor Monte Carlo

eb8680:
Resurrects #1977, with some bugfixes and a better understanding of the algorithm and complexity.

I would ideally like to get this PR merged before 1.0 (so that `config_enumerate(expand=False, num_samples=N, default='parallel')` behaves correctly), but followups including #2144 can wait.

Tested:
- Enumeration machinery is mostly exercised by existing tests
- Added test cases to `test_valid_models.py` and `test_enum.py`
- Added support to DMM example


pyro/poutine/enumerate_messenger.py
```diff
@@ -5,13 +5,68 @@
 from .runtime import _ENUM_ALLOCATOR
 
 
+def _tmc_sample(msg):
```
eb8680: This function is the only piece of new functionality in this PR.

pyro/infer/traceenum_elbo.py
```diff
@@ -50,6 +50,19 @@ def _check_model_guide_enumeration_constraint(model_enum_sites, guide_trace):
                                      .format(name, f.name))
 
 
+def _check_tmc_elbo_constraint(model_trace, guide_trace):
```
fritzo: Can you explain the issue this detects?

eb8680: `TraceEnum_ELBO` does not support taking multiple samples from nonreparameterized distributions in the model - it does not compute the corresponding DiCE factors and thus produces incorrect gradients.  This function checks for this case and conservatively raises a warning if any variables in the model are multiply sampled, even if they are reparameterized.

examples/dmm/dmm.py
```diff
@@ -190,7 +190,7 @@ def model(self, mini_batch, mini_batch_reversed, mini_batch_mask,
         # this marks that each datapoint is conditionally independent of the others
         with pyro.plate("z_minibatch", len(mini_batch)):
             # sample the latents z and observed x's one time step at a time
-            for t in range(1, T_max + 1):
+            for t in pyro.markov(range(1, T_max + 1)):
```
martinjankowiak: this change could use a comment

pyro/poutine/enumerate_messenger.py
```diff
@@ -5,13 +5,68 @@
...
+    # find batch dims that aren't plate dims
+    batch_shape = [1] * len(dist.batch_shape)
+    for f in msg["cond_indep_stack"]:
+        if f.vectorized:
+            batch_shape[f.dim] = f.size
+    batch_shape = tuple(batch_shape)
+
+    # sample a batch
+    sample_shape = (num_samples,) if batch_shape == dist.batch_shape else ()
+    fat_sample = dist(sample_shape=sample_shape)  # TODO thin before sampling
```
martinjankowiak: what'd the TODO about?

pyro/infer/traceenum_elbo.py
eb8680: `TraceEnum_ELBO` does not support taking multiple samples from nonreparameterized distributions in the model - it does not compute the corresponding DiCE factors and thus produces incorrect gradients. This function checks for this case and conservatively raises a warning if any variables in the model are multiply sampled, even if they are reparameterized.

pyro/poutine/enumerate_messenger.py
eb8680: If a sample site depends on `N > 1` upstream sample sites that have been multiply sampled, `fat_sample` will contain `num_samples ** (N-1)` too many samples, which are then discarded in the `while` loop below this line.  A more efficient implementation would apply the logic in this function to the parameters of `dist` before sampling, but doing this generically would require extra distribution plumbing.

tests/infer/test_enum.py
```diff
@@ -437,14 +437,15 @@ def guide():
     ]))
 
 
-@pytest.mark.parametrize("enumerate1", [None, "sequential", "parallel"])
-@pytest.mark.parametrize("enumerate2", [None, "sequential", "parallel"])
-@pytest.mark.parametrize("enumerate3", [None, "sequential", "parallel"])
+@pytest.mark.parametrize("num_samples", [None, 10000])
```
fritzo: Could we keep the old `enumerate=None` params? This isn't horribly ugly:
```py
ELBO_BERN_PARAMS = [
    (e1, e2, e3, num_samples)
    for e1 in [None, "sequential", "parallel"]
    for e2 in [None, "sequential", "parallel"]
    for e3 in [None, "sequential", "parallel"]
    for num_samples in [None, 10000]
    if num_samples is None or None not in (e1, e2, e3)
]
@pytest.mark.parametrize("enumerate1,enumerate2,enumerate3,num_samples",
                         ELBO_BERN_PARAMS)
...
```

tests/infer/test_enum.py
```diff
@@ -574,10 +579,11 @@ def guide():
     ]))
 
 
-@pytest.mark.parametrize("enumerate1", [None, "sequential", "parallel"])
-@pytest.mark.parametrize("enumerate2", [None, "sequential", "parallel"])
+@pytest.mark.parametrize("num_samples", [None, 10000])
```
fritzo: ditto, could we strictly extend the old test param grid?

tests/infer/test_enum.py
```diff
@@ -658,13 +664,14 @@ def guide():
     ]))
 
 
-@pytest.mark.parametrize("enumerate4", [None, "sequential", "parallel"])
-@pytest.mark.parametrize("enumerate3", [None, "sequential", "parallel"])
-@pytest.mark.parametrize("enumerate2", [None, "sequential", "parallel"])
-@pytest.mark.parametrize("enumerate1", [None, "sequential", "parallel"])
+@pytest.mark.parametrize("num_samples", [None, 100000])
```
fritzo: ditto

tests/infer/test_enum.py
```diff
@@ -715,12 +722,13 @@ def guide():
     ]))
 
 
-@pytest.mark.parametrize("enumerate3", [None, "sequential", "parallel"])
-@pytest.mark.parametrize("enumerate2", [None, "sequential", "parallel"])
-@pytest.mark.parametrize("enumerate1", [None, "sequential", "parallel"])
+@pytest.mark.parametrize("num_samples", [None, 2000])
```
fritzo: ditto

examples/dmm/dmm.py
martinjankowiak: some of the code snippets in dmm.ipynb would also need to be updated

tests/infer/test_enum.py
fritzo: You might even want to keep the whole grid for tmc sampling. IIRC this test is cheap.

fritzo:
@eb8680 if you don't have time to get this in by the 1.0 release, I'd be happy with a 2-line `NotImplementedError` if `num_samples > 0 and expand=False`. Then this PR would no longer technically be a breaking change. That would also give us room to announce this a a cool new feature in 1.1.

eb8680:
I should be able to finish this by this evening

examples/dmm/dmm.py
eb8680: > some of the code snippets in dmm.ipynb would also need to be updated

I added a comment to the script.  I'll leave additional explanation/advertising to a followup PR.

PR #2144, opened 2019-11-12
Add Tensor Monte Carlo loss function and tests

eb8680:
This PR implements the importance sampling estimate of marginal likelihood described in the paper [Tensor Monte Carlo](https://arxiv.org/abs/1806.08593) using tensor variable elimination and the enumeration machinery added in #2143 .

Tasks:
- [x] Add tests with plated models

Tested:
- [x] Added tests of TMC loss and gradients against ground-truth for categorical and gaussian chains
- [x] Added TMC to DMM example
- [x] Add to tests/infer/test_valid_models.py


eb8680:
> Could you add TensorMonteCarlo to a bunch of the parameter lists in tests/infer/test_valid_models.py, just wherever it seems relevant?

Done.

> Is there anything else you'd like reviewed in special detail?

In the last commit I unified the single-site DiCE factor computation in `pyro.infer.util.Dice.__init__` and `TensorMonteCarlo` into a helper function `compute_site_dice_factor` to avoid duplicating some rather intricate logic.  No behavior should have changed, but it would be good to get a second set of eyes on it.

pyro/infer/tmc.py
```diff
@@ -0,0 +1,195 @@
...
+    To take multiple samples, mark the site with
+    ``infer={'enumerate': 'parallel', 'num_samples': N}``.
+    To configure all sites in a model or guide at once,
+    use :func:`~pyro.infer.enum.config_enumerate`.
+    To enumerate or sample a sample site in the ``model``,
+    mark the site and ensure the site does not appear in the ``guide``.
+
+    This assumes restricted dependency structure on the model and guide:
+    variables outside of an :class:`~pyro.plate` can never depend on
+    variables inside that :class:`~pyro.plate`.
```
fritzo: Note I sometimes get inscrutable sphinx errors with "`" followed by "." in sphinx links. The workaround is to always leave a space before the ".".

fritzo:
What tests are failing? My internet is too flaky to connect to travis.

eb8680:
> What tests are failing?

A couple of JIT tests and examples that exercise edge cases of the `TraceEnum_ELBO` Dice logic.  It seems like there might be a type conversion issue upstream in PyTorch since the non-JIT versions of these tests pass, but I can work around it.

eb8680:
> Do I understand correctly that TMC is like our Enum implementation, except instead of proposing from the uniform distribution we propose from a guide distribution and need to account for that proposal distribution?

Yes, that's right.

> you might consider naming this TMC_ELBO or similar

Renamed to `TraceTMC_ELBO`.

>  I would find it helpful to see docstrings or comments-of-purpose on each of the _compute_*() functions.

Done.

PR #2145, opened 2019-11-12
Add Tensor Monte Carlo support to more existing examples

eb8680:
This PR follows #2144 by adding support for the `TensorMonteCarlo` implementation added there to more existing examples.


tests/test_examples.py
```diff
@@ -22,6 +22,11 @@
     'capture_recapture/cjs.py --num-steps=1 -m 3',
     'capture_recapture/cjs.py --num-steps=1 -m 4',
     'capture_recapture/cjs.py --num-steps=1 -m 5',
+    'capture_recapture/cjs.py --num-steps=1 -m 1 --tmc --tmc-num-samples=2',
```
martinjankowiak: why use tmc in scenarios where i can compute the exact marginal likelihood?

eb8680: You shouldn't, I just think it's neat that TMC makes Monte Carlo gradient estimation feasible in these models.

martinjankowiak: ok but anywhere this is true (hmm, cjs) we should make a comment to the effect that is illustrative only

examples/capture_recapture/cjs.py
```diff
@@ -263,9 +263,17 @@ def expose_fn(msg):
...
     # we need to use TraceEnum_ELBO.
-    elbo = TraceEnum_ELBO(max_plate_nesting=1, num_particles=20, vectorize_particles=True)
-    optim = Adam({'lr': args.learning_rate})
-    svi = SVI(model, guide, optim, elbo)
+    if args.tmc:
+        elbo = TraceTMC_ELBO(max_plate_nesting=1)
+        tmc_model = poutine.infer_config(
+            model,
+            lambda msg: {"num_samples": args.tmc_num_samples, "expand": False} if msg["infer"].get("enumerate", None) == "parallel" else {})  # noqa: E501
+        optim = Adam({'lr': args.learning_rate})
```
fritzo: nit: you could move `optim` out of the conditional

examples/capture_recapture/cjs.py
```diff
@@ -290,5 +298,11 @@ def expose_fn(msg):
     parser.add_argument("-d", "--dataset", default="dipper", type=str)
     parser.add_argument("-n", "--num-steps", default=400, type=int)
     parser.add_argument("-lr", "--learning-rate", default=0.002, type=float)
+    parser.add_argument("--tmc", action='store_true',
+                        help="Use Tensor Monte Carlo instead of exact enumeration " +
```
fritzo: note: no need for "+" sign in multi-line string literals

examples/hmm.py
```diff
@@ -582,12 +582,22 @@ def main(args):
...
-    optim = Adam({'lr': args.learning_rate})
-    svi = SVI(model, guide, optim, elbo)
+    if args.tmc:
+        if args.jit:
+            raise NotImplementedError("jit support not yet added for TraceTMC_ELBO")
+        elbo = TraceTMC_ELBO(max_plate_nesting=1 if model is model_0 else 2)
+        tmc_model = poutine.infer_config(
+            model,
+            lambda msg: {"num_samples": args.tmc_num_samples, "expand": False} if msg["infer"].get("enumerate", None) == "parallel" else {})  # noqa: E501
+        optim = Adam({'lr': args.learning_rate})
```
fritzo: nit: pull `optim` out of conditional

examples/smcfilter.py
```diff
@@ -102,12 +99,50 @@ def main(args):
     logging.info("std: {}".format(z.variance ** 0.5))
 
 
+def tmc_run(args):
```
fritzo: Could you add some docstrings to `tmc_run()` and `smc_run` to inform the first-time reader that tmc is like smc but shares trajectories, and maybe adding some educational material describing similarities to other smc algorithms? This file seems like a good place for that material.

also nit: maybe call these `infer_smc`/`infer_tmc` or other descriptive name

examples/capture_recapture/cjs.py
```diff
@@ -263,9 +263,17 @@ def expose_fn(msg):
 
     # since we enumerate the discrete random variables,
     # we need to use TraceEnum_ELBO.
```
fritzo: ... use either TraceEnum_ELBO or TraceTMC_ELBO.

examples/smcfilter.py
eb8680: Since `TraceTMC_ELBO` isn't directly comparable to `SMCFilter` in functionality (e.g. no backward posterior sampling impemented, filtering vs smoothing) and is still pretty unstable, I'll just revert the changes to this example for now and make a more complete version once I've addressed those issues.

tests/test_examples.py
```diff
@@ -115,6 +137,10 @@ def xfail_jit(*args):
     'hmm.py --num-steps=1 --truncate=10 --model=4 --jit',
     'hmm.py --num-steps=1 --truncate=10 --model=5 --jit',
     'hmm.py --num-steps=1 --truncate=10 --model=7 --jit',
+    'hmm.py --num-steps=1 --truncate=10 --model=1 --tmc --tmc-num-samples=2 --jit',
```
fritzo: I think you'll need to wrap these in `xfail_jit(-)`

PR #2146, opened 2019-11-12
Add smoke test for PyroModule based linear regression

neerajprad:
I was exploring if we could use `PyroModule` to replace `random_module` but got into issues like:
 - Site names containing `.` and being rejected by `nn.Module`.
 - Site names not matching between guide/model traces.

This failing test captures the issues involved.

### DIFF
--- a/tests/infer/test_autoguide.py
+++ b/tests/infer/test_autoguide.py
@@ -18,8 +18,9 @@ from pyro.infer.autoguide import (AutoCallable, AutoDelta, AutoDiagonalNormal, A
                                   AutoGuideList, AutoIAFNormal, AutoLaplaceApproximation, AutoLowRankMultivariateNormal,
                                   AutoMultivariateNormal, init_to_feasible, init_to_mean, init_to_median,
                                   init_to_sample)
-from pyro.nn.module import PyroModule, PyroParam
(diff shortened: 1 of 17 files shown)


tests/infer/test_autoguide.py
```diff
@@ -568,10 +568,53 @@ def forward(self):
...
+            self.weight = PyroSample(lambda self: dist.Normal(0., 1.).expand([out_features, in_features])
+                                     .to_event(2))
+            self.bias = PyroSample(lambda self: dist.Normal(0., 10.).expand([out_features]).to_event(1))
+
+    class LinearRegression(PyroModule):
+        def __init__(self):
+            super().__init__()
+            self.linear = RandomLinear(D, 1)
+
+        def forward(self, x, y):
```
neerajprad: @fritzo - Does this look correct?

fritzo: Looks right to me.

martinjankowiak:
@neerajprad :+1: it is my belief that we should consider deprecating lift/random_module in favor of PyroModule. the former paradigm seems to cause a lot of confusion among users

neerajprad:
The tests pass now. This is so much cleaner than `random_module`. 😆

neerajprad:
Ready to merge.

PR #2148, opened 2019-11-13
Support saving of PyroModules in PyroParam store

fritzo:
Addresses #2120 

This aims to preserve our old serializability path, allowing `PyroModule`s and `AutoGuide`s to be saved and loaded via a side-effect of `pyro.get_param_store().load()`. This also fixes a bug whereby the param store was populated with partially qualified names during nested module construction, e.g. using @fehiepsi 's example:
```py
class Child(PyroModule):
    def __init__(self, x):
        super().__init__()
        self.a = nn.Parameter(torch.tensor(x))

class Family(PyroModule):
    def __init__(self):
        self.a = Child(1.0)  # Child.__setattr__('a', ...) used to trigger a premature pyro.param()
        self.v = Child(2.0)
```
To achieve these ends, I have
1. Renamed `._pyro_cache()` to `._pyro_context()` and restricted pyro primitives to be called only when the context is active (inside `__call__` or methods marked `@pyro_method`). Thus no pyro.param statements will be called during `__init__()`. Not foolproof, but less buggy.
2. Clearly defined precedence for when PyroModule takes precedence over ParamStore, and added comments to that effect.
3. In `PyroModule.__delattr__`, remove parameters from the param store.

## Tested
- added a new serialization test
- added some `len(pyro.get_param_store())` assertions to avoid duplicate param name errors.

### DIFF
--- a/tests/nn/test_module.py
+++ b/tests/nn/test_module.py
@@ -297,14 +297,15 @@ def test_pyro_serialize():
     actual = module()  # triggers saving in param store
     assert_equal(actual[0], module.x)
     assert_equal(actual[1], module.y)
-    assert len(pyro.get_param_store()) == 2
+    assert set(pyro.get_param_store().keys()) == {"x", "y"}
+    assert_equal(pyro.param("x").detach(), module.x.detach())
+    assert_equal(pyro.param("y").detach(), module.y.detach())
 
-    # Work around https://github.com/pytorch/pytorch/issues/27972
     pyro.get_param_store().save("/tmp/pyro_module.pt")
     pyro.clear_param_store()
     assert len(pyro.get_param_store()) == 0
     pyro.get_param_store().load("/tmp/pyro_module.pt")
-    assert len(pyro.get_param_store()) == 2
+    assert set(pyro.get_param_store().keys()) == {"x", "y"}
     actual = MyModule()
     actual()
 
--- a/tests/nn/test_module.py
+++ b/tests/nn/test_module.py
@@ -62,14 +62,27 @@ def test_svi_smoke():
 
 
 def test_names():
-    root = PyroModule()
-    root.x = nn.Parameter(torch.tensor(0.))
-    root.y = PyroParam(torch.tensor(1.), constraint=constraints.positive)
-    root.m = nn.Module()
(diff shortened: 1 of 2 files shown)


tests/nn/test_module.py
```diff
@@ -249,19 +257,57 @@ def gather(self):
...
+    assert_equal(actual[0], module.x)
+    assert_equal(actual[1], module.y)
+    assert len(pyro.get_param_store()) == 2
+
+    # Work around https://github.com/pytorch/pytorch/issues/27972
+    pyro.get_param_store().save("/tmp/pyro_module.pt")
+    pyro.clear_param_store()
+    assert len(pyro.get_param_store()) == 0
+    pyro.get_param_store().load("/tmp/pyro_module.pt")
+    assert len(pyro.get_param_store()) == 2
```
neerajprad: Should we also check that the values in the param store before `MyModule` is called matches the module attributes after call? Feel free to disregard this if it is already exercised by some other test.

neerajprad: Btw..I'm happy to merge this and resolve these issues in follow-up PRs.

fritzo:
@neerajprad ready to merge after tests pass.

tests/nn/test_module.py
```diff
@@ -249,19 +257,57 @@ def gather(self):
...
+
+    assert isinstance(module.x, torch.Tensor)
+    assert len(pyro.get_param_store()) == 0
+
+    actual = module()  # triggers saving in param store
+    assert_equal(actual[0], module.x)
+    assert_equal(actual[1], module.y)
+    assert len(pyro.get_param_store()) == 2
+
+    # Work around https://github.com/pytorch/pytorch/issues/27972
```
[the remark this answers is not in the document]
fritzo: done

tests/nn/test_module.py
fritzo: done

fehiepsi:
> Is this ready to be merged?

The PR looks good to me! I am playing around with some hierarchy patterns. If I notice any issue, I will raise them in https://github.com/pyro-ppl/pyro/issues/2120#issuecomment-553210426 instead.

PR #2150, opened 2019-11-13
Recursively set attributes for dynamic autoguides

neerajprad:
This resolves one of the issues in #2146, where `PyroSample` statements in submodules throw errors in dynamic autoguides like `AutoDelta` because we try to set attributes with names like `module.x` which throws an error in `nn.Module`. This instead does a recursive `setattr` - initializing `PyroModule`s recursively if the prefix attributes do not exist. We can merge this separately or as part of #2146.


neerajprad:
> Do you want to add tests to this PR, or wait for #2146?

I'll just merge this along with your changes into #2146, if this looks good.

PR #2151, opened 2019-11-13
Add a causal effect vae example

fritzo:
Paper: http://papers.nips.cc/paper/7223-causal-effect-inference-with-deep-latent-variable-models.pdf
Original code: https://github.com/AMLab-Amsterdam/CEVAE

I plan to refactor this into reusable module `pyro.contrib.cevae` in a follow-up PR.

## Tested
- verified sane results on the synthetic dataset from the paper
- added a smoke test to test_examples.py


examples/cevae.py
```diff
@@ -0,0 +1,65 @@
...
+        raise NotImplementedError
+
+    def y_dist(self, t, y):
+        raise NotImplementedError
+
+    def z_dist(self, t, y, x):
+        raise NotImplementedError
+
+
+def ite(model, guide, x, num_samples=100):
```
fritzo: @eb8680 am I using `poutine.do` correctly here? I am literally transcribing
image

eb8680: If you want `ite` to correspond to the posterior predictive distributions in section 4 of the CEVAE paper then unfortunately no - `poutine.do` doesn't do backdoor adjustment, it just mutilates a model according to the definition of the do-operator.  

The expectations in the ITE are taken with respect to the *model*'s posterior marginal over `y | x, do(t=T)`, which is in turn computed by averaging over samples of the *marginal* distribution of `z` in the *guide*, which requires yet another marginalization over the auxiliary distributions over `y` and `t` in the *guide*.  

That is, to compute the ITE for the causal model in the paper, we do not intervene on `t` in the guide and we average over two distinct sets of samples from `y` - one from the guide and one from the model, which we return:
```py
guide_plate = pyro.plate('guide_particles', num_samples, dim=-2)
model_plate = pyro.plate('model_particles', num_samples, dim=-3)
with poutine.block(hide=['y', 't']), guide_plate:
    guide_trace = poutine.trace(poutine.block(guide, hide=['y', 't'])).get_trace(x)
with model_plate, guide_plate:
    y0 = poutine.replay(poutine.do(model, data=dict(t=0)), guide_trace)(x)
```

I believe these tedious and error-prone manipulations could be automated almost entirely if we implement a more complete semantics for counterfactuals, such as the formalism described in [Single World Intervention Graphs: A Primer](https://pdfs.semanticscholar.org/07bb/cb458109d2663acc0d098e8913892389a2a7.pdf).

examples/cevae.py
```diff
@@ -0,0 +1,170 @@
...
+    model = Model(args)
+    guide = Guide(args)
+
+    dataset = TensorDataset(x, t, y)
+    dataloader = DataLoader(dataset, batch_size=args.batch_size, shuffle=True)
+    num_steps = args.num_epochs * len(dataloader)
+    optim = ClippedAdam({"lr": args.learning_ratee,
+                         "lrd": args.learning_rate_decay ** (1 / num_steps)})
+
+    # We construct the CEVAE loss by applying poutine.do to the guide.
```
fritzo: @eb8680 Does this look right?

examples/cevae.py
```diff
@@ -0,0 +1,311 @@
...
+        loc0, scale0 = self.z0_nn(y_x)
+        loc1, scale1 = self.z1_nn(y_x)
+        loc = torch.where(t.bool().unsqueeze(-1), loc1, loc0)
+        scale = torch.where(t.bool().unsqueeze(-1), scale1, scale0)
+        return dist.Normal(loc, scale).to_event(1)
+
+
+class TraceCausalEffect_ELBO(Trace_ELBO):
+    """
+    The CEVAE objective (to maximize) is::
```
eb8680: nit: cite [1]

examples/cevae.py
```diff
@@ -0,0 +1,311 @@
...
+        self.z1_nn = DiagNormalNet([1 + args.feature_dim] +
+                                   [args.hidden_dim] * args.num_layers +
+                                   [args.latent_dim])
+
+    def forward(self, x, t=None, y=None, size=None):
+        if size is None:
+            size = x.size(0)
+        with pyro.plate("data", size, subsample=x):
+            t = pyro.sample("t", self.t_dist(x),
+                            obs=t, infer={"is_auxiliary": True})
```
eb8680: Add comments here describing the effect of this `is_auxiliary` annotation?

examples/cevae.py
```diff
@@ -0,0 +1,311 @@
...
+                                   [args.hidden_dim] * args.num_layers +
+                                   [args.latent_dim])
+        self.z1_nn = DiagNormalNet([1 + args.feature_dim] +
+                                   [args.hidden_dim] * args.num_layers +
+                                   [args.latent_dim])
+
+    def forward(self, x, t=None, y=None, size=None):
+        if size is None:
+            size = x.size(0)
+        with pyro.plate("data", size, subsample=x):
```
eb8680: nit: use `subsample_size=x.size(0)` explicitly here for clarity since `x` isn't actually an index

examples/cevae.py
```diff
@@ -0,0 +1,311 @@
...
+        self.y0_nn = BernoulliNet([args.latent_dim] +
+                                  [args.hidden_dim] * args.num_layers)
+        self.y1_nn = BernoulliNet([args.latent_dim] +
+                                  [args.hidden_dim] * args.num_layers)
+        self.t_nn = BernoulliNet([args.latent_dim])
+
+    def forward(self, x, t=None, y=None, size=None):
+        if size is None:
+            size = x.size(0)
+        with pyro.plate("data", size, subsample=x):
```
eb8680: nit: use `subsample_size=x.size(0)` explicitly here for clarity since `x` isn't actually an index

fritzo:
Thanks for the detailed review, @eb8680 !
> make a Jupyter notebook version with ...

Good idea. Do you mind if I do that in a follow up PR, refactoring logic into `pyro.contrib.cevae` and then using that in a tutorial?

examples/cevae.py
```diff
@@ -0,0 +1,311 @@
...
+            layers.append(nn.Linear(in_size, out_size))
+            layers.append(nn.ELU())
+        layers.pop(-1)
+        super().__init__(*layers)
+
+
+def nn_init(m):
+    if isinstance(m, nn.Linear):
+        m.weight.data.normal_(0, 0.1 / m.in_features ** 0.5)
+        m.bias.data.fill_(0)
```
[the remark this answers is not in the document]
fritzo: I was just following the paper, but indeed default works fine. I'll delete this.

examples/cevae.py
fritzo: Oh it turns out I can omit these (as I learned from discussion in #2196).

examples/cevae.py
fritzo: Hmm, this version avoids triggering a `_Subsample` site, whereas the version `subsample_size=len(x)` would trigger extra sampling. While I realize this is a bit of a hack since `x` is not an index tensor, it is a hack that we often use to avoid the extra computation.

examples/cevae.py
fritzo: done.

PR #2152, opened 2019-11-13
Add named_pyro_params method to PyroModule

fehiepsi:
This PR adds an analogous method to `named_parameters()` of nn.Module, which yields all Pyro params.

The issue https://github.com/pyro-ppl/pyro/issues/2120#issuecomment-553247132 is also fixed in this PR.

### DIFF
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -69,6 +69,16 @@ class _Context:
             self.cache[name] = value
 
 
+def _get_pyro_params(module):
+    for name in module._parameters:
+        if name.endswith("_unconstrained"):
+            constrained_name = name[:-len("_unconstrained")]
+            if isinstance(module, PyroModule) and constrained_name in module._pyro_params:
+                yield constrained_name, getattr(module, constrained_name)
+                continue
+        yield name, module._parameters[name]
+
+
 class PyroModule(torch.nn.Module):
     """
     Subclass of :class:`torch.nn.Module` that supports setting of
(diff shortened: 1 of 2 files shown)


tests/nn/test_module.py
```diff
@@ -129,6 +134,31 @@ def __init__(self):
...
+
+    f = Family()
+    f.c = Child(3.)
+    assert_equal(f().detach(), torch.tensor(3.))
+
+    g = Family()
+    g.c = Child(4.)
+    assert_equal(g().detach(), torch.tensor(4.))
+    # FIXME: failing test
+    assert_equal(f().detach(), torch.tensor(3.))
```
fehiepsi: @fritzo I give up dealing with this failing test. :( I think that failing is the expected behavior here (because we only have 1 global ParamStoreDict for 2 global PyroModule instances). Some possible solutions are
+ Set `_pyro_name = id(self)` by default. This will create ugly names for param sites in a trace.
+ Allow users to set `_pyro_name` in __init__ method.

fehiepsi: Nvm, this can be resolved with `pyro.clear_param_store()` or combining these two PyroModules into one. The later one will resolve the conflict of `f`'s parameters and `g`'s parameters in ParamStore.

pyro/nn/module.py
```diff
@@ -69,6 +69,16 @@ def set(self, name, value):
             self.cache[name] = value
 
 
+def _get_pyro_params(module):
```
fritzo: nit: this would be slightly faster as
```py
if not isinstance(module, PyroModule):
    yield from module._parameters.items()
    return
_pyro_params = module._pyro_params
for ...
    if ...
        if constrained_name in _pyro_params:
            ...
   ...
```
but no need to change :smile:

PR #2155, opened 2019-11-14
Make Predictive a PyroModule

### LINKED ISSUE #2154: Refactor Predictive to make it a serializable PyroModule
While working on #2083, I found that the Predictive class is not working for AutoGuides as expected. 

One issue is that using a Predictive module where we set model and trained guide as submodules throws an error `submodule model has executed outside of supermodule".format(name)`.

It will be nice to refactor it so that it:
 - [ ] Returns correct results for autoguides.
 - [ ] Supports serializability as `ScriptModule`.
 - [ ] Faster serializability for the sequential predictive function.

neerajprad:
Attempts to fix #2154 - added a failing test.

### DIFF
--- a/pyro/infer/predictive.py
+++ b/pyro/infer/predictive.py
@@ -5,7 +5,6 @@ import torch
 
 import pyro
 import pyro.poutine as poutine
-from pyro.nn import PyroModule
 from pyro.poutine.util import prune_subsample_sites
 
 
@@ -46,7 +45,7 @@ def _predictive_sequential(model, posterior_samples, model_args, model_kwargs,
                 for site, shape in return_site_shapes.items()}
 
 
-def _predictive(model, posterior_samples, num_samples, return_sites=None,
+def _predictive(model, posterior_samples, num_samples, return_sites=(),
                 return_trace=False, parallel=False, model_args=(), model_kwargs={}):
     max_plate_nesting = _guess_max_plate_nesting(model, model_args, model_kwargs)
     vectorize = pyro.plate("_num_predictive_samples", num_samples, dim=-max_plate_nesting-1)
@@ -67,10 +66,15 @@ def _predictive(model, posterior_samples, num_samples, return_sites=None,
     for site in model_trace.stochastic_nodes + model_trace.observation_nodes:
         append_ndim = max_plate_nesting - len(model_trace.nodes[site]["fn"].batch_shape)
         site_shape = (num_samples,) + (1,) * append_ndim + model_trace.nodes[site]['value'].shape
-        if return_sites is not None:
+        # non-empty return-sites
+        if return_sites:
             if site in return_sites:
                 return_site_shapes[site] = site_shape
-        # include all sites not in posterior samples by default
+        # special case (for guides): include all sites
+        elif return_sites is None:
+            return_site_shapes[site] = site_shape
+        # default case: return sites = ()
+        # include all sites not in posterior samples
         elif site not in posterior_samples:
             return_site_shapes[site] = site_shape
 
@@ -100,7 +104,7 @@ def _predictive(model, posterior_samples, num_samples, return_sites=None,
     return predictions
 
 
-class Predictive(PyroModule):
+class Predictive(torch.nn.Module):
     """
     This class is used to construct predictive distribution. The predictive distribution is obtained
     by running model conditioned on latent samples from `posterior_samples`.
@@ -124,7 +128,7 @@ class Predictive(PyroModule):
         all batch dims correctly annotated via :class:`~pyro.plate`. Default is `False`.
     """
     def __init__(self, model, posterior_samples=None, guide=None, num_samples=None,
-                 return_sites=None, parallel=False):
+                 return_sites=(), parallel=False):
         super().__init__()
         if posterior_samples is None:
             if num_samples is None:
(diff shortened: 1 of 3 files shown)


pyro/infer/predictive.py
```diff
@@ -163,13 +180,15 @@ def get_samples(self, *args, **kwargs):
         """
         posterior_samples = self.posterior_samples
         if self.guide is not None:
-            # use return_sites='' as a special signal to return all sites
```
fehiepsi: I wonder if docs for `return_sites`: "by default only sample sites not present in `posterior_samples` are returned" is still valid now? The old behavior makes more sense to me.

neerajprad: You mean for the case where we have both a guide and non-empty `posterior_samples`?

neerajprad: I was thinking of actually raising an exception in that case, and adding support for it later if requested.

fehiepsi: I mean for the case `return_sites=None`, we will include all sites by default with this change. The former behavior is "only sample sites not present in posterior_samples are returned".

pyro/infer/predictive.py
```diff
@@ -66,15 +67,15 @@ def _predictive(model, posterior_samples, num_samples, return_sites=None,
...
         append_ndim = max_plate_nesting - len(model_trace.nodes[site]["fn"].batch_shape)
         site_shape = (num_samples,) + (1,) * append_ndim + model_trace.nodes[site]['value'].shape
-        if isinstance(return_sites, (list, tuple, set)):
+        if return_sites is not None:
             if site in return_sites:
                 return_site_shapes[site] = site_shape
         else:
-            if (return_sites is not None) or (site not in reshaped_samples):
-                return_site_shapes[site] = site_shape
+            # include all sites by default
```
fehiepsi: @neerajprad I meant this case.

neerajprad: That's a mistake, thanks for catching! Can I just use the `if site not in posterior_samples` check here, or do we really need the sentinel `''` value?

pyro/infer/predictive.py
```diff
@@ -99,7 +100,7 @@ def _predictive(model, posterior_samples, num_samples, return_sites=None,
     return predictions
 
 
-class Predictive:
+class Predictive(PyroModule):
```
fritzo: Hmm, this will currently break in a few ways:
1. in `PyroModule._pyro_set_supermodule()`, `assert not value._pyro_context.used` will fail
2. names will change in `model` and `guide`, invalidating the `PyroModule`-paramstore coupling.

Maybe we should make `Predictive` a `torch.nn.Module` instead?

neerajprad: Yes, I bumped into exactly those two issues while debugging. 😞 Let me try your suggestion instead.

pyro/infer/predictive.py
fehiepsi: I think using `''` is not needed if we raise an exception where both `guide` and `posterior_samples` are specified. I don't know in which situation users need both of them. :D

pyro/infer/predictive.py
fritzo: On further thought, I don't see how `PyroModule` is appropriate. `nn.Module` seems like the right way to serialize. But I don't fully understand the issues...

fritzo: For example if `model` and `guide` have `PyroSample` sites, those addresses must match exactly, and that can't happen in a single `PyroModule`, but is fine if they live in a common `nn.Module`.

neerajprad: You are right - the poutine calls should also be serialized correctly. I am chasing some unrelated issues though, still debugging.

pyro/infer/predictive.py
neerajprad: I think you original cases are indeed needed. :) Seems like we would need to distinguish between three cases, but I'm just using a `()` instead of `''`.
 - when `return_sites` is non empty we return samples from it.
 - when `return_sites = ()` (default), we return samples not in posterior samples.
 - when `return_sites = None` and a `guide` argument is provided (only used internally when a guide is provided), we return all sites. This is important because a guide will pass in `posterior_samples` but we should return those samples to the user.

fehiepsi: Yeah, that looks good to me!

tests/infer/test_autoguide.py
```diff
@@ -596,22 +597,77 @@ def test_linear_regression_smoke(auto_class, Elbo):
...
+        warnings.filterwarnings("ignore", category=torch.jit.TracerWarning)
+        traced_predictive = torch.jit.trace_module(predictive, {"call": (x,)})
+    f = io.BytesIO()
+    torch.jit.save(traced_predictive, f)
+    f.seek(0)
+    predictive_deser = torch.jit.load(f)
+    pyro.set_rng_seed(0)
+    samples_deser = predictive_deser.call(x)
+    # Note that the site values are different in the serialized guide
+    assert len(samples) == len(samples_deser)
```
neerajprad: This works but I the traced predictive module does not give exactly the same values as the original one (fixing the `rng` seed). I think this should be good for now, but I'll try to harden the test if I figure out why the traced module gives different results.

fritzo:
LGTM on the `nn.Module` refactoring. @fehiepsi have you reviewed the other changes?

pyro/nn/module.py
```diff
@@ -161,8 +164,11 @@ def _pyro_set_supermodule(self, name, context):
         self._pyro_context = context
         for key, value in self._modules.items():
             assert isinstance(value, PyroModule)
-            assert not value._pyro_context.used, \
+            assert not value._pyro_context.active, \
                 "submodule {} has executed outside of supermodule".format(name)
+            if value._pyro_context.used:
```
fritzo: Why is this needed?

neerajprad: Hmm..sorry, none of this is needed. This was from a debugging session that I thought was reverted in the PR. Will remove.

PR #2156, opened 2019-11-14
Tune defaults of Auto*Normal guides

### LINKED ISSUE #2120: Make autoguides more usable

fritzo:
Resolves #2120 

1. Reduce `init_scale` from 1.0 to 0.1. I typically use 0.1 for small models and 0.01 for large models. I choose 0.1 as a nice value for beginner users. Not so large that the model goes wild early in training (which I often see with 1.0), but not so small that it takes a long time for variance to increase.
2. Use a default heuristic of `rank ~ sqrt(latent_dim)` in `AutoLowRankMultivariateNormal`. This is a good computational tradeoff in practice. The old default of `rank=1` was silly.
3. Tune optimizers to reduce test time. `pytest -n auto tests/infer/test_autoguide.py` reduces from 42sec to 12sec on my laptop.

## Tested
- updated `test_mean` and `test_quantiles`
- used these defaults in small and large practical models

### DIFF
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -606,7 +606,7 @@ class AutoLowRankMultivariateNormal(AutoContinuous):
         svi = SVI(model, guide, ...)
 
     By default the ``cov_diag`` is initialized to a small constant and the
-    ``cov_factor`` is intialized randomly such that on average
+    ``cov_factor`` is initialized randomly such that on average
     ``cov_factor.matmul(cov_factor.t())`` has the same scale as ``cov_diag``.
 
     :param callable model: A generative model.
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -517,7 +517,7 @@ class AutoMultivariateNormal(AutoContinuous):
         svi = SVI(model, guide, ...)
 
     By default the mean vector is initialized by ``init_loc_fn()`` and the
-    Cholesky factor is initialized to the identity.
+    Cholesky factor is initialized to the identity times a small factor.
 
     :param callable model: A generative model.
     :param callable init_loc_fn: A per-site initialization function.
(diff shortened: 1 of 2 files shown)


pyro/infer/autoguide/guides.py
```diff
@@ -607,22 +605,24 @@ class AutoLowRankMultivariateNormal(AutoContinuous):
         guide = AutoLowRankMultivariateNormal(model, rank=10)
         svi = SVI(model, guide, ...)
 
-    By default the ``cov_diag`` is initialized to 1/2 and the ``cov_factor`` is
-    intialized randomly such that ``cov_factor.matmul(cov_factor.t())`` is half the
-    identity matrix.
+    By default the ``cov_diag`` is initialized to a small constant and the
+    ``cov_factor`` is intialized randomly such that on average
```
martinjankowiak: typo

Issue #2153, opened 2019-11-13
Implement single-world counterfactual semantics for causal inference


eb8680:
To support applications like the Causal Effect VAE #2151 I propose we implement the node-splitting transformation described in detail in [Single World Intervention Graphs: A Primer](https://pdfs.semanticscholar.org/07bb/cb458109d2663acc0d098e8913892389a2a7.pdf).  I believe a simple version of this would be quite straightforward and fully compatible with all existing inference algorithms - we'd just need a `CounterfactualMessenger` subsuming the functionality of `poutine.condition` and `poutine.do`:
```py
model = poutine.counterfactual(
    model,
    observations={'x': x_obs, 'y': None, 'z': z_obs},
    interventions={'x': None, 'y': y, 'z': z_cf}
)
```
There are three cases to handle in implementing SWIG-style node splitting:
1. When a variable appears in `observations` but not `interventions` (such as site `"x"` in this example) , the behavior for that variable should be identical to the behavior under `poutine.condition`: the variable is marked observed and its value is set to the observation (here `x_obs`).
2. When a variable appears in `interventions` but not `observations` (such as site `"y"` in this example), the behavior for that variable should be to set the site's value to the intervention and block any further processing of the site as in the current `poutine.do`.  However, it should also introduce a new sample site with a different name that reuses the stochastic function at the original site.  Note that this will have no statistical effect when the model is normalized, i.e. when there are no observations elsewhere in the model.
3. When a variable appears in both `interventions` and `observations` (such as site `"z"` in this example), the behavior should follow case 2 except that the fresh sample site should also be marked observed as in case 1 with the value of the observation (here `z_obs`).

robertness:
I think this is a good idea, as SWIGs are a good visual way of reasoning about counterfactuals on a single DAG.

Can you clarify the following?  How does backdoor adjustment come into play here?  Is there automatic estimation of causal effects?
> and will automatically perform backdoor adjustment when the model is not normalized.

I think, that from a design perspective, ideally new abstractions encourage (or at least won't discourage) Pearl-style counterfactuals on an SCM-style model.

An SCM can be specified as a probabilistic program where all the stochasticity is relegated to exogenous noise variables. To use Peter's and Schölkopf's canonical example (page 36):

```
def model(noise_dist):
   nt, nb = sample('noise', noise_dist)   
   t = sample('t', Dirac(nt))
   b = sample('b', Dirac(t * nb + (1-t) * (1-nb))
   return b
```
[...]

PR #2157, opened 2019-11-14
Rewrite poutine.do to represent single-world counterfactuals

eb8680:
Addresses #2153.

Before this PR, `poutine.do` and `poutine.condition` did not compose in an intuitive way to produce counterfactual queries.  After the changes in this PR, `poutine.do` modifies models to represent the joint distribution over a variable to be intervened on and a potential outcome after a particular intervention on that variable, as opposed to simply mutilating the model.

cc @robertness - what do you think of this change (proposed in #2153)?  Is it sensible?  Would it break a bunch of existing code?  I'm far from being an expert in causal inference but I'm hoping to find ways we can automate away more do-calculus drudgery in Pyro.

### DIFF
--- a/pyro/poutine/do_messenger.py
+++ b/pyro/poutine/do_messenger.py
@@ -1,4 +1,9 @@
+import numbers
 import warnings
+
+import torch
+
+from pyro.distributions import Delta, Distribution
 
 from .messenger import Messenger
 from .runtime import apply_stack
@@ -28,7 +33,8 @@ class DoMessenger(Messenger):
 
         >>> intervened_model = do(model, data={"z": torch.tensor(1.)})
 
-    This is equivalent to replacing `z = pyro.sample("z", ...)` with `z = value`
+    This is equivalent to replacing `z = pyro.sample("z", ...)` with
+    `z = pyro.sample("z__CF", dist.Delta(v=torch.tensor(1.)))`
     and introducing a fresh sample site pyro.sample("z", ...) whose value is not used elsewhere.
 
     References
@@ -47,7 +53,7 @@ class DoMessenger(Messenger):
 
     def _pyro_sample(self, msg):
         if msg.get('_intervener_id', None) != self._intervener_id and \
-                self.data.get(msg['name']) is not None:
+                msg['name'] in self.data:
 
             if msg.get('_intervener_id', None) is not None:
                 warnings.warn(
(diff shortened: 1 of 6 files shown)


pyro/poutine/do_messenger.py
```diff
@@ -0,0 +1,58 @@
...
+    :returns: stochastic function decorated with a :class:`~pyro.poutine.do_messenger.DoMessenger`
+    """
+    def __init__(self, data):
+        super(DoMessenger, self).__init__()
+        self.data = data
+        self._intervener_id = str(id(self))
+
+    def _pyro_sample(self, msg):
+        if msg.get('_intervener_id', None) != self._intervener_id and \
+                msg['name'] in self.data and \
```
fritzo: nit: simplify to `self.data.get(msg['name']) is not None`

tests/poutine/test_counterfactual.py
```diff
@@ -0,0 +1,69 @@
+import logging
+
+import pytest
+
+import pyro
+import pyro.distributions as dist
+import pyro.poutine as poutine
+from tests.common import assert_equal, assert_not_equal
+
+logger = logging.getLogger(__name__)
```
fritzo: logger is unused, causing flake8 failure

eb8680:
> Do I understand correctly

Yes, I believe so, except that

> Before this PR poutine.do() would block the guide from conditioning the model, and the model would erroneously draw x from the prior

is not quite right: under the current behavior the sample site `"x"` would return `intervention` in both model and guide but would not appear in either trace.

pyro/poutine/do_messenger.py
```diff
@@ -0,0 +1,96 @@
...
+            intervention = self.data[msg['name']]
+            msg['name'] = msg['name'] + "__CF"  # mangle old name
+
+            # if intervention is None, use same variable with different randomness
+            # this is equivalent to the posterior predictive distribution
+            if intervention is None:
+                intervention = msg['fn']
+
+            if isinstance(intervention, Distribution):
+                # stochastically intervened site is no longer observed
```
fritzo: Maybe we should leave the `Distribution` case `NotImplemented` for the release?

It seems more subtle and is untested. For example in the variational intervention case
```py
with poutine.do(...):
    with poutine.trace() as tr:
        guide()
    with poutine.replay(trace=tr.trace):
        model()
```
it seems that the samples drawn from the distributional intervention will not agree between the guide and the model.

Issue #2116, opened 2019-11-04
Misc tasks for 1.0 release


fritzo:
- [x] Replace `DeprecationWarning` with ~`UserWarning`~ `FutureWarning`.
- [x] Add stability statement to ~~README~~ release notes (describing what is and is not stable)
- [x] Mark unstable parts of the interface as EXPERIMENTAL (e.g. `pyro.nn.PyroModule`) #2158, ...
- [x] Try to fix codecov badge (no need to try hard tho)

fritzo:
Let me know if anything else needs to be marked EXPERIMENTAL.

PR #2158, opened 2019-11-14
Add more docs to PyroModule

fritzo:
Addresses #2116 

1. Adds more docs, including EXPERIMENTAL marker and description of pyromodule <--> param store synchronization.
2. Adds a `name` arg to `__init__`, mostly to help explain naming in docs.
3. Adds a `clear` function that removes data from both a `PyroModule` and the param store.

## Tested
- used `name` in autoguides
- added a test for `clear`

### DIFF
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -404,5 +404,7 @@ def clear(mod):
     assert isinstance(mod, PyroModule)
     for name in list(mod._pyro_params):
         delattr(mod, name)
+    for name in list(mod._parameters):
+        delattr(mod, name)
     for name in list(mod._modules):
         delattr(mod, name)
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -1,7 +1,8 @@
 """
-Pyro includes a special class :class:`~pyro.nn.module.PyroModule` whose
-attributes can be modified by Pyro effects. To create one of these attributes,
-use either the :class:`PyroParam` struct or the :class:`PyroSample` struct::
+Pyro includes an experimental class :class:`~pyro.nn.module.PyroModule`, a
+subclass of :class:`torch.nn.Module`, whose attributes can be modified by Pyro
+effects.  To create a poutine-aware attribute, use either the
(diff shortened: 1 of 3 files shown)


pyro/nn/module.py
```diff
@@ -81,16 +82,21 @@ def _get_pyro_params(module):
 
 class PyroModule(torch.nn.Module):
     """
-    Subclass of :class:`torch.nn.Module` that supports setting of
-    :class:`PyroParam` and :class:`PyroSample` .
+    EXPERIMENTAL Subclass of :class:`torch.nn.Module` whose attributes can be
```
martinjankowiak: does that mean we reserve the right to change the interface?

tests/nn/test_module.py
```diff
@@ -217,6 +217,39 @@ def test_constraints(shape, constraint_):
...
+
+    # mutate
+    for x in pyro.get_param_store().values():
+        x.unconstrained().data += torch.randn(())
+    state1 = m()
+    for x, y in zip(state0, state1):
+        assert not (x == y).all()
+
+    clear(m)
+    del m
```
fehiepsi: This will leave `x` in the param store because `clear` only remove root module's Pyro params and its submodules, while `x` is a `nn.Parameter` here. Is it expected? If so, it would be clearer to add `assert len(pyro.get_param_store()) == 1` here.
On the other hand, because all parameters of `Model` are nn.Parameter, `clear` and `del` here are not necessary. I think that defining `self.p.x = pyro.nn.PyroParam(torch.tensor(3.))` would be a better test.

pyro/nn/module.py
fritzo: Correct. I am especially uncertain about the synchronization policy between pyro param store and `PyroModule`.  We might soon declare this non-experimental in say pyro 1.1.

tests/nn/test_module.py
fritzo: Good catch! I've fixed and strengthened tests.

PR #2160, opened 2019-11-14
Make it clear when fullname is defined in PyroModule

fritzo:
Addresses #2161 

1. This makes adds a `._pyro_get_fullname()` method, to replace the use of `_make_name` in #2135 . This method also adds new checks to ensure that `fullname` is actually well-defined; these checks required some fixes to `__delattr__`. Hopefully these new checks will surface misuses earlier an aid debugging.
2. Removes dependency of `pyro.nn.module` on `pyro.distributions`, which was causing problems in #1932 . (relatedly, removes a dependency in lift_messenger.py)

## Tested

### DIFF
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -15,7 +15,6 @@ import torch
 from torch.distributions import constraints, transform_to
 
 import pyro
-import pyro.distributions as dist
 from pyro.poutine.runtime import _PYRO_PARAM_STORE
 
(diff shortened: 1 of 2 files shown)


pyro/nn/module.py
```diff
@@ -48,13 +47,11 @@ class _Context:
     def __init__(self):
         self.active = 0
         self.cache = {}
-        if __debug__:
-            self.used = False
+        self.used = False
```
fritzo: The `.used` field is now used outside of assertions.

Issue #2083, opened 2019-10-23
Update all tutorials / examples for Pyro 1.0 release


neerajprad:
We should update the tutorials / examples to illustrate best practices:

 - [ ] Check that all non-runnable tutorials give correct results. Alternatively, make an attempt to convert them into valid notebooks that users can play around with. 
 - [ ] Make an attempt to use the `Predictive` utility class in bayesian regression and baseball examples. We don't need to do this all over the place, but only in a few basic tutorials, which a first time user of Pyro is likely to start with. It is also instructive to see how to use the replay and trace handlers to do this directly.
 - [ ] A fully developed example to show how to serve model predictions in C++, for SVI and MCMC. #2073. 
 - [ ] There are quite a few examples where we write our own summary functions to summarize posterior distributions, e.g. in baseball, and bayesian regression examples, using mean, median and different quantiles. It will be nice if we can use the `summary` function from `mcmc.util` for this. Right now, the summary function just prints the summary information. It will be better if it returns a dict of summary information, and there is a separate function for pretty printing it. 
 - [ ] Remove `pyro.random_module` from the bayesian regression example, or make it more friendly for initial users, e.g. providing checks on prior values, etc.

neerajprad:
If you noticed anything else from the forums, please update this list.

fehiepsi:
> It will be better if it returns a dict of summary information, and there is a separate function for pretty printing it.

Sound reasonable to me. I'll refactor it and update related examples.

eb8680:
It would be nice to have the LDA example produce sensible numbers on a real dataset.

neerajprad:
> It would be nice to have the LDA example produce sensible numbers on a real dataset.

I initially starting using that for the 20 news dataset as an example for model serving (which was probably a bad idea), but it was a bit too slow for the large vocabulary size. I'll dig into that a bit, and update https://github.com/pyro-ppl/pyro/issues/2072 to gather ideas on how we can make it work on that.

PR #2162, opened 2019-11-14
Updates to the Bayesian Regression tutorial

neerajprad:
Addresses #2083, #2054. 

This updates the Bayesian Regression tutorial to:
 - Remove usage of `random_module` and `TracePredictive` in favor of `PyroModule` and `Predictive`.
 - Adds a section on model serving using `TorchScript`.
 - Mostly other small changes for clarification.

nbviewer [link](https://nbviewer.jupyter.org/github/pyro-ppl/pyro/blob/43a5f9e3b7057f066c42bcb301efdb1be34b576e/tutorial/source/bayesian_regression.ipynb)


tutorial/source/bayesian_regression.ipynb
```diff
@@ -92,333 +109,367 @@
...
     "          ylabel=\"log GDP (2000)\",\n",
     "          title=\"Non African Nations\")\n",
     "sns.scatterplot(african_nations[\"rugged\"], \n",
-    "            np.log(african_nations[\"rgdppc_2000\"]), \n",
-    "            ax=ax[1])\n",
+    "                african_nations[\"rgdppc_2000\"], \n",
+    "                ax=ax[1])\n",
     "ax[1].set(xlabel=\"Terrain Ruggedness Index\",\n",
     "          ylabel=\"log GDP (2000)\",\n",
-    "          title=\"Afri
```
fehiepsi: We don't need to register for unconstrained parameters. The original `nn.Parameter` also works.

fehiepsi: not important: I think you can also define `sigma` as PyroSample.

fehiepsi: Is it your intention to reuse the learned `linear_reg_model`?

fehiepsi: nit: `print(name, pyro.param(name).data)` will make the output cleaner

tutorial/source/bayesian_regression.ipynb
```diff
@@ -623,64 +678,194 @@
...
+   "metadata": {},
+   "source": []
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Model Serving via TorchScript\n",
+    "\n",
+    "Finally, note that the `model`, `guide` and the `predictive` utility class are all `torch.nn.Module` instances, and can be serialized as [TorchScript](https://pytorch.org/docs/stable/jit.html). \n",
```
fehiepsi: typo: `predictive` -> `Predictive`

tutorial/source/bayesian_regression.ipynb
fehiepsi: This comment is for the definition of `LinearRegression` class. I don't know why it displays a different position here.

fehiepsi: This is for the definition of `BayesianRegression` class.

fehiepsi: This comment is for the line `model = BayesianRegression(linear_reg_model)` in Using AutoGuide section.

fehiepsi: This is for the cell numbered 12.

martinjankowiak:
generally lgtm. comments:
- "Each time linear_module is called,"  => add "... the forward method..." ?
- you're using the data twice if you set the priors using the MLE estimate?
- `num_samples=1000` but not vectorized? isn't that super slow? incidentally, i thought we had `num_particles` everywhere....
- this isn't relevant to this tutorial but why aren't the returned tensors in `quantiles` detached?
- strange language: "Refer to the Model Serving section for an inside look at how you can roll one yourself"
- "Note that for lower values of ruggedness there are many more data points, and as such, lesser wiggle room for the line of best fit" this isn't right. the coefficients are global. the band gets thicker because weights are multiplied by inputs

tutorial/source/bayesian_regression.ipynb
neerajprad: You mean we can use `nn.Parameter` instead of `PyroParam`? That's true, but I wanted to write something that would work with constraints too. I'm happy to remove it if it seems confusing.

neerajprad: That was just to initialize the parameters to the ML estimates; maybe I should write a line to clarify that.

neerajprad: That's right, I didn't do that because we don't need to set it as an attribute, and I just wanted to show that we can call regular pyro primitives in the code too, but I should call this out probably.

neerajprad:
> "Each time linear_module is called," => add "... the forward method..." ?

Done.

> you're using the data twice if you set the priors using the MLE estimate?

Yeah, this is problematic. I wanted to mention that we could do a better initialization for the autoguide mean if needed, but this is using the dataset twice. It will be best to just initialize using some sensible priors instead.

> num_samples=1000 but not vectorized? isn't that super slow? incidentally, i thought we had num_particles everywhere....

I should remove the `num_samples` argument. Its from the earlier code, and is only used with `TracePredictive`. In any case, models with NNs and `.squeeze`s don't always work with our vectorization. 

> this isn't relevant to this tutorial but why aren't the returned tensors in quantiles detached?

Since `nn.Parameter` has `requires_grad=True`, we end up with tensors with `gradient_fn` whenever the guide is called. I think this would have been the case earlier too, but I'm looking into this separately.

> strange language: "Refer to the Model Serving section for an inside look at how you can roll one yourself"

Fixed.

> "Note that for lower values of ruggedness there are many more data points, and as such, lesser wiggle room for the line of best fit" this isn't right. the coefficients are global. the band gets thicker because weights are multiplied by inputs

Yes, that's plain wrong. Fixed.

tutorial/source/bayesian_regression.ipynb
neerajprad: Actually `PyroSample` won't work here.

neerajprad: Done.

tutorial/source/bayesian_regression.ipynb
```diff
@@ -16,7 +16,21 @@
...
+    "In this tutorial, we will first implement linear regression in PyTorch and learn point estimates for the parameters $w$ and $b$.  Then we will see how to incorporate uncertainty into our estimates by using Pyro to implement Bayesian regression. Additionally, we will learn how to use the Pyro's utility functions to do predictions and serve our model using `TorchScript`.\n",
+    "\n",
+    "## Tutorial Outline\n",
+    "\n",
+    " - [Setup](#Setup)\n",
+    "   - [Dataset](#Dataset)\n",
+    " - [Linear Regression](#Linear-Regression)\n",
+    "   - [Training with PyTorch Optimizers](#Training-with-PyTorch-Optimizers)\n",
+    "   - [Regression Fit](#Plotting-the-Regression-Fit)\n",
+    " - [Bayesian Regression with Pyro's SVI](#Bayesian-Regression-with-Pyro's-Stochastic-Variational-Inference-(SVI))\n",
```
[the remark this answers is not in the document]
neerajprad: Fixed.

neerajprad:
I completely forgot that now that we have `guide` as a module, we can simply turn off the `requires_grad` attribute to False on all parameters by doing `guide.requires_grad_(False)`. With that we don't end up computing `grad_fn` and the output is much cleaner.

tutorial/source/bayesian_regression.ipynb
```diff
@@ -92,337 +109,372 @@
...
     "          ylabel=\"log GDP (2000)\",\n",
     "          title=\"Non African Nations\")\n",
     "sns.scatterplot(african_nations[\"rugged\"], \n",
-    "            np.log(african_nations[\"rgdppc_2000\"]), \n",
-    "            ax=ax[1])\n",
+    "                african_nations[\"rgdppc_2000\"], \n",
+    "                ax=ax[1])\n",
     "ax[1].set(xlabel=\"Terrain Ruggedness Index\",\n",
     "          ylabel=\"log GDP (2000)\",\n",
-    "          title=\"Afri
```
fritzo: Are you sure this is necessary? I would have thought that we could simply
```py
class LinearRegression(nn.Linear, PyroModule):
    pass
```

neerajprad: I was using `PyroParam` earlier, but since we are using `nn.Parameter` now, I think we can remove this.

tutorial/source/bayesian_regression.ipynb
fehiepsi: I meant to define `self.sigma = PyroSample(dist.Uniform(0., 10.))` at the module construction. I think that it is a part of the Bayesian NN, so considering it an attribute of BayesianRegression module makes sense to me. Anyway, this is just a minor point to me. :)

neerajprad: That seems reasonable to me, will change.

neerajprad: Actually, one issue that I see with this pattern is that it is easy to forget that `self.sigma` is a special attribute and we need to access it in the correct context. The current model is:

```python
    def forward(self, x, y=None):
        sigma = pyro.sample("sigma", dist.Uniform(0, 10))
        mean = self.linear(x).squeeze(-1)
        with pyro.plate("data", x.shape[0]):
            obs = pyro.sample("obs", dist.Normal(mean, sigma), obs=y)
        return mean
```

if we change it to have `self.sigma = PyroSample(dist.Uniform(1, 10))` in `__init__` and change the forward method to the following, 

```python
    def forward(self, x, y=None):
        mean = self.linear(x).squeeze(-1)
        with pyro.plate("data", x.shape[0]):
            obs = pyro.sample("obs", dist.Normal(mean, self.sigma), obs=y)
        return mean
```
we'll silently do the unintended thing of actually sampling `sigma` that has a batch size equal to the data, since it is easy to forget that `self.sigma` needs to be accessed in the correct context, i.e. we'll need to add `sigma = self.sigma` outside the plate so as to trigger `getattr` outside the plate context. I'm happy to do this latter thing, but that will need a little bit more explaining in the notebook as to why we did that. I'll make the change based on feedback.

fritzo:
Looks great!

tutorial/source/bayesian_regression.ipynb
fehiepsi: Nice observation! This is easy to make mistakes. Thanks for pointing it out!

Issue #2161, opened 2019-11-14
Make PyroModule more usable and safe


fritzo:
This is a general issue to collect PRs related to `PyroModule`

PR #2164, opened 2019-11-15
Support syntax to create PyroModule mixin classes

fritzo:
Addresses #2161 

This makes it easier to turn various `nn.Module` subclasses into poutine-aware subclasses by mixing in `PyroModule` via syntax compatible with [PEP 484](https://www.python.org/dev/peps/pep-0484/):
```diff
- class LinearRegression(nn.Linear, PyroModule):
-     pass
- m = LinearRegression()
+ m = PyroModule[nn.Linear]
```
I like the possibility of this idiom, spelling out `PyroModule[...]` everywhere to make it clear that this is simply a modified version of another well-understood `nn.Module` subclass.

The two subtleties were:
- creating classes dynamically while memoizing their types
  (so that `PyroModule[nn.Linear] is PyroModule[nn.Linear]`), and
- supporting pickling

I have followed the convention that `PyroModule[MyModule].__name__ = PyroMyModule`, so e.g. `PyroModule[nn.Module] is PyroModule`.

## Tested
- added unit test
- tested serialization
- used `PyroModule[nn.Linear]` in the Bayesian regression tutorial

nbviewer [link](https://nbviewer.jupyter.org/github/pyro-ppl/pyro/blob/6c5d767/tutorial/source/bayesian_regression.ipynb)


martinjankowiak:
cool but shouldn't this have a more descriptive name than `Pyro`? i realize `PyroModule` is taken but perhaps there's something else?

fritzo:
I guess we need to trade off (1) having a short name so we can simpy write `Short(nn.Sequetial)` everywhere rather than making local definitions (with indirection) and (2) making this descriptive. Maybe we could call it `PyroMixinFactory` and encourage `import pyro.nn.PyroMixinFactory as Pyro`?

fritzo:
I mean, this code is delightfully readable:
```py
model = Pyro(nn.Sequential)(
    Pyro(nn.Linear)(28 * 28, 200),  # Bayesian
    nn.Sigmoid(),
    Pyro(nn.Linear)(200, 200),  # Bayesian
    nn.Sigmoid(),
    nn.Linear(200, 10),  # point estimate
)
```

martinjankowiak:
@fritzo yeah something like that sounds better to me 👍

martinjankowiak:
note the :-1: is for being "bayesian" about 28 * 28 * 200 weights ;)

fritzo:
What do you think of `pyronic(nn.Linear)`? We could even unify this with `@pyro_method` and decorate new methods as `@pyronic`?

martinjankowiak:
to my mind unification is a bit too automagical and may obscure important differences for users. i prefer something like encouraging uniform imports:`import pyro.nn.PyroMixinFactory as Pyro`

fritzo:
What do you think about this notation?
```py
model = PyroModule[nn.Sequential](
    PyroModule[nn.Linear](28 * 28, 200),  # Bayesian
    nn.Sigmoid(),
    PyroModule[nn.Linear](200, 200),  # Bayesian
    nn.Sigmoid(),
    nn.Linear(200, 10),  # point estimate
)
```
This also plays well with Python 3 type hinting ([PEP 544](https://www.python.org/dev/peps/pep-0544/)).

martinjankowiak:
that looks reasonable. what version of python does that require?

fritzo:
That works with all versions of Python. In newer Pythons it can also be used for type hinting, but that is just an added benefit.

I'll make that change and push, thanks for the feedback!

pyro/nn/module.py
```diff
@@ -77,7 +78,40 @@ def _get_pyro_params(module):
...
 
 
-class PyroModule(torch.nn.Module):
+# Inherit from ABCMeta because many practical nn.Module subclasses also inherit
+# from ABCMeta (e.g. EasyGuide), and this will avoid metclass conflict errors.
+class _PyroModuleMeta(ABCMeta):
+    _pyro_mixin_cache = {}
+
+    # Unpickling helper to create an empty object of type PyroModule[Module].
+    class _New:
```
neerajprad: I don't fully understand all the magic that happens here, but I have done a few tests and have been very satisfied with what I got. :)

Issue #2149, opened 2019-11-13
Problems with MeanFieldELBO [bug] with TransformedDistributions in guide and model


gamerDecathlete:
This is a copy of an issue I posted for pytorch: https://github.com/pytorch/pytorch/issues/29698

I think this is an issue with the _kl_independent_independent class which is summing over batches when it shouldn't...
### Guidelines

I think the Kl divergence is summing over the batch dimensions when it shouldn't, at least for the Gaussian case

Here's the transformed distribution function:
```py
def _kl_transformed_transformed(p, q):
    if p.transforms != q.transforms:
        raise NotImplementedError
    if p.event_shape != q.event_shape:
        raise NotImplementedError
    # extra_event_dim = len(p.event_shape) - len(p.base_dist.event_shape)
    extra_event_dim = len(p.event_shape)
    base_kl_divergence = kl_divergence(p.base_dist, q.base_dist) #call to indep_indep below
   #this will again sum over kl_divergence for each entry in batch
    return _sum_rightmost(base_kl_divergence, extra_event_dim)
```

Here's  independent_independent KL
```py
@register_kl(Independent, Independent)
def _kl_independent_independent(p, q):
    shared_ndims = min(p.reinterpreted_batch_ndims, q.reinterpreted_batch_ndims)
    p_ndims = p.reinterpreted_batch_ndims - shared_ndims
    q_ndims = q.reinterpreted_batch_ndims - shared_ndims
    p = Independent(p.base_dist, p_ndims) if p_ndims else p.base_dist
    q = Independent(q.base_dist, q_ndims) if q_ndims else q.base_dist
    kl = kl_divergence(p, q)
    if shared_ndims:
       #this line gets called when base_dist is Gaussian
        kl = sum_rightmost(kl, shared_ndims)
    return kl
```

## To Reproduce
I think something like this will do it:
```py
p = Gaussian(torch.zeros(3, 10), torch.ones(3, 10)).independent(1)
q = Gaussian(torch.ones(3,10) *2, torch.ones(3,10).independent(1)
kl = kl_divergence(p, q)
print(kl.shape) #should be something like [3, 1] but will instead output []
```
## expected behavior
Maybe I misunderstand kl_divergence function, but I don't think it should be summing over batch.

 - PyTorch Version (e.g., 1.0): 1.2
 - OS (e.g., Linux): Ubuntu 16.04
 - How you installed PyTorch (`conda`, `pip`, source): conda/ pip 
[...]

PR #2165, opened 2019-11-15
Fix shape of kl of 2 transformed distributions

fehiepsi:
Addresses #2149.

### DIFF
--- a/pyro/distributions/kl.py
+++ b/pyro/distributions/kl.py
@@ -50,7 +50,7 @@ def _kl_transformed_transformed(p, q):
         raise NotImplementedError
     if p.event_shape != q.event_shape:
         raise NotImplementedError
-    extra_event_dim = len(p.event_shape) - len(p.base_dist.event_shape)
+    extra_event_dim = len(p.base_dist.batch_shape) - len(p.batch_shape)
     base_kl_divergence = kl_divergence(p.base_dist, q.base_dist)
     return sum_rightmost(base_kl_divergence, extra_event_dim)
 
--- a/pyro/distributions/kl.py
+++ b/pyro/distributions/kl.py
@@ -1,6 +1,6 @@
 import math
 
-from torch.distributions import kl_divergence, register_kl
+from torch.distributions import TransformedDistribution, kl_divergence, register_kl
 
 from pyro.distributions.delta import Delta
 from pyro.distributions.distribution import Distribution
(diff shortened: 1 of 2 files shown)


pyro/distributions/kl.py
```diff
@@ -43,4 +43,16 @@ def _kl_independent_mvn(p, q):
...
 
 
+# TODO: move upstream
+@register_kl(TransformedDistribution, TransformedDistribution)
+def _kl_transformed_transformed(p, q):
+    if p.transforms != q.transforms:
+        raise NotImplementedError
+    if p.event_shape != q.event_shape:
+        raise NotImplementedError
+    extra_event_dim = len(p.event_shape) - len(p.base_dist.event_shape)
```
fritzo: Should this be `len(p.batch_shape) - len(p.base_dist.batch_shape)`? I seem to remember a `Transform` that changed `event_dim`, which I believe would introduce skew here.

fehiepsi: The upstream version `extra_event_dim = len(p.event_shape)` seems incorrect to me. However, I am not sure if this change will cover all the cases. Discussions are welcomed!

fehiepsi: Yeah, LKJ transforms a vector into a matrix. It is a bit subtle. I'll think more about this.

fehiepsi: Nice observation! Currently, transformed distributions assume that [`p.batch_shape + p.event_shape == p.base_dist.batch_shape + p.base_dist.event_shape`](https://github.com/pytorch/pytorch/blob/master/torch/distributions/transformed_distribution.py#L56). So I think that we can't construct a transformed distribution using a Transform that increases `event_dim`. Assuming that that issue is fixed, then using your formula would be better. I'll make the change. Thanks!

fritzo: Good observation, we should fix that upstream!

fehiepsi: It would be tricky. We'll need to provide more information to Transform class, e.g. to deal with the case:
```
d = dist.TransformedDistribution(dist.Normal(0, 1).expand((3,)), transforms.StickBreakingTransform())
d.event_shape == d.sample().shape  # 3 != 4
```
I think that users are less likely to use those transforms to create a transformed distribution. So I would prefer to skip it for now. :D

PR #2166, opened 2019-11-15
Raise FutureWarning for deprecated methods/classes

neerajprad:
Addresses #2116. Raises a louder FutureWarning which isn't silenced by default specially on versions 3.5, 3.6. 
- [x] FutureWarning added for:
   - [x] SVI.run
   - [x] TracePredictive
   - [x] mcmc.predictive
   - [x] random_module
 - [x] Updates to the bayesian_regression part ii tutorial to remove dependence on `SVI.run`.


tests/infer/test_abstract_infer.py
```diff
@@ -5,10 +5,10 @@
 import pyro.distributions as dist
 import pyro.optim as optim
 import pyro.poutine as poutine
-from pyro.infer.autoguide import AutoDelta, AutoDiagonalNormal, AutoLaplaceApproximation
-from pyro.infer import SVI, TracePredictive, Trace_ELBO
```
neerajprad: TracePredictive has been deprecated for the last two releases. All the functionality exists in `Predictive`, and these tests are covered there. As such, I thought it safe to remove them since we don't intend to make any changes to `TracePredictive` but instead remove it completely in the next release.

neerajprad:
@fritzo - what do you think about using [FutureWarning](https://docs.python.org/3/library/exceptions.html#FutureWarning) instead of the default `UserWarning` that we are using now?

.codecov.yml
```diff
@@ -7,4 +7,4 @@ coverage:
   round: nearest
   precision: 2
 
-comment: off
+comment: false
```
fritzo: 🎉

pyro/infer/svi.py
```diff
@@ -40,6 +40,10 @@ def __init__(self,
                  num_samples=10,
```
fritzo: Can we add a similar warning and default for `num_samples`? Maybe we should change both defaults to `None` or "deprecated".

pyro/infer/abstract_infer.py
```diff
@@ -289,8 +293,8 @@ def __init__(self, model, posterior, num_samples, keep_sites=None):
         self.num_samples = num_samples
         self.keep_sites = keep_sites
         super(TracePredictive, self).__init__()
-        warnings.warn('This class will be removed in the next release. Use the '
-                      '`pyro.infer.Predictive` class instead.', DeprecationWarning)
+        warnings.warn('This class is deprecated and will be removed in the next release. '
```
fritzo: Maybe we should say "in a future release" rather than "in the next" (here and elsewhere)?

pyro/infer/svi.py
neerajprad: Yes, good idea. Will fix.

neerajprad:
@eb8680, @martinjankowiak - I have modified some OED code as to prepare for a future deprecation of `.run` methods. Does that seem okay to you?

pyro/infer/predictive.py
```diff
@@ -63,7 +63,7 @@ def _predictive(model, posterior_samples, num_samples, return_sites=(),
         return trace
 
     return_site_shapes = {}
-    for site in model_trace.stochastic_nodes + model_trace.observation_nodes:
+    for site in model_trace.stochastic_nodes:
```
neerajprad: @fehiepsi - To save travis runs for minor changes, I included this here. The motivation for this is that when we pass in observations to `Predictive` we shouldn't record `observation_nodes`. Does that seem reasonable to you (I forget the motivation for this, but it doesn't seem like we should need this now)?

neerajprad: This is also exercised in `bayesian_regression_ii.ipynb`.

fehiepsi: This is at least useful for your deterministic `dist.Delta` trick. ;)

neerajprad: Ahh..that's a good trick. Let us merge this if tests pass, and I'll address this in the other PR.

neerajprad: okay, reverted.

PR #2168, opened 2019-11-15
Fix SMCFilter, change interface

### LINKED ISSUE #2138: Resampling bug in SMCFilter?
@tuananhle7 [points out](https://github.com/pyro-ppl/pyro/pull/1947#issuecomment-551885868)

> I was looking at [this example](https://github.com/pyro-ppl/pyro/blob/dev/examples/smcfilter.py) and the [source code](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/smcfilter.py) in order to understand the implementation.

> I don't understand how resampling works:

> Assume that at time `t - 1`, we have particles `{z_(t - 1)k}` and weights `{w_(t - 1)k}` (`k = 1, ..., K`) which I think (?) are stored in [`_values`](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/smcfilter.py#L42) and [`_log_weights`](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/smcfilter.py#L43).

> The [resampling step](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/smcfilter.py#L73) produces ancestral indices `a_(t - 1)k` (`k = 1, ..., K`) which are marginally distributed according to a Categorical distribution parameterized by the normalized weights ([here](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/smcfilter.py#L123) I think). Then `z_(t - 1)k` which is stored in `_values` is updated to be `z_(t - 1)(a_(t - 1)k)` [here](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/smcfilter.py#L124).

> At time `t`, we should propose `z_tk` from `guide(z_t(k - 1))` where `z_t(k - 1)` is already resampled. However, [here](https://github.com/pyro-ppl/pyro/blob/dev/examples/smcfilter.py#L60-L61), `z_tk` is proposed from a guide which is conditioned on `self.z` (attribute of guide) which doesn't seem to interact with `_values` which stores the resampled values.

cc @chanjed @karalets

fritzo:
Fixes #2138 

This changes the `SCMFilter` interface to require `model` and `guide` to store any random state in a `state` dict that is managed by the `SMCFilter` class. As part of this refactoring, users now get more control over what state is stored across iterations, thereby fixing the quadratic growth issue that plagued our original implementation.

This isn't the slickest possible interface, but it is clear, easy to understand, and forwards compatible with the possibility of moving that `state` dict into a `SMCModel` class with `nn.Module`-like `getattr` access.

## Tested
- updated existing tests
- added a stronger test against `GaussianHMM`
- verified sane results from the example, increase steps from 50 -> 500

### DIFF
--- a/examples/smcfilter.py
+++ b/examples/smcfilter.py
@@ -50,13 +50,13 @@ class SimpleHarmonicModel_Guide:
 
     def init(self, state, initial):
         self.t = 0
-        state["z"] = pyro.sample("z_init", dist.Delta(initial, event_dim=1))
+        pyro.sample("z_init", dist.Delta(initial, event_dim=1))
 
     def step(self, state, y=None):
         self.t += 1
 
         # Proposal distribution
-        state["z"] = pyro.sample(
+        pyro.sample(
             "z_{}".format(self.t),
             dist.Normal(state["z"].matmul(self.model.A), torch.tensor([1., 1.])).to_event(1))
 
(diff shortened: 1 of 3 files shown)


fritzo:
@martinjankowiak ready to merge? I'll add stronger [GaussianHMM tests](https://github.com/pyro-ppl/pyro/blob/fix-smc-resample-2-test/tests/infer/test_smcfilter.py#L177) later, they appear finicky.

PR #2170, opened 2019-11-16
Avoid PyroMeta inheriting from ABCMeta

fritzo:
Addresses #2161 

This reverts an inheritance change from #2164 so that `PyroModule`s (and hence `AutoGuide` and many other derived classes) do not inherit from `ABCMeta`. Previously I thought it would be preferable to make `PyroModule[m]` work for classes `m` with metaclass deriving from `ABCMeta` (e.g. `EasyGuide`). Now I believe that will cause more issues than simply working around the one `EasyGuide` case and later adding auto-metaclass logic (with which this PR is forward compatible).

I've also tried to make the `PyroModule[-]` logic easier to understand by creating a local class.

## Tested

### DIFF
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -10,7 +10,6 @@ effects.  To create a poutine-aware attribute, use either the
 
 """
 import functools
-from abc import ABCMeta
 from collections import OrderedDict, namedtuple
 
 import torch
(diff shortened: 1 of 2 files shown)


pyro/nn/module.py
```diff
@@ -98,15 +95,13 @@ def __getitem__(cls, Module):
...
+            # Unpickling helper to load an object of type PyroModule[Module].
+            def __reduce__(self):
+                state = getattr(self, '__getstate__', self.__dict__.copy)()
+                return _PyroModuleMeta._New, (Module,), state
 
-        name = "Pyro" + Module.__name__
-        bases = (Module, PyroModule)
-        dct = {"__reduce__": __reduce__}
-        result = type(name, bases, dct)
+        result.__name__ = "Pyro" + Module.__name__
```
neerajprad: Thanks for simplifying!

PR #2176, opened 2019-11-16
Remove torch.einsum patch

fritzo:
The patch has dwindled to a no-op, and can now be removed. Simon Wang also complained this was hacky.


neerajprad:
I think travis is going to be busy for a while, so feel free to push this to the 1.0.0 branch directly.

fritzo:
:+1: I'll close and push this into 1.0

PR #2177, opened 2019-11-17
Return namedtuple from .call()

fritzo:
Replaces the `tuple` return from `AutoGuide.call()` and `Predictive.call()` with a dynamically-defined `namedtuple` object.

## Backwards compatibility

- the returned value is still a subclass of `tuple`
- the ordering is still sorted alphabetically by key

## Tested
- refactoring and serialization is exercised by autoguide tests
- added a unit test

### DIFF
--- a/pyro/infer/util.py
+++ b/pyro/infer/util.py
@@ -1,6 +1,6 @@
 import math
 import numbers
-from collections import Counter, defaultdict
+from collections import Counter, defaultdict, namedtuple
 
 import torch
 from opt_einsum import shared_intermediates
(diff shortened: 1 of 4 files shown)


pyro/infer/util.py
```diff
@@ -32,6 +32,33 @@ def torch_item(x):
     return x if isinstance(x, numbers.Number) else x.item()
 
 
+def dict_to_namedtuple(dct):
+    """
+    Converts a dictionary to a namedtuple.
+    """
+    assert isinstance(dct, dict)
+    dct = {key if key.isidentifier() and not key.startswith("_") else
```
fritzo: This seems unsafe. Any suggestions?

neerajprad: I think this should work for all `PyroModule`s? For this to work with arbitrary names like those from `pyro.module` (which someone may use with `Predictive`), we could use the hash of the str as the key suffix if it is not a valid identifier. We can also cache metadata to map the actual key names to the ones used in the namedtuple and share it with `namedtuple_to_dict` to get the dict with the original keys.

neerajprad: On second thoughts, caching is best avoided and won't work with separate sessions. Maybe something like below might work?

```python
def make_key(keys):
    keys = sorted(keys)
    exists_counter = Counter()
    id_keys = []
    for key in keys:
        # replace nonidentifier characters with '' followed by all leading non alphanumeric chars
        id_key = re.sub(r'^[^a-zA-z_]*', '', re.sub(r'[^a-zA-Z0-9_]', '', key))
        cnt = exists_counter[id_key]
        exists_counter[id_key] += 1
        if cnt:
            id_key = id_key + '_{}'.format(cnt)
        id_keys.append(id_key)
    return id_keys
```

fritzo: I think we'll want an invertible mapping, some sort of `encode`,`decode` pair for python identifiers that do not begin with `_`.

fritzo:
Closing as this appears infeasible.

PR #2178, opened 2019-11-17
Implement EnergyDistance likelihood-free inference algorithm

fritzo:
This implements an `SVI`-compatible loss for posterior predictive energy distance, a multidimensional generalization of CRPS. This can be computed as long as `num_particles >= 2`, and has cost quadratic in number of particles. From the docs:

image

The motivation is to perform likelihood-free inference in models with heavy-tailed likelihoods such as the [Levy stable distribution](https://en.wikipedia.org/wiki/Stable_distribution). In these models, the posterior predictive may have infinite variance, but energy distance can nevertheless be Monte Carlo estimated.

This implementation allows an optional Bayesian regularization by including a `-log p(z)` term in the loss. Note that many likelihood-free approaches approximate only the likelihood
```
# typical likelihood-free loss (not energy distance):
z ~ q(z|x)                    # posterior
x' ~ p(x|z)                   # posterior predictive
loss = distance(x', data)     # generalized likelihood
     + log q(z|x) - log p(z)  # KL(q,p)
```
By contrast, the energy distance includes a generalized entropy term, and approximates both the likelihood and entropy(guide), hence regularization requires only a `log p(z)` term.
```
# CRPS loss:
z ~ q(z|x)                           # posterior
x' ~ p(x|z)                          # posterior predictive
loss = energy_dist(x',x) - log p(z)
     = |x-x'|                        # generalized likelihood
     - 1/2 |x'-copy(x')| - log p(z)  # generalized KL(q,p)
```

 The approach is to:
- Include loss terms for both energy distance and optionally `-log p(z)`.
- Restrict to reparametrized guides and likelihoods (checked when `is_validation_enabled`). We could relax this in a follow-up PR by using DiCE-weighted particles.

I plan to add an example in a future PR #2198 ([notebook](https://nbviewer.jupyter.org/github/pyro-ppl/pyro/blob/trace-crps-example/tutorial/source/likelihood_free.ipynb))

## Tested
- [x] test_valid_models.py
  - including a test with `dist.Stable` for which `.log_prob()` is undefined.
- [x] test univariate convergence
- [x] test multivariate convergence
- [x] tested on a real dataset:
  I tested on a large multivariate time series forecasting problem with heavy-tailed likelihoods (power-folded-student-T) and found:
  - Each svi step was ~3x slower and used ~2x more memory than `Trace_ELBO`. Note much of this slowdown is due to the slow `Gamma.rsample()` method used by the likelihood.
  - But convergence was more more robust and took fewer iterations (after I worked around NANs in `StudentT.rsample()`, I believe due to flaky `_standard_gamma()`). Overall time was comparable to `Trace_ELBO`.
  - Convergence appears to be roughly exponential in this problem (down to the noise floor):
    image
  - The final fit looked great, but uncertainty was underestimated, just as it was with `Trace_ELBO`. This may simply suggest a poor model fit.


pyro/infer/trace_crps.py
```diff
@@ -0,0 +1,94 @@
...
+    This is a non-Bayesian, likelihood-free method; no densities are evaluated.
+    """
+    def __init__(self,
+                 num_particles=2,
+                 max_plate_nesting=float('inf')):
+        assert num_particles >= 2
+        self.max_plate_nesting = max_plate_nesting
+        self.vectorize_particles = True
+
+    def _get_samples(self, model, guide, *args, **kwargs):
```
fritzo: @eb8680 thanks for explaining how to draw posterior predictive samples on Friday! Can you review my use of `uncondition` here? I'm especially unsure whether `tr2` will record the observations.

eb8680: Hmm, I don't think so, `tr2` will end up with samples rather than the observation here.

pyro/infer/svi.py
```diff
@@ -66,7 +66,8 @@ def __init__(self,
             if loss_and_grads is None:
                 def _loss_and_grads(*args, **kwargs):
                     loss_val = loss(*args, **kwargs)
-                    loss_val.backward(retain_graph=True)
+                    if getattr(loss_val, 'requires_grad', False):
```
fritzo: This ensures match with logic in the `ELBO` case.

pyro/poutine/uncondition_messenger.py
```diff
@@ -19,6 +19,7 @@ def _pyro_sample(self, msg):
         if msg["is_observed"]:
             msg["is_observed"] = False
             msg["infer"]["was_observed"] = True
+            msg["infer"]["obs"] = msg["value"]
```
fritzo: @eb8680 I was able to get things working with this change to `uncondition`. Is it reasonable?

eb8680: Looks fine to me

pyro/infer/trace_crps.py
```diff
@@ -0,0 +1,176 @@
...
+    return scale_and_mask(error, scale, mask)
+
+
+class Trace_CRPS:
+    """
+    Posterior predictive CRPS loss with optional ``KL(q,p)`` regularization.
+
+    This is a likelihood-free method, and can be used for likelihoods without
+    tractible density functions. CRPS is a robust loss function, and is well
+    defined for any distribution with finite absolute moment ``E[|data|]``.
```
martinjankowiak: data or latent?

pyro/infer/trace_crps.py
```diff
@@ -0,0 +1,176 @@
...
+
+    This requires static model structure, fully reparametrized guide, and
```
martinjankowiak: fully => a fully

pyro/infer/trace_crps.py
```diff
@@ -0,0 +1,176 @@
...
+    [1] `Strictly Proper Scoring Rules, Prediction, and Estimation`
+    Tilmann Gneiting, Adrian E. Raftery (2007)
+    https://www.stat.washington.edu/raftery/Research/PDF/Gneiting2007jasa.pdf
+
+    :param int num_particles: The number of particles/samples used to form the
+        gradient estimators. Must be at least 2.
+    :param int max_plate_nesting: Optional bound on max number of nested
+        :func:`pyro.plate` contexts. If omitted, ELBO may guess a valid value
+        by running the (model,guide) pair once, however this guess may be
+        incorrect if model or guide structure is dynamic.
```
martinjankowiak: although we're disallowing dynamic anyway?

pyro/infer/trace_crps.py
fritzo: data; that is the space where we compute CRPS.

fritzo:
@martinjankowiak ready for final review. After this and #2181 merge I'll create an example combining them.

fritzo:
I'm not entirely satisfied with the form of prior regularization here. Perhaps we should instead use an energy-score prior term here as well, something like
```
energy_dist(q(z), p(z)) = E_{Z~q,Z'~p} ||Z-Z'|| - 1/2 E_{Z,Z'~q} ||Z-Z'|| - 1/2 E_{Z,Z'~p} ||Z-Z'||
```
but that would have the disadvantage of being scale-sensitive to latent variables (I like that the `CRPS - log p(z)` form is scale-sensitive to observed variables but is scale invariant for latents).

PR #2181, opened 2019-11-21
Implement Levy Stable distribution

fritzo:
This implements the four-parameter [Stable distribution](https://en.wikipedia.org/wiki/Stable_distribution), a flexible model of heavy-tailed data.

This distribution is unusual in that its density function is intractable (implementations often use FFTs to compute an integral, for each datapoint!). However a reparametrized sampler can be implemented by modifying the classic Chambers-Mallows-Stuck sampler (which generalizes the usual Box-Muller sampler for Normal distributions). The modifications are:
1. Use Nolan's `S0` parametrization, so that samples are continuous in the (alpha,beta) parameters (note the more common `S1` parametrization is discontinuous at the midpoint `alpha=1` 😔).
2. Patch the hole at `alpha=1`. I have avoided deriving a Pade approximation or similar, and have instead linearly interpolated between rsamples on either side of the hole.

This distribution is intended for the likelihood-free inference algorithm `Trace_CRPS` #2178.

## Tested
This adds around 12sec of new tests.

- [x] added to generic tests (though most xfail due to non-implemented `.log_prob()`)
- [x] shape tests
- [x] test density against [scipy.stats.levy_stable.cdf](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.levy_stable.html#scipy.stats.levy_stable) (which is finicky and slow, needing to perform numerical quadrature at each sample value).


tests/distributions/test_stable.py
```diff
@@ -0,0 +1,57 @@
...
+    assert d.batch_shape == batch_shape
+
+    x = d.rsample(sample_shape)
+    assert x.shape == sample_shape + batch_shape
+
+    x.sum().backward()
+
+
+@pytest.mark.parametrize("beta", [-0.6, -0.3, 0.0, 0.3, 0.6])
+@pytest.mark.parametrize("alpha", [0.2, 0.5, 0.8, 0.99, 1.0, 1.01, 1.2, 1.5, 1.8])
```
fehiepsi: is it stable for cases around boundaries of `alpha` and `beta`, e.g. `alpha=0.01 or 2`, `beta=-1 or 1`?

fritzo: the scipy reference implementation seems to break down in edge cases.

fritzo:
> Re GP: ... relationship between Stable distribution and GP likelihoods ... ?

If we had a `GaussianHMM.rsmaple()` #2182, then we could train a model with GP latent state and Stable likelihood using `Trace_CRPS` #2178.

martinjankowiak:
nice! i wont have time for any CR till Monday so feel free to merge with @fehiepsi's approval

tests/distributions/test_stable.py
fritzo: Oh, I have added more edge cases! Note that scipy breaks down at `alpha=0.995` and `alpha=1.005`, so I have instead temporarily increased `RADIUS` so that interpolation logic is checked.

fritzo:
Thanks for review @fehiepsi!

tests/distributions/test_stable.py
fehiepsi: Thanks for explaining!

PR #2183, opened 2019-11-22
Port BART preprocessing script from sandbox to Pyro repo

fritzo:
Original source:
https://github.com/pyro-ppl/sandbox/blob/master/2019-08-time-series/bart/preprocess.py

I have included caching logic and @neerajprad has hosted the final compressed 63MB .pkl.bz2 file, which downloads and uncompresses quickly. Preprocessing from scratch takes ~200MB of download bandwidth, a few GB of temp space and a couple processor hours, but can serve as a backup in case the cache is down.

I have also added pyro.contrib.examples to our documentation.

The motivation is to enable lots of new multivariate time series examples and tutorials.

## Tested
- ran locally

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -32,6 +32,7 @@ Pyro Documentation
    contrib.autoname
    contrib.bnn
    contrib.easyguide
+   contrib.examples
    contrib.gp
    contrib.minipyro
    contrib.oed
--- a/pyro/contrib/examples/bart.py
+++ b/pyro/contrib/examples/bart.py
@@ -1,0 +1,159 @@
+import argparse
+import csv
+import datetime
+import gzip
+import logging
+import multiprocessing
+import os
+import subprocess
+import sys
+import urllib
+
+import torch
+
+from pyro.contrib.examples.util import get_data_directory
+
+DATA = get_data_directory(__file__)
+
+# https://www.bart.gov/about/reports/ridership
+SOURCE_DIR = "http://64.111.127.166/origin-destination/"
+SOURCE_FILES = [
+    "date-hour-soo-dest-2011.csv.gz",
+    "date-hour-soo-dest-2012.csv.gz",
+    "date-hour-soo-dest-2013.csv.gz",
+    "date-hour-soo-dest-2014.csv.gz",
+    "date-hour-soo-dest-2015.csv.gz",
+    "date-hour-soo-dest-2016.csv.gz",
+    "date-hour-soo-dest-2017.csv.gz",
+    "date-hour-soo-dest-2018.csv.gz",
+]
+
+
+def _load_hourly_od(basename):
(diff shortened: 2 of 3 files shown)


pyro/contrib/examples/bart.py
```diff
@@ -0,0 +1,159 @@
...
+import os
+import subprocess
+import sys
+import urllib
+
+import torch
+
+from pyro.contrib.examples.util import get_data_directory
+
+DATA = get_data_directory(__file__)
```
fritzo: I'm not sure what is the safest path to use here. I think it would be great if we followed the convention that all data was stored in `$REPO_ROOT/.data/` so that `$REPO_ROOT/examples` and `$REPO_ROOT/tutorials` could share `.data/`, but I believe that would require some refactoring.

docs/source/index.rst
```diff
@@ -32,6 +32,7 @@ Pyro Documentation
    contrib.autoname
    contrib.bnn
    contrib.easyguide
+   contrib.examples
```
neerajprad: Do you intend for `contrib.examples` to house utilities for examples? Maybe we could also just move our examples module here too (not in this PR).

pyro/contrib/examples/bart.py
neerajprad: I agree, that will be ideal. Right now, all folders in examples have their own `.data` directory. But it is not too bad, since most of the tutorials/examples don't share that much data, and those that do can probably be organized into a single folder. Some of the examples also use `.data` to dump their outputs. I think the only organizing principle we have right now is that for an example we should only have to download the dataset once, and it shouldn't depend on where the example is run from.

neerajprad:
I am trying to run this and generate the `.pkl` files.

docs/source/index.rst
fritzo: I think it's ok for examples to live outside of the repo, and nice for some data downloaders to live in the repo. Often R packages are organized that way, I believe.

neerajprad: Oh..I didn't mean the examples directory. I prefer that being outside the main package. I thought we had a separate `pyro.examples` module, but I was mistaken - it was already in `pyro.contrib`.

pyro/contrib/examples/bart.py
```diff
@@ -0,0 +1,179 @@
...
+    Load a dataset of hourly origin-destination ridership counts for every pair
+    of BART stations during the years 2011-2018.
+
+    **Source** https://www.bart.gov/about/reports/ridership
+
+    This downloads and preprocesses the dataset the first time it is called,
+    requiring about 300MB of file transfer and storing a few GB of temp files.
+    On subsequent calls this reads from a cached ``.pkl.bz2``.
+
+    :param bool force: whether to force download and recomputation
```
neerajprad: Can we remove the force arg now (it doesn't seem to be used)?

PR #2184, opened 2019-11-22
Adding a tutorial on Dirichlet process mixture modeling

m-k-S:
Hi, I wrote a tutorial on Dirichlet process mixture models in Pyro, located [here](https://github.com/m-k-S/pyro/blob/dev/tutorial/source/dirichlet_process_mixture.ipynb). It contains an implementation of a Dirichlet process Gaussian mixture as well as a Dirichlet process Poisson mixture. It also contains a small section on model criticism in Pyro using the log predictive distribution and [Posterior Dispersion Indices](https://arxiv.org/pdf/1605.07604.pdf) (Kucukelbir, 2016). 

I would appreciate any thoughts, suggestions, and/or criticism!


fehiepsi:
Thank @m-k-S, the content looks really nice! I am happy to take the first pass for this tutorial.

fehiepsi:
@m-k-S For some reason, your notebook raw is displayed in 1 single line, which makes it hard to comment for specific codes. Could you save your notebook in the usual way? Here are my comments regarding code cells:
+ I think you missed `import torch`
+ There are some 2-space indentations in the code. 4-space is prefered.
+ `from tqdm.notebook import tqdm` -> `from tqdm import tqdm` (notebook widgets are not rendered well)
+ Bayes_Centers_05 -> Bayes_Centers_01
+ Is using "from statsmodels.graphics.tsaplots import plot_acf" necessary? I think you can replace it with
```
autocorr_ax.acorr(np.asarray(losses), detrend=lambda x: x - x.mean(), maxlags=750, usevlines=False, marker=',')
autocorr_ax.set_xlim(0, 750)
autocorr_ax.axhline(0, ls="--", c="k", lw=1)
```
+ Some speed improvements (not important):
     - `mixed_weight` implementation can be made faster with
```python

def mix_weights(beta):
    beta1m_cumprod = (1 - beta).cumprod(-1)
    return F.pad(beta, (0, 1), value=1) * F.pad(beta1m_cumprod, (1, 0), value=1)
```
 +
    - use `pyro.param("p", lambda: ..., constraint)` (the `lambda` expression will only be executed once when creating the first value of that param `p`)
+ I don't understand what is computed in `log_posterior_predictive_eval`. Could you explain it? In particular, what is the role of `torch.exp(Poisson(...).log_prob(...))` in `log_prob` computation? And do the weights `post_pred_weights` play any role here? I guess you want to use `mixture_of_poisson` here...

About explanations, here are my opinions. Feel free to make adjustment as you like:
+ The title "Bayesian Nonparametrics in Pyro" is not specific. There are other nonparametric models in Pyro such as Gaussian Process.
+ Definition of Dirichlet Process:
    - what is G(Omega_1)? I think it is better to have a sentence say that G is also a "discrete" probability distribution over Omega.
+ Definition of Chinese Restaurant Process:
    - missing definition of Omega: probably add "If we associate to each table 𝑡 a draw from a base distribution 𝐺0 **over Omega**"?
    - probability mass `𝑛𝑡` -> unnormalized probability mass `nt`
+ The context for the definition of "The variational family 𝑞" is missing. Probably it is better to mention it after you have defined the `model`?
+ ELBO Behavior:
    - The plot of autocorrelation is computed after "detrending" (see above). I am not sure if it is a proper way to justify that "it takes around 500 iterations to achieve convergence". Could you provide some references?
+ Dirichlet Mixture Model for Long Term Solar Observations:
    - It would be clearer to have comments mention that Bayes_Rates and Bayes_Weights are the posterior means of `beta` and `lambda`.
+ Criticism:
     - Could you add some references for `"Mean Predictive"` and `Posterior Dispersion Indices` and how to interpret those values?
+ Because the tutorial does inference for the "truncated" DPMM, it would be nice if there is a paragraph mentioning practical usages of "truncated" DPMM. I would love to know if "truncated" is a recommended way in practice. :)

m-k-S:
Hi @fehiepsi,

Thank you so much for the detailed feedback. I have implemented almost all of the changes you suggested above. 

Regarding the `log_posterior_predictive_eval`: as per [Variational inference for Dirichlet process mixtures](http://www.cs.columbia.edu/~blei/papers/BleiJordan2004.pdf), the posterior predictive for the Dirichlet process mixture is exactly equal to the scale-mixture of exponential family distributions, where the weights are the mixture weights given by the Bayes estimators of the 'stick breaking' latent variables, and the cluster parameters are those given by the Bayes estimators. I did indeed forget to multiply each probability by the corresponding `post_pred_weight` term, so thank you for catching that. I corrected this code and re-ran it, which you can see in the updated notebook.

Regarding Dirichlet processes: G is a probability distribution drawn from DP(alpha, G_0) –– in reality, it is a continuous distribution. Therefore, when we take a disjoint union Omega_1, ... Omega_k of our support Omega, G(Omega_i) is the probability measure of Omega_i under this new distribution. The key property of DPs is that the "vector" (G(Omega_1), ..., G(Omega_k)) is equivalent to a draw from the Dirichlet distribution Dir(alpha * G_0(Omega_1), ..., alpha * G_0(Omega_k)), where G_0(Omega_i) is the probability measure of Omega_i under G_0.

Regarding truncation: this is a bit of a nuanced point which I think is difficult to convey. I'll attempt to give my (probably limited) understanding of it: a typical mixture model (let's use GMMs just for an example) with a Dirichlet **distribution** prior on the weights is a priori assigning equivalent "importance" to each cluster; in other words, before we even begin to perform inference on a GMM, we first must fix a number $K$ of the clusters that we think exist in the data, and then during posterior computation the weights for each of the $K$ clusters are adjusted according only to the number of points assigned to each cluster.

In contrast, when we use a Dirichlet **process* prior on our weights, we are (in theory) adding the number of clusters present in the data to our model. However, in the stick-breaking formulation of DPMMs, this would require infinitely many samples, and so practically for computation, we must assign a truncation point T. In this case, though, T represents not how many clusters we think exist in our data, but rather the assertion, "there are definitely no more than T clusters in our data". In other words, `mix_weights(beta)[T] == 1`. The Dirichlet process prior thus tends to yield mixture weights that are *much* more heavily skewed towards the left (depending on your setting for alpha), which means that if T was far greater than the "ground truth" number of clusters K in the data, there will (in theory) be approximately T-K clusters with negligible weight, which we can essentially ignore.

This might be clearer if you look at the Dirichlet process Poisson mixture I compute in my notebook - T is set to 20 (I am asserting that there are no more than 20 clusters in the data) but practically, only 7 clusters have non-negligible weight. If I had trained a regular Poisson mixture model on this data with T set to 20, it would've attempted to fit 20 clusters to the data, which is an incorrect model assumption.

Hopefully this helps! Please let me know if anything is unclear.

Perhaps you could let me know if this makes sense? I can add

tutorial/source/dirichlet_process_mixture.ipynb
```diff
@@ -0,0 +1,790 @@
+{
```
fehiepsi: You would need to add this tutorial to the Contributed section of the file https://github.com/pyro-ppl/pyro/blob/dev/tutorial/source/index.rst

tutorial/source/dirichlet_process_mixture.ipynb
```diff
@@ -0,0 +1,790 @@
...
+        "import numpy as np\n",
+        "import matplotlib.pyplot as plt\n",
+        "import matplotlib.cm as cm\n",
+        "from tqdm import tqdm\n",
+        "\n",
+        "from pyro.distributions import *\n",
+        "import pyro\n",
+        "from pyro.optim import Adam\n",
+        "from pyro.infer import SVI, Trace_ELBO\n",
+        "from torch.distributions import constraints"
```
fehiepsi: Could you enable validation and set random seed here?
```
assert pyro.__version__.startswith('1.0.0')
pyro.enable_validation(True)       # can help with debugging
pyro.set_rng_seed(0)
```
It would be clearer if you move `from torch.distributions import constraints` to the above import block.

m-k-S:
Hi @fehiepsi,

Hope you had a good Thanksgiving, and thank you for another round of thoughtful suggestions!

Could you briefly explain how Pyro computes the mean and variance of the posterior predictive? It seems to me that in most cases, the posterior predictive has no closed form and thus the way I was taught to compute expectations of functions with respect to the posterior predictive was through Monte Carlo integration (as I had implemented initially). 

The "mean predictive" is just the mean value of the posterior predictive evaluated at each point in the holdout set; as far as I know, it is essentially a measure of how well your model does at 'explaining' points from the population distribution that it hasn't seen before. We would expect that since the train subset and the test subset came from the same initial set of observations, the posterior predictive would take high values at these held out points.

The posterior dispersion index (again, to the best of my knowledge) is the variance with respect to the posterior of the posterior predictive (evaluated at a test point x'), divided by the expectation with respect to the posterior of the posterior predictive (evaluated at the same test point x'). This essentially lets you discover the datapoints "whose likelihoods are changing rapidly at the posterior mean estimate of the latent variables", which implies that the model has low 'confidence' in whatever predictive value it is assigning to said point. Since computation of these PDIs requires expectations with respect to the posterior, I performed Monte Carlo integration by taking S=100 samples from the variational family and evaluating the posterior predictive generated by each of these 100 samples at every datapoint in the holdout set. The only point with a significantly high PDI is on the right tail of the empirical test distribution (which makes sense visually and intuitively) and I am pretty confident that these results are correct. Fwiw, everything I know about this I received secondhand from the author of the PDI paper (Alp Kucukelbir), so perhaps I am missing nuances. As far as I understand, though, the PDI is not meant for comparing two models from different classes (eg, comparing nonparametric regression to linear regression) but rather comparing different settings of the latent variables in the same model class (in this case, the DPMM) [so this is for model criticism rather than model comparison, re: your last question]

fehiepsi:
Thanks, you too! ;)

Yeah, the discussion is going to be more interesting. Let's discuss further those points regarding log predictive.

> how Pyro computes the mean and variance of the posterior predictive? ... expectations of functions with respect to the posterior predictive was through Monte Carlo integration

Yup, we compute this way, for example in [baseball example](https://github.com/pyro-ppl/numpyro/blob/master/examples/baseball.py#L152). The difference is we use the *same* set of posterior samples for each testing data point, while you use a *different* set of posterior samples for each testing data point and for each computation (mean/var/...). Probably this is not important but the former is more applicable in MCMC context (where we only have 1 set of posterior samples) and makes the computation easier to vectorize. It is also what is implemented in Algorithm 1 (section 2.4) of the PDI paper.

> log mean predictive

IIUC then your implementation of `log_posterior_predictive_eval` computes `mu(n)` in formula (1) of PSI paper. To get log mu(n), I think you would need to take `log` of the output. Note that log of the expectation of probabilities can be implemented as log_mean_exp(log probabilities) for numerical stability. (in my previous comment, `mean_log_prob = log_prob.logsumexp(-1) - math.log(S)` does this purpose: compute log_mu(n) for each n). Anyway, if you only want to compute `mu(n)`, then we just need to rename `log_posterior_predictive_eval` to `posterior_predictive_eval`?

> mean predictive

In the PDI paper, I just found `mu(n)`, `mu_log(n)`, and I know that the sum of log of those `mu(n)`s is log pointwise predictive density, which is used to compute information criterions for model selection purpose. But none of those is "mean predictive". This is probably a new terminology (at least to me) so it would be nice to provide some context in the tutorial so users can follow up and interpret that value. Anyway, I found that the `Mean Predictive:  0.9945551` is quite surprised. Why do you think that this is a reasonable result? I have not yet seen this good performance when "Bayesian" is involved. :D

> the variance with respect to the posterior of the posterior predictive (evaluated at a test point x'), 

I think the paper PDI causes a bit of confusion regarding this aspect. The introduction talks about PDI but the remaining content introduces `WAPDI` (which is what is implemented in my previous comment). If you want to compute PDI, I guess given `log_prob` (with shape num_data x num_posterior_samples):
```
prob = log_prob.exp()
# or we can compute probability directly (logsumexp trick is just more stable when computing log_prob)
# prob = (post_pred_weights * Poisson(post_pred_clusters).log_prob(data_test.reshape(-1, 1, 1)).exp()).sum(-1)
mean_prob = prob.mean(-1)
var_prob = prob.var(-1)
```
I think by PDI, the authors mean WAPDI (as reflected in the Algorithm 1 of [published paper](http://proceedings.mlr.press/v70/kucukelbir17a/kucukelbir17a.pdf)) and the introduction just provides some intuition for WAPDI.

m-k-S:
Ah, I see what is unclear now, and I don't have an answer off the top of my head. I am meeting with the authors later this week and will try to get clarification then. Thanks again!

fehiepsi:
Yeah, that's great to have some clarifications from original authors! Waiting to hear about it...

m-k-S:
Hi @fehiepsi,

Sorry for the delay. 

Would you mind explaining a bit more what the `Predictive` class is doing? When I compute the mean log predictive using the Predictive class and your code on the _training set_, I get a value of around -1500, which is extremely low given that I am evaluating on the same data that is used to train. Here is the code I used:

```
# Using pyro's built in posterior predictive class:
posterior = Predictive(guide, num_samples=S, return_sites=["beta", "lambda"])(data)
post_pred_weights = mix_weights(posterior["beta"])
post_pred_clusters = posterior["lambda"]
# log_prob shape = N_test x S
log_prob = (post_pred_weights.log() + Poisson(post_pred_clusters).log_prob(data.reshape(-1, 1, 1))).logsumexp(-1)
mean_log_prob = log_prob.logsumexp(-1) - np.log(S)
log_posterior_predictive = mean_log_prob.sum(-1)
```

I was able to get clarification from Prof. Kucukelbir and can confirm that I am using PDI's correctly. The "mean predictive" is not a widely used term, but it is just the mean value of the posterior predictive on the holdout set and is a bit more interpretable than the mean log predictive (since it is a mean empirical probability).

fehiepsi:
> what the Predictive class is doing

It gives you samples of `beta` and `lambda` from the posterior (the `guide`), which are `beta_mc` and `lambda_mc` in your computation.

> mean_log_prob

I think a better name of this variable is `log_mean_prob` (which is `log(mu(n))`). Sorry if the name caused confusion!

> I get a value of around -1500

This is a usual value AFAIK. Because this is the *sum* of log (mean) prob of *each* data points.

> mean value of the posterior predictive on the holdout set and is a bit more interpretable than the mean log predictive

Could you clarify what are *mean predictive* and *mean log predictive* (as I said in one of the comments, those are strange terminologies to me)? I was concerned about if we should use the formula in the introduction or in the content, i.e. `PDI = sigma^2(n) / mu(n)` or `WAPDI = sigma_log^2(n) / log(mu(n))`. Both of them use `mu(n)` (which is *mean predictive* I guess). None of them use `mu_log(n)` (which is *mean log predictive* I guess).

Or by posterior predictive, you mean `mu(n)`, and mean predictive, you mean `mean(mu(n))`? I am really confused, so I think it is best to add some definitions with formulas in the tutorial, especially for terminologies which are not wisely used ;)

(About interpretation, I am sure why the paper provides definitions, intuitions, computations, and experiments of WAPDI, but uses a different formula in the introduction, but let's skip discussing about this.) If you want to use the formula in the introduction, then could you address the following two points:
+ use the same posterior samples in the calculation of predictive values
+ fix `log_posterior_predictive_eval`: the formula in that function computes the negative of probability (i.e. `-mu(n)`), not `log_posterior_predictive`. Hence, the formula for `Mean Predictive` computes `exp(-prob)` (i.e. `exp(-mu(n))`. It is near 1 not because the model does a good job but because `prob` is pretty small. Could you please verify this point?

m-k-S:
I see! That all makes sense. I will write up a small section about how to perform model evaluation and how to optimize hyperparameters using intrinsic metrics (eg log mean predictive, PDI) and remove the unclear parts.

fehiepsi:
> Does that perhaps clarify?

Thanks, I think I understand your terminologies now. So mean predictive is `mean(mu(n))`, predictive distribution is `mu(n)`, and mean log predictive is `mean(log(mu(n)))`. That makes sense to me!

tutorial/source/dirichlet_process_mixture.ipynb
```diff
@@ -0,0 +1,799 @@
...
...GOFHSCuAV0nN0IyI2KTMwMzMrRy2Jf0jpUZiZWd10mfgjYqWkjYCt\ngbULk35fWlRmZlaaLhO/pOOAbwCbAzOADwP3AXuWGpmZmZWilpO7JwEtwNyI+CjwIeBvpUZlZmal\nqSXxL4+IVwAkrRkRs4D3lhuWmZmVpZaTu/MlDQZ+CfxG0vPAM+WGZWZmZanl5O7BefC7kj4ObARM\nLjUqMzMrTS01fiTtDLwnIi6XtCkwDPhrqZH1kVHj++dvVFdxzZ14wIBbd2/Wa2b1U8tVPd8BdiNd\nznk56ZLOq4Ddyw3NzMzKUMvJ3UNJXTIvA8j99GxYZlBmZlaeWhL/iogIch/8ktYtNyQzMytTLYn/\nBknnARtJOha4Hbi43LDMzKwsXSb+iDgD+BVwC/BB4N8j4tyulpN0saSF+SEureMmSHpW0vT82r83\nwZuZWfd1eHJX0vnAyRHxYkTcBtzWzXVfCvwn6YRw0Y8j4qxursvMzPpIZzX+J4EHJR3ZkxVHxD3A\n8z2KyszMStNh4o+IM0kdsY2VNEXSoZIOaX31YpsnSHokNwVt3Iv1mJlZD3T16MVnSXfpvgc4qPA6\nsIfbO590P8AYYD5wdkczShonaZqkaYsWLerh5szMrK3O2vi3JSXqecBOETG/txuLiAWF9V9IOmnc\n0byTgEkALS0t0dttm5lZ0tmdu9cBJ0bE7X21MUnDCz8gnwZmdja/mZn1vc4S/5iIWNHTFUu6mnSO\nYIikZ4BTgT0ljSHdDDYX+EpP129mZj3TYeLvTdLPyx/RzuiLerNOMzPrvVru3DUzsybSYeKXNCX/\nPaN+4ZiZWdk6a+MfLmlX4GBJ1wAqToyIh0qNzMzMStFZ4v8e8F3gXcA5baYFsFdZQZkfeNJWmQ+X\nMauazk7uXgdcJ+m7EXFaHWMyM7MS1fLM3dMkHQzskUdNjYgOb7wyM7P+rcureiSdDpwIPJpfJ0r6\nYdmBmZlZOWp52PoBpJu53gCQdBnwR+CUMgMzM7Ny1Hod/+DC8EZlBGJmZvVRS43/dOCPku4iXdK5\nBzC+1KjMzKw0tZzcvVrSVODDedTJEfFcqVGZmVlpaqnxk3vUvKXkWKxOfI+AWbW5rx4zs4px4jcz\nq5hOE7+kQZL+VK9gzMysfF09c3clMFvSyDrFY2ZmJavl5O7GwCxJD
```
fehiepsi: Beside fixing log_posterior_predictive_eval computation, could you review this formula? I am not sure but probably we should move the sum over `t` inside the expectation, and remove one duplicated E_q...

m-k-S: This formula is correct, these are independent expectations and thus they are able to factor as such. Similarly, expectations are linear so we can take the sum outside of the expectation.

m-k-S:
Hi @fehiepsi,

I've modified the first part of the model criticism section to use the log predictive as you have defined it (as a sum over the log predictive values for _each_ test point), and I used that value to demonstrate how we would find an approximately optimal value of a hyperparameter.

I've removed the section on posterior dispersion indices as it seems like it could be potentially confusing for those who might view this tutorial. However, if I continue to work on probabilistic programming in my research, I may attempt at contributing a PDI method in the core `pyro.infer` code, as that seems like a neat challenge and probably very useful to probabilistic programmers using Pyro :)

Let me know if there's anything else that should be changed!

Thanks again.

tutorial/source/dirichlet_process_mixture.ipynb
fehiepsi: Hmm, is it better to use a formula aligned with the implementation (without having to know those facts)? But I don't have a strong opinion on this. LGTM!

fehiepsi:
> contributing a PDI method in the core pyro.infer

Yeah, PDI seems like a nice way to evaluate a model. I would propose to have it in [ops.stats](https://github.com/pyro-ppl/pyro/blob/dev/pyro/ops/stats.py) module. :)

> I've removed the section on posterior dispersion indices

I think your explanation and reference would be more than enough and it is nice to have it. The original confusion is how to compute those terms (mainly log prob vs prob) so I thought (initially) that you wanted to use WAPDI. Would you mind adding it again with a correction on how PDI is computed? Don't worry if you still want to remove it.

> there's anything else that should be changed

Thanks for your contribution! The content looks good to me. Could you resolve the conflict at the `index.rst` file? I'll ask someone to take another pass on this great tutorial. :)

m-k-S:
Thanks @fehiepsi,

I fixed the conflict. I'll save the PDI work for when I understand it better and can implement it somewhere in the core Pyro code.

Is there anything else I need to do?

m-k-S:
Hi @martinjankowiak ,

Thank you! I've made the two changes as you've suggested. However, I'm not sure what you mean when you say that my guide distributions aren't more flexible – could you elaborate?

Thanks again!

martinjankowiak:
@m-k-S i just mean that you could have more free parameters (instead of all ones for one of the beta distribution parameters, etc.)

m-k-S:
Oh, I see. The first Beta parameter is set to 1 because that is what it is set to in the model as well. The covariance of the multivariate normal (q_mu) is set to identity just for convenience; would you prefer me to add a line stating this explicitly in the tutorial?

martinjankowiak:
@m-k-S well, unless you have good reasons for not doing so, i think it'd be better if you just make the q's more general: learn both beta parameters, learn the variance, etc. what do you think?

m-k-S:
Personally, I would prefer to keep it as is, for simplicity and clarity. I believe it is highly unlikely that the results would change even if those parameters are made free. I could include a short sentence or two explaining this decision, however, if you think that would be helpful.

martinjankowiak:
ok great sounds good let's do that instead. thanks!

Am Di., 17. Dez. 2019 um 08:54 Uhr schrieb MKS <notifications@github.com>:

PR #2185, opened 2019-11-22
add comment in Poutine tutorial that the first version of LogJointMessenger can only handle observed samples

dchudz:
It looks to me like when `LogJointMessenger` is introduced in the tutorial, the messenger can only handle observed values (because `_pyro_post_sample` is not yet introduced).

I guess it's relatively clear, but it took me a couple readings to see that that's what was going on -- so I thought a comment might help.

Apologies if I've misunderstood anything!


dchudz:
Oh, and I wanted to say that I loved the poutine tutorial (thanks!)  - really helpful for my understanding of pyro.

PR #2191, opened 2019-11-23
Reweighted Wake-Sleep

iffsid:
This implements `ReweightedWakeSleep` (RWS) following [1].
This is particularly useful for learning in models that involve stochastic branching, as described in [2].

It allows running inference in different modes based on the `insomnia` parameter. 
Denoting the model parameters as *theta* and guide parameters as *phi*, when `insomnia=1.0`, it simply runs the *wake_theta* and *wake_phi* updates.
When `insomnia=0.0`, it runs standard wake-sleep; i.e. the *wake_theta* update and the *sleep-phi* update.
When `0.0 <= insomnia <= 1.0`, the updates for *phi* are a weighted combination of the *wake-phi* and *sleep-phi* updates.

Note that when the model has no learnable parameters (indicated using `model_has_params=False`) and `insomnia=0.0`, this reduces to running `CSIS`.

[1] `Reweighted Wake-Sleep`,
    Jörg Bornschein, Yoshua Bengio

[2] `Revisiting Reweighted Wake-Sleep for Models with Stochastic Control Flow`,
    Tuan Anh Le, Adam R. Kosiorek, N. Siddharth, Yee Whye Teh, Frank Wood

**co-authored-with: @tuananhle7**

## Changes to other `pyro` aspects

* We have introduced the ability to `detach` a trace (in `trace_struct.py`), resulting in the sampled values at relevant sites being detached themselves. 
* Consequently, the `get_importance_trace` method (currently in `enum.py`) is instrumented with an additional flag `detach=False`, which allows computing traces as usual, but with appropriate detaching of the guide when required.
* Also, as with `CSIS`, being able to run the *sleep-phi* updates needs obtaining a handle on the observations in the guide---for which we expect explicit use of an 'observations' keyworded argument when defining the guide.
* Since RWS does not define a single *unified* objective, both the `loss` and `loss_and_grads` methods return a `tuple` of losses---one each for the model (*theta*) and guide (*phi*). `SVI` is hence suitably adapted to handle the return of a `tuple`.

We are open to suggestions/edits on these changes to help simplify/unify integration.

## Tested

* basic unit tests

[edit]: The 'rsvi' version of the `test_exponential_gamma` test fails for RWS due to an `is` assertion check that compares a non-detached and detached variable. This is bypassed by setting it up as a test that's expected to fail (as `xfail`).

cc: @fritzo


iffsid:
I don't believe the failing test is a numerics issue; this is what the error shows:
```
self = ShapeAugmentedGamma(rate: torch.Size([3, 1]), concentration: torch.Size([3, 1]))
boosted_x = tensor([[0.6212],
        [0.3735],
        [0.6635]])

    def score_parts(self, boosted_x=None):
        if boosted_x is None:
            boosted_x = self._unboost_x_cache[0]
>       assert boosted_x is self._unboost_x_cache[0]
E       AssertionError

pyro/distributions/testing/rejection_gamma.py:142: AssertionError
```
(above snippet taken from CI checks [here](https://travis-ci.com/pyro-ppl/pyro/jobs/259792730))

Suspect it's potentially a mismatch between model and inference; will investigate further.

fritzo:
Oh that might be due to `site["value"]=site["value"].detach()`. Is it possible to `.requires_grad_(False)` in-place?

iffsid:
No, that triggers a `RuntimeError`.

```
self = <pyro.poutine.trace_struct.Trace object at 0x7efccb160470>

    def detach(self):
        """
        Detach values (in-place) at each sample site of the trace.
        """
        for _, site in self.nodes.items():
            if site["type"] == "sample":
                # site["value"] = site["value"].detach()
>               site["value"].requires_grad_(False)
E               RuntimeError: you can only change requires_grad flags of leaf variables. If you want to use a computed variable in a subgraph that doesn't require differentiation use var_no_grad = var.detach().

pyro/poutine/trace_struct.py:262: RuntimeError
```
I also tried out `.detach_()`, but that fails because it doesn't work when the tensor is a `view`.

Would it be reasonable to change the tripped assertion to a `torch.equal` check?

fritzo:
I think it's fine if you disable that test for rws. The assertion should test with `is`, not equal. We could add more general metadata, but I don't there will be many users of `Rejector` combined with rws.

tuananhle7:
> I think it's fine if you disable that test for rws. The assertion should test with `is`, not equal. We could add more general metadata, but I don't there will be many users of `Rejector` combined with rws, and the error is not silent.

Is something like [this](https://github.com/pyro-ppl/pyro/pull/2191/commits/ebbd5aa8f51183ede1f9e578f2a972b5f1ef8688) OK?

cc @iffsid

pyro/infer/rws.py
```diff
@@ -0,0 +1,234 @@
...
+                                                  vectorize_particles=vectorize_particles,
+                                                  strict_enumeration_warning=strict_enumeration_warning)
+        self.insomnia = insomnia
+        self.model_has_params = model_has_params
+        self.num_sleep_particles = num_particles if num_sleep_particles is None else num_sleep_particles
+
+        assert(insomnia >= 0 and insomnia <= 1), \
+            "insomnia should be in [0, 1]"
+
+        print("Running {} {}".format(
```
martinjankowiak: i don't think we want this print statement (we don't do this anywhere else in pyro)

iffsid: done.

pyro/infer/enum.py
```diff
@@ -33,12 +33,14 @@ def iter_discrete_extend(trace, site, **ignored):
         yield extended_trace
 
 
-def get_importance_trace(graph_type, max_plate_nesting, model, guide, *args, **kwargs):
+def get_importance_trace(graph_type, max_plate_nesting, model, guide, *args, detach=False, **kwargs):
```
fritzo: @neerajprad this interface is becoming awkward. WDYT of replacing `*args, **kwargs` with `args, kwargs` in a later cleanup PR, either before or after this merges?

pyro/infer/svi.py
```diff
@@ -128,4 +132,8 @@ def step(self, *args, **kwargs):
         # zero gradients
         pyro.infer.util.zero_grads(params)
 
-        return torch_item(loss)
+        # RWS: handle loss = (wake_theta_loss, wake_phi_loss)
```
fritzo: I think it would be cleaner to (1) handle arbitrary `tuple` and `namedtuple` objects
```py
if isinstance(loss, tuple):  # including namedtuple
    return type(loss)(map(torch_item, loss))
else:
    return torch_item(loss)
```
and (2) move this logic into `torch_item` and `torch_isnan` so that svi.py is unmodified. WDYT?

pyro/infer/rws.py
```diff
@@ -0,0 +1,227 @@
...
+        :returns: returns model loss and guide loss
+        :rtype: float, float
+
+        Computes the re-weighted wake-sleep estimators for the model (wake-theta) and the
+          guide (insomnia * wake-phi + (1 - insomnia) * sleep-phi).
+        """
+        with torch.no_grad():
+            wake_theta_loss, phi_loss = self._loss(model, guide, *args, **kwargs)
+
+        return wake_theta_loss, phi_loss
```
fritzo: I am a bit wary of returning tuples from these functions rather than scalars, however the changes in this PR seem surprisingly minimal.

pyro/infer/rws.py
```diff
@@ -0,0 +1,227 @@
...
+        return wake_theta_loss.detach(), phi_loss.detach()
+
+    def _vectorized_num_sleep_particles(self, fn):
+        """
+        Copy of `_vectorised_num_particles` that uses `num_sleep_particles`.
+        """
+        def wrapped_fn(*args, **kwargs):
+            if self.num_sleep_particles == 1:
+                return fn(*args, **kwargs)
+            with pyro.plate("num_sleep_particles_vectorized", self.num_sleep_particles, dim=-self.max_plate_nesting):
```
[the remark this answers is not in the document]
fritzo: Also could you ensure there are tests with say `num_particles=2` and `num_sleep_particles=3`?

pyro/infer/rws.py
fritzo: Is there a good reason to return multiple losses? I am very tempted to return
```py
wake_theta_loss + phi_loss
```
and use the simpler alternative to the try-except logic in this PR:
```py
(wake_theta_loss + phi_loss).backward()
```

pyro/infer/rws.py
```diff
@@ -0,0 +1,227 @@
...
+    def loss_and_grads(self, model, guide, *args, **kwargs):
+        """
+        :returns: returns model loss and guide loss
+        :rtype: float
+
+        Computes the RWS estimators for the model (wake-theta) and the guide (wake-phi).
+        Performs backward as appropriate on both, using num_particle many samples/particles.
+        """
+        wake_theta_loss, phi_loss = self._loss(model, guide, *args, **kwargs)
+        try:
```
fritzo: simplify to
```py
(wake_theta_loss + phi_loss).backward()
```

pyro/infer/rws.py
```diff
@@ -0,0 +1,227 @@
...
+            if self.num_sleep_particles == 1:
+                return fn(*args, **kwargs)
+            with pyro.plate("num_sleep_particles_vectorized", self.num_sleep_particles, dim=-self.max_plate_nesting):
+                return fn(*args, **kwargs)
+
+        return wrapped_fn
+
+    @staticmethod
+    def _get_matched_trace(model_trace, guide, *args, **kwargs):
+        # TODO: hardcoded kwarg 'observations'?
```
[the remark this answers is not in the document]
iffsid: So, I'm not quite sure how we could avoid it, really.

Basically, when doing _sleep-phi_,  we need to switch the processing order from _wake-*_; we first sample the (unconditioned) model trace, and then replay that on the guide. 

Because the guide doesn't `sample` the observation variable, but is instead given it, we need to match things up somehow for the `replay`. We hence using the same procedure as with `CSIS` by forcing use of a named keyword `observations` to hold the observations that the guide uses.

pyro/infer/rws.py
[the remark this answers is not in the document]
iffsid: I might have misunderstood your comment, but `num_particles` and `num_sleep_particles` are never _both_ employed for a single trace. 

`num_particles` is only used for the _wake-*_ computation.
`num_sleep_particles` is only used for the _sleep-phi_ computation, which draws _entirely new traces_ from the model and guide (in that order) using the `_vectorized_num_sleep_particles` method.

pyro/infer/svi.py
iffsid: Completely fine with whatever the best-fit approach for pyro ought to be. We just went with this approach to minimise edits. :)

pyro/infer/rws.py
iffsid: I'm not sure it's correct to merge them---mainly because they are losses corresponding to different _objectives_, not just gradients w.r.t different variables.

The _wake_theta_loss_ computes the gradients with respect to (only) model parameters through the `iwae` objective, whereas the _phi_loss_ computes gradients w.r.t (only) guide parameters through (a combination of) the `csis` objective and a reweighted (self-normalised IS) version of the `csis` objective.

pyro/infer/rws.py
iffsid: Apart from the reason to keep the losses separate as we discuss elsewhere, this check serves a use case where there are no model (_theta_) parameters to learn, but where we would like to employ learning for the guide (_phi_) parameters through some combination of _wake_phi_ and _sleep_phi_ updates (even just _wake_phi_).

It's a `try-catch` block because it isn't straightforward to figure out if a model has parameters to learn without effectively expanding *all* of it ahead of time, which may not always be possible.

fritzo: Do I understand correctly that the try-catch block fails when `wake_theta_loss` does not depend on any learnable parameters? If so then you should first simplify to
```py
if wake_theta_loss.requires_grad:
    wake_theta_loss.backward()
```
and then simplify further to the equivalent
```py
(wake_theta_loss + phi_loss).backward()
```
```

File: pyro/infer/rws.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (fritzo) [reply pyro/infer/rws.py]: Thanks for explaining!

File: pyro/infer/rws.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (fritzo) [reply pyro/infer/rws.py]: Oh, I don't understand the CSIS use of `kwargs["observations"]` either. @eb8680 can you explain?

@iffsid can you explain the TODO?

File: pyro/infer/rws.py [hunk 8]
```diff
@@ -0,0 +1,227 @@
...
+from pyro.infer.elbo import ELBO
+from pyro.infer.enum import get_importance_trace
+from pyro.infer.util import is_validation_enabled
+from pyro.poutine.util import prune_subsample_sites
+from pyro.util import check_if_enumerated, check_model_guide_match, warn_if_nan
+
+
+class ReweightedWakeSleep(ELBO):
+    r"""
+    An implementation of Reweighted Wake Sleep following reference [1].
```
fritzo: Could you also add a comment explaining that the guide needs to take an optional kwarg `observations` which is a dict?

pyro/infer/svi.py
fritzo: On second thought I guess this change is fine here. It would be nice to use the more general form though, `type(loss)(map(torch_item, loss))`.

pyro/infer/rws.py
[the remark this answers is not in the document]
eb8680: The guide distribution `q(z | x)` depends in general on the values of observed variables `x` in the model's joint distribution `p(x, z)`.  In the wake phase, those values are given as data, but in the sleep phase, the data is ignored and those values are instead drawn from the model `p(x | z)`, and need to be extracted from a model trace and passed to the guide.  They can't currently be passed with an effect handler because the dependence of `q` on `x` is deterministic and arbitrary.  I suppose after #2196 we could update `CSIS` and `RWS` to require the guide to have `deterministic` calls around the values `x` and accomplish the same thing with `condition`, although that doesn't seem any more elegant than the current solution.

fritzo:
Could you also add a section in docs/source/inference_algos.rst, e.g. following Importance?

pyro/infer/rws.py
fritzo: Ok, that makes sense that you'd want to return multiple separate losses. Thanks for explaining.

pyro/infer/enum.py
neerajprad: I think it makes sense to only use this `*model_args, **model_kwargs` pattern for user-facing interface, and even then, its debatable if there are too many other arguments. I'll put up a fix.

pyro/infer/rws.py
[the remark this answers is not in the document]
iffsid: @fritzo The TODO was just to ensure that we explicitly discussed this issue in the PR. :)
I'll drop the TODO comment and add a note in the docs about the explicit `observations` keyword as well.

pyro/infer/rws.py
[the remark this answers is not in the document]
tuananhle7: done https://github.com/pyro-ppl/pyro/pull/2191/commits/f0e2d79c6e4a2d8da154398ca35db0f4f031bc44

pyro/infer/rws.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+
+    def __init__(self,
+                 num_particles=2,
+                 insomnia=1.,
+                 model_has_params=True,
+                 num_sleep_particles=None,
+                 vectorize_particles=True,
+                 max_plate_nesting=float('inf'),
+                 max_iarange_nesting=None,  # DEPRECATED
```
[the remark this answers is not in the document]
tuananhle7: done in  https://github.com/pyro-ppl/pyro/pull/2191/commits/3b8f00ebe13b924ba91a099e050b826aff9c41ef

tuananhle7:
> Could you also add a section in docs/source/inference_algos.rst, e.g. following `Importance`?

done in https://github.com/pyro-ppl/pyro/pull/2191/commits/a4105e557fe47bec5825da5103cc4dea5ebd2247

pyro/infer/svi.py
iffsid: Done in `0a48f5c8`.

pyro/infer/rws.py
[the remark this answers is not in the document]
iffsid: Done in `7532af20`.

pyro/infer/rws.py
iffsid: Done in `7532af20`.

iffsid:
> Hi @iffsid @tuananhle7 , thanks for this PR. We are planning to publish a minor 1.1 release this Friday. If you think you'll have time to respond to the minor comments in this PR, we'd be happy to prioritize review so this can make the 1.1 release. No rush though if you're busy.

Thanks, that'd be great!
We'll sort out the edits for the minor comments asap.

pyro/infer/svi.py
```diff
@@ -103,7 +103,11 @@ def evaluate_loss(self, *args, **kwargs):
         Evaluate the loss function. Any args or kwargs are passed to the model and guide.
         """
         with torch.no_grad():
-            return torch_item(self.loss(self.model, self.guide, *args, **kwargs))
+            loss = self.loss(self.model, self.guide, *args, **kwargs)
+            if isinstance(loss, tuple):
+                return torch_item(loss[0]), torch_item(loss[1])
```
fritzo: ditto: can you support longer and namedtuples here:
```py
return type(loss)(map(torch_item, loss))
```

pyro/poutine/trace_struct.py
```diff
@@ -252,6 +252,15 @@ def compute_score_parts(self):
                     warn_if_nan(site["log_prob_sum"], "log_prob_sum at site '{}'".format(name))
                     warn_if_inf(site["log_prob_sum"], "log_prob_sum at site '{}'".format(name), allow_neginf=True)
 
+    def detach(self):
```
fritzo: nit: could you rename this method to `detach_()` to follow PyTorch naming conventions for in-place ops?

pyro/infer/rws.py
```diff
@@ -0,0 +1,242 @@
...
+        O(`num_particles`) samples from the guide,
+        O(`num_sleep_particles`) from the model,
+        O(`num_particles` + `num_sleep_particles`) `log_prob` evaluations of the guide, and
+        O(`num_particles`) evaluations of the model
+
+
+    .. note:: This is particularly useful for models with stochastic branching,
+        as described in [2].
+
+    .. note:: This returns _two_ losses, one each for the model and the guide.
```
fritzo: Could you mention the correspondence theta=model, phi=guide here? Or even go farther and use variable names like `wake_model_loss` and `guide_loss` in the code (only if you feel that does not stray too far from your paper).

pyro/infer/rws.py
```diff
@@ -0,0 +1,242 @@
...
+        Performs backward as appropriate on both, using num_particle many samples/particles.
+        """
+        wake_theta_loss, phi_loss = self._loss(model, guide, *args, **kwargs)
+        try:
+            wake_theta_loss.backward(retain_graph=True)
+        except RuntimeError:
+            pass
+        phi_loss.backward()
+
+        return wake_theta_loss.detach(), phi_loss.detach()
```
fritzo: nit: I believe all losses currently return floats. Could you 
```py
return ....detach().item(), ....detach().item()
```
here?

pyro/poutine/trace_struct.py
iffsid: I presume it's alright to define an in-place op that _not_ a pure side effect? 
This method also returns the detached trace.

fritzo: Yes that's fine, in PyTorch, `.detach()` creates a detached copy, `.detach_()` detaches in-place, and both return the result (which in the latter case is the same object). I just thought it would be nice if your function were named similarly `.detach_()` in case someone eventually wants to implement a `.detach() = .copy().detach_()`.

pyro/infer/svi.py
iffsid: Ah, missed that one. Done in 60e277a3.

pyro/poutine/trace_struct.py
iffsid: Nevermind. I changed it so it is actually just a pure side effect (in 357f529e).

pyro/infer/rws.py
iffsid: Done. Added correspondence (60486ebd)

pyro/infer/rws.py
iffsid: Done in 7d7b5f16

pyro/infer/rws.py
iffsid: Yes, can simplify that. Done in 92dfa729.

fritzo:
Note you'll need to merge with dev and resolve conflicts in tests/infer/test_inference.py

PR #2193, opened 2019-11-27
Added zero inflated negative binomial

kashif:
Also refactored the zero inflated poisson into a base class.


kashif:
thanks @fehiepsi  any idea why the test

```
pytest tests/infer/test_enum.py::test_elbo_zip
```

is failing? Calling the secondary distribution's `.expand()` seems to set the type of the parameter to `Size`...

pyro/distributions/zero_inflated.py
```diff
@@ -0,0 +1,99 @@
...
+    def mean(self):
+        return (1 - self.gate) * self.distribution.mean
+
+    @lazy_property
+    def variance(self):
+        return (1 - self.gate) * (
+            self.distribution.mean ** 2 + self.distribution.variance
+        ) - (self.mean) ** 2
+
+    def expand(self, batch_shape):
```
fehiepsi: @kashif I think you would need to follow PyTorch convention here:
```
        new = self._get_checked_instance(type(self), _instance)
        gate = self.gate.expand(batch_shape)
        distribution = self.distribution.expand(batch_shape)
        super(type(self), new).__init__(gate, distribution, validate_args=False)
        new._validate_args = self._validate_args
        return new
```

I would recommend to change `distribution` to `base_dist`, which is also a convention name in Pyro. :)

kashif: thanks will do!

kashif:
thanks @fehiepsi

PR #2197, opened 2019-12-01
Add a to_pyro_module_() helper

fritzo:
Addresses #2161 

This aims to make it easier to convert 3rd party `nn.Module` instances to `PyroModule`s, as described in @simeneide's [forum post](https://forum.pyro.ai/t/the-usage-of-pyromodule/1443).

The safest way to do this appears to be to update the `.__class__` attribute in-place:
```py
module.__class__ = PyroModule[module.__class__]
module._pyro_... = ...  # initialize PyroModule attributes
```
While this appears intrusive, it is equivalent to replacing `nn.Linear -> PyroModule[nn.Linear]` and so forth in the 3rd-party code that created the instance. It should therefore also be compatible with serialization.

## 🐛 Bugfixes

This PR also fixes two bugs surfaced by new tests:

- A bug whereby `._pyro_set_supermodule()` was being bypassed due to `nn.Module` subclasses calling `.add_module()` rather than `.__setitem__()` directly.
- A recursion bug in `tests.common.assert_close("foo", "foo")`

## Tested
- added a test that `to_pyro_module_(m)` is equivalent to recursively inserting `PyroModule[-]`
- strengthen existing tests

### DIFF
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -502,9 +502,11 @@ def to_pyro_module_(m, recurse=True):
         assert isinstance(model[0], PyroModule[nn.Linear])
 
         # Now we can attempt to be fully Bayesian:
-        for name, value in list(model.named_parameters()):
-            deep_setattr(name, PyroSample(prior=
-                dist.Normal(0, 1).expand(value.shape).to_event(value.dim())))
+        for m in model.modules():
+            for name, value in list(m.named_parameters(recurse=False)):
+                setattr(m, name, PyroSample(prior=dist.Normal(0, 1)
+                                                      .expand(value.shape)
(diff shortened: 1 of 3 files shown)


pyro/nn/module.py
```diff
@@ -479,3 +480,69 @@ def clear(mod):
...
+    # Reregister parameters and submodules.
+    for name, value in list(m._parameters.items()):
+        setattr(m, name, value)
+    for name, value in list(m._modules.items()):
+        if recurse:
+            to_pyro_module_(value)
+        setattr(m, name, value)
+
+
+def deep_setattr(obj, name, value):
```
fritzo: This is added simply to support the docstring example.

PR #2198, opened 2019-12-02
Add tutorial using EnergyDistance and Stable distribution

fritzo:
This adds a tutorial on likelihood-free inference using `EnergyDistance` #2178 and a `Stable` distribution #2181 . The problem is heavy-tailed stochastic volatility estimation, similar to the PyMC3 tutorial https://docs.pymc.io/notebooks/stochastic_volatility.html

## Tasks
- [ ] Fix convergence issues
- [ ] Simplify inference (e.g. use defaults wherever possible)

View notebook [here](https://nbviewer.jupyter.org/github/pyro-ppl/pyro/blob/trace-crps-example/tutorial/source/likelihood_free.ipynb)


fritzo:
Replaced by #2253

PR #2199, opened 2019-12-02
Implement uniform mixture sampling in TraceTMC_ELBO

eb8680:
This PR adds a different default strategy for merging particle populations in Tensor Monte Carlo: instead of propagating a single set of particles, new samples from each variable are drawn from a mixture over all possible combinations of parent particles.

Tasks:
- [x] Sample different ancestors for each particle
- [x] Refactor TMC sample functions to look similar and use a single `Vindex` call
- [x] Make new strategy the default TMC implementation

Tested:
- Added to existing TMC tests


eb8680:
I think I'd like to get this into the 1.1 release if possible, it seems like a more sensible default than the diagonal approximation from #2143.  @fritzo we can go through this in person tomorrow if you want.

eb8680:
@fritzo I changed the default strategy back to diagonal per our discussion

PR #2203, opened 2019-12-03
Do not use varargs in internal functions / methods in pyro.infer

neerajprad:
Based on comment in https://github.com/pyro-ppl/pyro/pull/2191#discussion_r352404596, this uses `args, kwargs` directly in internal methods / functions in `pyro.infer`.

I have tried to restrict this PR to the following  methods / functions:
 - `get_importance_trace`.
 - methods internal to `ELBO` - `_get_trace`, `_get_traces`, `_get_vectorized_trace`, `_guess_max_plate_nesting`.
 - `TraceMMD._differentiable_loss_parts`, `TraceGraph_ELBO. _loss_and_surrogate_loss `

This is an internal refactoring, and shouldn't affect existing tests. I think it is usually best to reserve the varargs pattern for user-facing interface. All internal functions / methods that get these from a user-facing API should just pass these along as is without doing additional packing and unpacking. I'm happy to revert this, or open it up for further discussion if reviewers think otherwise.

### DIFF
--- a/pyro/infer/discrete.py
+++ b/pyro/infer/discrete.py
@@ -208,9 +208,9 @@ class TraceEnumSample_ELBO(TraceEnum_ELBO):
                            first_available_dim=-2)(*args, **kwargs)
 
     """
-    def _get_trace(self, model, guide, *args, **kwargs):
+    def _get_trace(self, model, guide, args, kwargs):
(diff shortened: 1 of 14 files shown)


fritzo:
You'll need to update pyro/infer/rws.py

PR #2204, opened 2019-12-04
add MultivariateCauchy dist and LowerCholeskyAffine

martinjankowiak:


pyro/distributions/multivariate_cauchy.py
```diff
@@ -0,0 +1,51 @@
...
+    arg_constraints = {'loc': constraints.real, 'scale_tril': constraints.lower_triangular}
+    support = constraints.real
+    has_rsample = True
+
+    def __init__(self, loc, scale_tril, validate_args=None):
+        base_dist = _Cauchy(loc, 1.0).to_event(1)
+        transforms = [AffineTransform(torch.zeros_like(loc), scale_tril, event_dim=1)]
+        super(MultivariateCauchy, self).__init__(base_dist, transforms, validate_args=validate_args)
+
+    def expand(self, batch_shape, _instance=None):
```
martinjankowiak: will this work?

pyro/distributions/multivariate_cauchy.py
```diff
@@ -0,0 +1,51 @@
...
+    def scale(self):
+        return self.base_dist.scale_tril
+
+    @property
+    def mean(self):
+        return self.loc
+
+    @property
+    def variance(self):
+        raise ValueError("Distribution has infinite variance.")
```
martinjankowiak: is `ValueError` ok?

pyro/distributions/multivariate_cauchy.py
```diff
@@ -0,0 +1,51 @@
+import torch
```
martinjankowiak: suggestions for tests?

pyro/distributions/multivariate_cauchy.py
```diff
@@ -0,0 +1,51 @@
...
+from torch.distributions import constraints
+from torch.distributions.transforms import AffineTransform
+from torch.distributions.cauchy import Cauchy
+from pyro.distributions.torch import TransformedDistribution
+from pyro.distributions.torch_distribution import TorchDistributionMixin
+
+
+class _Cauchy(Cauchy, TorchDistributionMixin):
```
fritzo: Can you instead `from pyro.distributions.torch import Cauchy`?

pyro/distributions/multivariate_cauchy.py
fritzo: cc @neerajprad

pyro/distributions/multivariate_cauchy.py
```diff
@@ -0,0 +1,51 @@
...
+    :param torch.tensor loc: the location parameter of the distribution, a D-dimensional vector.
+    :param torch.tensor scale_tril: the scale parameter of the distribution, a D x D lower-triangular matrix.
+    """
+    arg_constraints = {'loc': constraints.real, 'scale_tril': constraints.lower_triangular}
+    support = constraints.real
+    has_rsample = True
+
+    def __init__(self, loc, scale_tril, validate_args=None):
+        base_dist = _Cauchy(loc, 1.0).to_event(1)
+        transforms = [AffineTransform(torch.zeros_like(loc), scale_tril, event_dim=1)]
```
martinjankowiak: oh wait this won't work affine is only for pointwise stuff

pyro/distributions/multivariate_cauchy.py
```diff
@@ -0,0 +1,51 @@
...
+    def loc(self):
+        return self.base_dist.loc
+
+    @property
+    def scale(self):
+        return self.base_dist.scale_tril
+
+    @property
+    def mean(self):
+        return self.loc
```
fritzo: This should return nan (since mean is undefined), and variance should return inf. See [cauchy.py](https://github.com/pytorch/pytorch/blob/master/torch/distributions/cauchy.py) for example:
```py
from torch._six import inf, nan

class MultivariateCauchy(...):
    ...
    @property
    def mean(self):
        return self.loc.new_full(self.shape(), nan)

    @property
    def variance(self):
        return self.loc.new_full(self.shape(), inf)
```

pyro/distributions/multivariate_cauchy.py
neerajprad: Yes, this should work fine.

pyro/distributions/multivariate_cauchy.py
```diff
@@ -0,0 +1,47 @@
...
+        new = self._get_checked_instance(MultivariateCauchy, _instance)
+        return super(MultivariateCauchy, self).expand(batch_shape, _instance=new)
+
+    @property
+    def loc(self):
+        return self.base_dist.loc
+
+    @property
+    def scale(self):
+        return self.base_dist.scale_tril
```
fritzo: This doesn't look right, won't you need to apply the `AffineTransform` and then extract the `.diagonal(dim1=-2, dim2=-1)`?

pyro/distributions/multivariate_cauchy.py
fritzo: add some shape tests to `tests/distributions/conftest.py`

pyro/distributions/multivariate_cauchy.py
neerajprad: Hmm..yeah, this won't work. I think you can probably easily adapt @fehiepsi's [MultivariateAffineTransform](https://github.com/pyro-ppl/numpyro/blob/033393afc92563c82ef9723907f36b8774065b9b/numpyro/distributions/transforms.py#L336) for this purpose, and we can later move it upstream to pytorch if it seems useful.

pyro/distributions/multivariate_cauchy.py
fritzo: return nan; see comment on `def mean(self):`

pyro/distributions/multivariate_cauchy.py
fritzo: Yeah, we should add that to `pyro.distributions.transforms` and later move it upstream.

pyro/distributions/multivariate_cauchy.py
martinjankowiak: i'm not sure `scale` isn't defined in `Distribution` so i'm not really sure what this is supposed to be... perhaps inappropriate here?

pyro/distributions/multivariate_cauchy.py
martinjankowiak: thanks, done

pyro/distributions/multivariate_cauchy.py
martinjankowiak: great done

tests/distributions/test_transforms.py
```diff
@@ -131,6 +131,17 @@ def test_householder_inverses(self):
         for input_dim in [2, 5, 10]:
             self._test_inverse(input_dim, T.householder(input_dim, count_transforms=2))
 
+    def test_lower_cholesky_affine(self):
```
martinjankowiak: tests for LowerCholeskyAffine

pyro/distributions/multivariate_cauchy.py
martinjankowiak: done

pyro/distributions/multivariate_cauchy.py
martinjankowiak: i removed this altogether

pyro/distributions/transforms/lower_cholesky_affine.py
```diff
@@ -0,0 +1,58 @@
...
+
+    :param loc: the fixed D-dimensional vector to shift the input by.
+    :type loc: torch.tensor
+    :param scale_tril: the D x D lower triangular matrix used in the transformation.
+    :type scale_tril: torch.tensor
+    """
+    codomain = constraints.real
+    bijective = True
+    event_dim = 1
+    volume_preserving = False
```
neerajprad: This seems to be unused?

martinjankowiak: yeah it's something stefan added thinking it might be useful later 🤷‍♂

neerajprad: Oh..this seems like something Stefan has introduced in our transforms module?

pyro/distributions/transforms/lower_cholesky_affine.py
```diff
@@ -0,0 +1,58 @@
+import torch
+from torch.distributions import constraints
+from torch.distributions.transforms import Transform
+
+from pyro.distributions.util import copy_docs_from
+
+
+@copy_docs_from(Transform)
+class LowerCholeskyAffine(Transform):
```
fehiepsi: In NumPyro, we also have [MultivariateAffineTransform transform](http://num.pyro.ai/en/stable/distributions.html#multivariateaffinetransform). Should we make the name consistent?

pyro/distributions/transforms/lower_cholesky_affine.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    A bijection of the form :math:`\\mathbf{y} = \\mathbf{L} \\mathbf{x} + \\mathbf{r}`
+    where `\\mathbf{L}` is a lower triangular matrix and `\\mathbf{r}` is a vector.
+
+    :param loc: the fixed D-dimensional vector to shift the input by.
+    :type loc: torch.tensor
+    :param scale_tril: the D x D lower triangular matrix used in the transformation.
+    :type scale_tril: torch.tensor
+    """
+    codomain = constraints.real
```
fehiepsi: I think domain and codomain should be `real_vector`.

pyro/distributions/transforms/lower_cholesky_affine.py
martinjankowiak: yes

pyro/distributions/transforms/lower_cholesky_affine.py
martinjankowiak: i think my name is better : ) not all invertible matrices are LowerCholesky

pyro/distributions/multivariate_cauchy.py
```diff
@@ -0,0 +1,39 @@
...
+    :param torch.tensor scale_tril: the scale parameter of the distribution, a D x D lower-triangular matrix.
+    """
+    arg_constraints = {'loc': constraints.real, 'scale_tril': constraints.lower_triangular}
+    support = constraints.real
+    has_rsample = True
+
+    def __init__(self, loc, scale_tril, validate_args=None):
+        base_dist = Cauchy(loc, 1.0).to_event(1)
+        transforms = [LowerCholeskyAffine(torch.zeros_like(loc), scale_tril)]
+        self.loc = loc
```
fritzo: I don't think this will work correctly with `.expand`. Maybe that's why those `@property`s were there before. Maybe we should
```py
@property
def loc(self):
    return self.base_dist.loc

@property
def scale_tril(self):
    return self.transforms[0].scle_tril
```

pyro/distributions/transforms/lower_cholesky_affine.py
fehiepsi: So we'll change the name in NumPyro when pyro-api is more matured. : )

pyro/distributions/multivariate_cauchy.py
```diff
@@ -0,0 +1,39 @@
...
+
+    :param torch.tensor loc: the location parameter of the distribution, a D-dimensional vector.
+    :param torch.tensor scale_tril: the scale parameter of the distribution, a D x D lower-triangular matrix.
+    """
+    arg_constraints = {'loc': constraints.real, 'scale_tril': constraints.lower_triangular}
+    support = constraints.real
+    has_rsample = True
+
+    def __init__(self, loc, scale_tril, validate_args=None):
+        base_dist = Cauchy(loc, 1.0).to_event(1)
```
fritzo: Whoa, are you sure you want
```py
Transformed(Cauchy(loc, ...), LowerCholeskyAffine(0, scale_tril))
```
I would think you'd want to match `MultivariateNormal` and put the loc in the transform
```py
TransformedDistribution(Cauchy(torch.zeros_like(loc), 1.).to_event(1),
                        LowerCholeskyAffine(loc, scale_tril))
```

fritzo: Otherwise you'll get weird coupling of the `loc` and `scale_tril` parameters, right?

martinjankowiak: maybe can you please explain? obviously math is the same...

fritzo: Hmm I am pretty sure the math is different:
```py
# This is the sampler you implemented:
scale_tril @ (noise + loc)

# This is the sampler I am suggesting:
scale_tril @ noise + loc
```
Which one of us is confused 😕 ?

fehiepsi: not related but may I ask a question here:

Can I use this LowerCholeskyAffine to define the MVT implementation at [this PR](https://github.com/pyro-ppl/pyro/pull/2025/files#diff-35ab0c5d82928aaf937d461c83519de2)? I got some trouble with writing a test to verify that the implementation is correct (I only verify the consistency with univariate case). I have a feeling that it might not be the same: MVT(df, loc, scale_tril) and TransformedDist(ST(df, 0, 1).expand([1])), Affine(loc, scale_tril))...

martinjankowiak: @fritzo ah yeah you're right of course thanks for catching! 😊

martinjankowiak: @fehiepsi you need to include the gamma variate outside the transform: z = mu + L z_unit / tau.sqrt() (or something like that)

pyro/distributions/multivariate_cauchy.py
neerajprad: Sorry I missed this refactoring. This will throw an error currently because `expanded_dist.loc` will be missing.

pyro/distributions/multivariate_cauchy.py
fehiepsi: Thank Martin! So they are different. I'll think more about how to write a proper test for it when revisiting that PR.

fehiepsi: Oh the second thought, I think the implementation here uses a different definition w.r.t. the one in [wikipedia](https://en.wikipedia.org/wiki/Cauchy_distribution#Multivariate_Cauchy_distribution). Consider `loc=0`, `scale_tril=id`, log_prob of MVC here is `dist.Cauchy(0, 1).log_prob(x).sum(-1)` so it has the form of `log(1 + x ** 2).sum(-1)`, while the wikipedia definition has the form of `log(1 + ||x|| ** 2)`... If so, I think we should rename this distribution to something else.

fritzo: @fehiepsi yes Martin and I were discussing radial symmetry offline. I agree `1 + ||x||**2` seems more parsimonious.

martinjankowiak:
closing this in favor of #2207 and a separate MultivariateStudentT PR

PR #2209, opened 2019-12-04
Add default TorchDistribution.expand method

### LINKED ISSUE #2186: FR ExpandedDistribution for easier custom distributions
As of #1377 it is onerous to define custom distributions: each new distribution requires a custom `.expand()` method --a method that is difficult to understand and inconsistently implemented among reference distributions. This issue proposes to resurrect `ReshapedDistribution` as a simpler `ExpandedDistribution` and to use that as a fallback in case `.expand()` is not implemented.

Request: classes deriving from `TorchDistribution` need not implement `.expand()`

neerajprad:
Resolves #2186. 

This adds a default `.expand` method in `TorchDistributon` by resurrecting the earlier `ReshapedDistribution`. 

**Tests:** All existing `expand` tests exercise both the distribution's expand method as well as the default one provided in `TorchDistribution`, and these pass.

### DIFF
--- a/pyro/distributions/torch_distribution.py
+++ b/pyro/distributions/torch_distribution.py
@@ -1,4 +1,5 @@
 import warnings
+from collections import OrderedDict
 
 import torch
 from torch.distributions import constraints
@@ -6,6 +7,7 @@ from torch.distributions.kl import kl_divergence, register_kl
 
 import pyro.distributions.torch
 from pyro.distributions.distribution import Distribution
+from pyro.distributions.score_parts import ScoreParts
 from pyro.distributions.util import broadcast_shape, scale_and_mask
 
 
@@ -197,7 +199,10 @@ class TorchDistribution(torch.distributions.Distribution, TorchDistributionMixin
     method to improve gradient estimates and set
     ``.has_enumerate_support = True``.
     """
-    pass
+    # Provides a default `.expand` method for Pyro distributions which overrides
+    # torch.distributions.Distribution.expand (throws a NotImplementedError)
+    def expand(self, batch_shape, _instance=None):
+        return ReshapedDistribution(self, batch_shape)
 
 
 class MaskedDistribution(TorchDistribution):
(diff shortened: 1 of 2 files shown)


pyro/distributions/torch_distribution.py
```diff
@@ -280,6 +285,108 @@ def variance(self):
...
+        shape = broadcast_shape(self.batch_shape, value.shape[:value.dim() - self.event_dim])
+        log_prob = self.base_dist.log_prob(value)
+        return log_prob.expand(shape)
+
+    def score_parts(self, value):
+        shape = broadcast_shape(self.batch_shape, value.shape[:value.dim() - self.event_dim])
+        log_prob, score_function, entropy_term = self.base_dist.score_parts(value)
+        if self.batch_shape != self.base_dist.batch_shape:
+            log_prob = log_prob.expand(shape)
+            score_function = score_function.expand(shape)
```
fritzo: I belive `score_function` and `entropy_term` are sometimes identically `float(0)`, so you'll need to
```py
if isinstance(score_function, torch.Tensor):
    score_function = score_function.expand(shape)
if isinstance(entropy_term, torch.Tensor):
    entropy_term = entropy_term.expand(shape)
```

pyro/distributions/torch_distribution.py
```diff
@@ -197,7 +199,10 @@ class TorchDistribution(torch.distributions.Distribution, TorchDistributionMixin
     method to improve gradient estimates and set
     ``.has_enumerate_support = True``.
     """
-    pass
+    # Provides a default `.expand` method for Pyro distributions which overrides
+    # torch.distributions.Distribution.expand (throws a NotImplementedError)
+    def expand(self, batch_shape, _instance=None):
```
fritzo: Do I understand correctly that this must live in `TorchDistribution` rather than `TorchDistributionMixin` due to mro?

pyro/distributions/torch_distribution.py
```diff
@@ -280,6 +285,108 @@ def variance(self):
         return self.base_dist.variance
 
 
+class ReshapedDistribution(TorchDistribution):
```
fritzo: Can we name this `ExpandedDistribution`? The original `Reshaped` implemented both `Independent()` and `.expand()`, whereas your new implementation serves only to implement `.expand()`.

tests/distributions/test_distributions.py
```diff
@@ -151,17 +152,22 @@ def test_expand_by(dist, sample_shape, shape_type):
 
 @pytest.mark.parametrize('sample_shape', [(), (2,), (2, 3)])
 @pytest.mark.parametrize('shape_type', [torch.Size, tuple, list])
-def test_expand_new_dim(dist, sample_shape, shape_type):
+@pytest.mark.parametrize('default', [False, True])
+def test_expand_new_dim(dist, sample_shape, shape_type, default):
     for idx in range(dist.get_num_test_data()):
         small = dist.pyro_dist(**dist.get_dist_params(idx))
         with xfail_if_not_implemented():
```
fritzo: Can we now remove any of the `xfail_if_not_implemented`s, here and in other tests below?

pyro/distributions/torch_distribution.py
neerajprad: That's right; otherwise we'll call the default `torch.distributions.Distribution.expand` which will throw a `NotImplementedError`. Maybe we can move this upstream to PyTorch, if it looks reasonable.

pyro/distributions/torch_distribution.py
neerajprad: Good point, will change.

tests/distributions/test_distributions.py
neerajprad: Done - I have restricted the scope of `xfail` so that it only applies to custom implementations.

pyro/distributions/torch_distribution.py
neerajprad: I have moved this to `TorchDistributionMixin` so that it is also available to `.expand_by` and set `expand = TorchDistributionMixin.expand` here.

PR #2210, opened 2019-12-05
Automatically generate handlers from Messengers

### LINKED ISSUE #2188: Check constructor argument types in Messengers and handlers
In [this topic on the forum](https://forum.pyro.ai/t/poutine-mask-intended-usage/1428/2), a user passed a mask tensor to `poutine.mask` as an argument `poutine.mask(mask)` rather than a keyword argument `poutine.mask(mask=mask)`, attempted to use the resulting `Messenger` as a context manager, and got an obscure error message as a result:
```
ValueError: Expected mask to be a boolean but got <class ‘NoneType’>
```
It would be good to raise more informative error messages in such cases that would make the issue clear (in this case, raising a `TypeError` or `ValueError` when initializing the `Messenger`'s `fn` attribute to something clearly not callable).

eb8680:
Resolves #2188.

This almost-pure refactoring PR:
1. Removes all the hand-defined boilerplate convenience functions like `poutine.trace` defined in `pyro/poutine/handlers.py` and moves or merges their docstrings into the corresponding `Messenger` classes.
2. Programatically generates new versions of these functions that behave identically (necessary to address #2188)
3. Adds a one-line check in the automatically generated versions that addresses #2188 
4. Adds a regression test that would have caught #2188 to `test/poutine/test_poutine.py`
5. Changes the name of `EnumerateMessenger` to `EnumMessenger` to avoid a name collision with the `enumerate` builtin during programmatic handler generation and be consistent with its preexisting alias `poutine.enum`

Almost all of the changes in this PR are copy-pasted docstrings and deleted boilerplate.

Tested:
- Pure refactoring exercised by existing tests
- Regression test in `test_poutine.py`
- Docs render correctly on my machine

### DIFF
--- a/pyro/poutine/block_messenger.py
+++ b/pyro/poutine/block_messenger.py
@@ -81,8 +81,8 @@ class BlockMessenger(Messenger):
         >>> def fn():
         ...     a = pyro.sample("a", dist.Normal(0., 1.))
         ...     return pyro.sample("b", dist.Normal(a, 1.))
-        >>> fn_inner = trace(fn)
-        >>> fn_outer = trace(block(fn_inner, hide=["a"]))
+        >>> fn_inner = pyro.poutine.trace(fn)
+        >>> fn_outer = pyro.poutine.trace(pyro.poutine.block(fn_inner, hide=["a"]))
         >>> trace_inner = fn_inner.get_trace()
         >>> trace_outer  = fn_outer.get_trace()
         >>> "a" in trace_inner
--- a/pyro/poutine/block_messenger.py
+++ b/pyro/poutine/block_messenger.py
@@ -63,11 +63,8 @@ def _make_default_hide_fn(hide_all, expose_all, hide, expose, hide_types, expose
 
 class BlockMessenger(Messenger):
     """
-    This Messenger selectively hides Pyro primitive sites from the outside world.
+    This handler selectively hides Pyro primitive sites from the outside world.
     Default behavior: block everything.
-    BlockMessenger has a flexible interface that allows users
-    to specify in several different ways
-    which sites should be hidden or exposed.
 
     A site is hidden if at least one of the following holds:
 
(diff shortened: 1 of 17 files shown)


pyro/poutine/handlers.py
```diff
@@ -69,359 +70,76 @@
...
+                "{} is not callable, did you mean to pass it as a keyword arg?".format(fn))
+        msngr = msngr_cls(*args, **kwargs)
+        return msngr(fn) if fn is not None else msngr
+
+    return handler
+
+
+_re1 = re.compile('(.)([A-Z][a-z]+)')
+_re2 = re.compile('([a-z0-9])([A-Z])')
+for _msngr_cls in _msngrs:
```
eb8680: Here is the programmatic handler generation; compare to `pyro/distributions/torch.py`

pyro/poutine/handlers.py
```diff
@@ -69,359 +70,76 @@
...
+    UnconditionMessenger,
+]
+
+
+def _make_handler(msngr_cls):
+
+    def handler(fn=None, *args, **kwargs):
+        if fn is not None and not callable(fn):
+            raise ValueError(
+                "{} is not callable, did you mean to pass it as a keyword arg?".format(fn))
```
eb8680: Here is the fix for #2188

pyro/poutine/handlers.py
```diff
@@ -69,359 +70,76 @@
...
+
+    return handler
+
+
+_re1 = re.compile('(.)([A-Z][a-z]+)')
+_re2 = re.compile('([a-z0-9])([A-Z])')
+for _msngr_cls in _msngrs:
+    _handler_name = _re2.sub(
+        r'\1_\2', _re1.sub(r'\1_\2', _msngr_cls.__name__.split("Messenger")[0])).lower()
+    _handler = _make_handler(_msngr_cls)
```
fritzo: `_handler.__name__ = _handler_name`?

pyro/poutine/infer_config_messenger.py
```diff
@@ -3,7 +3,13 @@
 
 class InferConfigMessenger(Messenger):
     """
-    Modifies contents of the infer kwarg at sample sites
+    Given a callable that contains Pyro primitive calls
```
fritzo:
```
Given a callable `fn` that contains Pyro primitive calls
and a callable `config_fn` taking a trace ...
```

pyro/poutine/handlers.py
eb8680: @neerajprad will this play nicely with serialization?  Do I need to use `functools.partial` here instead of a closure?

eb8680:
> I noticed EnumerateMessenger/EnumMessenger is not documented (otherwise we would be introducing a breaking change). Should it be documented?

Probably, but this PR is just refactoring - all of the documentation changes here are from copy-pasting handler docstrings into messengers, not rewriting them.

pyro/poutine/handlers.py
neerajprad: This should work fine! These will be available from the module namespace, so pickle shouldn't have a problem.

eb8680:
I'll update the Sphinx config/doc generation for `pyro.poutine` in a followup PR to make sure all `Messenger`s are included.

pyro/poutine/block_messenger.py
```diff
@@ -77,47 +74,38 @@ class BlockMessenger(Messenger):
...
     :param expose_fn: function that takes a site and returns True to expose the site
         or False/None to hide it.  If specified, all other parameters are ignored.
         Only specify one of hide_fn or expose_fn, not both.
-    :param bool hide_all: hide all sites
-    :param bool expose_all: expose all sites normally
-    :param list hide: list of site names to hide, rest will be exposed normally
-    :param list expose: list of site names to expose, rest will be hidden
-    :param list hide_types: list of site types to hide, rest will be exposed normally
-    :param list expose_types: list of site types to expose normally, rest will be hidden
+    :param hide: list of site names to hide
```
neerajprad: I think the previous docstring was more complete - `hide_all` and `expose_all` parameters are missing. Also maybe we should put in the type info back - some IDEs will do type checks based on the docstring.

neerajprad: Feel free to address this in your followup PR, btw.

eb8680: Good catch, I'll include this in the followup PR

PR #2211, opened 2019-12-06
fix guide_trace and model_trace in bayesian regression tutorial

Capri2014:


tutorial/source/bayesian_regression.ipynb
```diff
@@ -747,8 +747,8 @@
     "        \n",
     "    def forward(self, *args, **kwargs):\n",
     "        samples = {}\n",
-    "        guide_trace = poutine.trace(guide).get_trace(*args, **kwargs)\n",
-    "        model_trace = poutine.trace(poutine.replay(model, guide_trace)).get_trace(*args, **kwargs)\n",
+    "        guide_trace = poutine.trace(self.guide).get_trace(*args, **kwargs)\n",
```
neerajprad: Thanks for fixing this bug! I was wondering why this was even working in the first place - it seems that it was picking up the model and guide from the outer scope.

PR #2213, opened 2019-12-06
Implement GammaGaussianHMM

fehiepsi:
Addresses https://github.com/pyro-ppl/funsor/issues/190
~~Blocked by #2025~~

As a consequence of #2025, this PR implements a student-t filter which is exact and parallel scanned. Note that this requires the joint distribution of init_state, latent sites, obs sites is a StudentT distribution (hence, there is only one common degree of freedom). Relaxing this condition would require some sort of moment matching for tensordot op and would be non-exact.

### References

+ F. J. Giron and J. C. Rojano (1994) "Bayesian Kalman filtering with elliptically contoured errors"
+ Filip Tronarp, Toni Karvonen, and Simo Sarkka (2019)
   "Student's t-filters for noise scale estimation"
   https://users.aalto.fi/~ssarkka/pub/SPL2019.pdf

### TODO
- [x] double check `.filter()` method.
This method returns posteriors for both "multiplier" `s` and the final latent state, which should be enough for forecasting.
- [x] add tests similar to GaussianHMM for this class

Blocked by #2025


fehiepsi:
@fritzo All tests have passed locally in my system. Could you suggest some hmm examples for me to test this class? :)

fritzo:
@fehiepsi all of Martin's `pyro.contrib.timeseries` are based on `GaussianHMM`, so you could try those. Typically Martin used the EEG dataset as in examples. Also just search for `GaussianHMM`. Finally, search through http://pyro.ai/time

fehiepsi:
Thanks, @fritzo ! Starting with Martin's [time series examples](https://github.com/pyro-ppl/pyro/blob/dev/examples/contrib/timeseries/gp_models.py) looks like the right direction. If the class performs well, then I'll replace egg by another noisy dataset.

docs/source/distributions.rst
```diff
@@ -210,6 +210,13 @@ Rejector
     :undoc-members:
     :show-inheritance:
 
+RobustHMM
```
fritzo: Can we rename this to `StudentTHMM` or something more specific, so as to leave room for alternative robust time series models like `StableHMM`?

fritzo: Or maybe `GammaGaussianHMM`, to distinguish it from a collection of independent `StudentT` distributions.

fehiepsi: done

pyro/distributions/hmm.py
```diff
@@ -324,6 +349,166 @@ def filter(self, value):
...
+        assert isinstance(transition_matrix, torch.Tensor)
+        assert isinstance(transition_dist, MultivariateNormal)
+        assert isinstance(observation_matrix, torch.Tensor)
+        assert isinstance(observation_dist, MultivariateNormal)
+        hidden_dim, obs_dim = observation_matrix.shape[-2:]
+        assert initial_dist.event_shape == (hidden_dim,)
+        assert transition_matrix.shape[-2:] == (hidden_dim, hidden_dim)
+        assert transition_dist.event_shape == (hidden_dim,)
+        assert observation_dist.event_shape == (obs_dim,)
+        shape = broadcast_shape(initial_dist.batch_shape + (1,),
```
fritzo: Don't forget the new parameter,
```py
shape = broadcast_shape(scale_dist.batch_shape + (1,),
                        initial_dist.batch_shape + (1,),
                        ...,
```

tests/distributions/test_hmm.py
```diff
@@ -436,3 +456,111 @@ def test_gaussian_mrf_log_prob_block_diag(sample_shape, batch_shape, num_steps,
     actual_log_prob = d.log_prob(data)
     expected_log_prob = marginal_obs_dist.log_prob(data).sum(-1)
     assert_close(actual_log_prob, expected_log_prob)
+
+
+@pytest.mark.parametrize('obs_dim', [1, 2])
+@pytest.mark.parametrize('hidden_dim', [1, 3])
+@pytest.mark.parametrize('init_shape,trans_mat_shape,trans_mvn_shape,obs_mat_shape,obs_mvn_shape', [
```
fritzo: Could you add a leftmost column for `scale_shape` and a new row
```py
((5,), (), (6,), (), (), ()),
```

tests/distributions/test_hmm.py
```diff
@@ -436,3 +456,111 @@ def test_gaussian_mrf_log_prob_block_diag(sample_shape, batch_shape, num_steps,
...
+    ((), (6,), (6,), (6,), (6,)),
+    ((5,), (6,), (), (), ()),
+    ((), (5, 1), (6,), (), ()),
+    ((), (), (5, 1), (6,), ()),
+    ((), (), (), (5, 1), (6,)),
+    ((), (6,), (5, 1), (), ()),
+    ((), (), (6,), (5, 1), ()),
+    ((), (), (), (6,), (5, 1)),
+    ((5,), (), (), (), (6,)),
+    ((5,), (5, 6), (5, 6), (5, 6), (5, 6)),
```
fritzo: and change this row to
```py
((5,), (5,), (5, 6), (5, 6), (5, 6), (5, 6)),
```

tests/distributions/test_hmm.py
```diff
@@ -436,3 +456,111 @@ def test_gaussian_mrf_log_prob_block_diag(sample_shape, batch_shape, num_steps,
...
+                                  obs_mat_shape, obs_mvn_shape, hidden_dim, obs_dim):
+    init_dist = random_mvn(init_shape, hidden_dim)
+    trans_mat = torch.randn(trans_mat_shape + (hidden_dim, hidden_dim))
+    trans_dist = random_mvn(trans_mvn_shape, hidden_dim)
+    obs_mat = torch.randn(obs_mat_shape + (hidden_dim, obs_dim))
+    obs_dist = random_mvn(obs_mvn_shape, obs_dim)
+    scale_dist = random_gamma(init_shape)
+    d = dist.GammaGaussianHMM(scale_dist, init_dist, trans_mat, trans_dist, obs_mat, obs_dist)
+
+    shape = broadcast_shape(init_shape + (1,),
```
fritzo:
```py
shape = broadcast_shape(scale_shape + (1,), ...)
```

fehiepsi:
Thank you for reviewing and suggestions, @fritzo ! You got it right. I think we can split `_init` into `_scale` and `_init` but it seems not necessary.

fritzo:
Oh I do like the current organization `self._init = scale_prior + init_prior`.

PR #2215, opened 2019-12-06
Bump to version 1.1.0

neerajprad:
The only additional change is deprecating the `save_visualization` function and removing `graphviz` as a core dependency.


setup.py
```diff
@@ -78,7 +78,6 @@
     install_requires=[
         # if you add any additional libraries, please also
         # add them to `docs/requirements.txt`
-        'graphviz>=0.8',
```
fritzo: Should we add this to `EXTRAS_REQUIRE` so it is available to make docs, or is that unnecessary?

neerajprad: Oh..I thought this was added; yes, I'll add it to `EXTRAS_REQUIRE`.

Issue #2221, opened 2019-12-10
[FR] Automatic reparameterization


fritzo:
Could we implement automatic reparameterization? Maybe at least for autoguides?

A good test example would be Neal's funnel, as described in the paper.

See Gorinova, Moore, and Hoffman (2019)
https://arxiv.org/pdf/1906.03028.pdf

## Tasks
- [x] #2222 implement `poutine.param` and basic framework
- [x] #2235 generalize `LocScaleTransform` and implement an effect `pyro.infer.reparam.decenter()`
- [x] #2235 implement `TransformReparameterizer` (suggested by @fehiepsi ) ~~and an effect `pyro.infer.reparam.neutra()`~~
- [x] #2235 support `Independent` distributions
- ~~implement a Levy-Ito reparameterization~~
- ~~implement a `difference` or `cumsum` reparameterization for time series~~
- [ ] #2252 implement a Fourier transform reparameterization for time series

fehiepsi:
FYI, numpyro supports automatic reparameterization for both autoguide and
mcmc by defining that distribution as a transformed distribution. This is a
nice feature to have and can help for avoiding mistakes where support of
some “transformed” distributions depend on other variable, e.g. Uniform(0,
a) has dynamic support but currently autoguide will use a fixed transform
(depending on initial value of a). If we reparameterize here, we only need
to learn guide for U(0, 1) and that issue won’t happen.

However, the biggest problem for supporting this is non-trivial codes will
be added, so it is hard to review and follow the logic regarding this
feature in NumPyro. I will sketch the solution there below so we can
discuss if there is a better solution.

PR #2222, opened 2019-12-11
Implement poutine.reparam with aux variable support

fritzo:
Addresses #2221 

This adds a `poutine.reparam` that transforms a sample site (based on `site["fn"]`) into one or more auxiliary sample sites followed by a deterministic transformation. This generalizes [(Gorinova et al. 2019)](https://arxiv.org/pdf/1906.03028.pdf) by allowing multiple "overparameterized" sample sites for auxiliary variable methods.

The main motivation for this PR is to support [Stable](http://docs.pyro.ai/en/stable/distributions.html#stable) distributions in variational inference. Although `Stable.log_prob()` is undefined, the distribution can be reparameterized to a pair of auxiliary variables with defined `.log_prob()`; thus `poutine.reparam` can make a stable model tractable.

I have designed to allow custom `Reparameterizer` instances to be set per-site, either manually or via `@infer_config`. This PR implements only `LocScaleReparameterizer` and `StableReparameterizer`. 

Note the `Reparameterizer` class restricts the dependency structure, disallowing one auxiliary variable to depend on another. In principle we could replace a `pyro.sample` site with an arbitrary probabilistic program (and the proposed `pyro.reparam()` interface could be adapted to this general case via an `isinstance(-,Reparameterizer)` check). One advantage of an opinionated `Reparameterizer` class is that if forces separation of concerns between `pyro.distributions` and `pyro.poutine`: no pyro primitives need be used in `Reparameterizer` implementations.

## Tested
- [x] test `TrivialReparameterizer`
- [x] test `LocScaleReparameterizer`
- [x] test `StableReparameterizer`
- [x] smoke test for `Stable` inference with autoguides
- [x] test SVI in tests/infer/test_valid_models.py
- [x] test MHC and NUTS in tests/infer/mcmc/test_valid_models.py
- [x] ensure `StableReparameterizer` is continuous

See generated [docs](http://fritzo.org/temp/pyro/html/distributions.html#reparameterizers)

### DIFF
--- a/pyro/distributions/reparameterize.py
+++ b/pyro/distributions/reparameterize.py
@@ -66,7 +66,7 @@ class TrivialReparameterizer(Reparameterizer):
 class LocScaleReparameterizer(Reparameterizer):
     """
     Generic centering reparameterizer for distributions that are completely
-    spacified by parameters ``loc`` and ``scale``.
+    specified by parameters ``loc`` and ``scale``.
     """
     def get_dists(self, fn):
         loc = torch.zeros_like(fn.loc)
--- a/pyro/distributions/reparameterize.py
+++ b/pyro/distributions/reparameterize.py
@@ -71,7 +71,7 @@ class LocScaleReparameterizer(Reparameterizer):
     def get_dists(self, fn):
         loc = torch.zeros_like(fn.loc)
         scale = torch.ones_like(fn.scale)
-        new_fn = type(fn)(loc, scale)
+        new_fn = type(fn)(loc=loc, scale=scale)
         return OrderedDict([("centered", new_fn)])
 
     def transform_values(self, fn, values):
--- a/pyro/distributions/reparameterize.py
+++ b/pyro/distributions/reparameterize.py
@@ -28,21 +28,28 @@ class Reparameterizer(ABC):
     @abstractmethod
     def get_dists(self, fn):
         """
+        Constructs one or more auxiliary
+        :class:`~pyro.distribution.Distribution` s that will replace ``fn``.
+
         :param ~pyro.distributions.Distribution fn: A base distribution.
         :returns: An OrderedDict mapping name (suffix) to auxiliary distribution.
-        :rtype: OrderedDict
+        :rtype: ~collections.OrderedDict
         """
-        return OrderedDict([("trivial", fn)])
+        raise NotImplementedError
 
     @abstractmethod
     def transform_values(self, fn, values):
         """
+        Deterministically combines samples from the distributions returned by
+        :meth:`get_dists` into a single reparameterized sample from ``fn``.
+
         :param ~pyro.distributions.Distribution fn: A base distribution.
-        :param OrderedDict values:
+        :param ~collections.OrderedDict values: An OrderedDict mapping name (suffix) to
+            sample value drawn from corresponding auxiliary variable.
         :returns: A transformed value.
-        :rtype: Tensor
+        :rtype: ~torch.Tensor
         """
-        return values["trivial"]
+        raise NotImplementedError
 
 
 class TrivialReparameterizer(Reparameterizer):
--- a/pyro/distributions/reparameterize.py
+++ b/pyro/distributions/reparameterize.py
@@ -1,0 +1,71 @@
+from abc import ABC, abstractmethod
+from collections import OrderedDict
+
+import torch
+
+
+class Reparameterizer(ABC):
+    """
+    Base class for reparameterization transforms, generalizing [1] to handle
+    multiple distributions and auxiliary variables.
+
+    The interface specifies two methods to be used by inference algorithms in
+    the following pattern::
+
+        # Transform a site like this...
+        pyro.sample("x", dist.Normal(loc, scale))
+
+        # ...to one or more sample sites plus a deterministic site:
+        d = dist.Normal(loc, scale)
+        values = OrderedDict((name, pyro.sample("x_"+name, fn))
+                             for name, fn in reparam.get_dists(d))
+        pyro.deterministic("x", reparam.transform_values(d, values))
+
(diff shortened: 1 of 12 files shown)


tests/distributions/test_reparameterize.py
```diff
@@ -0,0 +1,107 @@
...
+            with pyro.plate("particles", 100000):
+                pyro.sample("x", dist.Normal(loc, scale))
+
+    value = poutine.trace(model).get_trace().nodes["x"]["value"]
+    expected_probe = normal_probe(value)
+
+    def config_fn(site):
+        if site["name"] == "x" and isinstance(site["fn"], dist.Normal):
+            return {"reparam": Reparam()}
+        return {}
```
fehiepsi: I guess we don't need to return an empty dict here?

fehiepsi:
In the future, it would be nice to have `TransformedReparameterizer` which handles transformed distribution. I would be very happy to use this pattern. :)
```
reparam_model = poutine.reparam(model, {"x": TransformedReparameterizer()})
```

pyro/poutine/reparam_messenger.py
```diff
@@ -0,0 +1,61 @@
...
+        # Combine auxiliary values via pyro.deterministic().
+        # TODO(https://github.com/pyro-ppl/pyro/issues/2214) refactor to
+        # use site type "deterministic" when it exists.
+        value = reparam.transform_values(msg["fn"], new_values)
+        assert isinstance(value, torch.Tensor)
+        if getattr(msg["fn"], "_validation_enabled", False):
+            # Validate while the original msg["fn"] is known.
+            msg["fn"]._validate_sample(value)
+        msg["value"] = value
+        msg["fn"] = Delta(value, event_dim=msg["fn"].event_dim).mask(False)
```
neerajprad: Not directly relevant to this PR, but I think this will probably not work for HMC. In that case, we need to be able to score samples from a base distribution (like Normal(0., 1.)) using the transformed distribution's log density. The samples themselves will be generated by HMC, and the delta function (IIUC) will result in the sample getting rejected because of having infinite energy. Just an FYI; I haven't thought fully about it, but it may be the case that both are sufficiently different to warrant separate solutions. It is also straightforward to do non-centered parametrization manually, so we can ignore the MCMC case for the time being.

fritzo: This is just a `pyro.deterministic()` call in-lined to avoid cyclic dependencies. Does HMC not yet support `pyro.deterministi()`?

fehiepsi: I think this will work for HMC.

fritzo:
@fehiepsi I like the `TransformedParameterizer` idea! Especially because that broadens our class of transforms to non-invertible transforms: reparameterization avoids the need to implement `.log_abs_det_jacobian()`.

pyro/poutine/reparam_messenger.py
neerajprad: Thanks for clarifying, I misread what this was doing. `pyro.deterministic` should work fine.

fehiepsi:
Yup, it is especially useful for flows with no `inverse` method (e.g. BNAFlow) and will make autoguide faster because, as you pointed out, no need to go through `transform.log_abs_det_jacobian`.

tests/distributions/test_reparameterize.py
fritzo: A dict is required by `infer_config`'s `config_fn`

pyro/distributions/reparameterize.py
```diff
@@ -0,0 +1,78 @@
...
+        return OrderedDict([("trivial", fn)])
+
+    def transform_values(self, fn, values):
+        return values["trivial"]
+
+
+class LocScaleReparameterizer(Reparameterizer):
+    """
+    Generic centering reparameterizer for distributions that are completely
+    spacified by parameters ``loc`` and ``scale``.
```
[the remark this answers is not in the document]
fritzo: done

fritzo:
@fehiepsi I have updated to support both notations:
```py
# Either specify in the reparam config:
@poutine.reparam(config={"x": LocScaleReparameterizer()})
...
# or mark a single site.
pyro.sample("y", dist.Normal(loc, scale),
            infer={"reparam": LocScaleReparameterizer})
```
That way I can still use `@config_infer` to match sites with `isinstance(site["fn"], dist.Stable)`). 

@neerajprad I have added smoke tests for `HMC` and `NUTS`.

pyro/poutine/reparam_messenger.py
```diff
@@ -0,0 +1,72 @@
...
+    :param dict config: Optional configuration mapping site name to
+        :class:`~pyro.distributions.reparameterize.Reparameterizer` object.
+    """
+    def __init__(self, config=None):
+        super().__init__()
+        if config is None:
+            config = {}
+        self.config = config
+
+    def _pyro_sample(self, msg):
```
fritzo: @eb8680 can you take a look to ensure I've implemented this correctly? I've followed `poutine.do`.

eb8680: I'll need to read the paper to understand so probably won't be able to get to this today

fritzo: Oh I'm not so concerned with matching up with the paper as ensuring that I have correctly transformed a single `sample` site into multiple `sample` sites plus a `pyro.deterministic` site.

fritzo:
Hmm I may be missing a `log_abs_det_jacobian` term in `.transform_values()`...

fehiepsi:
I think we don't need `log_abs_det_jacobian` in autoguide/mcmc because the inference happens in unconstrained space (assume base dists have real support, this means that we only need to compute `p(x)` rather than `p(t(x))` where `x` belongs to base dists' domain and `t` is the transform).

Issue #2223, opened 2019-12-11
Undocumented restrictions on discrete inference?


lsgos:
## Weird behaviour in discrete inference with shared mixture weights

pyro version - 1.1.0 (and earlier)

Consider the following pyro model, defining a mixture of gaussians 

```py
K = 5
d = 2
N_obs = 100
@infer.config_enumerate
def model(N=1):
    with pyro.plate('data_plate', N, dim=-2):
        mixing_weights = pyro.param('pi', th.ones(K) / K, constraint=constraints.simplex)
        means = pyro.sample('mu', D.Normal(th.zeros(K, d), th.ones(K, d)).to_event(2))
    
        with pyro.plate('observations', N_obs, dim=-1):
            s = pyro.sample('s', D.Categorical(mixing_weights))

            x = pyro.sample('x', D.Normal(I.Vindex(means)[..., s, :], 0.1).to_event(1))
```
This can be enumerated, and works fine. In particular, gradients flow to the mixing weights variable, as can be checked with the following snippet

```py
pyro.clear_param_store()
conditioned_model = pyro.condition(model, data={'x': x})
guide = infer.autoguide.AutoDelta(poutine.block(conditioned_model, hide=['s']))

elbo = infer.TraceEnum_ELBO(max_plate_nesting=2)
opt = pyro.optim.Adam({'lr': 1e-2})
svi = infer.SVI(conditioned_model, guide, opt, elbo)
# probably need 
losses = []
for i in range(1000):
    l = svi.step(x.size(0))
    if i % 100 == 0: 
        print(i, l)
        elbo.loss_and_grads(conditioned_model, guide, x.size(0))
        print(pyro.get_param_store()._params['pi'].grad)
    losses.append(l)
```

Now, if we add another random variable after x that depends on the same mixture component, everything breaks 

```py
pyro.clear_param_store()

K = 5
d = 2
N_obs = 100
@infer.config_enumerate
def model(N=1):
    with pyro.plate('data_plate', N, dim=-2):
        mixing_weights = pyro.param('pi', th.ones(K) / K, constraint=constraints.simplex)
        means = pyro.sample('mu', D.Normal(th.zeros(K, d), th.ones(K, d)).to_event(2))
    
        with pyro.plate('observations', N_obs, dim=-1):
            s = pyro.sample('s', D.Categorical(mixing_weights))

            x = pyro.sample('x', D.Normal(I.Vindex(means)[..., s, :], 0.1).to_event(1))
[...]

PR #2226, opened 2019-12-17
Pyro enum bugfix [WIP]

lsgos:
As suggested [here](https://github.com/pyro-ppl/pyro/issues/2223#issuecomment-565234047), this pull request adds a failing test which reproduces #2223. It currently makes no attempt to fix the issue, but should hopefully be useful when debugging.

If anything needs cleaning up please let me know. Help is appreciated - I'm not very familiar with the low level mechanics of the enumeration code atm.

I also found that the test for the multivariate student T was failing on cuda because it parameterises df as a float rather than a tensor, and then the samples inherit the device of df. The first commit here fixes this test by changing tensor_wrap to also wrap floats, sending these to cuda if wrapped in a tensors_default_to context. I squashed this here because it came up first when I was trying to hit my failing test.

### DIFF
--- a/pyro/infer/traceenum_elbo.py
+++ b/pyro/infer/traceenum_elbo.py
@@ -98,7 +98,7 @@ def _compute_model_factors(model_trace, guide_trace):
         if site["type"] == "sample":
             if name in guide_trace.nodes:
                 cost_sites.setdefault(ordering[name], []).append(site)
-                non_enum_dims.update(site["packed"]["log_prob"]._pyro_dims)
+                non_enum_dims.update(guide_trace.nodes[name]["packed"]["log_prob"]._pyro_dims)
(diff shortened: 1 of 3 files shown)


lsgos:
Do you want me to put the change to dist_fixture in a separate PR? I kind of agree it doesn't make sense to bundle it with this one, as it's off topic, but I think its a real (minor) bug - I wasn't able to run the test suite locally without it.

fritzo:
> Do you want me to put the change to dist_fixture in a separate PR?

Yes, actually could you also open up an issue with the failed test output. I'm not sure whether fixing the fixture is the right solution, or whether that indicates a missing coercion in a distribution `.__init__()` method which we should fix instead. Thanks!

PR #2229, opened 2019-12-18
Small typo in unused function in tutorial

alexander-held:
The second part of the basic tutorial [An Introduction to Inference in Pyro](http://pyro.ai/examples/intro_part_ii.html) shows how to condition on an observation in different ways. The first example `conditioned_scale` is used for the remainder of the tutorial, and a function `scale_obs(guess)` is defined to show an equivalent way of doing the conditioning. In order for this function to be equivalent, the standard deviation for one `dist.Normal()` call needs to be updated to match the value used in `scale(guess)`.

The function `scale_obs(guess)` changed by this is not used afterwards in the tutorial, so this changes no functionality.


fritzo:
Note CI builds are failing due to #2230 ; we can rerun tests after that merges.

PR #2233, opened 2019-12-23
Implement StudentT ops

fehiepsi:
The implementation will assume that `df > 2` throughout.

Right now, there is nothing implemented yet. I'll add more details below. The math is mostly corresponding to GammaGaussian op.

#### Reference

+ Robust Bayesian Filtering and Smoothing Using Student’s t Distribution https://arxiv.org/pdf/1703.02428.pdf

#### Tasks

- [ ] Implement StudentT op, except the sum method. 
- [ ] Consider to implement the "approximated" sum method <-  I'm not sure if this is necessary
- [ ] Implement the "approximated" tensordot
- [ ] Add test to verify that this match the sequential approximation mechanism in the above reference
- [ ] Add tests similar to Gaussian/GammaGaussian ops.

### DIFF
--- a/pyro/ops/studentt.py
+++ b/pyro/ops/studentt.py
@@ -1,0 +1,384 @@
+import math
+
+import torch
+from torch.distributions.utils import lazy_property
+from torch.nn.functional import pad
+
+from pyro.distributions.util import broadcast_shape
+
+
+class Gamma:
+    """
+    Non-normalized Gamma distribution.
+    """
+    def __init__(self, log_normalizer, concentration, rate):
+        self.log_normalizer = log_normalizer
+        self.concentration = concentration
+        self.rate = rate
+
+    def log_density(self, s):
+        """
+        Non-normalized log probability of Gamma distribution.
+
+        This is mainly used for testing.
+        """
+        return self.log_normalizer + (self.concentration - 1) * s.log() - self.rate * s
+
+    def logsumexp(self):
+        """
+        Integrates out the latent variable.
+        """
+        return self.log_normalizer + torch.lgamma(self.concentration) - \
+            self.concentration * self.rate.log()
+
+
+class StudentT:
+    """
+    Non-normalized StudentT distribution:
+
+        StudentT(x, s) ~ (concentration + 0.5 * dim - 1) * log(s)
+                              - rate * s - s * 0.5 * info_vec.T @ inv(precision) @ info_vec)
+                              - s * 0.5 * x.T @ precision @ x + s * x.T @ info_vec,
+
+    which will be reparameterized as
+
+        GammaGaussian(x, s) =: alpha * log(s) + s * (-0.5 * x.T @ precision @ x + x.T @ info_vec - beta).
+
+    The `s` variable plays the role of a mixing variable such that
+
+        p(x | s) ~ Gaussian(s * info_vec, s * precision).
+
+    Conditioned on `s`, this represents an arbitrary semidefinite quadratic function,
+    which can be interpreted as a rank-deficient Gaussian distribution.
+    The precision matrix may have zero eigenvalues, thus it may be impossible
+    to work directly with the covariance matrix.
+
+    :param torch.Tensor log_normalizer: a normalization constant, which is mainly used to keep
+        track of normalization terms during contractions.
+    :param torch.Tensor info_vec: information vector, which is a scaled version of the mean
+        ``info_vec = precision @ mean``. We use this represention to make gaussian contraction
+        fast and stable.
+    :param torch.Tensor precision: precision matrix of this gaussian.
+    :param torch.Tensor alpha: reparameterized shape parameter of the marginal Gamma distribution of
+        `s`. The shape parameter Gamma.concentration is reparameterized by:
+
+            alpha = Gamma.concentration + 0.5 * dim - 1
+
+    :param torch.Tensor beta: reparameterized rate parameter of the marginal Gamma distribution of
+        `s`. The rate parameter Gamma.rate is reparameterized by:
+
+            beta = Gamma.rate + 0.5 * info_vec.T @ inv(precision) @ info_vec
+    """
+    def __init__(self, log_normalizer, info_vec, precision, alpha, beta):
+        # NB: using info_vec instead of mean to deal with rank-deficient problem
+        assert info_vec.dim() >= 1
+        assert precision.dim() >= 2
+        assert precision.shape[-2:] == info_vec.shape[-1:] * 2
+        self.log_normalizer = log_normalizer
+        self.info_vec = info_vec
+        self.precision = precision
+        self.alpha = alpha
+        self.beta = beta
+
+    def dim(self):
+        return self.info_vec.size(-1)
+
+    @lazy_property
+    def batch_shape(self):
+        return broadcast_shape(self.log_normalizer.shape,
+                               self.info_vec.shape[:-1],
+                               self.precision.shape[:-2],
+                               self.alpha.shape,
+                               self.beta.shape)
+
+    def expand(self, batch_shape):
+        n = self.dim()
+        log_normalizer = self.log_normalizer.expand(batch_shape)
+        info_vec = self.info_vec.expand(batch_shape + (n,))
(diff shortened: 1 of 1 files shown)


fritzo:
What is required to implement `df < 2`? In particular many real-world time series exhibit df around 1.7-1.8.

fehiepsi:
@fritzo That condition is only used for moment-matching, where we need to match covariance of two student-ts. For example, given `a`, `b`, we want to find `c` such that St(b, 0, c * Id) ~ St(a, 0, Id). I believe that we can relax that condition as long as we have a better approximation.

In https://arxiv.org/abs/1703.02428 and https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=8822764 , the authors use KL for such approximation, hence `df > 2` is not required. However, using KL requires us to minimizing `KL(St(b, 0, c * Id), St(a, 0, Id))`, which is non-trivial. :(

fehiepsi:
I think that we can leverage tail behaviors of `St(b, 0, c * Id)` and `St(a, 0, Id)` to find `c` such that both distributions have the same 95% confidence interval... instead of using moment matching or kl as above. But I haven't come up with a formula yet.

fritzo:
When `df<2` the `df`-power moments still exist, so we could try to match those at least for `df >= 1`
```
mu = E[x]
sigma = (E|x - E[x]|^df)^(1/df)
```

fehiepsi:
Good idea!! I'll try to find the formula and match one of the moments < df.

fehiepsi:
This new paper https://arxiv.org/abs/1912.01607 computes general moments for Student's t-distribution. There are many choices for moment order, especially for multidimensional cases. I'll try to pick up a simple one from them.

pyro/distributions/multivariate_studentt.py
```diff
@@ -47,8 +47,7 @@ def covariance_matrix(self):
     @lazy_property
     def precision_matrix(self):
         identity = torch.eye(self.loc.size(-1), device=self.loc.device, dtype=self.loc.dtype)
-        scale_inv = identity.triangular_solve(self.scale_tril, upper=False).solution.transpose(-1, -2)
-        return torch.matmul(scale_inv.transpose(-1, -2), scale_inv)
+        return torch.cholesky_solve(identity, self.scale_tril)
```
fehiepsi: This is much more stable, which helps many tests.

pyro/ops/studentt.py
```diff
@@ -0,0 +1,324 @@
...
+    """
+    Non-normalized StudentT distribution:
+
+        StudentT(x) ~ Integral(p(x, s), s),
+
+    where
+
+        p(x, s) = GammaGaussian(log_normalizer, info_vec, precision, alpha, beta).
+
+    Note that different StudentT instances can have the same log density.
```
fritzo: Is this because `Studentt(l,i,p,a,b) = StudentT(l,i*t,p*t,a,b/t)` for any `t>0`? If so could you add that to this comment?

pyro/ops/studentt.py
```diff
@@ -0,0 +1,324 @@
...
+    def __init__(self, log_normalizer, info_vec, precision, alpha, beta, mask=None):
+        self.joint = GammaGaussian(log_normalizer, info_vec, precision, alpha, beta)
+        if mask is None:
+            self.mask = info_vec.new_full(info_vec.shape[-1:], True, dtype=torch.bool)
+        else:
+            assert mask.dim() == 1
+            self.mask = mask
+
+    @staticmethod
+    def from_joint(joint, mask=None):
```
fritzo: Can you add an `assert isinstance(joint, GammaGaussian)` to clarify?

pyro/ops/studentt.py
```diff
@@ -0,0 +1,324 @@
...
+        Marginalizing out variables on either side of the event dimension::
+
+            g.marginalize(left=n).event_logsumexp() = g.event_logsumexp()
+            g.marginalize(right=n).event_logsumexp() = g.event_logsumexp()
+
+        and for data ``x``:
+
+            g.condition(x).event_logsumexp()
+              = g.marginalize(left=g.dim() - x.size(-1)).log_density(x)
+        """
```
fritzo: Can you add `assert left == 0 or right == 0` to clarify usage restrictions?

pyro/ops/studentt.py
fehiepsi: Yes, it is (modulo alpha, beta reparameterization). I'll clarify it.

fehiepsi:
@fritzo Unfortunately, the implementation is only useful for sequential filtering. :( The reason is the `.condition` op does not asscociate with `.add`, that is `x + xy.condition(obs) != (x.event_pad(right=...) + xy).condition(y)`. A StudentTHMM can be implemented using StudentT op but parallel scan will be invalid. Do you think that this op is useful if we can't do parallel scan?

tests/ops/test_studentt.py
```diff
@@ -0,0 +1,351 @@
...
+    x, y = xy[..., :x_dim], xy[..., x_dim:]
+    y_pred = x.unsqueeze(-2).matmul(matrix).squeeze(-2)
+    actual_log_prob = st.log_density(xy)
+    expected_log_prob = y_mvt.log_prob(y - y_pred)
+    assert_close(actual_log_prob, expected_log_prob)
+
+
+@pytest.mark.parametrize("batch_shape", [(), (4,), (3, 2)], ids=str)
+@pytest.mark.parametrize("dim", [1, 2, 3])
+def test_studentt_multiple_representation(batch_shape, dim):
```
fehiepsi: @fritzo this test verifies multiple representations of a StudentT

fritzo:
> Do you think that this op is useful if we can't do parallel scan?

Well the main purpose of `Gaussian` and `GammaGaussian` was to implement parallel-scan algorithms, so maybe `StudentT` ops are not worth implementing.

What do you think about an alternative approach to `StudentTHMM`:
Following `StableHMM`, we could implement a `StudentTHMM` with `.rsample()` but no `.log_prob()` method, and implement auxiliary variable reparameterizers `StudentTReparam` and `StudentTHMMReparam` using scale-mixtures of `Normal` and `GaussiannHMM`, respectively. During inference we would learn (completely in parallel) the auxiliary Gamma-distributed "outlyingness" variables for each time step and observation step; then conditioned on "outlyingness" we could parallel-scan evaluate `GaussianHMM.log_prob()` or filter. Inference would then be exact usinng SVI, HMC, NUTS, etc. and would work for both latent and observed HMM observations. I think we would want to implement the `MultivariateStudentT` versions separately.

fehiepsi:
re not worth implementing: Agreed! I'll move some cleanups into a separate PR and leave the implementation here in case it is useful for benchmarking in the future.

re alternative approach to StudentTHMM inference: It is reasonable and should be easier to have than StableHMM. I think that it is also more useful than GammaGaussian, whose posteriors of the last state is more Gaussian-like for long time series with high dimensional outputs. I'll follow this approach instead.

fritzo:
> I'll follow this [StudentTHMM] approach instead.

Great! I think the code should be straight forward. I see the biggest open research question being:
"What kind of amortized guides are best for estimating the latent scale parameters"? I think we can share guide patterns across `StableHMM` and `StudentTHMM`.

fehiepsi:
I get your point. I'll think more about that question...

matthewfieger:
The reference for this pull request is interesting.  So is there a way to fit a t smoother with Pyro?  I am coming from the PyFlux library where, for example, I can fit a [dynamic linear model with student’s t noise](https://pyflux.readthedocs.io/en/latest/dyn_lin.html).  PyFlux is no longer being actively developed.  I would like to switch to Pyro to fit these types of models if possible.

fritzo:
Hi @matthewfieger , one thing we are working on is a `StudentTHMM` together with an auxiliary variable reparameterizer `StudentTReparam` similar to [StableHMMReparam](http://docs.pyro.ai/en/latest/infer.reparam.html#pyro.infer.reparam.stable.StableHMMReparam). This would allow `StudentT` processes to be represented as conditional `Gaussian` processes, permitting smoothing via SVI or HMC. This approach would be more accurate and more expensive than the moment-matching approach proposed in this PR.

Issue #2232, opened 2019-12-23
Where should pyro.reparam implementations live?


fritzo:
Regarding #2221 

Currently `pyro.reparam` supports only `Reparameterizer` instances. Since none of those implementations use any pyro primitives, they can live at the distributions level of abstraction, and live in the `pyro.distributions` module. However future `pyro.reparam` implementations may use pyro primitives, e.g. auto-decentering from [Gorinova,Moore,Hoffman 2019](https://arxiv.org/pdf/1906.03028) would require `pyro.param` statements; and condional dependencies among auxiliary variables would require `pyro.sample` statements (I plan to use this for a Levy-Ito decomposition).

Before next release we should decide on a proper place for `Reparameterizer`. Options include:
- Continue to live in `pyro.distributions` but use the `pyroapi` interface so as to be compatible with multiple backends.
- Move to `pyro.infer.reparam`.

I feel like reparameterizers are modeling gadgets that should be backend agnostic and not tied to the ppl language, hence the original placement in `pyro.distributions`. Yet they are simple and stable, so pyro.contrib doesn't seem appropriate.

cc @neerajprad

fritzo:
@fehiepsi @neerajprad Do you see any way to make this machinery generic across backends? (Even if we end up copying code among backends, to avoid dependencies)

fritzo:
After discussion with @neerajprad we believe the following is a good organization:
- pyro.distributions contains low-level distribution utilities
  - reparameterize
    - `Reparameterizer`
    - `LocScaleReparameterizer`
    - `TransformReparameterizer`
  - stable
    - `StableReparameterizer`
- pyro.poutine contains the handler
  - `poutine.reparam()`
- pyro.infer.reparam contains high-level inference effects built on `Reparameterizer` and `poutine.reparam()`
  - `decenter()`
  - `neutra()`
  - `levy_ito()`
  - `difference()`

PR #2235, opened 2019-12-24
Refactor reparameterization into pyro.infer.reparam

fritzo:
Addresses #2221 #2232 
Joint work with @fehiepsi and @neerajprad 

This refactors `pyro.reparam` and adds more examples of reparameterization:
1. Single-site `Transform` reparameterization as suggested by @fehiepsi.
2. Fully learnable decentering as in [(Gorinova et al. 2019)](https://arxiv.org/pdf/1906.03028.pdf). This is interesting because it creates new `pyro.param` sites inside the effect handler.
3. Stable reparameterization as in the original #2222. This is interesting because it introduces auxiliary variables, creating new `pyro.sample` statements inside the effect handler.
4. Symmetric Stable reparameterization. This is interesting because it creates an auxiliary random variable and then recurses to the `StableReparam`, and because the final distribution is not a `Delta` (rather it is a `Normal`) and can therefore be used for both latent and observed variables.
5. Symmetric stable reparameterization of a `StableHMM`. This is interesting because it calls other reparameterizers internally, and because the final distribution is a full `GaussianHMM`.
6. Multi-site `NeuTra` as suggested by @fehiepsi. This is interesting because the reparameterizer is stateful and is used across multiple sample sites.

After this PR, there is a single `pyro.reparam` poutine and a new module `pyro.infer.reparam` with the above six reparameterizers. Reparameterization is removed from `pyro.distributions`.

## Future plans
- Cumsum or banded-diagonal preconditioning for time series models.
- Levy-Ito decomposition of time series.

## Tasks
- [x] migrate existing tests
- [x] update docs (see [generated docs](http://fritzo.org/temp/pyro/html/infer.reparam.html))
- [x] implement a `StableHMM` (in a separate PR #2236)
- [x] test `LocScaleReparam`
- [x] test `TransformReparam`
- [x] test `StableReparam`
- [x] test `SymmetricStableReparam`
- [x] test `StableHMMReparam`
- [x] test `NeuTraReparam` (@neerajprad)


pyro/infer/reparam/transform.py
```diff
@@ -0,0 +1,42 @@
...
+    :param ~pyro.distributions.TorchDistribution base_dist: A base
+        distribution for the auxiliary latent variable.
+    :param ~torch.distributions.transforms.Transform transform: A bijective
+        transform defining forward and log abs det jacobian methods.
+    """
+    def __init__(self, base_dist, transform):
+        self.base_dist = base_dist
+        self.transform = transform
+
+    def __call__(self, name, fn, obs):
```
fritzo: The direction of x-z data dependency here depends on whether the site is observed or not, which might change due to another handler. This seems problematic for 'poutine.condition`, 'poutine.replay` etc.

pyro/infer/reparam/transform.py
```diff
@@ -0,0 +1,45 @@
...
+    [1] Hoffman, M. et al. (2019)
+        "NeuTra-lizing Bad Geometry in Hamiltonian Monte Carlo Using Neural Transport"
+        https://arxiv.org/abs/1903.03704
+
+    :param ~pyro.distributions.TorchDistribution base_dist: A base
+        distribution for the auxiliary latent variable.
+    :param ~torch.distributions.transforms.Transform transform: A bijective
+        transform defining forward and log abs det jacobian methods.
+    """
+    def __init__(self, base_dist, transform):
```
fehiepsi: I am not sure how useful this TransformReparam is, but I would prefer a TransformReparam which allows me to reparameterize the following site:
```
x = pyro.sample("x", TransformedDistribution(dist.Normal(0, 1), Affine(loc, scale)))
```
I expect the usage should be similar to `LocScaleReparam`.

fritzo: Good idea! I have updated as you suggest.

pyro/infer/reparam/cumsum.py
```diff
@@ -0,0 +1,28 @@
...
+    parameterization.
+    """
+    def __call__(self, name, fn, obs):
+        assert obs is None
+        assert fn.event_dim == 1
+        size = fn.event_shape[0]
+        scale = pyro.param("{}_scale", torch.tensor(1.),
+                           constraint=constraints.positive)
+        cumsum_fn = dist.Cauchy(0, scale).expand([size + 1])
+        cumsum_value = pyro.sample("{}_cumsum".format(name), cumsum_fn)
```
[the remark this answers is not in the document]
fritzo: I have updated the docs; let me know if you'd like further clarification. Note the new distribution doesn't really matter; its `.log_prob()` is ignored.

pyro/infer/reparam/neutra.py
```diff
@@ -0,0 +1,40 @@
...
+            # Sample a shared latent on first sample site.
+            posterior = self.guide.get_posterior()
+            x_unconstrained = pyro.sample("{}_latent".format(name), posterior)
+            log_density = -posterior.log_prob(x_unconstrained)
+            x_constrained = self.guide._unpack_latent(x_unconstrained)
+            self.x_constrained = list(reversed(x_constrained))
+
+        # Extract a single site's value from the shared latent.
+        site, value = self.x_unconstrained.pop()
+        log_density = log_density + fn.log_prob(value)
```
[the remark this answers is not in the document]
fritzo: Thanks for explaining, I think I understand now. Does this look correct now?

pyro/infer/reparam/cumsum.py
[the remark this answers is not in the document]
fehiepsi: Thanks! The implementation aligns with the docs, but I still don't understand why we need this reparam. Could you give me an example of "time series" and "increments" in the phrase "a time series is parameterized by increments"?

pyro/infer/reparam/neutra.py
```diff
@@ -0,0 +1,64 @@
...
+                                 "TransformedDistributions but got a posterior of type {}"
+                                 .format(type(posterior)))
+            t = dist.transforms.ComposeTransform(posterior.transforms)
+            z_unconstrained = pyro.sample("{}_latent".format(name),
+                                          posterior.base_dist.mask(False))
+
+            # Differentiably transform.
+            x_unconstrained = t(z_unconstrained)
+            log_density = t.log_abs_det_jacobian(z_unconstrained, x_unconstrained)
+            x_constrained = self.guide._unpack_latent(x_unconstrained)
```
fehiepsi: nit: I guess you mean `x_unconstrained_unpacked = self.guide._unpack_latent(x_unconstrained)`

pyro/infer/reparam/neutra.py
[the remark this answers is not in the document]
fehiepsi: I think that you are reparameterized the `guide` here instead of the `model`. IIUC, then we need to (assume that `fn` is latent of `model`) leverage the logic [here](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/autoguide/guides.py#L458).
```
site, unconstrained_value = self.x_unconstrained.pop()
transform = biject_to(fn.support)
value = transform(unconstrained_value)
logdet = transform.inv.log_abs_det_jacobian(value, unconstrained_value)
logdet = sum_rightmost(logdet, logdet.dim() - value.dim() + fn.event_dim)
log_density = log_density + fn.log_prob(value) + logdet
```

pyro/infer/reparam/transform.py
```diff
@@ -0,0 +1,31 @@
...
+    This reparameterization works only for latent variables, not likelihoods.
+    """
+    def __call__(self, name, fn, obs):
+        assert isinstance(fn, dist.TransformedDistribution)
+        assert obs is None
+
+        # Draw noise from the base distribution.
+        y = pyro.sample("{}_base".format(name), fn.base_dist)
+
+        # Differentiably transform.
```
fehiepsi: I think that we don't need to include `log_abs_det_jacobian` here. By reparameterizing (i.e. doing inference w.r.t. base value), we only need to include log_prob of base_dist.

pyro/infer/reparam/cumsum.py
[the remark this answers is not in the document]
fritzo: I have removed `CumsumReparam` because it wasn't exercising any novel aspect of the interface, and can safely be added to a follow-up PR.

pyro/infer/reparam/neutra.py
fritzo: oops, actually this line is correct but there is a typo below...

pyro/infer/reparam/transform.py
fritzo: Fixed.

pyro/infer/reparam/neutra.py
fehiepsi: But the `._unpack_latent` [method](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/autoguide/guides.py#L415) returns unconstrained values...

pyro/infer/reparam/neutra.py
```diff
@@ -0,0 +1,66 @@
...
+        log_density = 0.
+        if not self.x_constrained:  # On first sample site.
+            # Sample a shared latent.
+            posterior = self.guide.get_posterior()
+            if not isinstance(posterior, dist.TransformedDistribution):
+                raise ValueError("NeuTraReparam only supports guides whose posteriors are "
+                                 "TransformedDistributions but got a posterior of type {}"
+                                 .format(type(posterior)))
+            t = dist.transforms.ComposeTransform(posterior.transforms)
+            z_unconstrained = pyro.sample("{}_latent".format(name),
```
neerajprad: The name here will correspond to the first site's name, which could cause confusion. Can we rename this to `shared_latent` or something else?

pyro/infer/reparam/neutra.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    def __init__(self, guide):
+        if not isinstance(guide, AutoContinuous):
+            raise TypeError("NeuTraReparam expected an AutoContinuous guide, but got {}"
+                            .format(type(guide)))
+        self.guide
+        self.x_constrained = []
+
+    def __call__(self, name, fn, obs):
+        log_density = 0.
```
neerajprad: IIUC correctly, if we are to use this directly with HMC/NUTS, we will also probably need a `sample` statement corresponding to the `obs`, maybe something like 

```python
if obs is not None:
    return fn, obs
```

pyro/infer/reparam/neutra.py
```diff
@@ -0,0 +1,66 @@
...
+                                          posterior.base_dist.mask(False))
+
+            # Differentiably transform.
+            x_unconstrained = t(z_unconstrained)
+            log_density = t.log_abs_det_jacobian(z_unconstrained, x_unconstrained)
+            x_constrained = self.guide._unpack_latent(x_unconstrained)
+            self.x_constrained = list(reversed(x_constrained))
+
+        # Extract a single site's value from the shared latent.
+        site, value = self.x_constrained.pop()
```
neerajprad: Could we add an assert to ensure that the popped site's name is the same as name argument passed to `__call__`?

pyro/infer/reparam/neutra.py
neerajprad: I think we'll need to constrain and accumulate the jacobian adjustment terms into `log_density`.

pyro/infer/reparam/neutra.py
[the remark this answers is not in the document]
fritzo: done

pyro/infer/reparam/neutra.py
fritzo: Good point. I've added the following early exit:
```py
if name not in self.guide.prototype_trace.nodes:
    return fn, obs
```
Note that most reparameterizers do not support observation statements, and so I've asserted
```py
assert obs is None, "ThisReparam does not support observe statements"
```
The reason I'm opting to error rather than be permissive is to address the mismatch of behavior between the `obs=___` kwarg and `poutine.condition`. Due to poutine stack ordering, reparameterizers cannot know about `poutine.condition`s higher in the stack, but they do know about `obs=___` kwargs. For most reparameterizers `poutine.condition` will result in infinite loss, which is an easy-to-identify problem. Because the `poutine.condition` case is problematic, I have chosen to make the `obs=___` case also problematic (raising `AssertionError`), so that if users switch between the two idioms they will see identical `Reparam` behavior (i.e. either both good or both bad).

pyro/infer/reparam/neutra.py
fritzo: I can rename it to `{}_shared_latent`, but we do need some prefix to disambiguate multiple reparameterizers.

pyro/infer/reparam/neutra.py
neerajprad: Thanks for explaining. We use [poutine.condition](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/mcmc/util.py#L257) to compute the potential energy for samples from the integrator. Will that be problematic? 

```python
        cond_model = poutine.condition(self.model, params_constrained)
        model_trace = poutine.trace(cond_model).get_trace(*self.model_args,
                                                          **self.model_kwargs)
```

fritzo: It should be fine for MCMC to use of `poutine.condition` to set latent variables, since those latent variables will be uneffected by `poutine.reparam` statements (they are either not reparameterized sites, or they are auxiliary sites created by `poutine.reparam` and themselves are not reparameterized).

The only issue is when `poutine.condition` is used for likelihoods: reparameterization often results in a `Delta` simulating a `pyro.deterministic` and would lead to infinite loss when using `poutine.condition`. Therefore most reparameterizers do not support likelihoods (exceptions being `SymmetricStableReparam` and `StableHMMReparam` which produce `Normal`s rather than `Delta`s).

pyro/infer/reparam/levy_ito.py
```diff
@@ -0,0 +1,56 @@
...
+        init_shocks = pyro.sample("{}_init_shocks".format(name),
+                                  dist.Pareto(1, self.stability))
+        init_dist = dist.Normal(self.init_dist.scale,
+                                self.init_dist.loc + init_shocks).to_event(1)
+
+        # Sample shocks to latent state.
+        length = self.length
+        shock_times = pyro.sample("{}_trans_shock_times".format(name),
+                                  dist.Uniform(0, length).to_event(2))
+        # FIXME we need both positive and negative shocks.
```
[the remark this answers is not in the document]
fritzo: I have added a docstring and moved this into pyro/infer/reparam/stable.py

pyro/infer/reparam/neutra.py
```diff
@@ -0,0 +1,78 @@
...
+    Example usage::
+
+        # Step 1. Train a guide
+        guide = AutoIAFNormal(model)
+        svi = SVI(model, guide, ...)
+        # ...train the guide...
+
+        # Step 2. Use trained guide in NeuTra MCMC
+        neutra = NeuTraReparam(guide)
+        model = poutine.infer_config(lambda _: {"reparam": neutra})
```
neerajprad: This is probably missing the first argument - `model` to `infer_config`.

pyro/infer/reparam/neutra.py
```diff
@@ -0,0 +1,78 @@
...
+                                 "TransformedDistributions but got a posterior of type {}"
+                                 .format(type(posterior)))
+            t = dist.transforms.ComposeTransform(posterior.transforms)
+            z_unconstrained = pyro.sample("{}_shared_latent".format(name),
+                                          posterior.base_dist.mask(False))
+
+            # Differentiably transform.
+            x_unconstrained = t(z_unconstrained)
+            log_density = t.log_abs_det_jacobian(z_unconstrained, x_unconstrained)
+            self.x_unconstrained = list(reversed(self.guide._unpack_latent(x_unconstrained)))
```
neerajprad: I think we'll need to do `list(reversed(list(..)` here.

pyro/infer/reparam/neutra.py
neerajprad: Thanks, that makes sense!

pyro/infer/reparam/neutra.py
fritzo: reversed works on iterators

neerajprad: That's interesting. It doesn't work with generators though (I guess because it can be an infinite stream). 😞

fritzo: oh sorry. Could  you make the change, taking temporary ownership of this file while you write tests? (sorry this PR is so big, I just wanted to get the interface right, and the best way to do that was to bang on it from every angle at once)

neerajprad: Sounds good! I'll handle any minor changes (non-interface related) to get the test working.

pyro/infer/reparam/stable.py
```diff
@@ -0,0 +1,198 @@
...
+        one = proto.new_ones(proto.shape)
+        u = pyro.sample("{}_uniform".format(name),
+                        self._wrap(dist.Uniform(-half_pi, half_pi), event_dim))
+        e = pyro.sample("{}_exponential".format(name),
+                        self._wrap(dist.Exponential(one), event_dim))
+
+        # Differentiably transform to scale drawn from a totally-skewed stable variable.
+        _, z = _unsafe_standard_stable(fn.stability / 2, 1, u, e)
+        assert (z >= 0).all()
+        scale = fn.scale * z * (math.pi / 4 * fn.stability).cos().pow(2 / fn.stability)
```
fehiepsi: I think we should use `( 2 * z) ** 0.5` instead of `z` here (if z is a mixing variable). Btw, what is the role of `(math.pi / 4 * fn.stability).cos().pow(2 / fn.stability)`? I can't find relevant information in Proposition 3 of reference [1] (probably you used a different reference?).

pyro/infer/reparam/stable.py
```diff
@@ -0,0 +1,198 @@
...
+                                  dist.Pareto(1, stability.unsqueeze(-1).unsqueeze(-1).unsqueeze(-1))
+                                      .expand([2, self.num_shocks, hidden_dim])
+                                      .to_event(3))
+        pos_shock_times, neg_shock_times = shock_times.unbind(-3)
+        pos_shock_sizes, neg_shock_sizes = shock_sizes.unbind(-3)
+        pos_shocks = _densify_shocks(duration, pos_shock_times, pos_shock_sizes)
+        neg_shocks = _densify_shocks(duration, neg_shock_times, neg_shock_sizes)
+        # FIXME correctly scale pos_shocks and neg_shocks
+        shocks = pos_shocks - neg_shocks
+        # FIXME correct scale and loc
```
fehiepsi: I think that I understand your purpose here, but I can't find "shocks"-related references to verify the corresponding math. Section 1.2.6 of https://people.bath.ac.uk/ak257/book2e/book2e.pdf gives me no hint. :(

fritzo: For alternate explanations, see
- [wikipedia](https://en.wikipedia.org/wiki/L%C3%A9vy_process#L%C3%A9vy%E2%80%93It%C3%B4_decomposition)
- Theorem 4 in [Barndorff-Nielsen & Shepard](https://pdfs.semanticscholar.org/fe80/07ba98fafa23ddca98ac5d9b1cf1732f70bd.pdf)
- Jim Pitman's [lecture notes](https://www.stat.berkeley.edu/~pitman/s205s03/levy.pdf) which follow Kallenberg closely but in more readable syntax.

By "shocks", I mean the compound Poisson process component of the Levy-Ito representation. This is defined by $\Pi(dx)$ in the linked [Kyprianou (2013)](https://people.bath.ac.uk/ak257/book2e/book2e.pdf) and wikipedia, defined by (N, v(dy)) in Barndorff-Nielsen, and $\nu$ in Pitman's notes. Note the Levy-Ito decomposition of stochastic processes is closely related to the Levy-Khintchine theorem about characteristic functions of those processes. It seems like applied finance folks use the term "shocks", whereas probabilists use the term "jumps"; I can switch to "jumps" if you think that's clearer.

pyro/infer/reparam/stable.py
fritzo: Oops sorry wrong link 😊 I've fixed, [here it is](https://people.maths.ox.ac.uk/~howison/papers/levy.pdf
). Note that reference uses a different parameter ordering: `S_stability(scale, skew, loc)`. Let me know if you find a clearer source of this theorem. I really enjoyed that paper, it is very clear.

fehiepsi: :D In that paper, the mixing "scale" (not variance) will be `z ** 0.5`. In addition, Stable(2, 0, k, ...) is Normal(..., sqrt(2) * k), so as mentioned in my last comment, we should use `( 2 * z) ** 0.5` instead of `z` of `z` here.

Similarly, for the latter term, we should use `pow(1 / fn.stability)` instead of `pow(2 / fn.stability)`.

tests/distributions/test_stable.py
```diff
@@ -62,3 +63,13 @@ def cdf(x):
...
     stat, pvalue = kstest(sampler, cdf, N=num_samples)
     assert pvalue > 0.1, pvalue
+
+
+@pytest.mark.parametrize("loc", [0, 1, -1, 2, 2])
+@pytest.mark.parametrize("scale", [0.5, 1, 2])
+def test_normal(loc, scale):
+    num_samples = 100000
+    expected = dist.Normal(loc, scale).sample([num_samples])
+    actual = dist.Stable(2, 0, scale * 0.5 ** 0.5, loc).sample([num_samples])
```
fritzo: @fehiepsi `Stable(2, 0, scale * sqrt(2), loc) = Normal(loc, scale)`

fehiepsi: Yeah, I know about it. I guess you mean `scale / sqrt(2)` here?

pyro/infer/reparam/stable.py
fritzo: Thanks, I have updated to use `z.sqrt()`, but there is still a bug. Can you explain your `.pow(1/stability)` suggestion? I believe `.pow(2/stability)` is the correct translation of `cos(...).pow(alpha' / alpha)`.

fehiepsi: I think `.pow(2 / stability)` is the scale of `Y` in the above reference, and we have Z = Y^1/2 X. So we need to multiply `scale` with `...pow(1/stability)`

fehiepsi: Re bug: If there is still a bug, could you try to completely remove the latter term? In [this reference](https://arxiv.org/pdf/1604.07487.pdf) (at the end of page 8), IIUC there is no such latter term (though I think that your reference is correct).

tests/distributions/test_stable.py
fritzo: The test passes as-is. Our implementation agrees with scipy.stats. Wikipedia and scipy.stats disagree by a factor of 1.

pyro/infer/reparam/stable.py
fritzo: Thanks for explaining, that seems to fix the test!

pyro/infer/reparam/stable.py
fritzo: OK, I have renamed "shocks" to "jumps" to be more standard.

tests/distributions/test_stable.py
fehiepsi: Hmm, in the implementation, `scale * 0.5 ** 0.5` will be equal to `scale / sqrt(2)`. Similarly, in stable implementation, you divide scale to `sqrt(0.5)`, which is equivalent to multiply with `sqrt(2)`. Both are equivalent IMO. The implementation here agrees with wikipedia, so if scipy is different, probably scipy is wrong (assuming the tests pass)?

fritzo: Yes yes, I simply misread my own code 😊 Thanks for pointing out.

fritzo:
@fehiepsi @neerajprad I think this is now finished. I have removed `CumsumReparam` and the Levy-Ito decomposition from this PR, and will address them in subsequent PRs. The `StableHMMReparam` now simply uses `SymmetricStableReparam` under the hood, but its interface is forwards compatible with a future `num_jumps` arg to allow optional Levy-Ito decomposition.

pyro/infer/reparam/stable.py
fritzo: I have removed the Levy-Ito decomposition for now, and will implement it in a subsequent PR.

pyro/infer/reparam/neutra.py
fritzo: Renamed to `{}_shared_latent`.

pyro/infer/reparam/neutra.py
```diff
@@ -0,0 +1,104 @@
...
+        if not self.x_unconstrained:  # On first sample site.
+            # Sample a shared latent.
+            posterior = self.guide.get_posterior()
+            if not isinstance(posterior, dist.TransformedDistribution):
+                raise ValueError("NeuTraReparam only supports guides whose posteriors are "
+                                 "TransformedDistributions but got a posterior of type {}"
+                                 .format(type(posterior)))
+            self.transform = dist.transforms.ComposeTransform(posterior.transforms)
+            z_unconstrained = pyro.sample("{}_shared_latent".format(name),
+                                          posterior.base_dist.mask(False))
```
fehiepsi: Interesting! @neerajprad I guess `.mask(False)` is a nice pattern for improper prior (with initial values of MCMC can be obtained randomly through init_strategy, instead of through the fixed value specified in `param` statement).

pyro/infer/reparam/stable.py
```diff
@@ -0,0 +1,167 @@
...
+        assert (z >= 0).all()
+        scale = fn.scale * (2 ** 0.5) * (math.pi / 4 * fn.stability).cos().pow(1 / fn.stability) * z.sqrt()
+        scale = scale.clamp(min=torch.finfo(scale.dtype).tiny)
+
+        # Construct a scaled Gaussian.
+        new_fn = self._wrap(dist.Normal(fn.loc, scale), event_dim)
+        return new_fn, obs
+
+
+def _densify_jumps(duration, times, sparse_jumps):
```
fehiepsi: I think you can remove this.

pyro/infer/reparam/neutra.py
```diff
@@ -0,0 +1,104 @@
...
+            x_unconstrained = self.transform(z_unconstrained)
+            log_density = self.transform.log_abs_det_jacobian(z_unconstrained, x_unconstrained)
+            self.x_unconstrained = list(reversed(list(self.guide._unpack_latent(x_unconstrained))))
+
+        # Extract a single site's value from the shared latent.
+        site, unconstrained_value = self.x_unconstrained.pop()
+        assert name == site["name"], "model structure changed"
+        transform = biject_to(fn.support)
+        value = transform(unconstrained_value)
+        logdet = transform.inv.log_abs_det_jacobian(value, unconstrained_value)
```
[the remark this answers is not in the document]
fritzo: done

PR #2237, opened 2019-12-28
Fix autoguide latent unpacking with batched latent samples

neerajprad:
This fixes an issue with autoguide `_unpack_latent` method where unpacking might not work with batching. Needed for Neutra reparametrizer in #2235.

### DIFF
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -26,7 +26,7 @@ import pyro
 import pyro.distributions as dist
 import pyro.distributions.transforms as transforms
 import pyro.poutine as poutine
-from pyro.distributions.util import broadcast_shape, eye_like, sum_rightmost
+from pyro.distributions.util import eye_like, sum_rightmost
 from pyro.infer.autoguide.initialization import InitMessenger, init_to_median
 from pyro.infer.autoguide.utils import _product
 from pyro.infer.enum import config_enumerate
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -421,13 +421,9 @@ class AutoContinuous(AutoGuide):
         batch_shape = latent.shape[:-1]  # for plates outside of _setup_prototype, e.g. parallel particles
         pos = 0
         for name, site in self.prototype_trace.iter_stochastic_nodes():
-            constrained_shape = site["value"].shape
             unconstrained_shape = self._unconstrained_shapes[name]
             size = _product(unconstrained_shape)
-            event_dim = site["fn"].event_dim + len(unconstrained_shape) - len(constrained_shape)
-            unconstrained_shape = broadcast_shape(unconstrained_shape,
-                                                  batch_shape + (1,) * event_dim)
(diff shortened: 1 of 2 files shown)


pyro/infer/autoguide/guides.py
```diff
@@ -421,13 +421,9 @@ def _unpack_latent(self, latent):
         batch_shape = latent.shape[:-1]  # for plates outside of _setup_prototype, e.g. parallel particles
         pos = 0
         for name, site in self.prototype_trace.iter_stochastic_nodes():
-            constrained_shape = site["value"].shape
             unconstrained_shape = self._unconstrained_shapes[name]
             size = _product(unconstrained_shape)
-            event_dim = site["fn"].event_dim + len(unconstrained_shape) - len(constrained_shape)
```
neerajprad: I may be missing context as to why we need this, so please let me know if I have oversimplified! This seems to pass existing tests and a new one that I added.

fehiepsi: The change looks fine to me but I might miss some context too.

tests/infer/test_autoguide.py
```diff
@@ -449,12 +449,16 @@ def model():
...
-        return pyro.sample('x', dist.LKJCorrCholesky(2, torch.tensor(1.)))
+        with pyro.plate('N', 5):
+            return pyro.sample('x', dist.LKJCorrCholesky(2, torch.tensor(1.)))
 
     guide = AutoDiagonalNormal(model)
     assert guide()['x'].shape == model().shape
     latent = guide.sample_latent()
-    assert list(guide._unpack_latent(latent))[0][1].shape == (1,)
+    assert list(guide._unpack_latent(latent))[0][1].shape == (5, 1)
+    with pyro.plate('N', 100):
```
neerajprad: After looking at some of the failing autoguide tests in predictive, I was convinced that what we have is fine. To get this test to pass we will need to adjust this plate by `max_plate_dim` - `pyro.plate('N', 100, dim=-2)`. While `Predictive` does this by default, `MCMC.get_samples` does not consider `max_plate_dim`, so maybe we should change that so that those samples can be passed directly to other utilities like `guide._unpack_latent` without adjusting the sample dim.

neerajprad:
@fehiepsi - Thanks for taking a look. While the tests in `test_autoguide` all passed, this change has issues when we embed the model inside another plate, e.g. in `Predictive` for vectorized sampling. I finally concluded that this wouldn't be an issue if we just ensure that batch dims don't collide with plate dims in the model. I was directly using the samples from `mcmc.get_samples` which caused issues with `_unpack_latent`, but I think changing the behavior of `mcmc.get_samples()` may be a more fruitful discussion instead.

neerajprad:
> Do I understand correctly that this is an issue for utilities that generate samples of shape (num_samples, 1, 1, 1) + event_shape where the (1, 1, 1) is needed to avoid plates?

That's right, that is exactly the issue. 

> This seems to be a recurring issue. Maybe it is time to think about named dimension and whether we could adopt limited use of named dimensions to resolve issues like this.

Interesting idea. I think if we have some consistent naming practice for batch dims, we could use that information for alignment.

PR #2244, opened 2020-01-06
Fixed `ConditionedPlanar` bug to set `event_dim`

### LINKED ISSUE #2147: ConditionedPlanarFlow's event_dim is wrong [bug]
--------------------------------------------------------------------------------------------------
### Issue Description
Provide a brief description of the issue.

I think the ConditionedPlanarFlow's event_dim is incorrect. I think the class needs to be set as such:
```

class ConditionedPlanarFlow(Transform):
   event_dim=1 #this line is not in the current release version
    def __init__(self, bias=None, u=None, w=None):
        super(ConditionedPlanarFlow, self).__init__(cache_size=1)
        self.bias = bias
        self.u = u
        self.w = w
        self._cached_logDetJ = None

```
The Transform super class default is event_dim=0.

Maybe I misunderstand this class, but I notice the PlanarFlow class has event_dim=1, as do many of the other normalizing flows and when I tried using this flow for my models my ELBO would go negative. I think the confusion is that ConditionalPlanarFlow has the event_dim=1, but that isn't setting it when it the actual flows are created (from what I can tell anyways).

### Environment
 - Found on: Ubuntu 
 - PyTorch version: I think I was using 1.2 ...but again this is a super minor bug
You can see it in the the pyro docs http://docs.pyro.ai/en/stable/_modules/pyro/distributions/transforms/planar.html#ConditionalPlanarFlow

 - Pyro version: 3.6.10 (I think...pretty sure a single boolean value is pretty robust to version though)
Was originally found in...nightly build (not sure on term... I built from sourceb asically)

stefanwebb:
Resolves #2147

### DIFF
--- a/pyro/distributions/transforms/planar.py
+++ b/pyro/distributions/transforms/planar.py
@@ -14,6 +14,11 @@ from pyro.nn import DenseNN
 
 @copy_docs_from(Transform)
 class ConditionedPlanar(Transform):
+    domain = constraints.real
+    codomain = constraints.real
+    bijective = True
+    event_dim = 1
+
     def __init__(self, bias=None, u=None, w=None):
         super(ConditionedPlanar, self).__init__(cache_size=1)
         self.bias = bias


pyro/distributions/transforms/planar.py
```diff
@@ -14,6 +14,11 @@
 
 @copy_docs_from(Transform)
 class ConditionedPlanar(Transform):
+    domain = constraints.real
```
fritzo: we may want to change this to `constraints.real_vector` at some point, but this is fine for now

PR #2068, opened 2019-10-18
Added 1x1 invertible convolution as a transform (Glow, Kingma 2018)

bmazoure:
This PR implements 1x1 invertible convolutions (which boils down to multiple `nn.Linear` stacked together, but nevertheless useful when mixed with `AffineCoupling`) from [this paper](https://arxiv.org/abs/1807.03039), Eq. 9.

Issue #2066, opened 2019-10-18
Support for invertible convolution transforms


bmazoure:
Seeing that almost all normalizing flow models are implemented under `pyro.distributions.transforms`, would it be possible (in theory) to support convolutional invertible layers (see [here](https://arxiv.org/abs/1807.03039) )? Similarly to `InvertibleAutoregressiveFlow`, which would take a convolutional layer and compute the forward/backward passes together with the log_det_jacobian.
Just wondering if there would be any hidden issues that would arise if I was to implement it.

martinjankowiak:
no, i don't see any reason why it couldn't be added to `pyro.distributions.transforms`. contributions welcome!

cc @stefanwebb

PR #2246, opened 2020-01-06
Invertible 1x1 convolution

stefanwebb:
I've continued the feature request of @bmazoure in #2066 / #2068, changing a number of the implementation details, adding tests, docs, etc.

This should be good to go for implementing Glow normalizing flow, and Neural Spline FLow


pyro/distributions/transforms/generalized_channel_permute.py
```diff
@@ -0,0 +1,164 @@
...
+        # Unflatten x
+        return x.view_as(y)
+
+    def log_abs_det_jacobian(self, x, y):
+        """
+        Calculates the elementwise determinant of the log Jacobian, i.e. log(abs(det(dy/dx))).
+        """
+
+        h, w = x.shape[-2:]
+        log_det = h * w * torch.log(torch.abs(torch.diag(self.LU))).sum()
```
fritzo: The `.diag()` and `.sum()` look suspicious. If you intend to support batching (e.g. for multi-particle SVI), you'll need to restrict the `.sum()` to relevant dimensions, e.g.
```py
log_det = h * w * self.LU.diagonal(dim1=-2, dim2=-1).abs().log().sum((-3, -2, -1))
```

pyro/distributions/transforms/generalized_channel_permute.py
```diff
@@ -0,0 +1,164 @@
...
+
+        Invokes the bijection x=>y; in the prototypical context of a
+        :class:`~pyro.distributions.TransformedDistribution` `x` is a sample from the base distribution (or the output
+        of a previous transform)
+        """
+
+        # Extract the lower and upper matrices from the packed LU matrix
+        U = self.LU.triu()
+        L = self.LU.tril()
+        L.diagonal(dim1=-2, dim2=-1).fill_(1)
```
fritzo: You might want to avoid in-place ops here and elsewhere, e.g.
```py
L = self.LU.tril(-1) + torch.eye(self.LU.size(-1), dtype=self.LU.dtype, device=self.LU.device)
```
Actually I would be surprised if gradients worked with the in-place version. Do you have smoke tests ensuring gradients work?

tests/distributions/test_transforms.py
```diff
@@ -136,7 +136,7 @@ def test_poly_jacobians(self):
 
     def test_householder_inverses(self):
         for input_dim in [2, 5, 10]:
-            self._test_inverse(input_dim, T.householder(input_dim, count_transforms=2))
+            self._test_inverse([input_dim], T.householder(input_dim, count_transforms=2))
```
fritzo: This probably needs to be a tuple or `torch.Size` e.g. `(input_dim,)`.

stefanwebb: I've realized it works if I leave `input_dim` as is in all the other test methods, and use `shape` for `GeneralizedChannelPermute`

pyro/distributions/transforms/generalized_channel_permute.py
stefanwebb: This will work for the non-conditional version of `GeneralizedChannelPermute`. I will revisit this point when implementing the conditional version!

stefanwebb: I.e. it will work correctly for batching of different sizes for the non-conditional transform. See the following line where the `log_det` factor multiplies a vector of ones in the batch shape.

pyro/distributions/transforms/generalized_channel_permute.py
stefanwebb: There is no error when I call `backward()` through a loss involving this transform and a transformed distribution, although I'll use your idiom to be safe

stefanwebb:
Okay, I think I've addressed all the points in your review, @fritzo :)

Issue #2249, opened 2020-01-10
Switch from MIT license to Apache-2.0 license


fritzo:
As part of Pyro's move from Uber to the Linux Foundation, we are replacing the existing
> MIT license + Contributor License Agreement

with a simpler structure:

> Apache 2.0 license + no CLA

## Tasks

- [x] #2250 Include copyright notices on each file: `Copyright (c) 2017-2019 Uber Technologies, Inc.`
- [x] #2250 Include an SPDX short-form identifier in a comment at the top of each file in the repo, wherever reasonably possible. (addresses #2067 )
- [x] #2250 Switch from MIT to Apache-2.0 license
- [x] Disable Uber CLA

swinslow:
Hi @fritzo, on the SPDX identifiers, https://spdx.org/ids has more details about this (or I'm happy to weigh in if you have questions!)

On the copyright notices, another option that some of our projects have gone with has been to use a more general form of copyright notice, such as: "Copyright Contributors to the Pyro project." We actually just posted a blog about this [earlier today](https://www.linuxfoundation.org/blog/2020/01/copyright-notices-in-open-source-software-projects/) — noting it here in case Pyro wants to take a similar approach.

fritzo:
Thanks, @swinslow, can I include this line in each source file?
> Copyright (c)  2017-2018 Contributors to the Pyro project.

swinslow:
From the LF's perspective, I think that would be fine, though a couple of comments:

* You might consider omitting the years range. Including it tends to result in questions about e.g. "when do we need to update it in each file?" or "what if contributions were made in 2019 but it's not in the year range?" Because of those sorts of things, we've increasingly seen (and recommended) that projects may just omit year ranges.

* Similarly with the (c) symbol — fine to include if you like and no harm in doing so, though we've seen some projects omit it. Mostly because omitting it helps avoid the question of whether e.g. redistributions need to include multiple different lines in attribution notices, when different variations on that symbol inevitably appear — e.g. `(c)` vs. `(C)` vs. `Ⓒ` vs. `&copy;` vs...

* Lastly, just to note, you may also want to check with your own legal counsel about whether they have a preference for the notice format, since in any event you / they will be retaining ownership of the copyrights (no matter what notice you put on it!) I'm not able to provide you with legal advice on this, unfortunately — just trying to be transparent on that  :)

PR #2250, opened 2020-01-10
Switch to Apache 2.0 license

fritzo:
Addresses #2249 #2067 

1. Switch from MIT to Apache-2.0 license
    - update LICENSE.md
    - update setup.py
2. Add a script to add copyright notices and SPDX identifiers to all .py and .cpp files.
3. Apply the script to all source files

I plan to update and modify the script as we e.g. update the copyright year next year or add new file types.

After this PR merges we can disable the Uber CLA 🎉 

@neerajprad we will also need to switch licenses in other repos:
- https://github.com/pyro-ppl/numpyro
- https://github.com/pyro-ppl/funsor
- https://github.com/pyro-ppl/pyro-api

### DIFF
--- a/scripts/update_headers.py
+++ b/scripts/update_headers.py
@@ -1,0 +1,52 @@
+# Copyright (c) 2020 Contributors to the Pyro project.
+import os
+import glob
+
+root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+blacklist = ["/build/", "/dist/"]
+file_types = [
+    ("*.py", "# {}"),
+    ("*.cpp", "// {}"),
+]
+
+for basename, comment in file_types:
+    copyright_line = comment.format("Copyright (c) 2017-2020 Contributors to the Pyro project.\n")
+    # See https://spdx.org/ids-how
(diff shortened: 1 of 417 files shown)


fritzo:
cc @swinslow @ibrahimhaddad

scripts/update_headers.py
```diff
@@ -0,0 +1,54 @@
+# Copyright (c) 2020 Contributors to the Pyro project.
```
fritzo: @neerajprad here is the new script file

neerajprad: Thanks, @fritzo. Should we do this for the other projects in pyro-ppl too?

fritzo: Yes, we should do this for all repos in https://github.com/pyro-ppl.

neerajprad: Okay, I'll make the changes for numpyro and pyro-api.

fritzo:
@neerajprad I've updated to use @swinslow's recommended simpler copyright notices.

fritzo:
@neerajprad I reverted to preserve Uber copyrights on existing files; the `scripts/update_headers.py` can remain as-is, since we can use @swinslow's simplified copyright notice for new files.

PR #2252, opened 2020-01-12
Discrete Cosine reparameterizer

### LINKED ISSUE #2221: [FR] Automatic reparameterization

fritzo:
Resolves #2221 
Blocking #2253 
Joint work with @fehiepsi 

This implements a Discrete Cosine Transform reparameterizer, intended for use in time series models. The motivating model class is time series models where it is natural to specify a model in terms of increments (e.g. a Gaussian or StudentT or Stable process), but where that parameterization is extremely poorly conditioned, leading to heavily non-diagonal posterior covariances. By reparameterizing in the frequency domain we can nearly (sometimes exactly) diagonalize the posterior covariance.

Empirically this seems to work well for the stochastic volatility model in #2253, where log-volatility is a latent gaussian process. Using an `AutoDiagonalNormal` guide I get cleaner posterior median and lower loss.

## Tasks
- [x] implement low-level ops `dct` and `idct`
    @fritzo initially added torch-dct as a dependency.
    @fehiepsi then replaced this with a custom implementation.
- [x] test on a time series model, e.g. stochastic volatility #2253

### DIFF
--- a/pyro/ops/tensor_utils.py
+++ b/pyro/ops/tensor_utils.py
@@ -138,17 +138,17 @@ def idct(x):
     N = x.size(-1)
     scale = torch.cat([x.new_tensor([math.sqrt(N)]), x.new_full((N - 1,), math.sqrt(0.5 * N))])
     x = x * scale
-    # Step 1, solve X = cos(k) * Yr - sin(k) * Yi
+    # Step 1, solve X = cos(k) * Yr + sin(k) * Yi
     # We know that Y[1:] is conjugate to Y[:0:-1], hence
     # X[:0:-1] = sin(k) * Yr[1:] + cos(k) * Yi[1:]
     # So Yr[1:] = cos(k) * X[1:] + sin(k) * X[:0:-1]
-    # and Yi[1:] = -sin(k) * X[1:] + cos(k) * X[:0:-1]
+    # and Yi[1:] = sin(k) * X[1:] - cos(k) * X[:0:-1]
     # In addition, Yi[0] = 0, Yr[0] = X[0]
-    # In other words, Y = complex_mul(e^-ik, X + i[0, X[:0:-1]])
-    xi = torch.nn.functional.pad(x[..., 1:], (0, 1)).flip(-1)
+    # In other words, Y = complex_mul(e^ik, X - i[0, X[:0:-1]])
+    xi = torch.nn.functional.pad(-x[..., 1:], (0, 1)).flip(-1)
     X = torch.stack([x, xi], dim=-1)
     coef_real = torch.cos(torch.linspace(0, 0.5 * math.pi, N + 1))
-    coef = torch.stack([coef_real[:-1], -coef_real[1:].flip(-1)], dim=-1)
+    coef = torch.stack([coef_real[:-1], coef_real[1:].flip(-1)], dim=-1)
     Y = _complex_mul(coef, X)
     # Step 2
     y = torch.irfft(Y, 1, onesided=False)
--- a/pyro/ops/tensor_utils.py
+++ b/pyro/ops/tensor_utils.py
@@ -92,6 +92,7 @@ def dct_ii(x):
 
     :param Tensor x:
     """
+    return x  # FIXME
     raise NotImplementedError("TODO")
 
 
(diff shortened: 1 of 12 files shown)


fehiepsi:
@fritzo You can use the following version.

```
def _real_of_complex_mul(a, b):
    ar, ai = a.unbind(-1)
    br, bi = b.unbind(-1)
    return ar * br - ai * bi

def dtc(x):
    # Ref: http://fourier.eng.hmc.edu/e161/lectures/dct/node2.html
    # Step 1
    y = torch.cat([x[..., ::2], x[..., 1::2].flip(-1)], dim=-1)
    # Step 2
    Y = torch.rfft(y, 1, onesided=False)
    # Step 3
    coef_real = torch.cos(torch.linspace(0, 0.5 * math.pi, x.size(-1) + 1))
    coef = torch.stack([coef_real[:-1], -coef_real[1:].flip(-1)], dim=-1)
    # XXX: scipy.fftpack.dct has a factor of 2
    return _real_of_complex_mul(coef, Y)


def idtc(x):
    # Step 1
    coef_real = torch.cos(torch.linspace(0, 0.5 * math.pi, x.size(-1) + 1))
    coef = torch.stack([coef_real[:-1], coef_real[1:].flip(-1)], dim=-1)
    Y = x.unsqueeze(-1) * coef
    # Step 2
    y = torch.irfft(Y, 1, onesided=False)
    # Step 3
    return torch.stack([y, y.flip(-1)], axis=-1).reshape(x.shape[:-1] + (-1,))[..., :x.size(-1)]
```

fritzo:
@fehiepsi does that guarantee orthogonality, so that the Jacobian is 1? Do you think we should add those instead of depending on torch-dct?

fehiepsi:
Oh, good point. We need that for jacobian. I will edit it soon.

fritzo:
@fehiepsi If you'd like to use those versions, can you push to this branch and add a test against `scipy.fftpack.dct(..., norm="ortho")`?

pyro/ops/tensor_utils.py
```diff
@@ -110,3 +110,48 @@ def repeated_matmul(M, n):
...
+    # In other words, Y = complex_mul(e^ik, X - i[0, X[:0:-1]])
+    xi = torch.nn.functional.pad(-x[..., 1:], (0, 1)).flip(-1)
+    X = torch.stack([x, xi], dim=-1)
+    coef_real = torch.cos(torch.linspace(0, 0.5 * math.pi, N + 1))
+    coef = torch.stack([coef_real[:-1], coef_real[1:].flip(-1)], dim=-1)
+    Y = _complex_mul(coef, X)
+    # Step 2
+    y = torch.irfft(Y, 1, onesided=False)
+    # Step 3
+    # FIXME: something is wrong here
```
fritzo: Is something still wrong here?

fritzo:
Thanks @fehiepsi ! I have added docs to `dct` and `idct` and removed the dependency on `torch-dct`. I think this is ready for review. Can you in particular check `DiscreteCosineReparam.__call__()`?

pyro/ops/tensor_utils.py
fehiepsi: Ah, it is legacy. Btw, it seems that padding methods will not work here (because `dct(padding(x))` will have a length different from x)... this is different from `convolve` where we apply fft and ifft in the same method.

fritzo: ok, thanks for thinking about it.

PR #2253, opened 2020-01-12
Add tutorial on inference with Stable distributions

fritzo:
~~Blocked by #2252~~
~~Blocked by #2259~~

This adds a tutorial demonstrating use of the `StableReparam`, `SymmetricStableReparam`, and `DiscreteCosineReparam` for inference in a non-Gaussian stochastic volatility model.

## Tasks
- [x] create a dataloader in `pyro.contrib.examples`
- [x] upload dataset to cdn @neerajprad 

See [notebook](https://nbviewer.jupyter.org/github/pyro-ppl/pyro/blob/stable-example/tutorial/source/stable.ipynb)


pyro/contrib/examples/finance.py
```diff
@@ -0,0 +1,25 @@
...
+import urllib
+
+import pandas as pd
+
+from pyro.contrib.examples.util import get_data_directory
+
+DATA = get_data_directory(__file__)
+
+# https://finance.yahoo.com/quote/%5EGSPC/history/
+CACHE_URL = "https://fritzo.org/media/snp500.csv.bz2"
```
fritzo: @neerajprad Could you upload this to our CDN and either update this url or comment so I can update this url?

fehiepsi:
@fritzo I think that it is harder to review notebooks with `svg` format, especially for future updates. I would propose to have an environment flag to change the format to svg when we build tutorials.

fritzo:
@fehiepsi thanks, I'll remove the svg formatting.

pyro/contrib/examples/finance.py
neerajprad: The dataset is at - https://d2hg8soec8ck9v.cloudfront.net/datasets/snp500.csv.bz2

fritzo: Thanks!

martinjankowiak:
awesome! initial comments/suggestions:
- typos 'usinng', 'in to ... tractible ...', 'distributoin'
- comment on 'stability = 1.977' and 'stability = 1.965' below?
- explicate terms in volatility model more?
- what's with the scale of the figure in cell [11] ?
- comment on final figure?

fritzo:
@martinjankowiak ready for another review round.

martinjankowiak:
- 'significantly less than 1' => 'significantly less than 2' ?
- i still find the description of the form of the volatility model confusing. this is no doubt in part because i'm not familiar with finance models of this sort. but insofar as other readers may also not be very familiar, i think some more explication might be called for?

fritzo:
> the description of the form of the volatility model

sure, maybe I'll link to wikipedia or something. it's just brownian motion rated by geometric brownian motion.

fritzo:
Thanks for reviewing!

PR #2256, opened 2020-01-14
Rename stable reparam to make way for skewed reparam

fritzo:
1. Renames `StableReparam` -> `LatentStableReparam` to make way for a possible future `StableRepram` that works also for likelihoods and nonzero skew. (This new reparamterizer appears out of scope for the 1.2 release.)
2. Drops requirement for homogeneous `stability` parameter of `StableHMM`, after @martinjankowiak observed that this is not required by any of our inference algorithms.
3. Refactors interface of low-level `_unsafe_standard_stable()` to avoid an unnecessary tensor op when used in reparameterizers.

## Tested

### DIFF
--- a/pyro/distributions/stable.py
+++ b/pyro/distributions/stable.py
@@ -27,13 +27,12 @@ def _unsafe_standard_stable(alpha, beta, V, W):
     v = b.atan() + alpha * V
     Z = v.sin() / ((1 + b * b).rsqrt() * V.cos()).pow(inv_alpha) \
         * ((v - V).cos() / W).pow(inv_alpha - 1)
-    Z.data[~(Z.data == Z.data)] = 0  # drop occasional NANs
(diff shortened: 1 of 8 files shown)


pyro/distributions/stable.py
```diff
@@ -27,13 +27,12 @@ def _unsafe_standard_stable(alpha, beta, V, W):
...
+    Z.data[Z.data != Z.data] = 0  # drop occasional NANs
 
-    # Convert to Nolan's parametrization S^0 so that samples depend
-    # continuously on (alpha,beta), allowing us to interpolate around the hole
-    # at alpha=1.
-    X = Z - b
-    return X, Z
+    # Return both the standard Zolotarev parameterization Z and the shift b to
+    # convert to Nolan's parametrization S^0 where samples depend continuously
+    # on (alpha,beta), allowing interpolate around the hole at alpha=1.
```
martinjankowiak: grammar typo

PR #2259, opened 2020-01-15
Implement general StableReparam

fritzo:
~~Blocked by #2261~~
~~Blocked by #2262~~ 

This adds the final member of the `Stable` reparameterizers: a general `StableReparam` that emits a conditionally `Normal` distribution. The reparameterization splits the skewed stable input random variable into a symmetric stable r.v. + a totally skewed stable r.v., then defers to other strategies.

<details>

image

</details>

This also refactors the low-level `_standard_sample()` helpers to take a `coords` parameter, which can be either "S" or "S0".

## Tasks
- [x] get tests to pass
- [x] test inference on real data, e.g. #2253 (see [results](https://nbviewer.jupyter.org/github/pyro-ppl/pyro/blob/stable-example/tutorial/source/stable.ipynb))
- [x] ensure totally-skewed auxiliary variable parameterization is continuous
    (pointed out by @martinjankowiak )
- [x] Update `LinearHMMReparam` docs and tests after #2261 merges
- [x] Add KS test

### DIFF
--- a/pyro/infer/reparam/stable.py
+++ b/pyro/infer/reparam/stable.py
@@ -99,13 +99,14 @@ class SymmetricStableReparam(Reparam):
                         self._wrap(dist.Exponential(one), event_dim))
 
         # Differentiably transform to scale drawn from a totally-skewed stable variable.
-        z = _unsafe_standard_stable(fn.stability / 2, 1, u, e, coords="S")
+        a = fn.stability
+        z = _unsafe_standard_stable(a / 2, 1, u, e, coords="S")
         assert (z >= 0).all()
-        scale = fn.scale * (2 ** 0.5) * (math.pi / 4 * fn.stability).cos().pow(1 / fn.stability) * z.sqrt()
+        scale = fn.scale * (math.pi / 4 * a).cos().pow(a.reciprocal()) * z.sqrt()
         scale = scale.clamp(min=torch.finfo(scale.dtype).tiny)
 
-        # Construct a scaled Gaussian.
-        new_fn = self._wrap(dist.Normal(fn.loc, scale), event_dim)
+        # Construct a scaled Gaussian, using Stable(2,0,s,m) == Normal(m,s*sqrt(2)).
+        new_fn = self._wrap(dist.Normal(fn.loc, scale * (2 ** 0.5)), event_dim)
         return new_fn, obs
 
 
@@ -146,28 +147,23 @@ class StableReparam(Reparam):
         #
         # To derive the parameters of S and T, we solve the equations
         #
-        #   S.stability = a      T.stability = a
-        #   S.skew = 0           T.skew = sgn(b)
-        #   S.loc = 0            T.loc = 0
+        #   T.stability = a            S.stability = a
+        #   T.skew = sgn(b)            S.skew = 0
+        #   T.loc = 0                  S.loc = 0
         #
-        #   s = 1 = (S.scale**a + T.scale**a)**(1/a)
+        #   s = (S.scale**a + T.scale**a)**(1/a) = 1       # by step 1.
         #
         #       S.skew * S.scale**a + T.skew * T.scale**a
         #   b = ----------------------------------------- = sgn(b) * T.scale**a
         #                S.scale**a + T.scale**a
-        # whence
+        # yielding
         #
-        #  T.scale = |b| ** (1/a)
-        #  S.scale = (1 - |b|) ** (1/a)
-        a = fn.stability
-        a_inv = a.reciprocal()
-        skew_abs = fn.skew.abs()
-        t_scale = skew_abs.pow(a_inv)
-        s_scale = (1 - skew_abs).pow(a_inv)
+        #   T.scale = |b| ** (1/a)     S.scale = (1 - |b|) ** (1/a)
 
         # Draw parameter-free noise.
-        half_pi = a.new_full(a.shape, math.pi / 2)
-        one = a.new_ones(a.shape)
+        proto = fn.stability
+        half_pi = proto.new_full(proto.shape, math.pi / 2)
+        one = proto.new_ones(proto.shape)
         zu = pyro.sample("{}_z_uniform".format(name),
                          self._wrap(dist.Uniform(-half_pi, half_pi), event_dim))
         ze = pyro.sample("{}_z_exponential".format(name),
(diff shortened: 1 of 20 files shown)


pyro/infer/reparam/stable.py
```diff
@@ -99,54 +99,92 @@ def __call__(self, name, fn, obs):
...
+                         self._wrap(dist.Exponential(one), event_dim))
+        tu = pyro.sample("{}_t_uniform".format(name),
+                         self._wrap(dist.Uniform(-half_pi, half_pi), event_dim))
+        te = pyro.sample("{}_t_exponential".format(name),
+                         self._wrap(dist.Exponential(one), event_dim))
+
+        # Differentiably transform.
+        a = fn.stability
+        z = _unsafe_standard_stable(a / 2, 1, zu, ze, coords="S")
+        t = _standard_stable(a, 1, tu, te, coords="S") * fn.skew.sign()
```
fritzo: This version ensures continuity in `tu`, whereas the alternative
```py
t = _standard_stable(a, fn.skew.sign(), tu, te, coords="S")
```
is discontinuous. Thanks @martinjankowiak for pointing out the importance of verifying this.

pyro/infer/reparam/stable.py
```diff
@@ -99,11 +99,92 @@ def __call__(self, name, fn, obs):
...
+        #   T.loc = 0                  S.loc = 0
+        #
+        #   s = (S.scale**a + T.scale**a)**(1/a) = 1       # by step 1.
+        #
+        #       S.skew * S.scale**a + T.skew * T.scale**a
+        #   b = ----------------------------------------- = sgn(b) * T.scale**a
+        #                S.scale**a + T.scale**a
+        # yielding
+        #
+        #   T.scale = |b| ** (1/a)     S.scale = (1 - |b|) ** (1/a)
```
fehiepsi: This looks correct to me. Nicely done!

fritzo:
It looks like tests are failing at `stability=1` due to a real bug in `Stable.sample()`:
image
I am attempting to fix...

pyro/infer/reparam/stable.py
```diff
@@ -99,11 +99,92 @@ def __call__(self, name, fn, obs):
...
+    def __call__(self, name, fn, obs):
+        fn, event_dim = self._unwrap(fn)
+        assert isinstance(fn, dist.Stable)
+
+        # Strategy: Let X ~ S0(a,b,s,m) be the stable variable of interest.
+        # 1. WLOG scale and shift so s=1 and m=0, additionally shifting to convert
+        #    from Zolotarev's S parameterization to Nolan's S0 parameterization.
+        # 2. Decompose X = S + T, where
+        #    S ~ S(a,0,...,0) is symmetric and
+        #    T ~ S(a,sgn(b),...,0) is totally skewed.
```
fehiepsi: Unless returning a Normal distribution is important, would it be more stable to decompose into 2 totally skewed distributions? (so we have 4 auxiliary random variables instead of 5 as in this PR?)

fehiepsi:
> tests are failing at stability=1

You were using interpolation at this hole. It might be better to use the formula in Nolan ref for stability=1 and interpolate between `1 - eps` and `1`, rather than `1 - eps` and `1 + eps`.

pyro/infer/reparam/stable.py
fritzo: It is indeed important to return a Normal distribution, that way
1. the `.log_prob()` method is define so we can use this as a likelihood, and 
2. we can use this in a `LinearHMM` via `LinearHMMReparam`, where the `Normal` is used as a component distribution in a `GaussianHMM`, permitting parallel-scan inference.

fritzo:
The interpolation seems to be doing well even very close to 1
image
but if we go one digit closer there is total loss of precision:
image
I suspect there is catastrophic canceling somewhere...

fritzo:
I'm beginning to suspect SciPy's sampler is the buggy one:
image

fehiepsi:
Hmm, let me try a plot for this. Looks interesting...

pyro/infer/reparam/stable.py
fehiepsi: Thanks for explaining!

fritzo:
```py
import math
import torch
import numpy as np
from pyro.distributions.stable import Stable
from matplotlib import pyplot
%matplotlib inline
```
```py
num_samples = 1000000
alpha = 1.00000001
beta = 0
expected = levy_stable.rvs(alpha, beta, size=num_samples)
actual = Stable(alpha, beta).sample([num_samples])
if alpha != 1:
    expected -= beta * np.tan(math.pi / 2 * alpha)  # Convert from S0 to S coords
pyplot.figure(figsize=(9,4))
pyplot.hist(expected, bins=200, range=(-10, 10), color='b', alpha=0.5, label="scipy")
pyplot.hist(actual, bins=200, range=(-10, 10), color='r', alpha=0.5, label="pyro")
pyplot.xlim(-10, 10)
pyplot.title("stability = {}, skew = {}".format(alpha, beta))
pyplot.legend(loc="best");
```

fehiepsi:
FYI, the plot for alpha=1 (using the separate formula in Nolan ref)  is the same as yours.

pyro/infer/reparam/stable.py
```diff
@@ -99,11 +99,92 @@ def __call__(self, name, fn, obs):
...
+
+        # Differentiably transform.
+        a = fn.stability
+        z = _unsafe_standard_stable(a / 2, 1, zu, ze, coords="S")
+        t = _standard_stable(a, one, tu, te, coords="S") * fn.skew.sign()
+        a_inv = a.reciprocal()
+        skew_abs = fn.skew.abs()
+        t_scale = skew_abs.pow(a_inv)
+        s_scale = (1 - skew_abs).pow(a_inv)
+        loc = fn.loc + fn.scale * (t * t_scale - fn.skew * (math.pi / 2 * a).tan())
```
fritzo: I think the last remaining issue is that this line is discontinuous in `a` at `a=1`, leading to test failures of the KS test in `tests/infer/reparam/test_stable.py::test_distribution`.

pyro/infer/reparam/stable.py
```diff
@@ -99,11 +99,101 @@ def __call__(self, name, fn, obs):
...
+
+        # Differentiably transform.
+        a = fn.stability
+        z = _unsafe_standard_stable(a / 2, 1, zu, ze, coords="S")
+        t = _standard_stable(a, one, tu, te, coords="S")
+        a_inv = a.reciprocal()
+        skew_abs = fn.skew.abs()
+        t_scale = skew_abs.pow(a_inv) * fn.skew.sign()
+        s_scale = (1 - skew_abs).pow(a_inv)
+        # FIXME The following line is discontinuous at a=1.
```
fritzo: @fehiepsi I don't see an easy way to fix this, and as @martinjankowiak points out most interesting datasets have `stability > 1.5`. Therefore I've simply added a FIXME, a warning in the docstring, and a validation check. We can think about this in the future if we ever need `stability` near 1.

Ready to merge, pending review.

fritzo: Oh wait, I think I see the fix :smile: The following function **is** continuous except at the hole
```
(t_scale - fn.skew) * tan(pi / 2 * a)
  = skew.sign() * (skew.abs().pow(1/a) - skew.abs()) * tan(pi / 2 * a)
                  ----------------------------------   ---------------
                             has a root                has an asymptote
```
so we simply need to work around the canceling root. I'll quick add that as another interpolation trick...

fritzo: Fixed :smile: Ready to merge, pending review.

fehiepsi: Brilliant!

I spent the whole night to figure out a solution but unsuccessful. Pretty happy when seeing your solution after waking up! :D

pyro/infer/reparam/stable.py
```diff
@@ -99,11 +99,122 @@ def __call__(self, name, fn, obs):
...
+        scale = scale.clamp(min=torch.finfo(scale.dtype).tiny)
+
+        # Construct a scaled Gaussian, using Stable(2,0,s,m) == Normal(m,s*sqrt(2)).
+        new_fn = self._wrap(dist.Normal(loc, scale * (2 ** 0.5)), event_dim)
+        return new_fn, obs
+
+
+def _unsafe_shift(a, skew_abs, t_scale):
+    # At a=1 the lhs has a root and the rhs has an asymptote.
+    return (t_scale - skew_abs) * (math.pi / 2 * a).tan()
```
fehiepsi: I think it will be easier to understand the cancelation and patching logic here if you remove `t_scale` arg.

fehiepsi: Ah, I see, you want to save computation when no hole happens.

fritzo:
> will StableReparam still work with `coords=S`?

Good point, I'll add assertions to the `*StableReparam` classes. I don't intend the "S" version to be used in inference, only for generating data.

pyro/infer/reparam/stable.py
fritzo: Ha ha I also spent the whole night looking for a solution :D

Issue #2254, opened 2020-01-13
FR StudentTReparam, StudentTHMM


fritzo:
This proposes to implement reparameterizers for `StudentT` distributions by introducing latent `Gamma`-distributed variables, following the pattern of of symmetric `Stable` distributions.

## Tasks
- [ ] `StudentTReparam` following [SymmetricStableReparam](http://docs.pyro.ai/en/latest/infer.reparam.html#pyro.infer.reparam.stable.SymmetricStableReparam). #2265 
- ~~`StudentTHMM` following [StableHMM](http://docs.pyro.ai/en/latest/distributions.html#pyro.distributions.StableHMM). Note this version should use batched `StudentT` distributions rather than the elliptical `MultivariateStudentT` distribution; the latter distribution might be separately generalized to an HMM.~~
    UPDATE this is obsoleted by the new generic `LinearHMM` #2260
- ~~`StudentTHMMReparam` following [StableHMMReparam](http://docs.pyro.ai/en/latest/infer.reparam.html#pyro.infer.reparam.stable.StableHMMReparam).~~
  UPDATE this is obsoleted by the new generic `LinearHMMReparam` class in #2259

fritzo:
@fehiepsi I should have some time to implement this maybe next week, or did you want to implement it? It's nearly identical to the `Stable` versions.

fehiepsi:
@fritzo I can implement them later this week. Thanks for bringing them up!

PR #2261, opened 2020-01-15
Generalize StableHMMReparam to LinearHMMReparam

fritzo:
Following #2260 , this generalizes `StableHMMReparam` to a flexible `LinearHMMReparam` that allows many combinations of reparameterization strategies and allows `TransformedDistributions` such as `LogNormal` for observation distributions, as suggested by @martinjankowiak.

I have pulled this out from #2259 as it was becoming too complex.  From that old PR:
> This renames `StableHMMReparam` -> `LinearHMMReparam` and allows it to use different reparameterization strategies for each of `init`, `trans`, and `obs` sites. The motivation is to:
> 1. Allow all 2**3 = 8 combinations of {`StableReparam`,`SymmetricStableReparam`};
> 2. Address #2254 , which will be able to use `LinearHMMReparam(StudentTReparam(), ...)`; and
> 3. Make way for a future general `LinearHMM` class #2260 whose component distributions can mix and match different distributions such as `Stable`, `Normal`, `LogNormal` (as suggested by @martinjankowiak), or `StudentT`. (as discussed with @eb8680 )

This also adds some pattern matching logic to `LinearHMM` and makes `.to_event()` a little smarter, allowing it to *remove* event dimensions if given a negative number (greatly simplifying `LinearHMMReparam` code).

## Tested
- refactoring is exercised by existing stable_hmm tests
- added new tests for transformed hmm
- added unit tests for `to_event()`


fehiepsi:
@fritzo Just needs to be confirmed so I can review in the right direction: this PR changes the logic of LinearHMM sampling so that it will apply noise's transform to the output, rather than the noise?

fritzo:
@fehiepsi yes, the noise should be applied before the transform. That was the original intent of `LinearHMM`. This is important to allow transforms to change the observation domain/constraint, e.g. it doesn't make sense to **add** process noise to `LogNormal` observation noise because that would violate the positivity constraint. Instead we assume the noise adds to the initial distribution.

If this seems too weird, I would be open to reverting the transforms bit and instead forcing users to apply transforms directly to the `LinearHMM` distribution, but that would require extra logic inside `LinearReparamHMM` and extra logic on the user side to propagate `constraint` information.

fehiepsi:
Oh, LogNormal is a nice motivating example. Thanks! I'll review and if I find some points not reasonable, I'll let you know.

fritzo:
Thanks for reviewing!

PR #2264, opened 2020-01-16
improve speed and numerical stability of scale_tril to precision

fehiepsi:
This PR uses cholesky_solve to convert scale_tril to precision in MVN and MVT distributions.

For MVN, the PR https://github.com/pytorch/pytorch/pull/32092 does not make it into 1.4.0, so I overwrite the implementation in Pyro. We can remove it for the next PyTorch release.

### DIFF
--- a/pyro/distributions/torch.py
+++ b/pyro/distributions/torch.py
@@ -3,6 +3,7 @@
 
 import torch
 from torch.distributions import constraints
+from torch.distributions.utils import lazy_property
 
 from pyro.distributions.constraints import IndependentConstraint
 from pyro.distributions.torch_distribution import TorchDistributionMixin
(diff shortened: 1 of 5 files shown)


pyro/distributions/torch.py
```diff
@@ -40,6 +41,13 @@ def enumerate_support(self, expand=True):
 class MultivariateNormal(torch.distributions.MultivariateNormal, TorchDistributionMixin):
     support = IndependentConstraint(constraints.real, 1)  # TODO move upstream
 
+    # TODO: remove this in the PyTorch release > 1.4.0
```
fritzo: Can you confirm this is exercised in tests?

tests/distributions/test_mvt.py
```diff
@@ -12,6 +12,41 @@
...
+    (), (2,), (3, 2),
+])
+@pytest.mark.parametrize('cov_shape', [
+    (), (2,), (3, 2),
+])
+@pytest.mark.parametrize('dim', [
+    1, 3, 5,
+])
+def test_shape(df_shape, loc_shape, cov_shape, dim):
+    mvt = random_mvt(df_shape, loc_shape, cov_shape, dim)
```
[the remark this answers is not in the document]
fehiepsi: I have added a smoke test for the precision matrix backward.

tests/distributions/test_mvn.py
```diff
@@ -0,0 +1,42 @@
...
+from pyro.distributions import MultivariateNormal
+from tests.common import assert_equal
+
+
+def random_mvn(loc_shape, cov_shape, dim):
+    """
+    Generate a random MultivariateNormal distribution for testing.
+    """
+    rank = dim + dim
+    loc = torch.randn(loc_shape + (dim,))
```
fritzo: nit: why not `.requires_grad_()` here?

fehiepsi: That should be fine too. I'll do it instead. :)

PR #2266, opened 2020-01-17
Fix for mcmc api transforms None for potential_fn

### LINKED ISSUE #2272: [bug] MCMC run throws exception when using num_chains>1 as of Pyro 1.2.0.
### Issue Description

MCMC run throws exception when using num_chains>1 as of Pyro 1.2.0.  When num_chains is set to 1 the below code will execute in both Pyro 1.1.0 and Pyro 1.2.0. When num_chains is set to >1 the below code will execute in Pyro 1.1.0, but raises an AttributeException as of Pyro 1.2.0. 

```
AttributeError                            Traceback (most recent call last)
<ipython-input-13-5a1da811afef> in <module>()
      8 mcmc = MCMC(nuts_kernel, num_samples=25, warmup_steps=100, num_chains=cpu_count)
      9 
---> 10 mcmc.run(y)

~/anaconda3/envs/pytorch_p36/lib/python3.6/site-packages/pyro/poutine/messenger.py in _context_wrap(context, fn, *args, **kwargs)
      9 def _context_wrap(context, fn, *args, **kwargs):
     10     with context:
---> 11         return fn(*args, **kwargs)
     12 
     13 

~/anaconda3/envs/pytorch_p36/lib/python3.6/site-packages/pyro/infer/mcmc/api.py in run(self, *args, **kwargs)
    397 
    398         # transform samples back to constrained space
--> 399         for name, transform in self.transforms.items():
    400             z_acc[name] = transform.inv(z_acc[name])
    401         self._samples = z_acc

AttributeError: 'NoneType' object has no attribute 'items'
```

### Environment
- OS X and Python 3.6 / rhel fedora and Python 3.6
- PyTorch 1.4.0
- Pyro version
    - Throws error on 1.2.0
    - Does not throw error on 1.1.0

### Code Snippet
```python
import numpy as np

import torch
print(torch.__version__)
assert torch.__version__.startswith('1.4.0')

import pyro
from pyro.infer import MCMC, NUTS
print(pyro.__version__)
assert pyro.__version__.startswith('1.2.0') or pyro.__version__.startswith('1.1.0')

# set 
seed = 2020
pyro.set_rng_seed(seed)

def nrv(n=1000, loc=5, scale=2, seed=2020):
    np.random.seed(2020)
    z = np.random.normal(loc, scale, size=n)
    y = torch.from_numpy(z).type(torch.float)
    return y

def model(y):
    scale_guess = torch.tensor(20, dtype=torch.float)
    scale_prior = pyro.sample('scale', dist.HalfNormal(scale_guess))
    mean_guess = torch.tensor(0, dtype=torch.float)
    mean_prior = pyro.sample('loc', dist.Normal(mean_guess, scale_prior), obs=y)

def main(num_chains):
    nuts_kernel = NUTS(model, adapt_step_size=True)
    mcmc = MCMC(nuts_kernel, num_samples=25, warmup_steps=100, num_chains=chains)
    mcmc.run(y)

if __name__ == '__main__':
    main(1)
    main(2)
```

ludkinm:
Fixes #2272. 

This PR fixes a logic gap for assign `self.transforms` in the `MCMC` api.
It can be the case that both `MCMC.transforms is None` and `MCMC.kernel.transforms is None`
in which case MCMC.transforms should be set to an empty dict.
This bug appears when using HMC/NUTS with a potential_fn.

Note sure on the style guide for inline comments so will leave them as is.

### DIFF
--- a/pyro/infer/mcmc/api.py
+++ b/pyro/infer/mcmc/api.py
@@ -385,13 +385,13 @@ class MCMC(object):
         # If transforms is not explicitly provided, infer automatically using
         # model args, kwargs.
         if self.transforms is None:
-            if hasattr(self.kernel, 'transforms'):
-                if self.kernel.transforms is not None:
-                    self.transforms = self.kernel.transforms
-            elif self.kernel.model:
-                _, _, self.transforms, _ = initialize_model(self.kernel.model,
-                                                            model_args=args,
-                                                            model_kwargs=kwargs)
+            if hasattr(self.kernel, 'transforms'): # If the kernel has transforms
+                if self.kernel.transforms is not None: # and they are not None
+                    self.transforms = self.kernel.transforms # then use the kernels transforms
+                else:
+                    self.transforms = {}
+            elif self.kernel.model: # else if the kernel doesn't define transforms, try using the model
+                _, _, self.transforms, _ = initialize_model(self.kernel.model, model_args=args, model_kwargs=kwargs)
             else:
                 self.transforms = {}
 
--- a/pyro/infer/mcmc/api.py
+++ b/pyro/infer/mcmc/api.py
@@ -385,12 +385,12 @@ class MCMC(object):
         # If transforms is not explicitly provided, infer automatically using
         # model args, kwargs.
         if self.transforms is None:
-            if hasattr(self.kernel, 'transforms'): # If the kernel has transforms
-                if self.kernel.transforms is not None: # and they are not None
+            if hasattr(self.kernel, 'transforms'):  # If the kernel has transforms
+                if self.kernel.transforms is not None:  # and they are not None
                     self.transforms = self.kernel.transforms # then use the kernels transforms
                 else:
                     self.transforms = {}
-            elif self.kernel.model: # else if the kernel doesn't define transforms, try using the model
+            elif self.kernel.model:  # else if the kernel doesn't define transforms, try using the model
                 _, _, self.transforms, _ = initialize_model(self.kernel.model, model_args=args, model_kwargs=kwargs)
             else:
                 self.transforms = {}
--- a/pyro/infer/mcmc/api.py
+++ b/pyro/infer/mcmc/api.py
@@ -385,15 +385,15 @@ class MCMC(object):
         # If transforms is not explicitly provided, infer automatically using
         # model args, kwargs.
         if self.transforms is None:
-            if hasattr(self.kernel, 'transforms'):
-                # Use kernel transforms if not `None`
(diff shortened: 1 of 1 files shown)


pyro/infer/mcmc/api.py
```diff
@@ -385,13 +385,13 @@ def run(self, *args, **kwargs):
...
         # model args, kwargs.
         if self.transforms is None:
-            if hasattr(self.kernel, 'transforms'):
-                if self.kernel.transforms is not None:
-                    self.transforms = self.kernel.transforms
-            elif self.kernel.model:
-                _, _, self.transforms, _ = initialize_model(self.kernel.model,
-                                                            model_args=args,
-                                                            model_kwargs=kwargs)
+            if hasattr(self.kernel, 'transforms'): # If the kernel has transforms
```
fritzo: two spaces before `#` character

pyro/infer/mcmc/api.py
```diff
@@ -385,13 +385,13 @@ def run(self, *args, **kwargs):
...
-            if hasattr(self.kernel, 'transforms'):
-                if self.kernel.transforms is not None:
-                    self.transforms = self.kernel.transforms
-            elif self.kernel.model:
-                _, _, self.transforms, _ = initialize_model(self.kernel.model,
-                                                            model_args=args,
-                                                            model_kwargs=kwargs)
+            if hasattr(self.kernel, 'transforms'):  # If the kernel has transforms
+                if self.kernel.transforms is not None:  # and they are not None
+                    self.transforms = self.kernel.transforms # then use the kernels transforms
```
neerajprad: I think the lint is failing here. Btw, you could run it locally using `make lint`.

neerajprad:
I just pushed a fix to release a patch for 1.2 with this fix.

neerajprad:
There are a few issues still with multiprocessing, I'll either fix it here or put out a separate PR.

neerajprad:
ludkinm - Could you review the minor fixes that I made for #2272, and let me know if that's in line with your use case?

neerajprad:
@fehiepsi - can we merge this? I would like to fix #2272 and release a patch update.

fehiepsi:
@neerajprad your review is still in change request status, which does not allow me to merge.

fehiepsi:
Sorry, I just know that I can dismiss your review :)

ludkinm:
Sorry I have been away from the office so didn't respond.
@neerajprad thanks for your fix, it works in my use case too - my python is rusty and forgot that if statements are evaluated lazily.

PR #2268, opened 2020-01-17
Port neutra example from numpyro

neerajprad:
This ports the neutra example from numpyro to test the NeutraReparam reparametrizer. I am not seeing good results with the IAFNormal autoguide and specially with the default setting where it uses a single flow.

TODOs:
 - [ ] Try on another dataset - maybe high dimensional Neal's funnel.
 - [ ] Have a more general flow based autoguide to wrap over BNAF which is not currently available as an autoguide.
 - [ ] The samples from the warped posterior look far from gaussian, and this probably needs to be investigated. 

[neutra.pdf](https://github.com/pyro-ppl/pyro/files/4078250/neutra.pdf)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ examples/*/results
 examples/*/raw
 examples/dmm/*.pkl
 examples/*.csv
+examples/*.pdf
 examples/mixed_hmm/*csv
 pyro/contrib/examples/processed
 pyro/contrib/examples/results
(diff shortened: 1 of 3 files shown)


fehiepsi:
I think bnaf will solve the issue easily. :D

neerajprad:
Closing this in favor of #2294 which is probably better suited for discussion. Once the issues with the neutra hmc are resolved, I'll convert the notebook into a python script.

PR #2269, opened 2020-01-17
Fix tutorial build for the website

neerajprad:
Minor changes so that html build doesn't throw warnings and level 1 headers are only used for tutorial titles (otherwise the left sidebar would get populated with all the subtitles if they are at level 1).


fritzo:
@neerajprad is the DMM tutorial new since last release? If so I'll add it to the 1.2 release docs.

fehiepsi:
It is new since the last release. I added it to `hackmd` but it does not appear in the release note.

fritzo:
@fehiepsi thanks, I'll update the release notes. (I probably copied hackmd -> github before you finished editing)

PR #2271, opened 2020-01-20
bugfix #2270

### LINKED ISSUE #2270: MaskedMixture distribution doesn't work properly when Delta distribution is used as a component
### Issue Description
MaskedMixture fails when Delta distribution is used as a component distribution. In particular, sampling from MaskedMixture outputs wrong values and also alters the Delta distribution used as a component.

### Code Snippet
```py
import sys
import torch
import pyro

print(sys.version)
print(pyro.__version__)
print(torch.__version__)

delta = dist.Delta(torch.tensor([0.]))
gamma = dist.Gamma(torch.ones(2)*100., torch.ones(1))
m = torch.tensor([0, 1]).bool()

print("\nDelta dist before sampling:", delta)

masked_mixture = dist.MaskedMixture(m, delta, gamma)
print("\nSample masked mixture:", masked_mixture.sample())

print("\nDelta dist after sampling:", delta)
```
returns:
```
3.7.6 (default, Jan  8 2020, 19:59:22)
[GCC 7.3.0]
1.2.0
1.4.0

Delta dist before sampling: Delta(v: tensor([0.]), log_density: tensor([0.]))

Sample masked mixture: tensor([103.3137, 103.3137])

Delta dist after sampling: Delta(v: tensor([103.3137]), log_density: tensor([0.]))
```
possible solution (at least it fixes the example above) is to use torch.where in pyro/distributions/mixture.py file like below:

```diff
  def sample(self, sample_shape=torch.Size()):
      mask = self.mask.expand(sample_shape + self.batch_shape) if sample_shape else self.mask
-     result = self.component0.sample(sample_shape)
-     result[mask] = self.component1.sample(sample_shape)[mask]
+     result = torch.where(mask,
+                          self.component1.sample(sample_shape),
+                          self.component0.sample(sample_shape))
      return result
```

ordabayevy:
Resolves #2270 
Using torch.where seems to resolve the issues related to using the Delta distribution as a component in the MaskedMixture.

### DIFF
--- a/pyro/distributions/mixture.py
+++ b/pyro/distributions/mixture.py
@@ -102,6 +102,7 @@ class MaskedMixture(TorchDistribution):
 
     def sample(self, sample_shape=torch.Size()):
         mask = self.mask.expand(sample_shape + self.batch_shape) if sample_shape else self.mask
+        mask = mask.reshape(mask.shape + (1,) * len(self.event_shape))
         result = torch.where(mask,
                              self.component1.sample(sample_shape),
                              self.component0.sample(sample_shape))
--- a/pyro/distributions/mixture.py
+++ b/pyro/distributions/mixture.py
@@ -101,16 +101,16 @@ class MaskedMixture(TorchDistribution):
             return type(self)(mask, component0, component1)
 
     def sample(self, sample_shape=torch.Size()):
-        mask = self.mask.expand(sample_shape + self.batch_shape) if sample_shape else self.mask
-        mask = mask.reshape(mask.shape + (1,) * len(self.event_shape))
+        mask = self.mask.reshape(self.mask.shape + (1,) * self.event_dim)
+        mask = mask.expand(sample_shape + self.shape())
         result = torch.where(mask,
                              self.component1.sample(sample_shape),
                              self.component0.sample(sample_shape))
         return result
 
     def rsample(self, sample_shape=torch.Size()):
-        mask = self.mask.expand(sample_shape + self.batch_shape) if sample_shape else self.mask
-        mask = mask.reshape(mask.shape + (1,) * len(self.event_shape))
+        mask = self.mask.reshape(self.mask.shape + (1,) * self.event_dim)
+        mask = mask.expand(sample_shape + self.shape())
         result = torch.where(mask,
                              self.component1.rsample(sample_shape),
                              self.component0.rsample(sample_shape))
--- a/pyro/distributions/mixture.py
+++ b/pyro/distributions/mixture.py
@@ -102,14 +102,16 @@ class MaskedMixture(TorchDistribution):
 
     def sample(self, sample_shape=torch.Size()):
         mask = self.mask.expand(sample_shape + self.batch_shape) if sample_shape else self.mask
-        result = self.component0.sample(sample_shape)
-        result[mask] = self.component1.sample(sample_shape)[mask]
+        result = torch.where(mask,
+                             self.component1.sample(sample_shape),
+                             self.component0.sample(sample_shape))
         return result
 
     def rsample(self, sample_shape=torch.Size()):
         mask = self.mask.expand(sample_shape + self.batch_shape) if sample_shape else self.mask
-        result = self.component0.rsample(sample_shape)
-        result[mask] = self.component1.rsample(sample_shape)[mask]
+        result = torch.where(mask,
+                             self.component1.rsample(sample_shape),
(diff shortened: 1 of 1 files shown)


pyro/distributions/mixture.py
```diff
@@ -102,14 +102,16 @@ def expand(self, batch_shape):
 
     def sample(self, sample_shape=torch.Size()):
         mask = self.mask.expand(sample_shape + self.batch_shape) if sample_shape else self.mask
-        result = self.component0.sample(sample_shape)
-        result[mask] = self.component1.sample(sample_shape)[mask]
+        result = torch.where(mask,
```
fehiepsi: I think you would need to reshape `mask` here: `mask.reshape(mask.shape + (1,) * len(self.event_shape))`

fritzo: Good catch, it looks like this is causing `tests/distributions/test_mixture.py::test_masked_mixture_multivariate` to fail. @ordabayevy can you see if this fixes the test?

ordabayevy: @fritzo yeah, I am looking into it

ordabayevy: @fritzo where can I find instructions on how to run the tests? I looked [here](https://github.com/pyro-ppl/pyro/tree/dev/tests) but couldn't build pytorch binaries in docker. Some help would be greatly appreciated since I am new to many of these things.

ordabayevy: I've run the test locally as `pytest -q test_mixture.py` and it passed it with @fehiepsi suggestion.

pyro/distributions/mixture.py
```diff
@@ -102,14 +102,17 @@ def expand(self, batch_shape):
...
+        mask = mask.reshape(mask.shape + (1,) * len(self.event_shape))
+        result = torch.where(mask,
+                             self.component1.sample(sample_shape),
+                             self.component0.sample(sample_shape))
         return result
 
     def rsample(self, sample_shape=torch.Size()):
         mask = self.mask.expand(sample_shape + self.batch_shape) if sample_shape else self.mask
-        result = self.component0.rsample(sample_shape)
-        result[mask] = self.component1.rsample(sample_shape)[mask]
```
fehiepsi: @ordabayevy Could you also reshape the mask here?

pyro/distributions/mixture.py
```diff
@@ -122,8 +125,9 @@ def log_prob(self, value):
         mask = self.mask
         if mask.shape != mask_shape:
             mask = mask.expand(mask_shape)
-        result = self.component0.log_prob(value)
-        result[mask] = self.component1.log_prob(value)[mask]
```
fehiepsi: and here too

pyro/distributions/mixture.py
ordabayevy: @fehiepsi ok, thanks. Reshaping in the log_prob seems to be unnecessary.

fehiepsi: I think it is required. The `mask` shape there also does not include `event_shape`.

pyro/distributions/mixture.py
```diff
@@ -102,14 +102,18 @@ def expand(self, batch_shape):
...
-        result = self.component0.sample(sample_shape)
-        result[mask] = self.component1.sample(sample_shape)[mask]
+        mask = mask.reshape(mask.shape + (1,) * len(self.event_shape))
+        result = torch.where(mask,
+                             self.component1.sample(sample_shape),
+                             self.component0.sample(sample_shape))
         return result
 
     def rsample(self, sample_shape=torch.Size()):
         mask = self.mask.expand(sample_shape + self.batch_shape) if sample_shape else self.mask
```
fritzo: I think this should be
```py
mask = self.mask.reshape(self.mask.shape + (1,) * self.event_dim)
mask = mask.expand(sample_shape + self.shape())
```

pyro/distributions/mixture.py
ordabayevy: including it there fails the test, presumably because the output of the log_prob has the shape of `sample_shape + batch_shape` without the `event_shape`

pyro/distributions/mixture.py
ordabayevy: @fritzo that looks good to me. I will submit the commit.

pyro/distributions/mixture.py
fehiepsi: Opps, you are right!!!

Issue #2277, opened 2020-01-22
Reparametrization trick is incorrect in presence of discontinuity


fritzo:
This issue was raised by @hongseok-yang.

In models where likelihoods discontinuously depend on a latent variable, it is incorrect to use the parameterization trick for that variable in the guide. A correct workaround is to substitute the guide distribution with a non-reparameterized copy. For example:
```diff
  def model(data):
      z = pyro.sample("z", dist.Normal(0,1))
      p = z.round().clamp(min=0.2, max=0.8)
      pyro.sample("x", dist.Bernoulli(p), obs=tensor(0.))

  def guide(data):
      loc = pyro.param("loc", torch.tensor(0.))
      scale = pyro.param("scale", torch.tensor(1.), constraint=positive)
-    z = pyro.sample("z", dist.Normal(loc, 1))
+    z = pyro.sample("z", NonreparameterizedNormal(loc, 1))
```

## Tasks
- [ ] allow manual user-specified disabling of the reparameterization trick for a single variable.
- [ ] add validation logic to detect when the issue occurs
- [ ] add a tutorial on "caveats of variational inference"

PR #2278, opened 2020-01-22
Add in-place method Distribution.has_rsample_()

fritzo:
Addresses #2277 

This adds a `.has_rsample_()` in-place method similar to torch's `.requires_grad_()`. This is intended use in guides where the reparametrization trick is invalid.

The minimal change I could find to support @hongseok-yang's request was to 
1. add an attribute override setter to `Distribution` class and
2. preserve any overridden `.has_rsample` attributes in `@broadcast`.

This should be orthogonal to all inference algorithms. Example usage:
```py
def model(data):
    z = pyro.sample("z", dist.Normal(0,1))
    p = z.round().clamp(min=0.2, max=0.8)  # discontinuous
    pyro.sample("x", dist.Bernoulli(p), obs=tensor(0.))

def guide(data):
    loc = pyro.param("loc", torch.tensor(0.))
    scale = pyro.param("scale", torch.tensor(1.), constraint=positive)
    pyro.sample("z", dist.Normal(loc, scale).has_rsample_(False))  # <------ HERE
```

## Tested
- smoke tests
- gradient tests

### DIFF
--- a/pyro/distributions/distribution.py
+++ b/pyro/distributions/distribution.py
@@ -83,6 +83,10 @@ class Distribution(object, metaclass=ABCMeta):
         distributions should override this method to compute correct
         `.score_function` and `.entropy_term` parts.
 
+        Setting ``.has_rsample`` on a distribution instance will determine
+        whether inference engines use reparameterized samplers or the score
+        function estimator.
+
         :param torch.Tensor x: A single value or batch of values.
         :return: A `ScoreParts` object containing parts of the ELBO estimator.
         :rtype: ScoreParts
@@ -112,3 +116,21 @@ class Distribution(object, metaclass=ABCMeta):
         :rtype: iterator
         """
         raise NotImplementedError("Support not implemented for {}".format(type(self)))
+
+    def has_rsample_(self, value):
+        """
+        Force reparameterized or detached sampling on a single distribution
+        instance. This sets the ``.has_rsample`` attribute in-place.
+
+        This is useful to instruct inference algorithms to avoid
+        reparameterized gradients for variables that discontinuously determine
+        downstream control flow.
+
+        :param bool value: Whether samples be pathwise differentiable.
+        :return: self
+        :rtype: Distribution
+        """
+        if not (value is True or value is False):
+            raise ValueError("Expected value in {False,True}, actual {}".format(value))
+        self.has_rsample = value
+        return self
--- a/pyro/distributions/distribution.py
+++ b/pyro/distributions/distribution.py
@@ -84,8 +84,8 @@ class Distribution(object, metaclass=ABCMeta):
         `.score_function` and `.entropy_term` parts.
 
         Setting ``.has_rsample`` on a distribution instance will determine
-        whether inference engines use reparameterized samplers or the score
-        function estimator.
+        whether inference engines like :class:`~pyro.infer.svi.SVI` use
+        reparameterized samplers or the score function estimator.
 
         :param torch.Tensor x: A single value or batch of values.
         :return: A `ScoreParts` object containing parts of the ELBO estimator.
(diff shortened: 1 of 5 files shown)


pyro/distributions/distribution.py
```diff
@@ -83,6 +83,10 @@ def score_parts(self, x, *args, **kwargs):
         distributions should override this method to compute correct
         `.score_function` and `.entropy_term` parts.
 
+        Setting ``.has_rsample`` on a distribution instance will determine
+        whether inference engines use reparameterized samplers or the score
```
martinjankowiak: plus something like the following?:  
gradient-based inference engines like :class:`SVI`

pyro/distributions/distribution.py
```diff
@@ -112,3 +116,21 @@ def enumerate_support(self, expand=True):
...
+    def has_rsample_(self, value):
+        """
+        Force reparameterized or detached sampling on a single distribution
+        instance. This sets the ``.has_rsample`` attribute in-place.
+
+        This is useful to instruct inference algorithms to avoid
+        reparameterized gradients for variables that discontinuously determine
+        downstream control flow.
+
+        :param bool value: Whether samples be pathwise differentiable.
```
martinjankowiak: nit: +should?

pyro/poutine/broadcast_messenger.py
```diff
@@ -63,4 +63,6 @@ def _pyro_sample(msg):
             for i in range(-len(target_batch_shape) + 1, 1):
                 if target_batch_shape[i] is None:
                     target_batch_shape[i] = actual_batch_shape[i] if len(actual_batch_shape) >= -i else 1
-            msg["fn"] = msg["fn"].expand(target_batch_shape)
+            msg["fn"] = dist.expand(target_batch_shape)
+            if "has_rsample" in dist.__dict__:
```
fritzo: @neerajprad This is kind of gross, but otherwise we'd need to update O(#distributions)-many implementations of `.expand()`. Any ideas for alternatives?

tests/infer/test_valid_models.py
```diff
@@ -197,6 +197,47 @@ def guide():
     assert_error(model, guide, Elbo(), match='Multiple sample sites named')
 
 
+@pytest.mark.parametrize("has_rsample", [False, True])
+@pytest.mark.parametrize("Elbo", [Trace_ELBO, TraceGraph_ELBO, TraceEnum_ELBO, TraceTMC_ELBO])
+def test_set_has_rsample_ok(has_rsample, Elbo):
```
martinjankowiak: can you please comment on the main goal of this and the following test?

pyro/poutine/broadcast_messenger.py
neerajprad: I think we can probably add this check to `TorchDistribution._get_checked_instance`, but we can do it in PyTorch directly to make it available for PyTorch users.

tests/infer/test_valid_models.py
fritzo: Done. Note this test isn't likely to fail now, but Hongseok and students may implement stricter checks for #2277, and those checks should take care to validate this model.

pyro/poutine/broadcast_messenger.py
fritzo: @neerajprad that's a great idea. I've made the logic a little safer, but I do like the idea of updating `._get_checked_instance` to copy this and possibly other custom attributes; that way other possible occurrences of `.expand()` won't silently break.

fritzo: @neerajprad on further inspection, `_get_checked_instance()` doesn't seem like quite the right place. Rather I think we can implement a revised interface that copies `._validate_args` and any other of a set of whitelisted attributes. Of course this is all PyTorch work, out of scope for this PR.

fritzo:
bump

PR #2279, opened 2020-01-24
Expand MaskedDistribution batch shape to the broadcast of mask and base_dist batch shapes

ordabayevy:
### Issue Description
Add feature to expand the `batch_shape` of the `MaskedDistribution` to the `broadcast_shape(mask.shape, base_dist.batch_shape)` which seems to me like an intuitive behavior. These changes fail `tests/distributions/test_mask.py::test_mask_invalid_shape` test most likely because of the differences in the intended behavior.

### DIFF
--- a/pyro/distributions/torch_distribution.py
+++ b/pyro/distributions/torch_distribution.py
@@ -252,7 +252,6 @@ class MaskedDistribution(TorchDistribution):
     arg_constraints = {}
 
     def __init__(self, base_dist, mask):
-        self.base_dist = base_dist
         if isinstance(mask, bool):
             self._mask = mask
         else:
@@ -262,6 +261,7 @@ class MaskedDistribution(TorchDistribution):
             if base_dist.batch_shape != batch_shape:
                 base_dist = base_dist.expand(batch_shape)
             self._mask = mask.bool()
+        self.base_dist = base_dist
         super(MaskedDistribution, self).__init__(base_dist.batch_shape, base_dist.event_shape)
 
 
--- a/pyro/distributions/torch_distribution.py
+++ b/pyro/distributions/torch_distribution.py
@@ -256,11 +256,14 @@ class MaskedDistribution(TorchDistribution):
         if isinstance(mask, bool):
             self._mask = mask
         else:
-            if broadcast_shape(mask.shape, base_dist.batch_shape) != base_dist.batch_shape:
-                raise ValueError("Expected mask.shape to be broadcastable to base_dist.batch_shape, "
-                                 "actual {} vs {}".format(mask.shape, base_dist.batch_shape))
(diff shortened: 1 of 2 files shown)


pyro/distributions/torch_distribution.py
```diff
@@ -252,16 +252,19 @@ class MaskedDistribution(TorchDistribution):
...
+            batch_shape = broadcast_shape(mask.shape, base_dist.batch_shape)
+            if mask.shape != batch_shape:
+                mask = mask.expand(batch_shape)
+            if base_dist.batch_shape != batch_shape:
+                base_dist = base_dist.expand(batch_shape)
             self._mask = mask.bool()
+        self.base_dist = base_dist
         super(MaskedDistribution, self).__init__(base_dist.batch_shape, base_dist.event_shape)
 
+
```
fritzo: nit: delete extraneous line

ordabayevy:
@fritzo another change I would like to propose is to implement the same logic in `MaskMessenger`. 
```
+        if msg["done"] or msg["type"] != "sample": # not sure if conditions are right here
+            return None
+
         msg["mask"] = self.mask if msg["mask"] is None else self.mask & msg["mask"]
+        msg["fn"] = MaskedDistribution(msg["fn"], msg["mask"])
         return None
```
For example, when the mask is enumerated the sampling shape would automatically be expanded to the shape of the enumerated mask (model1 would behave like model2).
```
@config_enumerate
 def model1():
     with pyro.plate("data", 5, dim=-2):
         with pyro.plate("components", 2, dim=-1):
             mask = pyro.sample("mask", dist.Categorical(torch.ones(2)))
             with pyro.poutine.mask(mask=mask.bool()):
                 x = pyro.sample("x", dist.Normal(0., 1.))
                 # x.shape = (5,2) not expanded to mask.shape

 @config_enumerate
 def model2():
     with pyro.plate("data", 5, dim=-2):
         with pyro.plate("components", 2, dim=-1):
             mask = pyro.sample("mask", dist.Categorical(torch.ones(2)))
             with pyro.poutine.mask(mask=mask.bool()):
                 x = pyro.sample("x", dist.Normal(0., 1.).expand(mask.shape))
                 # x.shape = (2,5,2) expanded to mask.shape
```

fritzo:
@ordabayevy I like your suggested changes to `MaskMessenger`, but we'll need to think through implications on the `scale_and_mask()` helper. It may be that, with your suggestion of using `MaskedDistribution` we can eliminate much of the complex logic involving `scale_and_mask()`, and possibly even eliminate the `mask` field from trace sites.

If you'd like to make this change, can you put up a PR where we can discuss?

ordabayevy:
@fritzo ok, I will. Yeah, I was also thinking that this way masking of `log_prob` can be handled by the `MaskedDistribution`

Issue #2281, opened 2020-01-25
Create an autoguide to flexibly use any normalizing flow


neerajprad:
We have many nice normalizing flows added by @stefanwebb, but going by our current design, we may need to separately wrap each of these in an Autoguide. Instead, I propose having a generic `AutoNormalizingFlow` guide that can taken in any normalizing flow (a `TranformModule` instance). 

Additionally:
 - It will be nice to be able to specify the number of flows. 
 - We should be able to delegate creating these transforms to an initialization function. That way users can specify other parameters like hidden dims, etc.

@stefanwebb, @fehiepsi : WDYT? I have a rough draft ready that I'll push shortly. 

I need this to do some experiments for the neural transport HMC.

fehiepsi:
Resolved at #2282

PR #2282, opened 2020-01-25
Add AutoNormalizingFlow autoguide

neerajprad:
Addresses #2281. 

This creates a generic AutoNormalizingFlow autoguide that can take in any normalizing flow (an instance of `TransformModule`). The `AutoIAFNormal` just wraps over this more general autoguide, though we can also deprecate it since it is not doing much.

Testing: Existing `AutoIAFNormal` tests should exercise this code path. I have also added `AutoIAFNormal` and an `AutoBNAFNormal` to one of the inference tests. 

 - This uses the transform factory functions by @stefanwebb along with a `make_transforms` utility so that the `init_transform_fn` can return a `TransformModule` instance when provided with the latent dimension of the model.
 - I had to add a `ComposeTransformModule` that inherits from `ComposeTransform` and `ModuleList`. This is because transforms need to be an instance of `nn.Module` for Pyro to correctly register their parameters in the param store. I couldn't have simply used `self.transforms = ComposeTransform(...)` because that wouldn't register the parameters of the individual transforms.

**TODOs:**
 - [x] Add another test for BlockAutoregressive flow. 
 - [x] `test_neutra` fails currently with "RuntimeError: Trying to backward through the graph a second time, but the buffers have already been freed." After much debugging, I have isolated the issue to the cache used by `AffineAutoregressive`, but I am not yet sure what's the best way to handle this. ~#2283, This is resolved by setting `retain_graph=True` in the verlet integrator.~ This can be resolved by cloning the tensors to get around the caching by id.
 - [x] Add to docs.

cc. @stefanwebb


pyro/infer/autoguide/guides.py
```diff
@@ -656,44 +657,75 @@ def _loc_scale(self, *args, **kwargs):
...
     :param callable init_loc_fn: A per-site initialization function.
         See :ref:`autoguide-initialization` section for available functions.
+    :param init_tranform_kwargs: other keyword arguments taken by
+        `init_transform_fn`.
     """
 
-    def __init__(self, model, hidden_dim=None, init_loc_fn=init_to_median):
-        self.hidden_dim = hidden_dim
-        self.arn = None
+    def __init__(self, model, num_flows, init_transform_fn, init_loc_fn=init_to_median,
```
fehiepsi: `init_transform_fn` and `init_transform_kwargs` are a bit confusing. The class name is *Flow. We also have the arg `num_flows`. And not all TransformModule can be used here (only transforms have input_dim=output_dim are valid). I would suggest other names: `flow_generator` and `flow_kwargs`. :D

pyro/distributions/torch_transform.py
```diff
@@ -15,3 +15,19 @@ def __init__(self, *args, **kwargs):
 
     def __hash__(self):
         return super(torch.nn.Module, self).__hash__()
+
+
+class ComposeTransformModule(torch.distributions.ComposeTransform, torch.nn.ModuleList):
+    """
+    This allows us to use a list of `TransformModule` in the same way as
+    :class:`~torch.distributions.transform.ComposeTransform`. This is needed
+    so that transform parameters are automatically  registered by Pyro's param
```
martinjankowiak: typo: 2 spaces after automatically

pyro/infer/autoguide/guides.py
```diff
@@ -656,44 +657,75 @@ def _loc_scale(self, *args, **kwargs):
...
         super().__init__(model, init_loc_fn=init_loc_fn)
 
     def get_posterior(self, *args, **kwargs):
         """
         Returns a diagonal Normal posterior distribution transformed by
-        :class:`~pyro.distributions.transforms.iaf.InverseAutoregressiveFlow`.
+        :class:`~pyro.distributions.torch_tranform.TransformModule`.
         """
         if self.latent_dim == 1:
             raise ValueError('latent dim = 1. Consider using AutoDiagonalNormal instead')
```
martinjankowiak: i'm not sure if this `ValueError` is still needed? @stefanwebb are any of the current flows non-trivial in `D=1`?

pyro/infer/autoguide/guides.py
```diff
@@ -656,44 +657,75 @@ def _loc_scale(self, *args, **kwargs):
...
     """
 
-    def __init__(self, model, hidden_dim=None, init_loc_fn=init_to_median):
-        self.hidden_dim = hidden_dim
-        self.arn = None
+    def __init__(self, model, num_flows, init_transform_fn, init_loc_fn=init_to_median,
+                 **init_transform_kwargs):
+        self.num_flows = num_flows
+        self.transform = None
+        self._init_transform_fn = init_transform_fn
```
martinjankowiak: i think the general idea of using these pre-made factories is a good one :+1:

pyro/infer/autoguide/guides.py
stefanwebb: I would suggest changing `num_flows` to `num_transforms` and keeping everything in terms of "transforms" rather than "flows". I understand a "normalizing flow" to be comprised of a number of bijective transforms.

pyro/infer/autoguide/guides.py
stefanwebb: Normalizing flows usually contain a number of different transforms of different types. You can compose any of the transforms in the library into a normalizing flow...

Does it make sense then to pass a single transform function that can be repeated? If you wanted a more complex flow, you'd have to write a helper function, something like:

```
def make_transforms(latent_dim=1):
  transforms = [T.affine_autoregressive(...), T.batch_norm(...), T.affine_autoregressive(...)]
  return T.ComposeTransforms(transforms)
```

pyro/infer/autoguide/guides.py
stefanwebb: Not yet, but `SplineCoupling` and some other ones will be (to be pushed shortly). So I recommend removing this assertion

pyro/infer/autoguide/guides.py
```diff
@@ -656,44 +657,75 @@ def _loc_scale(self, *args, **kwargs):
...
+        guide = AutoIAFNormal(model, hidden_dim=latent_dim)
+        svi = SVI(model, guide, ...)
+
+    :param callable model: a generative model
+    :param int hidden_dim: number of hidden dimensions in the IAF
+    :param callable init_loc_fn: A per-site initialization function.
+        See :ref:`autoguide-initialization` section for available functions.
+    """
+
+    def __init__(self, model, num_flows=1, hidden_dim=None, init_loc_fn=init_to_median):
```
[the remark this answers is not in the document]
stefanwebb: Yeah, I recommend changing to `num_transforms` and making the default argument `1`

pyro/infer/autoguide/guides.py
stefanwebb: Also, not all transforms have helper functions (like the ones from PyTorch, and a few that I didn't add myself to Pyro...)

neerajprad: @stefanwebb - I added `num_flows` since that seemed like a common use case. Many experiments that I have seen just use a single transform repeated n times, so this is to accommodate for that use case without having to write custom helper functions. For more complex combinations, I foresee users using a `make_transforms` function like you suggest.

pyro/infer/autoguide/guides.py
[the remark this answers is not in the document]
neerajprad: Done.

pyro/infer/autoguide/guides.py
neerajprad: Done; changed to `transforms` everywhere.

pyro/infer/autoguide/guides.py
neerajprad: Okay, removed this.

pyro/infer/autoguide/guides.py
fritzo: @neerajprad re `num_flows`: I agree with @stefanwebb and would aim to "program with functions" rather than "with *args,**kwargs" if possible. You might replace the `num_flows` arg with a new higher-order helper function like
```py
def iterated(repeats, base_fn, *args, **kwargs):
    assert isinstance(repeats, int) and repeats >= 0
    return ComposeTransformModule([base_fn(*args, **kwargs)
                                   for _ in range(repeats)])
```
That way I think you could replace `num_flows` with something like
```diff
- AutoNormalizingFlow(init_transform_fn, num_flows)
+ AutoNormalizingFlow(partial(iterated, num_flows, init_transform_fn))
```

pyro/ops/integrator.py
```diff
@@ -83,7 +83,7 @@ def potential_grad(potential_fn, z):
         else:
             raise e
 
-    grads = grad(potential_energy, z_nodes)
+    grads = grad(potential_energy, z_nodes, retain_graph=True)
```
neerajprad: @fritzo, @fehiepsi - Ref to our discussion in #2283. I had to do this by default in the integrator, since this issue is not specific to HMC initialization, and we bump into the same issue when a proposal might be rejected at any point. Unless `retain_graph` is `True` we will end up hitting the cache but the cached value's `grad` would have been cleared. Storing the graph for all samples is not needed, but it is hard to say apriori when we need to do this. I didn't see any significant memory usage increase at least on our tests and examples, but let me know if this seems like a dicey change.

pyro/infer/autoguide/guides.py
neerajprad: Done.

pyro/ops/integrator.py
fehiepsi: I just worry that this will retain graphs of every leap_frog step. From #2283, if this just happens at `find_reasonable_step_size`, then it might be better to clone the variable in this function.

neerajprad: I think we can get away by just cloning in one part of the codebase, I'll make that change. We are not making any in place updates in the leapfrog integrator, but I agree that its best to avoid `retain_graph=True`.

neerajprad: I was able to resolve this by cloning. I still face an issue when trying to jit compile where the tracer complains that it cannot add a constant tensor with `requires_grad=True`; there's an xfailing test for that.

pyro/infer/autoguide/guides.py
```diff
@@ -656,44 +656,76 @@ def _loc_scale(self, *args, **kwargs):
...
-        return iaf_dist
+        a bijective transform.
+        """
+        if self.transform is None:
+            self.transform = self._init_transform_fn(self.latent_dim)
+        flow_dist = dist.TransformedDistribution(dist.Normal(0., 1.).expand([self.latent_dim]), self.transform)
+        return flow_dist
+
+
+class AutoIAFNormal(AutoNormalizingFlow):
```
neerajprad: @martinjankowiak - I have only kept `AutoIAFNormal` here since we had it earlier, but if needed, we can add convenience wrappers for other transforms like `AutoBNAFNormal`. Right now it seems pretty convenient to just use `make_transform` and `AutoNormalizingFlow` directly, but I'm happy to add more if anyone requests it.

martinjankowiak: yeah that seems fine for now. we'll probably want a `AutoBNAFNormal` but let's do that in another PR

pyro/infer/autoguide/guides.py
```diff
@@ -656,44 +656,76 @@ def _loc_scale(self, *args, **kwargs):
...
+        :func:`~pyro.distributions.transforms.affine_autoregressive`.
+    """
+
+    def __init__(self, model, hidden_dim=None, init_loc_fn=init_to_median, num_transforms=1, **init_transform_kwargs):
+        super().__init__(model,
+                         init_transform_fn=functools.partial(make_transform, affine_autoregressive,
+                                                             repeats=num_transforms,
+                                                             hidden_dims=hidden_dim,
+                                                             **init_transform_kwargs),
+                         init_loc_fn=init_loc_fn)
```
martinjankowiak: one potentially confusing issue for users is that currently these init functions are sort of vacuous and/or have unexpected behavior in the presence of transforms

neerajprad: Hmm..I hadn't thought about that. 

@fritzo - are we even using the initialization if we aren't calling `_init_loc()`?

fritzo: @neerajprad `init_loc_fn()` is still used by `InitMessenger` when creating the prototype trace. However if you don't care about the values in that trace, I would suggest hiding this from `AutoNormalizingFlow`.

neerajprad: Thanks for explaining, @fritzo. I think, in that case, it probably won't have much of an effect but we should remove it so as to avoid confusion. This will be a breaking change though for `AuotIAFNormal` and will require us to change a few tests where we use the init_fn arg with parametrized autoguide classes, so I would like to do it in a separate PR, if that's okay.

fritzo: @neerajprad I think we can avoid breaking changes if you want:
1. Remove `init_loc_fn` from the new `AutoNormalizingFlow` class (and passing either `init_to_feasible` or `init_to_sample` to `super().__init__()` since the default `init_to_median()` is more expensive and unused).
2. Preserve the `init_loc_fn` kwarg to `AutoIAFNormal` but discard its value?

neerajprad: Okay, done. I have added a `FutureWarning` so that we can remove it at some point, and just to warn users that it doesn't have any effect in practice. According to our guidelines, I suppose we shouldn't actually remove it until Pyro 2.0? 🙂

pyro/infer/autoguide/guides.py
```diff
@@ -656,7 +656,48 @@ def _loc_scale(self, *args, **kwargs):
...
+                                 repeats=2)
+        guide = AutoNormalizingFlow(model, transform_init)
+        svi = SVI(model, guide, ...)
+
+    :param callable model: a generative model
+    :param init_transform_fn: a callable which when provided with the latent
+        dimension returns an instance of :class:`~torch.distributions.Transform`
+        , or :class:`~pyro.distributions.TransformModule` if the transform has
+        trainable params.
+    :param callable init_loc_fn: A per-site initialization function.
```
fritzo: remove docs

pyro/distributions/torch_transform.py
```diff
@@ -15,3 +15,19 @@ def __init__(self, *args, **kwargs):
...
+
+class ComposeTransformModule(torch.distributions.ComposeTransform, torch.nn.ModuleList):
+    """
+    This allows us to use a list of `TransformModule` in the same way as
+    :class:`~torch.distributions.transform.ComposeTransform`. This is needed
+    so that transform parameters are automatically registered by Pyro's param
+    store when used in :class:`~pyro.nn.module.PyroModule` instances.
+    """
+    def __init__(self, parts):
+        super(ComposeTransformModule, self).__init__(parts)
```
fritzo: nit: `super().__init__(parts)`

neerajprad:
Thanks for the detailed review and help with the design - @fritzo, @martinjankowiak, @stefanwebb, @fehiepsi!

PR #2283, opened 2020-01-26
Patch Transform to unset cache on call to backward

neerajprad:
This problem was seen in #2282 when we use caching with one of the transforms, and can be seen in the following failure pattern (see test below too):

```python
x = torch.tensor(2., requires_grad=True)
t = ExpTransform(cache_size=1)
torch.autograd.grad(t(x), x)
# Error on second call
torch.autograd.grad(t(x), x)
```
, where the second call throws `RuntimeError: Trying to backward through the graph a second time`. 

To get around this, we can add a backward hook to the transformed values in `__call__` and `_inv_call` to unset `cached_x_y` after the first `grad` call. This ensures that we rebuild the autograd graph instead of reusing the previous transformed value on the second `grad` call even if the original argument remains the same.

### DIFF
--- a/pyro/distributions/torch_patch.py
+++ b/pyro/distributions/torch_patch.py
@@ -44,10 +44,14 @@ def _Transform__call__(self, x):
         return y_old
     y = self._call(x)
 
-    if self._cache_size > 0:
-        # unset cached value after call to backward
+    if self._cache_size > 0 and x.requires_grad:
+        weakself = weakref.ref(self)
+
+        # hook to unset cached value after call to backward
         def _hook(_):
-            self._cached_x_y = None, None
+            self = weakself()
+            if self is not None:
+                self._cached_x_y = None, None
 
         y.register_hook(_hook)
     self._cached_x_y = x, y
@@ -66,10 +70,14 @@ def _Transform_inv_call(self, y):
         return x_old
     x = self._inverse(y)
 
-    if self._cache_size > 0:
-        # unset cached value after call to backward
+    if self._cache_size > 0 and y.requires_grad:
+        weakself = weakref.ref(self)
+
+        # hook to unset cached value after call to backward
         def _hook(_):
-            self._cached_x_y = None, None
+            self = weakself()
+            if self is not None:
+                self._cached_x_y = None, None
 
         x.register_hook(_hook)
     self._cached_x_y = x, y
--- a/pyro/distributions/torch_patch.py
+++ b/pyro/distributions/torch_patch.py
@@ -33,6 +33,49 @@ def patch_dependency(target, root_module=torch):
     return decorator
 
 
+# See: tests.distributions.transforms.test_transforms.py::test_transform_backward
+# TODO: should be moved to PyTorch if this is a reasonable pattern
+@patch_dependency('torch.distributions.transforms.Transform.__call__')
+def _Transform__call__(self, x):
+    if self._cache_size == 0:
+        return self._call(x)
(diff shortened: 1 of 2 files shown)


pyro/distributions/torch_patch.py
```diff
@@ -33,6 +33,49 @@ def decorator(new_fn):
...
+    if x is x_old:
+        return y_old
+    y = self._call(x)
+
+    if self._cache_size > 0:
+        # unset cached value after call to backward
+        def _hook(_):
+            self._cached_x_y = None, None
+
+        y.register_hook(_hook)
```
fritzo: I believe this creates a cyclic reference (unless PyTorch uses weakrefs in hook logic).

@apaszke do you have recommendations for avoiding cyclic dependencies in hooks?

neerajprad: Good catch. I think one way might be to capture the handle as `self._handle = y.register_hook(_hook)` and to remove it `self._handle.remove()` when a new value of `y` is provided. I'll have to test it out, but there might be a simpler solution.

fritzo: Maybe
```py
weakself = weakref.ref(self)
def _hook(_):
    self = weakself()
    if self is not None:
        self._cached_x_y = None, None
```

neerajprad: That should work great!

pyro/distributions/torch_patch.py
```diff
@@ -33,6 +33,57 @@ def decorator(new_fn):
...
+@patch_dependency('torch.distributions.transforms.Transform.__call__')
+def _Transform__call__(self, x):
+    if self._cache_size == 0:
+        return self._call(x)
+    x_old, y_old = self._cached_x_y
+    if x is x_old:
+        return y_old
+    y = self._call(x)
+
+    if self._cache_size > 0 and x.requires_grad:
```
fritzo: How about restricting to leaf variables, since only they would be updated in-place?
```py
if self._cache_size > 0 and x.requires_grad and x.is_leaf:
```

fritzo:
@neerajprad I can see how this pattern would be useful, but I think we disagree on what it is useful for :smile: I think it would be useful to clear the cache when a leaf variable has ben updated in-place by an optimizer. However for the issue stated ("Trying to backward through the graph a second time"), I believe a better solution would be to use `retain_graph=True`. I don't know if there is a programmatic way to do this locally; ideally we could run the transform with a
```py
with torch.enable_retain_graph():
   ...
```

fritzo:
For example the following works fine:
```py
x = torch.tensor(2., requires_grad=True)
t = ExpTransform(cache_size=1)
torch.autograd.grad(t(x), x, retain_graph=True)
# Error on second call
torch.autograd.grad(t(x), x)
```

neerajprad:
I think using `retain_graph=True` will work in this case, but my expectation when using transforms is that they behave like stateless functions (despite the caching), and each time we call `t(x)` the autograd graph is recomputed by default. This case got triggered because we had some initial `x` in HMC that was used to find the optimal step size, and the same `x` was later used to initiate the MCMC chain in a different part of the code. This could also probably happen if the proposal is rejected and we compute the gradient again. We might be able to save compute by retaining the graph I suppose, but I'll need to think if it is safe to do so always.

neerajprad:
Thanks for the review and valuable comments, @fritzo. We can revisit this in PyTorch if other users ever face this. For the HMC issue, I will go with your suggestion of using `retain_graph=True` during initialization.

PR #2284, opened 2020-01-26
Use MaskedDistribution in MaskMessenger

ordabayevy:
As mentioned in #2279 make use of the `MaskedDistribution` in the `MaskMessenger`. Benefits could be:

- unify the logic of `MaskedDistribution` and `MaskMessenger`
- automatic broadcasting to the mask shape when the mask is enumerated
- possibly eliminating the `mask` field from trace sites @fritzo

The discussion is much welcomed.


ordabayevy:
Current build fails to pass `test_enum.py::test_elbo_enumerate_plate_2`. While trying to debug it I found out that in `pyro/infer/traceenum_elbo.py::_compute_model_factors` log_factors uses `site["packed"]["unscaled_log_prob"]` which is not masked. @fritzo can you give me a hint why that is the case? And does it imply that applying the mask to a distribution directly vs using the `poutine.mask` context manager in the model can behave differently?
```
with pyro.poutine.mask(mask=mask):
   pyro.sample("x", dist, infer={"enumerate": "parallel"})
```
vs
`pyro.sample("x", dist.mask(mask), infer={"enumerate": "parallel"})`

pyro/poutine/mask_messenger.py
```diff
@@ -31,5 +32,9 @@ def __init__(self, mask):
         self.mask = mask
 
     def _process_message(self, msg):
+        if msg["done"] or msg["type"] != "sample":
+            return None
+
         msg["mask"] = self.mask if msg["mask"] is None else self.mask & msg["mask"]
+        msg["fn"] = MaskedDistribution(msg["fn"], msg["mask"])
```
fritzo: If we replaced this with `msg["fn"].mask(msg["mask"])` then we could add short-circuit logic to avoid multiply wrapping a distribution in `MaskedDistribution. Just a thought.

fritzo:
Hi @ordabayevy, first thanks for taking on this ambitious and valuable refactoring. There is some tricky logic in `_compute_model_factors()` where we [avoid scaling](https://github.com/pyro-ppl/pyro/blob/73af4ff/pyro/infer/traceenum_elbo.py#L128). I believe using your suggestion we can remove these lines and replace `"masked_log_prob"` with `"unscaled_log_prob"` everywhere.

In fact, I believe we could simplify the `scale_and_mask` helpers to `scale_tensor()`, omitting masking logic. I think this is the biggest benefit of your refactoring, simplifying all that masking logic.

One issues is that increased use of `MaskedDistribution` may break existing pattern-matching logic; for example `TraceMeanField_ELBO` uses analytic KLs, and I think we would need to add `MaskedDistribution` patterns to `pyro/distributions/kl.py`. Similarly [pyro.infer.reparam.Reparam](https://github.com/pyro-ppl/pyro/blob/73af4ff/pyro/infer/reparam/reparam.py) has helpers to unwrap `Independent` distributions, and we may need to extend these to unwrap `MaskedDistribution`s as well. This cleanup can be done in a follow-up PR, but should be done before release.

Also I'm happy to chat on PyTorch slack if you want to walk over any code.

fritzo:
> does ... applying the mask to a distribution directly vs using the poutine.mask context manager in the model can behave differently?

I believe the two should be equivalent, but I may be forgetting details.

ordabayevy:
> Also I'm happy to chat on PyTorch slack if you want to walk over any code.
That would be great, however, I am not on PyTorch slack yet. Do I need an invitation?

fritzo:
Request an invitation [here](https://docs.google.com/forms/d/e/1FAIpQLSeADnUNW36fjKjYzyHDOzEB_abKQE9b6gqqW9NXse6O0MWh0A/viewform)

ordabayevy:
@fritzo sorry, it took me some time to get a better grasp of enumeration and dice loss algorithms before implementing new changes.

I tried to separate the logic of masking and scaling with `mask_tensor` and `scale_tensor`. Now `MaskMessenger` applies masking to sample sites except for those that are marginalized in the model (this is done kinda in a hacky way through `ReplayMessenger`).

Another change I made is that `mask.shape` and `base_dist.batch_shape` are not broadcasted anymore in `MaskedDistribution`. Mask is only applied when calculating the `log_prob` and `score_parts` so that in the guide sampling has the shape of the `base_dist.batch_shape` (this leads to the possibly undesired behavior of `log_prob` having a different shape broadcasted to `mask.shape`).

Minor note: I noticed a clash between the `mask` attribute of `MaskedMixture` and `mask` method of `MaskedDistribution`.

Overall, I am not sure if these changes simplified the logic of masking or made it more complicated.

PR #2285, opened 2020-01-26
Implement GaussianHMM.posterior().rsample()

### LINKED ISSUE #2182: FR parallel-scan GaussianHMM.rsample()
`GaussianHMM.rsample()` would enable a powerful class of guides for time-series models with non-Gaussian likelihoods. Consider for example a guide:
```py
from pyro.distributions import GaussianHMM
from pyro.ops.gaussian import Gaussian

def model(data):
    p_z = GaussianHMM(*my_prior)  # prior, say a Matern GP
    z = pyro.sample("z", p_z)  # samples from prior
    p_x = dist.Normal(*noisy_nonlinear_function(z))
    pyro.sample("x", p_x, obs=data)  # likelihood

def guide(data):
    p_z = GaussianHMM(*my_prior)  # same as in the model
    p_x_given_z = Gaussian(*my_nn(data))  # independent over time
    p_z_given_x = forward_filter(p_z, p_x_given_z)  # updates trans_dist
    assert isinstance(p_z_given_x, GaussianHMM)  # posterior is still Gaussian
    pyro.sample("z", p_z_given_x)  # samples from posterior via GaussianHMM.rsample()
```
In this guide, `my_nn` acts as a weak encoder, forming a gaussian update `(log_normalizer, info_vec, precision)` independently at each time step, kind of a neural EP. In principle it is easy to fuse these weak likelihoods with a strong prior, e.g. a Matern GP, and then draw a posterior sample using parallel-scan.

This guide would work well with either continuous observations and `Trace_ELBO` loss, or discrete observations (e.g. `Poisson`) and a `Trace_CRPS` loss. I'd like to add this class of guides to `pyro.contrib.timeseries`.

fritzo:
Resolves #2182 

1. Adds a method `GaussianHMM.posterior(likelihood)` that produces a new `GaussianHMM` instance incorporating the original `GaussianHMM` thought of as a prior, and the `likelihood` distribution (a MVN or Independent(Normal)).
2. Implements a parallel-scan forward-filter backward-rsample algorithm `_sequential_gaussian_filter_sample()`.
3. Refactors `GaussianHMM.rsample()` to use parallel scan, and removes old lazy logic to support the previous linear `.rsample()` implementation from #2238 .

## Tested
- [x] shape test for `_sequential_gaussian_filter_sample()`
- [x] shape test for `GaussianHMM.posterior()`
- [x] distribution test for `GaussianHMM.posterior().rsample()`
- [x] distribution test for `GaussianHMM.posterior().log_prob()`
- [x] test the computed `log_normalizer`


pyro/distributions/hmm.py
```diff
@@ -318,39 +393,23 @@ def __init__(self, initial_dist, transition_matrix, transition_dist,
...
-        return mvn_to_gaussian(self.initial_dist).expand(self.batch_shape)
-
-    @lazy_property
-    def _trans(self):
-        return matrix_and_mvn_to_gaussian(self.transition_matrix, self.transition_dist)
-
-    @lazy_property
-    def _obs(self):
-        return matrix_and_mvn_to_gaussian(self.observation_matrix, self.observation_dist)
+        self._init = mvn_to_gaussian(initial_dist).expand(batch_shape)
```
fritzo: The next few changes simply revert the eager->lazy refactoring of #2238

pyro/distributions/hmm.py
```diff
@@ -405,6 +465,38 @@ def filter(self, value):
...
+        :rtype: GaussianHMM
+        """
+        assert (isinstance(likelihood, torch.distributions.MultivariateNormal) or
+                (isinstance(likelihood, torch.distributions.Independent) and
+                 isinstance(likelihood.base_dist, torch.distributions.Normal)))
+        new = self._get_checked_instance(GaussianHMM)
+        new.hidden_dim = self.hidden_dim
+        new.obs_dim = self.obs_dim
+        new._trans = self._trans
+        new._obs = self._obs + mvn_to_gaussian(likelihood).event_pad(left=self.hidden_dim)
```
fritzo: Hmm at this point the distribution is no longer normalized, so `.log_prob()` will be invalid. I think I'll change to compute the log normalizer here. We may even be able to return that for use in inference.

pyro/distributions/hmm.py
```diff
@@ -90,6 +89,82 @@ def _sequential_gaussian_tensordot(gaussian):
...
+            cond = cond.reshape(shape + (-1, 2 * state_dim))  # [z0z2]
+            sample = joint.condition(cond).rsample()  # [z1]
+            sample = torch.cat([sample, result[..., -1:, :]], dim=-2)  # [z1, z3]
+            result = torch.stack([
+                result[..., :-1, :],  # [z0, z2]
+                sample,  # [z1, z3]
+            ], dim=-2)  # [[z0, z1], [z2, z3]]
+            result = result.reshape(shape + (-1, state_dim))  # [z0, z1, z2, z3]
+
+    return result[..., 1:, :]  # [z1, z2, z3, ...]
```
fehiepsi: Amazing work!!

pyro/distributions/hmm.py
fritzo: I've fixed and updated to use the new `.conjugate_update()` interface of #2289 .

tests/distributions/test_hmm.py
```diff
@@ -355,26 +387,35 @@ def test_gaussian_hmm_distribution(diag, sample_shape, batch_shape, num_steps, h
     expected_log_prob = logp.log_density(unrolled_data)
     assert_close(actual_log_prob, expected_log_prob)
 
-    # Test mean and covariance.
+    d_posterior, log_normalizer = d.conjugate_update(like_dist)
+    assert_close(d.log_prob(data) + like_dist.log_prob(data),
+                 d_posterior.log_prob(data) + log_normalizer)
```
fritzo: This test is similar to tests/distributions/test_conjugate_update.py in #2289

pyro/distributions/hmm.py
fritzo: Well, I think this demonstrates that we have found a useful abstraction in `Gaussian` and Funsors. @eb8680 and I are working on the generalized Funsor version in https://github.com/pyro-ppl/funsor/issues/304

fritzo:
After this merges, I can add `GaussianGMM` `ConjugateReparam` tests to #2289

Issue #2288, opened 2020-01-28
False negatives in check_model_guide_match()


fritzo:
@mbabadi writes:

> After a certain version of Pyro (I didn't keep track), model sampling sites without matching guide sampling sites are acceptable (no exception is thrown). I believe the current behavior is that Pyro effectively sets the posterior to the prior. We were used to the fail-fast behavior, and this led to wrong inferences in an application where we forgot a guide sampling statement by mistake.

fritzo:
@mbabadi could you provide an example? I haven't been able to reproduce this and I don't see where the error might be happening in [check_model_guide_match()](https://github.com/pyro-ppl/pyro/blob/73af4ff/pyro/util.py#L172). Feel free to close this issue if the problem has been solved.

Issue #1723, opened 2019-01-30
[FR] Conjugate operations on existing distributions


fritzo:
Exact variable elimination of continuous random variables requires a number of new distribution methods. Some of the mathematics for these is already implemented in [pyro.contrib.tracking](http://pyro.ai/examples/ekf.html). To make variable elimination algorithms generic, we should move this math into new methods of subclasses of `TorchDistribution` or new functions in `pyro.distributions.conjugate`.

Note this issue concerns only low-level distributions methods, not high-level inference algorithms. 

## Interface

This is WIP.

Univariate distributions can perform "Bayesian fusion" aka "filter update", equivalent to multiplying densities pointwise, i.e. adding log densities pointwise.
```py
def fuse(dist1, dist2):
    assert type(dist1) is type(dist2)
    # returns (log_prob, updated dist)
    # dispatch on type-type pairs, following kl_divergence()
```
Conjugate pairs can be conditioned on distribution-valued observations (see #988 )
```py
def observe(param_dist, obs_dist):
    # returns (log_prob, updated dist of type param_dist)
    # dispatch on type-type pairs
```
Multivariate distributions like `MultivariateNormal` can encode multiple univariate distributions. Maybe we can support these by allowing transforms?
```py
def transform_and_observe(param_dist, transform, obs_dist):
    # returns (log_prob, updated dist of type param_dist)
```

## Tasks:

- [ ] settle on interface
- [ ] implement methods for `Normal`
- [ ] implement methods for `Beta`
- [ ] implement methods for `Binomial`: `.as_posterior()`
- [ ] implement methods for `MultivariateNormal`
- [ ] implement methods for `Dirichlet`

fritzo:
Closing in favor of [Funsor](https://github.com/pyro-ppl/funsor), which can be used in a number of ways:
- as a complete backend for pyro
- to implement custom Pyro distributions
- as a DSL within Pyro models, constructing factors via `pyro.factor` statements.

PR #2289, opened 2020-01-28
Implement .conjugate_update() and ConjugateReparam

fritzo:
Addresses #1723, #2182 
Similar to https://github.com/pytorch/pytorch/pull/32177

This adds `ConjugateReparam` as a new facility to handle conjugate distributions.

The motivation is to handle the new conjugate pair (`GaussianHMM`, `Independent(MultivariateNormal)`) as in #2285. Additionally this may be able to replace `pyro.contrib.conjugate`.

## Changes

- [x] Adds a `ConjugateReparam` class for reparameterization.
- [x] Adds a new method interface `Distribution.conugate_update()`, as needed in #2285.
    - [x] `Beta` for `Bernoulli` and `Binomial`
    - [x] `Dirichlet` for `Categorical` and `Multinomial`
    - [x] `Gamma` for `Poisson`
- [x] Adds logic to allow `poutine.trace` to ignore sites sampled from the prior. This is needed because conjugate-reparameterized sites are sampled from the posterior (or a learned approximate posterior), and thus should not be handled by SVI or MCMC.
- [x] Adds functionality to `Reparam` to have access to the `*args,**kwargs` provided to a function that it decorates (typically model inputs, but possibly inputs to a sub-model as subroutine). This is needed for data-dependent conjugate updates.

## Triaged
- This does not implement conjugation operations : `p(x|z)` -> `q(z|x)`, so the onus is on the user to provide exact or approximate `q(z|x)`. For example in a Gamma-Poisson model, the user must hand-conjugate to manually construct a `Gamma` equivalent to the funsor expression
    ```py
    p_x_given_z = Poisson(Variable("z"), value=data)
    q_z_given_x = p_x_given_z - p_x_given_z.reduce(ops.logaddexp)  # a Gamma funsor
    ```
    I'd like to see what patterns we develop before adding helpers.

## Tested
- [x] `log_normalizer` computation in posterior methods
- [x] `ConjugateReparam` on a Beta-Binomial model
- [x] `ConjugateReparam` on a `GaussianHMM`
- [x] combine `ConjugateReparam`+`LinearHMMReparam`+`StableReparam`
- [x] inference in SVI
- [x] inference for HMC
- [x] data-dependent likelihood
- [x] test that spurious model sites raise an error #2288

### DIFF
--- a/pyro/distributions/distribution.py
+++ b/pyro/distributions/distribution.py
@@ -112,3 +112,22 @@ class Distribution(object, metaclass=ABCMeta):
         :rtype: iterator
         """
         raise NotImplementedError("Support not implemented for {}".format(type(self)))
+
+    def conjugate_update(self, other):
+        """
+        Creates an updated distribution fusing information from another
+        compatible distribution. This is supported by only a few conjugate
+        distributions.
+
+        This should satisfy::
+
+            fg, log_normalizer = f.conjugate_update(g)
+            assert f.log_prob(x) + g.log_prob(x) == fg.log_prob(x) + log_normalizer
+
+        :param other: A distribution representing ``p(data|self.probs)`` but
+            normalized over ``self.probs`` rather than ``data``.
+        :return: a pair ``(updated,log_normalizer)`` where ``updated`` is an
+            updated distribution of type ``type(self)``, and ``log_normalizer``
+            is a :class:`~torch.Tensor` representing the normalization factor.
+        """
+        raise NotImplementedError("{} does not support .conjugate_update()".format(type(self)))
--- a/pyro/distributions/distribution.py
+++ b/pyro/distributions/distribution.py
@@ -111,7 +111,7 @@ class Distribution(object, metaclass=ABCMeta):
         :return: An iterator over the distribution's discrete support.
         :rtype: iterator
         """
-        raise NotImplementedError("Support not implemented for {}".format(type(self)))
+        raise NotImplementedError("Support not implemented for {}".format(type(self).__name__))
 
     def conjugate_update(self, other):
         """
@@ -130,4 +130,5 @@ class Distribution(object, metaclass=ABCMeta):
             updated distribution of type ``type(self)``, and ``log_normalizer``
             is a :class:`~torch.Tensor` representing the normalization factor.
         """
-        raise NotImplementedError("{} does not support .conjugate_update()".format(type(self)))
+        raise NotImplementedError("{} does not support .conjugate_update()"
+                                  .format(type(self).__name__))
--- a/pyro/distributions/distribution.py
+++ b/pyro/distributions/distribution.py
@@ -132,7 +132,7 @@ class Distribution(object, metaclass=ABCMeta):
         :func:`~funsor.pyro.convert.tensor_to_funsor` ::
 
             dist_to_funsor(f) + dist_to_funsor(g)
-              == dist_to_funsor(gf) + tensor_to_funsor(log_normalizer)
+              == dist_to_funsor(fg) + tensor_to_funsor(log_normalizer)
 
         :param other: A distribution representing ``p(data|latent)`` but
             normalized over ``latent`` rather than ``data``. Here ``latent``
--- a/pyro/distributions/distribution.py
+++ b/pyro/distributions/distribution.py
@@ -119,13 +119,25 @@ class Distribution(object, metaclass=ABCMeta):
         compatible distribution. This is supported by only a few conjugate
         distributions.
 
-        This should satisfy::
+        This should satisfy the equation::
 
             fg, log_normalizer = f.conjugate_update(g)
             assert f.log_prob(x) + g.log_prob(x) == fg.log_prob(x) + log_normalizer
 
-        :param other: A distribution representing ``p(data|self.probs)`` but
-            normalized over ``self.probs`` rather than ``data``.
+        Note this is equivalent to :obj:`funsor.ops.add` on
+        :class:`~funsor.terms.Funsor` distributions, but we return a lazy sum
+        ``(updated, log_normalizer)`` because PyTorch distributions must be
+        normalized.  Thus :meth:`conjugate_update` should commute with
+        :func:`~funsor.pyro.convert.dist_to_funsor` and
+        :func:`~funsor.pyro.convert.tensor_to_funsor` ::
+
+            dist_to_funsor(f) + dist_to_funsor(g)
+              == dist_to_funsor(gf) + tensor_to_funsor(log_normalizer)
+
+        :param other: A distribution representing ``p(data|latent)`` but
+            normalized over ``latent`` rather than ``data``. Here ``latent``
+            is a candidate sample from ``self`` and ``data`` is a ground
+            observation of unrelated type.
         :return: a pair ``(updated,log_normalizer)`` where ``updated`` is an
             updated distribution of type ``type(self)``, and ``log_normalizer``
             is a :class:`~torch.Tensor` representing the normalization factor.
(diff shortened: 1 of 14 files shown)


pyro/poutine/reparam_messenger.py
```diff
@@ -13,6 +13,10 @@ class ReparamMessenger(Messenger):
     constructor.  See the :mod:`pyro.infer.reparam` module for available
     reparameterizers.
 
+    Note some reparameterizers can examine the ``*args,**kwargs`` inputs of
```
fritzo: The changes in this file are to allow data-dependent reparameterizers, e.g. `q(z|x)` where `x` is input as model `*args,**kwargs`.

pyro/poutine/trace_messenger.py
```diff
@@ -118,7 +118,7 @@ def _reset(self):
         super()._reset()
 
     def _pyro_post_sample(self, msg):
-        if not self.param_only:
+        if not self.param_only and not msg["infer"].get("_do_not_trace"):
```
fritzo: This is a bit of a hack, but the simplest alternative I can think of is to add lots of logic to `get_importance_trace()` and MCMC utilities to discard relevant traced values. I am open to alternatives that cost < 5loc.

tests/infer/test_valid_models.py
```diff
@@ -116,6 +116,27 @@ def guide():
         assert_ok(model, guide, elbo)
 
 
+@pytest.mark.parametrize("Elbo", [
```
fritzo: I've added this test to sanity check #2288 in the presence of new tracing logic for `"_do_not_trace"`.

pyro/poutine/trace_messenger.py
eb8680: This seems fine but can you add an assertion
```
assert msg["infer"].get("is_auxiliary") and not msg["is_observed"])
```
and set `is_auxiliary` in `ConjugateReparam.__call__`?

fritzo: done.

pyro/distributions/distribution.py
```diff
@@ -111,4 +111,24 @@ def enumerate_support(self, expand=True):
...
+        compatible distribution. This is supported by only a few conjugate
+        distributions.
+
+        This should satisfy::
+
+            fg, log_normalizer = f.conjugate_update(g)
+            assert f.log_prob(x) + g.log_prob(x) == fg.log_prob(x) + log_normalizer
+
+        :param other: A distribution representing ``p(data|self.probs)`` but
+            normalized over ``self.probs`` rather than ``data``.
```
fehiepsi: It took me a while to understand what's going on here... (it is not easy to interpret your `likelihood` terms in the tests). Nice idea! 

nit: probably you want to replace `self.probs` by `latent`.

fritzo: Thanks for reviewing! For intuition those likelihood terms in the test are "conjugated" i.e. "transposed-and-normalized" likelihoods, which are easier to interpret as Funsors (see Triaged comment in this PR description).

fritzo: Done, I've also added a dosctring note on the relation to `Funsor.__add__()`.

pyro/infer/reparam/conjugate.py
```diff
@@ -0,0 +1,88 @@
...
+
+    :param guide: A likelihood distribution or a callable returning a
+        guide distribution. Only a few distributions are supported, depending
+        on the prior distribution's
+        :meth:`~pyro.distributions.Distribution.conjugate_update`
+        implementation.
+    :type guide: ~pyro.distributions.Distribution or callable
+    """
+    def __init__(self, guide):
+        self.guide = guide
```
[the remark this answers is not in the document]
fritzo: We do not currently need to call `super().__init__()` because `Reparam` has no `__init__()` method; however I could add `super().__init__()` calls to all `Reparam` subclasses if you would like.

pyro/distributions/distribution.py
fehiepsi: Understood!

pyro/distributions/torch_distribution.py
```diff
@@ -318,6 +318,12 @@ def mean(self):
     def variance(self):
         return self.base_dist.variance
 
+    def conjugate_update(self, other):
```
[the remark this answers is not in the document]
fritzo: Done, and I've also added `Independent.conjugate_update()`.

Note that current `.conjugate_update()` methods use pattern matching and I suspect we will later need to add unwrapping logic in those patterns (for `Independent`, `Masked`, `Expanded` etc.). Currently the methods only destructure the left hand side, not the argument.

pyro/infer/reparam/conjugate.py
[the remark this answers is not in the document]
neerajprad: Thanks for explaining, we can probably add it later if `Reparam` gets an `__init__`. I was just wondering where `self.args_kwargs` was coming from, but I see that  it is being set in `ReparamMessenger`.

tests/infer/reparam/test_conjugate.py
```diff
@@ -0,0 +1,210 @@
...
+            pyro.sample("x", dist.Normal(z, 1).to_event(2), obs=data)
+
+    # Test that we can combine these two reparameterizers.
+    reparam_model = poutine.reparam(model, {
+        "z": LinearHMMReparam(StableReparam(), StableReparam(), StableReparam()),
+    })
+    reparam_model = poutine.reparam(reparam_model, {
+        "z": ConjugateReparam(dist.Normal(data, 1).to_event(2)),
+    })
+    reparam_guide = AutoDiagonalNormal(reparam_model)  # Models auxiliary variables.
```
fritzo: This test is the culmination of months of reparameterizer work. It starts with an inhomogeneous Levy Stable process, applies `LinearHMMReparam` which calls `StableReparam` three times under the hood resulting in a conditional `GaussianHMM`, then applies `ConjugateReparam` to make a conjugate update to the conditional `GaussianHMM`, and finally uses an autoguide to model the auxiliary variables introduced by `StableReparam()`. Running the model calls parallel-scan forward-filter backward-sample. Hurray for composable inference 🎉

pyro/infer/reparam/conjugate.py
```diff
@@ -0,0 +1,89 @@
...
+        # Compute importance weight. Let p(z) be the original fn, q(z|x) be
+        # the guide, and u(z) be the conjugate_updated distribution. Then
+        #   normalizer = p(z) q(z|x) / u(z).
+        # Since we've sampled from u(z) instead of p(z), we
+        # need an importance weight
+        #   p(z) / u(z) = normalizer / q(z|x)                       (Eqn 1)
+        # Note that q(z|x) is often approximate; in the exact case
+        #   q(z|x) = p(x|z) / integral p(x|z) dz
+        # and since this site and the downstream likelihood site will have combined density
+        #   normalizer q(z|x) p(x|z) = normalizer / integral p(x|z) dz
```
neerajprad: Should this be `(normalizer / q(z|x)) p(x|z) = normalizer integral p(x|z) dz `?

fritzo: Good catch, I've fixed the comment.

fritzo:
@neerajprad thanks for help with this PR and careful review!

neerajprad:
Ready to merge? I'm looking forward to getting rid of `contrib.conjugate`. 😄

fritzo:
Ready to merge!  Note we'll probably want to refactor and optimize before we eliminate `pyro.contrib.conjugate` (e.g. split the `Delta.log_density` into two parts, half of which can be blocked if we also block the downstream likelihood, as we have discussed).

Issue #1706, opened 2019-01-08
Adaptation in beta-binomial model results in slow mixing with float tensor


neerajprad:
This issue was already discovered by @fehiepsi, and discussed in https://github.com/uber/pyro/pull/1678, but I was recently bitten by it again. As such, this is to track this issue and discuss any solutions / fixes, if there is indeed something that can be fixed. 

e.g. model that works very well with default tensor type `torch.DoubleTensor` (as set currently), but is slow with `torch.FloatTensor`. The reason for the slowness is that during the adaptation, we keep decreasing the step size to a very small value. This results in very slow mixing of the chains, as can be seen from the diagnostics.

```python
import pymc3 as pm
import pyro

import pyro.distributions as dist
from pyro.infer.mcmc import MCMC, NUTS
import theano.tensor as tt

import torch

torch.set_printoptions(precision=10)
torch.set_default_tensor_type('torch.DoubleTensor')


def pm_model(data):
    with pm.Model() as model:
        p = pm.Beta('p', 1., 1.)
        p_print = tt.printing.Print('p')(p)
        pm.Binomial('obs', data['n'], p_print, observed=data['x'])
    return model


def pyro_model(data):
    p = pyro.sample('p', dist.Beta(1., 1.))
    pyro.sample('obs', dist.Binomial(data['n'], p), obs=data['x'])


def get_samples_pymc(data, num_samples=200, warmup_steps=200):
    data = {k: v.numpy() for k, v in data.items()}
    with pm_model(data):
        trace = pm.sample(draws=num_samples, tune=warmup_steps, chains=2)


def get_samples_pyro(data, num_samples=200, warmup_steps=200):
    nuts_kernel = NUTS(pyro_model,
                       adapt_step_size=True,
                       adapt_mass_matrix=True,
                       jit_compile=True,
                       full_mass=True)
    mcmc_run = MCMC(nuts_kernel,
                    num_samples=num_samples,
                    warmup_steps=warmup_steps,
                    num_chains=2).run(data)
    print(mcmc_run.marginal(sites=['p']).diagnostics())


data = {'n': torch.tensor(5000000.), 'x': torch.tensor(3849.)}
#get_samples_pymc(data)
get_samples_pyro(data)
```

It seems that the reason is that our precision when using float tensors is too low - instead of fine-tuning the step size around a small value to meet the acceptance probability threshold, we end up massively decreasing it. Since we don't see this issue with pymc3, I am guessing that it could point to:
[...]

PR #2291, opened 2020-01-29
Add cheap test for log joint computation in neutra reparametrizer

neerajprad:
I have been trying to understand why the results from neutraHMC are sub par on a few examples that I have tried out, and this test is to just validate the log joint computation (which looks fine). I am still debugging other issues with the example.

### DIFF
--- a/pyro/infer/reparam/neutra.py
+++ b/pyro/infer/reparam/neutra.py
@@ -49,7 +49,7 @@ class NeuTraReparam(Reparam):
         self.x_unconstrained = []
 
     def _reparam_config(self, site):
-        if site["name"] in self.guide.prototype_trace:
+        if site["name"] in self.guide.prototype_trace and not site['is_observed']:
             return self
 
     def reparam(self, fn=None):
--- a/tests/infer/reparam/test_neutra.py
+++ b/tests/infer/reparam/test_neutra.py
@@ -40,7 +40,7 @@ def test_neals_funnel_smoke():
     model = neutra.reparam(neals_funnel)
     nuts = NUTS(model)
     mcmc = MCMC(nuts, num_samples=50, warmup_steps=50)
-    mcmc.run()
+    mcmc.run(dim)
     samples = mcmc.get_samples()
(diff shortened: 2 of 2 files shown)


fritzo:
Test looks reasonable, but is there a way to check for sign errors in e.g. the `log_abs_det_jacobian`?  I guess you could flip the sign in the implementation and check that results are even worse 🤔

neerajprad:
> Test looks reasonable, but is there a way to check for sign errors in e.g. the log_abs_det_jacobian?

I tried that but didn't see any improvements. The guide is actually learning something useful but running HMC on the warped space seems to take a long time and the samples look particularly bad. I might add more tests during the course of debugging, thanks for the tip!

fehiepsi:
> running HMC on the warped space seems to take a long time

I think the slowness is expected. There might be some discontinuities in log_prob computation. Did you try BNAF instead of IAF? (I didn't have any success with iaf :(

neerajprad:
> I think the slowness is expected. There might be some discontinuities in log_prob computation. Did you try BNAF instead of IAF? (I didn't have any success with iaf :(

I'll update the other PR so that you can follow where I am up. I have even simplified to using a single mode banana gaussian distribution (with a multi-modal distribution, results can be pretty random when fitting the guide since SVI can choose to focus on only one of the modes, so I wanted to test with a simpler setting). With that, IAF does a pretty reasonable job of fitting (not in the tails though), and I ideally wanted to see HMC fill out the tails, but the results are quite terrible with the samples falling along one vertical line, and pointing to some more serious bug. I am not too concerned about the speed until this issue is rectified.

fehiepsi:
Yup, let's try to isolate the issue. I can recall this [old issue](https://github.com/pyro-ppl/pyro/issues/1706#issuecomment-453151242) with discontinuity, which might be relevant. When you make the other PR, I will try to see if the issue happens with NumPyro.

neerajprad:
Thanks, Du. I just created https://github.com/pyro-ppl/pyro/pull/2294 to discuss the issue.

PR #2293, opened 2020-01-30
Update BART dataset to include 2019 data

fritzo:
Updated dataset is at http://fritzo.org/media/bart_full.pkl.bz2


neerajprad:
The dataset should be updated on S3.

PR #2294, opened 2020-01-30
Notebook to explore Neutra reparametrizer

neerajprad:
This is to replace https://github.com/pyro-ppl/pyro/pull/2268 in favor of a temporary notebook for discussion. Once we have discussed the issues involved and this example is working, we can convert it into a script. For discussion purposes, I think a notebook will be a better format. 

This has existing changes from #2282. See [notebook](https://github.com/pyro-ppl/pyro/blob/0cc222c785617a55a173a0e8a2cf5b9999b55e74/tutorial/source/neutra_hmc.ipynb)


fehiepsi:
@neerajprad I did some experiments with numpyro in [this gist](https://gist.github.com/fehiepsi/ad865d22bdd0759a8beb44767015451d). Some of my observations so far:
+ There is a bug in bnaf implementation in numpyro (I will make a PR for it soon).
+ The result in numpyro is better (I used num_flows=1 but it is also better with num_flows=5; for this simple problem, num_flows=1 is already good enough IMO). I am not sure what is the reason with Pyro - may be `eps` stuffs in Pyro implementation of bnaf plays some roles here...
+ We can see neutraHMC does a better job than bnaf alone; it is also much better than vanilla HMC in terms of n_eff. However, I still can't recover the full tail. This is mainly due to the fact that BNAF transform sets an upper bound to the codomain (see last cells in my gist). I think `tanh` activation plays a role in this limitation. I'll see if `residual` can help for it.

neerajprad:
Thanks for helping debug this, @fehiepsi. 

> The result in numpyro is better (I used num_flows=1 but it is also better with num_flows=5; for this simple problem, num_flows=1 is already good enough IMO)

Another thing that I noticed is that BNAF is slower than IAF in Pyro though it is expected to be faster in practice. Maybe that has to do with the default parameters like number of hidden units, but I haven't looked into it in detail.

That said, even with the fit that we see in Pyro, I would expect the neutra reparametrizer to not completely fail as it is doing now, which makes me suspect that there is something else outside just the autoguide fit.

fehiepsi:
@neerajprad I got pretty good result by adding `num_particles=10` to AutoContinuousELBO and can recover posterior tails for both bnaf and neutra. About the slowness, I think it is expected because by default, BNAF (with hidden_factors=[8, 8], we have hidden_dims=[16, 16] for this 2D data) has higher hidden_dim than IAF (with hidden_dims=[3 * 2 + 1] = [7] in make_autoregressive or hidden_dims[2] in AutoIAFNormal)).

I would suggest plotting the potential_fn of neutra_model (with only 1 flow for simplicity) to see if there are problems with it. It is surprising to me that autoguide can give moderate samples but neutraHMC fails. I highly suspect that discontinuity is happening somewhere here.

martinjankowiak:
@fehiepsi maybe you can plot some random 1-dimensional subsections of the potential function with high granularity and see if you find any sharp discontinuities....

neerajprad:
> I would suggest plotting the potential_fn of neutra_model (with only 1 flow for simplicity) to see if there are problems with it.

Thanks for the tip. It does seem that all the samples in the warped space are squished along the x axis and the PE function looks really off (or, I could be doing something odd).

neerajprad:
My previous plots were misleading since I was plotting the potential energy directly (in the log scale). The revised potential energy plots look much more reasonable. I have plotted the results with the affine transform which is actually not bad at all. Thanks @martinjankowiak for the suggestion. So it does seem that the problem is somewhere in the normalizing flow code.

neerajprad:
Okay, I think there were just some issues in my code, which are fixes. Things seem to run fine and AutoBNAF gives better results at least in terms of effective sample size and no divergences (vanilla HMC results in 4 divergent transitions). Visually, the quality of the samples is clearly better than the affine transform, but unclear if it is better than vanilla on this simple toy dataset. Thanks for helping debug this, @fehiepsi, @martinjankowiak.

PR #2295, opened 2020-01-30
Block non-stochastic sites from AutoGuide prototype_trace

fritzo:
As @neerajprad observed in #2291, `AutoGuide.prototype_trace` currently records all sites, which I find unexpected. This PR changes to record only stochastic sites, saving memory and slightly simplifying logic in neutra.

### DIFF
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -58,7 +58,7 @@ def _deep_setattr(obj, key, val):
     setattr(obj, rpart, val)
 
 
-def _prototype_hide_fn(msg):
+def prototype_hide_fn(msg):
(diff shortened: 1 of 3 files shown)


pyro/infer/autoguide/guides.py
```diff
@@ -58,6 +58,11 @@ def _getattr(obj, attr):
     setattr(obj, rpart, val)
 
 
+def prototype_hide_fn(msg):
+    # Record only stochastic sites in the prototype_trace.
+    return msg["type"] != "sample" or msg["is_observed"] or site_is_subsample(msg)
```
neerajprad: Nice! Thanks for the change.

Issue #2292, opened 2020-01-30
Gradient on transforms currently fail with in-place modification to tensor attributes


neerajprad:
This issue follows from the discussion in #2283. If we have a suitable path forward, I may reopen this in PyTorch, but this is to just jot down possible failure modes which may not be obvious at a glance. 

Many of our normalizing flows use `cache_size=1` since computing the inverse is not feasible. e.g. `BlockAutoregressive`. Transforms use identity checks when storing (x, y) pairs. As such, caching will be an issue when tensors are modified in-place. While we rarely do that in Pyro, a more subtle error arises when we change tensor attributes like `requires_grad`.

There are three kinds of issues (other than those arising from direct in-place modification of tensor values) that can happen:

1. Transforming x and later trying to differentiate wrt `x.requires_grad_(True)`.

```python
>>> t = ExpTransform(cache_size=1)
>>> x = torch.tensor(2.)
>>> t(x)
>>> x.requires_grad_(True)
>>> torch.autograd.grad(t(x), x)

RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
```

2. Differentiating w.r.t. the same tensor twice.

```python
>>> t = ExpTransform(cache_size=1)
>>> x = torch.tensor(2., requires_grad=True)
>>> torch.autograd.grad(t(x), x)  # success
>>> torch.autograd.grad(t(x), x)

RuntimeError: Trying to backward through the graph a second time, but the buffers have already been freed. Specify retain_graph=True when calling backward the first time
```

3. Exception thrown when JIT tracing.
JIT tracing can throw errors like:
```
RuntimeError: Cannot insert a Tensor that requires grad as a constant. Consider making it a parameter or input, or detaching the gradient
```
  TODO: isolate a minimal example. 

Note that to resolve all of these issues #2283 is not sufficient (it will only resolve 2.). Cloning tensors can resolve 1 and 2, but I am still facing issues with 3.

andersdot:
I'm hitting the issue #3 above with JIT tracing. I have a pre-trained pytorch normalizing flow model as a prior, which I wrapped into a pyro distribution. When I run `MCMC` using the `NUTS` kernel with `jit_compile=False` it works (though very slow) but with `jit_compile=True`, I get the error you show above

`RuntimeError: Cannot insert a Tensor that requires grad as a constant. Consider making it a parameter or input, or detaching the gradient`

fritzo:
[...]

PR #2298, opened 2020-02-04
Add NeuTra reparam example

neerajprad:
This adds a NeuTra reparametrizer example on a banana shaped distribution. See [neutra.pdf](https://github.com/pyro-ppl/pyro/files/4150898/neutra.pdf), following discussion in #2294. 

Regarding the API, here are a few things that might need further discussion (separately from this PR):
 - I have changed the `AutoDiagonalNormal` and `AutoMultivariateNormal` guides to use a `TransformedDistribution` so that we can use it with `NeuTraReparam`. 
   - Should we expose `transform` and `base_dist` be available as properties for these classes directly?
   - I think these two are good enough, but I'm not sure if `LowRankMultivariateNormal` should also be written as a `TransformedDistribution`. 
 - Currently, we require users to set `guide.requires_grad_(False)` before passing this to `NeuTraReparam`, since we don't want guide parameters to be updated while running HMC. We can probably do this inside `NeuTraRepam` itself by default (if the guide is a `PyroModule` instance), but I'm not sure if changing the `requires_grad` flag for parameters in the param store is something that might be a surprising side effect for users. Alternative suggestions are welcome. cc. @martinjankowiak

### DIFF
--- a/examples/neutra.py
+++ b/examples/neutra.py
@@ -208,4 +208,3 @@ if __name__ == '__main__':
 
     args = parser.parse_args()
     main(args)
-
--- a/examples/neutra.py
+++ b/examples/neutra.py
@@ -164,7 +164,7 @@ def main(args):
             title='Posterior \n(BNAF autoguide)')
     sns.kdeplot(guide_samples[:, 0], guide_samples[:, 1], ax=ax6)
 
-    # 3(b). Draw samples using NeuTra HMC
+    # 4(b). Draw samples using NeuTra HMC
     logging.info('\nDrawing samples using BNAF autoguide + NeuTra HMC ...')
     neutra = NeuTraReparam(guide.requires_grad_(False))
     neutra_model = poutine.reparam(model, config=lambda _: neutra)
--- a/examples/neutra.py
+++ b/examples/neutra.py
@@ -1,0 +1,211 @@
+"""
+This example illustrates the use of `NeuTraReparam` to run neural transport HMC [1]
+on a toy model that draws from a banana-shaped bivariate distribution [2]. We first
+train an autoguide by using `AutoNormalizingFlow` that learns a transformation from
+a simple latent space (isotropic gaussian) to the more complex geometry of the
+posterior. Subsequently, we use `NeuTraReparam` to run HMC and draw samples from this
+simplified "warped" posterior. Finally, we use our learnt transformation to transform
+these samples back to the original space. For comparison, we also draw samples from
+a NeuTra-reparametrized model that uses a much simpler `AutoDiagonalNormal` guide.
+
+References:
+----------
+[1] Hoffman, M., Sountsov, P., Dillon, J. V., Langmore, I., Tran, D., and Vasudevan,
+ S. Neutra-lizing bad geometry in hamiltonian monte carlo using neural transport.
+ arXiv preprint arXiv:1903.03704, 2019.
+[2] Wang Z., Broccardo M., and Song J. Hamiltonian Monte Carlo Methods for Subset
+ Simulation in Reliability Analysis. arXiv preprint arXiv:1706.01435, 2018.
+"""
+
+import argparse
+import logging
+from functools import partial
+
+import torch
+import matplotlib.pyplot as plt
+import seaborn as sns
+from matplotlib.gridspec import GridSpec
+
+import pyro
+from pyro import optim, poutine
+import pyro.distributions as dist
+from pyro.distributions import constraints
+from pyro.distributions.transforms import iterated, block_autoregressive
+from pyro.infer import MCMC, NUTS, SVI, Trace_ELBO
+from pyro.infer.autoguide import AutoDiagonalNormal, AutoNormalizingFlow
+from pyro.infer.reparam import NeuTraReparam
+
+
+logging.basicConfig(format='%(message)s', level=logging.INFO)
+
+
+class BananaShaped(dist.TorchDistribution):
+    support = constraints.real_vector
+
+    def __init__(self, a, b, rho=0.9):
+        self.a = a
+        self.b = b
+        self.rho = rho
+        self.mvn = dist.MultivariateNormal(torch.tensor([0., 0.]),
+                                           covariance_matrix=torch.tensor([[1., self.rho], [self.rho, 1.]]))
+        super().__init__(event_shape=(2,))
+
+    def sample(self, sample_shape=()):
+        u = self.mvn.sample(sample_shape)
+        u0, u1 = u[..., 0], u[..., 1]
+        a, b = self.a, self.b
+        x = a * u0
+        y = (u1 / a) + b * (u0 ** 2 + a ** 2)
+        return torch.stack([x, y], -1)
+
+    def log_prob(self, x):
+        x, y = x[..., 0], x[..., 1]
+        a, b = self.a, self.b
+        u0 = x / a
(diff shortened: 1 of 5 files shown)


examples/neutra.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def sample(self, sample_shape=()):
+        u = self.mvn.sample(sample_shape)
+        u0, u1 = u[..., 0], u[..., 1]
+        a, b = self.a, self.b
+        x = a * u0
+        y = (u1 / a) + b * (u0 ** 2 + a ** 2)
+        return torch.stack([x, y], -1)
+
+    def log_prob(self, x):
```
martinjankowiak: this is a properly normalized probability?

neerajprad: It should be since the determinant of the jacobian is 1.

pyro/infer/autoguide/guides.py
```diff
@@ -589,7 +591,9 @@ def get_posterior(self, *args, **kwargs):
         """
         Returns a diagonal Normal posterior distribution.
         """
-        return dist.Normal(self.loc, self.scale).to_event(1)
+        transform = dist.transforms.AffineTransform(self.loc, self.scale)
+        return dist.TransformedDistribution(dist.Normal(0., 1.).expand([self.latent_dim]).to_event(1),
+                                            transform)
```
fritzo: 1. Please measure memory and time overhead introduced by this refactoring. This is a heavily used autoguide, and should be kept simple and fast. I am even inclined to introduce a new method `get_posterior_transformed_dist()` or something so as not to obfuscate this code.

2. Please take care with device (the proposed change disregards non-default device):
    ```py
    base_dist = dist.Normal(torch.zeros_like(self.loc), torch.ones_like(self.scale))
    return dist.TransformedDistribution(base_dist.to_event(1), transform)
    ```

pyro/infer/autoguide/guides.py
```diff
@@ -545,7 +545,9 @@ def get_posterior(self, *args, **kwargs):
         """
         Returns a MultivariateNormal posterior distribution.
         """
-        return dist.MultivariateNormal(self.loc, scale_tril=self.scale_tril)
+        transform = dist.transforms.LowerCholeskyAffine(self.loc, scale_tril=self.scale_tril)
```
fritzo: Please measure memory and time overhead introduced by this refactoring. This is a heavily used autoguide, and should be kept simple and fast. I am even inclined to introduce a new method `.get_posterior_transformed_dist()` or something so as not to obfuscate this code.

fritzo:
> we require users to set `guide.requires_grad_(False)`

Interesting use case. From our 1.0 perspective, `PyroModule`s are the source of truth and the param store is only a legacy convenience for optimization-based inference. So it seems fine to support `nn.Module` methods like `.requires_grad_()`. I guess we could even add a method `.freeze()` or something that converted all `nn.Parameter`s and `PyroParam`s to registered buffers; this would improve performance for constrained parameters since `transform_to` would be called only once during `.freeze()` rather than each time parameters were accessed.

> Should we expose `transform` and `base_dist` ... as properties for these classes directly?

Good idea, I find this cleaner than the proposed refactoring of `.get_posterior()`. Maybe `.get_base_dist()` and `.get_transform()` which together create a `TransformedDistribution` that is equivalent to the simple `dist.Normal(self.loc, self.scale).to_event(1)` returned by `.get_posterior()`?

```py
dist.TransformedDistribution(..., dist.transforms.AffineTransform(...))
```
This is neat! I wonder if it's worth adding a multivariate de-centering reparameterizer, say a `LocScaleTrilReparam` analogous to the existing `LocScaleReparam`. Or we could implement this as a `TransformedDistributionReparam` that re-wrote `MultivariateNormal` as `TransformedDistribution(Normal(zeros,ones), ...)` as you have done in this PR, and maybe registered rewritings of other distributions as `TransformedDistribution` (we could then apply `TransformReparam` on the result).

pyro/infer/autoguide/guides.py
neerajprad: I think we should add `get_base_dist` and `get_transform` methods, and keep the existing `get_posterior` methods as is (we can add a default `get_posterior` though as you suggest in `AutoContinuous`).

fehiepsi: Agreed! We should make separate `get_base_dist` and `get_transform` methods.

pyro/infer/autoguide/guides.py
neerajprad: Reverted as discussed.

neerajprad:
> From our 1.0 perspective, PyroModules are the source of truth and the param store is only a legacy convenience for optimization-based inference. So it seems fine to support nn.Module methods like .requires_grad_().

Note that we already have `.requires_grad_` available to us via `nn.Module` which we are using in the examples and tests to freeze the guide parameters after training (though that will still run `transform_to` as you mention). @martinjankowiak and I were discussing if we should set `guide.requires_grad_(False)` by default inside `NeuTraReparam` rather than expect users to do so (current behavior). I think doing it inside `NeuTra` is fine too, except that it will have the side effect that the guide passed into `__init__` may have its parameters' `requires_grad` flag toggled to `False`.

> I guess we could even add a method .freeze() or something that converted all nn.Parameters and PyroParams to registered buffers

That will be a nice optimization. We could also modify the `nn.Module.train` and `nn.Module.eval` methods to freeze / unfreeze. I think we could potentially get bigger savings if we are able to jit compile neutra reparameterized models (#2292), but I'll keep this in mind since this is a more general change. 

> I wonder if it's worth adding a multivariate de-centering reparameterizer, say a LocScaleTrilReparam analogous to the existing LocScaleReparam. 

That will be nice, we have all the pieces in place already!

examples/neutra.py
```diff
@@ -0,0 +1,211 @@
...
+    fit_guide(guide, args)
+    with pyro.plate('N', args.num_samples):
+        guide_samples = guide()['x'].detach().cpu().numpy()
+
+    ax6.contourf(x1, x2, p, cmap='OrRd')
+    ax6.set(xlabel='x0', ylabel='x1', xlim=xlim, ylim=ylim,
+            title='Posterior \n(BNAF autoguide)')
+    sns.kdeplot(guide_samples[:, 0], guide_samples[:, 1], ax=ax6)
+
+    # 3(b). Draw samples using NeuTra HMC
```
fritzo: nit: `4(b)`?

pyro/infer/autoguide/guides.py
```diff
@@ -721,10 +721,10 @@ class AutoNormalizingFlow(AutoContinuous):
     """
 
     def __init__(self, model, init_transform_fn):
+        super().__init__(model, init_loc_fn=init_to_feasible)
```
fritzo: Do I understand correctly that this failed because `self.transform` was a `TransformModule` and thus required `nn.Module.__init__()` to be called before setting that attribute?

examples/neutra.py
neerajprad: Thanks for the catch, just fixed.

pyro/infer/autoguide/guides.py
neerajprad: `transform` is just set to None in `__init__` to begin with, but this was needed to call `PyroModule.__init__` before setting `self._prototype_tensor`.

neerajprad:
Thanks for the review and design discussion, @fritzo, @martinjankowiak!

PR #2299, opened 2020-02-04
Add pyro.contrib.timeseries.stable with tutorial

fritzo:
This adds a log-Stable Cox process in a new contrib module. The process is:
```
z[t] --> z[t+1]   stable
 |         |
 V         V
y[t]     y[t+1]   stable
 |         |
 V         V
x[t]     x[t+1]   poisson
```
but inference is tractable by combining
- an `AutoDiagonalNormal` guide over global parameters `q(globals)`
- a bunch of auxiliary variable reparameterizers `p(z,y)=p(z,y|aux)p(aux)`
- an `AutoDiagonalNormal` guide over the auxiliary variables `q(aux)`
- an EP guide for `q(y|x)`
- parallel-scan rsample for `p(y|x,aux,globals)` (sampled directly from the posterior)

## Tasks
- [x] ensure the model is registered as a sub-`nn.Module`
- [ ] work out conjugacy stuff with @neerajprad 
- [x] resolve nan grads wrt `stability` parameter
- [ ] demonstrate univariate forcasting on BART data
- [ ] demonstrate multivariate forcasting on BART data
- [ ] demonstrate event detection

## Tested
- [ ] smoke test (currently fails)
- [ ] validate inference on BART data


pyro/infer/reparam/conjugate.py
```diff
@@ -88,3 +88,34 @@ def __call__(self, name, fn, obs):
...
+        value = pyro.sample("{}_obs".format(name), fn)
+
+        # Expose the latent sample.
+        latent = getattr(value, "_pyro_latent", None)
+        if latent is not None:  # Try to use the latent cached by fn.sample().
+            pyro.deterministic("{}_latent".format(name), latent,
+                               event_dim=fn.event_dim + latent.dim() - value.dim())
+        else:  # Fall back to constucting and sampling from the posterior.
+            latent_fn = fn.get_posterior(value)
+            latent = pyro.sample("{}_latent".format(name), latent_fn)
```
fritzo: @neerajprad wdyt of this conjugacy pattern together with the `._pyro_latent` logic elsewhere in this PR? Note it's rather brittle, e.g. the [torch.stack() logic in Predictive](https://github.com/pyro-ppl/pyro/blob/6e967362ada8ca1661561826656d5cd6ad296ec4/pyro/infer/predictive.py#L47) would need to be applied *after* `ConjugateLatentParam`.

neerajprad: I just had a couple of questions to help me understand the changes involved:
 - where is the `get_posterior` method defined?
 - how does this relate to `ConjugateReparam` - are these complimentary approaches (i.e. this is only used for compound distributions) or do we use one for inference and the other for predictions / getting posterior samples?

fritzo: - I have not yet defined `.get_posterior()`, actually I'm still wavering between `.rsample_posterior()` which returns a value, and `.get_posterior()` which returns a distribution (like your `.posterior()` in pyro.contrib.conjugate)
- This PR renames `ConjugateReparam` to `ConjugateUpdateReparam`. The two should be complementary in that `ConjugateUpdateReparam` maps an already-factorized (prior,likelihood) pair of distributions and produces a posterior, whereas `ConjugateLatentReparam` takes a bundled prior+likelihood martinal distribution and pulls out the implicit latent variable. In this PR I am stacking z->y->x and using `ConjugateUpdateReparam` on `p(y|x)` and `ConjugateLatentReparam` on `p(z|y)`.

Note I feel these kludges all point towards a need for Funsors or a move towards Funsors.

pyro/infer/reparam/conjugate.py
```diff
@@ -88,3 +88,34 @@ def __call__(self, name, fn, obs):
...
+    variable of a conjugate pair distribution.
+
+    This only works for a small set of conjugate pair distributions, including
+    :class:`~pyro.distributions.BetaBinomial` ,
+    :class:`~pyro.distributions.DirichletMultinomial` ,
+    :class:`~pyro.distributions.GammaPoisson` , and
+    :class:`~pyro.distributions.GaussianHMM` .
+    """
+    def __call__(self, name, fn, obs):
+        assert obs is None, "ConjugateLatentReparam does not support observe statements"
```
neerajprad: If we don't support observed sites, how do we handle a model with a single observed from a compound distribution `pyro.sample('obs', BetaBinomial(c1, c0, count), obs=...)`?

fritzo: Good point, I think we can actually remove this using something like
```py
if obs is None:
    value = pyro.sample("{}_obs".format(name), fn)
else:
    value = obs
```

neerajprad: I see, I think we'll also need to convert the `latent` in the `else` block into a deterministic site in that case, and add the log prob contribution from the observed `value`.

pyro/infer/reparam/conjugate.py
fritzo: Interesting. I guess one way to implement the plated observation feature would be to define
```py
posterior = prior.conjugate_update(
    Likelihood.conjugate(data).conjugate_reduce(dim=...))
```
where
- `.conjugate_update()` is like `Funsor.__add__`,
- `.conjugate_reduce(dim)` is like `Funsor.reduce(ops.add, name)` in https://github.com/pytorch/pytorch/pull/32177, and
- `Likelihood.conjugate` is a static method of each class `Likelihood` (e.g. `Poisson`) that produces a corresponding `Gamma`
One issue I'm facing is that `.conjugate(data)` is not always a proper distribution even when `prior + L.conjugate(data)` is proper, e.g. `Poisson.conjugate(0) = Gamma(0, 1)`. I'm tempted to allow improper distributions like `Gamma(0,1)` as we would in Funsor, i.e.
```py
Gamma(0,1, value="z") = Poisson("z", value=0) + const.  # true in Funsor
```
and we'd probably disallow `.sample()` from improper distributions.

fehiepsi:
Just to correct the implementation of LogNormalCoxGuide (in case someone is using it):
```python
    def __call__(self, data):
        # See https://en.wikipedia.org/wiki/Gamma_distribution#Logarithmic_expectation_and_variance
        shape = data + 1
        loc = shape.digamma()
        scale = shape.polygamma(1).sqrt()
        return dist.Normal(loc, scale).to_event(self.event_dim)
```

fritzo:
@fehiepsi feel free to push directly to this branch and take over the PR. I won't have time to work on it for a while.

PR #2300, opened 2020-02-05
Add shape validation to LinearHMMReparam, GaussianHMM

fritzo:
Addresses #2299 

This enables partial validation of the `GaussianHMM` sample and of the `LinearHMMReparam`.

The motivation was a silent bug due when using `LinearHMMReparam` on a homogeneous `LinearHMM`, i.e. one whose duration is 1. By special convention our HMM distributions allow broadcasting of event_shape since time is part of the event shape and HMMs are often constructed as time-homogeneous systems.

This PR adds an attribute `._broadcast_time` so that `LinearHMMReparam` can disallow subsequent broadcasting of time; since that broadcasting would also need to apply to any time-local auxiliary variables created by `LinearHMMReparam`.

Note this is reminiscent of `Multinomial` whose `total_count` is ignored in `.log_prob()`.

## Tested
- added a death test

### DIFF
--- a/pyro/distributions/hmm.py
+++ b/pyro/distributions/hmm.py
@@ -361,13 +361,18 @@ class GaussianHMM(TorchDistribution):
         This should have event_shape ``(obs_dim,)``.
     :type observation_dist: ~torch.distributions.MultivariateNormal or
         ~torch.distributions.Independent of ~torch.distributions.Normal
+    :param bool broadcast_time: Whether to allow broadcasting of the time axis.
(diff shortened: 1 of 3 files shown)


fritzo:
@neerajprad I'm open to other ideas if you have suggestions. Maybe we could instead add a `duration` kwarg and set `event_shape` based on that?

fritzo:
@neerajprad I'll sketch up a `duration` solution for comparison...

fritzo:
Closing in favor of #2301

PR #2301, opened 2020-02-05
Add duration validation to HMMs

fritzo:
Addresses #2299 
Replaces #2300 

This adds a `duration` kwarg to Pyro's HMM classes and enables `._validate_sample()` on relevant methods. The crux is that HMMs can broadcast along the time axis, allowing cheap specification of homogeneous HMMs with time-unbroadcasted parameters.

This also fixes a duration broadcasting bug in `LinearHMMReparam` whereby unbroadcasted component distributions incorrectly led to unbroadcasted auxiliary variables.

@neerajprad this currently introduces minor backwards incompatibility, but I think this fixes more errors than it introduces. WDYT?
- `GaussianHMM.rsample()` and `LinearHMM.rsample()` will error when `event_shape[0] == 1` and `duration` is unspecified (which I think we want; this caused a bug in #2299 that took me a day to diagnose)

## Tested
- new validation logic is exercised by existing tests
- added a death test for `LinearHMMReparam`
- added a regression test for the `LinearHMMReparam` shape bug

### DIFF
--- a/pyro/distributions/hmm.py
+++ b/pyro/distributions/hmm.py
@@ -425,7 +425,9 @@ class GaussianHMM(HiddenMarkovModel):
         This should have event_shape ``(obs_dim,)``.
     :type observation_dist: ~torch.distributions.MultivariateNormal or
         ~torch.distributions.Independent of ~torch.distributions.Normal
-    :param int duration: Optional
+    :param int duration: Optional size of the time axis ``event_shape[0]``.
+        This is required when sampling from homogeneous HMMs whose parameters
+        are not expanded along the time axis.
     """
     has_rsample = True
     arg_constraints = {}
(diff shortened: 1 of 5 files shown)


pyro/distributions/hmm.py
```diff
@@ -188,7 +188,57 @@ def _sequential_gamma_gaussian_tensordot(gamma_gaussian):
...
+            event_shape = torch.Size((duration,) + event_shape[1:])
+        self._duration = duration
+        super().__init__(batch_shape, event_shape, validate_args)
+
+    @property
+    def duration(self):
+        """
+        Returns the size of the time axis, or None if unknown.
+        """
+        return self._duration
```
fritzo: The `.duration` attribute is modeled after `.batch_shape` and `.event_shape`

fritzo:
Thanks for reviewing! Ready to merge.

PR #2304, opened 2020-02-09
Add GaussianHMM.posterior_rsample, fix .rsample bugs

fritzo:
1. Adds a `GaussianHMM.posterior_rsample()` method
2. Fixes an expand bug in `GaussianHMM.rsample()`
3. Fixes an expand bug in `LinearHMM.rsample()`
4. Marks conjugacy-related methods EXPERIMENTAL (those introduced since last release)

## Tested
- [x] unit test `GaussianHMM.posterior_rsample()`
- [x] regression test for `GaussianHMM.rsample()`
- [x] regression test for `LinearHMM.rsample()`

### DIFF
--- a/pyro/distributions/hmm.py
+++ b/pyro/distributions/hmm.py
@@ -930,7 +930,8 @@ class LinearHMM(HiddenMarkovModel):
         init = self.initial_dist.rsample(sample_shape)
         trans = self.transition_dist.expand(self.batch_shape + (self.duration,)).rsample(sample_shape)
         obs = self.observation_dist.expand(self.batch_shape + (self.duration,)).rsample(sample_shape)
-        z = _linear_integrate(init, self.transition_matrix, trans)
+        trans_matrix = self.transition_matrix.expand(self.batch_shape + (self.duration, -1, -1))
+        z = _linear_integrate(init, trans_matrix, trans)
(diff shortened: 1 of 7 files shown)


fritzo:
@neerajprad as per today's discussion, I have marked new conjugacy-related methods EXPERIMENTAL. I'd like to use this PR's `.rsample_posterior()` in experiments until we have a funsor replacement.

pyro/distributions/hmm.py
```diff
@@ -506,14 +506,22 @@ def rsample(self, sample_shape=torch.Size()):
         obs_dim = self.obs_dim
         hidden_dim = self.hidden_dim
         trans = self._trans + self._obs.marginalize(right=self.obs_dim).event_pad(left=self.hidden_dim)
-        if trans.batch_shape[-1] != self.duration:
```
neerajprad: Is the issue here that `batch_shape` might be empty?

fritzo: That's one issue. Another issue is that `trans` is a `Gaussian` and has multiple underlying tensors some of which may be expanded and others unexpanded. Some of the ops after this line require all those underlying tensors to be expanded to the same shape. We could fix those underlying ops, but that seemed difficult so instead I've added manual `x = x.expand(x.batch_shape)` in a few places.

neerajprad: Thanks for explaining.

Issue #2214, opened 2019-12-06
[FR] New site type for pyro.deterministic


neerajprad:
Mentioned by @fritzo in https://github.com/pyro-ppl/pyro/pull/2196#issuecomment-560525534. Also related - https://github.com/pyro-ppl/pyro/pull/237.

Here is one reason to change it (which might not have been explicitly mentioned in that PR). I wanted to change our tutorials / examples to start using `pyro.deterministic` but the following behavior (under Predictive) will be unintuitive for users. It is obvious what is happening once you realize deterministic is an alias for a particular sample statement, and how broadcasting works (in the presence of event dims etc. which is probably too much to ask).

```python
    def forward(self, x, y=None):
        sigma = pyro.sample("sigma", dist.Uniform(0., 10.))
        mean = pyro.deterministic("mean", self.linear(x).squeeze(-1))
        with pyro.plate("data", x.shape[0]):
            obs = pyro.sample("obs", dist.Normal(mean, sigma), obs=y)
```

Here, the size of the mean variable will be `[1, len(x)]` under Predictive and not `[len(x)]` which one might expect. The reason is that `mean` is just a sample site under the hood and will prepend `1` to the left based on `max_plate_nesting` (in this case `1`, but since the remaining dims of mean are considered event dims we prepend 1).

PR #2306, opened 2020-02-13
Mark PyroModule non-experimental

fritzo:
This removes the "EXPERIMENTAL" marker from PyroModule, since we have not seen any issues in the last few months.

@neerajprad do you think `Predictive` is ready to graduate? No hurry.

In #2304 I've similarly marked some not-yet-released things EXPERIMENTAL.

### DIFF
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -2,10 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 """
-Pyro includes an experimental class :class:`~pyro.nn.module.PyroModule`, a
-subclass of :class:`torch.nn.Module`, whose attributes can be modified by Pyro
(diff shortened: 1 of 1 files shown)


neerajprad:
> @neerajprad do you think Predictive is ready to graduate?

I think the interface is stable, and we should be able to remove the experimental tag. The only thing that I want to address is behavior of deterministic sites (#2214).  @ahmadsalim, @fehiepsi - wdyt?

ahmadsalim:
@neerajprad Thanks for asking.

I think it would be nice that `pyro.deterministic` had a consistent shape under `Predictive`.

fehiepsi:
@neerajprad I am not sure how complicated it is in Pyro, but it would be nice to have. :)

PR #2308, opened 2020-02-13
Add tutorial for Boosting Black Box Variational Inference

lorenzkuhn:
In this pull request, we (@gideonite, @sharrison5, @TNU-yaoy and I) would like to submit a tutorial on implementing [Boosting Black Box Variational Inference](https://arxiv.org/pdf/1806.02185.pdf) using Pyro. In the tutorial, we summarise the paper, show how boosting black box Variational Inference can be implemented as an extension of Pyro's SVI and apply the method to the problem of approximating a bimodal posterior.

I've run `make test-tutorial`(which passes), `make scrub` and `make license`. The changes to `neutra.py` and `test_neutra.py` were caused by `make license`.

We look forward to your feedback!


fritzo:
@lorenzkuhn this looks great!  LGTM on code and documentation.

@martinjankowiak can you review the relbo math?

martinjankowiak:
thanks @lorenzkuhn looks great!

here are some nits/comments/suggestions:
- the "Variational Inference/ELBO" section is inconsistent about s(z)/q(z); i.e. s => q
- typo 'mulitmodal'
- "in our case, we include loc in both the model and the guide." do you mean z?
- note that the way you use `Trace_ELBO` leads to extra samples being drawn. either comment or fix?

lorenzkuhn:
Thanks a lot for the feedback @fritzo and @martinjankowiak, that sounds good. 

@martinjankowiak : Thanks, I'll fix the issues from the first three bullet points. What would be the recommended approach to avoid drawing extra samples in Trace_ELBO?

martinjankowiak:
@lorenzkuhn i think the easiest thing to do would be to replace

```
loss_fn = pyro.infer.Trace_ELBO(max_plate_nesting=1).differentiable_loss(model,
                                                                             guide,
                                                                             *args,
                                                                             **kwargs)
```
with a simple custom elbo along the lines of what's [here](https://github.com/pyro-ppl/pyro/blob/c9249333def232d8273ff6e88402fe6afd477b26/pyro/contrib/minipyro.py#L319), i.e. this way you can reuse your guide trace.
@fritzo wdyt?

fritzo:
As I understand, @martinjankowiak is concerned with the `guide` being run twice in `relbo()`. I believe this could be avoided by passing a traced guide to `elbo` and using that trace to compute the approximation. WDYT of the following refactoring?
```py
def relbo(model, guide, *args, **kwargs):
    approximation = kwargs.pop('approximation')

    # We first compute the elbo, but record a guide trace for use below.
    traced_guide = trace(guide)
    elbo = pyro.infer.Trace_ELBO(max_plate_nesting=1)
    loss_fn = elbo.differentiable_loss(model, traced_guide, *args, **kwargs)

    # We do not want to update parameters of previously fitted components
    # and thus block all parameters in the approximation apart from z.
    guide_trace = traced_guide.trace
    replayed_approximation = trace(replay(block(approximation, expose=['z']), guide_trace))
    approximation_trace = replayed_approximation.get_trace(*args, **kwargs)

    relbo = -loss_fn - approximation_trace.log_prob_sum()
    
    # By convention, the negative (R)ELBO is returned.
    return -relbo
```

fritzo:
@lorenzkuhn FYI we're planning on a 1.3 release on Monday. There's no rush 😄but if you would like to get this in by Monday morning, we can publish it with the release.

lorenzkuhn:
Ah, I see, thanks @martinjankowiak and @fritzo. I've refactored the relbo as you suggested @fritzo    and fixed the typos/inconsistencies that @martinjankowiak found :thumbsup:

PR #2307, opened 2020-02-13
Add an experimental contrib.funsor library for staging the transition to Funsor

eb8680:
This PR contains a first draft of the machinery necessary to port Pyro's inference algorithms to Funsor while preserving backward compatibility with model code, as originally proposed in our design doc.  I'm posting it in an incomplete state for design discussion.

The key component is a pair of effectful wrappers for the funsor utility functions `to_funsor` and `to_data` (see https://github.com/pyro-ppl/funsor/pull/316) that use the information contained in Pyro's batch dimensions and `pyro.markov` contexts to automatically convert positional tensors to funsors with meaningfully named inputs and vice versa, and a `NamedMessenger` that overrides the behavior of `to_funsor` and `to_data` and should be a drop-in replacement for `pyro.markov` so that we can continue using old positional dimension style in existing model code while working entirely with named Funsor dimensions in inference code.  

The implementation details of `NamedMessenger` aren't that important and will need to change quite a bit to acommodate plates and global dimensions; the current version is more of an existence proof that this overall design centered around `to_funsor` and `to_data` is feasible and can preserve backwards compatibility with models.  This design is also backend-agnostic and should be immediately reusable in NumPyro.

To illustrate how this would dramatically simplify our inference code, I've included a sketch of Funsor-based enumeration and variable elimination in `examples/contrib/funsor/enum_sketch.py`.

Plan for integrating the changes here into Pyro:
1. Finish a working version of the `contrib.funsor` library
2. Break it up into small pieces and merge each piece into `pyro.contrib.funsor` (see below)
3. Make any necessary changes/bugfixes upstream in Funsor and cut a Funsor release
4. Add `funsor[torch]` as an optional requirement
5. Cut a Pyro release with the `contrib.funsor` package
6. Profile extensively
6. Begin replacing components in Pyro proper with the versions in `contrib.funsor`
7. Delete or deprecate unused code resulting from the refactor

Tentative sequence of PRs expected based on the contents of this branch:
1. `make_handler` change, and initial `pyro.contrib.funsor` library creation: `primitives`, `to_funsor`, `to_data`, `markov`, `pyroapi`, and `test_named.py`.
3. First introductory tutorial
4. `GlobalNamedMessenger`, `BaseEnumMessenger`, and a second tutorial
5. Exact-only `EnumMessenger`, `trace`, `replay`, and `test_valid_models.py`
6. Vectorized `PlateMessenger` and `test_valid_models_plate.py`
[...]

PR #2309, opened 2020-02-15
High-level forecasting interface: contrib.forecast

fritzo:
Addresses #2311
See [design doc](https://docs.google.com/document/d/1jqQ_7mCnggqqmICzux2Fr9yvHO5AoVpBdmjNC0i6JAY).
See [generated documentation](http://fritzo.org/temp/pyro/html/contrib.forecast.html).

This is the first minimal coherent PR of a forecasting framework. Users specify a model by subclassing the `ForecatingModel` and implementing a single `.model()` method (following @eb8680's suggestion and differing from the design doc), which can contain plates and reparameterizers. The final step in the `.model()` method is to call the `self.predict()` method, possibly inside plates indicating hierarchical organization. For example:
```py
class MyModel(ForecastingModel):
    def model(self, zero_data, covariates):
        loc = zero_data[..., :1, :]
        scale = pyro.sample("scale", dist.LogNormal(loc, 1).to_event(1))
        with self.time_plate:
            jumps = pyro.sample("jumps", dist.Normal(0, scale).to_event(1))
        prediction = jumps.cumsum(-2)
        noise_dist = dist.Laplace(zero_data, 1)
        self.predict(noise_dist, prediction)

forecaster = Forecaster(MyModel(), data, covariates)
samples = forecaster(data, covariates, num_samples=100)
```
There is considerable complexity hidden behind the `.predict()` method, enabling the `.model()` method to be used for both training and forecasting, and freeing users from error-prone shape munging. I not so worried about this complexity because it is a side effect of positional dimensions and should be eliminated after @eb8680's big Funsor refactor #2307.

## Tested
- [x] `GaussianHMM.prefix_condition()`
- [x] univariate `prefix_condition()`
- [x] `Forecaster.fit()`
- [x] `Forecaster()` drawing samples


docs/source/primitives.rst
```diff
@@ -12,6 +12,8 @@ Primitives
     :members:
     :undoc-members:
 
+.. autofunction:: pyro.plate_stack
+
```
fritzo: unrelated cleanup

examples/neutra.py
```diff
@@ -1,3 +1,6 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
```
fritzo: unrelated cleanup

pyro/contrib/forecast/forecaster.py
```diff
@@ -0,0 +1,221 @@
...
+
+    @abstractmethod
+    def model(self, zero_data, covariates):
+        """
+        Generative model definition.
+
+        Implementations must call the :meth:`predict` method exactly once.
+
+        Implementations must draw all time-dependent noise inside the
+        :meth:`time_plate` .  The prediction passed to :meth:`predict` must be
```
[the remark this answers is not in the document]
fritzo: Are you referring to the space between the tick and the period? I always include this because sphinx [restrictions on quoting](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#gotchas). I don't fully understand the rules, but after having spent hours debugging subtle sphinx breakages, I now always surround tick links with space. Do you have a clearer understanding of sphinx restrictions?

pyro/contrib/forecast/util.py
```diff
@@ -0,0 +1,228 @@
...
+        assert model_value.shape[:-2] == guide_value.shape[:-2]
+        assert model_value.size(-2) > guide_value.size(-2)
+        assert model_value.size(-1) == guide_value.size(-1)
+        split = guide_value.size(-2)
+        msg["value"] = torch.cat([guide_value, model_value[..., split:, :]], dim=-2)
+
+
+class PrefixConditionMessenger(Messenger):
+    """
+    EXPERIMENTAL Given a prefix of t-many observations, condition a (t+f)-long
```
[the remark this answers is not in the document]
fritzo: Actually this messenger is indeed conditioning on observations, but almost all of our noise models are independent so conditioning reduces to truncating. The one nontrivial case is `GaussianHMM` where `prefix_condition()` triggers the `GaussianHMM.filter()` method and does actual conditioning.

fritzo:
@neerajprad Thanks for reviewing! Note that pyro/contrib/forecast/util.py is intended for internal use only. I have omitted underscores because I am importing across files. I can additionally mark INTERNAL if you think that's safer.

pyro/contrib/forecast/forecaster.py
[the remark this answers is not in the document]
neerajprad: That and the double space before `The` - my keyboard has a double keystroke problem with many keys, which turns out to be a common problem with recent macbooks. I thought you might have the same issue with yours. I didn't know about the sphinx linking issue, I have faced it before and didn't realize that I could get around by putting a space after the backtick, I'll keep that in mind for next time.

pyro/contrib/forecast/util.py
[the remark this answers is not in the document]
neerajprad: Ahh..I think I understand this now. Thanks for clarifying!

neerajprad:
> I have omitted underscores because I am importing across files. I can additionally mark INTERNAL if you think that's safer.

I think since it is marked as experimental, it should be fine. It is nice that it is documented and if we have users who end up using it, we can get to know about other use cases and patterns.

pyro/contrib/forecast/util.py
```diff
@@ -0,0 +1,228 @@
...
+
+import pyro.distributions as dist
+from pyro.poutine.messenger import Messenger
+from pyro.poutine.util import site_is_subsample
+
+
+class PrefixReplayMessenger(Messenger):
+    """
+    EXPERIMENTAL Given a trace of training data, replay a model with batched
+    sites extended to include both training and forecast time, using the guide
```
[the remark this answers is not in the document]
fritzo: Oops good catch. This should not assume dim=-2. I think this will need a two part solution:
1. inside the `time_plate` set `dim=-1 - msg["fn"].event_dim`
2. outside the `time_plate` we still need to support [reparametrizer logic](https://github.com/pyro-ppl/pyro/blob/ce76668176b6df2d46f5d3276a232b913740c0d8/pyro/infer/reparam/hmm.py#L80), so we can search to find which dim differs.

I'll push a fix soon.

pyro/contrib/forecast/forecaster.py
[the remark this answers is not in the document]
fritzo: Well I've removed the spaces and the docs built fine 🤷‍♂

pyro/contrib/forecast/util.py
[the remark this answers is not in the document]
fritzo: OK, I've changed to simply search for whatever dimension differs, ignoring the `time_plate`. I've also added a new test case where `dim=-` that would have broken before your comment.

pyro/contrib/forecast/forecaster.py
```diff
@@ -0,0 +1,220 @@
...
+    def time_plate(self):
+        """
+        :returns: A plate named "time" with size ``covariates.size(-2)`` and
+            ``dim=-1``. This is available only during model execution.
+        :rtype: :class:`~pyro.plate`
+        """
+        assert self._time_plate is not None, ".time_plate accessed outside of .model()"
+        return self._time_plate
+
+    def predict(self, noise_dist, prediction):
```
martinjankowiak: should there be a comment that this is analogous to an observe statement in a typical pyro model?

pyro/contrib/forecast/forecaster.py
```diff
@@ -0,0 +1,220 @@
...
+        noise_dist = reshape_batch(noise_dist, noise_dist.batch_shape + (1,))
+        data = self._data.unsqueeze(-3)
+        prediction = prediction.unsqueeze(-3)
+
+        # Create a sample site.
+        t_obs = data.size(-2)
+        t_cov = prediction.size(-2)
+        if t_obs == t_cov:  # training
+            pyro.sample("residual", noise_dist, obs=data - prediction)
+            self._forecast = data.new_zeros(data.shape[:-2] + (0,) + data.shape[-1:])
```
martinjankowiak: why fill this during training?

pyro/contrib/forecast/forecaster.py
```diff
@@ -0,0 +1,220 @@
...
+                 num_steps=1001,
+                 log_every=100,
+                 init_scale=0.1):
+        assert data.size(-2) == covariates.size(-2)
+        super().__init__()
+        self.model = model
+        self.guide = AutoNormal(self.model, init_loc_fn=init_to_sample, init_scale=init_scale)
+        optim = ClippedAdam({"lr": learning_rate, "betas": betas,
+                             "lrd": learning_rate_decay ** (1 / num_steps)})
+        elbo = Trace_ELBO()
```
martinjankowiak: support `vectorize_particles=True`?

pyro/contrib/forecast/util.py
```diff
@@ -0,0 +1,231 @@
...
+
+for _type in UNIVARIATE_DISTS:
+    prefix_condition.register(_type)(_prefix_condition_univariate)
+
+
+@prefix_condition.register(dist.MultivariateNormal)
+def _(d, data):
+    t = data.size(-2)
+    loc = d.loc[..., t:, :]
+    scale_tril = d.scale_tril[..., t:, :, :]
```
martinjankowiak: this sort of thing only works because everything has been (possibly) expanded by now right?

pyro/distributions/hmm.py
```diff
@@ -596,6 +596,55 @@ def conjugate_update(self, other):
         new._validate_args = self.__dict__.get('_validate_args')
         return new, log_normalizer
 
+    def prefix_condition(self, data):
+        """
+        EXPERIMENTAL Given self has ``event_shape == (t+f, d)`` and data ``x``
+        of shape ``batch_shape + (t, d)``, find a conditional distribution of
```
martinjankowiak: nit: find => compute?

martinjankowiak: nit: find => compute/construct?

tests/contrib/forecast/test_forecaster.py
```diff
@@ -0,0 +1,119 @@
...
+    def model(self, zero_data, covariates):
+        with pyro.plate_stack("batch", zero_data.shape[:-2], rightmost_dim=-2):
+            loc = zero_data[..., :1, 0]
+            scale = pyro.sample("scale", dist.LogNormal(loc, 1))
+
+            with self.time_plate:
+                jumps = pyro.sample("jumps", dist.Normal(0, scale))
+            prediction = jumps.cumsum(-1).unsqueeze(-1) + zero_data
+
+            noise_dist = dist.Laplace(zero_data, 1).to_event(2)
```
martinjankowiak: i guess this `to_event` logic will be tricky for some users...

pyro/contrib/forecast/forecaster.py
fritzo: It seems safer to keep a consistent return type and contract that
```py
assert isinstance(result, torch.Tensor)
assert result.dtype == data.dtype
assert result.device == data.device
assert result.shape == data.shape[:-1] + covariates.shape[-2:-1] + data.shape[-1:]
```
Note this costs nothing; when training the tensor is empty `self._forecast.numel() == 0`.

pyro/contrib/forecast/util.py
fritzo: That's right, all inputs are expanded via `broadcast_all()`. Note that we could be more efficient using Du's `._unbroadcasted_scale_tril` but that optimization seems premature at this time.

tests/contrib/forecast/test_forecaster.py
fritzo: Good point, I can add an adapter inside the `.predict()` method
```py
noise_dist = noise_dist.to_event(2 - niose_dist.event_dim)
```
I'll update shortly...

pyro/contrib/forecast/forecaster.py
fritzo: done

pyro/distributions/hmm.py
fritzo: done

tests/contrib/forecast/test_forecaster.py
fritzo: done, and updated tests

pyro/contrib/forecast/forecaster.py
fritzo: done

pyro/distributions/hmm.py
fritzo: done

pyro/contrib/forecast/forecaster.py
```diff
@@ -182,14 +196,17 @@ def __init__(self, model, data, covariates, *,
                  learning_rate_decay=0.1,
                  num_steps=1001,
                  log_every=100,
-                 init_scale=0.1):
+                 init_scale=0.1,
+                 num_particles=1,
```
martinjankowiak: do we need docstrings for num_particles, learning_rate_decay, etc.?

tests/contrib/forecast/test_forecaster.py
martinjankowiak: cool great this automation works

pyro/contrib/forecast/forecaster.py
fritzo: Do you mind if I add those in a future PR to avoid blocking follow-up PRs?

Issue #2311, opened 2020-02-16
Add a contrib.forecasting module


fritzo:
Related to #2090 but focused on a narrow class of models, inference methods, and forecasting queries. See [design doc](https://docs.google.com/document/d/1jqQ_7mCnggqqmICzux2Fr9yvHO5AoVpBdmjNC0i6JAY).

- [x] #2309 implement a `GaussianHMM.prefix_condition()` method
- [x] #2309 implement a `PrefixReplayMessenger`
- [x] #2309 create a `ForecastingModel` base class and `Forecaster` wrapper
- [x] #2309 support Stable, StudentT, LinearHMM reparameterization
- [x] #2312 expose `.get_dist()` method of contrib.timeseries models
- [x] #2313 add seasonality helpers
- [x] #2335 add more seasonality helpers
- [x] #2316 add an eval/backtesting module
- [x] #2318 add a simple example
- [x] #2320 support `DiscreteCosineReparam` over `time_plate`, or support DCT in the guide
- [x] #2323 support subsampling in `AutoNormal`
- [x] #2330 support warm-starting in `backtest()`
- [x] #2331 support more likelihood distributions
- [x] #2332 add duration kwargs to `GaussianHMM`s in contrib.timeseries
- [x] #2334 support independent `GaussianHMM`s via `IndependentHMM` wrapper
- [x] #2339 add a univariate tutorial
- [x] #2343 add a state space model tutorial
- [x] #2348 add a multivariate tutorial
- [x] initial release

PR #2313, opened 2020-02-19
Add periodic ops for seasonal forecasting

fritzo:
Addresses #2311 

This adds two helpers for defining fine-grained seasonal components in forecasting models. While these helpers are linear dynamical systems and could in principle be encoded in `GaussianHMM` matrices, they would lead to prohibitively high-dimensional latent state, e.g. 168 dimensions for an hour-of-week model. By instead keeping this part of the model variational, we can retain O(1) cost per time step, independent of period.

Example using #2309 :
```py
def model(self, zero_data, covariates):
    period = 24
    duration = zero_data.size(-2)

    # This time-independent noise will be variationally inferred.
    with pyro.plate("init_season_plate", period,  dim=-1):
        init_season = pyro.sample("init_season", dist.Normal(0, 1))
    with time_plate:
        noise_season = pyro.sample("noise_season", dist.Normal(0, 0.01))

    # This is cheap and deterministic:
    season = (periodic_repeat(init_season, duration, dim=-1) +
              periodic_cumsum(noise_season, period, dim=-1))

    prediction = season.unsqueeze(-1)
    noise_dist = ...
    self.predict(noise_dist, prediction)
```
I plan to use these in a future `contrib.forecast` tutorial.

## Tested
- [x] unit test for `periodic_repeat()`
- [x] unit test for `periodic_cumsum()`

### DIFF
--- a/pyro/ops/tensor_utils.py
+++ b/pyro/ops/tensor_utils.py
@@ -61,7 +61,7 @@ def periodic_repeat(tensor, size, dim):
     """
     assert isinstance(size, int) and size >= 0
     assert isinstance(dim, int)
-    if dim > 0:
+    if dim >= 0:
         dim -= tensor.dim()
 
     period = tensor.size(dim)
@@ -89,7 +89,7 @@ def periodic_cumsum(tensor, period, dim):
     """
     assert isinstance(period, int) and period > 0
     assert isinstance(dim, int)
-    if dim > 0:
+    if dim >= 0:
         dim -= tensor.dim()
 
     # Pad to even size.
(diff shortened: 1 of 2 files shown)


pyro/ops/tensor_utils.py
```diff
@@ -39,6 +39,76 @@ def block_diagonal(mat, block_size):
...
+
+    This is useful for computing drifting seasonality in time series models.
+
+    :param torch.Tensor tensor: A tensor of differences.
+    :param int period: The period of repetition.
+    :param int dim: The tensor dimension along which to accumulate.
+    """
+    assert isinstance(period, int) and period > 0
+    assert isinstance(dim, int)
+    if dim > 0:
```
neerajprad: Should this be `dim >= 0`, otherwise for dim = 0, I think the padding might be incorrect?

fritzo: Good catch, fixed.

PR #2316, opened 2020-02-19
Add evaluation module to contrib.forecast

fritzo:
Addresses #2311 
See [generated docs](http://fritzo.org/temp/pyro/html/contrib.forecast.html#module-pyro.contrib.forecast.evaluate).

This adds a simple evaluation module for forecasting. This includes:
- three metrics (MAE, RMSE, CRPS) which are appropriate for evaluating (median, mean, confidence intervals) respectively.
- a backtesting framework that trains and evaluates a model multiple times on a moving window of data.

I intend this module to stay pretty simple, where we defer more complex evaluation to mature frameworks like [gluonts.evaluation](https://gluon-ts.mxnet.io/api/gluonts/gluonts.evaluation.html). I plan to use this in a future forecasting tutorial.

This also adds documentation requested by @martinjankowiak in #2309.

## Review requests

@neerajprad could you please review logic and math?

@martinjankowiak could you please ensure the interface is useful for your experiments?

## Tested
- [x] shape tests without `transform`
- [x] shape tests with `transform`


docs/source/infer.autoguide.rst
```diff
@@ -27,6 +27,14 @@ AutoCallable
...
     :show-inheritance:
 
+AutoNormal
+----------
+.. autoclass:: pyro.infer.autoguide.AutoNormal
+    :members:
+    :undoc-members:
+    :special-members: __call__
+    :show-inheritance:
+
```
fritzo: Adds link target that was missed in #2050

pyro/contrib/forecast/forecaster.py
```diff
@@ -182,13 +182,30 @@ class Forecaster(nn.Module):
...
+        :class:`~pyro.optim.optim.ClippedAdam`. Note this is the total decay
+        over all ``num_steps``, not the per-step decay factor.
+    :param int num_steps: Number of :class:`~pyro.infer.svi.SVI` steps.
+    :param float init_scale: Initial uncertainty scale of the
+        :class:`~pyro.infer.autoguide.AutoNormal` guide.
+    :param int num_particles: Number of particles used to compute the
+        :class:`~pyro.infer.elbo.ELBO`.
+    :param bool vectorize_particles: If ``num_particles > 1``, determines
+        whether to vectorize computation of the :class:`~pyro.infer.elbo.ELBO`.
+        Defaults to True. Set to False for models with dynamic control flow.
```
fritzo: As requested by @martinjankowiak in #2309

pyro/contrib/forecast/evaluate.py
```diff
@@ -0,0 +1,163 @@
...
+@torch.no_grad()
+def eval_mae(pred, truth):
+    """
+    Evaluate mean absolute error, using sample median as point estimate.
+
+    :param torch.Tensor pred: Forecasted samples.
+    :param torch.Tensor truth: Ground truth.
+    :rtype: float
+    """
+    pred = pred.median(0).values
```
neerajprad: Should we also support MAE and RMSE for the full set of predictions (not just the median/mean), or do you think CRPS is better reserved for that?

pyro/contrib/forecast/evaluate.py
```diff
@@ -0,0 +1,163 @@
...
+
+    result = {}
+    for t1 in range(start, stop, stride):
+        t0 = 0 if train_window is None else t1 - train_window
+        t2 = duration if test_window is None else t1 + test_window
+        assert 0 <= t0 < t1 < t2 <= duration
+        logger.info("Evaluating on window ({}, {}, {})".format(t0, t1, t2))
+
+        # Train a forecaster on the training window.
+        pyro.set_rng_seed(seed)
```
neerajprad: Is it okay to reuse the seed for each window? We can move it outside but I guess the `seed` argument below won't be particularly meaningful in that case.

pyro/contrib/forecast/evaluate.py
```diff
@@ -0,0 +1,163 @@
...
+    assert data.size(-2) == covariates.size(-2)
+    assert isinstance(min_train_window, int) and min_train_window >= 1
+    assert isinstance(min_test_window, int) and min_test_window >= 1
+    if metrics is None:
+        metrics = DEFAULT_METRICS
+    assert metrics, "no metrics specified"
+
+    duration = data.size(-2)
+    if test_window is None:
+        stop = duration - min_test_window
```
neerajprad: Should stop be `duration - min_test_window + 1` given that upper limit is exclusive?

pyro/contrib/forecast/evaluate.py
fritzo: No, MAE and RMSE don't make sense for monte carlo samples. CRPS is a good metric for monte carlo samples.

fritzo: In particular, mean can be seen as a minimizer of RMSE, median as a minimizer of MAE, and correctly-distributed samples as a minimizer of CRPS. Each pair is tightly coupled.

pyro/contrib/forecast/evaluate.py
fritzo: I think it is safer to reset the seed for each window, that way we can e.g. compute with large stride, then fill in the dict with smaller stride later. You're right that "seed" is then duplicated in all records. I'll switch to adding seed as an index column, that way we can test with multiple seeds and eventually store in a single dict.

pyro/contrib/forecast/evaluate.py
fritzo: Good catch, fixed.

pyro/contrib/forecast/evaluate.py
fritzo: OK, I've switched to a list output format. That way we can run with multiple random seeds and aggregate multiple experiments.

pyro/contrib/forecast/evaluate.py
neerajprad: Makes sense.

pyro/contrib/forecast/evaluate.py
```diff
@@ -0,0 +1,170 @@
...
+        start = min_train_window
+    else:
+        start = train_window
+
+    results = []
+    for t1 in range(start, stop, stride):
+        t0 = 0 if train_window is None else t1 - train_window
+        t2 = duration if test_window is None else t1 + test_window
+        assert 0 <= t0 < t1 < t2 <= duration
+        logger.info("Evaluating on window ({}, {}, {})".format(t0, t1, t2))
```
martinjankowiak: maybe it'd be more legible if this used slice syntax? `[3:9], [9:12]`

fritzo: done

fritzo:
Thanks for reviewing @neerajprad and @martinjankowiak !

PR #2318, opened 2020-02-20
Add a simple contrib.forecasting example

fritzo:
Addresses #2311 

This adds a simple forecasting example testing the `ForecastingModel` and `backtest()` function on BART data (arrivals to and departures from Embarcadero from 2011-2019). I've kept this minimal and simple, and plan to later create an extended tutorial as a jupyter notebook.

## Tested
- [x] added to test_examples.py


pyro/contrib/forecast/forecaster.py
```diff
@@ -11,7 +11,7 @@
 import pyro.distributions as dist
 import pyro.poutine as poutine
 from pyro.infer import SVI, Trace_ELBO
-from pyro.infer.autoguide import AutoNormal, init_to_sample
+from pyro.infer.autoguide import AutoDiagonalNormal, init_to_sample
```
fritzo: This works around a bug in `AutoNormal` triggered by the LKJ distribution. I'll fix that bug later...

examples/contrib/forecast/bart.py
```diff
@@ -0,0 +1,149 @@
...
+                                  dist.LogNormal(torch.zeros(2), 0.1).to_event(1))
+        trans_corr = pyro.sample("trans_corr",
+                                 dist.LKJCorrCholesky(2, torch.ones(())))
+        trans_scale_tril = trans_scale.unsqueeze(-1) * trans_corr
+
+        obs_scale = pyro.sample("obs_scale",
+                                dist.LogNormal(torch.zeros(2), 0.1).to_event(1))
+        obs_corr = pyro.sample("obs_corr",
+                               dist.LKJCorrCholesky(2, torch.ones(())))
+        obs_scale_tril = obs_scale.unsqueeze(-1) * obs_corr
```
fritzo: @fehiepsi did I get this linear algebra correct, for sampling a `scale_tril` matrix?

martinjankowiak: would it be helpful to pepper this with a bunch of assert shape statements? (it would be to me anyway)

examples/contrib/forecast/bart.py
```diff
@@ -0,0 +1,149 @@
...
+        # linear functions to transform our diagonal Normal noise to nontrivial
+        # samples from a Gaussian process.
+        prediction = (periodic_repeat(season_init, duration, dim=-2) +
+                      periodic_cumsum(season_noise, period, dim=-2))
+
+        # Construct a joint noise model. This model is a GaussianHMM, whose
+        # .rsample() and .log_prob() methods are parallelized over time; this
+        # this entire model is parallelized over time.
+        init_dist = dist.Normal(torch.zeros(2), 100).to_event(1)
+        trans_mat = trans_timescale.neg().exp().diag_embed()
```
martinjankowiak: ou process?

examples/contrib/forecast/bart.py
```diff
@@ -0,0 +1,149 @@
...
+    # the same size and dtype as the real data (but of course the generative
+    # model shouldn't depend on the value of the data it generates!), and a
+    # tensor of covariates. Our simple model depends on no covariates, so we
+    # simply pass in an empty tensor (see  the preprocess() function above).
+    def model(self, zero_data, covariates):
+        duration = zero_data.size(-2)
+        period = 24 * 7
+
+        # Sample global parameters.
+        noise_scale = pyro.sample("noise_scale",
```
martinjankowiak: this suggests to me that it'd be nice if the default guide could also take a MAP arg that demotes random variables to have dirac delta variational distributions

fritzo: I've intentionally avoided `AutoDelta` and MAP guides because that don't play well with reparameterization. In particular I believe they are incorrect / weird for `Stable` reparameterizers, and I'd like to avoid silent errors in model specification, esp. for a distribution people already distrust.

examples/contrib/forecast/bart.py
fritzo: Yes, this is an OU process around the seasonal GP

examples/contrib/forecast/bart.py
martinjankowiak: is this even a problem for the globals? shouldn't they be fine?

fritzo: The problem I'm concerned with doesn't depend on local-vs-global; it is due to `AutoDelta(model)` behaving differently from `AutoDelta(poutine.reparam(model, ...))` even for trivial single-site models.

But maybe I misunderstand your concern. Do you simply want to be non-variational about some variables? If so you can use `pyro.param` rather than `pyro.sample` in the model. In case of globals I doubt the prior matters much anyway. I'm only avoiding `pyro.param` here as an academic exercise to demonstrate we can be fully Bayesian.

martinjankowiak: just trying to understand the design constraints---thanks for explaining. 

my question is motivated by the intuition that i'd generally expect being bayesian about noise scale parameters (for example ) using VI could be problematic. so in practice the probably more sensible alternatives are MLE (`param`) and MAP (`AutoDelta`)

examples/contrib/forecast/bart.py
fritzo: done.

PR #2320, opened 2020-02-21
Implement a forecasting-specific optimizer

fritzo:
Addresses #2311 

This implements an `DCTAdam` optimizer similar to `ClippedAdam` but also maintaining `Adam` params in the frequency domain.

The motivation is that, while DCT reparameterizers appear to work well for time series models, (1) the `DiscreteCosineReparam` is not compatible with other `ForecastingModel` poutines, and (2) it is difficult to know whether each random variable will have an approximately time-diagonal or frequency-diagonal posterior. This PR address (1) by moving DCT logic into an optimizer, and addresses (2) by using both time- and frequency- domain optimizers for each parameter.

This also:
- fixes a bug in `AutoNormal`, exercised by examples/contrib/forecast/bart.py
- fixes the `backtest()` interface to input `model_fn` rather than `model`, so that parameter state is not leaked across training window

## Results

Results on bart.py aren't great: learning seems actually 20% slower. Thus I've disabled this feature by default. I'd like to merge EXPERIMENTAL and use this on a number of datasets. If it doesn't seem to help on other datasets we can remove it.

## Tests
- [x] unit tests for `dct()`,`idct()` dim kwarg
- [x] unit tests for `next_fast_len()`
- [x] unit tests for optimizer
- [x] used in examples/contrib/forecast/bart.py


pyro/contrib/forecast/guides.py
```diff
@@ -0,0 +1,139 @@
...
+                        f_scale = _deep_getattr(self.f_scales, name)
+                        f = pyro.sample(name + "_frequency",
+                                        dist.Normal(f_loc, f_scale).to_event(event_dim),
+                                        infer={"is_auxiliary": True})
+                    loc = loc + idct(f, dim=-1 - event_dim)
+                    stack.enter_context(plates["time"])
+
+                # Sample from a transformed diagonal normal.
+                base_dist = dist.Normal(loc, scale).to_event(event_dim)
+                pyro.sample(name, dist.TransformedDistribution(base_dist, transform))
```
fritzo: @fehiepsi @martinjankowiak can you help me convince myself that this auxiliary variable method is correct?

fritzo:
@fehiepsi could you review the autoguide logic and the new dct functionality?

pyro/optim/dct_adam.py
```diff
@@ -0,0 +1,113 @@
...
+    return torch.cat([time_domain, freq_domain], dim=dim)
+
+
+def _transform_inverse(x, dim, duration):
+    assert not x.requires_grad
+    assert dim < 0
+    dots = (slice(None),) * (x.dim() + dim)
+    left = dots + (slice(None, duration),)
+    right = dots + (slice(duration, None),)
+    return idct(x[right], dim)[left].add_(x[left])
```
fehiepsi: Makes sense to me, given that dct is a linear transform. Nice observation!

pyro/infer/autoguide/guides.py
```diff
@@ -386,24 +386,24 @@ def __init__(self, model, init_loc_fn=init_to_feasible, init_scale=0.1):
...
+            self._event_dims[name] = event_dim
 
             # Collect independence contexts.
             self._cond_indep_stacks[name] = site["cond_indep_stack"]
             init_scale = torch.full_like(init_loc, self._init_scale)
 
-            _deep_setattr(self.locs, name, nn.Parameter(init_loc))
-            _deep_setattr(self.scales, name, PyroParam(init_scale, constraints.positive))
+            _deep_setattr(self.locs, name, PyroParam(init_loc, constraints.real, event_dim))
+            _deep_setattr(self.scales, name, PyroParam(init_scale, constraints.positive, event_dim))
```
fehiepsi: This looks correct to me.

pyro/ops/tensor_utils.py
```diff
@@ -109,6 +109,37 @@ def periodic_cumsum(tensor, period, dim):
     return result
 
 
+_NEXT_FAST_LEN = {}
+
+
+def next_fast_len(size):
```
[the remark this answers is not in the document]
fritzo: done

fritzo:
Thanks for review @fehiepsi !

Issue #2220, opened 2019-12-09
Batching and plating


tuananhle7:
co-authored with @iffsid 

Statistically, batching allows us to trade off the speed of taking gradient steps for lower variance gradient estimators. This can be done in a for loop. However, batching is most useful for parallelizing the computation of the log probs of the model and the guide by using low level numerical optimizations.

Plating is a useful construct for defining conditionally independent but identical random variables. While this is useful for defining hierarchical models, it is unsuited for batching since it **changes the definition of the model**. This leads to issues like IWAE-based algorithms failing silently, additional multiplicative factors in ELBO-based gradients, and being able to define nonsensical guides.

We propose defining a special keyword for plates that are supposed to indicate batching which circumvents the above problems.

### More details

Consider a model `p(z)p(x|z)` where `z` and `x` can have multiple sites and each site can have different shapes, depending on the distribution batch shapes but also the plates internal to the model. Given a dataset of training `xs`, we often want to take the gradient of `log p(x)`, averaged over a batch (subset of `xs`), as an estimator of `E_{xs}[grad log p(x)]`. In [practice](https://pyro.ai/examples/vae.html), it is [suggested](https://pyro.ai/examples/svi_part_ii.html#Subsampling-when-there-are-only-local-random-variables) that we **change** the model by adding a batching plate so that the model becomes `p(z_{1:B}, x_{1:B}) = prod_b p(z_b)p(x_b | z_b)`.

#### Problem 1: IWAE-based models silently fail

In IWAE-based models like [RenyiELBO](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/renyi_elbo.py) and [RWS](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/rws.py), we want to compute the model objective as `logsumexp` over the particle dimension and then average over the batch dimension. However, since there is no distinction between a batch dimension and a dimension of an actual plate, IWAE-based models are forced to sum over the batch dimension before taking the `logsumexp` over the particle dimension ([here](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/renyi_elbo.py#L99) and [here](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/rws.py#L130)). This is wrong since the order of these operation matters.

#### Problem 2: ELBO-based gradients have wrong multiplicative factors

[...]

PR #2321, opened 2020-02-21
Support plates in RenyiELBO

fritzo:
Addresses #2220 

This adds plate support to `RenyiELBO`. Once this merges, similar changes can be applied to `TraceAdaptiveTail_ELBO` and `ReweightedWakeSleep`.

## Tested
- [x] strengthened inference tests
- [x] added weak tests with various batch shapes
- ... other test suggestions welcome...

### DIFF
--- a/pyro/infer/renyi_elbo.py
+++ b/pyro/infer/renyi_elbo.py
@@ -9,7 +9,7 @@ import torch
 from pyro.distributions.util import is_identically_zero
 from pyro.infer.elbo import ELBO
 from pyro.infer.enum import get_importance_trace
-from pyro.infer.util import is_validation_enabled, torch_item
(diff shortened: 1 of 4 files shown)


fritzo:
cc @tuananhle7 @iffsid

iffsid:
cool, thanks!
We'll setup a PR once this merges. :)

fehiepsi:
The change looks reasonable to me. Just to verify my understanding, to support the previous behavior, I should declare different plates for z and x?
```
def model():
    with pyro.plate('z', 2):
        pyro.sample('z', ...)
    with pyro.plate('x', 2):
        pyro.sample('x', ...)
```

fritzo:
@fehiepsi yes good point, you can indeed declare different plates to achieve previous behavior.

PR #2324, opened 2020-02-23
Optimizing tests in GaussianChainTests

sleepy-owl:
Hi,

Some integration tests like `test_elbo_reparameterized_N_is_8` in `GaussianChainTests` take much longer to run than other tests. Reducing the iterations and adjusting the learning rate seems to reduce the running time (from ~`49s` to ~`10s` on my local machine). Note that all assertions still pass with these new values of parameters. 

Do you think that it is worth optimizing such tests? If yes, I can help optimize other slow running tests. Please let me know what you guys think and if you have any other suggestions/directions to look into.

Thanks!

### DIFF
--- a/tests/integration_tests/test_conjugate_gaussian_models.py
+++ b/tests/integration_tests/test_conjugate_gaussian_models.py
@@ -110,7 +110,7 @@ class GaussianChainTests(GaussianChain):
 
     def test_elbo_reparameterized_N_is_8(self):
         self.setup_chain(8)
-        self.do_elbo_test(True, 5000, 0.0015, 0.03, difficulty=1.0)
+        self.do_elbo_test(True, 1000, 0.008, 0.03, difficulty=1.0)
 
     @pytest.mark.skip("CI" in os.environ and os.environ["CI"] == "true",
                       "Skip slow test in travis.")
--- a/tests/integration_tests/test_conjugate_gaussian_models.py
+++ b/tests/integration_tests/test_conjugate_gaussian_models.py
@@ -110,7 +110,7 @@ class GaussianChainTests(GaussianChain):
 
     def test_elbo_reparameterized_N_is_8(self):
         self.setup_chain(8)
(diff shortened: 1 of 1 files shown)


fritzo:
Hi @sleepy-owl, sure we'd love help optimizing slow tests! Note some of the test parameters are set so that tests pass on GPU as well as CPU. We don't test GPU on CI because travis doesn't have GPUs. If you do optimize tests, could you also check they work with GPU by running `make test-cuda` or select individual tests via something like that make target:
```sh
CUDA_TEST=1 PYRO_TENSOR_TYPE=torch.cuda.DoubleTensor pytest -vx --stage unit --lax
CUDA_TEST=1 pytest -vx tests/test_examples.py::test_cuda
```
Thanks for any help!

fritzo:
cc @neerajprad who maintains our GPU tests

martinjankowiak:
hi @sleepy-owl thanks for taking a look at this! these are very old tests so no doubt there's room for improvement here. i'm not sure how random number seed generation is currently fixed in these tests, but before you make any changes it'd also be good to have some confidence that the tests pass under multiple seeds. we don't want, e.g., that a future version of pytorch shifts the randomness around a bit and tests start failing

sleepy-owl:
Thanks for the suggestions, @fritzo and @martinjankowiak! I will take a look at running the tests on `GPU` and running with `different seeds` and get back to you.

sleepy-owl:
Hi @martinjankowiak  and @fritzo,

I checked that the test passes on GPU using 
`CUDA_TEST=1 pytest -vx tests/integration_tests/test_conjugate_gaussian_models.py::GaussianChainTests::test_elbo_reparameterized_N_is_8`

Also, I removed all the seed setting statements in the test and ran the test multiple times (~50). The test passes 49 out of 50 times (i.e. with multiple seeds). It also passed when the original seed is set. 

I hope this should be good enough to merge. If you guys agree with this change, I can work on other slow tests in this repo.

Thanks!

PR #2325, opened 2020-02-24
Optimize gaussian op for scalar matrix

fehiepsi:
As suggested by @fritzo , we can improve the speed of GaussianHMM for the case `trans_dim, obs_dim == (1, 1)` by using scalar ops instead of linear algebra ops. This PR incorporates that enhancement.

Current tests (with trans_dim=obs_dim=1) already cover the implementation.

Profiling result: with
```
num_steps, obs_dim, hidden_dim = 6000, 1, 1
sample_shape = (5,)
batch_shape = (4,)
```
+ forward+backward: 52.9 ms (vs non-optimize one: 191 ms)
+ only forward: 12 ms (vs non-optimize one: 52.7 ms)

So we can observe 4x improvement with this PR.

### DIFF
--- a/pyro/ops/gaussian.py
+++ b/pyro/ops/gaussian.py
@@ -8,6 +8,7 @@ from torch.distributions.utils import lazy_property
 from torch.nn.functional import pad
 
 from pyro.distributions.util import broadcast_shape
+from pyro.ops.tensor_utils import cholesky, matmul, triangular_solve
 
 
 class Gaussian:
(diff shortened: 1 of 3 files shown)


pyro/ops/gaussian.py
```diff
@@ -170,10 +171,10 @@ def condition(self, value):
         P_bb = self.precision[..., n:, n:]
         b = value
 
-        info_vec = info_a - P_ab.matmul(b.unsqueeze(-1)).squeeze(-1)
+        info_vec = info_a - matmul(P_ab, b.unsqueeze(-1)).squeeze(-1)
```
fritzo: Maybe it's better to factor this out as `_mv(mat, vec)` or `matvecmul` or sth, both to speed up scalar case and to visually clean up usage. WDYT?

fehiepsi: Yup, I'll do that. It can be helpful for non-scalar vector too (though I don't expect much from it :).

PR #2329, opened 2020-02-27
Support warm starting in backtest()

fritzo:
Addresses #2311 

## Tested
- [x] smoke test with many window shapes

### DIFF
--- a/pyro/contrib/forecast/evaluate.py
+++ b/pyro/contrib/forecast/evaluate.py
@@ -70,6 +70,7 @@ def backtest(data, covariates, model_fn, guide_fn=None, *,
              stride=1,
              seed=1234567890,
              num_samples=100,
+             warm_start=False,
              forecaster_options={}):
     """
     Backtest a forecasting model on a moving window of (train,test) data.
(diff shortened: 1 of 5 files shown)


pyro/contrib/forecast/evaluate.py
```diff
@@ -173,3 +189,38 @@ def backtest(data, covariates, model_fn, guide_fn=None, *,
         del pred
 
     return results
+
+
+def _warm_start_param_store(t0, t1, t0_old, t1_old):
```
fritzo: @fehiepsi can you please review the indexing logic in this function?

fritzo:
@martinjankowiak please let me know if you need any additional functionality.
💯if you get a chance to test on your problem before merging 🙂

pyro/contrib/forecast/evaluate.py
```diff
@@ -173,3 +189,38 @@ def backtest(data, covariates, model_fn, guide_fn=None, *,
...
+        if time_dim is None:
+            continue
+        assert time_dim < 0
+        time_dim += constrained_param.dim() - param.dim()
+        with torch.no_grad():
+
+            # Truncate params that have just left the window.
+            if t0 != t0_old:
+                dots = (slice(None),) * (param.dim() + time_dim)
+                param = param[dots + (slice(t0_old - t0, None),)]
```
fehiepsi: I think `t0_old - t0` should be `t0 - t0_old`. In addition, you might not need the condition `if t0 != t0_old`.

fritzo:
Closing as this will not handle `LinearHMM` auxiliary parameters because they occur outside of the time plate. We'll need something closer to `PrefixReplayMessenger` logic...

PR #2335, opened 2020-03-03
Add a helper to handle seasonality via regression

fritzo:
Addresses #2311 

This adds a `periodic_features()` helper that can be used to generate covariates for forecasting models.

## Tested
- [x] added shape and bound tests

### DIFF
--- a/pyro/ops/tensor_utils.py
+++ b/pyro/ops/tensor_utils.py
@@ -118,8 +118,8 @@ def periodic_features(duration, max_period=None, min_period=None, **options):
     treated via regression. When only ``max_period`` is specified this
     generates periodic features at all length scales. When also ``min_period``
     is specified this generates periodic features at large length scales, but
-    omits high frequecy features. This is useful when combining regression for
-    long seasonality with other techinques like :func:`periodic_repeat` and
+    omits high frequency features. This is useful when combining regression for
+    long seasonality with other techniques like :func:`periodic_repeat` and
     :func:`periodic_cumsum` for short time scales. For example, to combine
     regress yearly seasonality down to the scale of one week one could set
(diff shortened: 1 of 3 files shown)


pyro/ops/tensor_utils.py
```diff
@@ -109,6 +109,39 @@ def periodic_cumsum(tensor, period, dim):
...
+    treated via regression, and short seasonality can be treated via the
+    :func:`periodic_repeat` and :func:`periodic_cumsum` helpers. For example,
+    to combine yearly with weekly seasonality one could set
+    ``max_period=365.25`` and ``min_period=7``.
+
+    :param int duration: Number of discrete time steps.
+    :param float max_period: Optional max period, defaults to ``duration``.
+    :param float min_period: Optional min period (exclusive), defaults to 2 = Nyquist cutoff.
+    :param \*\*options: Tensor construction options, e.g. ``dtype`` and ``device``.
+    :returns: A ``(duration, 2 * ceil(max_period - min_period) - 2)``-shaped tensor of features
```
fehiepsi: nit: `2 * ceil(max_period - min_period) - 2` -> `2 * floor(max_period / min_period)`

fritzo: oops, will fix docs. but I'm pretty sure `ceil(x) - 1 != floor(x)` and that we need `ceil` here. Am I mistaken?

fehiepsi: nope, you are right. `arange(1, 3.)` returns `1, 2`

fritzo: fixed.

fritzo:
Thanks for your careful review, @fehiepsi !

PR #2342, opened 2020-03-05
improve stability of SymmetricStableReparam

martinjankowiak:
i've been running into NaN gradients using SymmetricStableReparam. restricting the range of the uniform auxiliary seems to help. suggestions for better/other ways to improve stability?

to reproduce numerical issues run 

`python multivariate.py --noise-model stable --clip-norm 5.0 --learning-rate 0.01`

on the `stablets` branch of pyro-ppl/sandbox (see folder `2020-03-forecasting`). 

if you set `epsilon=0.0` you should see NaN parameters shortly after

`INFO     Training on window [0:619], testing on window [619:620]`

with this diff the script should run to completion.

### DIFF
--- a/pyro/infer/reparam/stable.py
+++ b/pyro/infer/reparam/stable.py
@@ -79,7 +79,13 @@ class SymmetricStableReparam(Reparam):
     [1] Alvaro Cartea and Sam Howison (2009)
         "Option Pricing with Levy-Stable Processes"
         https://pdfs.semanticscholar.org/4d66/c91b136b2a38117dd16c2693679f5341c616.pdf
+
+    :param float epsilon: Small numerical factor that helps with stability at the price of a
+        small bias in the auxiliary sampler. Defaults to 1.0e-4. Set to 0.0 to avoid bias.
     """
+    def __init__(self, epsilon=1.0e-4):
+        self.epsilon = epsilon
+
     def __call__(self, name, fn, obs):
         fn, event_dim = self._unwrap(fn)
         assert isinstance(fn, dist.Stable) and fn.coords == "S0"
@@ -94,7 +100,7 @@ class SymmetricStableReparam(Reparam):
         half_pi = proto.new_full(proto.shape, math.pi / 2)
         one = proto.new_ones(proto.shape)
         u = pyro.sample("{}_uniform".format(name),
-                        self._wrap(dist.Uniform(-half_pi, half_pi), event_dim))
+                        self._wrap(dist.Uniform(-half_pi + self.epsilon, half_pi - self.epsilon), event_dim))
         e = pyro.sample("{}_exponential".format(name),
                         self._wrap(dist.Exponential(one), event_dim))
 


pyro/infer/reparam/stable.py
```diff
@@ -94,7 +100,7 @@ def __call__(self, name, fn, obs):
         half_pi = proto.new_full(proto.shape, math.pi / 2)
         one = proto.new_ones(proto.shape)
         u = pyro.sample("{}_uniform".format(name),
-                        self._wrap(dist.Uniform(-half_pi, half_pi), event_dim))
+                        self._wrap(dist.Uniform(-half_pi + self.epsilon, half_pi - self.epsilon), event_dim))
```
martinjankowiak: it's likely similar effects could be achieved by instead modifying computations in `_unsafe_standard_stable`.

perhaps `(v - V).cos()`?  (i haven't thought through the math in much detail)

fehiepsi: This looks like the right direction. I’ll take a look.

fehiepsi: You pointed out the right place. It would be better to use
```
(v - V).cos().clamp_(min=torch.finfo(v.dtype).tiny)
```
This is precision issue
```
alpha = torch.tensor(0.99)
half_pi = math.pi / 2
assert half_pi > (half_pi - (1 - alpha) * 1e-6)
```
fails in single precision; hence cosine of RHS < 0.

fritzo: This seems reasonable. It would be nice to implement it in a way that applies to all stable logic, not only this reparameterizer, e.g. in `_unsafe_standard_stable()` as you suggest. Maybe we can add something like
```py
def _unsafe_standard_stable(..., tol=1e-4):
    if epsilon:
        assert 0 < tol < 1
        u = u * (1 - tol)
    ...
```
and expose `tol` through each reparameterizer and/or the Stable distribution? @fehiepsi WDYT?

fritzo: @fehiepsi great, I'd love to find a solution that did not expose parameters like `tol` and `epsilon`!

fritzo:
I'm having some successs with
```diff
diff --git a/pyro/distributions/stable.py b/pyro/distributions/stable.py
index 32a2f11e..aa215be1 100644
--- a/pyro/distributions/stable.py
+++ b/pyro/distributions/stable.py
@@ -10,11 +10,17 @@ from torch.distributions.utils import broadcast_all
 from pyro.distributions.torch_distribution import TorchDistribution


-def _unsafe_standard_stable(alpha, beta, V, W, coords):
+def _unsafe_standard_stable(alpha, beta, V, W, coords, tol=1e-6):
     # Implements a noisily reparametrized version of the sampler
     # Chambers-Mallows-Stuck method as corrected by Weron [1,3] and simplified
     # by Nolan [4]. This will fail if alpha is close to 1.

+    # Numerically stabilize.
+    if tol:
+        assert 0 < tol < 1
+        V = V.mul(1 - tol)
+        W = W + tol
+
     # Differentiably transform noise via parameters.
     assert V.shape == W.shape
     inv_alpha = alpha.reciprocal()
```

fritzo:
@martinjankowiak can you see if the `_unsafe_standard_stable(..., tol=1e-6)` version works for you? I've included that in #2343 .

fehiepsi:
@martinjankowiak @fritzo How about the following solution?
+ clamp at V to make V belong to (-pi/2, pi/2)
+ clamp at (v - V) because `v - V = half_pi - (half_pi + V) * (1 - alpha)`; when `(half_pi + V) * (1 - alpha)` small, `v - V` can be greater than `half_pi` (precision issue). Even if we make `half_pi + V > 1e-6`, when alpha near 1, that term is still small. 
```diff
@@ -18,10 +18,16 @@ def _unsafe_standard_stable(alpha, beta, V, W, coords):
     # Differentiably transform noise via parameters.
     assert V.shape == W.shape
     inv_alpha = alpha.reciprocal()
-    b = beta * (math.pi / 2 * alpha).tan()
-    v = b.atan() + alpha * V
+    half_pi = math.pi / 2
+    finfo = torch.finfo(V.dtype)
+    # make V belong to the open interval (-pi/2, pi/2)
+    V = V.clamp(min=-half_pi + finfo.eps)
+    ha = half_pi * alpha
+    b = beta * ha.tan()
+    # +/- `ha` term to keep the precision of alpha * (V + half_pi) when V ~ -half_pi
+    v = b.atan() - ha + alpha * (V + half_pi)
     Z = v.sin() / ((1 + b * b).rsqrt() * V.cos()).pow(inv_alpha) \
-        * ((v - V).cos() / W).pow(inv_alpha - 1)
+        * ((v - V).cos().clamp(min=finfo.tiny) / W).pow(inv_alpha - 1)
     Z.data[Z.data != Z.data] = 0  # drop occasional NANs
```

martinjankowiak:
> @martinjankowiak @fritzo How about the following solution?

seems reasonable. did you test this on multivariate.py? do you want to open a PR?

fehiepsi:
Yes, `multivariate.py` passes with that change. I'll make a PR for it.

fritzo:
Closing in favor of @fehiepsi's #2344

PR #2344, opened 2020-03-06
Fix NaN issue of unsafe_stable

fehiepsi:
Addresses #2342. This PR does not restrict the domain of the auxiliary uniform variable.

Test: pass @martinjankowiak 's `multivariate.py` example.


pyro/distributions/stable.py
```diff
@@ -18,10 +18,16 @@ def _unsafe_standard_stable(alpha, beta, V, W, coords):
     # Differentiably transform noise via parameters.
     assert V.shape == W.shape
     inv_alpha = alpha.reciprocal()
-    b = beta * (math.pi / 2 * alpha).tan()
-    v = b.atan() + alpha * V
+    half_pi = math.pi / 2
+    finfo = torch.finfo(V.dtype)
+    # make V belong to the open interval (-pi/2, pi/2)
+    V = V.clamp(min=-half_pi + finfo.eps)
```
fritzo: I think you'll need 2 * eps to avoid underflow, right? WDYT of
```py
V.data.clamp_(min=2 * finfo.eps - half_ip, max=half_pi - 2 * finfo.eps)
```

fehiepsi: `eps - half_pi` is a float number so underflow does not happen I think (I tested clamping with `eps-half_pi` and it gives a number greater than `-half_pi`). We don't need to clamp the upper point because `uniform` generates samples in [0, 1).

fritzo: My point was that `-half_pi + finfo.eps == -half_pi` because eps undeflows. I'd feel safer clamping more conservatively to something significantly inside the interval.

fritzo: The reason we need to clip both directions is that we can't rely on `uniform`'s generation behavior. When `_unsafe_standard_stable()` is called inside HMC it is created from a replayed transfomed real variable; when it is called inside SVI it is created from a transformed sample from normal or other distribution. Those transforms introduce noise so we need to clip from both sides.

fritzo:
@fehiepsi I have confirmed that making your fix a little more conservative fixes my tutorial 😄 
```diff
diff --git a/pyro/distributions/stable.py b/pyro/distributions/stable.py
index 32a2f11e..9047a147 100644
--- a/pyro/distributions/stable.py
+++ b/pyro/distributions/stable.py
@@ -18,10 +18,16 @@ def _unsafe_standard_stable(alpha, beta, V, W, coords):
     # Differentiably transform noise via parameters.
     assert V.shape == W.shape
     inv_alpha = alpha.reciprocal()
-    b = beta * (math.pi / 2 * alpha).tan()
-    v = b.atan() + alpha * V
+    half_pi = math.pi / 2
+    eps = torch.finfo(V.dtype).eps
+    # make V belong to the open interval (-pi/2, pi/2)
+    V = V.clamp(min=2 * eps - half_pi, max=half_pi - 2 * eps)
+    ha = half_pi * alpha
+    b = beta * ha.tan()
+    # +/- `ha` term to keep the precision of alpha * (V + half_pi) when V ~ -half_pi
+    v = b.atan() - ha + alpha * (V + half_pi)
     Z = v.sin() / ((1 + b * b).rsqrt() * V.cos()).pow(inv_alpha) \
-        * ((v - V).cos() / W).pow(inv_alpha - 1)
+        * ((v - V).cos().clamp(min=eps) / W).pow(inv_alpha - 1)
     Z.data[Z.data != Z.data] = 0  # drop occasional NANs
```
Nice work, your solution is cleaner than the clamping martin and I tried to do!

pyro/distributions/stable.py
fehiepsi: > Those transforms introduce noise

Yes, I noticed it too: it happens that V < -half_pi after transform. If we are going to clamp using 2eps then I agree that we should do it two sides. Thanks for explaining!

pyro/distributions/stable.py
```diff
@@ -18,10 +18,16 @@ def _unsafe_standard_stable(alpha, beta, V, W, coords):
...
+    eps = torch.finfo(V.dtype).eps
+    # make V belong to the open interval (-pi/2, pi/2)
+    V = V.clamp(min=2 * eps - half_pi, max=half_pi - 2 * eps)
+    ha = half_pi * alpha
+    b = beta * ha.tan()
+    # +/- `ha` term to keep the precision of alpha * (V + half_pi) when V ~ -half_pi
+    v = b.atan() - ha + alpha * (V + half_pi)
     Z = v.sin() / ((1 + b * b).rsqrt() * V.cos()).pow(inv_alpha) \
-        * ((v - V).cos() / W).pow(inv_alpha - 1)
+        * ((v - V).cos().clamp(min=eps) / W).pow(inv_alpha - 1)
```
fehiepsi: @fritzo Just to confirm: clamp by `eps` will make the result after `pow(inv_alpha-1)` a bit larger than using tiny (when `inv_alpha - 1 ~ 0.01`, it is 0.8 vs 0.4). Is it fine for you?

fehiepsi: nvm, we should concern grad issue too; using eps looks safer.

fritzo: That's fine with me. I was worried about the gradients: if the power is small and the value is tiny, then gradients can be very large.

Do you think we should similarly clamp the preceding line?
```py
((1 + b * b).rsqrt() * V.cos()).clamp(min=eps).pow(inv_alpha)
```

fritzo: Oh nevermind, I see the preceding line is already relatively safe due to the eps clamping of `V`.

PR #2347, opened 2020-03-07
add HMCForecaster

fehiepsi:
This PR supports HMC for Forecaster.

TODO:
- [x] add test

### DIFF
--- a/pyro/contrib/forecast/forecaster.py
+++ b/pyro/contrib/forecast/forecaster.py
@@ -10,8 +10,9 @@ import torch.nn as nn
 import pyro
 import pyro.distributions as dist
 import pyro.poutine as poutine
-from pyro.infer import SVI, Trace_ELBO
+from pyro.infer import MCMC, NUTS, SVI, Trace_ELBO
 from pyro.infer.autoguide import AutoNormal, init_to_sample
+from pyro.infer.predictive import _guess_max_plate_nesting
 from pyro.nn.module import PyroModule
 from pyro.optim import DCTAdam
 
@@ -282,4 +283,62 @@ class Forecaster(nn.Module):
         with PrefixReplayMessenger(tr.trace):
             with PrefixConditionMessenger(self.model._prefix_condition_data):
                 with pyro.plate("particles", num_samples, dim=dim):
+                    return self.model(data, covariates)
+
+
+class HMCForecaster(nn.Module):
+    """
+    Forecaster for a :class:`ForecastingModel` using Hamiltonian Monte Carlo.
+
+    On initialization, this will run MCMC to get posterior samples of the model.
+
+    After construction, this can be called to generate sample forecasts.
+
+    :param callable model: A forecasting model.
+    :param data: A 1D time series data.
+    :param covariates: An array of covariates with time dimension -2. For models not
+        using covariates, pass a shaped empty array ``np.empty((duration, 0))``.
+    :param int num_warmup: number of MCMC warmup steps.
+    :param int num_samples: number of MCMC samples.
+    :param int num_chains: number of parallel MCMC chains.
+    """
+    def __init__(self, model, data, covariates=None, *,
+                 num_warmup=1000, num_samples=1000, num_chains=1, seed=0):
+        assert data.size(-2) == covariates.size(-2)
+        super().__init__()
+        self.model = model
+        self.data = data
+        self.covariates = covariates
+        self.num_samples = num_samples * num_chains
+        max_plate_nesting = _guess_max_plate_nesting(model, (data, covariates), {})
+        self.max_plate_nesting = max(max_plate_nesting, 1)  # force a time plate
+
+        mcmc = MCMC(NUTS(model), num_warmup, num_samples, num_chains=num_chains)
+        mcmc.run(data, covariates)
+        mcmc.summary()
+
+        # inspect the model with particles plate = 1
+        with poutine.trace() as tr:
+            with pyro.plate("particles", 1, dim=-self.max_plate_nesting - 1):
+                model(data, covariates)
+
+        self._trace = tr.trace
+        samples = mcmc.get_samples()
+        for name, node in list(self._trace.nodes.items()):
+            if name not in samples:
+                del self._trace.nodes[name]
+            else:
+                sample = samples[name]
+                node['value'] = sample.reshape(
+                    (self.num_samples,) + (1,) * (node['value'].dim() - sample.dim()) + sample.shape[1:])
+
+    @torch.no_grad()
+    def forward(self, data, covariates):
+        assert data.size(-2) < covariates.size(-2)
+        assert self.max_plate_nesting >= 1
+        dim = -1 - self.max_plate_nesting
+
+        with PrefixReplayMessenger(self._trace):
+            with PrefixConditionMessenger(self.model._prefix_condition_data):
+                with pyro.plate("particles", self.num_samples, dim=dim):
                     return self.model(data, covariates)
--- a/pyro/contrib/forecast/forecaster.py
+++ b/pyro/contrib/forecast/forecaster.py
@@ -310,13 +310,13 @@ class HMCForecaster(nn.Module):
         Default to False.
     :param bool jit_compile: whether to use the PyTorch JIT to trace the log
         density computation, and use this optimized executable trace in the
-        integrator. Default to True.
+        integrator. Default to False.
     :param int max_tree_depth: Max depth of the binary tree created during the doubling
         scheme of :class:`~pyro.infer.mcmc.nuts.NUTS` sampler. Default to 10.
     """
     def __init__(self, model, data, covariates=None, *,
                  num_warmup=1000, num_samples=1000, num_chains=1,
-                 dense_mass=False, jit_compile=True, max_tree_depth=10):
+                 dense_mass=False, jit_compile=False, max_tree_depth=10):
         assert data.size(-2) == covariates.size(-2)
         super().__init__()
         self.model = model
--- a/pyro/contrib/forecast/forecaster.py
+++ b/pyro/contrib/forecast/forecaster.py
@@ -294,10 +294,14 @@ class HMCForecaster(nn.Module):
 
     After construction, this can be called to generate sample forecasts.
 
-    :param callable model: A forecasting model.
-    :param data: A 1D time series data.
-    :param covariates: An array of covariates with time dimension -2. For models not
-        using covariates, pass a shaped empty array ``np.empty((duration, 0))``.
+    :param ForecastingModel model: A forecasting model subclass instance.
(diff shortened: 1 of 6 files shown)


pyro/contrib/forecast/forecaster.py
```diff
@@ -283,3 +284,64 @@ def forward(self, data, covariates, num_samples):
...
+        for name, node in list(self._trace.nodes.items()):
+            if name not in samples:
+                del self._trace.nodes[name]
+            else:
+                sample = samples[name]
+                node['value'] = sample.reshape(
+                    (self.num_samples,) + (1,) * (node['value'].dim() - sample.dim()) + sample.shape[1:])
+
+    @torch.no_grad()
+    def forward(self, data, covariates):
```
fehiepsi: Do we need to support `num_samples` (by subsampling from mcmc samples) here? I recall that @martinjankowiak does not like that idea because MCMC samples are expensive.

fritzo: I would like to keep the interfaces identical, so yes I'd like to keep a `num_samples` kwarg. This is also required for `crps` evaluation and plotting quantiles.

Can we not simply reuse the existing samples?

pyro/contrib/forecast/forecaster.py
```diff
@@ -283,3 +284,64 @@ def forward(self, data, covariates, num_samples):
...
+        For models not using covariates, pass a shaped empty tensor
+        ``torch.empty(duration, 0)``.
+    :type covariates: ~torch.Tensor
+
+    :param int num_warmup: number of MCMC warmup steps.
+    :param int num_samples: number of MCMC samples.
+    :param int num_chains: number of parallel MCMC chains.
+    """
+    def __init__(self, model, data, covariates=None, *,
+                 num_warmup=1000, num_samples=1000, num_chains=1):
```
martinjankowiak: can we please add support for more hmc options? `full_mass`, `max_tree_depth`, maybe jit?

fehiepsi: Sure, I'll add them. `jit` is important btw, I'll enable it by default.

pyro/contrib/forecast/forecaster.py
fehiepsi: So I'll support that.

> Can we not simply reuse the existing samples?

In other frameworks, they use subsampling (Stan, Rlgt, PyMC3).

fehiepsi: @fritzo I just support `num_samples` with subsampling.

tests/contrib/forecast/test_forecaster.py
```diff
@@ -103,17 +103,24 @@ def model(self, zero_data, covariates):
 @pytest.mark.parametrize("t_obs", [1, 7])
 @pytest.mark.parametrize("t_forecast", [1, 3])
 @pytest.mark.parametrize("batch_shape", [(), (4,), (3, 2)], ids=str)
-@pytest.mark.parametrize("cov_dim", [0, 1, 6])
+@pytest.mark.parametrize("cov_dim", [0, 1, 5])
 @pytest.mark.parametrize("obs_dim", [1, 2])
 @pytest.mark.parametrize("dct_gradients", [False, True])
 @pytest.mark.parametrize("Model", [Model0, Model1, Model2, Model3, Model4])
-def test_smoke(Model, batch_shape, t_obs, t_forecast, obs_dim, cov_dim, dct_gradients):
+@pytest.mark.parametrize("engine", ["svi", "hmc"])
```
fehiepsi: `hmc` test is slow (~3m in my system) :(

fritzo: can we specify fewer samples and shorter warmup etc. to speed it up?

pyro/contrib/forecast/forecaster.py
```diff
@@ -283,3 +284,79 @@ def forward(self, data, covariates, num_samples):
...
+    :param int num_warmup: number of MCMC warmup steps.
+    :param int num_samples: number of MCMC samples.
+    :param int num_chains: number of parallel MCMC chains.
+    :param bool dense_mass: a flag to decide if mass matrix is dense or diagonal.
+        Default to False.
+    :param bool jit_compile: whether to use the PyTorch JIT to trace the log
+        density computation, and use this optimized executable trace in the
+        integrator. Default to True.
+    :param int max_tree_depth: Max depth of the binary tree created during the doubling
+        scheme of NUTS sampler. Default to 10.
```
fritzo: nit: could you add a sphinx link to the NUTS class?

tests/contrib/forecast/test_forecaster.py
fehiepsi: awesome! Time is reduced to 30s now. Thanks a lot!

fehiepsi:
@fritzo This does not add HMCForecaster to backtest. Could you suggest an api for it in `backtest`? I'll fill details then.

fritzo:
@fehiepsi I think we add a `forcaster_fn` kwarg to `backtest()` and then remove the `guide_fn` kwarg (because `guide_fn` can the be simulated by wrapping `Forecaster` in a smarter `forecaster_fn`). Because the interfaces are the same, `backtest()` should still work. You should be able to ignore the `warm_start` logic because `HMCForecaster` does not accept a `warm_start` kwarg.
```diff
diff --git a/pyro/contrib/forecast/evaluate.py b/pyro/contrib/forecast/evaluate.py
index 1b551bfd..8a5fe4d2 100644
--- a/pyro/contrib/forecast/evaluate.py
+++ b/pyro/contrib/forecast/evaluate.py
@@ -66,7 +66,8 @@ DEFAULT_METRICS = {
 }
 
 
-def backtest(data, covariates, model_fn, guide_fn=None, *,
+def backtest(data, covariates, model_fn, *,
+             forecaster_fn=Forecaster,
              metrics=None,
              transform=None,
              train_window=None,
@@ -162,9 +163,7 @@ def backtest(data, covariates, model_fn, guide_fn=None, *,
         train_data = data[..., t0:t1, :]
         train_covariates = covariates[..., t0:t1, :]
         model = model_fn()
-        guide = None if guide_fn is None else guide_fn()
-        forecaster = Forecaster(model, train_data, train_covariates,
-                                guide=guide, **forecaster_options)
+        forecaster = forecaster_fn(model, train_data, train_covariates,
+                                   **forecaster_options)
 
         # Forecast forward to testing window.
         test_covariates = covariates[..., t0:t2, :]
```

fritzo:
@fehiepsi can you also add a half sentence to the forecasting overview docs with a sphinx link to the `HMCForecaster` class?
https://github.com/pyro-ppl/pyro/blob/dev/docs/source/contrib.forecast.rst

fehiepsi:
Thanks, @fritzo ! I have incorporated your suggestions.

fehiepsi:
@fritzo @martinjankowiak I run HMCForecaster for tutorial i in [this gist](https://gist.github.com/fehiepsi/8bc0cd43289a60e79c176d73912a7fb2). For some reasons, with `jit_compile=True`, we got an error with Model2 (the traceback can be seen in the above gist). So I set `jit_compile=False` by default, to investigate later. It seems that HMC gives better results, but it is slow.

pyro/contrib/forecast/forecaster.py
```diff
@@ -283,3 +284,80 @@ def forward(self, data, covariates, num_samples):
...
+    :type data: ~torch.Tensor
+    :param covariates: A tensor of covariates with time dimension -2.
+        For models not using covariates, pass a shaped empty tensor
+        ``torch.empty(duration, 0)``.
+    :type covariates: ~torch.Tensor
+
+    :param int num_warmup: number of MCMC warmup steps.
+    :param int num_samples: number of MCMC samples.
+    :param int num_chains: number of parallel MCMC chains.
+    :param bool dense_mass: a flag to decide if mass matrix is dense or diagonal.
```
martinjankowiak: nit: A flag that controls whether the mass matrix is dense or diagonal. Defaults to False.

pyro/contrib/forecast/forecaster.py
```diff
@@ -283,3 +284,80 @@ def forward(self, data, covariates, num_samples):
...
+        Default to False.
+    :param bool jit_compile: whether to use the PyTorch JIT to trace the log
+        density computation, and use this optimized executable trace in the
+        integrator. Default to False.
```
martinjankowiak: nit: Defaults

pyro/contrib/forecast/forecaster.py
```diff
@@ -283,3 +284,80 @@ def forward(self, data, covariates, num_samples):
...
+    :param int max_tree_depth: Max depth of the binary tree created during the doubling
+        scheme of :class:`~pyro.infer.mcmc.nuts.NUTS` sampler. Default to 10.
```
martinjankowiak: nit: "scheme of the"
nit: Defaults

pyro/contrib/forecast/forecaster.py
```diff
@@ -283,3 +284,80 @@ def forward(self, data, covariates, num_samples):
...
+        with poutine.trace() as tr:
+            with pyro.plate("particles", 1, dim=-self.max_plate_nesting - 1):
+                model(data, covariates)
+
+        self._trace = tr.trace
+        self._samples = mcmc.get_samples()
+        self._num_samples = num_samples * num_chains
+        for name, node in list(self._trace.nodes.items()):
+            if name not in self._samples:
+                del self._trace.nodes[name]
```
martinjankowiak: what is this doing?

pyro/contrib/forecast/forecaster.py
```diff
@@ -283,3 +284,80 @@ def forward(self, data, covariates, num_samples):
...
+
+    @torch.no_grad()
+    def forward(self, data, covariates, num_samples):
+        assert data.size(-2) < covariates.size(-2)
+        assert self.max_plate_nesting >= 1
+        dim = -1 - self.max_plate_nesting
+
+        weights = torch.ones(self._num_samples, device=data.device)
+        indices = torch.multinomial(weights, num_samples, replacement=True)
```
martinjankowiak: perhaps replacement should be False if `num_samples <= self._num_samples`?

martinjankowiak:
@fehiepsi i can confirm this seems to give reasonable results on a univariate model i've been playing around with. nice work!

pyro/contrib/forecast/forecaster.py
fritzo: as I understand this is garbage collecting param statements to save memory

fehiepsi: In addition, this is to remove sites such as residuals, which should not be replayed in forward method.

martinjankowiak:
@fritzo ready to merge as far as i'm concerned

PR #2348, opened 2020-03-07
Add a forecasting tutorial on hierarchical models

fritzo:
Addresses #2311 

This also makes some slight changes to improve usability:
- fixes a restriction in `crps_empirical()`
- disables subsampling `Forecaster` after training is done

See notebook via [nbviewer](https://nbviewer.jupyter.org/github/pyro-ppl/pyro/blob/forecast-tutorial-iii/tutorial/source/forecasting_iii.ipynb).


fritzo:
> do we need expanding `scale` in Model2?

Yes, that is needed so that `noise_dist` has full time extent. I started to automate in that logic in `Forecaster.predict()` but it got too complex for this PR. Mind if I defer that to a follow-up PR?

martinjankowiak:
another brilliant tutorial!
nits/typos/suggestions/questions/comments:
- typo "is conditionally" => "are"
- "# Now outside of the time plate" isn't this confusing since these non-pyro-statements can occur anywhere?
- typo "O(num_stations ** 1 * duration)"
- is there missing data or did you remove that?
- "adjust gradient estimates" => "adjust per-parameter learning rates"? or do you mean something else?
- typo: runing

fritzo:
> "# Now outside of the time plate" isn't this confusing since these non-pyro-statements can occur anywhere?

Well there's a subtle point that you shouldn't mix elements within a plate, and that the following would be invalid in some cases:
```py
with self.time_plate:
    x = pyro.sample("X", dist.Normal(0, 1))
    y = x.cumsum(-1)
    z = pyro.sample("z", dist.Normal(y, 1))
```
My comment aims to nudge users to avoid this subtle behavior by avoiding any mixing inside the time plate. Let me know if you can think of a clearer way to nudge users.

> is there missing data or did you remove that?

The data is fully observed, but some of the counts are zero, e.g. Antioch station is all zero until it opened in 2018. However I'm intentionally modeling these long all-zero sections as a Stable process: the process can easily jump from zero to some nonzero state.

martinjankowiak:
yay 3 awesome tutorials!

fritzo:
Thanks for reviewing!

PR #2359, opened 2020-03-10
Tidy `pyro.distributions.transforms` tests

stefanwebb:
As the library of bijective transforms has grown (`pyro.distributions.transforms`), the code to test it (`tests/distributions/test_transforms.py`) has become unwieldy... There are up to three tests per transform (Jacobian, inverse, and output shape) in separate functions, and lots of duplicated code.

I first took the idea from the implementer of the lower Cholesky affine test to combine all tests for a given transform into one method. Then I made a helper method to avoid duplicated code.

In the process, I discovered that some element-wise transforms were not having their Jacobian tested. And also that some volume preserving transforms were not having their Jacobian tested for being 0.

One side effect of this PR is that it should make it easier for contributors to write tests for their bijections.

### DIFF
--- a/tests/distributions/test_transforms.py
+++ b/tests/distributions/test_transforms.py
@@ -136,14 +136,14 @@ class TransformTests(TestCase):
 
     def test_elu(self):
         # NOTE: Need following hack since helper function mistakenly doesn't take input dim
-        self._test(lambda input_dim: T.elu(), jacobian=False)
+        self._test(lambda input_dim: T.elu())
 
     def test_householder(self):
-        self._test(partial(T.householder, count_transforms=2), jacobian=False)
+        self._test(partial(T.householder, count_transforms=2))
 
     def test_leaky_relu(self):
         # NOTE: Need following hack since helper function mistakenly doesn't take input dim
-        self._test(lambda input_dim: T.leaky_relu(), jacobian=False)
+        self._test(lambda input_dim: T.leaky_relu())
 
     def test_lower_cholesky_affine(self):
         # NOTE: Need following hack since helper function unimplemented
--- a/tests/distributions/test_transforms.py
+++ b/tests/distributions/test_transforms.py
@@ -9,6 +9,8 @@ import torch
 import pyro.distributions as dist
 import pyro.distributions.transforms as T
 
+from functools import partial
+
 pytestmark = pytest.mark.init(rng_seed=123)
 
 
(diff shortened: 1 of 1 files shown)


stefanwebb:
I'll wait until this is merged before revisting #2246

tests/distributions/test_transforms.py
```diff
@@ -81,208 +83,97 @@ def _test_shape(self, base_shape, transform):
...
+    def test_affine_autoregressive(self):
+        for stable in [True, False]:
+            self._test(partial(T.affine_autoregressive, stable=stable))
+
+    # Affine Coupling
+    def test_affine_coupling(self):
+        self._test(T.affine_coupling)
+
+    # Batchnorm
+    def test_batchnorm(self):
```
martinjankowiak: what's going on in `test_batchnorm`?

tests/distributions/test_transforms.py
```diff
@@ -81,208 +83,97 @@ def _test_shape(self, base_shape, transform):
...
+        self._test_conditional(T.conditional_planar, inverse=False)
 
-    def test_cond_planar_jacobians(self):
-        observed_dim = 3
-        for input_dim in [2, 5, 10]:
-            z = torch.rand(observed_dim)
-            transform = T.conditional_planar(input_dim, observed_dim)
-            self._test_jacobian(input_dim, transform.condition(z))
+    def test_discrete_cosine(self):
+        # NOTE: Need following hack since helper function unimplemented
```
martinjankowiak: not sure if i would call these "hacks"?

tests/distributions/test_transforms.py
```diff
@@ -81,208 +83,97 @@ def _test_shape(self, base_shape, transform):
...
+        self._test(lambda input_dim: T.DiscreteCosineTransform())
 
-    def test_poly_jacobians(self):
-        for input_dim in [2, 5, 10]:
-            self._test_jacobian(input_dim, T.polynomial(input_dim))
+    def test_elu(self):
+        # NOTE: Need following hack since helper function mistakenly doesn't take input dim
```
martinjankowiak: do you intend to fix this separately?

stefanwebb: Yes, I'd like to clean up this in a separate PR. And some other clean-ups like removing the superfluous `Transform` suffix from some of the transform names... I think I will need to think about backwards compatibility here (a warning message saying classes/methods are deprecated)

tests/distributions/test_transforms.py
stefanwebb: Well, it's much cleaner if the help function is in the relevant transform module. I will add this in another (short) PR!

tests/distributions/test_transforms.py
stefanwebb: From the docs, the reason we need `transform.eval()` is explained by:

> Also, when the module is set to training mode, the sample mean and variance on the current minibatch are used in place of the smoothed averages, :math:`\\hat{\\mu}` and :math:`\\hat{\\sigma^2}`, for the inverse operator. For this reason it is not the case that :math:`x=g(g^{-1}(x))` during training, i.e., that the inverse operation is the inverse of the forward one.

I'm not entirely certain if the following line is necessary: `transform._inverse(torch.normal(torch.arange(0., input_dim), torch.arange(1., 1. + input_dim) / input_dim))`. What it does is populate the smoothed averages

stefanwebb: I think populating the smoothed averages is necessary actually since if you don't do that then they are all 0's and 1's, and if all dimensions of the smoothed average are the same this could obscure an error in the transform

PR #2360, opened 2020-03-10
Fix nan issue of stable reparam

fehiepsi:
Resolves an issue of #2299.

TODO:
- [x] fix the instability around `skew ~ 0`. I want to make `skew` has a consistent sign there.
- [x] ~fix nan issue when learning rate is high~ this is an inference issue, not the instability of stable distribution.

### DIFF
--- a/pyro/infer/reparam/stable.py
+++ b/pyro/infer/reparam/stable.py
@@ -178,11 +178,12 @@ class StableReparam(Reparam):
         z = _unsafe_standard_stable(a / 2, 1, zu, ze, coords="S")
         t = _standard_stable(a, one, tu, te, coords="S0")
         a_inv = a.reciprocal()
+        eps = torch.finfo(a.dtype).tiny
         skew_abs = fn.skew.abs()
-        t_scale = skew_abs.clamp(min=torch.finfo(a.dtype).tiny).pow(a_inv)
+        t_scale = skew_abs.pow(a_inv)
         s_scale = (1 - skew_abs).pow(a_inv)
-        shift = _safe_shift(a, skew_abs, t_scale)
-        loc = fn.loc + fn.scale * fn.skew.sign() * (t * t_scale + shift)
+        shift = _unsafe_shift(a, fn.skew, t_scale)
+        loc = fn.loc + fn.scale * (fn.skew.sign() * t_scale * t + shift)
         scale = fn.scale * s_scale * z.sqrt() * (math.pi / 4 * a).cos().pow(a_inv)
         scale = scale.clamp(min=torch.finfo(scale.dtype).tiny)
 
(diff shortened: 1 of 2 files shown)


tests/infer/reparam/test_stable.py
```diff
@@ -55,7 +58,7 @@ def model():
         expected_grads = grad(expected_m.sum(), params, retain_graph=True)
         actual_grads = grad(actual_m.sum(), params, retain_graph=True)
         assert_close(actual_grads[0], expected_grads[0], atol=0.2)
-        assert_close(actual_grads[1], expected_grads[1], atol=0.1)
+        assert_close(actual_grads[1][skew != 0], expected_grads[1][skew != 0], atol=0.1)
```
fehiepsi: this test is unstable when skew ~ 0 (say skew < 1e-3). Here we only want to make sure that nan does not happen.

pyro/infer/reparam/stable.py
```diff
@@ -179,7 +179,7 @@ def __call__(self, name, fn, obs):
         t = _standard_stable(a, one, tu, te, coords="S0")
         a_inv = a.reciprocal()
         skew_abs = fn.skew.abs()
-        t_scale = skew_abs.pow(a_inv)
+        t_scale = skew_abs.clamp(min=torch.finfo(a.dtype).tiny).pow(a_inv)
         s_scale = (1 - skew_abs).pow(a_inv)
```
fritzo: Maybe we should clamp both of these to `.clamp(min=torch.finfo(a.dtype).eps)`?

fehiepsi: Yes, we can clamp so to prevent the edge case `skew=+-1`. `eps` should be fine too. I was just worried about `grad` at skew=0. Current version returns `grad(skew) = 0.`, which is far from the expected values.

pyro/infer/reparam/stable.py
```diff
@@ -191,18 +192,18 @@ def __call__(self, name, fn, obs):
         return new_fn, obs
 
 
-def _unsafe_shift(a, skew_abs, t_scale):
+def _unsafe_shift(a, skew, t_scale):
     # At a=1 the lhs has a root and the rhs has an asymptote.
-    return (t_scale - skew_abs) * (math.pi / 2 * a).tan()
+    return (skew.sign() * t_scale - skew) * (math.pi / 2 * a).tan()
```
fehiepsi: I want to keep the term `skew * tan(a pi / 2)` to preserve grad w.r.t. skew, so I multiplied `skew.sign()` to the previous shift term.

fritzo:
@fehiepsi I ran a few of the notebooks and looks like this might even improve fit a tiny bit. Thanks for addressing!

PR #2363, opened 2020-03-10
Support batched forecasting

fritzo:
This adds support for drawing many samples in batches from `Forecaster.__call__()`.

The motivating use case is large models and short forecasts, where models may be trained using subsampling but where the `O(num_samples * full_data)` is too large to fit in memory for large `num_samples` (as e.g. in M5 forecasts). In this case we can sequentially draw say 10 samples at a time to generate very precise quantiles.

This also cleans up docs a bit.

Note I needed to move the `torch.no_grad()` context manager for HMC tests to pass.

## Tested
- [x] added batched call to unit tests

### DIFF
--- a/docs/source/contrib.forecast.rst
+++ b/docs/source/contrib.forecast.rst
@@ -47,7 +47,7 @@ Forecaster Interface
 ---------------------
 .. automodule:: pyro.contrib.forecast.forecaster
     :members:
-    :undoc-members:
+    :special-members: __call__
     :show-inheritance:
     :member-order: bysource
 
--- a/pyro/contrib/forecast/forecaster.py
+++ b/pyro/contrib/forecast/forecaster.py
@@ -281,27 +281,61 @@ class Forecaster(nn.Module):
         self.max_plate_nesting = elbo.max_plate_nesting
         self.losses = losses
 
-    @torch.no_grad()
-    def forward(self, data, covariates, num_samples):
+    def __call__(self, data, covariates, num_samples, batch_size=None):
+        """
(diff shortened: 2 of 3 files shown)


docs/source/contrib.forecast.rst
```diff
@@ -47,7 +47,7 @@ Forecaster Interface
 ---------------------
 .. automodule:: pyro.contrib.forecast.forecaster
     :members:
-    :undoc-members:
+    :special-members: __call__
```
fritzo: This hides the internal `.forward()` and exposes the public `.__call__()`.

pyro/contrib/forecast/forecaster.py
```diff
@@ -354,20 +389,54 @@ def __init__(self, model, data, covariates=None, *,
...
+        :param data: A tensor dataset with time dimension -2.
+        :type data: ~torch.Tensor
+        :param covariates: A tensor of covariates with time dimension -2.
+            For models not using covariates, pass a shaped empty tensor
+            ``torch.empty(duration, 0)``.
+        :type covariates: ~torch.Tensor
+        :param int num_samples: The number of samples to generate.
+        :param int batch_size: Optional batch size for sampling. This is useful
+            for generating many samples from models with large memory
+            footprint. By defaults to ``num_samples``.
```
martinjankowiak: ditto

pyro/contrib/forecast/forecaster.py
```diff
@@ -281,27 +281,61 @@ def __init__(self, model, data, covariates, *,
...
```
martinjankowiak: typo: By defaults to ``num_samples``

PR #2365, opened 2020-03-13
Fix LocScaleReparam, log params in backtest()

fritzo:
1. Fixes a naming bug in `LocScaleReparam` whereby all loc-scale-reparameterized sites shared a single centeredness parameter. After this PR each such site learns a separate centeredness parameter (as originally intended).

2. Also adds param logging to `contrib.forecast.evaluate.backtest()`, which logging led me to discover the loc-scale bug. After this PR all scalar params will be printed and saved (as floats) to backtest results. My immediate motivation is to inspect global parameters `stability`, `skew` and `df` of `Stable` and `StudentT` distributions.

## Tested
- [x] covered by existing tests
- [x] ran locally and examined logging output

### DIFF
--- a/pyro/contrib/forecast/evaluate.py
+++ b/pyro/contrib/forecast/evaluate.py
@@ -200,12 +200,19 @@ def backtest(data, covariates, model_fn, *,
             "num_samples": num_samples,
             "train_walltime": train_walltime,
             "test_walltime": test_walltime,
+            "params": {},
         }
         results.append(result)
         for name, fn in metrics.items():
             result[name] = fn(pred, truth)
-            if isinstance(result[name], (int, float)):
-                logger.debug("{} = {}".format(name, result[name]))
+        for name, value in pyro.get_param_store().items():
+            if value.numel() == 1:
+                value = value.cpu().item()
+                result["params"][name] = value
+        for dct in (result, result["params"]):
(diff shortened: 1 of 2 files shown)


pyro/contrib/forecast/evaluate.py
```diff
@@ -200,12 +200,19 @@ def forecaster_options_fn(*args, **kwargs):
...
             "test_walltime": test_walltime,
+            "params": {},
         }
         results.append(result)
         for name, fn in metrics.items():
             result[name] = fn(pred, truth)
-            if isinstance(result[name], (int, float)):
-                logger.debug("{} = {}".format(name, result[name]))
+        for name, value in pyro.get_param_store().items():
+            if value.numel() == 1:
```
martinjankowiak: make the numel boundary toggleable so that users have an easy way to inspect params? e.g. they can even choose inf

fritzo: I'm trying to avoid logging any `Tensor`s, instead logging only Python objects. One reason for this design choice is that a single big tensor renders the entire data structure unreadable (visually, at a terminal). One option that would still satisfy this design constraint is to log scalar statistics of each non-scalar tensor, e.g. min, max, median, mean. That seems like a reasonable feature for a follow-up PR.

PR #2323, opened 2020-02-23
Support subsampling in AutoDelta, AutoNormal

fritzo:
Resolves #2322
Addresses #2311 

This supports subsampling by adding a `create_plates()` function to `AutoDelta.__init__()` and `AutoNormal.__init__()`. For example with origin-destination-time data:
```py
def model(odt_data, subsample_origins=None, subsample_destins=None):
    # ...user defined...

def create_plates(odt_data, subsample_origins=None, subsample_destins=None):
    O, D, T = odt_data.shape
    return [pyro.plate("origin", size=O, subsample_size=subsample_origins, dim=-3),
            pyro.plate("destin", size=D, subsample_size=subsample_destins, dim=-2)]

guide = AutoNormal(model, create_plates=create_plates)
```

This also includes the following fixes:
-  introduces an experimental `pyro.subsample()` primitive (pair coded with @eb8680)
- changes `dist.Uniform` to support unbroadcasted params for broadcastable `.support` shape

## Tested
- [x] `dist.Uniform` refactoring is covered by existing tests
- [x] tests for `pyro.subsample()`
- [x] `AutoDelta` tests
- [x] `AutoNormal` tests
- [x] forecaster tests

PR #2370, opened 2020-03-15
Support replay of plate(..., subsample=...)

fritzo:
Follows up #2323 by allowing `pyro.plate(..., subsample=...)` syntax in the `create_plates()` function used in autoguides and the `contrib.forecast` module.

The motivation is the [hierarchical bart forecasting tutorial](https://pyro.ai/examples/forecasting_iii.html#Subsampling), which I'd like to switch to dependent subsampling as in the unit test in this PR; I'll wait until after this PR merges.

This will introduce a small amount of overhead in that some new `_Subsample` sites will appear in traces, but those appear to be needed so as to support trace-replay behavior. I believe this should not add any tensor op overhead. In fact I believe this removes a `torch.arange()` in some cases, since the same `arange` can be used by the guide and model. This also simplifies the non-subsample case by replacing our old `jit_compatible_arange()` with a cheaper `torch.arange()`.

## Tested
- [x] added a regression test
- [x] tested locally on a model that wasn't working before this PR

### DIFF
--- a/pyro/poutine/subsample_messenger.py
+++ b/pyro/poutine/subsample_messenger.py
@@ -5,7 +5,7 @@ import torch
 
 from pyro.distributions.distribution import Distribution
 from pyro.poutine.util import is_validation_enabled
-from pyro.util import ignore_jit_warnings, jit_compatible_arange
+from pyro.util import ignore_jit_warnings
 
 from .indep_messenger import CondIndepStackFrame, IndepMessenger
 from .runtime import apply_stack
@@ -47,7 +47,7 @@ class _Subsample(Distribution):
             raise NotImplementedError
         subsample_size = self.subsample_size
         if subsample_size is None or subsample_size >= self.size:
-            result = jit_compatible_arange(self.size, device=self.device)
+            result = torch.arange(self.size, device=self.device)
         else:
             result = torch.multinomial(torch.ones(self.size), self.subsample_size,
                                        replacement=False).to(self.device)
@@ -87,7 +87,7 @@ class SubsampleMessenger(IndepMessenger):
             assert subsample is None
             size = -1  # This is PyTorch convention for "arbitrary size"
             subsample_size = -1
-        elif subsample is None:
+        else:
             msg = {
                 "type": "sample",
                 "name": name,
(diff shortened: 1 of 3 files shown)


pyro/poutine/subsample_messenger.py
```diff
@@ -87,15 +87,15 @@ def _subsample(name, size=None, subsample_size=None, subsample=None, use_cuda=No
...
+        else:
             msg = {
                 "type": "sample",
                 "name": name,
                 "fn": _Subsample(size, subsample_size, use_cuda, device),
                 "is_observed": False,
                 "args": (),
                 "kwargs": {},
-                "value": None,
+                "value": subsample,
```
fritzo: Before this PR, if `subsample` was specified in the guide, no `_Subsample` site would be created, and hence the subsampling could not be traced and replayed. After this PR a `_Subsample` site is created whenever `size` is specified (and hence whenever subsampling is possible).

pyro/poutine/subsample_messenger.py
```diff
@@ -47,7 +47,7 @@ def sample(self, sample_shape=torch.Size()):
             raise NotImplementedError
         subsample_size = self.subsample_size
         if subsample_size is None or subsample_size >= self.size:
-            result = jit_compatible_arange(self.size, device=self.device)
+            result = torch.arange(self.size, device=self.device)
```
fritzo: I've also removed our obsolete helper `jit_compatible_arange`, saving a few tensor ops in the most common case.

tests/infer/test_autoguide.py
```diff
@@ -745,3 +745,36 @@ def create_plates(batch, subsample, full_size):
...
+    def create_plates(data):
+        size, size = data.shape
+        origin_plate = pyro.plate("origin", size, subsample_size=5, dim=-2)
+        if independent:
+            destin_plate = pyro.plate("destin", size, subsample_size=5, dim=-1)
+        else:
+            with origin_plate as subsample:
+                pass
+            destin_plate = pyro.plate("destin", size, subsample=subsample, dim=-1)
+        return origin_plate, destin_plate
```
fritzo: Before this PR, only the independent case works. After, the dependent case also works, using `pyro.plate(..., subsample=...)`.

fritzo:
> do you want to document this use case in the `plate` or `subsample` docstrings?

This use case is quite involved. I think the right place for documentation is in tutorials. I was planning on adding some new explanation to the [subsampling section](https://pyro.ai/examples/forecasting_iii.html#Subsampling) of the forecasting tutorial when I update that model. I guess we could also use an explanation in an autoguide tutorial...

PR #2374, opened 2020-03-17
Address numerical stability of Geometric log_prob

fehiepsi:
This PR computes the log_prob of Geometric distribution based on `logits` instead of `probs`. Currently, the log_prob is defined as follows
```python
def log_prob(self, value):
    ...
    return value * (-probs).log1p() + self.probs.log()
```
which will lead to `nan` when `probs=0` or `probs=1`.

@fritzo I also add Geometric to forecast univariate dist. This likelihood is quite stable for many univariate count series (with a lot of zeros) that I tested on.

### DIFF
--- a/pyro/distributions/torch.py
+++ b/pyro/distributions/torch.py
@@ -97,6 +97,7 @@ class MultivariateNormal(torch.distributions.MultivariateNormal, TorchDistributi
 
 
 class Geometric(torch.distributions.Geometric, TorchDistributionMixin):
+    # TODO: move upstream
     def log_prob(self, value):
         if self._validate_args:
             self._validate_sample(value)
--- a/pyro/distributions/torch.py
+++ b/pyro/distributions/torch.py
@@ -3,6 +3,7 @@
 
 import torch
 from torch.distributions import constraints
+from torch.distributions.utils import lazy_property
 
 from pyro.distributions.constraints import IndependentConstraint
 from pyro.distributions.torch_distribution import TorchDistributionMixin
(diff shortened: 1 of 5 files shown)


tests/distributions/conftest.py
```diff
@@ -427,6 +427,23 @@ def __init__(self, loc, scale):
             scipy_arg_fn=lambda rate: ((np.array(rate),), {}),
             prec=0.08,
             is_discrete=True),
+    Fixture(pyro_dist=dist.Geometric,
+            scipy_dist=sp.geom,
+            examples=[
+                {'logits': [2.0],
```
neerajprad: Can we also directly test with `probs=0.` and `probs=1.` or would that cause issues with scipy?

fritzo:
Can you also add to the `DISTS` list in [test/contrib/forecast/test_util.py](https://github.com/pyro-ppl/pyro/blob/dev/tests/contrib/forecast/test_util.py#L12)?

tests/distributions/conftest.py
fehiepsi: I haven't tested but I guess it will return `-inf` in scipy and a negative value ~-20 in torch (which is similar to Bernoulli or other distributions with logits). Let me verify...

pyro/distributions/torch.py
```diff
@@ -96,12 +95,13 @@ def _log_normalizer(d):
 class MultivariateNormal(torch.distributions.MultivariateNormal, TorchDistributionMixin):
     support = IndependentConstraint(constraints.real, 1)  # TODO move upstream
 
-    # TODO: remove this in the PyTorch release > 1.4.0
```
fehiepsi: opss, PyTorch newest release is still 1.4.0. Will revert back this...

fritzo:
Sounds reasonable. Another option to numerically stabilize `Poisson` models is to use a custom `init_loc_fn`:
```py
def init_loc_fn(site):
    if site["name"] == "rate":  # or whatever
        return count_data.clamp(min=0.1)
    return itit_to_sample(site)  # fall back to default

forecaster = Forecaster(..., init_loc_fn=init_loc_fn)
```

tests/distributions/conftest.py
fehiepsi: @neerajprad Geometric will raise an error if `probs=0` in its construction. For `probs=1`, scipy returns -inf while Geometric returns a small negative number (in log scale). This is also the case of `torch.distributions.Bernoulli(probs=torch.tensor(0.)).log_prob(1.)`.

neerajprad: Sounds good, thanks for checking. LGTM.

PR #2376, opened 2020-03-19
Improve the performance of rsample of GaussianHMM with Normal obs

### LINKED ISSUE #2375: O(obs_dim) algorithms for GaussianHMM methods
@martinjankowiak is trying to forecast with a `GaussianHMM` likelihood with large `obs_dim` and small `hidden_dim` however the time complexity is awful for forecasting (it's fine for training). We could solve this by implementing `AffineNormal.event_permute()`, `.condition()` and `.rsample()` as  used in `GaussianHMM.rsample()` [code](https://github.com/pyro-ppl/pyro/blob/fca6c9546e57b31e5d327ecb7fd9f6340ac3dc66/pyro/distributions/hmm.py#L514).

When I implemented those `AffineNormal` method I took a correct but expensive shortcut and simply converted to a `Gaussian` and now we need to pay off that technical debt e.g. [here](https://github.com/pyro-ppl/pyro/blob/fca6c9546e57b31e5d327ecb7fd9f6340ac3dc66/pyro/ops/gaussian.py#L302).

We should test the solution with a local test (or a `@pytest.mark.skip`'d test) with say `obs_dim=5000`, `hidden_dim=1` to make sure all relevant ops are cheap. It's tough to determine just by reading code.

fehiepsi:
Resolves #2375.

Test for GaussianHMM with duration=10, obs=1000, we get 0.56s against 4.84s as before.

With duration=1000, obs=5000, it took 24s in my system. I can't test for the old implementation because of OOM.

TODO:
- [x] make sure all tests pass, test `left_condition`
- [x] profiling on a Gaussian with large obs_dim

@martinjankowiak Could you try on your problem? I believe this would be much faster than before.

### DIFF
--- a/pyro/ops/gaussian.py
+++ b/pyro/ops/gaussian.py
@@ -308,10 +308,20 @@ class AffineNormal:
         """
         if value.size(-1) == self.matrix.size(-2):
             loc = matvecmul(self.matrix.transpose(-1, -2), value) + self.loc
-            return torch.distributions.Independent(
-                torch.distributions.Normal(loc, scale=self.scale), 1)
+            matrix = value.new_zeros(loc.shape[:-1] + (0, loc.size(-1)))
+            scale = self.scale.expand(loc.shape)
+            return AffineNormal(matrix, loc, scale)
         else:
             return self.to_gaussian().left_condition(value)
+
+    def rsample(self, sample_shape=torch.Size()):
+        """
+        Reparameterized sampler.
+        """
+        assert self.matrix.size(-2) == 0
+        shape = sample_shape + self.batch_shape + self.loc.shape[-1:]
+        noise = torch.randn(shape, dtype=self.loc.dtype, device=self.loc.device)
+        return self.loc + noise * self.scale
 
     def to_gaussian(self):
         if self._gaussian is None:
--- a/pyro/ops/gaussian.py
+++ b/pyro/ops/gaussian.py
@@ -7,7 +7,6 @@ import torch
 from torch.distributions.utils import lazy_property
 from torch.nn.functional import pad
 
-import pyro.distributions as dist
 from pyro.distributions.util import broadcast_shape
 from pyro.ops.tensor_utils import cholesky, matmul, matvecmul, triangular_solve
 
@@ -302,15 +301,22 @@ class AffineNormal:
             return self.to_gaussian().condition(value)
 
     def left_condition(self, value):
+        """
+        If ``value.size(-1) == x_dim``, this returns a Normal distribution with
+        ``event_dim=1``. After applying this method, the cost to draw a sample is
+        ``O(y_dim)`` instead of ``O(y_dim ** 3)``.
+        """
         if value.size(-1) == self.matrix.size(-2):
             loc = matvecmul(self.matrix.transpose(-1, -2), value) + self.loc
-            return dist.Normal(loc, self.scale).to_event(1)
+            return torch.distributions.Independent(
+                torch.distributions.Normal(loc, scale=self.scale), 1)
         else:
             return self.to_gaussian().left_condition(value)
 
     def to_gaussian(self):
         if self._gaussian is None:
-            mvn = dist.Normal(self.loc, scale=self.scale).to_event(1)
+            mvn = torch.distributions.Independent(
+                torch.distributions.Normal(self.loc, scale=self.scale), 1)
             y_gaussian = mvn_to_gaussian(mvn)
             self._gaussian = _matrix_and_gaussian_to_gaussian(self.matrix, y_gaussian)
         return self._gaussian
--- a/pyro/ops/gaussian.py
+++ b/pyro/ops/gaussian.py
@@ -318,7 +318,8 @@ class AffineNormal:
         """
         Reparameterized sampler.
         """
-        assert self.matrix.size(-2) == 0
+        if self.matrix.size(-2) > 0:
+            raise NotImplementedError
         shape = sample_shape + self.batch_shape + self.loc.shape[-1:]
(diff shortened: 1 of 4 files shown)


tests/distributions/test_hmm.py
```diff
@@ -334,6 +335,24 @@ def test_gaussian_hmm_shape(diag, init_shape, trans_mat_shape, trans_mvn_shape,
         assert d2.event_shape == (f, obs_dim)
 
 
+def test_gaussian_hmm_high_obs_dim():
+    hidden_dim = 1
+    obs_dim = 1000
+    duration = 10
```
fehiepsi: Those are limits in my system to be able to run the test with the old implementation. With the new implementation in this PR, we don't have OOM issue.

pyro/ops/gaussian.py
```diff
@@ -251,33 +277,49 @@ class AffineNormal:
...
-        precision = matmul(prec_sqrt, prec_sqrt.transpose(-1, -2))
-        delta = (value - self.loc) / self.scale
-        info_vec = matvecmul(prec_sqrt, delta)
-        log_normalizer = (-0.5 * self.loc.size(-1) * math.log(2 * math.pi)
-                          - 0.5 * delta.pow(2).sum(-1) - self.scale.log().sum(-1))
-        return Gaussian(log_normalizer, info_vec, precision)
+        if value.size(-1) == self.matrix.size(-2):
+            loc = matvecmul(self.matrix.transpose(-1, -2), value) + self.loc
+            return torch.distributions.Independent(
+                torch.distributions.Normal(loc, scale=self.scale), 1)
```
fehiepsi: This makes the output a bit inconsistent. Should we merge `rsample` with `left_condition`, e.g. `rsample(sample_shape=..., left_condition=None)`? I am open to any suggestions for having a consistent output.

pyro/ops/gaussian.py
```diff
@@ -251,33 +277,49 @@ class AffineNormal:
...
+            precision = matmul(prec_sqrt, prec_sqrt.transpose(-1, -2))
+            delta = (value - self.loc) / self.scale
+            info_vec = matvecmul(prec_sqrt, delta)
+            log_normalizer = (-0.5 * self.loc.size(-1) * math.log(2 * math.pi)
+                              - 0.5 * delta.pow(2).sum(-1) - self.scale.log().sum(-1))
+            return Gaussian(log_normalizer, info_vec, precision)
+        else:
+            return self.to_gaussian().condition(value)
+
+    def left_condition(self, value):
```
fehiepsi: I made this method because `event_permute` does not preserve a cheap form of AffineNormal. More precisely, given y = x @ M + mvn, we can't represent `x` as a noisy affine of `y` unless M is invertible.

martinjankowiak:
@fehiepsi i can confirm i got a 5x speed-up for my problem with `obs_dim=1440` 👍

pyro/ops/gaussian.py
fritzo: I see, the output type depends on input dims...

I think it's fine to implement a fused method `.left_condition_rsample(value, sample_shape=torch.Size())`.

Does the following alternative make sense: we could return a `Gaussian` here with event_dim=1 and an extra batch dimension, then update `Gaussian.rsample()` to handle the special case of `evet_dim=1`, as you did with Funsor. I guess in this case the type would be consistent but the `batch_shape` would be inconsistent?

fehiepsi: I think we can implement a fused method. Supporting `event_dim=1` in Gaussian would require a lot of ops. On second thought, I think we can return an AffineNormal here and add a `rsample` method to `AffineNormal`. Like Gaussian, which only supports .rsample for full rank precision, here we support `.rsample` for empty matrix. WDYT?

fritzo: sounds reasonable, would `.matrix is None` correspond to an identity matrix?

fehiepsi: It is just simply a zeros/empty matrix with 0 size: `.matrix = torch.zeros(0, loc.size(-1))`.

pyro/ops/gaussian.py
```diff
@@ -251,33 +277,59 @@ class AffineNormal:
...
-        prec_sqrt = self.matrix / self.scale.unsqueeze(-2)
-        precision = matmul(prec_sqrt, prec_sqrt.transpose(-1, -2))
-        delta = (value - self.loc) / self.scale
-        info_vec = matvecmul(prec_sqrt, delta)
-        log_normalizer = (-0.5 * self.loc.size(-1) * math.log(2 * math.pi)
-                          - 0.5 * delta.pow(2).sum(-1) - self.scale.log().sum(-1))
-        return Gaussian(log_normalizer, info_vec, precision)
+        Reparameterized sampler.
+        """
+        assert self.matrix.size(-2) == 0
```
fritzo: nit:
```py
if matrix.size(-2) > 0:
    raise NotImplementedError
```

pyro/ops/gaussian.py
```diff
@@ -305,7 +357,15 @@ def __add__(self, other):
         return self.to_gaussian() + other
 
     def marginalize(self, left=0, right=0):
-        return self.to_gaussian().marginalize(left, right)
+        assert left == 0
```
fritzo: nit:
```py
if left > 0:
    raise NotImplementedError
```

pyro/ops/gaussian.py
```diff
@@ -305,7 +357,15 @@ def __add__(self, other):
...
+        if right == self.loc.size(-1):
```
fritzo: wait, can we replace the above assertion with
```py
if left == 0 and right == self.loc.size(-1):
    n = ...
```

fehiepsi:
Thank @martinjankowiak and @fritzo for verifying and comments!

PR #2377, opened 2020-03-20
Fix bugs at reshape_batch LinearHMM

fehiepsi:
This fixes several issues at reshape_batch(LinearHMM):

+ In LinearHMM, we always expand init_dist, trans_dist, obs_dist; so when reshaping, we should also expand for consistency. This fixes a bug where `reshape_batch(hmm, (1,)).rsample().shape[0] != 1`.
+ Keep the `.transforms` after reshaping.
+ Workaround the case LinearHMM's observation_dist is LogNormal: `self.observation_dist` will be `torch.distributions.Normal`, which is not registered in forecast's UNIVARIATE_DIST.

### DIFF
--- a/pyro/contrib/forecast/util.py
+++ b/pyro/contrib/forecast/util.py
@@ -326,13 +326,54 @@ def _(d, batch_shape):
     new.transition_dist = trans_dist
     new.observation_matrix = obs_mat
     new.observation_dist = obs_dist
+    transforms = []
     for transform in d.transforms:
-        assert isinstance(transform, (transforms.AbsTransform,
-                                      transforms.ExpTransform,
-                                      transforms.SigmoidTransform)), \
+        assert type(transform) in UNIVARIATE_TRANSFORMS, \
             "Currently, reshape_batch only supports AbsTransform, " + \
             "ExpTransform, SigmoidTransform transform"
-    new.transforms = d.transforms
+        current_shape = d.observation_dist.shape()
+        new_shape = obs_dist.shape()
+        transforms.append(reshape_transform_batch(transform, current_shape, new_shape))
+    new.transforms = transforms
     super(dist.LinearHMM, new).__init__(d.duration, batch_shape, d.event_shape,
                                         validate_args=d._validate_args)
     return new
+
+
+UNIVARIATE_TRANSFORMS = {
+    transforms.AbsTransform: (),
+    transforms.AffineTransform: ("loc", "scale"),
+    transforms.ExpTransform: (),
+    transforms.PowerTransform: ("exponent",),
+    transforms.SigmoidTransform: (),
+}
+
+
+@singledispatch
+def reshape_transform_batch(t, current_batch_shape, new_batch_shape):
+    """
+    EXPERIMENTAL Given a transform ``t``, reshape to different batch shape
+    of same number of elements.
+
+    This is typically used to correct the transform parameters' shapes after
+    reshaping the base distribution of a transformed distribution.
+
+    :param t: A transform.
+    :type t: ~torch.distributions.transforms.Transform
+    :param tuple current_batch_shape: The current batch shape.
+    :param tuple new_batch_shape: A new batch shape.
+    :returns: A transform with the same type but given new batch shape.
+    :rtype: ~torch.distributions.transforms.Transform
+    """
+    raise NotImplementedError("reshape_transform_batch() does not suport {}".format(type(t)))
+
+
+def _reshape_batch_univariate_transform(t, current_batch_shape, new_batch_shape):
+    params = {name: getattr(t, name).expand(current_batch_shape).reshape(new_batch_shape)
+              for name in UNIVARIATE_TRANSFORMS[type(t)]}
+    params["cache_size"] = t._cache_size
+    return type(t)(**params)
+
+
+for _type in UNIVARIATE_TRANSFORMS:
+    reshape_transform_batch.register(_type)(_reshape_batch_univariate_transform)
--- a/pyro/contrib/forecast/util.py
+++ b/pyro/contrib/forecast/util.py
@@ -4,7 +4,7 @@
 from functools import singledispatch
 
 import torch
-from torch.distributions import transform_to
+from torch.distributions import transform_to, transforms
 
 import pyro.distributions as dist
 from pyro.poutine.messenger import Messenger
@@ -326,6 +326,12 @@ def _(d, batch_shape):
     new.transition_dist = trans_dist
     new.observation_matrix = obs_mat
     new.observation_dist = obs_dist
+    for transform in d.transforms:
+        assert isinstance(transform, (transforms.AbsTransform,
+                                      transforms.ExpTransform,
+                                      transforms.SigmoidTransform)), \
+            "Currently, reshape_batch only supports AbsTransform, " + \
+            "ExpTransform, SigmoidTransform transform"
     new.transforms = d.transforms
     super(dist.LinearHMM, new).__init__(d.duration, batch_shape, d.event_shape,
                                         validate_args=d._validate_args)
--- a/pyro/contrib/forecast/util.py
+++ b/pyro/contrib/forecast/util.py
@@ -312,30 +312,21 @@ def _(d, batch_shape):
 
 @reshape_batch.register(dist.LinearHMM)
 def _(d, batch_shape):
-    init_dist = d.initial_dist
-    if init_dist.batch_shape:
-        init_dist = init_dist.expand(d.batch_shape)
-        init_dist = reshape_batch(init_dist, batch_shape)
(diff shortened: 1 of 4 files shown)


pyro/distributions/hmm.py
```diff
@@ -942,6 +942,9 @@ def __init__(self, initial_dist, transition_matrix, transition_dist,
             else:
                 break
         if not observation_dist.event_shape:
+            if isinstance(observation_dist, torch.distributions.Normal) and \
+                    not isinstance(observation_dist, TorchDistributionMixin):
+                observation_dist = Normal(observation_dist.loc, observation_dist.scale)
```
fehiepsi: workaround observation_dist is LogNormal

pyro/distributions/hmm.py
```diff
@@ -980,7 +983,8 @@ def log_prob(self, value):
 
     def rsample(self, sample_shape=torch.Size()):
         assert self.duration is not None
-        init = self.initial_dist.rsample(sample_shape)
+        # XXX we might not need to expand here because we already expand eagerly in the construction
+        init = self.initial_dist.expand(self.batch_shape).rsample(sample_shape)
```
fehiepsi: should we remove `expand` here and the following ones?

tests/contrib/forecast/test_util.py
```diff
@@ -90,8 +91,7 @@ def test_reshape_batch(Dist, batch_shape, duration, dim):
 
     actual = reshape_batch(d, batch_shape + (1,))
     assert type(actual) is type(d)
-    if batch_shape:
-        assert actual.batch_shape == batch_shape + (1,)
-    else:
-        assert actual.batch_shape in ((), batch_shape + (1,))
```
fehiepsi: It seems to me that only LinearHMM is intended to have `actual.batch_shape == ()`. Is there a reason behind?

pyro/distributions/hmm.py
fritzo: Can you add that as a comment?

We should really fix `pyro.distributions.torch.LogNormal`. Another option is to force:
```py
if type(observation_dist) is torch.distributions.Normal):
    observation_dist.__class__ = dist.Normal
```

fehiepsi: Sounds reasonable, I'll do that.

pyro/distributions/hmm.py
fritzo: Sure, but could you add assertions in case the `__init__()` method later chages?
```py
assert self.initial_dist.batch_shape == self.batch_shape
assert self.transition_dist.batch_shape == self.batch_shape + (self.duration,)
assert self.observation_dist.batch_shape == self.batch_shape + (self.duration,)
```

pyro/contrib/forecast/util.py
```diff
@@ -312,30 +312,21 @@ def _(d, batch_shape):
...
+
+    new = d._get_checked_instance(dist.LinearHMM)
+    new.hidden_dim = d.hidden_dim
+    new.obs_dim = d.obs_dim
+    new.initial_dist = init_dist
+    new.transition_matrix = trans_mat
+    new.transition_dist = trans_dist
+    new.observation_matrix = obs_mat
+    new.observation_dist = obs_dist
+    new.transforms = d.transforms
```
fritzo: Can you either add an assertion here that the transforms are not batched, or add some code to `reshape_batch` transform objects? I suspect we're missing a bunch of low-level helpers like `Transform.batch_shape`, so it's fine to punt and add a could commonly used `@singledispatch` patterns to this file, e.g. for `LogTransform`.

pyro/distributions/torch.py
```diff
@@ -101,6 +101,17 @@ def log_prob(self, value):
         return (-value - 1) * torch.nn.functional.softplus(self.logits) + self.logits
 
 
+class LogNormal(torch.distributions.LogNormal, TorchDistributionMixin):
+    def __init__(self, loc, scale, validate_args=None):
+        base_dist = Normal(loc, scale)
```
fritzo: Can you add a comment that

```py
# This differs from torch.distributions.LogNormal only in that base_dist is
# a pyro.distributions.Normal rather than a torch.distributions.Normal.
```

pyro/contrib/forecast/util.py
```diff
@@ -312,30 +312,21 @@ def _(d, batch_shape):
...
-
-    return dist.LinearHMM(init_dist, trans_mat, trans_dist, obs_mat, obs_dist,
-                          duration=d.duration)
+    init_dist = reshape_batch(d.initial_dist, batch_shape)
+    trans_mat = d.transition_matrix.reshape(batch_shape + (-1, d.hidden_dim, d.hidden_dim))
+    trans_dist = reshape_batch(d.transition_dist, batch_shape + (-1,))
+    obs_mat = d.observation_matrix.reshape(batch_shape + (-1, d.hidden_dim, d.obs_dim))
+    obs_dist = reshape_batch(d.observation_dist, batch_shape + (-1,))
+
+    new = d._get_checked_instance(dist.LinearHMM)
```
fritzo: What is the reason for using `_get_checked_instance()` rather than a simple `LinearHMM(...)` constructor?

pyro/distributions/torch.py
```diff
@@ -112,6 +123,10 @@ def precision_matrix(self):
             self._batch_shape + self._event_shape + self._event_shape)
 
 
+class Normal(torch.distributions.Normal, TorchDistributionMixin):
```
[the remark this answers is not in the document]
fehiepsi: Yes.

pyro/contrib/forecast/util.py
fehiepsi: The constructor changes the meaning of `observation_dist` so I think using `_get_checked_instance` will simplify pieces of stuff related to `obs_dist` and `transform`. In addition, this is a bit cheaper and more aligned with `expand` method and other `reshape_batch` code.

pyro/contrib/forecast/util.py
fehiepsi: I think for ExpTransform, SigmoidTransform, AbsTransform, it is fine here. Do you want to support AffineTransform and PowerTransform?

fehiepsi:
@fritzo , sorry for the late response. Thanks for your comments and suggestions!

About reshape_batch of transforms, let's support AffineTransform or other ones if needed. As you mentioned, we just need to use @singledispatch to reshape the fields `loc`/`transform`/`exponent` from them.

pyro/distributions/hmm.py
fehiepsi: Sorry for the confusion, I have reverted the changes. `expand` is required in the case: those trans_dist, obs_dist has time_dim=1, but duration > 1.

pyro/contrib/forecast/util.py
fritzo: `PowerTransform` would be nice. I see no reason to add `AffineTransform`, but we might as well since it's so easy.

fehiepsi: I'll add support for them.

fehiepsi:
@fritzo I just add reshape pattern for univariate transforms. Does it align with what you had in mind?

pyro/distributions/torch.py
```diff
@@ -101,6 +101,19 @@ def log_prob(self, value):
...
+    def __init__(self, loc, scale, validate_args=None):
+        base_dist = Normal(loc, scale)
+        # This differs from torch.distributions.LogNormal only in that base_dist is
+        # a pyro.distributions.Normal rather than a torch.distributions.Normal.
+        super(torch.distributions.LogNormal, self).__init__(
+            base_dist, torch.distributions.transforms.ExpTransform(), validate_args=validate_args)
+
+    def expand(self, batch_shape, _instance=None):
+        new = self._get_checked_instance(LogNormal, _instance)
+        return super(torch.distributions.LogNormal, self).expand(batch_shape, _instance=new)
```
fritzo, line 114: Can you confirm this results in a `.base_dist` of the correct `Normal` type?

fehiepsi: Yes, it is
```python
import torch
from pyro.distributions.torch import Normal, TorchDistributionMixin

class LogNormal(torch.distributions.LogNormal, TorchDistributionMixin):
    def __init__(self, loc, scale, validate_args=None):
        base_dist = Normal(loc, scale)
        super(torch.distributions.LogNormal, self).__init__(
            base_dist, torch.distributions.transforms.ExpTransform(), validate_args=validate_args)

        new = self._get_checked_instance(LogNormal, _instance)
        return super(torch.distributions.LogNormal, self).expand(batch_shape, _instance=new)

x = LogNormal(0, 1).expand([3])
assert repr(type(x.base_dist)) == "<class 'pyro.distributions.torch.Normal'>"
```

PR #2379, opened 2020-03-22
Update links

mollerhoj:


pyro/contrib/minipyro.py
```diff
@@ -26,9 +26,9 @@
 # Pyro keeps track of two kinds of global state:
 # i)  The effect handler stack, which enables non-standard interpretations of
 #     Pyro primitives like sample();
-#     See http://docs.pyro.ai/en/0.3.1/poutine.html
+#     See http://docs.pyro.ai/en/1.3.0/poutine.html
```
neerajprad: Could you change these links to point to `stable` instead so that we don't have to update it after every release? e.g. http://docs.pyro.ai/en/stable/parameters.html

mollerhoj: I considered that, but the links might not exist in future versions. I've seen this problem before in documentation. This at least points to something independent of what happens in the future. Do you agree?

fritzo: I think `stable` is safest. If we wanted full reproducibility it would be best to point to a file that existed in git. But even the pyro.ai domain isn't guaranteed to exist. `stable` seems like a good compromise for maintainability.

mollerhoj:
Should be good to go now :-)

PR #2381, opened 2020-03-24
Conditional affine coupling layer

stefanwebb:
The title says it all

### DIFF
--- a/pyro/distributions/transforms/affine_coupling.py
+++ b/pyro/distributions/transforms/affine_coupling.py
@@ -177,12 +177,12 @@ class ConditionalAffineCoupling(ConditionalTransformModule):
     >>> base_dist = dist.Normal(torch.zeros(input_dim), torch.ones(input_dim))
     >>> param_dims = [input_dim-split_dim, input_dim-split_dim]
     >>> hypernet = ConditionalDenseNN(split_dim, context_dim, [10*input_dim],
-    >>> param_dims)
+    ... param_dims)
     >>> transform = ConditionalAffineCoupling(split_dim, hypernet)
     >>> pyro.module("my_transform", transform)  # doctest: +SKIP
     >>> z = torch.rand(batch_size, context_dim)
     >>> flow_dist = dist.ConditionalTransformedDistribution(base_dist,
-    >>> [transform]).condition(z)
+    ... [transform]).condition(z)
     >>> flow_dist.sample(sample_shape=torch.Size([batch_size]))  # doctest: +SKIP
 
     The inverse of the Bijector is required when, e.g., scoring the log density of a
--- a/pyro/distributions/transforms/affine_coupling.py
+++ b/pyro/distributions/transforms/affine_coupling.py
@@ -16,20 +16,24 @@ from functools import partial
 @copy_docs_from(TransformModule)
 class AffineCoupling(TransformModule):
     """
-    An implementation of the affine coupling layer of RealNVP (Dinh et al., 2017) that uses the bijective transform,
+    An implementation of the affine coupling layer of RealNVP (Dinh et al., 2017)
+    that uses the bijective transform,
 
         :math:`\\mathbf{y}_{1:d} = \\mathbf{x}_{1:d}`
         :math:`\\mathbf{y}_{(d+1):D} = \\mu + \\sigma\\odot\\mathbf{x}_{(d+1):D}`
 
-    where :math:`\\mathbf{x}` are the inputs, :math:`\\mathbf{y}` are the outputs, e.g. :math:`\\mathbf{x}_{1:d}`
-    represents the first :math:`d` elements of the inputs, and :math:`\\mu,\\sigma` are shift and translation
-    parameters calculated as the output of a function inputting only :math:`\\mathbf{x}_{1:d}`.
+    where :math:`\\mathbf{x}` are the inputs, :math:`\\mathbf{y}` are the outputs,
+    e.g. :math:`\\mathbf{x}_{1:d}` represents the first :math:`d` elements of the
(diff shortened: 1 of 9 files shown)


pyro/distributions/transforms/affine_coupling.py
```diff
@@ -125,11 +128,86 @@ def log_abs_det_jacobian(self, x, y):
...
+
+    Example usage:
+
+    >>> from pyro.nn import DenseNN
+    >>> input_dim = 10
+    >>> split_dim = 6
+    >>> base_dist = dist.Normal(torch.zeros(input_dim), torch.ones(input_dim))
+    >>> hypernet = DenseNN(split_dim, [10*input_dim], [input_dim-split_dim, input_dim-split_dim])
+    >>> transform = AffineCoupling(split_dim, hypernet)
+    >>> pyro.module("my_transform", transform)  # doctest: +SKIP
```
neerajprad: nit: tweak the example to reflect usage of `ConditionalAffineCoupling`?

pyro/distributions/transforms/affine_coupling.py
```diff
@@ -125,11 +128,86 @@ def log_abs_det_jacobian(self, x, y):
...
+    Bijector when its forward operation is called, e.g., when sampling from
+    :class:`~pyro.distributions.TransformedDistribution`. However, if the cached value isn't available, either because
+    it was overwritten during sampling a new value or an arbitary value is being scored, it will calculate it manually.
+
+    This is an operation that scales as O(1), i.e. constant in the input dimension. So in general, it is cheap
+    to sample *and* score (an arbitrary value) from :class:`~pyro.distributions.transforms.ConditionalAffineCoupling`.
+
+    :param split_dim: Zero-indexed dimension :math:`d` upon which to perform input/output split for transformation.
+    :type split_dim: int
+    :param nn: A neural network whose forward call returns a real-valued mean and logit-scale as a tuple. The input
```
neerajprad: Maybe specify that the nn constructor must take in a `context` argument to supplement the input dims?

tests/distributions/test_transforms.py
```diff
@@ -123,6 +123,9 @@ def test_block_autoregressive_jacobians(self):
         for residual in [None, 'normal', 'gated']:
             self._test(partial(T.block_autoregressive, residual=residual), inverse=False)
 
+    def test_conditional_affine_coupling(self):
+        self._test_conditional(T.conditional_affine_coupling)
```
neerajprad: I missed this test; thanks for adding!

pyro/distributions/transforms/affine_coupling.py
```diff
@@ -53,13 +56,13 @@ class AffineCoupling(TransformModule):
 
     :param split_dim: Zero-indexed dimension :math:`d` upon which to perform input/output split for transformation.
     :type split_dim: int
-    :param hypernet: an autoregressive neural network whose forward call returns a real-valued
-        mean and logit-scale as a tuple. The input should have final dimension split_dim and the output final
-        dimension input_dim-split_dim for each member of the tuple.
-    :type hypernet: callable
-    :param log_scale_min_clip: The minimum value for clipping the log(scale) from the autoregressive NN
+    :param nn: a neural network whose forward call returns a real-valued mean and logit-scale as a tuple. The
```
fritzo: Changing a param name of a documented class is not backwards compatible, so you should hold off renaming until a 2.0 release that allows breaking changes. I think we're tentatively planning on a 2.0 release in 2021 including major language redesign. Keep this in mind when choosing public interfaces 😉

stefanwebb: Hehe sorry about that! I've switched it back. Good to know we can reconsider the design for a 2.0 release - I think we can make some really great changes then to consolidate the bijective transform library

tests/distributions/test_transforms.py
stefanwebb: No problem!

pyro/distributions/transforms/affine_coupling.py
stefanwebb: Done!

pyro/distributions/transforms/affine_coupling.py
stefanwebb: Done!

stefanwebb:
Okay, made the requested changes :)

stefanwebb:
A problem with Travis?

neerajprad:
@stefanwebb - you'll need to merge changes from dev.

stefanwebb:
> @stefanwebb - you'll need to merge changes from dev.

✔️

pyro/distributions/transforms/affine_coupling.py
neerajprad: @fritzo - The attribute name is changed but the argument remains the same. That's fine by me since this maintains the current interface, or at least the extent to which I think we have committed. However if we consider attributes (those without leading underscores) to be part of the interface and if some code accesses it directly, it will break. Should we commit to keeping that backwards compatible too?

stefanwebb: I'm happy either way, provided there is freedom to change things around for a 2.0 release :)

fritzo: @neerajprad kwargs are part of the public interface, e.g. the proposed change would have broken
```py
transform = AffineCoupling(100, hypernet=MyMLP(100, 100))
```

neerajprad: I agree that would be problematic. The case I was talking about would be someone using `transform.hypernet = ...` later, which would break with this change (they would have to change it to `transform.nn = ...`).

PR #2382, opened 2020-03-24
Fix docs for MCMC

neerajprad:
- add docs for `MCMC.run`.
 - adds `MCMCKernel` class to the docs.
 - updates documentation regarding default `num_warmup`.

### DIFF
--- a/pyro/infer/mcmc/api.py
+++ b/pyro/infer/mcmc/api.py
@@ -273,7 +273,7 @@ class MCMC:
         excluding the samples discarded during the warmup phase.
     :param int warmup_steps: Number of warmup iterations. The samples generated
         during the warmup phase are discarded. If not provided, default is
-        half of `num_samples`.
+        is the same as `num_samples`.
     :param int num_chains: Number of MCMC chains to run in parallel. Depending on
         whether `num_chains` is 1 or more than 1, this class internally dispatches
         to either `_UnarySampler` or `_MultiSampler`.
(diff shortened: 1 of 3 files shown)


pyro/poutine/handlers.py
```diff
@@ -104,7 +104,7 @@ def handler(fn=None, *args, **kwargs):
             raise ValueError(
                 "{} is not callable, did you mean to pass it as a keyword arg?".format(fn))
         msngr = msngr_cls(*args, **kwargs)
-        return msngr(fn) if fn is not None else msngr
+        return functools.update_wrapper(msngr(fn), fn) if fn is not None else msngr
```
neerajprad: Without this, methods that use poutines as decorators will have their docstring reset. cc @eb8680 - does this seem fine?

pyro/infer/mcmc/api.py
```diff
@@ -350,6 +350,26 @@ def __init__(self, kernel, num_samples, warmup_steps=None, initial_params=None,
...
+
+            nuts_kernel = NUTS(model)
+            mcmc = MCMC(nuts_kernel, num_samples=500)
+            mcmc.run(data)
+            samples = mcmc.get_samples()
+
+        :param args: optional arguments taken by
+            :meth:`MCMCKernel.setup <pyro.infer.mcmc.mcmc_kernel.MCMCKernel.setup>`.
+        :param kwargs: optional keywords arguments taken by
+            :meth:`MCMCKernel.setup <pyro.infer.mcmc.mcmc_kernel.MCMCKernel.setup>`.
```
fritzo: Oh nice, I was unfamiliar with this syntax, but will start using it!

pyro/poutine/handlers.py
eb8680: LGTM

neerajprad: Hmm..seems like this is breaking something - `make doctest` and `tests.poutine.test_poutines.test_decorator_interface_primitives` seem to fail. I'll investigate and update soon.

fehiepsi:
Thanks, @neerajprad !

PR #2385, opened 2020-03-26
Aesthetic improvements to transform docstrings

stefanwebb:
The main purpose of this PR is to restrict the docstrings of `pyro.distributions.transforms` classes to be 80 chars long, for easy readability in Jupyter.

I have also made some other cosmetic changes like a consistent format for the references and removing the sample numbers after each code example in the docstring, which was the same for all transforms


stefanwebb:
I built the docs locally, but didn't check the rendering - hang on

stefanwebb:
It renders correctly!

PR #2387, opened 2020-03-26
Remove uses of `@torch.jit.script`

driazati:
This changes uses of `@torch.jit.script` to `@torch.jit._script_if_tracing`. The TorchScript compiler has a lot of lazy static initialization, so including these decorators will impact the time to execute `import pyro` for everyone, not just jit users. https://github.com/pytorch/pytorch/pull/34935 added a workaround for this so that if a function has this lazy script decorator, it can still be used to handle control flow in tracing.

cc @fritzo

### DIFF
--- a/pyro/distributions/hmm.py
+++ b/pyro/distributions/hmm.py
@@ -13,7 +13,7 @@ from pyro.ops.gaussian import Gaussian, gaussian_tensordot, matrix_and_mvn_to_ga
 from pyro.ops.tensor_utils import cholesky, cholesky_solve
 
 
-@torch.jit.script
+@torch.jit._script_if_tracing
 def _linear_integrate(init, trans, shift):
     """
     Integrate the inhomogeneous linear shifterence equation::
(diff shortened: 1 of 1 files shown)


fritzo:
@driazati thanks for fixing this!

Any chance you could make `_script_if_tracing()` public? If you do, we'll start using that for similar cases in Pyro 🙂

fritzo:
@driazati did some form of `._script_if_tracing()` make it into PyTorch 1.5?

jamestwebber:
> @driazati did some form of `._script_if_tracing()` make it into PyTorch 1.5?

Looks like it didn't make it (looking at [this](https://github.com/pytorch/pytorch/blob/v1.5.1/torch/jit/__init__.py)) but it will be in [1.6.0](https://github.com/pytorch/pytorch/blob/v1.6.0/torch/jit/__init__.py#L1583).

I wonder why they make it a "private" method if it's intended to be used by libraries.

fritzo:
Replacing with https://github.com/pyro-ppl/pyro/pull/2653 in anticipation of https://github.com/pytorch/pytorch/issues/45921

PR #2388, opened 2020-03-27
Allow unused params in CSIS inference

fritzo:
Addresses a comment on the forum
https://forum.pyro.ai/t/how-to-turn-on-allow-unused-true-through-pyro/1725

### DIFF
--- a/pyro/infer/csis.py
+++ b/pyro/infer/csis.py
@@ -111,7 +111,7 @@ class CSIS(Importance):
             if grads:
                 guide_params = set(site["value"].unconstrained()
                                    for site in particle_param_capture.trace.nodes.values())
-                guide_grads = torch.autograd.grad(particle_loss, guide_params)
+                guide_grads = torch.autograd.grad(particle_loss, guide_params, allow_unused=True)
                 for guide_grad, guide_param in zip(guide_grads, guide_params):
                     guide_param.grad = guide_grad if guide_param.grad is None else guide_param.grad + guide_grad
 


pyro/infer/csis.py
```diff
@@ -111,7 +111,7 @@ def loss_and_grads(self, grads, batch, *args, **kwargs):
             if grads:
                 guide_params = set(site["value"].unconstrained()
                                    for site in particle_param_capture.trace.nodes.values())
-                guide_grads = torch.autograd.grad(particle_loss, guide_params)
+                guide_grads = torch.autograd.grad(particle_loss, guide_params, allow_unused=True)
```
neerajprad: Looks fine. Just to confirm the issue - this shouldn't happen unless `guide_params` contains parameters that aren't used in the computation of `particle_loss`. Is that possible or common?

fritzo:
@DexiongYung does this fix your issue?

DexiongYung:
I believe so

pyro/infer/csis.py
fritzo: It is certainly possible e.g. in models with dynamic control flow with a dead branch, e.g.
```py
def model(data):
    a = pyro.param("a", torch.tensor(+1.))
    b = pyro.param("b", torch.tensor(-1.))
    x = pyro.sample("x", dist.Bernoulli(0.5))
    pyro.sample("y", dist.Normal(a if x else b, 1.), obs=data)
```
I don't recall having seen this error in any recent model, so I'm not sure how common it is.

neerajprad: Thanks for the snippet, makes sense.

Issue #790, opened 2018-02-20
Sampling from truncated Gaussian


bmazoure:
Is there currently a way to sample from new distributions which are similar to the ones already implemented in Pyro?

For instance, in the case of a multivariate truncated Gaussian we would need to define how to compute the gradient as a piecewise function and I am not sure where to do so.

martinjankowiak:
hello, you might take a look here:

https://github.com/probtorch/pytorch/pull/121

fritzo:
Hi @bmazoure you could also define a [Rejector distribution](https://github.com/uber/pyro/blob/dev/pyro/distributions/rejector.py) for a truncated normal, but this would require you to implement the probability of acceptance `log_scale` as a function of your truncation plane. For example to truncate by ensuring `sample[0] > min_x0`, you should be able to define
```py
class TruncatedMVN(dist.Rejector):
    def __init__(self, loc, covariance_matrix, min_x0):
        propose = dist.MultivariateNormal(loc, covariance_matrix)

        def log_prob_accept(x):
            return (x[0] > min_x0).type_as(x).log()

        scale_0 = torch.sqrt(covariance_matrix[0, 0])
        log_scale = torch.log(1 - dist.Normal(loc[0], scale_0).cdf(min_x0))
        super(TruncatedMVN, self).__init__(propose, log_prob_accept, log_scale)
```

bmazoure:
Will try these suggestions, thanks!

dobos:
I'm trying to use a rejector to limit a Pareto distribution, which is wrapped from a torch distribution:

```
class TruncatedPareto(Rejector):
    def __init__(self, scale, alpha, upper_limit, validate_args=None):
        propose = Pareto(scale, alpha, validate_args=validate_args)

        def log_prob_accept(x):
            return (x < upper_limit).type_as(x).log()

        log_scale = torch.Tensor(alpha) * torch.log(torch.Tensor([scale / upper_limit]))
        super(TruncatedPareto, self).__init__(propose, log_prob_accept, log_scale)
```

I can sample from it by calling `.sample()` but inside an MCMC, I get the error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-353-ebb20659df6e> in <module>
      2 kernel = NUTS(conditioned_model)
      3 mcmc = MCMC(kernel, num_samples=10, warmup_steps=2)
----> 4 mcmc.run(meta)
      5 posterior_samples = mcmc.get_samples()

[...]

PR #2389, opened 2020-03-27
Add super().__init__() call to Rejector

fritzo:
Fixes a bug noticed by @dobos here: https://github.com/pyro-ppl/pyro/issues/790#issuecomment-605084361

## Breaking change

This will break some code, as in e.g. `RejectionStandardGamma`, but I see no way of obtaining the needed shape metadata without a little breakage.

## Tested
- refactoring is covered by existing tests

### DIFF
--- a/pyro/distributions/testing/rejection_gamma.py
+++ b/pyro/distributions/testing/rejection_gamma.py
@@ -39,7 +39,8 @@ class RejectionStandardGamma(Rejector):
         # Compute log scale using Gamma.log_prob().
         x = new._d.detach()  # just an arbitrary x.
         log_scale = new.propose_log_prob(x) + new.log_prob_accept(x) - new.log_prob(x)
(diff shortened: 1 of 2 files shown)


pyro/distributions/testing/rejection_gamma.py
```diff
@@ -26,7 +26,8 @@ def __init__(self, concentration):
         # Compute log scale using Gamma.log_prob().
         x = self._d.detach()  # just an arbitrary x.
         log_scale = self.propose_log_prob(x) + self.log_prob_accept(x) - self.log_prob(x)
-        super().__init__(self.propose, self.log_prob_accept, log_scale)
+        super().__init__(self.propose, self.log_prob_accept, log_scale,
```
neerajprad: Some of the distribution tests are failing. I think we may need to change the `.expand` below as well?

fritzo:
bump

PR #2391, opened 2020-03-29
Fix PyroModule[nn.GRU] and other RNNs

### LINKED ISSUE #2390: PyroModule incompatible with torch.nn.RNN
As pointed out [on the forum](https://forum.pyro.ai/t/pyromodule-pyrosample-pyroparam-and-gru/1731),  `PyroModule` is incompatible with `torch.nn.RNN` due to [internal caching logic](https://github.com/pytorch/pytorch/blob/e90c32f11fc54d017609e68fb80e19350e199864/torch/nn/modules/rnn.py#L88).

I believe a workaround is to call `my_rnn._apply(lambda t: t)` inside a supermodule's `.forward()` method. There may be a more general fix.

fritzo:
Fixes #2390 

This fixes an issue with Bayesian RNNs caused by a caching mechanism in `nn.RNNBase`. The solution is a bit hacky in that it overwrites a private attribute, but I see no alternative solution.

## Tested
- [x]  added a regression test with `nn.GRU`

### DIFF
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -97,8 +97,10 @@ class _PyroModuleMeta(type):
             return PyroModule
         if Module in _PyroModuleMeta._pyro_mixin_cache:
             return _PyroModuleMeta._pyro_mixin_cache[Module]
+        bases = [PyroModule[b] for b in Module.__bases__
+                 if issubclass(b, torch.nn.Module)]
 
-        class result(Module, PyroModule):
+        class result(Module, *bases):
             # Unpickling helper to load an object of type PyroModule[Module].
             def __reduce__(self):
                 state = getattr(self, '__getstate__', self.__dict__.copy)()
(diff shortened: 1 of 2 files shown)


pyro/nn/module.py
```diff
@@ -97,8 +97,10 @@ def __getitem__(cls, Module):
             return PyroModule
         if Module in _PyroModuleMeta._pyro_mixin_cache:
             return _PyroModuleMeta._pyro_mixin_cache[Module]
+        bases = [PyroModule[b] for b in Module.__bases__
+                 if issubclass(b, torch.nn.Module)]
 
-        class result(Module, PyroModule):
+        class result(Module, *bases):
```
fritzo: This change seems clean and robust. It ensures distributivity of `PyroModule[-]`. For example if
```py
issubclass(GRU, RNNBase)
```
then
```py
issubclass(PyroModule[GRU], PyroModule[RNNBase])
```

neerajprad:
LGTM, merging this.

Issue #2386, opened 2020-03-26
FR: subsampling-aware PyTorch optimizers


fritzo:
@martinjankowiak suggested that, now that #1796 makes `pyro.param` aware of subsampling, we could in principle make PyTorch optimizers whose gradient statistics are updated only for those elements that appear in a subsample. This would give lower-variance gradient estimates and would be cheaper.

@martinjankowiak also points out that an alternative is to amortize the guide.

> you can imagine two implementations:
> * one that explicitly makes use of subsample indices
> * one that uses grad != 0 as a proxy
> are there circumstances under which the proxy isn’t good enough?

fehiepsi:
I believe this is important for my experiments so I want to fix this soon.

martinjankowiak:
@fehiepsi i think the easiest way to get something that works would be to add an option to `ClippedAdam` like `ignore_zero_gradient_stats`=`True`/`False`

you'd then need to keep track of `state['step']` on a per-coordinate basis (i.e. the optimizer would need more memory) and then you'd just need to change the updates to do vectorized/masked updates of the statistics

fritzo:
@martinjankowiak can you explain why it is necessary to keep track of coordinate-wise `state['step']`? I would think that `step` could be approximated as global, since the poisson approximation concentrates (in contrast to the other statistics).

martinjankowiak:
well you could make that approximation. but i was assuming a pretty generic
optimizer that makes few assumptions (just don't update when grad is zero)

Am Di., 31. März 2020 um 13:10 Uhr schrieb Fritz Obermeyer <
notifications@github.com>:

fritzo:
I agree the coordinate-wise `step` is parsimonious. Another parsimonious assumption could be that the gradient distribution is a zero-inflated Normal, so that with slight modifications, the usual `Adam` statistics can learn that distribution's parameters. Both versions seem reasonable.

PR #2393, opened 2020-04-02
make DCTAdam acknowledge subsample

fehiepsi:
Addresses #2386.

There are some changes that I haven't optimized such as separating the case `mask` is full index. I think it would be fine because the optimizer step is cheap. Please let me know if we need to do profiling those changes.

TODO:
- [x] add tests
- [x] incorporate reviewers' ideas

### DIFF
--- a/pyro/optim/dct_adam.py
+++ b/pyro/optim/dct_adam.py
@@ -69,10 +69,13 @@ class DCTAdam(Optimizer):
         numerical stability (default: 1e-8)
     :param float clip_norm: magnitude of norm to which gradients are clipped (default: 10.0)
     :param float lrd: rate at which learning rate decays (default: 1.0)
+    :param bool aware_subsample: whether to update gradient statistics only for
+        those elements that appear in a subsample (default: False).
     """
     def __init__(self, params, lr=1e-3, betas=(0.9, 0.999), eps=1e-8,
-                 clip_norm=10.0, lrd=1.0):
-        defaults = dict(lr=lr, betas=betas, eps=eps, clip_norm=clip_norm, lrd=lrd)
+                 clip_norm=10.0, lrd=1.0, aware_subsample=False):
+        defaults = dict(lr=lr, betas=betas, eps=eps, clip_norm=clip_norm, lrd=lrd,
+                        aware_subsample=aware_subsample)
         super().__init__(params, defaults)
 
     def step(self, closure=None):
@@ -93,7 +96,7 @@ class DCTAdam(Optimizer):
                     continue
 
                 subsample = getattr(p, "_pyro_subsample", {})
-                if subsample:
+                if subsample and group['aware_subsample']:
                     self._step_param_subsample(group, p, subsample)
                 else:
                     self._step_param(group, p)
--- a/pyro/optim/dct_adam.py
+++ b/pyro/optim/dct_adam.py
@@ -144,7 +144,7 @@ class DCTAdam(Optimizer):
     def _step_param_subsample(self, group, p, subsample):
         mask = _get_mask(p, subsample)
 
-        grad = p.grad.data[mask]
+        grad = p.grad.data.masked_select(mask)
         grad.clamp_(-group['clip_norm'], group['clip_norm'])
 
         # Transform selected parameters via dct.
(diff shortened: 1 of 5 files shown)


pyro/optim/dct_adam.py
```diff
@@ -33,6 +31,20 @@ def _transform_inverse(x, dim, duration):
     return idct(x[right], dim)[left].add_(x[left])
 
 
+def _get_mask(x, indices):
+    # create a bollean mask from subsample indices
```
martinjankowiak, line 35: typo: boolean

pyro/optim/dct_adam.py
```diff
@@ -33,6 +31,20 @@ def _transform_inverse(x, dim, duration):
     return idct(x[right], dim)[left].add_(x[left])
 
 
+def _get_mask(x, indices):
```
martinjankowiak, line 34: i'm having a bit of trouble understanding this helper. can you please either add more comments or add a test?

pyro/optim/dct_adam.py
```diff
@@ -86,31 +101,36 @@ def step(self, closure=None):
...
-                    p.data.addcdiv_(-step_size, exp_avg, denom)
+                    step = exp_avg[mask] / denom
+                    p.data[mask] += (step.mul_(-step_size))
                 else:
-                    step = _transform_inverse(exp_avg / denom, time_dim, duration)
-                    p.data.add_(step.mul_(-step_size))
+                    step = _transform_inverse(exp_avg[mask] / denom, time_dim, duration)
+                    p.data[mask] += (step.mul_(-step_size))
+
+                if hasattr(p, "_pyro_subsample"):
```
martinjankowiak, line 133: why is this necessary?

pyro/poutine/subsample_messenger.py
```diff
@@ -149,4 +149,17 @@ def _postprocess_message(self, msg):
                             .format(self.name, self.size, self.dim, statement, shape))
                     # Subsample parameters with known batch semantics.
                     if self.subsample_size < self.size:
-                        msg["value"] = msg["value"].index_select(dim, self._indices)
+                        value = msg["value"]
```
martinjankowiak, line 152: why was the following logic necessary?

pyro/optim/dct_adam.py
```diff
@@ -33,6 +31,20 @@ def _transform_inverse(x, dim, duration):
     return idct(x[right], dim)[left].add_(x[left])
 
 
+def _get_mask(x, indices):
+    # create a bollean mask from subsample indices
+    mask = x.new_zeros(x.shape, dtype=torch.long, device=x.device)
+    if len(indices) > 0:
+        size = x.dim()
+        idx = []
+        for dim in range(size):
```
fritzo, line 40: nit: `for dim in range(-x.dim(), 0):` would avoid the need for `neg_dim`

pyro/optim/dct_adam.py
fehiepsi: This is not necessarily required. I just want to make sure those extra attributes are cleared if, for some reason, users want to run SVI with subsampling at the early training phase and run SVI without subsampling at later steps. This might be not a good place to put that logic. Do you have other suggestions?

pyro/poutine/subsample_messenger.py
fehiepsi: The logic is used to add `_pyro_subsample` attribute to the parameters which need to be optimized. I think that checking `grad=0` might affect non-subsampled parameters (e.g. when `loss = 2 * param +...` and `param` has the initial value is `0`).

pyro/optim/dct_adam.py
fehiepsi: Thanks!

pyro/optim/dct_adam.py
fehiepsi: thanks!

pyro/optim/dct_adam.py
fehiepsi: I just add some comments to explain the logic. :)

fehiepsi:
Thank @fritzo and @martinjankowiak for reviewing and helpful suggestions at #2386! I have made corresponding changes. There is one minor issue is whether we clear "pyro_subsample" attribute at `DCTAdam.step` or someplace else. It is not important IMO so we might just skip it.

pyro/optim/dct_adam.py
```diff
@@ -73,44 +91,95 @@ def step(self, closure=None):
...
+            # Exponential moving average of squared gradient values
+            state['exp_avg_sq'] = torch.zeros_like(p)
+
+        exp_avg, exp_avg_sq = state['exp_avg'], state['exp_avg_sq']
+        beta1, beta2 = group['betas']
+
+        state['step'][mask] += 1
+
+        # Decay the first and second moment running average coefficient
+        exp_avg.masked_scatter_(mask, exp_avg[mask].mul_(beta1).add_(1 - beta1, grad))
```
fehiepsi, line 172: @martinjankowiak using `masked_scatter_` is faster than `exp_avg[mask] = ...`

fehiepsi:
Hmm, looks like Travis is not working today.

pyro/poutine/subsample_messenger.py
fritzo: It seems cleanest to me to use purely mask-based logic if possible, and assume it "almost surely" works. Can you think of a counterexample other than the first step example above?

fehiepsi:
> why 54 -> 78?

It is because of those `x[mask]`. I have removed duplicated ones and seen the speed much improved. (another source is `beta1 ** state_step`; before this PR, it is scalar op; now it is tensor op)

pyro/poutine/subsample_messenger.py
fehiepsi: I can't come up with a good example which shows the shortcomings of using `mask = (grad != 0)`: one example I have in mind is when a parameter moves to a local maximum area, which can have nonzero measure due to precision issue - to get out of it, we need to update the parameter by using grad information from past steps.

pyro/optim/dct_adam.py
```diff
@@ -73,44 +94,98 @@ def step(self, closure=None):
...
+            state['step'] = torch.zeros_like(p)
+            # Exponential moving average of gradient values
+            state['exp_avg'] = torch.zeros_like(p)
+            # Exponential moving average of squared gradient values
+            state['exp_avg_sq'] = torch.zeros_like(p)
+
+        beta1, beta2 = group['betas']
+
+        state_step = state['step'].masked_select(mask).add_(1)
+        state['step'].masked_scatter_(mask, state_step)
```
fritzo, line 172: Interesting pattern! Consider abstracting this out as a `masked_add_()` helper in `pyro.distributions.util`.

pyro/optim/dct_adam.py
```diff
@@ -51,10 +69,13 @@ class DCTAdam(Optimizer):
         numerical stability (default: 1e-8)
     :param float clip_norm: magnitude of norm to which gradients are clipped (default: 10.0)
     :param float lrd: rate at which learning rate decays (default: 1.0)
+    :param bool aware_subsample: whether to update gradient statistics only for
```
fritzo, line 72: nit: could we name this arg `subsample_aware`?

pyro/optim/dct_adam.py
fehiepsi: I would like to skip this abstraction. At other places, we not only use `add` but also `mul` or `addcmul`.

Issue #2394, opened 2020-04-03
Raise error in MCMC when the model has subsample sites


fehiepsi:
Raised at [this forum topic](https://forum.pyro.ai/t/subsampling-in-a-plate-makes-a-program-grind-to-a-halt/1748/3)

PR #2397, opened 2020-04-05
raise error if mcmc observes subsample site

fehiepsi:
Addresses #2394.

### DIFF
--- a/pyro/infer/mcmc/util.py
+++ b/pyro/infer/mcmc/util.py
@@ -383,8 +383,9 @@ def initialize_model(model, model_args=(), model_kwargs={}, transforms=None, max
     has_enumerable_sites = False
     prototype_samples = {}
     for name, node in model_trace.iter_stochastic_nodes():
-        if isinstance(node["fn"], _Subsample):
-            if node["fn"].subsample_size:
+        fn = node["fn"]
+        if isinstance(fn, _Subsample):
+            if fn.subsample_size is not None and fn.subsample_size < fn.size:
                 raise NotImplementedError("HMC/NUTS does not support model with subsample sites.")
             continue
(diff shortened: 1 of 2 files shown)


pyro/infer/mcmc/util.py
```diff
@@ -384,6 +384,8 @@ def initialize_model(model, model_args=(), model_kwargs={}, transforms=None, max
     prototype_samples = {}
     for name, node in model_trace.iter_stochastic_nodes():
         if isinstance(node["fn"], _Subsample):
+            if node["fn"].subsample_size:
```
fritzo, line 387: I think we can be more permissive and check:
```py
fn = node["fn"]
if fn.subsample_size is not None or fn.subsample_size < fn.size:
    raise ...
```
This allows passing `subsample_size=len(data)` for models that support either MCMC or SVI.

tests/infer/mcmc/test_mcmc_util.py
```diff
@@ -68,3 +68,14 @@ def test_model_with_param(jit_compile, num_chains):
     kernel = NUTS(model_with_param, jit_compile=jit_compile, ignore_jit_warnings=True)
     mcmc = MCMC(kernel, 10, num_chains=num_chains, mp_context="spawn")
     mcmc.run()
+
+
+def test_model_with_subsample():
+    def model():
+        with pyro.plate("J", 10, subsample_size=5):
```
fritzo, line 75: we could add a test that `pyro.plate("J", 10, subsample_size=10)` works without error.

neerajprad:
Not sure why we are seeing two travis PR checks and one is stalled. I'll reopen this PR.

fehiepsi:
Opss... I'll make a new PR instead.

PR #2398, opened 2020-04-05
support NegativeBinomial in forecast

fehiepsi:
NegativeBinomial is another parameterization of GammaPoisson. I want to add support for the zero-inflated version of it.

@fritzo This is the last piece I want to include in this patch release. :)

### DIFF
--- a/pyro/distributions/zero_inflated.py
+++ b/pyro/distributions/zero_inflated.py
@@ -97,6 +97,10 @@ class ZeroInflatedNegativeBinomial(ZeroInflatedDistribution):
     :param torch.Tensor probs: Event probabilities of success in the half open interval [0, 1).
     :param torch.Tensor logits: Event log-odds for probabilities of success.
     """
+    arg_constraints = {"gate": constraints.unit_interval,
+                       "total_count": constraints.greater_than_eq(0),
+                       "probs": constraints.half_open_interval(0., 1.),
+                       "logits": constraints.real}
     support = constraints.nonnegative_integer
 
     def __init__(self, gate, total_count, probs=None, logits=None, validate_args=None):
@@ -110,3 +114,15 @@ class ZeroInflatedNegativeBinomial(ZeroInflatedDistribution):
         super().__init__(
             gate, base_dist, validate_args=validate_args
         )
+
+    @property
+    def total_count(self):
+        return self.base_dist.total_count
+
+    @property
+    def probs(self):
+        return self.base_dist.probs
+
+    @property
+    def logits(self):
+        return self.base_dist.logits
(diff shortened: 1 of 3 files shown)


pyro/distributions/zero_inflated.py
```diff
@@ -110,3 +114,15 @@ def __init__(self, gate, total_count, probs=None, logits=None, validate_args=Non
         super().__init__(
             gate, base_dist, validate_args=validate_args
```
fritzo, line 115: Now that `arg_constraints` is specified, args will be validated twice. I think we could avoid double validation similar to how it's done in `.expand()`, i.e.
```py
super().__init__(gate, base_dist, validate_args=False)
self._validate_args = validate_args
```
WDYT?

fehiepsi: I guess you meant `base_dist = NegativeBinomial(..., validate_args=False)`? We need to validate the `gate` too.

fritzo: Hmm ok that version works too, just make sure you set `._validate_args` afterwards so that `.log_prob()` will validate:
```py
base_dist = NegativeBinomial(..., validate_args=False)
base_dist._validate_args = validate_args
```

pyro/distributions/zero_inflated.py
```diff
@@ -76,7 +74,7 @@ class ZeroInflatedPoisson(ZeroInflatedDistribution):
     support = constraints.nonnegative_integer
 
     def __init__(self, gate, rate, validate_args=None):
-        base_dist = Poisson(rate=rate, validate_args=validate_args)
+        base_dist = Poisson(rate=rate, validate_args=False)
```
fritzo, line 77:
```py
base_dist._validate_args = validate_args
```

pyro/distributions/zero_inflated.py
```diff
@@ -97,16 +95,32 @@ class ZeroInflatedNegativeBinomial(ZeroInflatedDistribution):
...
+                       "logits": constraints.real}
     support = constraints.nonnegative_integer
 
     def __init__(self, gate, total_count, probs=None, logits=None, validate_args=None):
         base_dist = NegativeBinomial(
             total_count=total_count,
             probs=probs,
             logits=logits,
-            validate_args=validate_args,
+            validate_args=False,
```
fritzo, line 109: base_dist._validate_args = validate_args

pyro/distributions/zero_inflated.py
fehiepsi: You are right, thanks!

pyro/distributions/zero_inflated.py
fehiepsi: done

Issue #2076, opened 2019-10-22
Composing distributions.transforms with TanhTransform


bmazoure:
I am trying to compose a Normalizing Flow (Sylvester) with a bijective (Tanh) transform to run the following code:
```
dim = 2
mu = torch.zeros(dim)
log_std = torch.zeros(dim)
transforms = [distributions.transforms.SylvesterFlow(input_dim=dim),
                           distributions.transforms.TanhTransforms()]
distributions.TransformedDistribution(distributions.Normal(mu,log_std.exp()), transforms)
```

but get the following:
```
~/sandbox/pyro_fork/pyro/distributions/transforms/sylvester.py in _inverse(self, y)
    127         """
    128 
--> 129         raise KeyError("SylvesterFlow expected to find key in intermediates cache but didn't")
    130 
    131     def log_abs_det_jacobian(self, x, y):

KeyError: "SylvesterFlow expected to find key in intermediates cache but didn't"
```

From what I understand, only Householder and IAF have their analytical inverse coded, which is causing the error. 
AFAIK, the inversion happens by looking up the value y in cache, and if the corresponding x is found, it is returned. However, since Tanh is bijective, shouldn't this cause no trouble, as the value y=Tanh(NF(x)) will be cached and returned?

eb8680:
cc @stefanwebb

stefanwebb:
I think you're right, there is a bug when composing a transform with an analytic inverse to one without.

The inverse of Tanh is calculated analytically, which then differs (in reference) from the value stored in the cache of the Sylvester transform.

The fix is to make Tanh cache its value - I'll add this change to #1932

PR #2401, opened 2020-04-09
Patch so that `cache_size=1` is default in `torch.distributions.transforms.Transform` and elsewhere

stefanwebb:
This PR addresses #2076  

Essentially, when you compose a transform that requires caching for its inverse, and some of the element-wise transforms from PyTorch you can get into strife because the latter don't always do caching by default, and some don't even give you the option of setting caching (e.g. `TanhTransform` and `SigmoidTransform`)!

Here is an example of code that will raise an exception without the patches to the PyTorch transform module:

```py
import torch
import pyro.distributions.transforms as T

torch.manual_seed(0)
dim = 2
mu = torch.zeros(dim)
log_std = torch.zeros(dim)
transforms = [T.Sylvester(input_dim=dim), T.TanhTransform()]
d = dist.TransformedDistribution(dist.Normal(mu, log_std.exp()), transforms)

y = d.sample()
print(d.log_prob(y))

```

I add the discussion label because I just want to run past @fritzo, do you think it's okay to set `Transform` to do caching by default, rather than adding a `cache_size=1` keyword argument to every element-wise transform?

### DIFF
--- a/pyro/distributions/torch_patch.py
+++ b/pyro/distributions/torch_patch.py
@@ -5,6 +5,8 @@ import functools
 import weakref
 
 import torch
+from torch.distributions.transforms import Transform, PowerTransform, AffineTransform
+from torch.distributions.utils import broadcast_all
 
 assert torch.__version__.startswith('1.')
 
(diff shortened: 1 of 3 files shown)


stefanwebb:
Ah I see! Let me do a mock up of option (2) to see if that would work.

Yes I think all transforms that don't require caching should give you the option at least of caching them in a keyword arg to `__init__` (with default value 0) so will PR that to PyTorch

fritzo:
@stefanwebb I needed the `ComposeTransform` fix to solve another problem. Can we collaborate on fixing this upstream in Pytorch https://github.com/probtorch/pytorch/pull/153 ?

stefanwebb:
I'm going to close this because it's superseded by https://github.com/probtorch/pytorch/pull/153

PR #2402, opened 2020-04-09
`ConditionalAffineAutoregressive`

stefanwebb:
Version of `AffineAutoregressive` that conditions on a context variable. Required for, e.g., reproducing experiments in IAF/MAF papers


pyro/distributions/transforms/affine_autoregressive.py
```diff
@@ -220,6 +223,106 @@ def _inverse_stable(self, y):
...
+    codomain = constraints.real
+    bijective = True
+    event_dim = 1
+
+    def __init__(self, hypernet, **kwargs):
+        super().__init__()
+        self.nn = hypernet
+        self.kwargs = kwargs
+
+    def condition(self, context):
```
neerajprad, line 321: Consider adding a small docstring for `condition`.

pyro/distributions/transforms/affine_autoregressive.py
```diff
@@ -220,6 +223,106 @@ def _inverse_stable(self, y):
...
+        cond_nn = partial(self.nn, context=context)
+        return AffineAutoregressive(cond_nn, **self.kwargs)
```
neerajprad, line 323: Do we have any error checking on `kwargs`, i.e. if an invalid kwarg is passed would we raise an exception anywhere?

pyro/distributions/transforms/affine_autoregressive.py
stefanwebb: Done!

pyro/distributions/transforms/affine_autoregressive.py
stefanwebb: An exception will be thrown in the initializer of `AffineAutoregressive`

pyro/distributions/transforms/affine_autoregressive.py
```diff
@@ -5,10 +5,13 @@
...
 from torch.distributions import constraints
 
+from pyro.distributions.conditional import ConditionalTransformModule
 from pyro.distributions.torch_transform import TransformModule
 from pyro.distributions.util import copy_docs_from
 from pyro.distributions.transforms.utils import clamp_preserve_gradients
-from pyro.nn import AutoRegressiveNN
+from pyro.nn import ConditionalAutoRegressiveNN, AutoRegressiveNN
+
+from functools import partial
```
fritzo, line 14: nit: can you normalize import order using `isort`?
```sh
pip install isort
isort -iy  pyro/distributions/transforms
```

stefanwebb: Done for this and other transforms!

fritzo:
@stefanwebb can you resolve the merge conflict?

stefanwebb:
Done

pyro/distributions/transforms/affine_autoregressive.py
```diff
@@ -220,6 +223,110 @@ def _inverse_stable(self, y):
...
+    value isn't available, either because it was overwritten during sampling a new
+    value or an arbitary value is being scored, it will calculate it manually. Note
+    that this is an operation that scales as O(D) where D is the input dimension,
+    and so should be avoided for large dimensional uses. So in general, it is cheap
+    to sample from IAF and score a value that was sampled by IAF, but expensive to
+    score an arbitrary value.
+
+    :param autoregressive_nn: an autoregressive neural network whose forward call
+        returns a real-valued mean and logit-scale as a tuple
+    :type autoregressive_nn: nn.Module
```
fritzo, line 283: I think you need to replace this docstring with a description of the `hypernet` arg?

pyro/distributions/transforms/affine_autoregressive.py
```diff
@@ -220,6 +223,110 @@ def _inverse_stable(self, y):
         return x
 
 
+@copy_docs_from(ConditionalTransformModule)
+class ConditionalAffineAutoregressive(ConditionalTransformModule):
+    """
```
fritzo, line 228: note: if you use [raw strings](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals) you can replace all your double backslashes with single backslashes. Feel free to ignore
```py
r"""
:math:`\sigma`
"""
```

stefanwebb: Hang on a sec, I'll make this change

stefanwebb:
Hold off on merging this PR, I need to fix a bug I discovered...

pyro/distributions/transforms/affine_autoregressive.py
stefanwebb: done

pyro/distributions/transforms/affine_autoregressive.py
stefanwebb: done

pyro/distributions/transforms/affine_autoregressive.py
```diff
@@ -220,6 +221,111 @@ def _inverse_stable(self, y):
...
+        self.kwargs = kwargs
+
+    def condition(self, context):
+        """
+        Conditions on a context variable, returning a non-conditional transform of
+        of type :class:`~pyro.distributions.transforms.AffineAutoregressive`.
+        """
+
+        # Note that nn.condition doesn't copy the weights of the ConditionalAutoregressiveNN
+        return AffineAutoregressive(self.nn.condition(context), **self.kwargs)
```
stefanwebb, line 326: @fritzo Can you please review this line? I have made a design change here. Previously I was using `partial` to fix the context that is used in the `ConditionalAutoRegressiveNN`, but this resulted in some bugs... You need to pass an `nn.Module` class to `AffineAutoregressive` rather than a function completion so that it registers the parameters and you can inspect the properties inside like the permutation (used in the tests)

stefanwebb:
I think it should be good to go, subject to the design issue I had above

pyro/distributions/transforms/affine_autoregressive.py
stefanwebb: I think in a different programming language, I would subclass `ConditionalAutoRegressiveNN` to include a context property, overwrite `.forward` to use this context, then copy self.nn into a new object of this type. That way you could share the weights between multiple `self.nn.condition(context)` objects rather than have the context rewritten each time it is called

pyro/distributions/transforms/affine_autoregressive.py
```diff
@@ -220,6 +221,111 @@ def _inverse_stable(self, y):
...
+    ... [transform]).condition(z)
+    >>> flow_dist.sample(sample_shape=torch.Size([batch_size]))  # doctest: +SKIP
+
+    The inverse of the Bijector is required when, e.g., scoring the log density of a
+    sample with :class:`~pyro.distributions.TransformedDistribution`. This
+    implementation caches the inverse of the Bijector when its forward operation is
+    called, e.g., when sampling from
+    :class:`~pyro.distributions.TransformedDistribution`. However, if the cached
+    value isn't available, either because it was overwritten during sampling a new
+    value or an arbitary value is being scored, it will calculate it manually. Note
```
martinjankowiak, line 273: typo: arbitary

stefanwebb: Fixed!

pyro/distributions/transforms/affine_autoregressive.py
martinjankowiak: @fritzo @stefanwebb have you guys resolved this design issue? otherwise lgtm

fritzo: Sorry I let this drop. I don't fully understand the design issue. Specifically what contract is made by `nn.condition()`? Does it always return a neural net? Which of `nn` and `nn.condition(...)` is ever serialized or trained? What bugs were caused by unregistered parameters? @stefanwebb let me know if you'd like to zoom to discuss any time.

fritzo:
I believe @stefanwebb and I discussed changing `.condition()` either to explicitly act as an in-place op or to create a shallow copy of the conditioned nn.

stefanwebb:
Yes that's right, sorry I haven't been on top of this...

There is still some question of the best way to get the parameters out of a composition of transforms so I'm going to go with the `PartialModule` solution Fritz and I discussed that combines a function partial with a Torch `nn.Module` so the conditioned distribution can have the parameters extracted from it.

Will ping you when it's ready for review!

stefanwebb:
I realized that we don't expect the result of a `transform.condition(z)` to be an `nn.Module` so opted for a simpler solution! @martinjankowiak it's ready for review :)

martinjankowiak:
@stefanwebb is there a preferred merge order for any of these PRs?

stefanwebb:
> @stefanwebb is there a preferred merge order for any of these PRs?

@martinjankowiak in chronological order, skipping the #2442, which needs some tests

PR #2403, opened 2020-04-09
Add a PyroModule tutorial

### LINKED ISSUE #2396: FR Tutorial on PyroModule
Addresses #2161 

Many people have requested a tutorial on [PyroModule](http://docs.pyro.ai/en/stable/nn.html#module-pyro.nn.module), e.g. on the [forum](https://forum.pyro.ai/t/the-usage-of-pyromodule/1443).

fritzo:
Resolves #2396 

This adds a simple tutorial on `PyroModule`. It is very basic. I welcome requests for additional material.

## Tested
- [x] built locally

View on [github](https://github.com/pyro-ppl/pyro/blob/module-tutorial/tutorial/source/modules.ipynb) or [nbviewer](https://nbviewer.jupyter.org/github/pyro-ppl/pyro/blob/module-tutorial/tutorial/source/modules.ipynb).

### DIFF
--- a/tutorial/Makefile
+++ b/tutorial/Makefile
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    ?= -E -W
+SPHINXOPTS    ?= -E -W -j 8
 SPHINXBUILD   = python -msphinx
 SPHINXPROJ    = PyroTutorials
 SOURCEDIR     = source
--- a/tutorial/source/index.rst
+++ b/tutorial/source/index.rst
@@ -26,6 +26,7 @@ Welcome to Pyro Examples and Tutorials!
    jit
    minipyro
    effect_handlers
+   modules
 
 .. toctree::
    :maxdepth: 2
--- a/tutorial/source/modules.ipynb
+++ b/tutorial/source/modules.ipynb
@@ -1,0 +1,604 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Modules in Pyro\n",
+    "\n",
+    "\n",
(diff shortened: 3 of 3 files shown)


tutorial/Makefile
```diff
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    ?= -E -W
+SPHINXOPTS    ?= -E -W -j 8
```
fritzo, line 5: This speeds up local building of tutorials, which is nice since we now have 50+ tutorials

martinjankowiak:
looks great! 

some typos i noticed:

- typo "but allowing"
- typo "wich"
- typo wrt "to_pyro_module_" above cell 5

fehiepsi:
Thanks for adding the tutorial! I have a few comments:
+ It might be helpful to mention hierarchical prior (using `lambda: ...` in PyroSample)

+ Those statements are a bit of conflict.

> A PyroModule model may contain nn.Module attributes, but not vice versa.
An nn.Module may contain both a PyroModule model and PyroModule guide (e.g. Predictive).

neerajprad:
This was super useful for me too! 

I only had some minor comments - feel free to address as you see fit:

* Summary
  - "A PyroModule model may contain nn.Module attributes, but not vice versa." - maybe we should comment on why - namespace collision and direct to the naming section?
* How PyroModule works
  - jit_module --> jit.trace_module (if this is intentional, its fine).
  - e.g. nn.GRU is a subclass of nn.RNN -> Consider expanding that `PyroModule[nn.GRU]` will be a subclass of `PyroModule[nn.RNN]` using this syntax. 
* How to make a PyroModule Bayesian
  - "to" missing in the heading
  - consider adding a note as to why we need `to_event(2)` (also used in later models), and direct to the shapes tutorial.
  - It is implicit but may be worth explicitly mentioning that weight is a special attribute and accessing `linear.weight` results in a new sample from the prior. Another thing that we should mention somewhere is that attribute access (for `PyroSample`, `PyroParam`) triggers usual pyro effects, which can sometimes be surprising (specially their interaction with `pyro.plate`) and means that unlike regular attributes, we need to be careful as to where we are accessing these.

   ```python
   with pyro.plate(“N”, N):
       x = linear.x # will have leading shape N.
   ```

fritzo:
> typo wrt "to_pyro_module_" above cell 5

@martinjankowiak can you point out the typo? I don't see it.

martinjankowiak:
end of the sentence starting "The third way to create a..."

fritzo:
Thanks for the detailed reviews!

neerajprad:
@fritzo - Sorry, I missed that you were planning to add more documentation to this tutorial in https://github.com/pyro-ppl/pyro/pull/2405. Do you want to do that as part of this PR, or is this good to merge?

fritzo:
Ready to merge, I'll add `@PyroSample` stuff in a follow-up PR.

Issue #2407, opened 2020-04-13
FR ContinuousBinomial, ContinuousBetaBinomial distributions


fritzo:
This issue proposes two distributions that could enable HMC and reparameterized SVI inference for models with count-valued latents, such as epidemiological [compartment models](https://en.wikipedia.org/wiki/Compartmental_models_in_epidemiology) (see [design doc](https://docs.google.com/document/d/1FmU3-8uAhUDyBDyD2Hyr69slJfm9x14tQPFto683VSw/edit#heading=h.u8o53kynqty9)).

Models such as SIR and SEIR typically use `Binomial` or overdispersed `BetaBinomial` distributions for transitions (or approximate these as `Censored(Poission)` or overdispersed `Censored(NegativeBinomial)`) and `Binomial` likelihoods (e.g. see [this prototype](https://github.com/pyro-ppl/pyro/blob/f952a43abded81e5f1925c2585a4c389b15d4049/tests/infer/test_smcfilter.py#L234)). In these cases, the `S,E,I` variables are integers. This issue proposes to replace the integers `total_count` with positive real numbers denoting a binary mixture model over the floor and ceiling of `total_count`. The semantics of the transition and likelihood distributions differ in that transition will be real->real, but likelihood will be real->integer.

## `ContinuousBinomial` likelihood

This is a simple two-component mixture model. Here's a sketch:

<details>

```python
class ContinuousBinomial(TorchDistribution):
    arg_constraints = {"total_count": constraints.positive,
                       "probs": constraints.positive}
    support = constraints.nonnegative_integer  # Note lack of upper bound.
    ...
    def sample(self, sample_shape=torch.Size()):
        lb = self.total_count.floor()
        probs = lb + 1 - self.total_count
        bern = probs.bernoulli()
        total_count = (lb + bern).long()
        return dist.Binomial(self, total_count, self.probs)
    def log_prob(self, value):
        # The first part is a mixture model.
        lb = self.total_count.floor().long()
        total_count = torch.stack([lb, ub + 1], dim=-1)
        logits = torch.stack([ub - self.total_count, self.total_count - lb], dim=-1).log()
        log_prob = dist.BetaBinomial(self.concentration0.unsqeeze(-1),
                                     self.concentration1.unsqeeze(-1),
                                     total_count).log_prob(value.unsqueeze(-1))
        # Unbounded support is required by actual applications.
        log_prob.masked_fill_(value.unsqueeze(-1) > ub, float(-inf))
[...]

PR #2409, opened 2020-04-14
Implement dequantized binomial distributions

fritzo:
Addresses #2407 

This implements a collection of new distributions to allow gradient-based inference for models with discrete latent count variables, especially `Binomial`. These distributions can be seen as introducing auxiliary Bernoulli variables and then collapsing out the discrete latent variables.
- `ExtendedBinomial` generalizes `Binomial` to a allow real-valued `total_count` parameter.
- `ExtendedBetaBinomial` generalizes `BetaBinomial` to a allow real-valued `total_count` parameter.
- `DequantizedDistribution(-)` converts a distribution over nonnegative integers into a distribution over positive real numbers. The motivating example is `DequantizedDistribution(ExtendedBetaBinomial(...))` for use in compartmental models.

This does not implement `.rsample()`, which is required for SVI; it will be added in a follow-up PR.

## Tested
- [x] plotted triangular distribution of `DequantizedDistribution` (see [notebook](https://github.com/pyro-ppl/sandbox/blob/3f58c9e2b958ceacf4987759087089efc1b7107d/2020-04-dequantized/dequantized_binomial.ipynb))
- [x] tests for dequantized `.sample()`
- [ ] generic distribution tests
- [ ] smoke tests for MCMC inference


pyro/distributions/quantized.py
```diff
@@ -0,0 +1,256 @@
...
+        lb = n.detach().floor()
+        ub = lb + 1
+        weights = n - lb
+        quantized = torch.cat([lb, ub], dim=-1)
+        return weights, quantized
+
+    def sample(self, sample_shape=torch.Size()):
+        shape = self._extended_shape(sample_shape)
+        weights, total_count = self._mixture
+        bern = weights.expand(shape + (1,)).bernoulli()
```
fehiepsi, line 117: I think you need to `squeeze` the result here.

pyro/distributions/quantized.py
```diff
@@ -0,0 +1,256 @@
...
+
+        # Model 2.
+        x ~ int_dist
+        y = dequantize(x, int_dist)
+        probs, counts = quantize(y)
+        b ~ Bernoulli(probs)
+        z = b + counts[..., 0]
+    """
+    neighbor = count.new_full(count.shape, 0.5)
+    neighbor.bernoulli_().mul_(2).add_(-1).add_(count).abs_()
```
fehiepsi, line 48: I think we should use `uniform_()` instead of `bernoulli_()` here; but even so, the math here is not quite convincing. I'll think more about this...

fritzo: I've rewritten this and factored it into the `DequantizedDistribution` class. I think the math is clearer now, but I still haven't written tests.

fehiepsi: Nice use of triangular distribution! I have tried to construct something similar (linear density on one side with predefined mean). I'll double check the math.

pyro/distributions/dequantized.py
```diff
@@ -0,0 +1,93 @@
...
+        new._validate_args = self._validate_args
+        return new
+
+    @torch.no_grad()
+    def sample(self, sample_shape=torch.Size()):
+        value = self.base_dist.sample(sample_shape)
+        # Add uniform triangular noise in [-1,1]
+        noise = value.new_empty(value.shape).uniform_(-0.25, 0.25)
+        sign = noise.sign()
+        noise.abs_().sqrt_().mul_(sign).add_(0.5)
```
fehiepsi, line 65: I think we add 0.5 before taking the sign

pyro/distributions/dequantized.py
```diff
@@ -0,0 +1,93 @@
...
+        batch_shape = torch.Size(batch_shape)
+        new.base_dist = self.base_dist.expand(batch_shape)
+        super(DequantizedDistribution, new).__init__(batch_shape, validate_args=False)
+        new._validate_args = self._validate_args
+        return new
+
+    @torch.no_grad()
+    def sample(self, sample_shape=torch.Size()):
+        value = self.base_dist.sample(sample_shape)
+        # Add uniform triangular noise in [-1,1]
```
fehiepsi, line 62: Looking like this will not make Model1 and Model2 equivalent:
p_model2 (z) = E_{u in (0, 1)} [u*p(z) + (1-u)*p(z-1)] + E_{u in (0, 1)} [u*p(z) + (1-u)*p(z+1)] = p(z) + (p(z-1) + p(z+1))/2 != p(z)

fritzo: Agreed, I think the best we can hope for is approximate equivalence with a good bound of 0.5 on earth mover distance error. It might be possible to implement distribution-specific dequantized distributions with tighter bounds, but I think it will be difficult to make their gradients well-behaved at the boundaries (0 and total_count).

pyro/distributions/dequantized.py
fritzo: Good catch. I think it's -0.5 though, right?

fritzo: OK, I've fixed and added a plot to the PR description (after Tested section).

tests/infer/mcmc/test_valid_models.py
```diff
@@ -463,6 +468,65 @@ def model():
...
                         dist.ExtendedBinomial(i, response_rate),
                         obs=data)
+        logger.debug(" ".join(map("{:0.3g}".format, i)))
+
+    mcmc_kernel = kernel(model, max_plate_nesting=1, **kwargs)
+    assert_ok(mcmc_kernel)
+
+
+@pytest.mark.parametrize("kernel, kwargs", [
+    (HMC, {"adapt_step_size": True, "num_steps": 3}),
```
fehiepsi, line 478: @fritzo you can set `trajectory_length=1` to reduce the number of `num_steps`. In HMC, `num_steps` is a dynamic value:
```
num_steps = trajectory_length / step_size
```

pyro/distributions/dequantized.py
fehiepsi: I think we can come up with an equivalent procedure by uniformly generating the noise in the interval (-C / p(n), C / p(n)) where C is small enough such that the interval belongs to (-1, 1) for all `n`. I'll make a notebook to show the equivalent after the dinner.

fritzo: Oh good point, `torch.uniform(0,1) + torch.uniform(-1,0)` is triangular :smile:

fehiepsi: My point is as long as we use a noise with the density mean proportional to 1 / p(n), then quantize(dequantize(x)) has the same distribution as x. The uniform noise as in my last comment preserves that property but dequantize(x) is discontinuous. We can use something like trapezoidal but the density is not monotone (it bends up and down). The triangular noise is nice but it does not preserve that relationship. I am a bit worried about this inconsistency
```python
d = dist.DequantizedDistribution(dist.Poisson(0.1))
x = torch.linspace(-1, 5, 1000)
z = d.sample((1000000,))
z = (z - z.floor()).bernoulli() + z.floor()
plt.figure(figsize=(8, 3), dpi=300).patch.set_color('white')
plt.hist(z, bins=200, range=(-1,5), alpha=0.5)
plt.hist(d.base_dist.sample((1000000,)), bins=200, range=(-1,5), alpha=0.5);
```

pyro/distributions/dequantized.py
```diff
@@ -0,0 +1,82 @@
...
+
+    For example the following two models result in approximately identical
+    distributions over ``z``, conditioned on inputs ``n, p``.
+
+        # Model 1.
+        z ~ Binomial(n, p)
+
+        # Model 2.
+        r ~ DequantizedDistribution(Binomial(n, p))
+        b ~ Bernoulli(1 + floor(r) - r)      # Quantize.
```
fehiepsi, line 33: I think this is `r - floor(r)`, e.g. when r ~ 4.99, the Bernoulli prob should be ~ 0.99

pyro/distributions/dequantized.py
```diff
@@ -0,0 +1,82 @@
...
+        new._validate_args = self._validate_args
+        return new
+
+    @torch.no_grad()
+    def sample(self, sample_shape=torch.Size()):
+        value = self.base_dist.sample(sample_shape)
+        # Add uniform triangular noise in [-1,1]
+        noise = value.new_empty(value.shape + (2,)).uniform_(-0.5, 0.5).sum(-1)
+        # Ensure result is nonnegative.
+        return noise.sub_(value).abs_()
```
fehiepsi, line 65: nit: `add_` is better than `sub_`

pyro/distributions/dequantized.py
```diff
@@ -0,0 +1,82 @@
...
+        with torch.no_grad():
+            lb = value.floor()
+            ub = lb + 1
+            quantized = torch.stack([lb, ub])
+
+        log_prob = self.base_dist.log_prob(quantized)
+        # Account for asymmetry at the interval [0,1].
+        log_prob = torch.where(quantized == 0, log_prob + math.log(2), log_prob)
+        # Allow unbounded support (required by many applications).
+        log_prob = log_prob.masked_fill((value < 0) | (quantized > ub), -math.inf)
```
fehiepsi, line 80: when `quantized > ub` happens? I guess you want to account the case `base_dist` has bounded support? If so, we can disable `validate_args` at `log_prob = self.base_dist.log_prob(quantized)` and make a check `self.base_dist.support.check(quantized)` here instead of `quantized > ub`

pyro/distributions/dequantized.py
```diff
@@ -0,0 +1,82 @@
...
+        ``.support == constraints.nonnegative_integer`` and
+        ``.event_shape == ()``.
+    :type base_dist: ~pyro.distributions.TorchDistribution
+    """
+    arg_constraints = {}
+    support = constraints.positive  # Note lack of upper bound.
+    has_rsample = False  # TODO implement .rsample() based on base_dist.cdf().
+
+    def __init__(self, base_dist, validate_args=None):
+        self.base_dist = base_dist
```
fehiepsi, line 48: nit: add an assertation `assert base_dist.event_shape == ()`

pyro/distributions/dequantized.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # Model 2.
+        r ~ DequantizedDistribution(Binomial(n, p))
+        b ~ Bernoulli(1 + floor(r) - r)      # Quantize.
+        z = floor(r) + b
+
+    Earth mover distance error is upper bounded by 1/3.
+
+    :param base_dist: A distribution with
+        ``.support == constraints.nonnegative_integer`` and
```
fehiepsi, line 39: I think `constraints.integer_interval` is also valid

pyro/distributions/extended.py
```diff
@@ -0,0 +1,155 @@
...
+    components. If ``n = total_count`` then the two components are:
+
+    - ``Binomial(floor(n), probs)`` with weight ``1 + floor(n) - n``, and
+    - ``Binomial(floor(n) + 1, probs)`` with weight ``n - floor(n)``.
+
+    The following two models result in identical distributions over ``z``,
+    conditioned on inputs ``n, p``.
+
+        # Model 1.
+        b ~ Bernoulli(1 + floor(n) - n)         # Quantize.
```
fehiepsi, line 30: I think `1 + floor(n) - n` should be `n - floor(n)`

pyro/distributions/dequantized.py
fehiepsi: On second thought, I think it can be fine to use `triangular` here. Could you add a note so we can revisit in the future? for example,
```
# we use triangular noise here instead of other candidates because
#   + the density will be continuous
#   + the density will have similar monotonicity as base_dist
#   + the density is easy to compute
```

pyro/distributions/dequantized.py
fritzo: Most of the math is similar but we would need a separate implementation to handle `integer_interval` since `nonnegative_integer` needs special handling at zero (e.g. the `.abs()` below). I think the cleanest way to do this is to choose good names for the two distributions, e.g. `DequantizedCount` and `DequantizedInt` or something. Do you have recommendations?

pyro/distributions/dequantized.py
fritzo: Good point. For the SEIR use case we will be using the `Extended*` distributions which intentionally avoid an upper bound, so I think this won't be an issue.

More generally, I think we would need a third `Dequantized` class to handle `integer_interval`. Maybe we name them `RelaxedInteger`, `RelaxedIntegerInterval` and `RelaxedNonnegativeInteger`?

fehiepsi: I don't see why we need another class for integer_interval. I think we can just simply assume that `log_prob(x > upper_bound) = -inf` for those enumerate supported distributions and consider its support as `nonnegative_integer`. WDYT?

fritzo: Yes, I think that makes sense. If the auxiliary variable method #2410 works out, I think we can replace `DequantizedDistribution` with the simple `quantize()` function in that PR. Then following your idea I think we can use the same `quantize()` function for bounded and unbounded logic; we'll just need to set the appropriate bounds of the input variables and clamp the output, e.g.
```py
lb, ub = ...  # integers
x_aux ~ Uniform(lb - 0.5, ub + 0.5)
x = quantize("x", x).clamp(min=lb, max=ub)
```

fritzo:
Closing in favor of #2410 . I think all the math from this PR can be moved to that PR.
@fehiepsi thanks again for the careful review and push to find an exact method!

PR #2410, opened 2020-04-15
Example: discrete SIR models with HMC inference

fritzo:
Addresses #2426 
Replaces #2407, #2409

This adds an extended example using HMC for a simple epidemiological model: SIR with discrete time and discrete observations. This uses an auxiliary variable method to reparameterize the discrete latent dynamic state as a continuous random variable amenable to gradient-based inference.

I plan to treat this example as documentation and experimentation for a subsequent `pyro.contrib.epidemic` module that provides pre-built models for SIR, SEIR, and variants. That module will be more complex, including richer transition and observation distributions.

Here is the [derivation](https://github.com/fritzo/notebooks/blob/master/cubic_spline_dist.ipynb) for the cubic spline distribution.

See [generated tutorial](http://fritzo.org/temp/pyro/html/sir_hmc.html).

## Tested
- [x] smoke test locally: `python sir_hmc.py --population=10 --enum`
- [x] smoke test locally: `python sir_hmc.py --population=10000 --sequential`
- [x] smoke test locally: `python sir_hmc.py --population=10000 --sequential --dct`
- [x] smoke test locally: `python sir_hmc.py --population=10000`
- [x] smoke test locally: `python sir_hmc.py --population=10000 --duration=100`
- [x] smoke test locally: forecasting and plotting
- [x] added to test_examples.py (costs under 20 seconds)

## Experiments

- [x] compare discrete versus continuous inference results

<details>

```sh
% python examples/sir_hmc.py --enum --jit  # discrete
Generated data:
0 0 1 0 1 0 0 1 2 1
...
response_rate: truth = 0.5, estimate = 0.747 ± 0.179
infection_rate: truth = 0.3, estimate = 0.601 ± 0.205
recovery_rate: truth = 0.3, estimate = 0.162 ± 0.129
(pyro) (fritzo@fritz-air-2:~/pyro) [sir-hmc]
```
```sh
% python examples/sir_hmc.py --jit  # continuous
Generated data:
0 0 1 0 1 0 0 1 2 1
...
response_rate: truth = 0.5, estimate = 0.741 ± 0.123
infection_rate: truth = 0.3, estimate = 0.618 ± 0.21
recovery_rate: truth = 0.3, estimate = 0.212 ± 0.141
```

</details>

- [x] test spline order 1 versus 3. Results: order 3 results in ~10x faster mixing (per iteration) but ~2x slower.

<details>

Linear spline:
linear-spline-mcmc

Cubic spline:
cubic-spline-mcmc

Linear stats:
[...]

PR #2414, opened 2020-04-15
Add extended Binomial and BetaBinomial distributions

fritzo:
Addresses #2407, #2410 
Replaces #2409

This adds two new distributions: `ExtendedBinomial` and `ExtendedBetaBinomial` plus a new `constraints.integer`. These allow some type checking, while setting `log_prob` to `-inf` for values outside of the extended support.

I've marked these EXPERIMENTAL because we may be able to implement this as a mixin class factory similar to `PyroModule[-]`, i.e. we may be able to replace them with `Extended[Binomial]`. However that seems like too much meta magic for a first PR.

## Tested
- [x] standard distribution tests
- [x] tests on values outside of support
- [x] death tests on invalid inputs


pyro/distributions/extended.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    EXPERIMENTAL :class:`~pyro.distributions.BetaBinomial` distribution
+    extended to have logical support the entire integers. Numerical support is
+    still the integer interval ``[0, total_count]``
+    """
+    support = constraints.integer
+
+    def log_prob(self, value):
+        result = super().log_prob(value)
+        invalid = ~super().support.check(value)
```
neerajprad, line 36: IIUC..without `validate_args`, this should return `-inf` for values outside the support, but will also not do any sample validation (e.g. check that the values are non-integers). So this change lets us use our existing sample validation logic but relaxes it for `log_prob` so that any integer values outside the support are assigned `-inf`?

fritzo: Correct. The difference from NumPyro is that, when `validate_args=True`, there is still some (slightly weaker) validation of shapes and integer values.

fritzo:
Thanks for reviewing!

PR #2416, opened 2020-04-16
Fix retry logic in MCMC _get_init_params()

fritzo:
The original logic tries to gather a set of size `num_chains` each attempt, even if some samples already exist. This version is cheaper and easier to understand.

### DIFF
--- a/pyro/infer/mcmc/util.py
+++ b/pyro/infer/mcmc/util.py
@@ -310,7 +310,7 @@ def _get_init_params(model, model_args, model_kwargs, transforms, potential_fn, 
 
     params_per_chain = defaultdict(list)
     num_found = 0
-    for attempt in range(num_chains + max_tries_initial_params):
+    for attempt in range(num_chains * max_tries_initial_params):
         if strategy == "uniform":
             params = {k: dist.Uniform(v.new_full(v.shape, -2), v.new_full(v.shape, 2)).sample()
                       for k, v in params.items()}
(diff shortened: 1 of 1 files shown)


pyro/infer/mcmc/util.py
```diff
@@ -303,32 +303,32 @@ def get_potential_fn(self, jit_compile=False, skip_jit_warnings=True, jit_option
...
-                for k, v in params.items():
-                    params_per_chain[k].append(v)
-                n += 1
-        if num_chains == 1:
-            return {k: v[0] for k, v in params_per_chain.items()}
-        else:
-            return {k: torch.stack(v) for k, v in params_per_chain.items()}
+    params_per_chain = defaultdict(list)
+    num_found = 0
+    for attempt in range(num_chains * max_tries_initial_params):
```
fehiepsi, line 313: I think the difference is at this line: before, we use `max_tries_initial_params`, now we use `num_chains * max_tries_initial_params`?

fehiepsi: > The original logic tries to gather a set of size num_chains each attempt

No, we didn't reset `n` after exiting the `while` loop.

fehiepsi: Opss, I get your point. The old logic is wrong (my bad).

fritzo: Oh, I see, thanks for pointing that out!

Issue #2415, opened 2020-04-16
API issues with MCMC initialization


neerajprad:
- [x] We should be able to provide `initial_params` in the constrained space which makes for a more natural choice.
 - [x] We use the `initialize_model` utility at a few places including within the `MCMC` class and the `HMCKernel` class. This function is used to get the value for the transforms, convert a model representation into a potential energy function, and get initial parameter values. If a user specifies an explicit value for `initial_params` and we call this function to get transforms or potential energy function, we shouldn't error if a viable initial parameter was not found. 
 - [x] Expose initialization strategy, e.g. sampling from uniform in the unconstrained space or the prior within the `MCMC` class.

cc. @fritzo, @fehiepsi

neerajprad:
@fehiepsi - what do you think about refactoring out finding the initial parameters into its own function? That will make it easier to skip it if the user provides valid initial params.

fehiepsi:
Agreed! We can even use `autoguide` initialization as in NumPyro. I'll make a PR for it.

fritzo:
Maybe add a new `init_params` in the constrained space and deprecate the `initial_params`, thus preserving backwards compatibility? Kind of hacky tho.

neerajprad:
Closing this issue - it looks like #2417 and #2420 should have resolved all of these. Regarding (1), we need to use init_strategy instead of `initial_params` to specify initialization in the constrained space, since the latter is also used when a user passes in a potential energy function directly. Feel free to reopen if there are any related issues that you observe.

PR #2417, opened 2020-04-16
Expose initialization strategy in HMC

fehiepsi:
Addresses part of https://github.com/pyro-ppl/pyro/issues/2415.

There is one issue is: if users want to change keywords of init strategies, they have to use `partial`, e.g. `partial(init_to_median, num_samples=20)`. I find that it is a bit inconvenient for users. In NumPyro, users can set `init_strategy=` `init_to_median(20)` or `init_to_uniform(3)`, `init_to_value({"x": 10})`, `init_to_feasible()`.

### DIFF
--- a/pyro/infer/autoguide/__init__.py
+++ b/pyro/infer/autoguide/__init__.py
@@ -6,7 +6,9 @@ from pyro.infer.autoguide.guides import (AutoCallable, AutoContinuous, AutoDelta
                                          AutoLaplaceApproximation, AutoLowRankMultivariateNormal,
                                          AutoMultivariateNormal, AutoNormal, AutoNormalizingFlow)
 from pyro.infer.autoguide.utils import mean_field_entropy
-from pyro.infer.autoguide.initialization import init_to_feasible, init_to_mean, init_to_median, init_to_sample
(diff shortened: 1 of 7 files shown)


fehiepsi:
@fritzo `initial_params` is originally useful when users use `potential_fn` instead of `model`.

> Re: usability, you could use the standard poutine trick:

Thanks, yeah we should do it!

fehiepsi:
> Do I understand correctly that this allows me to specify a subset of init_params to do so in the constrained space

Yes, you just need to specify a subset. The remaining params use the `uniform` strategy (default in MCMC).

neerajprad:
@fritzo - does this address your concern of initialization code being triggered at various points in the flow despite the user specifying initial params?

fehiepsi:
@neerajprad This does not. I think we should address it.

fehiepsi:
I am fixing some typos. I'll address that too.

neerajprad:
Thanks for clarifying, @fehiepsi. That might involve some refactoring, look forward to seeing your PR. Let me know if I can help.

fritzo:
> does this address your concern of initialization code being triggered

Yes this partially addresses it. The problem was, the model was running 100 random initializations even if I provided a single good one. After This PR, if I provide a single good initialization through `init_to_value`, no random initializations are sampled.

PR #2418, opened 2020-04-16
Fix backward printing in warn_if_nan, warn_if_inf

fritzo:
This ensures that backward nans warn with the correct (filename,lineno).

This also makes the `warn_if_*(-)` functions return their value, so it can be used in-line as in:
```diff
- x = y.log()
+ x = warn_if_nan(y.log())
```
and that line is now printed in the warning message.

This also updates to replace `float('inf')` with `math.inf`.

## Tested
- [x] verified output locally
- [x] updated tests

### DIFF
--- a/pyro/util.py
+++ b/pyro/util.py
@@ -2,15 +2,17 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import functools
+import math
 import numbers
 import random
+import sys
 import timeit
 import warnings
 from collections import defaultdict
+from contextlib import contextmanager
 from itertools import zip_longest
 
 import torch
-from contextlib import contextmanager
 
 from pyro.poutine.util import site_is_subsample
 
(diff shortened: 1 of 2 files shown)


pyro/util.py
```diff
@@ -62,34 +64,66 @@ def torch_isinf(x):
...
     """
+    if filename is None:
+        try:
+            frame = sys._getframe(1)
+        except ValueError:
+            filename = "sys"
+            lineno = 1
+        else:
+            filename = frame.f_code.co_filename
+            lineno = frame.f_lineno
```
fritzo, line 84: Compare this logic to [warnings.py](https://github.com/python/cpython/blob/518835f3354d6672e61c9f52348c1e4a2533ea00/Lib/warnings.py#L298). Note `._getframe()` is private because it is a CPython detail, but PyTorch is limited to CPython.

PR #2420, opened 2020-04-16
avoid finding init_params when unnecessary

fehiepsi:
Addresses #2415.

This copies changes by @fritzo in #2410 to fix the issue. Thanks!

I also fixed some typos in `modules` tutorial.

### DIFF
--- a/pyro/infer/mcmc/api.py
+++ b/pyro/infer/mcmc/api.py
@@ -412,7 +412,8 @@ class MCMC:
             elif self.kernel.model:
                 _, _, self.transforms, _ = initialize_model(self.kernel.model,
                                                             model_args=args,
-                                                            model_kwargs=kwargs)
+                                                            model_kwargs=kwargs,
+                                                            initial_params={})
             # Assign default value
             else:
                 self.transforms = {}
--- a/pyro/infer/mcmc/util.py
+++ b/pyro/infer/mcmc/util.py
@@ -403,11 +403,11 @@ def initialize_model(model, model_args=(), model_kwargs={}, transforms=None, max
     trace_prob_evaluator = TraceEinsumEvaluator(model_trace,
                                                 has_enumerable_sites,
                                                 max_plate_nesting)
-    prototype_params = {k: transforms[k](v) for k, v in prototype_samples.items()}
 
     pe_maker = _PEMaker(model, model_args, model_kwargs, trace_prob_evaluator, transforms)
 
     if initial_params is None:
+        prototype_params = {k: transforms[k](v) for k, v in prototype_samples.items()}
         # Note that we deliberately do not exercise jit compilation here so as to
         # enable potential_fn to be picklable (a torch._C.Function cannot be pickled).
         initial_params = _find_valid_initial_params(model, model_args, model_kwargs, transforms,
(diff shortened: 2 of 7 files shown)


fehiepsi:
@fritzo Even with #2417, we still have unnecessary procedure to find valid initial_params for multi-chains.

pyro/infer/mcmc/api.py
```diff
@@ -412,7 +412,8 @@ def model(data):
             elif self.kernel.model:
                 _, _, self.transforms, _ = initialize_model(self.kernel.model,
                                                             model_args=args,
-                                                            model_kwargs=kwargs)
+                                                            model_kwargs=kwargs,
+                                                            initial_params={})
```
fritzo, line 416: Do you want to default this to `None` rather than `{}`?

pyro/infer/mcmc/util.py
```diff
@@ -331,7 +331,7 @@ def _find_valid_initial_params(model, model_args, model_kwargs, transforms, pote
 
 def initialize_model(model, model_args=(), model_kwargs={}, transforms=None, max_plate_nesting=None,
                      jit_compile=False, jit_options=None, skip_jit_warnings=False, num_chains=1,
-                     init_strategy=init_to_uniform):
+                     init_strategy=init_to_uniform, initial_params=None):
```
fehiepsi, line 334: Another way is to make a hack: if num_chains=0, we return initial_params=None. So we don't have to introduce this new arg.

pyro/infer/mcmc/api.py
fehiepsi: If we set this to None, it will try to find valid initial params.

pyro/infer/mcmc/util.py
```diff
@@ -405,13 +407,14 @@ def initialize_model(model, model_args=(), model_kwargs={}, transforms=None, max
...
 
-    # Note that we deliberately do not exercise jit compilation here so as to
-    # enable potential_fn to be picklable (a torch._C.Function cannot be pickled).
-    init_params = _find_valid_initial_params(model, model_args, model_kwargs, transforms,
-                                             pe_maker.get_potential_fn(), prototype_params,
-                                             num_chains=num_chains, init_strategy=init_strategy)
+    if initial_params is None:
+        # Note that we deliberately do not exercise jit compilation here so as to
+        # enable potential_fn to be picklable (a torch._C.Function cannot be pickled).
+        initial_params = _find_valid_initial_params(model, model_args, model_kwargs, transforms,
```
neerajprad, line 413: We can move `prototype_params` also inside this block.

neerajprad:
> Even with #2417, we still have unnecessary procedure to find valid initial_params for multi-chains.

Could you provide some more context on what happens with mp? Maybe we can track it in the main issue if it is hard to resolve in this PR.

pyro/infer/mcmc/api.py
```diff
@@ -412,7 +412,8 @@ def model(data):
             elif self.kernel.model:
```
fehiepsi, line 412: @neerajprad In mp, I think `self.kernel.transforms` will be None, so we need this branch to define `self.transforms`.

neerajprad: I see, but with your PR this won't try to find valid initial parameter set again, right?

neerajprad: Just getting the transforms should be fast enough.

pyro/infer/mcmc/api.py
fritzo: Thanks for explaining!

pyro/infer/mcmc/api.py
fehiepsi: Yes, I think this PR solves all the known issues.

neerajprad: That's great! This was actually simpler than what I was expecting. Nicely done.

pyro/infer/mcmc/util.py
fehiepsi: Good catch, thanks!

PR #2421, opened 2020-04-16
fix DPMM tutorial and update make scrub

fehiepsi:
Currently, [DPMM tutorial](http://pyro.ai/examples/dirichlet_process_mixture.html) execution seems to be in arbitrary order. In addition, the notebook gives errors for Poisson likelihood models. Actually, the data points are not integers. To fix the issue, I added `.round()` at
+ `data = torch.tensor(df['sunspot.year'].values, dtype=torch.float32)`
-> `data = torch.tensor(df['sunspot.year'].values, dtype=torch.float32).round()`
+ `data = torch.tensor(df.drop(df_test.index)['sunspot.year'].values, dtype=torch.float)`
-> `data = torch.tensor(df.drop(df_test.index)['sunspot.year'].values, dtype=torch.float).round()`,
which gives quite similar results.

The diff is quite large because the current notebook is copied from colab, which has been formatted a bit different from the usual notebook. The changes only include the above `.round()` ops and reorder imports. I have rerun the notebook locally for this PR.


fritzo:
Thanks for fixing this! Culd you resolve merge conflicts?

fehiepsi:
Thanks for reviewing, @fritzo !

PR #2422, opened 2020-04-17
Stop using deprecated PyTorch function signatures.

gchanan:
Various function overloads were previously "deprecated" in PyTorch in that they didn't show up as options in documentation or error messages.
 
These functions are listed here: https://github.com/pytorch/pytorch/blob/release/1.5/tools/autograd/deprecated.yaml.

PyTorch 1.5 now makes these UserWarnings, which trips up test_cevae because the pytest configuration is to error on non-Deprecation warnings.

This change should have no other effect.

### DIFF
--- a/pyro/optim/clipped_adam.py
+++ b/pyro/optim/clipped_adam.py
@@ -68,11 +68,11 @@ class ClippedAdam(Optimizer):
                 state['step'] += 1
 
                 if group['weight_decay'] != 0:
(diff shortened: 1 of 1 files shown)


gchanan:
thanks for merging, I only fixed the issues that caused test failures.  I'll add another PR for adagrad_rmsprop.py.

fritzo:
Oh thanks!

PR #2425, opened 2020-04-20
Implement structured mass matrix in HMC

fehiepsi:
Requested by @fritzo. This is also a solution for the issue https://github.com/pyro-ppl/numpyro/pull/536.

For models as in #2410, we are most interested in global parameters. It is likely that those global parameters are correlated. We would need a scheme that can adapt a dense mass matrix of global parameters and adapt diagonal mass for remaining "auxiliary" variables.

In this PR, I add an additional semantic for `full_mass` args. If they are True or False, the behavior is the same as before. Otherwise, users can specify a list of tuples of site names for it. For example, assume that the model has latent variables "x", "y", "z", the meaning of each structure is defined as follows:
+ `full_mass=[("x", "y")]`: using dense mass matrix for the joint (x, y), diagonal mass matrix for z
+ `full_mass=[]` (or `full_mass=False`): using diagonal mass matrix for the joint (x, y, z)
+ `full_mass=[("x", "y", "z")]` (or `full_mass=True`): using dense mass matrix for the joint (x, y, z))
+ `full_mass=[("x",), ("y",), ("z")]`: using dense mass matrices for each of x, y, z.

### Changes
+ Now, `kernel.inverse_mass_matrix` is an OrderedDict that maps tuples of site names to the correspondent inverse mass matrices. Users can still set `kernel.adapter.inverse_mass_matrix = tensor` (this is not important though) as before but we might deprecate in the future.
+ `WarmupAdapter.is_diag_mass` is changed to `dense_mass` to incorporate structured mass matrix.

### TODO
- [x] add tests


fritzo:
Interface looks great!

tests/infer/mcmc/test_nuts.py
```diff
@@ -380,3 +380,31 @@ def model(data):
...
+    cov[2, 2] = z_var
+
+    # smoke tests
+    for dense_mass in [True, False]:
+        kernel = NUTS(model, jit_compile=True, ignore_jit_warnings=True, full_mass=dense_mass)
+        mcmc = MCMC(kernel, num_samples=1, warmup_steps=1)
+        mcmc.run(cov)
+        assert kernel.inverse_mass_matrix[("x", "y", "z")].dim() == 1 + int(dense_mass)
+
+    kernel = NUTS(model, jit_compile=True, ignore_jit_warnings=True, full_mass=[("x", "y")])
```
fritzo, line 406: Could we make this a little more complex and add a `w` variable that is full, so `full_mass=[("w",), ("x","y")]`? Then `w` will be full, `(x,y)` will be full, and `z` will be diagonal? This seems like the most general block diagonal structure.

pyro/infer/mcmc/hmc.py
```diff
@@ -251,14 +258,32 @@ def _initialize_model_properties(self, model_args, model_kwargs):
...
-            initial_mass_matrix = eye_like(site_value, mass_matrix_size)
+            msg = "full_mass should be a list of tuples of site names."
+            dense_sites_list = self._adapter.dense_mass
+            assert isinstance(dense_sites_list, list), msg
+            for dense_sites in dense_sites_list:
+                assert dense_sites and isinstance(dense_sites, tuple), msg
+                for name in dense_sites:
+                    assert isinstance(name, str) and name in self.initial_params, msg
+        diag_sites = tuple(sorted([name for name in self.initial_params
+                                   if name not in set().union(*dense_sites_list)]))
```
fritzo, line 274: nit: avoid computing `set().union()` each time through the loop, using `dense_sites_set`.

tests/infer/mcmc/test_nuts.py
fehiepsi: Sure, I'll add.

pyro/infer/mcmc/hmc.py
```diff
@@ -251,14 +258,32 @@ def _initialize_model_properties(self, model_args, model_kwargs):
...
+            dense_sites_list = [tuple(sorted(self.initial_params))]
         else:
-            initial_mass_matrix = eye_like(site_value, mass_matrix_size)
+            msg = "full_mass should be a list of tuples of site names."
+            dense_sites_list = self._adapter.dense_mass
+            assert isinstance(dense_sites_list, list), msg
+            for dense_sites in dense_sites_list:
+                assert dense_sites and isinstance(dense_sites, tuple), msg
+                for name in dense_sites:
+                    assert isinstance(name, str) and name in self.initial_params, msg
```
fritzo, line 272: Can we assert there are no duplicates? I think that's as easy as
```py
for dense_sites in dense_sites_list:
    ...
    assert len(dense_sites) == len(set(dense_sites_))
    ...
dense_sites_set = set().union(*dense_sites_list)
assert len(dense_sites_set) == sum(map(len, dense_sites_list))
```

pyro/infer/mcmc/hmc.py
```diff
@@ -201,16 +204,20 @@ def _find_reasonable_step_size(self, z):
...
-        for name, param in sorted(self.initial_params.items()):
-            next_pos = pos + param.numel()
-            r[name] = r_flat[pos:next_pos].reshape(param.shape)
-            pos = next_pos
-        assert pos == r_flat.size(0)
-        return r, r_flat
+        r_flat_dict = {}
+        for site_names in self.inverse_mass_matrix:
+            r_flat = pyro.sample(
+                name + "_" + ",".join(site_names), self._adapter.r_dist[site_names])
```
fritzo, line 211: This could result in name conflict, e.g. if we have names `loc_scale` and `loc` and `scale` (yes I've done that 😊). Do you think it's worth working around with positional naming, e.g.
```py
for i, site_name in enumerate(self.invere_mass_matrix):
    ... pyro.sample("{}_r_{}".format(name, i), ...)
```

pyro/infer/mcmc/hmc.py
fehiepsi: This is asserted in 3 lines below.

pyro/infer/mcmc/hmc.py
fehiepsi: Good point! I was worried that there is a name conflict here. I think we can just simply use the name `"{}_{}".format(name, str(site_names))` here, so if `site_names` is `("x", "y", "z")`, the site will have name `name_('x', 'y', 'z')`.

fritzo:
Yikes @fehiepsi I would like to make a full mass matrix between three scalar variables and the first element of a bit tensor (the other elements of that tensor can be diagonal). What should I do?

fehiepsi:
> I guess I could implement this via a UncatReparam

Yes, I think this helper reparam can solve your usage case. Otherwise, you can manually `uncat` the required latent.

PR #2427, opened 2020-04-20
Add header check to lint stage

fritzo:
This fixes a bug in update_headers.py and adds an automatic check to the lint stage on travis letting contributors know if they need to `make license`. If this ever becomes too flaky I'm happy to remove that check.

### DIFF
--- a/scripts/update_headers.py
+++ b/scripts/update_headers.py
@@ -1,8 +1,10 @@
 # Copyright Contributors to the Pyro project.
 # SPDX-License-Identifier: Apache-2.0
 
-import os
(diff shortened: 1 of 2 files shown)


scripts/update_headers.py
```diff
@@ -40,7 +47,6 @@
         lineno += 1
         while lines[lineno].startswith(comment.format("Copyright")):
             lineno += 1
-            changed = True
```
fritzo, line 43: This bug was triggering on spline.py.

fritzo:
bump

Issue #2426, opened 2020-04-20
FR Discrete compartmental models for epidemiology


fritzo:
This issue tracks tasks related to discrete compartmental models.
See [design doc](https://docs.google.com/document/d/1FmU3-8uAhUDyBDyD2Hyr69slJfm9x14tQPFto683VSw/edit?usp=sharing) for motivation. 

## Modeling
- [x] #2410 Initial prototype using HMC + enumeration
- [x] #2437 Implement a framework in `pyro.contrib.epidemiology`
- [x] #2446 Implement an SEIR model @fritzo
- [x] #2450 #2451 Generalize to overdispersed noise via `BetaBinomial` @fritzo
- [x] #2457 Handle observations accumulated over an interval
- [x] #2460 Allow stochastic init state and start date estimation
- [x] #2466 Add a regional model
- [x] #2474 Add a `CoalescentTimes` distribution
- [x] #2514 Eliminate need for `ComparmentalModel.transition_bwd()`
- [x] #2517 Add a model with time-varying global parameters
- [x] #2558 ~~#2468~~ Incorporate phylogeny observations into the likelihood (see [OverdispersedSEIRModel](https://github.com/pyro-ppl/pyro/blob/sir-phylogenetic-likelihood/pyro/contrib/epidemiology/seir.py))
- [x] #2558, #2559 Add example of branching flows #2552
- [ ] Add example of causal intervention (ideally [interactive on nbviewer](https://minrk-ipywidgets.readthedocs.io/en/latest/embedding.html#rendering-interactive-widgets-on-nbviewer))
- [ ] Add example of cyclic flows #2552
- [ ] Support enumeration of discrete globals, e.g. to condition on multiple phylogenetic samples

## Inference
- [x] #2428 Use block diagonal mass matrix after #2425 is merged 
- [x] #2429 Fix global parameters that are easy to physically measure
- [x] #2430 Try a frequency-smoothed DCT transform @fritzo
- [x] #2434 Try larger enumeration windows @martinjankowiak 
- [x] #2452 Use SMC to generically initialize to a feasible trajectory
- [x] #2465 Try structured adaptation matrices in HMC @fehiepsi 
- [x] #2492 #2493 Try a wavelet transform @martinjankowiak 
- [x] #2495 Implement a `SplitReparam` to split off low-frequency components of Haar reparams
- [x] #2510 #2513 #2522 Try continuous relaxed inference
- [x] #2498 Make all `Binomial` distributions overdispersed
[...]

PR #2434, opened 2020-04-22
add quintic spline quantization strategy to sir_hmc.py

martinjankowiak:
addresses #2426

### DIFF
--- a/examples/sir_hmc.py
+++ b/examples/sir_hmc.py
@@ -317,7 +317,7 @@ def continuous_model(data, population, args):
     for t, datum in poutine.markov(enumerate(data)):
         S_prev, I_prev = S_curr, I_curr
         S_curr = quantize("S_{}".format(t), S_aux[..., t], min=0, max=population, spline_order=args.spline_order)
-        I_curr = quantize("I_{}".format(t), I_aux[..., t], min=0, max=population, spine_order=args.spline_order)
+        I_curr = quantize("I_{}".format(t), I_aux[..., t], min=0, max=population, spline_order=args.spline_order)
 
         # Now we reverse the computation.
         S2I = S_prev - S_curr
--- a/examples/sir_hmc.py
+++ b/examples/sir_hmc.py
@@ -73,12 +73,12 @@ def global_model(population):
     return rate_s, prob_i, rho
 
 
-def discrete_model(data, population):
+def discrete_model(args, data):
     # Sample global parameters.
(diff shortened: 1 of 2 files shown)


martinjankowiak:
lmk if i should add new args to test_examples

fritzo:
Could you add test args to `test_examples.py`:
```py
'sir_hmc.py -t=2 -w=2 -n=4 -d=100 -p=10000 -f 2 -s0=5',
```

examples/sir_hmc.py
```diff
@@ -242,35 +242,64 @@ def hook_fn(kernel, *unused):
...
+            302 - 100 * ss + 10 * ssss,
+            302 - 100 * tt + 10 * tttt,
+            55 + 115 * s + 70 * ss - 9 * sss - 25 * ssss + 7 * sssss,
+            2 + 10 * s + 20 * ss + 20 * sss + 10 * ssss - 7 * sssss,
+            2 * sssss
+        ], dim=-1) * (1/840)
+
+    q = pyro.sample("Q_" + name, dist.Categorical(probs)).type_as(x_real)
+
+    x = lb + q - 1 if spline_order == 3 else lb + q - 3
```
fritzo, line 292: nit: consider simplifying to 
```py
x = lb + q - {3: 1, 5: 3}[spline_order]
```

examples/sir_hmc.py
```diff
@@ -242,35 +242,64 @@ def hook_fn(kernel, *unused):
...
     x = torch.min(x, 2 * max + 1 - x)
+
     return pyro.deterministic(name, x)
 
 
 # Now we can define another equivalent model.
 
 @config_enumerate
-def continuous_model(data, population):
+def continuous_model(data, population, args):
```
fritzo, line 302: nit: Since this is user facing, could you please clean up so that all models have the same signature and that signature is `(data, args)`, replacing `population` by `args.population` below?

examples/sir_hmc.py
martinjankowiak: done

examples/sir_hmc.py
martinjankowiak: done

examples/sir_hmc.py
martinjankowiak: done

examples/sir_hmc.py
martinjankowiak: done please check piping in case i missed anything

PR #2436, opened 2020-04-23
Added permutation of `AutoRegressiveNN` to buffer

stefanwebb:
The `ConditionalAutoRegressiveNN` and `AutoRegressiveNN` classes did not register their permutation matrices in a buffer, which means that part of the object state will not get saved/loaded and introduces the potential for subtle bugs


stefanwebb:
I think it may not be worth the effort in this case... You couldn't write a test that generalizes across the cases where it's needed, e.g. like the Jacobian test for transforms

PR #2437, opened 2020-04-23
Start a contrib.epidemiology module

fritzo:
Addresses #2426 

This refactors examples/sir_hmc.py into a more reusable framework that separates (1) clear user-facing model specification from (2) intricate effect handler code for inference and prediction. I have preserved the existing examples/sir_hmc.py as a minipyro-like concrete explanation of the new module; it thus serves as architecture documentation. This also makes small changes to sir_hmc.py to keep the two versions aligned (the old concrete and new abstract versions).

Whereas `sir_hmc.py` contains four models, I was able to reduce user-facing modeling code down to a few methods and only a single duplication: we need a forward `.transition_fwd()` method and also a `.transition_bwd()` method; these contain reverse versions of dynamic equations and cannot easily be automatically generated.

After this PR I plan to simplify sir_hmc.py by moving all the fancy DCT and spline stuff into the new module.

## Tested
- [x] Added a simple unit test
- [x] Ran plotting code locally to verify results are the same as `sir_hmc.py`


docs/source/contrib.epidemiology.rst
```diff
@@ -0,0 +1,26 @@
...
+.. warning:: Code in ``pyro.contrib.forecast`` is under development.
+    This code makes no guarantee about maintaining backwards compatibility.
+
+``pyro.contrib.epidemiology`` is a framework for experimenting with a restricted
+class of stochastic discrete-time discrete-count compartmental models. This
+framework implements **inference** via Hamiltonian Monte Carlo, **prediction**
+of latent variables, and **forecasting** future trajectories.
+
+For explanation of the underlying machinery in this framework, see the
+`SIR with HMC tutorial <http://pyro.ai/examples/sir_hcm.html>`_
```
martinjankowiak, line 14: typo

pyro/contrib/epidemiology/compartmental.py
```diff
@@ -0,0 +1,427 @@
...
+    def _generative_model(self, forecast=0):
+        """
+        Forward generative model used for simulation and forecasting.
+        """
+        # Sample global parameters.
+        params = self.global_model()
+
+        # Sample initial values.
+        state = self.initialize(params)
+        state = {i: torch.tensor(float(value)) for i, value in state.items()}
```
martinjankowiak, line 339: what is this for exactly?

pyro/contrib/epidemiology/compartmental.py
```diff
@@ -0,0 +1,427 @@
...
+            if site["type"] == "sample":
+                logp[name] = site["fn"].log_prob(site["value"])
+
+        # Manually perform variable elimination.
+        logp = sum(logp.values())
+        logp = logp.reshape(T, Q ** C, Q ** C)
+        logp = pyro.distributions.hmm._sequential_logmatmulexp(logp)
+        logp = logp.reshape(-1).logsumexp(0)
+        warn_if_nan(logp)
+        pyro.factor("transition", logp)
```
martinjankowiak, line 427: what happened to the previous `-log(4)`?

fritzo: I have neglected that from this model because (1) it does not affect inference (it does not change gradients), and (2) we will probably be moving to stochastic initial state soon anyway.

pyro/contrib/epidemiology/compartmental.py
fritzo: `self.initialize` currently returns deterministic initial states. Because those states are used in `torch.nn.functional.pad` they must be Python scalars. However in this function those initial states need to start out as tensors.

The `state` dict in this function acts similarly to the `state` dict in SMCFilter: it is a framework-managed storage location where users can read and write values. Note the `self.transition_fwd(params, state, t)` call below updates this state dict in-place.

PR #2442, opened 2020-04-25
Move spline function from `transforms.Spline` to utils

stefanwebb:
I've refactored the code so that the spline function is now in a utils module


stefanwebb:
It would be good to get some feedback on the public interface of the spline method. Is it okay that it expects it's inputs to be in a certain range or sum to 1 without validating this? Should the `Sigmoid` and `Softmax` transforms to make the inputs valid happen in the `Spline` transform, or the spline method in the ops module?

stefanwebb:
Also, it is sufficient that this method doesn't have any direct tests, but only tests by the class `Spline` that uses it? It will have the forward/inverse operations validated, the shape tested, and the log-det-abs-Jacobian tested this way

docs/source/ops.rst
```diff
@@ -93,3 +93,13 @@ State Space Model and GP Utilities
     :undoc-members:
     :show-inheritance:
     :member-order: bysource
+
+
+Numerical Methods Utilities
```
martinjankowiak, line 98: a bit vague?

stefanwebb: Do you think "Numerical Methods for Interpolation" might be a better title? If someone implements ODE solvers in the future for stochasic ODEs, should that be under a separate heading like "Numerical Methods for Differential Equations"?

stefanwebb:
> haven't had a chance to think about the api but if you want this to be used more generally (which i guess would motivate it being put in `ops`) you'll probably need additional tests

@martinjankowiak if I make tests to ensure the inverse holds as well as the log(det(Jacobian)) is correct would that suffice?

fritzo:
What is the status of this? @stefanwebb are you blocked on martin and me? It looks like you'll need to resolve a merge conflict.

stefanwebb:
@fritzo I'm not blocked, although it would be good to clarify what are the minimal tests you think would be required for the spline function now it's publicly exposed? (Btw, I'm away from the office this week)

docs/source/ops.rst
fritzo: Well you could first add smoke test exercising the new functions and methods with scalar and batched inputs. If you want to add further tests, you could add say a spline approximation of a non-polynomial function like `sin(x)` or `sin(x) exp(-x**2/200)` and `assert_close(spline_approx, truth)`.

fritzo:
> what are the minimal tests you think would be required

Well you could first add smoke test exercising the new functions and methods with scalar and batched inputs. If you want to add further tests, you could add say a spline approximation of a non-polynomial function like `sin(x)` or `sin(x) exp(-x**2/200)` and `assert_close(spline_approx, truth)`.

stefanwebb:
> > what are the minimal tests you think would be required
> 
> Well you could first add smoke test exercising the new functions and methods with scalar and batched inputs. If you want to add further tests, you could add say a spline approximation of a non-polynomial function like `sin(x)` or `sin(x) exp(-x**2/200)` and `assert_close(spline_approx, truth)`.

This will be quite a bit of work for something that is lower priority than some other improvements. I will come back to this later on...

fritzo:
> [testing] will be quite a bit of work

No worries. I'm ok merging as long as the spline functionality is exercised thoroughly in `SplineTransform` tests.

fritzo:
@stefanwebb do you want to get this into 1.4 release? I'm happy to merge as soon as you resolve merge conflicts.

fritzo:
closing since work seems to have moved to https://github.com/stefanwebb/flowtorch
feel free to reopen

stefanwebb:
Yes, let's close this one :)

PR #2443, opened 2020-04-25
Added with_cache() and refactored basic element-wise transforms

stefanwebb:
Added `.with_cache()` to classes inheriting from `Transform` in Pyro to match https://github.com/pytorch/pytorch/pull/36882

Also, did some refactoring that is orthogonal to this while I was digging around in transform code


pyro/distributions/transforms/basic.py
```diff
@@ -0,0 +1,107 @@
...
+def leaky_relu():
+    """
+    A helper function to create a
+    :class:`~pyro.distributions.transforms.LeakyReLUTransform` object for
+    consistency with other helpers.
+    """
+    return LeakyReLUTransform()
+
+
+class TanhTransform(Transform):
```
fritzo, line 74: It looks like there is already a Tanh transform upstream.
https://pytorch.org/docs/stable/distributions.html#torch.distributions.transforms.TanhTransform

If that is missing from PyTorch 1.4, we could simply bumpy Pyro's version dependency to 1.5. Also we should really implement all of these upstream and add a note `# TODO move this upstream` to the short-term implementations here.

pyro/distributions/transforms/neural_autoregressive.py
```diff
@@ -1,122 +1,20 @@
-# Copyright (c) 2017-2019 Uber Technologies, Inc.
+# Copyright Contributors to the Pyro project.
```
fritzo, line 1: Can you please preserve the old copyright below the new, as per our Uber council?

pyro/distributions/transforms/neural_autoregressive.py
```diff
@@ -215,7 +113,7 @@ def _call(self, x):
         return y
 
     # This method returns log(abs(det(dy/dx)), which is equal to -log(abs(det(dx/dy))
-    def log_abs_det_jacobian(self, x, y):
+    def log_abs_det_jacobian(self, x, y, eps=1e-8):
```
fritzo, line 116: Can we make this an instance property rather than a method property? That way `.inv.log_abs_det_jacobian()` is can also customize this value.

pyro/distributions/transforms/basic.py
fritzo: OK I see you are merely moving these methods. But would you mind adding some `# TODO move upstream` comments?

stefanwebb: I think this is covered by #2467, but will do!

stefanwebb:
Made all requested changes

pyro/distributions/transforms/discrete_cosine.py
```diff
@@ -73,3 +73,8 @@ def _inverse(self, y):
 
     def log_abs_det_jacobian(self, x, y):
         return x.new_zeros((1,) * self.event_dim)
+
+    def with_cache(self, cache_size=1):
+        if self._cache_size == cache_size:
+            return self
+        return DiscreteCosineTransform(-self.event_dim, self.smooth, cache_size)
```
fritzo, line 80: Please specify `cache_size=cache_size` as a kwarg. In particular, I plan on adding new kwargs to `DiscreteCosineTransform` and will start using the `*,` separator to force kwarg specification.

pyro/distributions/transforms/lower_cholesky_affine.py
```diff
@@ -64,3 +64,8 @@ def log_abs_det_jacobian(self, x, y):
         """
         return torch.ones(x.size()[:-1], dtype=x.dtype, layout=x.layout, device=x.device) * \
             self.scale_tril.diag().log().sum()
+
+    def with_cache(self, cache_size=1):
+        if self._cache_size == cache_size:
+            return self
+        return LowerCholeskyAffine(self.loc, self.scale_tril, cache_size)
```
fritzo, line 71: Please specify `cache_size=cache_size` as a kwarg

pyro/distributions/transforms/permute.py
```diff
@@ -91,6 +91,11 @@ def log_abs_det_jacobian(self, x, y):
 
         return torch.zeros(x.size()[:-1], dtype=x.dtype, layout=x.layout, device=x.device)
 
+    def with_cache(self, cache_size=1):
+        if self._cache_size == cache_size:
+            return self
+        return Permute(self.permutation, cache_size)
```
fritzo, line 97: `cache_size=cache_size`

pyro/distributions/transforms/discrete_cosine.py
stefanwebb: I've made this change

PR #2444, opened 2020-04-26
Conditional spline

stefanwebb:
Conditional version of element-wise Spline


stefanwebb:
I've realized there is a bug in my Spline reorganization... Please hold off on reviewing until I've pushed a fix

pyro/distributions/transforms/spline.py
```diff
@@ -365,3 +471,71 @@ def spline(input_dim, **kwargs):
...
+    nn = DenseNN(context_dim,
+                 hidden_dims,
+                 param_dims=[input_dim * count_bins,
+                             input_dim * count_bins,
+                             input_dim * (count_bins - 1),
+                             input_dim * count_bins])
+    return ConditionalSpline(nn, input_dim, count_bins, bound=bound)
+
+
+if __name__ == "__main__":
```
martinjankowiak, line 506: remove?

pyro/distributions/transforms/spline.py
```diff
@@ -33,15 +37,28 @@ def _select_bins(x, idx):
...
-    extra_dims = len(idx.shape) - len(x.shape)
-    expanded_x = x.expand(idx.shape[:extra_dims] + (-1,) * len(x.shape))
-    return expanded_x.gather(-1, idx).squeeze(-1)
+
+    """
+    idx ~ (context_batch_dims, input_dim, 1)
+    x ~ (batch_dims, input_dim, count_bins)
+    """
+
+    # Broadcast dimensions of idx over x
```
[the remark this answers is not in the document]
stefanwebb: By the way, the reason I do this is because `gather` doesn't broadcast... Yes there must be a better way to ensure the input batch dims and the context variable batch dims line up

pyro/distributions/transforms/spline.py
```diff
@@ -8,18 +8,22 @@
...
+from pyro.distributions.conditional import ConditionalTransformModule
 from pyro.distributions.torch_transform import TransformModule
 from pyro.distributions.util import copy_docs_from
+from pyro.nn import DenseNN
 
 
 def _search_sorted(bin_locations, inputs, eps=1e-6):
     """
-    Searches for which bin an input belongs to (in a way that is parallelizable and amenable to autodiff)
+    Searches for which bin an input belongs to (in a way that is parallelizable and
```
fritzo, line 25: FYI it looks like `torch.searchsorted()` is coming soon https://github.com/pytorch/pytorch/pull/34577 https://github.com/pytorch/pytorch/issues/7284 I guess we should try to follow that interface to make it easy to switch once it is released.

pyro/distributions/transforms/spline.py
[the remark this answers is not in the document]
fritzo: I don't know a better way offhand.

pyro/distributions/transforms/spline.py
```diff
@@ -33,15 +37,28 @@ def _select_bins(x, idx):
...
+    """
+
+    # Broadcast dimensions of idx over x
+    if len(idx.shape) >= len(x.shape):
+        x = x.reshape((1,) * (len(idx.shape) - len(x.shape)) + x.shape)
+        x = x.expand(idx.shape[:-2] + (-1,) * 2)
+    # Broadcast dimensions of x over idx
+    elif len(idx.shape) <= len(x.shape):
+        idx = idx.reshape((1,) * (len(x.shape) - len(idx.shape)) + idx.shape)
+        idx = idx.expand(x.shape[:-2] + (-1,) * 2)
```
fritzo, line 53: These conditions overlap and look dubious. Why do you use `<=` and `>=` rather than strict inequalities? Why is `.expand()` logic gated by `.dim()` conditions rather than `.shape` conditions?

pyro/distributions/transforms/spline.py
stefanwebb: I've changed the signature of `_searchsorted` so it matches `torch.searchsorted()`. The latter has many more features, however, you should be able to replace `_searchsorted` with `torch.searchsorted` as is without breaking anything

pyro/distributions/transforms/spline.py
stefanwebb: I have to look into this again... What I'm trying to take into account is that the context variable may have batch dimensions, which then effect the batch dimensions of the variable `idx`. We have to expand `idx` so that its batch dimensions match up with that of `x` before performing `gather`.

What I mean by batch dimensions on `context` is that we can have a single context variable that is produces parameters that are shared by all elements of the input variable batch, or we can have a batch of context variables so that each input variable in the batch has its own parameters

stefanwebb: See the latest version for a resolution of this issue

stefanwebb:
I think I've fixed this PR now... I had to figure out that the batch dimensions of the context variable (z for p(x | z)) must broadcast over the batch dimensions of the input variable (x), and not the other way around or both. This is a convention that is followed by the other conditional transforms

fritzo:
@stefanwebb I don't know what's wrong with travis. You might try merging in dev and pushing to trigger another build.

pyro/distributions/transforms/spline.py
```diff
@@ -188,60 +207,57 @@ def _monotonic_rational_spline(inputs, widths, heights, derivatives, lambdas,
...
     """
+    domain = constraints.real
+    codomain = constraints.real
+    bijective = True
+    event_dim = 0
 
-    def __init__(self, input_dim, count_bins=8, bound=3., order='linear'):
-        super().__init__()
+    def __init__(self, widths=None, heights=None, derivatives=None, lambdas=None, bound=3.0, order='linear'):
+        super().__init__(cache_size=1)
```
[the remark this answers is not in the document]
stefanwebb: Can you elaborate on this commit? `ConditionedSpline` is the only class that inherits from `Transform` directly, whereas `Spline` and `ConditionalSpline` will have `cache_size=1` implicitly. Do you mean you think there should be an option to set `cache_size` in these two classes?

stefanwebb:
@fritzo yes, the maths hasn't changed at all and there isn't anything in particular that needs reviewing I think. The main change I made since last time is in `_select_bins`: https://github.com/pyro-ppl/pyro/pull/2444/files#diff-225bd72ad1983bd159be030ab40a7a07R42

pyro/distributions/transforms/spline.py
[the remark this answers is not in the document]
fritzo: Thanks for explaining, I was just curious about the default value and whether it differed.

PR #2445, opened 2020-04-26
Simplify compartmental model interface

fritzo:
Addresses #2426 

This is a refactoring to prepare for a future SEIR model.

1. Moves the `num_quant_bins` from `.__init__()` (which henceforth accept only model-related args) to `.fit()` (which accepts inference-related args). This reduces the burden of creating a new model class.
2. Renames `SIRModel` to `SimpleSIRModel` and adds documentation that "this model is intended to be forked, not customized"
3. Fixes definition of timescale `tau` so that it is actually the mean (there was an off-by-1 error).

## Tested
- [x] added new test configs to test/contrib/epidemiology

### DIFF
--- a/pyro/contrib/epidemiology/sir.py
+++ b/pyro/contrib/epidemiology/sir.py
@@ -26,13 +26,9 @@ class SimpleSIRModel(CompartmentalModel):
         ``I``). Must be greater than 1.
     :param iterable data: Time series of new observed infections.
     :param int data: Time series of new observed infections.
-    :param int num_quant_bins: The number of quantization bins to use.
-        Note that computational cost is exponential in `num_quant_bins`.
-        Defaults to 4.
     """
 
-    def __init__(self, population, recovery_time, data, *,
-                 num_quant_bins=4):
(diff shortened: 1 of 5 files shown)


pyro/contrib/epidemiology/sir.py
```diff
@@ -10,26 +10,35 @@
...
     :param int data: Time series of new observed infections.
-    :param int num_quant_bins: The number of quantization bins to use. Note that
-            computational cost is exponential in `num_quant_bins`. Defaults to 4.
+    :param int num_quant_bins: The number of quantization bins to use.
+        Note that computational cost is exponential in `num_quant_bins`.
+        Defaults to 4.
     """
 
     def __init__(self, population, recovery_time, data, *,
                  num_quant_bins=4):
```
martinjankowiak, line 35: i don't see where `num_quant_bins` is piped here?

fritzo: oops, thanks for catching, I've removed.

PR #2446, opened 2020-04-26
Implement a simple SEIR model

fritzo:
Addresses #2426 

This adds a simple SEIR model with `Binomial` transitions and likelihoods. I plan to follow this up with an overdispersed SEIR model with `BetaBinomial` infections, but that will require deeper modifications from the `SimpleSIR`s model.

@martinjankowiak you should be able to test mixing by adding the `-e` command line arg e.g.
```sh
python examples/contrib/epidemiology/sir.py -e 2 -p 10000 -d 60
```

## Tested
- [x] added unit tests
- [x] added cases to test_examples.py
- [x] ran locally on a largish example


examples/contrib/epidemiology/sir.py
```diff
@@ -11,26 +11,35 @@
...
     extended_data = [None] * (args.duration + args.forecast)
-    model = SimpleSIRModel(args.population, args.recovery_time, extended_data)
+    model = Model(args.population, args.incubation_time, args.recovery_time,
+                  extended_data)
     for attempt in range(100):
         samples = model.generate({"R0": args.basic_reproduction_number,
                                   "rho": args.response_rate})
         obs = samples["obs"][:args.duration]
-        S2I = samples["S2I"]
+        new_I = samples.get("S2I", samples.get("E2I"))
```
martinjankowiak, line 35: what's going on here?

pyro/contrib/epidemiology/seir.py
```diff
@@ -0,0 +1,139 @@
...
+        self.incubation_time = incubation_time
+
+        assert isinstance(recovery_time, float)
+        assert recovery_time > 1
+        self.recovery_time = recovery_time
+
+        self.data = data
+
+    series = ("S2E", "E2I", "I2R", "obs")
+    full_mass = [("R0", "rho")]
```
martinjankowiak, line 51: should this be passed through `infer` like `num_quant_bins`?

examples/contrib/epidemiology/sir.py
fritzo: In the SIR models, S2I is noisily observed. In SEIR models, E2I is noisily observed. This one-liner works for either model class.

pyro/contrib/epidemiology/seir.py
fritzo: Indeed it is already passed like `num_quant_bins`.
https://github.com/pyro-ppl/pyro/blob/a56ecbaf687a6220aa02dc361dbd1d117843d6f6/pyro/contrib/epidemiology/compartmental.py#L247
This class attribute merely sets the default value.

fritzo:
Thanks for reviewing @martinjankowiak !

Issue #2441, opened 2020-04-25
[FR] Simple arithmetic transformation operator overloading


ecotner:
Hi, I've been trying to teach myself `pyro` for the last couple weeks, and it is a great package; I'm learning quite a bit about Bayesian methods from going through the tutorials and doing side projects. In the course of my studies, I found that there is one thing that would make creating models with deterministic transformations even more transparent and simple: overloading simple arithmetic operators on `Distribution`s so that you can quickly shift/scale/invert them. I'll give you an example:

Say I have a gamma-distributed RV that I want to shift in the positive direction; in `pyro` I would have to make that distribution like so:
```python
from pyro.distributions import Gamma, TransformedDistribution
from pyro.distributions.transforms AffineTransform

X = Gamma(1, 1)
X = TransformedDistribution(X, AffineTransform(1, 1))
```
it actually took me quite a long time to figure out that this was even possible, and it would be much simpler if I could do a short one-liner like `X = 1 + Gamma(1, 1)`, but this throws an error because addition between distributions and integers isn't defined. This extends to other simple binary/unary operations such as subtraction, multiplication, division, negation, etc. I believe `pymc3` already has this kind of feature built in, so it would be great if we could do this with pyro distributions too.

I was playing around with some mixin class that would make this sort of thing more transparent (I'm not sure how you'd implement this in reality; maybe directly add the dunder methods to `Distribution` so other subclasses can inherit from it?):
```python
class ArithmeticMixin:
    def __add__(self, x: float):
        return _TransformedDistribution(self, AffineTransform(x, 1))
    
    def __radd__(self, x: float):
        return _TransformedDistribution(self, AffineTransform(x, 1))
    
    def __sub__(self, x: float):
        return _TransformedDistribution(self, AffineTransform(-x, 1))
    
    def __rsub__(self, x: float):
        return _TransformedDistribution(self, AffineTransform(x, -1))
    
    def __mul__(self, x: float):
        return _TransformedDistribution(self, AffineTransform(0, x))
    
    def __rmul__(self, x: float):
        return _TransformedDistribution(self, AffineTransform(0, x))
    
    def __truediv__(self, x: float):
        return _TransformedDistribution(self, AffineTransform(0, 1/x))
    
    def __neg__(self):
        return _TransformedDistribution(self, AffineTransform(0, -1))
    
    def __abs__(self):
        return _TransformedDistribution(self, AbsTransform())

[...]

PR #2448, opened 2020-04-27
adding RandomVariable container class around Distribution

ecotner:
* simple container class that encapsulates a `Distribution`
* arithmetic operators like +, -, /, **, etc. overloaded to perform transformations on the domain of the distribution
* easy method `self.transform` to apply other transformations and return another `RandomVariable`

## Documented
- [x] docstring for `.rv` that points to `RandomVariable` class
- [x] add a file docs/source/contrib.randomvariable.rst following one of the other contrib.*.rst files
- [x] add a contrib.randomvariable to  docs/source/index.rst
- [x] distring for `.dist`
- [x] example usage with method chaining in `RandomVariable`

## Tested
- [x] distribution support tests


eb8680:
@ecotner thanks for the PR.  I stand by what I said in #2441, which is that this design breaks down when users (quite reasonably) try to build expressions that are not invertible or contain more than one random variable, especially if the random variables are not jointly independent.  We could raise `NotImplementedError`s or `ValueError`s in these cases, but it's generally best to avoid designing expressive APIs that frequently violate user intuition in non-obvious ways.  

Even disregarding the multiple-variables issue, I can also foresee a lot of feature creep and maintenance burden, up to reimplementing large swaths of the PyTorch tensor API as `torch.distributions.Transform`s.  I seem to recall that this problem has been discussed in PyTorch issues about adding more complicated shape manipulations to `torch.distributions.transforms`, but I'm having trouble finding the issues at the moment.

I think the best way to proceed with a feature that most closely resembles this PR is to resurrect the `RandomVariable` implementations in Funsor (https://github.com/pyro-ppl/funsor/pull/130).  There's been a lot of work done since then that would make wrapping distributions and ops/transforms generically (https://github.com/pyro-ppl/funsor/issues/159) fairly straightforward, and I am happy to guide you through the process if you're interested.  

However, I still think a better long-term solution to the multiple-variables problem is having `pyro.sample` statements emit `funsor.Variable`s.

fritzo:
@ecotner This is indeed a slick little DSL for specifying transforms, and I'd like to be able to expose it in Pyro. I think the safest way to do so in light of @eb8680 valid concerns is to:

1. Implement your DSL in a contrib package, say `pyro.contrib.transform` or `pyro.contrib.randomvariable`.
2. Clearly separate the `Distribution` DSL from the `RandomVariable` DSL, so as to avoid API creep and potential conflict with inference-oriented operations on distributions.

Regarding (2), we could see the `Distribution` DSL as providing chainable methods `.to_event()`, `.mask()`, and `.expand()`; whereas the `RandomVariable` DSL provides chainable methods `.transform()` `.__add__()` (for which we could add a torch-like alias `.add()`), `.__mul__()` etc. I think it would be pretty safe to add a single distribution propert `.rv` that switched from the `Distribution` DSL to the `RandomVariable` DSL, and provide a random variable property `.dist` to convert back to the `Distribution` DSL. Example usage could then be like
```py
d = dist.Uniform(0, 1).rv.add(1).mul(0.5).tanh().dist.mask(my_mask).to_event(1)
#   \________________/   \_____________________/     \________________________/
#    Distribution DSL       RandomVariable DSL            Distribution DSL
```
where the `.rv` switches to the `RandomVariable` DSL and the `.dist` switches back.

I realize this is a significant change from your original mixin idea, but at least it keeps your nice method chaining syntax. Would this limited design something you'd still like to work on?

eb8680:
@ecotner - @fritzo's proposal seems like a good first step, we often start new `pyro.contrib` modules when experimenting with new API designs.  I agree with you that an interface like this is a worthy design goal, and I certainly don't want to discourage you from contributing!

ecotner:
Thanks guys, those are definitely valid concerns. I wasn't even aware of `funsor` or your plans to integrate it into `pyro`, so I'm not surprised my idea could potentially break some of the future design elements you had in mind.

I like the idea of easily switching between `Distribution` <--> `RandomVariable`; and the syntax you outlined seems pretty slick itself! (getting strong `pandas` vibes). Since `RV.__add__` could basically call `RV.add` under the hood, you would be free to use either method chaining or operator overloading as you saw fit. And I had no assumptions that the mixin idea would even be feasible; my original post was simply to illustrate the concept and give an example that demonstrated the desired functionality without making modifications to the `pyro` source code.

@eb8680 I'm not discouraged at all; I don't think I've ever had a PR accepted on the first try :wink:.

ecotner:
So, I moved everything over to `pyro.contrib` and added the method chaining, but I'm getting a problem which I believe to be a circular import error that is potentially preventing me from adding the `.rv` property to `Distribution`. Basically, `RandomVariable` needs `TransformedDistribution`, which needs `Distribution`, which needs `RandomVariable`... etc.

I found a way around it by moving the import inside the `.rv` function, but it seems a little hacky. Thoughts?

Also, do you guys follow the `isort` format anymore? I ran `make format` and it modified the import statements of like 130 different files. I'm pretty sure that's why my build is failing, but I didn't want to change so many files at once...

pyro/distributions/distribution.py
```diff
@@ -166,3 +166,8 @@ def has_rsample_(self, value):
             raise ValueError("Expected value in {False,True}, actual {}".format(value))
         self.has_rsample = value
         return self
+
+    @property
+    def rv(self):
```
fritzo, line 171: Could you add a big docstring here with an EXPERIMENTAL warning and example usage, something like
```
"""
EXPERIMENTAL Switch to the Random Variable DSL for transforming random variables.

Example usage::

    # This should be equivalent to an Exponential distribution.
    Uniform(0,1).rv.log().neg().dist

:returns: A :class:`~pyro.contrib.randomvariable.random_variable.RandomVariable`
    object wrapping this distribution.
:rtype: ~pyro.contrib.randomvariable.random_variable.RandomVariable
"""
```

pyro/contrib/randomvariable/random_variable.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def transform(self, t: Transform):
+        """Performs a transformation on the distribution underlying the RV.
+
+        :param t: The transformation (or sequence of transformations) to be
+            applied to the distribution. There are many examples to be found in
+            `torch.distributions.transforms` and `pyro.distributions.transforms`,
+            or you can subclass directly from `Transform`.
+        :type t: `Transform`
+        """
```
fritzo, line 120: You might add `:rtype: RandomVariable`

fritzo:
@eb8680 are you ok with this experimental `.rv` interface? If so I'm ready to merge.

PR #2450, opened 2020-04-28
Add a factory for Poisson / NegativeBinomial / Binomial / BetaBinomial

fritzo:
Addresses #2426 

This creates a new helper `infection_dist()` that creates one of a Poisson, NegativeBinomial, Binomial, or BetaBinomial depending on the quantitative values of its arguments. Of these four distributions, I intend only to use the Binomial and BetaBinomial; the others are present more as a bridge to limiting parameterizations that are more common in epidemiology.

image

This PR also refactors the SIR and SEIR models to use the new helper. This refactoring prepares for overdispersed versions of these distributions in a follow-up PR #2451.

## Tested
- [x] added a new unit test confirming approximate agreement in four limiting cases
- [x] visually validated agreement in [a notebook](https://github.com/fritzo/notebooks/blob/master/infection_dist.ipynb)
- [x] ran `sir.py -p 10000 -d 60 -f 30 --plot` as a sanity check

### DIFF
--- a/pyro/contrib/epidemiology/distributions.py
+++ b/pyro/contrib/epidemiology/distributions.py
@@ -28,7 +28,7 @@ def infection_dist(*,
 
     The overdispersed distributions (Negative-Binomial and Beta-Binomial
     returned when ``concentration < ∞``) are useful for modeling superspreader
-    individuals [2,3]. The finitely supported distributions Binomial and
+    individuals [1,2]. The finitely supported distributions Binomial and
     Negative-Binomial are useful in small populations and in probabilistic
     programming systems where truncation or censoring are expensive [3].
 
@@ -58,7 +58,7 @@ def infection_dist(*,
     :param population: The total number of individuals in a population.
         This defaults to an infinite population.
     :concentration: The concentration or dispersion parameter ``k`` in
-        overdispersed models of superspreaders [2,3]. This defaults to minimum
+        overdispersed models of superspreaders [1,2]. This defaults to minimum
         variance ``concentration = ∞``.
     """
     # Convert to colloquial variable names.
--- a/pyro/contrib/epidemiology/distributions.py
+++ b/pyro/contrib/epidemiology/distributions.py
@@ -1,0 +1,101 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import math
+
+import torch
+
+import pyro.distributions as dist
+
+
+def infection_dist(*,
+                   individual_rate,
+                   num_infectious,
+                   num_susceptible=math.inf,
+                   population=math.inf,
+                   concentration=math.inf):
+    """
+    Create a :class:`~pyro.distributions.Distribution` over the number of new
+    infections at a discrete time step.
+
+    This returns a Poisson, Negative-Binomial, Binomial, or Beta-Binomial
+    distribution depending on whether :param:`population` and
+    :param:`concentration` are finite. In Pyro models, the population is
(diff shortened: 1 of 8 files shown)


pyro/contrib/epidemiology/distributions.py
```diff
@@ -0,0 +1,98 @@
...
+            # Return an overdispersed Negative-Binomial distribution.
+            combined_k = k * I
+            logits = torch.as_tensor(R / k).log()
+            return dist.NegativeBinomial(combined_k, logits=logits)
+    else:
+        # Compute the probability that any given (susceptible, infectious)
+        # pair of individuals results in an infection at this time step.
+        p = torch.as_tensor(R / N).clamp(max=1 - 1e-6)
+        # Combine infections from all individuals.
+        combined_p = p.neg().log1p().mul(I).expm1().neg()  # = 1 - (1 - p)**I
```
fehiepsi, line 85: I think this is not consistent w.r.t. before. Previously, `combined_p` is `1 - (e^-rate_s) ** I = 1 - (e^-p) ** I`.

pyro/contrib/epidemiology/distributions.py
```diff
@@ -0,0 +1,98 @@
...
+    distribution depending on whether ``population`` and ``concentration`` are
+    finite. In Pyro models, the population is usually finite. In the limit
+    ``population → ∞`` and ``num_susceptible/population → 1``, the Binomial
+    converges to Poisson and the Beta-Binomial converges to Negative-Binomial.
+    In the limit ``concentration → ∞``, the Negative-Binomial converges to
+    Poisson and the Beta-Binomial converges to Binomial.
+
+    The overdispersed distributions (Negative-Binomial and Beta-Binomial
+    returned when ``concentration < ∞``) are useful for modeling superspreader
+    individuals [2,3]. The finitely supported distributions Binomial and
```
martinjankowiak, line 31: are these the right refs?

martinjankowiak: => [1, 2]?

pyro/contrib/epidemiology/distributions.py
```diff
@@ -0,0 +1,98 @@
...
+        ``R0 / tau`` where ``R0`` is the basic reproductive number and ``tau``
+        is the mean duration of infectiousness.
+    :param num_infectious: The number of infectious individuals at this
+        time step, sometimes ``I``, sometimes ``E+I``.
+    :param num_susceptible: The number ``S`` of susceptible individuals at this
+        time step. This defaults to an infinite population.
+    :param population: The total number of individuals in a population.
+        This defaults to an infinite population.
+    :concentration: The concentration or dispersion parameter ``k`` in
+        overdispersed models of superspreaders [2,3]. This defaults to minimum
```
martinjankowiak, line 61: same here?

pyro/contrib/epidemiology/distributions.py
fritzo: That's right, I am changing the formula. This PR seems to make it more mathematically plausible.

fritzo: @fehiepsi I am not super confident in these formulas, but the plots do show they are at least in agreement. Let me know if you have any suggestions.

fehiepsi: This seems to be similar to [Reed-Frost model](https://en.wikipedia.org/wiki/Reed%E2%80%93Frost_model#Mathematics).

fritzo:
Thanks for reviewing!

PR #2451, opened 2020-04-28
Implement overdispersed SIR and SEIR models

fritzo:
Addresses #2426

This adds two new compartmental models with overdispersed infection distributions to model superspreading individuals.

## Tested
- [x] added unit tests
- [x] added to `examples/contrib/epidemiology/sir.py`
- [x] sanity checked locally


pyro/contrib/epidemiology/seir.py
```diff
@@ -136,3 +108,129 @@ def transition_bwd(self, params, prev, curr, t):
...
+        and Incidence Time Series"
+        https://academic.oup.com/mbe/article/34/11/2982/3952784
+
+    :param int population: Total ``population = S + E + I + R``.
+    :param float incubation_time: Mean incubation time (duration in state
+        ``E``). Must be greater than 1.
+    :param float recovery_time: Mean recovery time (duration in state
+        ``I``). Must be greater than 1.
+    :param iterable data: Time series of new observed infections, i.e. a
+        Binomial subset of the ``E -> I`` transitions at each time step.
```
martinjankowiak, line 152: nit: i guess this is already merged but isn't 'binomial subset' strange language? maybe something like 
"i.e. a noisily observed time series of the E->I transitions at each time step"

pyro/contrib/epidemiology/seir.py
```diff
@@ -136,3 +108,129 @@ def transition_bwd(self, params, prev, curr, t):
...
+    implicit: ``R = population - S - E - I``) with transitions
+    ``S -> E -> I -> R``.
+
+    This model accounts for superspreading (overdispersed individual
+    reproductive number) by assuming each infected individual infects
+    BetaBinomial-many susceptible individuals, where the BetaBinomial
+    distribution acts as an overdispersed Binomial distribution, adapting the
+    more standard NegativeBinomial distribution that acts as an overdispersed
+    Poisson distribution [1,2] to the setting of finite populations. To
+    preserve Markov structure, we follow [2] and assume all infections by a
```
martinjankowiak, line 131: > To  preserve Markov structure, we follow [2] and assume all infections by a single individual occur on the single time step where that individual makes an ``I -> R`` transition.

can you please explain in more detail?

pyro/contrib/epidemiology/sir.py
```diff
@@ -115,3 +94,115 @@ def transition_bwd(self, params, prev, curr, t):
...
+
+    def transition_fwd(self, params, state, t):
+        R0, k, prob_i, rho = params
+
+        # Sample flows between compartments.
+        I2R = pyro.sample("I2R_{}".format(t),
+                          dist.Binomial(state["I"], prob_i))
+        c1 = (k * I2R).clamp(min=1e-6)
+        c0 = k / R0 * self.population
+        S2I = pyro.sample("S2I_{}".format(t),
```
martinjankowiak, line 176: why not use the dist helper?

fritzo: oops, I must have missed this. Will fix!

pyro/contrib/epidemiology/seir.py
fritzo: How about this?
```
:param iterable data: Time series of new observed infections.
    Each time step is Binomial distributed between 0 and the number of ``E -> I`` transitions.
    This allows false negative but no false positives.
```

martinjankowiak: lgtm. note this also occurs below

fritzo: done

pyro/contrib/epidemiology/seir.py
fritzo: done

fritzo:
Thanks for reviewing!

PR #2452, opened 2020-04-28
Use SMC to initialize compartmental models

fritzo:
Addresses #2426 

This replaces custom `.heuristic()` methods with a default generic `SMCFilter`-based initialization strategy, greatly simplifying the process of creating new compartmental models. Users can still override this default `.heuristic()` if needed.

:tada: hooray for composable inference :tada:

## Tested
- [x] ran `python sir.py -p 10000 -d 60 -f 30 --verbose --plot` and confirmed good posteriors and mixing:
```
         mean       std    median      5.0%     95.0%     n_eff     r_hat
 R0      1.48      0.08      1.47      1.36      1.61     16.89      1.09
rho      0.48      0.03      0.49      0.44      0.53     29.28      1.14
```
image

### DIFF
--- a/pyro/contrib/epidemiology/compartmental.py
+++ b/pyro/contrib/epidemiology/compartmental.py
@@ -14,7 +14,7 @@ import pyro.distributions as dist
 import pyro.distributions.hmm
 import pyro.poutine as poutine
 from pyro.distributions.transforms import DiscreteCosineTransform
-from pyro.infer import MCMC, NUTS, infer_discrete
+from pyro.infer import MCMC, NUTS, SMCFilter, infer_discrete
 from pyro.infer.autoguide import init_to_value
 from pyro.infer.reparam import DiscreteCosineReparam
 from pyro.util import warn_if_nan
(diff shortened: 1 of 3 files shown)


pyro/contrib/epidemiology/compartmental.py
```diff
@@ -433,3 +456,45 @@ def enum_shape(position):
...
+            if site["type"] == "sample":
+                state[name] = site["value"]
+
+        self.t = 0
+        state.update(self.model.initialize(params))
+        self.step(state)  # Take one step since model.initialize is deterministic.
+
+    def step(self, state):
+        with poutine.block(), poutine.condition(data=state):
+            params = self.model.global_model()
```
martinjankowiak, line 482: why is this needed exactly?

fritzo: Recall that `SMCFilter` tracks all state in a `state` dict and periodically resamples all that state. In this use case the `state` dict contains values of all sample sites that we will eventually replay. In particular it contains sample sites for global variables. Now since resampling changes those sites, and since `global_model()` not only draws samples but performs computation on those samples, we need to re-perform that computation after each resampling. Hence we block (to avoid this being traced again), and condition on the `SMCFilter`-managed `state` to get the latest resampled `params` each step.

PR #2457, opened 2020-05-01
Add example of sparsely observed SIR model

fritzo:
Addresses #2426 

This adds an example model with sparsely observed cumulative infections. This model is interesting because it preserves Markov structure by adding an auxiliary variable for fully observed cumulative infections (as discussed with @eb8680).

@eb8680 two notes:
1. Regarding our models with Lucy, I think there is enough complexity that we should probably create a model in say `examples/contrib/epidemiology/epi_phy.py` that has a complete complex model and script. That is, I think it would be most educational to illustrate each model feature independently in the `pyro.contrib.epidemiology` module, and then combine them in `examples/`.
2. The `S2O` flow looks a little like a chemical reaction with multiple products.

## Tested
- [x] added unit tests

### DIFF
--- a/pyro/contrib/epidemiology/sir.py
+++ b/pyro/contrib/epidemiology/sir.py
@@ -224,7 +224,7 @@ class SparseSIRModel(CompartmentalModel):
     population - S - I``) with transitions ``S -> I -> R``.
 
     This model allows observations of **cumulative** infections at uneven time
-    intervals. To preserve Markov structure (and hence tractible inference)
+    intervals. To preserve Markov structure (and hence tractable inference)
     this model adds an auxiliary compartment ``O`` denoting the fully-observed
     cumulative number of observations at each time point. At observed times
     (when ``mask[t] == True``) ``O`` must exactly match the provided data;
--- a/pyro/contrib/epidemiology/sir.py
+++ b/pyro/contrib/epidemiology/sir.py
@@ -22,7 +22,6 @@ class SimpleSIRModel(CompartmentalModel):
     :param int population: Total ``population = S + I + R``.
     :param float recovery_time: Mean recovery time (duration in state
         ``I``). Must be greater than 1.
-    :param iterable data: Time series of new observed infections.
     :param iterable data: Time series of new observed infections. Each time
         step is Binomial distributed between 0 and the number of ``S -> I``
         transitions. This allows false negative but no false positives.
@@ -136,7 +135,6 @@ class OverdispersedSIRModel(CompartmentalModel):
     :param int population: Total ``population = S + I + R``.
     :param float recovery_time: Mean recovery time (duration in state
         ``I``). Must be greater than 1.
-    :param iterable data: Time series of new observed infections.
     :param iterable data: Time series of new observed infections. Each time
         step is Binomial distributed between 0 and the number of ``S -> I``
         transitions. This allows false negative but no false positives.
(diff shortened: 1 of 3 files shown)


pyro/contrib/epidemiology/sir.py
```diff
@@ -212,3 +210,111 @@ def transition_bwd(self, params, prev, curr, t):
...
+
+    To customize this model we recommend forking and editing this class.
+
+    This is a stochastic discrete-time discrete-state model with four
+    compartments: "S" for susceptible, "I" for infected, and "R" for
+    recovered individuals (the recovered individuals are implicit: ``R =
+    population - S - I``) with transitions ``S -> I -> R``.
+
+    This model allows observations of **cumulative** infections at uneven time
+    intervals. To preserve Markov structure (and hence tractible inference)
```
martinjankowiak, line 227: typo: tractable

pyro/contrib/epidemiology/sir.py
```diff
@@ -212,3 +210,111 @@ def transition_bwd(self, params, prev, curr, t):
...
+        # Update compartments with flows.
+        state["S"] = state["S"] - S2I
+        state["I"] = state["I"] + S2I - I2R
+        state["O"] = state["O"] + S2O
+
+        # Condition on cumulative observations.
+        mask_t = self.mask[t] if t < self.duration else False
+        data_t = self.data[t] if t < self.duration else None
+        pyro.sample("obs_{}".format(t),
+                    dist.Delta(state["O"]).mask(mask_t),
```
martinjankowiak, line 292: why is the auxiliary necessary? or does this just play better with the structure of `CompartmentalModel`?

fritzo: tl;dr The auxiliary variable is needed to preserve Markov structure.

The observations in this model are aggregated over intervals: `obs=S2I[t_prev+1:t_curr+1].sum()` where `t_prev` is the time of the last observation and `t_curr` is the time of the current observation. In our enumeration strategy, this would couple all `t_curr-t_prev`-many enumeration variables, growing exponentially in the number of variables. While the non-parallel-scan enumeration strategy could handle this, it would be prohibitively expensive, and would not allow e.g. large gaps in sensor data (as e.g. when a government shuts down or runs out of tests). The trick we're using is to add an auxiliary variable for the entire cumulative observation trajectory (with the same likelihood as in the usual SIR models), and then `Delta`-clamp that auxiliary to the true observations at a few sparse time steps. This makes more work for HMC adds one enumeration variable per time step and increases the complexity of variable elimination by a constant factor of `Q**2`, but crucially this factor is independent of gap size.

I had been struggling with this issue for a while since Lucy's model simulates 4 times per day but is observed only once. The only alternative I could see was to do parallel-scan variable elimination where each `DiscreteHMM` state covered the joint distribution over an entire day (four time steps), resulting in complexity `Q**(2 * 4 * 2)` for an SIR model or `Q**(3 * 4 * 2)` for an SEIR model.

martinjankowiak: thanks for the explanation. to clarify though: if all you had was occasional missing data you wouldn't need this construction. this is really for the cumulative case

fritzo: Correct. It appears the cumulative case is more common in epidemiology.

PR #2458, opened 2020-05-01
cast to float in dmm.py

martinjankowiak:
i noticed that dmm.py was reporting integer NLLs on evaluation. i guess because pytorch has changed how casting works. i wonder if this might be happening elsewhere?


martinjankowiak:
> the return type of svi.evaluate_loss()

(python) float

> the datatype of val_seq_length and test_seq_lengths?

torch.int64

PR #2462, opened 2020-05-05
Return `pyro.distributions.TransformedDistribution` instead of `torch.distributions.TransformedDistribution` when conditioning

pawni:
Hi all,

I started working with Pyro's conditional transforms. However, the `ConditionalTransformedDistribution` returns a pytorch `TransformedDistribution` instead of the pyro version. If you want to use this distribution with `pyro.sample(.)` it needs to base the pyro version though.

### DIFF
--- a/pyro/distributions/conditional.py
+++ b/pyro/distributions/conditional.py
@@ -3,9 +3,10 @@
 
 from abc import ABC, abstractmethod
 
-import pyro
 import torch
 import torch.nn
+
+from .torch import TransformedDistribution
 
 
 class ConditionalDistribution(ABC):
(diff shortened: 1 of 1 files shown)


pyro/distributions/conditional.py
```diff
@@ -3,6 +3,7 @@
 
 from abc import ABC, abstractmethod
 
+import pyro
```
fritzo, line 6: Could you avoid the top level import and instead
```py
from .torch import TransformedDistribution
```
and use `TransformedDistribution` below? The reason is that `pyro.distributions` tries to be self-contained and avoid ever importing `pyro`.

pawni: Sure thing! Just changed that.

PR #2465, opened 2020-05-07
Implement arrow-head precision adaptation scheme

fehiepsi:
As pointed out by @fritzo, with a Gaussian posterior, we can use grad samples to estimate mass matrix (instead of using samples to estimate inverse mass matrix). Indeed, if z has covariance M-1, then `grad(potential, z) = M z` will have covariance M.

### Tests

+ All tests pass
+ I printed out the inverse mass matrix of the beta_bernoulli model, and it matches the expected one.

### TODO

- [x] ~Subclass WarmupAdapter using the precision adaptation scheme. Probably putting the implementation in `contrib.epidemiology`.~ After #2473, it is enough to create a new ArrowheadMassMatrix class.
- [x] Implement [arrowhead](https://en.wikipedia.org/wiki/Arrowhead_matrix) mass matrix adaptation.
- [x] ~Implement `arrowhead_precison_to_scale_tril`~ Various utilities are implemented and tested: `arrowhead_sqrt`, `arrowhead_sqrt_to_sqrt_inverse`, `arrowhead_sqrt_inverse_to_inverse`,...
- [x] Optimize the implementations. I expect that the complexity is similar to diagonal one if the `head` dimension is small w.r.t. the full dimension. Indeed, it is O(N x head_size^2) in time and O(N x head_size) in memory. :100: 
- [x] Test on some models


pyro/infer/mcmc/adaptation.py
```diff
@@ -194,25 +178,394 @@ def step(self, t, z, accept_prob):
...
+
+    A, B = x.top[:, :head_size], x.top[:, head_size:]
+    top = A.t().matmul(x.top)
+    bottom_left = top[:, head_size:].t()
+    # the following matmul operator is O(N^2 x head_size)
+    bottom_right = B.t().matmul(B) + x.bottom_diag.pow(2).diag()
+    return torch.cat([top, torch.cat([bottom_left, bottom_right], -1)], 0)
+
+
+class ArrowheadMassMatrix:
```
fehiepsi, line 425: @fritzo This mass matrix adapter has complexity O(matrix_size x head_size^2) in time and O(matrix_size x head_size) in memory. So the performance should be similar to the diagonal mass matrix when head_size << matrix_size. It is nice to try on the SIR models. There are a few features we can add easily (which I can add in the follow-up PRs if needed) such as:
+ a pack/unpack mechanism to move important parts of a latent variable to the head part of the arrowhead matrix. I think this is the feature that you requested in #2425.
+ specify the initial mass matrix (which is learned using SVI) in the constructor

pyro/infer/mcmc/adaptation.py
```diff
@@ -194,25 +178,394 @@ def step(self, t, z, accept_prob):
...
+        """
+        Computes the gradient of kinetic energy w.r.t. the momentum `r`.
+        It is equivalent to compute velocity given the momentum `r`.
+        :param dict r: a dictionary maps site names to a tensor momentum.
+        :returns: a dictionary maps site names to the corresponding gradient
+        """
+        v = {}
+        for site_names, mass_matrix_sqrt_inverse in self._mass_matrix_sqrt_inverse.items():
+            r_flat = torch.cat([r[site_name].reshape(-1) for site_name in site_names])
+            # NB: using inverse_mass_matrix as in BlockMassMatrix will cost
```
fehiepsi, line 524: @fritzo I think you mentioned about the matrix-vector product here.

tests/infer/mcmc/test_nuts.py
```diff
@@ -412,3 +411,21 @@ def model(cov):
...
+    def model(data):
+        concentration = torch.tensor([1.0, 1.0, 1.0])
+        p_latent = pyro.sample('p_latent', dist.Dirichlet(concentration))
+        pyro.sample("obs", dist.Categorical(p_latent), obs=data)
+        return p_latent
+
+    true_probs = torch.tensor([0.1, 0.6, 0.3])
+    data = dist.Categorical(true_probs).sample(sample_shape=(torch.Size((2000,))))
+    nuts_kernel = NUTS(model, jit_compile=True, ignore_jit_warnings=True)
+    nuts_kernel.mass_matrix_adapter = ArrowheadMassMatrix(head_size=1)
```
fritzo, line 426: nit: could you use `head_size=2` in this test and `concentration=torch.ones(4)`?

pyro/infer/mcmc/adaptation.py
```diff
@@ -8,9 +8,11 @@
 
 import pyro
 from pyro.ops.dual_averaging import DualAveraging
-from pyro.ops.welford import WelfordCovariance
+from pyro.ops.welford import WelfordCovariance, WelfordArrowheadCovariance
 
 adapt_window = namedtuple("adapt_window", ["start", "end"])
+ArrowheadMatrix = namedtuple("ArrowheadMatrix", ["top", "bottom_diag"])
```
fritzo, line 14: Do you assume everywhere that an `ArrowheadMatrix` is symmetric? (i would expect so)

pyro/ops/welford.py
```diff
@@ -48,3 +48,47 @@ def get_covariance(self, regularize=True):
...
+        delta_post = sample - self._mean
+        if self.head_size > 0:
+            self._m2_top = self._m2_top + torch.ger(delta_post[:self.head_size], delta_pre)
+        else:
+            self._m2_top = sample.new_empty(0, sample.size(0))
+        self._m2_bottom_diag = self._m2_bottom_diag + delta_post[self.head_size:] * delta_pre[self.head_size:]
+
+    def get_covariance(self, regularize=True):
+        """
+        Gets the covariance in arrowhead form: (top, bottom_diag) where `top = cov[head_size:]`
```
fritzo, line 80: do you mean `top = cov[:head_size]`?

pyro/infer/mcmc/adaptation.py
```diff
@@ -8,9 +8,11 @@
...
+ArrowheadUpperSqrtMatrix = namedtuple("ArrowheadUpperSqrtMatrix", ["top", "bottom_diag"])
```
fritzo, line 15: Consider moving these namedtuples and the math helpers to a file `pyro.ops.arrowhead`. I think these could be marked `EXPERIMENTAL`. Then you could `import pyro.ops.arrowhead as arrowhead`.

pyro/infer/mcmc/adaptation.py
```diff
@@ -350,3 +352,247 @@ def unscale(self, r):
...
+            adapt_scheme.update(z_grad_flat)
+
+    def end_adaptation(self):
+        """
+        Updates the current mass matrix using the adaptation scheme.
+        """
+        mass_matrix = {}
+        for site_names, adapt_scheme in self._adapt_scheme.items():
+            top, bottom_diag = adapt_scheme.get_covariance(regularize=True)
+            mass_matrix[site_names] = ArrowheadMatrix(top, bottom_diag)
```
fritzo, line 531: Do I understand correctly that 'covariance' here means "covariance of z gradients" rather than "covariance of samples"?

pyro/infer/mcmc/adaptation.py
```diff
@@ -350,3 +352,247 @@ def unscale(self, r):
...
+    :param head_size: Head sizes of arrowhead blocks. If this is an integer, all arrowhead blocks
+        will have the same head size. Otherwise, this is a dict that maps site names to the
+        corresponding arrowhead mass matrix. This generalizes the block structure because
+        the special cases
+            + `head_size=0` corresponds to a diagonal mass matrix
+            + `head_size=mass_matrix_size` corresponds to a dense mass matrix
+    :type head_size: int or dict
+    :param float init_scale: initial scale to construct the initial mass matrix.
+    """
+    def __init__(self, head_size=0, init_scale=1.):
```
fehiepsi, line 443: Add `regularize=True` and `regularize_size=(1e-3, 5)` args.

pyro/infer/mcmc/adaptation.py
```diff
@@ -350,3 +352,247 @@ def unscale(self, r):
...
+    A, B = x.top[:, :head_size], x.top[:, head_size:]
+    # NB: the complexity is O(N * head_size^2)
+    # ref: https://en.wikipedia.org/wiki/Schur_complement#Background
+    Dsqrt = x.bottom_diag.sqrt()
+    B_Dsqrt = B / Dsqrt.unsqueeze(-2)  # shape: head_size x N
+    schur_complement = A - B_Dsqrt.matmul(B_Dsqrt.t())  # complexity: head_size^2 x N
+    # we will decompose schur_complement to U @ U.T (so that the sqrt matrix
+    # is upper triangular) using some `flip` operators:
+    #   flip(cholesky(flip(schur_complement)))
+    top_left = torch.flip(torch.flip(schur_complement, (-2, -1)).cholesky(), (-2, -1))
```
martinjankowiak, line 372: this isn't important but why are these flips necessary? it seems like you should be able to avoid them....

pyro/infer/mcmc/adaptation.py
martinjankowiak: i agree that it would be nice to move some of this to `ops` 👍

pyro/infer/mcmc/adaptation.py
fehiepsi: You're right that those flips aren't necessary. But having an upper triangular form is convenient for testing. This also makes those sqrt forms consistent with BlockMassMatrix (its sqrt/sqrt_inverse is also upper triangular). And we just flip several times (at the end of each adaptation windows) during adaptation, so their effects are negligible.

tests/infer/mcmc/test_nuts.py
fehiepsi: @fritzo A more comprehensive test is in `test_arrowhead_mass`. I will use this dirichlet categorical model as a smoke test for gradient adaptation strategy.

pyro/infer/mcmc/adaptation.py
fehiepsi: I have moved those stuffs to ops and renamed the lengthy names into `triu_inverse`, `triu_matvecmul`, `triu_gram` (because those maths are true for general positive definite triu arrowhead matrices).

pyro/infer/mcmc/adaptation.py
```diff
@@ -348,5 +349,182 @@ def unscale(self, r):
...
+
+
+class ArrowheadMassMatrix:
+    """
+    EXPERIMENTAL This class is used to adapt (inverse) mass matrix and provide useful
+    methods to calculate algebraic terms which involves the mass matrix.
+
+    The mass matrix will have arrowhead structure, which can be specified by
+    using the method :meth:`configure` with the corresponding structured
+    `mass_matrix_shape` arg together with the arg `head_size` in the constructor.
```
fritzo, line 363: nit: update comment about `head_size`

tests/infer/mcmc/test_nuts.py
```diff
@@ -412,3 +411,52 @@ def model(cov):
...
+    def model(data):
+        concentration = torch.tensor([1.0, 1.0, 1.0])
+        p_latent = pyro.sample('p_latent', dist.Dirichlet(concentration))
+        pyro.sample("obs", dist.Categorical(p_latent), obs=data)
+        return p_latent
+
+    true_probs = torch.tensor([0.1, 0.6, 0.3])
+    data = dist.Categorical(true_probs).sample(sample_shape=(torch.Size((2000,))))
+    nuts_kernel = NUTS(model, jit_compile=True, ignore_jit_warnings=True)
+    nuts_kernel.mass_matrix_adapter = ArrowheadMassMatrix()
```
martinjankowiak, line 457: i'm still a bit confused by the API. what will the headsize here be?

pyro/infer/mcmc/adaptation.py
fehiepsi: Oops... nice catch, thanks!

tests/infer/mcmc/test_nuts.py
fehiepsi: Here `head_size=0` :D (if `full_mass=True`, head_size=3). This basically tests if grad_adapt works. Does it make sense to you?

@fritzo Is this the behavior you expected?

fritzo: Yes, this is the behavior I expected. In particular, if we would see
- `full_mass=False` ==> `head_size=0` 
- `full_mass=True` ==> `head_size=3` 
- `full_mass=[("p_latent",)]` ==> `head_size=3`

PR #2467, opened 2020-05-07
Update to PyTorch 1.5

fritzo:
- [x] update version pins
- [x] remove `VonMises` which is now upstream (this was giving me warnings whenever I make docs)
- [x] remove `MultivariateNormal.precision_matrix` which is now upstream
- [x] remove `TanhTransform` now that it is available upstream

Test Failures:
 - [x] Fix pickle deprecation in favor of torch.save/load for tensor objects.
 - [x] Fix deprecated signature of in-place tensor ops.
 - [x] Remove deprecated `epoch` arg from scheduler. cc. @jpchen 
 - [x] MCMC with pyro.param statements does not work.
 - [x] NamedTensor not supported in tracer (test_jit errors).


tests/distributions/test_von_mises.py
```diff
@@ -9,10 +9,53 @@
 from torch import optim
 
 from pyro.distributions import VonMises, VonMises3D
-from pyro.distributions.von_mises import _log_modified_bessel_fn
 from tests.common import skipif_param
 
 
+def _eval_poly(y, coef):
```
fritzo, line 15: This testing code was moved from `von_mises.py` which was removed.

fritzo:
@stefanwebb do you recall anything else we should remove? Should we remove `TanhTransform` now that it is available upstream?

pyro/distributions/transforms/neural_autoregressive.py
```diff
@@ -82,33 +82,6 @@ def leaky_relu():
...
-
-    def _call(self, x):
-        return torch.tanh(x)
-
-    def _inverse(self, y):
-        eps = torch.finfo(y.dtype).eps
-        return self.atanh(y.clamp(min=-1. + eps, max=1. - eps))
-
-    def log_abs_det_jacobian(self, x, y):
-        return - 2. * (x - math.log(2.) + F.softplus(- 2. * x))
```
fritzo, line 109: @stefanwebb note that the torch version does not clamp and instead suggests using `cache_size=1`

https://github.com/pytorch/pytorch/blob/99349393ba9bc9b21a6fcb5c7cefd1738f9a6b98/torch/distributions/transforms.py#L431-L438

tests/distributions/test_von_mises.py
neerajprad: Should we move all of these tests to pytorch too? cc. @ahmadsalim

ahmadsalim: Yes, I can do that 😄

pyro/distributions/transforms/neural_autoregressive.py
stefanwebb: I think it should be fine to remove the `Tanh` from Pyro and use the one from PyTorch that doesn't do clamping

stefanwebb:
I can't think of anything else that should be removed in the transforms library

tests/distributions/test_von_mises.py
fritzo: @ahmadsalim I think it's fine for those tests to live in Pyro for now. I'm pretty unhappy with PyTorch's test_distributions.py file: it is huge, monolithic, and uses poor testing practices. I'd like to split it up and start using PyTest. After we do that it should be easy to simply copy these tests over rather than rewrite them to avoid pytest (and later have to rewrite them to use pytest).

fritzo: @neerajprad maybe we can put some time into splitting up PyTorch's test_distributions.py and starting to use pytest there https://github.com/pytorch/pytorch/issues/11578 , and afterwards move @ahmadsalim 's tests?

neerajprad: That sounds good. It will be worth reorganizing our monolithic `test_distributions.py` using pytest first before adding more to the file. Most pytorch devs already use pytest as a test runner, but for some reason there has been reluctance to add it as a test dependency and use all  the testing utilities that it offers.

neerajprad:
I think most fixes are in, except two:
 - NamedTensors isn't supported by JIT and enumeration examples fail with jit enabled (this happens somewhere in Dice). `RuntimeError: NYI: Named tensors are not supported with the tracer` @fritzo - how should we resolve this?
 - There's an odd issue where `pyro.param` statements don't work with JIT in MCMC and fail with `RuntimeError: !t.requires_grad() INTERNAL ASSERT FAILED at ../torch/csrc/jit/ir/constants.cpp:102, please report a bug to PyTorch.` This is despite @fehiepsi's fix from https://github.com/pyro-ppl/pyro/pull/2355 and is probably just a PyTorch bug, but we need to isolate the issue.

fritzo:
> Named tensors isn't supported by the JIT

I wasn't aware of any use of named tensors in Pyro. @eb8680 are you aware of any usage? This might simply be a PyTorch bug.

eb8680:
I think it's just a PyTorch bug

fritzo:
@neerajprad wdyt of simply marking these with `pytest.mark.skipif(torch.__version__ == "1.5.0", reason="PyTorch bug")`? It would be nice to also make it easy for PyTorch devs to run exactly these tests, but I don't know a way to do so other than by grepping. One nice aspect of this strategy is that the bugs will be loud if PyTorch devs test Pyro the next PyTorch release.

neerajprad:
> It would be nice to also make it easy for PyTorch devs to run exactly these tests, but I don't know a way to do so other than by grepping. One nice aspect of this strategy is that the bugs will be loud if PyTorch devs test Pyro with the next PyTorch release.

I think that sounds reasonable. I would still like to find a work-around if possible and update the torch dependency (I think all tweaks are backward compatible). At least in the last release, the test suite was run and they made a best effort attempt at fixing most of the bugs that were surfaced. The more obscure ones like the named tensor bug, I wouldn't expect that our tests will be very useful for them to diagnose the issue, and they'll likely need a smaller reproducible example from us that is independent of Pyro.

fritzo:
Looks like there was an easy fix for the spurious named tensor error: `probs[mask] -> probs.masked_select(mask)`. I think that was the last error 🤞

neerajprad:
Ready to merge?

fritzo:
Ready!

Issue #2464, opened 2020-05-07
AffineCoupling uses NHWC, PyTorch uses NCHW [feature request]


akern40:
### Issue Description
The `AffineCoupling` transformation module splits the incoming tensor along its last dimension:
```{python}
x1, x2 = x[..., :self.split_dim], x[..., self.split_dim:]
```
However, affine coupling layers split their inputs along channels. Since the convention in PyTorch is to use NCHW ordering, a naive usage of the `AffineCoupling` transformation would split data along the width of a tensor.

This could be fixed with two `Permute` modules that change the data from NCHW to NHWC and back again, but it seems like Pyro should conform to the NCHW standard of PyTorch.

### Replication
To replicate, you can perform the following:
```{python}
import torch, pyro

loc = torch.zeros(512, 32, 14, 14)  # A common NCHW set of values
scale = torch.eye(14)
base_dist = pyro.distributions.MultivariateNormal(loc, covariance_matrix=scale)

# A basic convolutional NN for the hypernet
hypernet = torch.nn.Sequential(
    torch.nn.Conv2d(16, 512, 3, padding=1),
    torch.nn.ReLU(),
    torch.nn.Conv2d(512, 512, 1),
    torch.nn.ReLU(),
    torch.nn.Conv2d(512, 32, 3, padding=1)
)
hnet_fn = lambda x: hypernet(x).chunk(2, dim=1)

ac = pyro.distributions.transforms.AffineCoupling(16, hnet_fn)
y = ac(x)
```
I'd expect that `AffineCoupling` would break up the tensor along the channel dimension (canonically dimension 1) at the 16th channel, but instead, it breaks it up into two tensors: one of shape `[512, 32, 14, 0]`, and the other of size `[512, 32, 14, 14]`. This gives the following error:
```{python}
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-245-bb8c77bf15c5> in <module>()
----> 1 ac(base.sample())

/torch/distributions/transforms.py in __call__(self, x)
    127         if x is x_old:
    128             return y_old
--> 129         y = self._call(x)
    130         self._cached_x_y = x, y
    131         return y

/pyro/distributions/transforms/affine_coupling.py in _call(self, x)
    105         x1, x2 = x[..., :self.split_dim], x[..., self.split_dim:]
    106 
--> 107         mean, log_scale = self.nn(x1)
[...]

PR #2472, opened 2020-05-09
Added `dim` keyword arg

stefanwebb:
This adds a `dim` keyword to the `transforms.Permute` and `transform.AffineCoupling` to define which dimension these transforms operate on. As discussed in #2464 this extra flexibility is useful when dealing with matrix/tensor/etc. valued random variables


pyro/distributions/transforms/affine_coupling.py
```diff
@@ -83,12 +86,16 @@ class AffineCoupling(TransformModule):
     domain = constraints.real
     codomain = constraints.real
     bijective = True
-    event_dim = 1
 
-    def __init__(self, split_dim, hypernet, log_scale_min_clip=-5., log_scale_max_clip=3.):
+    def __init__(self, split_dim, hypernet, dim=-1, log_scale_min_clip=-5., log_scale_max_clip=3.):
```
fritzo, line 90: nit: Could you separate args from kwarg with a `*,` to avoid unintentional breakage due to this change:
```py
def __init__(self, split_dim, hypernet, *, dim=-1, log_scale_min_clip=-5., log_scale_max_clip=3.):
```
This forces users to specify optional kwargs by name rather than position. This may break the 0.01 users who previously specified those args positionally, but at least it will not break their code silently.

pyro/distributions/transforms/affine_coupling.py
```diff
@@ -102,7 +109,7 @@ def _call(self, x):
         :class:`~pyro.distributions.TransformedDistribution` `x` is a sample from
         the base distribution (or the output of a previous transform)
         """
-        x1, x2 = x[..., :self.split_dim], x[..., self.split_dim:]
+        x1, x2 = x.split([self.split_dim, x.size(self.dim) - self.split_dim], dim=self.dim)
```
fritzo, line 112: Nice, I hadn't know about that torch function!

stefanwebb:
@akern40 has found a few bugs in how the dimensions are handled, I'll make the changes plus Fritz' requests

pyro/distributions/transforms/affine_coupling.py
stefanwebb: Done!

stefanwebb:
Made the requested changes!

akern40:
> Could we add some simple smoke tests to ensure shapes are handled correctly with dim= -1, -2, -3?

@stefanwebb do you want to handle that, or I can?

stefanwebb:
@fritzo @akern40 In the process of adding tests I discovered a number of issues with adding a dim keyword to these classes... I had to rethink how tests were handled for `event_dim>1` classes and add in some NN resizing.

I think in an interface redesign I would make the helper functions to create transforms take an `event_shape` rather than an `input_dim`, or perhaps they could input a distribution or random variable that already has an event_shape.

All tests pass now!

PR #2473, opened 2020-05-09
Refactor HMC to support structured mass matrix

fehiepsi:
This is a part of #2465.

### Motivation

There are several motivations for this refactoring:
+ It will be cleaner to separate out logics involving inverse_mass_matrix. Now, all ingredients are available in BlockMassMatrix class.

+ It will be faster because some duplicated work can be removed. For example, to compute kinetic energy, we compute r^T @ (M^-1 @ r). Equivalently, we can compute it using the formula ((r^T @ L) @ (L.T @ r)) where L is Cholesky of M^-1 (the Cholesky of M^-1 is available when we convert `precision` to `scale_tril` in `r_dist`). Now, we can carry the term `L.T @ r` (in the implementation, I called this `r_unscaled`) around to compute e.g. U-turn condition (this time, there will be no matrix-vector product).

+ It will be easier to implement other mass matrix structures. For example, ArrowheadMassMatrix in #2465 can be implemented with the same methods as in BlockMassMatrix (but instead of storing `inverse_mass_matrix`, we will store `mass_matrix`). Kernel classes HMC/NUTS only need to access `scale` / `unscale` / `kinetic_grad` method to compute required terms.

### API changes

+ `r_dist`, `inverse_mass_matrix` are removed from `WarmupAdapter` class because all the logics are deferred to mass_matrix_adapter.
+ velocity_verlet now requires `kinetic_grad` explicitly instead of `inverse_mass_matrix`. I think this is good because it allows us to use non-euclidean kinetic energy (not important though). In addition, I will exploit the structure of the mass matrix to save computational cost for kinetic_grad in Arrowhead adapter.
+ Now, in adaptation, I use WelfordArrowheadCovariance to cover all three cases: dense (`head_size = matrix_size`), diagonal (`head_size = 0`), and arrowhead.

### Profiling

 + logistic regression model gives the same results and faster than before (for both kernels HMC/NUTS)
+ neutra example gives similar result (somehow the result is a bit better now) and faster


fehiepsi:
@fritzo @neerajprad Please let me know what you think about the change. I believe this refactoring is good to have (especially, it will save many matrix-vector products when using dense mass).

pyro/infer/mcmc/hmc.py
```diff
@@ -194,34 +186,37 @@ def _find_reasonable_step_size(self, z):
...
+        for site_names, value in self.inverse_mass_matrix.items():
+            size = value.size(0)
+            r_unscaled[site_names] = pyro.sample(
+                "{}_{}".format(name, site_names), dist.Normal(value.new_zeros(size), value.new_ones(size)))
+
+        r = self.mass_matrix_adapter.scale(r_unscaled, r_prototype=self.initial_params)
+        return r, r_unscaled
+
+    @property
+    def mass_matrix_adapter(self):
```
fehiepsi, line 210: This is exposed to allow users to choose another MassMatrix class.

pyro/infer/mcmc/nuts.py
```diff
@@ -30,8 +28,8 @@ def _logaddexp(x, y):
 # weight is the number of valid points in case we use slice sampling
 #   and is the log sum of (unnormalized) probabilites of valid points
 #   when we use multinomial sampling
-_TreeInfo = namedtuple("TreeInfo", ["z_left", "r_left", "z_left_grads",
-                                    "z_right", "r_right", "z_right_grads",
+_TreeInfo = namedtuple("TreeInfo", ["z_left", "r_left", "r_left_unscaled", "z_left_grads",
+                                    "z_right", "r_right", "r_right_unscaled", "z_right_grads",
```
fehiepsi, line 32: We don't want to recompute those unscaled terms. So I expand the TreeInfo to store that extra information.

pyro/infer/mcmc/hmc.py
```diff
@@ -322,28 +319,27 @@ def sample(self, params):
         # recompute PE when cache is cleared
         if z is None:
             z = params
-            potential_energy = self.potential_fn(z)
-            self._cache(z, potential_energy)
+            z_grads, potential_energy = potential_grad(self.potential_fn, z)
+            self._cache(z, potential_energy, z_grads)
```
fehiepsi, line 323: Here we cache `z_grads` too for mass matrix adapters which require it.

pyro/infer/mcmc/hmc.py
```diff
@@ -194,34 +187,38 @@ def _find_reasonable_step_size(self, z):
...
-                pos = next_pos
-            assert pos == r_flat.size(0)
-            r_flat_dict[site_names] = r_flat
-        return r, r_flat_dict
+        r_unscaled = {}
+        for site_names, value in self.inverse_mass_matrix.items():
+            size = value.size(0)
+            r_unscaled[site_names] = pyro.sample(
+                "{}_{}".format(name, site_names),
+                NonreparameterizedNormal(value.new_zeros(size), value.new_ones(size)))
```
fritzo: Could you explain what's going on here, and maybe add a comment explaining the use of `NonreparameterizedNormal` rather than `Normal`?

pyro/infer/mcmc/nuts.py
```diff
@@ -162,30 +160,23 @@ def __init__(self,
         # Here, as suggested in [1], we set dE_max = 1000.
         self._max_sliced_energy = 1000
 
-    def _is_turning(self, r_left, r_right, r_sum):
+    def _is_turning(self, r_left_unscaled, r_right_unscaled, r_sum):
```
fritzo, line 163: I'm curious, does this mean the turning condition is defined in the original coordinates or in the adapted coordinates? I would have thought it would be defined in the adapted coordinates, as it is for reparameterized models.

pyro/ops/welford.py
```diff
@@ -48,3 +48,46 @@ def get_covariance(self, regularize=True):
                 scaled_cov.view(-1)[::scaled_cov.size(0) + 1] += shrinkage
                 cov = scaled_cov
         return cov
+
+
+class WelfordArrowheadCovariance:
```
fritzo: Would this interface also be able to support arrowhead precision structure, or would we need to replace `.get_covariance()` with a matrix-vector-product interface?

tests/ops/test_welford.py
```diff
@@ -35,7 +35,7 @@ def test_welford_diagonal(n_samples, dim_size):
 @pytest.mark.parametrize('n_samples,dim_size', [(1000, 1),
                                                 (1000, 7),
                                                 (1, 1)])
-@pytest.mark.init(rng_seed=7)
+# @pytest.mark.init(rng_seed=7)
```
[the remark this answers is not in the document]
fehiepsi: I have removed it in recent commits.

tests/infer/mcmc/test_nuts.py
```diff
@@ -181,8 +181,10 @@ def model(data):
                        adapt_mass_matrix=adapt_mass_matrix, full_mass=full_mass)
     mcmc = MCMC(nuts_kernel, num_samples=400, warmup_steps=200)
     mcmc.run(data)
+    print(mcmc.kernel.inverse_mass_matrix)
```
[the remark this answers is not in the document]
fehiepsi: I have removed this line (I used it just to test my implementation).

pyro/infer/mcmc/hmc.py
fehiepsi: There is nothing special here. It is a bit faster than `Normal` I guess (I'll add a comment). I think `torch.normal(loc, scale)` is faster than `loc  + scale * torch.randn(shape)`.

pyro/infer/mcmc/nuts.py
fehiepsi: Originally, the turning condition is defined by `r_left . (z_right - z_left)`, which is approximated by `r_left . M^-1 . sum(r)` (because for each leapfrog step, `z' - z = M^-1 r'`). So if we unscale those `r`, it is `r_left_unscaled . sum(r_unscaled)`. I think we need to keep the definition of the turning condition.

pyro/ops/welford.py
fehiepsi: Hi @fritzo, I have moved WelfordArrowheadCovariance to #2465. In #2465, `get_covariance` will return the `top` part and `bottom_diag` part of the arrowhead matrix (to save memory).

pyro/infer/mcmc/adaptation.py
```diff
@@ -194,25 +176,177 @@ def step(self, t, z, accept_prob):
...
+            self._inverse_mass_matrix[site_names] = inverse_mass_matrix
+            self._mass_matrix_sqrt[site_names] = mass_matrix_sqrt
+            self._mass_matrix_sqrt_inverse[site_names] = mass_matrix_sqrt_inverse
 
-    @property
-    def r_dist(self):
-        return self._r_dist
+    def configure(self, mass_matrix_shape, adapt_mass_matrix=True, options={}):
+        """
+        Setups initial mass matrix.
```
neerajprad, line 254: nit: `Sets up`..

pyro/infer/mcmc/adaptation.py
```diff
@@ -194,25 +176,177 @@ def step(self, t, z, accept_prob):
...
+                v[site_name] = v_flat[pos:next_pos].reshape(r[site_name].shape)
+                pos = next_pos
+        return v
+
+    def scale(self, r_unscaled, r_prototype):
+        """
+        Computes `M^{1/2} @ r_unscaled`.
+
+        Note that `r` is generated from a gaussian with scale `mass_matrix_sqrt`.
+        This method will unscale it.
```
neerajprad, line 319: Should be `scale it`.

pyro/infer/mcmc/adaptation.py
```diff
@@ -194,25 +176,177 @@ def step(self, t, z, accept_prob):
...
+def _upper_triangular_inverse(x):
+    if x.dim() == 1:
+        return x.reciprocal()
+    else:
+        identity = torch.eye(x.size(-1), dtype=x.dtype, device=x.device)
+        return torch.triangular_solve(identity, x, upper=True)[0]
+
+
+class BlockMassMatrix:
+    """
```
neerajprad, line 210: Do you foresee any changes to the interface? If so, we should mark this as experimental in the docstring.

pyro/infer/mcmc/adaptation.py
```diff
@@ -194,25 +176,177 @@ def step(self, t, z, accept_prob):
...
+    def update(self, z, z_grad):
+        """
+        Updates the adaptation scheme using the new sample `z` or its grad `z_grad`.
+
+        :param dict z: the current value.
+        :param dict z_grad: grad of the current value.
+        """
+        for site_names, adapt_scheme in self._adapt_scheme.items():
+            z_flat = torch.cat([z[name].reshape(-1) for name in site_names])
+            adapt_scheme.update(z_flat.detach())
```
neerajprad, line 282: nit: Maybe detach earlier `z[name].detach()` (or even earlier if possible)?

fehiepsi: yes, I think it is better.

pyro/infer/mcmc/adaptation.py
fehiepsi: I think the interface is good. We might add more features to the constructor if needed. But I'll mark this as Experimental.

fehiepsi:
@fritzo @neerajprad Many thanks for reviewing!

pyro/ops/welford.py
fritzo: I'm confused. I see this could possibly return a compact representation of an arrowhead covariance matrix, but my question is about arrowhead precision matrices. In contrast to arrowhead covariance matrices, the covariance corresponding to an arrowhead precision matrix is dense. Stated another way, if we implement not only a `WelfordArrowheadCovariance` but also a `WelfordArrowheadPrecision` whose covariance is dense, will this `.get_covariance()` method still be tractable, or will we need to implement a custom matrix-vector product op (which is cheap for arrowhead-precisions; we can solve against a sparse precision matrix).

fehiepsi: Yes, in #2465, we learn arrowhead precision matrix (i.e. arrowhead mass matrix). The covariance (inverse mass matrix) will be dense. But the sqrt of the covariance will be a sparse triangular matrix. We will leverage this sparse structure to cheaply compute matrix-vector product with O(N x head_size^2) cost.

fritzo: Thanks for explaining!

PR #2468, opened 2020-05-08
Add coalescent likelihoods to contrib.epidemiology

eb8680:
Addresses #2426 
Pair-coded with @fritzo 

This PR adds a new example model that includes a phylogenetic term.

(part of this PR was moved to #2474:)

<details>
This PR adds a phylogenetic likelihood translated from EpiGenMCMC to `contrib.epidemiology`, and adds a new example model that includes a phylogenetic term.  There are two implementations of the likelihood here, a vectorized and non-vectorized version (in coalescent event sequence and hence in simulation time). In smoke testing with the example in this PR, the two appear to behave identically in `transition_fwd`, and the vectorized version should be easy (with a bit of judicious masking) to integrate with our parallel-scan inference machinery.
</details>

## Tasks:
- [x] Add sequential likelihood implementation
- [x] Add naive vectorized likelihood implementation
- [x] Add a temporary script for preprocessing data
- [x] Add a simple example and use it as a smoke test with some real data
- [x] Integrate the vectorized likelihood together with some naively tensorized data into `transition_bwd`
- [x] Convert likelihoods to `TorchDistribution`s
- [x] Add unit tests that check math and broadcasting
- [x] Add documentation
- [x] Investigate less memory-intensive approaches to vectorized likelihood evaluation, perhaps by converting the data into PyTorch sparse tensors
- [x] Usability: allow non-sorted `coal_times` or at least explain error
- [x] Usability: consider allowing negative `coal_times`
- [ ] Add unit tests that check against some ground-truth likelihoods (blocked by lack of ground truth)
- [ ] Remove temporary preprocessing script with R/Rpy2 dependency

PR #2474, opened 2020-05-10
Implement a coalescent time distribution

fritzo:
Addresses #2426 
Blocking #2468 

This adds two new distributions `CoalescentTimes` and `CoalescentTimesWithRate` and a class `CoalescentRateLikelihood` implementing a phylogenetic likelihood
```
p(coalescent times | sample times, base rate)
```
by splitting a phylogeny into fixed sample times (i.e. EPI nodes, leaves) and unknown coalescent times (i.e. internal NODE nodes). For `...WithRates`, the base rate is specified as a vector representing a piecewise constant base coalescent rate on a uniform grid (e.g. computed from `S` and `I` time series in SIR models).

The design aims to make it cheap to enumerate over multiple phylogenetic samples (e.g. from BEAST). Two crucial observations are: (1) among phylogeny samples, the leaves are fixed, hence the total number of nodes is fixed; and (2) we can avoid summing over intervals by precomputing `rate.cumsum(-1)` and using this for all samples. Exploiting these facts leads to an `O(T + S N log(N))` algorithm where `T` is the number of simulation time steps, `N` is the number of leaves in the phylogeny, and `S` is the number of sampled phylogenies. This PR does not implement batching over different phylogenetic problems (e.g. one phylogeny per region in regional models); that would require more careful masking and padding and will be left to a possible future PR.

These are low level distribution and do not implement preprocessing to transform phylogenetic trees into `(coal_times, leaf_times)` pairs. This PR is intended as a low-level helper for use in #2468 and an example using `pyro.contrib.epidemiology`.

## Tested
- [x] shape tests
- [x] compare against uniform rate
- [x] added smoke tests for use in an SEIR model
- [x] visually sanity check samples in [a notebook](https://github.com/fritzo/notebooks/blob/master/coalescent.ipynb)


pyro/distributions/coalescent.py
```diff
@@ -0,0 +1,306 @@
...
+    @functools.wraps(fn)
+    def memoized_fn(*args):
+        key = tuple(map(id, args))
+        if key not in cache:
+            cache[key] = fn(*args)
+            for arg in args:
+                weakref.finalize(arg, cache.pop, key, None)
+        return cache[key]
+
+    return memoized_fn
```
fritzo, line 223: @neerajprad could you review this code? I wonder if it is interacting poorly with PyTorch?

fritzo: No the failure occurs even without this, so this can't be the problem...

fritzo:
@eb8680 FYI I am adding an interface that plays better with `.transition_bwd()` as we discussed.

fritzo:
I've added a `CoalescentRateLikelihood` class compatible with both vectorized and sequential operation, and copied more of your #2468 into this to exercise the likelihood in an SEIR model. However this PR still deals entirely with `(leaf_times, coal_times)`.

I think it would be cleanest to merge this PR, then update #2468 to use `CoalescentRateLikelihood` and update the example script there to parse a `.nwk` phylogeny and extract to `(leaf_times, coal_times)` either using your r2py code or the `Bio.Phylo` python library.

pyro/contrib/epidemiology/seir.py
```diff
@@ -240,3 +262,8 @@ def transition_bwd(self, params, prev, curr, t):
         pyro.sample("obs_{}".format(t),
                     dist.ExtendedBinomial(S2E, rho),
                     obs=self.data[t])
+        if self.coal_likelihood is not None:
+            R = R0 * prev["S"] / self.population
+            coal_rate = R * (1. + 1. / k) / (tau_i * prev["I"] + 1e-8)
+            pyro.factor("coalescent_{}".format(t),
+                        self.coal_likelihood(coal_rate, t))
```
eb8680, line 269: This seems like a good interface

pyro/distributions/coalescent.py
```diff
@@ -0,0 +1,428 @@
...
+        i = coal_times.floor().clamp(min=0).long()
+        rates = phylogeny.coal_binomial * _gather(self.rate_grid, -1, i)
+        log_prob = log_prob + safe_log(rates).sum(-1)
+
+        batch_shape = broadcast_shape(self.batch_shape, value.shape[:-1])
+        log_prob = log_prob.expand(batch_shape)
+        return log_prob
+
+
+class CoalescentRateLikelihood:
```
eb8680, line 195: Should this be a `nn.Module`?

fritzo: At the moment there is no reason to make this an `nn.Module`. It is a little more similar to a `Distribution`, but doesn't quite follow the distribution interface (indeed I've given it a `validate_args` kwarg like a distribution). Moreover in typical usage the parameters are all fixed rather than learned.

BTW this is a very natural object from the perspective of Funsor. Consider a funsor
```py
leaf_times = Tensor(...)
coal_times = Tensor(...)
rate_grid = Variable("rate_grid", reals(duration))
f = CoalescentTimesWithRate(leaf_times, rate_grid).log_prob(coal_times)
```
Then this likelihood represents roughly the funsor
```py
f(rate_grid=rate_grid["t"])
```
I look forward to the bright future when we won't need special classes like this 🚀

fritzo:
Thanks for reviewing!

PR #2486, opened 2020-05-16
Add ESS thoreshold logic to SMCFilter

fritzo:
Addresses #2426 

This adds an ESS-based threshold to `SMCFilter` to decide when to importance resample (resolving a long-standing TODO). The default is to resample when `ESS < num_particles / 2`. I've also threaded this through the epidemiological `CompartmentalModel` because I've found a lower threshold can help preserve diversity and find a feasible solution in #2468.

I've also added a couple checks to raise `ValueError` asap when `SMCFilter` fails to find a feasible hypothesis.

## Tested
- [x] logic updates are covered by existing tests for both `SMCFilter` and `CompartmentalModel`
- [x] manually logged ess values to sanity check dynamics
- [x] evaluated improved performance on epidemiological models in #2468

### DIFF
--- a/pyro/infer/smcfilter.py
+++ b/pyro/infer/smcfilter.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import contextlib
+import math
 
 import torch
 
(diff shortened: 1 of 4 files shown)


pyro/infer/smcfilter.py
```diff
@@ -104,16 +111,28 @@ def _update_weights(self, model_trace, guide_trace):
...
 
     def _maybe_importance_resample(self):
-        if self.state:  # TODO check perplexity
+        if not self.state:
+            return
+        # Decide whether to resample based on ESS.
+        logp = self.state._log_weights
+        logp -= logp.logsumexp(dim=-1)
+        ess = logp.mul(2).exp().sum().reciprocal()
+        if ess < self.ess_threshold * self.num_particles:
```
fritzo, line 135: Derivation:
```
ess = 1 / sum(p[i] ** 2 for i in range(N))
    = 1 / sum((logp[i] * 2).exp() for i in range(N))
    = logp.mul(2).exp().sum().reciprocal()
```

PR #2488, opened 2020-05-16
Use systematic resampling in SMCFilter

fritzo:
Addresses #2426 

This PR
1. switches from multinomial resampling to systematic resampling (i.e. quasi-random sampling) in the `SMCFilter`, increasing sample diversity.
2. switches from `ValueError` to a new exception `SMCFailed` and uses retry logic in `CompartmentalModel` in case of failure.

## Tested
- [x] added a unit test for `_systematic_sample()`
- [x] behavior change is covered by existing tests
- [x] ensured retry logic works on a hard test case


fritzo:
Thanks for reviewing!

PR #2484, opened 2020-05-15
Add Poisson approximation to Binomial.sample()

fritzo:
Addresses #2426 

This adds a `Poisson` approximation to the `.sample()` methods of `Binomial` and `BetaBinomial`, gated by an experimental kwarg `approx_sample_thresh`.

After this PR the approximation will be used in `pyro.contrib.epidemiology` with default threshold of 10000 when sampling and with a reduced threshold of 1000 during heuristic initialization (which can tolerate higher errors). This dramatically speeds up heuristic initialization (by ~100x for population 1000000) and appears to enable arbitrarily large populations, e.g. the following no longer crashes:
```sh
$ python examples/contrib/epidemiology/sir.py -e 4 -p 100000000
Simulating from a SimpleSEIRModel
Observed 1154/1746 infections:
51 53 66 54 55 67 53 56 43 44 40 44 63 65 61 63 56 69 86 65
INFO 	 Heuristically initializing...
INFO 	 Running inference...
```
@eb8680 you might now be able to run `sir.py` on your machine 😆

## Tested
- [x] accuracy tests
- [x] exercised in epidemiology models

PR #2490, opened 2020-05-19
Revise binomial approximation interface

fritzo:
Addresses #2426 

The aim of this PR is to lower compute cost of large regional models. Before this PR, regional models with large population are prohibitively slow to initialize due to `Binomial.sample()` cost. After this PR I can easily run regional models with large heterogeneous populations.

This follows #2484 by:
1. generalizing the Binomial approximation logic to allow heterogeneous `total_count` (e.g. for regional models with different populations in different regions);
2. making the binomial approximation a global variable (rather than an __init__ arg), thereby enabling approximation of all distributions used in `contrib.epidemiology` without user intervention; and
3. setting approximation thresholds for three methods of `CompartmentalModel`:
    - for `.heuristic()`, threshold = 100
    - for `.generate()`, threshold = 1000
    - for `.predict()`, threshold = 10000

## Tested
- [x] ran a model that freezes my compute before this PR and works after:
    `python examples/contrib/epidemiology/regional.py -p 1000000 -r 10 -d 60`
- [x] successfully ran a colab notebook with large heterogeneous population.

### DIFF
--- a/pyro/contrib/epidemiology/distributions.py
+++ b/pyro/contrib/epidemiology/distributions.py
@@ -8,25 +8,36 @@ import torch
 
 import pyro.distributions as dist
 
-_APPROX_SAMPLE_THRESH = 10000
-
 
 @contextmanager
 def set_approx_sample_thresh(thresh):
     """
-    Temporarily set global approx_sample_thresh in ``infection_dist``.
-    The default global value is 10000.
+    EXPERIMENTAL Temporarily set the global default value of
+    ``Binomial.approx_sample_thresh``, thereby decreasing the computational
+    complexity of sampling from :class:`~pyro.distributions.Binomial`,
+    :class:`~pyro.distributions.BetaBinomial`,
+    :class:`~pyro.distributions.ExtendedBinomial`,
+    :class:`~pyro.distributions.ExtendedBetaBinomial`, and distributions
+    returned by :func:`infection_dist`.
+
+    This is useful for sampling from very large ``total_count``.
+
+    This is used internally by
+    :class:`~pyro.contrib.epidemiology.compartmental.CompartmentalModel`.
 
     :param thresh: New temporary threshold.
     :type thresh: int or float.
     """
-    global _APPROX_SAMPLE_THRESH
-    old = _APPROX_SAMPLE_THRESH
+    assert isinstance(thresh, (float, int))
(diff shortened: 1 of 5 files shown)


pyro/contrib/epidemiology/distributions.py
```diff
@@ -8,34 +8,44 @@
 
 import pyro.distributions as dist
 
-_APPROX_SAMPLE_THRESH = 10000
```
fritzo, line 11: This global variable is now called `Binomial.approx_sample_thresh`

pyro/distributions/torch.py
```diff
@@ -32,33 +31,16 @@ def _log_normalizer(d):
...
-        new._validate_args = self._validate_args
-        return new
+    # EXPERIMENTAL threshold on total_count above which sampling will use a
+    # clamped Poisson approximation for Binomial samples. This is useful for
+    # sampling very large populations.
+    approx_sample_thresh = math.inf
 
     def sample(self, sample_shape=torch.Size()):
         if self.approx_sample_thresh < math.inf:
-            if self.approx_sample_thresh < self.total_count.min():
```
fritzo, line 60: Previously this would use exact sampling if any region had a small population; now we use exact sampling for small regions and approximate sampling for large regions.

pyro/distributions/torch.py
```diff
@@ -68,7 +50,15 @@ def sample(self, sample_shape=torch.Size()):
...
                     result = torch.min(result + shift, self.total_count)
-                    return torch.where(p < q, result, self.total_count - result)
+                    sample = torch.where(p < q, result, self.total_count - result)
+                # Draw exact samples for remaining items.
+                if exact.any():
+                    total_count = torch.where(exact, self.total_count,
+                                              torch.zeros_like(self.total_count))
+                    exact_sample = torch.distributions.Binomial(
+                        total_count, self.probs, validate_args=False).sample(sample_shape)
+                    sample = torch.where(exact, exact_sample, sample)
```
martinjankowiak, line 60: i'm confused. how does L60 work if L53 didn't execute?

fritzo: that can't happen; the indentation on L53 is for the `with torch.no_grad()`, not an if statement. L60 can only execute if L53 executes.

PR #2495, opened 2020-05-20
Implement a SplitReparam and use it in contrib.epidemiology

fritzo:
Addresses #2426 

This adds a `SplitReparam` reparameterizer to split a sample site tensor into multiple other tensors, as suggested by @fehiepsi. The motivating use case (also implemented in this PR) is splitting Haar-reparameterized sample sites into low- and high-frequency parts, then adding the low frequency parts to the `full_mass` matrix in HMC.

Note this reparameterizer is quite limited: it cannot generate samples because there is no standard way to split a distribution into multiple independent distribution. However in HMC and SVI require only `.log_prob()` to be implemented, not `.sample()`. Actually this PR needed to change HMC internals to avoid calling `.sample()` inadvertently during prototype tracing.

This PR also rebalances some test since `unit` was taking 37 minutes vs `integration_batch_1` taking 15 minutes.

## Tested
- [x] unit tests
- [x] added epidemiology tests
- [x] added to epidemiology examples and tests


pyro/infer/mcmc/util.py
```diff
@@ -382,7 +384,8 @@ def initialize_model(model, model_args=(), model_kwargs={}, transforms=None, max
     # No-op if model does not have any discrete latents.
     model = poutine.enum(config_enumerate(model),
                          first_available_dim=-1 - max_plate_nesting)
-    model_trace = poutine.trace(model).get_trace(*model_args, **model_kwargs)
+    prototype_model = poutine.trace(InitMessenger(init_strategy)(model))
+    model_trace = prototype_model.get_trace(*model_args, **model_kwargs)
```
fritzo, line 388: @fehiepsi could you please review changes in this file? The changes are needed to
1. avoid calling `.sample()` in constructing the prototype `model_trace`, and
2. avoid duplicating expensive initialization work by reusing that `model_trace` as the first trace in `_find_valid_initial_params()`.

pyro/contrib/epidemiology/compartmental.py
```diff
@@ -301,14 +301,23 @@ def fit(self, **options):
             that computational cost is exponential in `num_quant_bins`.
             Defaults to 4.
         :param bool haar: Whether to use a Haar wavelet reparameterizer.
+        :param int haar_full_mass: Number of low frequency Haar components to
```
martinjankowiak, line 304: is it better to subsume the bool arg entirely into the int arg?

fritzo: That seemed weird, since to trigger `haar` without full mass you would need to pass `haar=0`, so we'd need another distinct value to not use `haar` like `haar=-1`. At that point it is weird enough I'd rather simply expose two options.

Also we may some day add other reparameterizers and rename `haar_full_mass` to `low_freq_full_mass` or something.

fritzo: Do you want me to rename it now? Maybe to `lowpass_full_mass`?

martinjankowiak: probably unnecessary

pyro/contrib/epidemiology/compartmental.py
```diff
@@ -301,14 +301,23 @@ def fit(self, **options):
...
         self.num_quant_bins = options.pop("num_quant_bins", 4)
         haar = options.pop("haar", False)
+        assert isinstance(haar, bool)
+        haar_full_mass = options.pop("haar_full_mass", 0)
+        assert isinstance(haar_full_mass, int)
+        assert haar_full_mass >= 0
+        haar_full_mass = min(haar_full_mass, self.duration)
+        haar = haar or (haar_full_mass > 0)
 
         # Heuristically initialze to feasible latents.
```
martinjankowiak, line 322: nit: old typo

pyro/infer/reparam/split.py
```diff
@@ -0,0 +1,90 @@
...
+    def sample(self, sample_shape=torch.Size()):
+        raise NotImplementedError("SplitReparam does not support sampling")
+
+
+class SplitReparam(Reparam):
+    """
+    Reparameterizer to split a random variable along a dimension, similar to
+    :func:`torch.split`.
+
+    This is useful for treating different parts of a tensor with different
```
martinjankowiak, line 50: typo: different other

pyro/infer/mcmc/util.py
fehiepsi: IIUC, then with e.g. `init_to_uniform` strategy, we still [call sample](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/autoguide/initialization.py#L111) but won't for some other strategies? If so, the change looks great to me.

pyro/infer/reparam/split.py
```diff
@@ -0,0 +1,90 @@
...
+        super(_ImproperUniform, new).__init__(batch_shape, self.event_shape)
+        return new
+
+    def log_prob(self, value):
+        batch_shape = value.shape[:value.dim() - self.event_dim]
+        batch_shape = broadcast_shape(batch_shape, self.batch_shape)
+        return torch.zeros(()).expand(batch_shape)
+
+    def sample(self, sample_shape=torch.Size()):
+        raise NotImplementedError("SplitReparam does not support sampling")
```
fehiepsi, line 42: nit: `_ImproperUniform` does not support sampling

fritzo: > nit: _ImproperUniform does not support sampling

This user-facing exception mention the user-facing class; `_ImproperUniform` is an implementation detail.

> I think we can move this to the main distributions module.

I feel like this is too bespoke for general use. Until we find another use case I'd prefer to keep it private. We may find another way to implement this, e.g. `Delta(nan).mask(False)` would raise a Nan error rather than a `NotImplementedError`.

pyro/infer/mcmc/util.py
fritzo: That's correct, `init_to_uniform` still calls sample. Thanks for reviewing!

pyro/infer/reparam/split.py
fritzo: > with an sample_fn arg to generate prototype samples for HMC.

I believe this functionality is already cleanly accomplished by `InitMessenger`. Do you have another use case in mind?

fehiepsi: Nevermind, I think `dist.Foo(...).mask(False)` can serve the same purpose and initial values can be drawn directly from `Foo` distribution. This class is useful when we don't know the support before-hand.

PR #2498, opened 2020-05-22
Add overdispersed models to contrib.epidemiology

fritzo:
Addresses #2426 

This aims to improve fit to real data and improve mixing by relaxing all `Binomial` distributions to precisely overdispersed `BetaBinomial`s such that the large population limiting dynamics is an SDE rather than a deterministic ODE. These new distributions are used in two new models: `OverdispersedSIRModel` and `OverdispersedSEIRModel`.

## Questions

- [x] Is the proposed `overdispersion` parameter reasonable?
    - Yes; see [this notebook](https://github.com/fritzo/notebooks/blob/master/overdispersed_binomial.ipynb) for visualization
- [x] Should this be a global parameter, a local parameter, or an inferred variable?
    - An inferred variable works well, I've chosen the weak prior `Beta(2, 2)` 

## Tested
- [x] unit tests for distribution changes
- [x] unit tests verifying properties of overdispersion
- [x] smoke tests for use in SIR models
- [x] profiling: switching to `BetaBinomial`s results in ~20% slowdown of likelihood computation


pyro/contrib/epidemiology/distributions.py
```diff
@@ -38,12 +49,132 @@ def set_approx_sample_thresh(thresh):
...
+    :type probs: float or torch.Tensor
+    :param overdispersion: Amount of overdispersion, in the half open interval
+        [0,2). Defaults to zero.
+    :type overdispersion: float or torch.tensor
+    """
+    _validate_overdispersion(overdispersion)
+    if _is_zero(overdispersion):
+        return dist.ExtendedBinomial(total_count, probs)
+    if getattr(probs, "dtype", torch.float64) != torch.float64:
+        warnings.warn("binomial_dist is unstable for dtypes less than torch.float64; "
```
martinjankowiak, line 103: is this really the source?

fritzo: Good point. I've moved the warning to `CompartmentalModel.fit()` and `.predict()`.

fritzo:
@martinjankowiak thanks for your patience. I've isolated the changes to two new models. I plan to test these on real data in the next couple weeks.

PR #2500, opened 2020-05-24
Add a shifted Stirling approximation to log Beta function

fritzo:
Addresses #2426 
Blocking #2498 
See [derivation notebook](https://github.com/fritzo/notebooks/blob/master/beta_function.ipynb).

This implements a cheap approximate `log_beta()` function for use in `Binomial.log_prob()` and `BetaBinomial.log_prob()`. This is motivated by profiling results of #2498 which showed that `torch.lgamma()` was taking more than half of inference time. This PR also refactors to use the new approximation in `CompartimentalModel.heuristic()`, and I plan to use the approximation in more places in #2498.

## Accuracy

Users specify a `tol` parameter which is absolute error (in log space) and is worst for small values of `x,y` in `log_beta(x,y)`. When does this happen in `pyro.contrib.epidemiology`? The only circumstance when one of `x,y` is small is in superspreading models with small `k` and tiny infectious population, so `k * I <= 1`; in all other cases this approximation is more accurate than the stated `tol`.

## Speed

This approximation speeds up `log_beta()` by about 2.5x at loosest tolerance 0.1.

| time | function |
|-|-|
| 116ms | `log_beta(x, y)` |
| **46ms** | `log_beta_stirling(x, y, tol=0.1)` |
| 74ms | `log_beta_stirling(x, y, tol=0.05)` |
| 171ms | `log_beta_stirling(x, y, tol=0.02)` |

## Tested
- [x] added unit tests for `log_beta()` and `log_binomial()` at a variety of tolerances
- [x] added unit test of `Binomial.log_prob()` at a variety of tolerances
- [x] refactoring of `BetaBinomial` is covered by existing tests
- [x] new usage in `CompartmentalModel.heuristic` is covered by superspreader smoke tests

### DIFF
--- a/pyro/distributions/conjugate.py
+++ b/pyro/distributions/conjugate.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2017-2019 Uber Technologies, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
+import functools
 import numbers
 
 import torch
@@ -9,10 +10,7 @@ from torch.distributions.utils import broadcast_all
 
 from pyro.distributions.torch import Beta, Binomial, Dirichlet, Gamma, Multinomial, Poisson
 from pyro.distributions.torch_distribution import TorchDistribution
-
-
-def _log_beta(x, y):
-    return torch.lgamma(x) + torch.lgamma(y) - torch.lgamma(x + y)
+from pyro.ops.special import log_beta, log_beta_stirling
 
 
 def _log_beta_1(alpha, value, is_sparse):
@@ -46,6 +44,12 @@ class BetaBinomial(TorchDistribution):
     has_enumerate_support = True
     support = Binomial.support
 
+    # EXPERIMENTAL If set to a positive value, the .log_prob() method will use a
+    # shifted Sterling's approximation to the Beta function, reducing
+    # computational cost from 9 lgamma() evaluations to four log() evaluations
+    # plus arithmetic. Recommended values are between 0.1 and 0.01.
+    approx_log_prob_tol = 0.
+
     def __init__(self, concentration1, concentration0, total_count=1, validate_args=None):
         concentration1, concentration0, total_count = broadcast_all(
             concentration1, concentration0, total_count)
@@ -77,12 +81,17 @@ class BetaBinomial(TorchDistribution):
     def log_prob(self, value):
         if self._validate_args:
             self._validate_sample(value)
-        log_factorial_n = torch.lgamma(self.total_count + 1)
-        log_factorial_k = torch.lgamma(value + 1)
-        log_factorial_nmk = torch.lgamma(self.total_count - value + 1)
-        return (log_factorial_n - log_factorial_k - log_factorial_nmk +
-                _log_beta(value + self.concentration1, self.total_count - value + self.concentration0) -
-                _log_beta(self.concentration0, self.concentration1))
+
+        if self.approx_log_prob_tol > 0:
+            lbeta = functools.partial(log_beta_stirling, tol=self.approx_log_prob_tol)
+        else:
+            lbeta = log_beta
+
+        n = self.total_count
+        k = value
+        a = self.concentration1
+        b = self.concentration0
+        return lbeta(k + a, n - k + b) - lbeta(a, b) - lbeta(k + 1, n - k + 1) - n.log1p()
 
     @property
     def mean(self):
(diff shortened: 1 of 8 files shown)


pyro/distributions/conjugate.py
```diff
@@ -77,12 +81,17 @@ def sample(self, sample_shape=()):
...
+        if self.approx_log_prob_tol > 0:
+            lbeta = functools.partial(log_beta_stirling, tol=self.approx_log_prob_tol)
+        else:
+            lbeta = log_beta
+
+        n = self.total_count
+        k = value
+        a = self.concentration1
+        b = self.concentration0
+        return lbeta(k + a, n - k + b) - lbeta(a, b) - lbeta(k + 1, n - k + 1) - n.log1p()
```
fritzo, line 94: I have refactored this math to use the `log_beta()` interface rather than calling `lgamma()` directly.

pyro/ops/special.py
```diff
@@ -0,0 +1,67 @@
...
+    for _ in range(shift):
+        factors.append(xy / (x * y))
+        x = x + 1
+        y = y + 1
+        xy = xy + 1
+
+    log_factor = functools.reduce(operator.mul, factors).log()
+
+    return (log_factor + (x - 0.5) * x.log() + (y - 0.5) * y.log()
+            - (xy - 0.5) * xy.log() + (math.log(2 * math.pi) / 2 - shift))
```
fehiepsi, line 67: Where does this `shift` come from?

fritzo: Stirling's approximation of each of the `lgamma` terms is
```
lgamma(x) ~ (x-1/2) log(x) - x + log(2 pi)/2
```
Now when we use this to approximate a `log_beta` function the `x` terms usually cancel
```
log_beta(x,y) = lgamma(x) + lgamma(y) - lgamma(x+y)
              = (x-1/2) log(x) + (y-1/2) log(y) - (x+y-1/2) log(x+y)
              + -x + -y + (x+y)       # <--- these cancel if shift = 0
              + (1+1-1) log(2 pi)/2
```
Now the trick to make this more accurate is to use the `lgamma` recursion and increment each of `x`, `y`, and `x+y`, which is written in the code as
```
x = x + 1
y = y + 1
xy = xy + 1
```
When we do that, the first line works unchanged, the second line gets an extra `-1`, and the last line remains unchanged. And if we increment each of `x`, `y`, and `x+y` `shift`-many times and each time gets an extra `-1` then we can simply add the term `shift`.

fehiepsi: Oh yeah, it comes from the second line (I thought it cancelled out). Thanks for a clear explanation!

fritzo:
I'll quick add this approximation also to `dist.Binomial`, which I hadn't realized also uses `lgamma`

pyro/distributions/torch.py
```diff
@@ -61,6 +73,21 @@ def sample(self, sample_shape=torch.Size()):
...
+        n = self.total_count
+        k = value
+        # k * log(p) + (n - k) * log(1 - p) = k * (log(p) - log(1 - p)) + n * log(1 - p)
+        #     (case logit < 0)              = k * logit - n * log1p(e^logit)
+        #     (case logit > 0)              = k * logit - n * (log(p) - log(1 - p)) + n * log(p)
+        #                                   = k * logit - n * logit - n * log1p(e^-logit)
+        #     (merge two cases)             = k * logit - n * max(logit, 0) - n * log1p(e^-|logit|)
+        normalize_term = n * (_clamp_by_zero(self.logits) + self.logits.abs().neg().exp().log1p())
+        return (k * self.logits - normalize_term
+                + log_binomial(n, k, tol=self.approx_log_prob_tol))
```
fritzo, line 89: This was adapted from the [upstream implementation](https://github.com/pytorch/pytorch/blob/d035d05080729c30636ff30fcc068de3c7e9badd/torch/distributions/binomial.py#L108-L119):

<details>

```py
        log_factorial_n = torch.lgamma(self.total_count + 1)
        log_factorial_k = torch.lgamma(value + 1)
        log_factorial_nmk = torch.lgamma(self.total_count - value + 1)
        # k * log(p) + (n - k) * log(1 - p) = k * (log(p) - log(1 - p)) + n * log(1 - p)
        #     (case logit < 0)              = k * logit - n * log1p(e^logit)
        #     (case logit > 0)              = k * logit - n * (log(p) - log(1 - p)) + n * log(p)
        #                                   = k * logit - n * logit - n * log1p(e^-logit)
        #     (merge two cases)             = k * logit - n * max(logit, 0) - n * log1p(e^-|logit|)
        normalize_term = (self.total_count * _clamp_by_zero(self.logits)
                          + self.total_count * torch.log1p(torch.exp(-torch.abs(self.logits)))
                          - log_factorial_n)
        return value * self.logits - log_factorial_k - log_factorial_nmk - normalize_term
```

</details>

fritzo:
OK I've also used `log_beta()` in `Binomial.log_prob()`. Sorry for the duplicate reviews!

PR #2502, opened 2020-05-24
Use approximate log_beta() in .fit(), .predict()

fritzo:
Addresses #2426 
Blocking #2498 (from which this simpler PR was extracted)

This PR:

1. Fixes a nan gradient issue in `ExtendedBetaBinomial` related to https://github.com/pytorch/pytorch/issues/15506. This may have caused spurious rejections during heuristic initialization.
2. Starts using the fast `log_beta()` in `CompartmentalModel.fit()` and `.predict()` now that gradients are safe. This reduces likelihood compute cost by about 14% in `SuperspredingSEIRModel`, and will reduce #2498 more.
3. Moves `safe_log()` to `pyro.ops.special` which seems like a better home. (this change is non-breaking since `safe_log()` has not yet been released).

## Tested

- [x] added unit tests for nan gradients in `ExtendedBinomial` and `ExtendedBetaBinomial`
- [x] ran locally to verify sane behavior

### DIFF
--- a/pyro/distributions/extended.py
+++ b/pyro/distributions/extended.py
@@ -23,7 +23,7 @@ class ExtendedBinomial(Binomial):
 
     def log_prob(self, value):
         result = super().log_prob(value)
-        invalid = ~super().support.check(value)
+        invalid = (value < 0) | (value > self.total_count)
         return result.masked_fill(invalid, -math.inf)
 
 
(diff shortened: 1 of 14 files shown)


pyro/distributions/extended.py
```diff
@@ -40,6 +40,18 @@ class ExtendedBetaBinomial(BetaBinomial):
...
     def log_prob(self, value):
-        result = super().log_prob(value)
-        invalid = ~super().support.check(value)
+        if self._validate_args:
+            self._validate_sample(value)
+
+        total_count = self.total_count
+        invalid = (value < 0) | (value > total_count)
+        n = total_count.clamp(min=0)
+        k = value.masked_fill(invalid, 0)
```
fritzo, line 49: this is the crux of the nan fix

pyro/distributions/extended.py
```diff
@@ -23,7 +23,7 @@ class ExtendedBinomial(Binomial):
 
     def log_prob(self, value):
         result = super().log_prob(value)
-        invalid = ~super().support.check(value)
+        invalid = (value < 0) | (value > self.total_count)
```
martinjankowiak, line 26: curous: why this change?

fritzo: This is a little cheaper. The `integer_interval` additionally checks `value % 1 == 0`, but that is already checked in validation by the above line, and doesn't affect numerical stability.

fritzo:
Thanks for reviewing!

PR #2408, opened 2020-04-13
Create finding_MLE_MAP_simple_model.ipynb

phamtrongthang123:
Submitting tutorial on using pyro to find MLE and MAP on simple models (coin flip from https://pyro.ai/examples/svi_part_i.html#A-simple-example)

View on [github](https://github.com/phamtrongthang123/pyro/blob/patch-1/tutorial/source/finding_MLE_MAP_simple_model.ipynb) or [nbviewer](https://nbviewer.jupyter.org/github/phamtrongthang123/pyro/blob/patch-1/tutorial/source/finding_MLE_MAP_simple_model.ipynb)

PR #2507, opened 2020-05-26
add simple map/mle tutorial

martinjankowiak:
this builds off of the tutorial submitted in #2408 .

view [notebook](https://github.com/pyro-ppl/pyro/blob/mapmle/tutorial/source/mle_map.ipynb).


martinjankowiak:
@fritzo yikes nice catch!

PR #2509, opened 2020-05-27
Resurrect num_quant_bins=2 in contrib.epidemiology

fritzo:
Addresses #2426 

This resurrects the old `compute_bin_probs(s, num_quant_bins=2)`. While this works horribly for general inference, I'd like to use that helper for another spline purpose.

I've also switched from case-checking to return-early-with-fallthrough-exception to make further modifications simpler.

## Tested
- [x] added unit test cases

### DIFF
--- a/pyro/contrib/epidemiology/util.py
+++ b/pyro/contrib/epidemiology/util.py
@@ -106,7 +106,7 @@ W16 = [[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.1562511562511555e-07],
 W16 = numpy.array(W16)
 
 
-def compute_bin_probs(s, num_quant_bins=3):
+def compute_bin_probs(s, num_quant_bins):
     """
     Compute categorical probabilities for a quantization scheme with num_quant_bins many
     bins. `s` is a real-valued tensor with values in [0, 1]. Returns probabilities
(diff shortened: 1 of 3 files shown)


fritzo:
maybe this is a bad idea 🤔

fritzo:
@martinjankowiak would you mind if we resurrected this? I'd like to see how this feature interacts with other inference tricks, and it would ease my branch management if this were in dev.

pyro/contrib/epidemiology/util.py
```diff
@@ -106,16 +106,19 @@ def align_samples(samples, model, particle_dim):
...
     Compute categorical probabilities for a quantization scheme with num_quant_bins many
     bins. `s` is a real-valued tensor with values in [0, 1]. Returns probabilities
     of shape `s.shape` + `(num_quant_bins,)`
     """
-    if num_quant_bins not in [4, 8, 12, 16]:
-        raise ValueError("Supported quantization strategies have 4, 8, 12, or 16 bins")
 
     t = 1 - s
+
+    if num_quant_bins == 2:
```
martinjankowiak, line 118: why not the smoother cubic version?

fritzo: I think my/our early attempt at a 2-point cubic spline was ill-conceived and not very smooth:
image

PR #2512, opened 2020-05-30
Clean up contrib.epidemiology docs

fritzo:
Addresses #2426 

This PR:
1. Moves `CompartmentalModel.heuristic()` down in the class definition, since we no longer need custom overrides.
2. Adds example models to the docs TOC, making them more discoverable:
image

## Tested
- [x] built docs locally


fritzo:
Thanks for reviewing!  merge any time

PR #2514, opened 2020-05-31
Eliminate need for CompartmentalModel.transition_bwd()

fritzo:
Addresses #2426 

This follows ideas from #2513 to replace the `.transition_bwd()` method with a conditioned call of the `.transition_fwd()` method, significantly reducing user effort to create a new model. The additional bit of logic to reverse the flow computation is implemented in a new `.compute_flows()` method that by default assumes a linear flow terminating in "R", but can be overridden by users to perform arbitrary computations. Only one such override is needed by our current examples.

The new `.transition()` method is now called under more interpretations, so various logic needs to be generalized:
- `t < self.duration` now also needs to check for `slice` instances
- We now need to use `ExtendedBinomial` everywhere, not only in `.transition_bwd()`
- The SMC wrapper now needs to provide approximate variables like `I_approx`.

## Tested
- [x] manually ran a biggish example and verified sane posterior

### DIFF
--- a/pyro/contrib/epidemiology/compartmental.py
+++ b/pyro/contrib/epidemiology/compartmental.py
@@ -272,6 +272,8 @@ class CompartmentalModel(ABC):
             slice (for vectorized inference) or an integer time index. During
             prediction ``t`` will be integer time index.
         :type t: int or slice
+        :returns: A dict mapping flow name to tensor value.
+        :rtype: dict
         """
         flows = {}
         flow = 0
--- a/pyro/contrib/epidemiology/compartmental.py
+++ b/pyro/contrib/epidemiology/compartmental.py
@@ -182,10 +182,10 @@ class CompartmentalModel(ABC):
         Note that this method is called under multiple different
         interpretations, including batched and vectorized interpretations.
         During :meth:`generate` this is called to generate a single sample.
-        During :meth:`heuristic` this is called to generate a batch of sample
+        During :meth:`heuristic` this is called to generate a batch of samples
         for SMC.  During :meth:`fit` this is called both in vectorized form
-        (vectorizing over time) and insequential form (for a single time step);
-        both forms enumerate over discrete latent variables.  During
+        (vectorizing over time) and in sequential form (for a single time
+        step); both forms enumerate over discrete latent variables.  During
         :meth:`predict` this is called to forecast a batch of samples,
         conditioned on posterior samples for the time interval
         ``[0:duration]``.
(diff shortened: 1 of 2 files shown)


pyro/contrib/epidemiology/models.py
```diff
@@ -771,38 +623,23 @@ def transition_fwd(self, params, state, t):
...
+        mask_t = self.mask[t] if t_is_observed else False
+        data_t = self.data[t] if t_is_observed else None
         pyro.sample("obs_{}".format(t),
                     dist.Delta(state["O"]).mask(mask_t),
                     obs=data_t)
 
-    def transition_bwd(self, params, prev, curr, t):
-        R0, tau, rho = params
-
+    def compute_flows(self, prev, curr, t):
```
fritzo, line 633: This is the one model where we need a custom `.compute_flows()` method.

pyro/contrib/epidemiology/compartmental.py
```diff
@@ -172,50 +172,68 @@ def initialize(self, params):
...
 
         This inputs a current ``state`` and stochastically updates that
         state in-place.
 
-        Note that this method is called under two different interpretations.
+        Note that this method is called under multiple different
+        interpretations, including batched and vectorized interpretations.
         During :meth:`generate` this is called to generate a single sample.
-        During :meth:`predict` thsi is called to forecast a batch of samples,
+        During :meth:`heuristic` this is called to generate a batch of sample
```
martinjankowiak, line 185: typo: samples

pyro/contrib/epidemiology/compartmental.py
```diff
@@ -172,50 +172,68 @@ def initialize(self, params):
...
+        for SMC.  During :meth:`fit` this is called both in vectorized form
+        (vectorizing over time) and insequential form (for a single time step);
```
martinjankowiak, line 187: typo: insequential

pyro/contrib/epidemiology/models.py
```diff
@@ -717,6 +564,11 @@ class SparseSIRModel(CompartmentalModel):
     (when ``mask[t] == True``) ``O`` must exactly match the provided data;
     between observed times ``O`` stochastically imputes the provided data.
 
+    This model demonstrates how to implement a custom :meth:`compute_flows`
+    method. A custom method is needed in this model because inhabitants of the
+    ``S`` compartment can transition to both the ``I`` and ``o`` compartments,
```
martinjankowiak, line 569: typo: o versus O

PR #2516, opened 2020-06-02
Expose ImproperUniform distribution

fritzo:
Addresses #2426 
Follows up on @fehiepsi's comment https://github.com/pyro-ppl/pyro/pull/2495#pullrequestreview-415691686

This exposes `ImproperUniform` as a public distribution. It is currently used only in `SplitReparam`. I plan to use it also for sampling auxiliary variables in `CompartmentalModel._sequential_model()` and `._vectorized_model()`. Those classes currently use `dist.Uniform(...).mask(False).expand(shape).to_event()`, but I would like to generalize the logic to arbitrary constraints, while adding the safety of a not-implemented `.sample()` method.

## Tested
- [x] refactoring is covered by existing `SplitReparam` tests
- [x] added some unit tests

### DIFF
--- a/pyro/distributions/improper_uniform.py
+++ b/pyro/distributions/improper_uniform.py
@@ -37,12 +37,12 @@ class ImproperUniform(TorchDistribution):
 
     :param support: The support of the distribution.
     :type support: ~torch.distributions.constraints.Constraint
-    :param torch.Size batch_shape: Defaults to empty.
-    :param torch.Size event_shape: Defaults to empty.
+    :param torch.Size batch_shape: The batch shape.
+    :param torch.Size event_shape: The event shape.
     """
     arg_constraints = {}
 
-    def __init__(self, support, batch_shape=torch.Size(), event_shape=torch.Size()):
+    def __init__(self, support, batch_shape, event_shape):
         assert isinstance(support, constraints.Constraint)
         self._support = support
         super().__init__(batch_shape, event_shape)
--- a/pyro/distributions/improper_uniform.py
+++ b/pyro/distributions/improper_uniform.py
@@ -1,0 +1,67 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+from torch.distributions import constraints
+
+from .torch_distribution import TorchDistribution
+from .util import broadcast_shape
+
+
+class ImproperUniform(TorchDistribution):
+    """
+    Improper distribution with zero :meth:`log_prob` and undefined
+    :meth:`sample`.
+
+    This is useful for transforming a model from generative dag form to factor
+    graph form for use in HMC. For example the following are equal in
+    distribution::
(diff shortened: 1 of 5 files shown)


pyro/distributions/improper_uniform.py
```diff
@@ -0,0 +1,67 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+from torch.distributions import constraints
+
+from .torch_distribution import TorchDistribution
+from .util import broadcast_shape
+
+
+class ImproperUniform(TorchDistribution):
```
fehiepsi, line 11: Just to make sure that you already consider it: should it be better to use the shorter name `Improper`?

pyro/distributions/improper_uniform.py
```diff
@@ -0,0 +1,67 @@
...
+        xyz = dist.Normal(0, 1).expand([3]).to_event(1).mask(False)
+
+    :param support: The support of the distribution.
+    :type support: ~torch.distributions.constraints.Constraint
+    :param torch.Size batch_shape: Defaults to empty.
+    :param torch.Size event_shape: Defaults to empty.
+    """
+    arg_constraints = {}
+
+    def __init__(self, support, batch_shape=torch.Size(), event_shape=torch.Size()):
```
fehiepsi, line 45: I think `event_shape` is a requirement. Otherwise, we can set it to `None`, so that the default event_shape can be `(0,) * biject_to(support).event_dim`. This way, log_prob will have a more appropriate shape.

pyro/distributions/improper_uniform.py
fritzo: Hmm... I guess there are other improper distributions, e.g. `Unit`.

pyro/distributions/improper_uniform.py
fritzo: I think i'd rather make them both required than perform a computation on the constraint. Currently our constraints are pretty sloppy- we should be using `IndependentConstraint` more thoroughly and many of the computations would be wrong. This isn't a big issue yet because we don't use `biject_to(support).event_dim` for much.

fehiepsi: Yeah, I agree that we should require event_shape. But for batch_shape, I guess it is better to let it have default shape as before (i.e. `__init__(self, support, event_shape, batch_shape=torch.Size())`) to makes the code less verbose.

fritzo: I agree a default of `batch_shape=()` makes sense, but I find it disorienting for batch_shape to be specified on the right of event_shape: I always see things as `batch_shape, event_shape`. 🤷

fehiepsi: Haha, sounds reasonable to me.

fritzo:
Thanks for your careful review @fehiepsi!  I'm glad you care about interface elegance :smile:

PR #2517, opened 2020-06-03
Add a time-heterogeneous SIR model

fritzo:
Addresses #2426 

This adds an SIR model with time-varying response rate `rho` and reproductive number `Re`. The response rate is assumed to be piecewise constant, say depending on known testing capacity. The reproductive number undergoes Brownian motion in log space. The motivation is to handle real-world datasets where policy changes can effect reproductive number over time. Existing homogeneous models are a very poor fit.

Mechanically this introduces a new autovectorization strategy, which we might eventually generalize into a strategy for `poutine.markov` contexts. The strategy is to sample a time-batch of `curr` values as an `dist.IndependentUniform(...)`, then truncate and pad to get a time-batch of `prev` values, and finally replay the model with a time-batch of (`prev`,`curr`) pairs (this only works for Markov window of 1; longer windows would require more states). I.e. this implements the common inference motif of converting a sequential directed Markov chain model to a batched factor graph model.

The new strategy supports a wide range of time-varying variables, both time-independent and Markov with window size 1. I see this as the last major feature of `contrib.epidemiology` before release.

## Tested
- [x] smoke tests
- [x] added to examples/.../sir.py

### DIFF
--- a/pyro/contrib/epidemiology/models.py
+++ b/pyro/contrib/epidemiology/models.py
@@ -1,6 +1,8 @@
 # Copyright Contributors to the Pyro project.
 # SPDX-License-Identifier: Apache-2.0
 
+import re
+
 import torch
 from torch.nn.functional import pad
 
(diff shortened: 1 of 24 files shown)


pyro/contrib/epidemiology/models.py
```diff
@@ -551,6 +551,99 @@ def transition(self, params, state, t):
...
 
+class HeterogeneousSIRModel(CompartmentalModel):
+    """
+    Generalizes :class:`SimpleSIRModel` by allowing ``Re`` and ``rho`` to vary
+    in time.
+
+    To customize this model we recommend forking and editing this class.
+
+    In this model, the response rate ``rho`` is piecewise constant with unknown
+    value over three pieces. The reproductive number ``Re`` is a the product of
```
martinjankowiak, line 562: typo "a the"

fritzo: fixed

PR #2521, opened 2020-06-08
Eliminate need for CompartmentalModel class variables

fritzo:
Addresses #2426 

This eliminates the need for classes deriving from `CompartmentalModel` to define class variables `.full_mass` and `.series`, thus easing the task of creating new models. After this PR, those class variables are defined automatically via `@lazy_property`.

This also fixes an issue with sphinx 3.1 that breaks docs building of `@lazy_property` (this fix is blocking all Pyro development).

## Tested
- [x] added assertions checking `._full_mass` computation
- [x] ran `examples/.../sir.py --plot` to verify all series are created and plotted

### DIFF
--- a/pyro/distributions/torch_patch.py
+++ b/pyro/distributions/torch_patch.py
@@ -55,4 +55,10 @@ def _Multinomial_support(self):
     return torch.distributions.constraints.integer_interval(0, total_count)
 
 
+# This adds a __call__ method to satisfy sphinx.
+@patch_dependency('torch.distributions.utils.lazy_property.__call__')
+def _lazy_property__call__(self):
+    raise NotImplementedError
+
+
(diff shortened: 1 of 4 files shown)


pyro/distributions/torch_patch.py
```diff
@@ -55,4 +55,10 @@ def _Multinomial_support(self):
     return torch.distributions.constraints.integer_interval(0, total_count)
 
 
+# This adds a __call__ method to satisfy sphinx.
+@patch_dependency('torch.distributions.utils.lazy_property.__call__')
+def _lazy_property__call__(self):
+    raise NotImplementedError
```
fritzo, line 61: This is required for any Pyro development after the recent sphinx 3.1 release.

PR #2518, opened 2020-06-05
Add epidemiology tutorial with a regional SEIR model

fritzo:
Addresses #2426 
View [notebook](https://nbviewer.jupyter.org/github/pyro-ppl/pyro/blob/sir-tutorial-ii/tutorial/source/epi_regional.ipynb)

## Tasks
- [x] Fix initialization issues
- [ ] Figure out how to run on a GPU machine

PR #2523, opened 2020-06-13
Support HaarReparam for non-compartmental variables

fritzo:
Addresses #2426 

This follows up #2517 to support `HaarReparam` and `SplitReparam` for non-compartmental latent time series. After this PR `haar_full_mass` will trigger learning of covariance among the low frequency components of *all* time series, hopefully improving mixing in long-duration datasets.

**Summary:**
1. Factors out `haar` and `haar_full_mass` logic into a `_HaarSplitReparam` class.
2. Factors out sampling of all auxiliary variables into a `._sample_auxiliary()` method.
3. Changes non-compartmental variables to avoid plates, thereby permitting reparameterization along the time dimension (we can't reparameterize along a plate dimension; this was the crux).
4. Adds a new `HeterogeneousRegionalSIRModel` (ported from #2518) to exercise new code.
5. Cleans up model imports and docs, since this was getting unweildy.
6. Renames `Re` to `Rt` which is more standard.
7. Bumps pytest version to fix recent `pytest_cov` breakage.

From @martinjankowiak's comment in https://github.com/pyro-ppl/pyro/pull/2517#pullrequestreview-424883012:

> it really is getting quite complicated though so it's probably good that the current phase of feature addition is coming to an end.

I hope the refactoring in this PR helps to reduce complexity.

## Tested
- [x] refactoring exercised by existing tests
- [x] added a new `HeterogeneousRegionalSIRModel` with test
- [x] verified improved mixing in #2518

### DIFF
--- a/pyro/contrib/epidemiology/__init__.py
+++ b/pyro/contrib/epidemiology/__init__.py
@@ -3,22 +3,9 @@
 
 from .compartmental import CompartmentalModel
 from .distributions import beta_binomial_dist, binomial_dist, infection_dist
-from .models import (HeterogeneousSIRModel, OverdispersedSEIRModel, OverdispersedSIRModel, RegionalSIRModel,
-                     SimpleSEIRModel, SimpleSIRModel, SparseSIRModel, SuperspreadingSEIRModel, SuperspreadingSIRModel,
-                     UnknownStartSIRModel)
 
 __all__ = [
     "CompartmentalModel",
-    "HeterogeneousSIRModel",
-    "OverdispersedSEIRModel",
-    "OverdispersedSIRModel",
-    "RegionalSIRModel",
-    "SimpleSEIRModel",
-    "SimpleSIRModel",
-    "SparseSIRModel",
-    "SuperspreadingSEIRModel",
-    "SuperspreadingSIRModel",
-    "UnknownStartSIRModel",
     "beta_binomial_dist",
     "binomial_dist",
     "infection_dist",
--- a/pyro/contrib/epidemiology/compartmental.py
+++ b/pyro/contrib/epidemiology/compartmental.py
@@ -400,9 +400,7 @@ class CompartmentalModel(ABC):
         mcmc.run()
         self.samples = mcmc.get_samples()
         if haar:
-            print(f"DEBUG {list(self.samples.keys())}")
             haar.aux_to_user(self.samples)
-            print(f"DEBUG {list(self.samples.keys())}")
 
         # Unsqueeze samples to align particle dim for use in poutine.condition.
         # TODO refactor to an align_samples or particle_dim kwarg to MCMC.get_samples().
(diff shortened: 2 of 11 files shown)


pyro/contrib/epidemiology/compartmental.py
```diff
@@ -868,3 +849,77 @@ def init(self, state):
     def step(self, state):
         with poutine.block(hide_types=["observe"]):
             super().step(state.copy())
+
+
+class _HaarSplitReparam:
```
fritzo, line 854: @fehiepsi I found this class provides useful helpers user <--> aux, in addition to our usual `.reparam()` functionality. These extra helpers seem necessary for `poutine.reparam()` to play well with `init_to_value()`. That is, if we reparameterize a model, then a user's custom `init_to_value()` function will need to be converted to different coordinates. WDYT of generalizing these to be part of the `Reparam` interface, say as a pair of methods (`.aux_to_user()`, `.user_to_aux()`) or similar? (in follow-up PRs to Pyro and NumPyro).

pyro/contrib/epidemiology/__init__.py
```diff
@@ -3,22 +3,9 @@
...
-    "HeterogeneousSIRModel",
-    "OverdispersedSEIRModel",
-    "OverdispersedSIRModel",
-    "RegionalSIRModel",
-    "SimpleSEIRModel",
-    "SimpleSIRModel",
-    "SparseSIRModel",
-    "SuperspreadingSEIRModel",
-    "SuperspreadingSIRModel",
-    "UnknownStartSIRModel",
```
fritzo, line 21: Removing these because they are inteded more as examples than as reusable components.

pyro/contrib/epidemiology/compartmental.py
fehiepsi: > These extra helpers seem necessary for poutine.reparam() to play well with init_to_value().

I see. I agree that those helper methods would simplify much of the user code for this init strategy.

pyro/contrib/epidemiology/compartmental.py
```diff
@@ -777,28 +771,15 @@ def _relaxed_model(self):
...
         params = self.global_model()
+        auxiliary, non_compartmental = self._sample_auxiliary()
 
-        # Sample the compartmental continuous reparameterizing variable.
-        shape = (C, T) + R_shape
-        auxiliary = pyro.sample("auxiliary",
-                                dist.Uniform(-0.5, self.population + 0.5)
-                                    .mask(False).expand(shape).to_event())
-        assert auxiliary.shape == shape, "particle plates are not supported"
+        # Split tenors into current state.
```
martinjankowiak, line 780: typo

pyro/contrib/epidemiology/compartmental.py
```diff
@@ -868,3 +849,77 @@ def init(self, state):
...
+        return True
+
+    def reparam(self, model):
+        """
+        Wrap a model with ``poutine.reparam``.
+        """
+        # Transform to Haar coordinates.
+        config = {}
+        for name, dim in self.dims.items():
+            config[name] = HaarReparam(dim=dim, flip=True)
```
martinjankowiak, line 876: make `flip` an arg?

fritzo: I think that's unnecessary since we don't expose any plumbing to set that, and since `_HaarSplitReparam` is merely a minimal internal helper.

pyro/contrib/epidemiology/compartmental.py
fritzo: done

fritzo:
Thanks for reviewing!

PR #2524, opened 2020-06-13
Monotonic rational quadratic spline

stefanwebb:
Previously, `Spline` and `ConditionalSpline` could only use monotonic rational *linear* splines. I've added the functionality to use quadratic splines, as per the Neural Spline Flow paper


stefanwebb:
Hmm, there is some problem with Travis and the version of PyTest...

fritzo:
@stefanwebb I've fixed the pytest problems in #2523. You could wait until that merges or pull in the changes (setup.py and tests/integration_tests/test_conjugate_gaussian_models.py).

tests/distributions/test_transforms.py
```diff
@@ -196,7 +196,8 @@ def test_conditional_radial(self):
         self._test_conditional(T.conditional_radial, inverse=False)
 
     def test_conditional_spline(self):
-        self._test_conditional(T.conditional_spline)
+        for order in ['linear', 'quadratic']:
+            self._test_conditional(partial(T.conditional_spline, order=order))
```
fritzo, line 200: nit: Does pytest notation work here?
```py
@pytest.mark.parametrize("order", ["linear", "quadratic"])
def test_conditional_spline(self, order):
    self._test_conditional(partial(T.conditional_spline, order=order))
```
That generally gives clearer test reports.

tests/distributions/test_transforms.py
```diff
@@ -261,7 +262,8 @@ def test_radial(self):
         self._test(T.radial, inverse=False)
 
     def test_spline(self):
-        self._test(T.spline)
+        for order in ['linear', 'quadratic']:
+            self._test(partial(T.spline, order=order))
```
fritzo, line 266: ditto

fritzo:
You should be able to merge with dev now, to fix travis tests.

tests/distributions/test_transforms.py
stefanwebb: It appears you can't do this on class methods: https://stackoverflow.com/questions/18182251/does-pytest-parametrized-test-work-with-unittest-class-based-tests?noredirect=1&lq=1

A workaround would involve requiring an extra dependency... It is better to keep as is?

tests/distributions/test_transforms.py
stefanwebb: ditto

tests/distributions/test_transforms.py
fritzo: fine to keep as is.

PR #2526, opened 2020-06-14
Support jit in CompartmentalModel

fritzo:
Addresses #2426 

This supports `jit_compile=True` in `CompartmentalModel.fit()`. Changes include:
- plumbing arguments through interfaces and scripts
- replacing `Tensor.__iter__()` with `Tensor.unbind(0)` to issue the correct warning type, so it can be ignored by `ignore_jit_warnings=True`.

## Performance

Anecdotally about 1.6x speedup (with identical output):
| nojit | jit | |
|-|-|:-|
| 44 sec | 27 sec | `python examples/contrib/epidemiology/sir.py` |
| 89 sec | 56 sec | `python examples/contrib/epidemiology/regional.py -hfm=7` |

## Tested
- [x] added unit tests
- [x] added to `test_examples.py::test_jit`


examples/contrib/epidemiology/regional.py
```diff
@@ -149,6 +150,7 @@ def main(args):
     parser.add_argument("--single", action="store_false", dest="double")
     parser.add_argument("--rng-seed", default=0, type=int)
     parser.add_argument("--cuda", action="store_true")
+    parser.add_argument("--jit", action="store_true")
```
martinjankowiak, line 153: shouldn't this be on by default?

fritzo: Good question. How about we enable it on by default in the example files, but off by default in the `CompartmentalModel` class? I just worry that `jit_compile=True` disables validation, and that if we optimize for new users we should enable validation by default. The example scripts use predefined models and don't really need validation, whereas `CompartmentalModel` will be the base class for new models.

martinjankowiak: yeah that sounds like a reasonable middle ground

fritzo: done.

PR #2527, opened 2020-06-15
Misc fixes to contrib.epidemiology

fritzo:
Addresses #2426 

This implements three fixes in preparation for a tutorial notebook:
1. Support parallel chain MCMC. This requires pickle support, so I've moved the local function `heuristic()` to a method `._heuristic()`.
2. Fixes a shape error (and resolves a TODO) in `._concat_series()`.
3. Refactors relaxed logic to refer to a global `_RELAX_MIN_VARIANCE` variable, and lowers the value from 0.25 to 0.1, which seems to improve inference. (I may change again in the future, but future changes will be one-line changes rather than 6-line changes across 2 files).

## Tested
- [x] unit tests for `num_chains=1`
- [x] verified shape and numerical changes in a notebook

### DIFF
--- a/examples/contrib/epidemiology/sir.py
+++ b/examples/contrib/epidemiology/sir.py
@@ -91,6 +91,7 @@ def infer(args, model):
                      heuristic_ess_threshold=args.ess_threshold,
                      warmup_steps=args.warmup_steps,
                      num_samples=args.num_samples,
+                     num_chains=args.num_chains,
                      max_tree_depth=args.max_tree_depth,
                      arrowhead_mass=args.arrowhead_mass,
                      num_quant_bins=args.num_bins,
@@ -293,6 +294,7 @@ if __name__ == "__main__":
     parser.add_argument("-np", "--num-particles", default=1024, type=int)
     parser.add_argument("-ess", "--ess-threshold", default=0.5, type=float)
     parser.add_argument("-w", "--warmup-steps", type=int)
+    parser.add_argument("-c", "--num-chains", default=1, type=int)
     parser.add_argument("-t", "--max-tree-depth", default=5, type=int)
     parser.add_argument("-a", "--arrowhead-mass", action="store_true")
     parser.add_argument("-r", "--rng-seed", default=0, type=int)
--- a/examples/contrib/epidemiology/sir.py
+++ b/examples/contrib/epidemiology/sir.py
@@ -79,6 +79,7 @@ def generate_data(args):
 
 
 def infer(args, model):
+    parallel = args.num_chains > 1
     energies = []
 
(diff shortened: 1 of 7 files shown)


martinjankowiak:
> Refactors relaxed logic to refer to a global _RELAX_MIN_VARIANCE variable, and lowers the value from 0.25 to 0.1, which seems to improve inference. 

@fritzo any hypothesis as to why?

fritzo:
> > lowers `_RELAX_MIN_VARIANCE` from 0.25 to 0.1, which seems to improve inference.

> @fritzo any hypothesis as to why?

Yes, I believe there is a bias-variance tradeoff where very small `_RELAX_MIN_VARIANCE` leads to poor mixing because of infeasibility, but large `_RELAX_MIN_VARIANCE` leads to spurious infections due to leakage at zero. Anecdotally, the value 0.1 seems balance the tradeoff, but I have not yet done a thorough comparison.

pyro/infer/mcmc/api.py
```diff
@@ -408,8 +408,12 @@ def model(data):
         # If transforms is not explicitly provided, infer automatically using
         # model args, kwargs.
         if self.transforms is None:
+            # Try to initialize kernel.transforms using kernel.setup().
+            if getattr(self.kernel, "transforms", None) is None:
+                warmup_steps = 0
+                self.kernel.setup(warmup_steps, *args, **kwargs)
             # Use `kernel.transforms` when available
-            if hasattr(self.kernel, 'transforms') and self.kernel.transforms is not None:
+            if getattr(self.kernel, "transforms", None) is not None:
```
fritzo, line 416: This is required because manually calling `initialize_model()` below would ignore `init_strategy` and attempt to sample from `ImproperUniform`. Instead we let `kernel.setup()` call `initialize_model()` with proper arguments.

examples/contrib/epidemiology/sir.py
```diff
@@ -92,16 +93,17 @@ def hook_fn(kernel, *unused):
                      warmup_steps=args.warmup_steps,
                      num_samples=args.num_samples,
                      num_chains=args.num_chains,
+                     mp_context="spawn" if parallel else None,
```
martinjankowiak, line 96: unfortunately i've found this to be somewhat buggy (fair number of crashes)

fritzo: do you use "forkserver" instead?

@neerajprad do you have suggestions?

martinjankowiak: no i've used spawn. but i've tried to avoid it...

fritzo:
Phew 😅 thanks for reviewing!

PR #2540, opened 2020-06-27
SplineAutoregressive

stefanwebb:
SplineAutoregressive transform from Neural Spline Flow


fritzo:
@stefanwebb LGTM, but looks like you'll need to resolve merge conflicts

stefanwebb:
Okay, should be good to go now!

Issue #2361, opened 2020-03-10
RenyiELBO no longer compatible with sequential plates? [question] [documentation]


sharrison5:
### Issue Description
In Pyro 1.3.0, the behaviour of `RenyiELBO` changed. In particular, it now seems to be incompatible with sequential plates. See MWE below.

Questions:
 + Is there a way to keep using sequential plates? The plate I'm using can't naturally be replaced by a vectorised version (unlike in the example).
 + If a model/guide contains plates, is it safe to use `RenyiELBO` in Pyro < 1.3.0 (i.e. what was supported before #2321)?

Many thanks! :smile:

### Code Snippet
```py
import sys, torch, pyro
print(sys.version, torch.__version__, pyro.__version__)

def model(data):
    x = pyro.sample('x', pyro.distributions.Bernoulli(torch.tensor(0.5)))
    for i in pyro.plate('data_plate', len(data)):
        pyro.sample('data_{:d}'.format(i), pyro.distributions.Normal(x, scale=torch.tensor(0.1)), obs=data[i])

# Works as expected
# def model(data):
#     x = pyro.sample('x', pyro.distributions.Bernoulli(torch.tensor(0.5)))
#     with pyro.plate('data_plate', len(data)):
#         pyro.sample('data', pyro.distributions.Normal(x, scale=torch.tensor(0.1)), obs=data)

def guide(data):
    p = pyro.param('p', torch.tensor(0.5))
    x = pyro.sample('x', pyro.distributions.Bernoulli(p))

elbo = pyro.infer.TraceGraph_ELBO()
print(elbo.loss(model, guide, torch.randn([3])))

elbo = pyro.infer.RenyiELBO()
print(elbo.loss(model, guide, torch.randn([3])))
```

Output:
```
3.8.1 (default, Jan  8 2020, 16:15:59) 
[Clang 4.0.1 (tags/RELEASE_401/final)] 1.4.0 1.3.0
2.4235687255859375
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-6865281f40fd> in <module>
     15 
     16 elbo = pyro.infer.RenyiELBO()
---> 17 print(elbo.loss(model, guide, torch.randn([3])))

~/miniconda3/envs/pDCM-dev/lib/python3.8/site-packages/torch/autograd/grad_mode.py in decorate_no_grad(*args, **kwargs)
     47         def decorate_no_grad(*args, **kwargs):
     48             with self:
[...]

PR #2541, opened 2020-06-27
fix summation with plates when manually unrolling

iffsid:
Potential fix for manual unrolling of plate

issue: #2361

### DIFF
--- a/tests/infer/test_inference.py
+++ b/tests/infer/test_inference.py
@@ -767,11 +767,7 @@ def test_sequential_plating_sum():
     loss_fn = RenyiELBO(alpha=0, num_particles=30, vectorize_particles=True)
     svi = SVI(model, guide, adam, loss_fn)
 
-    for step in range(1001):
+    for step in range(1):
         loss = svi.step(data)
         if step % 20 == 0:
             logger.info("step {} loss = {:0.4g}".format(step, loss))
-
-    expected_p = torch.tensor(0.5)
-    actual_p = pyro.param("p").detach()
-    assert_close(actual_p, expected_p, atol=0.05)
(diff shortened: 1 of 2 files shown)


fehiepsi:
Thanks, @iffsid! Could you also add the failing example in #2361 to [this](https://github.com/pyro-ppl/pyro/blob/dev/tests/infer/test_inference.py#L747) so we can make sure that future changes do not have regression?

tests/infer/test_inference.py
```diff
@@ -745,3 +745,33 @@ def model():
...
+    svi = SVI(model, guide, adam, loss_fn)
+
+    for step in range(1001):
+        loss = svi.step(data)
+        if step % 20 == 0:
+            logger.info("step {} loss = {:0.4g}".format(step, loss))
+
+    expected_p = torch.tensor(0.5)
+    actual_p = pyro.param("p").detach()
+    assert_close(actual_p, expected_p, atol=0.05)
```
fehiepsi, line 777: It is surprising to me that without enumeration, SVI works for this pair of `model`/`guide`. Bernoulli is a discrete distribution, so grad can't pass through its samples. What happens if you change the initial parameter `p` to something different from `0.5`, for example `p = pyro.param('p', torch.tensor(0.2))`?

If SVI fails, I would suggest to only run 1 step `svi.step(data)` to verify `RenyiELBO` can run with a sequential plate. Alternatively, I guess you can change `Bernoulli` to some other continuous distribution, for example `dist.Normal(0, 1)` in `model` and `dist.Normal(p, 0.1)` in the guide, to make the inference work.

iffsid: Ah, I didn't actually do any of those checks---just put in a dummy check as we didn't care about the answer in this case.
I thought the use of `score_parts` within `RenyiElbo` handles the discrete Bernoulli?

In any case, I'll just run this for one step as we just need to be sure it 'runs'. :)

fehiepsi: You are right! `score_parts` might remedy that issue. Thanks for reminding me about it!

PR #2544, opened 2020-06-28
Test for whether `TransformModule`s work for density estimation

stefanwebb:
I've added a test for whether TransformModule's work for density estimation after discovering that `Spline` doesn't... I'll fix the bug in `Spline` in this PR

### DIFF
--- a/pyro/distributions/transforms/householder.py
+++ b/pyro/distributions/transforms/householder.py
@@ -3,6 +3,7 @@
 
 import math
 import warnings
+from functools import partial
 
 import torch
 import torch.nn as nn
@@ -12,8 +13,6 @@ from pyro.distributions.conditional import ConditionalTransformModule
 from pyro.distributions.torch_transform import TransformModule
 from pyro.distributions.util import copy_docs_from
 from pyro.nn import DenseNN
-
-from functools import partial
 
 
 class ConditionedHouseholder(Transform):
--- a/pyro/distributions/transforms/householder.py
+++ b/pyro/distributions/transforms/householder.py
@@ -13,6 +13,8 @@ from pyro.distributions.torch_transform import TransformModule
 from pyro.distributions.util import copy_docs_from
 from pyro.nn import DenseNN
 
+from functools import partial
+
 
 class ConditionedHouseholder(Transform):
     domain = constraints.real
@@ -21,15 +23,13 @@ class ConditionedHouseholder(Transform):
     event_dim = 1
     volume_preserving = True
 
-    def __init__(self):
+    def __init__(self, u_unnormed=None):
         super().__init__(cache_size=1)
-
-    def _params(self):
-        raise NotImplementedError()
+        self.u_unnormed = u_unnormed
 
     # Construct normalized vectors for Householder transform
     def u(self):
-        u_unnormed = self._params()
+        u_unnormed = self.u_unnormed() if callable(self.u_unnormed) else self.u_unnormed
         norm = torch.norm(u_unnormed, p=2, dim=-1, keepdim=True)
         return torch.div(u_unnormed, norm)
 
@@ -145,9 +145,6 @@ over-parametrization!".format(count_transforms, input_dim))
         stdv = 1. / math.sqrt(self.u_unnormed.size(-1))
         self.u_unnormed.data.uniform_(-stdv, stdv)
 
-    def _params(self):
-        return self.u_unnormed
-
 
 @copy_docs_from(ConditionalTransformModule)
 class ConditionalHouseholder(ConditionalTransformModule):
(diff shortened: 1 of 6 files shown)


stefanwebb:
I worked out that the bug was that several transforms didn't update their parameters for each call of the forward or inverse operation. This is a problem after you've taken a gradient step during learning...

We now test that density estimation is possible and it passes for all transforms!

stefanwebb:
@fritzo when this merges it will unblock #2542

pyro/distributions/transforms/householder.py
```diff
@@ -21,14 +21,17 @@ class ConditionedHouseholder(Transform):
     event_dim = 1
     volume_preserving = True
 
-    def __init__(self, u_unnormed=None):
+    def __init__(self):
         super().__init__(cache_size=1)
-        self.u_unnormed = u_unnormed
+
+    def _params(self):
+        raise NotImplementedError()
```
fritzo, line 28: This seems a bit awkward, creating an object then setting a private member after construction. What if instead we allow tensor-or-callable values in the constructor, similar to `pyro.param`?
```py
class ConditionedHousholder(Transform):
    def __init__(self, u_unnormed):
        super().__init__(cache_size=1)
        assert isinstance(u_unnormed, torch.Tensor) or callable(u_unnormed)
        self.u_unnormed = u_unnormed

    def u(self):
        u_unnormed = self.u_unnormed
        if callable(u_unnormed):
            u_unnormed = u_unnormed()
        ...
```
and correspondingly updated `ConditionalHouseholder`
```py
class ConditionalHouseholder(...):
    def _u_unnormed(self, context):
        # u_unnormed ~ (count_transforms, input_dim)
        # Hence, input_dim must divide
        u_unnormed = self.nn(context)
        if self.count_transforms == 1:
            u_unnormed = u_unnormed.unsqueeze(-2)
        else:
            u_unnormed = torch.stack(u_unnormed, dim=-2)
        return u_unnormed
        
    def condition(self, context):
        u_unnormed = functools.partial(self._u_unnormed, context)
        return ConditionedHouseholder(u_unnormed)
```

pyro/distributions/transforms/householder.py
```diff
@@ -218,14 +224,18 @@ def __init__(self, input_dim, nn, count_transforms=1):
...
-        if self.count_transforms == 1:
-            u_unnormed = u_unnormed.unsqueeze(-2)
-        else:
-            u_unnormed = torch.stack(u_unnormed, dim=-2)
-        return ConditionedHouseholder(u_unnormed)
+        def params():
+            # u_unnormed ~ (count_transforms, input_dim)
+            # Hence, input_dim must divide
+            u_unnormed = self.nn(context)
+            if self.count_transforms == 1:
```
fritzo, line 231: Not part of this PR, but it is an antipattern to make return type depend on a scalar value. Doing so breaks the ability to write generic code using this interface. Has this behavior been released yet?

pyro/distributions/transforms/householder.py
stefanwebb: Ah, that's a much better way of doing that! I'll make these changes :)

pyro/distributions/transforms/householder.py
stefanwebb: Yes I'm afraid it's been released... Can you elaborate on why this would break the ability to write generic code? I think the underlying problem is actually how `DenseNN` shapes it's outputs, and this is something we should think about more carefully during a major refactoring

fritzo: **warning:** the following is highly subjective

An example of some generic code you might like to write is, say, writing a `ConditionalCatTransform` that concatenates two conditional transforms. You might want to fuse their neural nets to save on tensor ops and share strength. How would you do this? Maybe
```py
class ConditionalCatTransform(...):
    def __init__(self, parts):
        super().__init__(self, cache_size=1)
        self.parts = parts  # ignoring ModuleList magic for demo purposes
        self.nn = memoize(cat_dense_nn([part.nn for part in self.parts]))
        end = 0
        for part in self.parts:
            beg, end = end, end + part.count_transforms
            # The following line assumes DenseNN returns a tuple:
            part.nn = lambda context: self.nn(context)[beg:end]
```
then we could define helpers `memoize = functools.lru_cache(max_size=1)` and
```py
def cat_dense_nn(parts):
    input_dims = parts[0].input_dims
    hidden_dims = parts[0].hidden_dims
    param_dims = sum([part.param_dims for part in parts], [])
    return DenseNN(input_dims, hidden_dims, param_dims) 
```
That's what we could have written if the types were consistent. It would have been pretty simple generic code. But it looks like `DenseNN` returns an output type that depends on an int value, so we would need to complicate our wrapper with extra logic:
```diff
  class ConditionalCatTransform(...):
      def __init__(self, parts):
          ...
          for part in self.parts:
              beg, end = end, end + part.count_transforms
-             part.nn = lambda context: self.nn(context)[beg:end]
+             if beg + 1 == end:
+                 part.nn = lamba context: self.nn(context)[beg]
+             else:
+                 part.nn = lambda context: self.nn(context)[beg:end]
```
Now that's just a little more complex. But I feel, as an author of complex code, that the complexity tax is best paid by one-off code, so that abstractions can be built tax free. I think this tax structure is one of the main differences between programming languages, e.g. R and MATLAB tend to tax library code, whereas Python and C tend to tax one-off code.

pyro/distributions/transforms/householder.py
stefanwebb: Done!

pyro/distributions/transforms/householder.py
stefanwebb: Okay, let's think about this when we have a chance to overhaul `DenseNN`

stefanwebb:
@fritzo I'm not sure what's wrong with the AIR tutorial in Travis...

fritzo:
@neerajprad Do you have any ideas why the air tutorial might be failing? I haven't been able to reproduce locally.

neerajprad:
I think https://github.com/pyro-ppl/pyro/pull/2549 should temporarily silence the FutureWarning from jupyter_client that's causing the tutorial tests to fail.

fritzo:
@stefanwebb can you try merging in dev and pushing to trigger ci?

stefanwebb:
Thanks @neerajprad! It all passes now 😄

PR #2551, opened 2020-07-07
Dense MatrixExponential Transform

stefanwebb:
This is an implementation of the `MatrixExponential` transform for a dense linear operator, y = exp(M)x. The transform was introduced in this recent paper: [The Convolution Exponential and Generalized Sylvester Flows](https://arxiv.org/abs/2006.01910).

The strength of the method is that it allows you to take any linear operation and make it into a bijection with an explicit inverse and tractable log-determinate-Jacobian. I've implemented a dense linear `MatrixExponential` transform here, although the method really shines for structured M on images and graphs (I leave `ConvMatrixExponential` and `GraphMatrixExponential` as well as `GeneralizedSylvester` for future PRs)

### DIFF
--- a/pyro/distributions/transforms/matrix_exponential.py
+++ b/pyro/distributions/transforms/matrix_exponential.py
@@ -1,0 +1,147 @@
+# Copyright (c) 2017-2019 Uber Technologies, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+import math
+
+import torch
+import torch.nn as nn
+from torch.distributions import Transform, constraints
+
+from pyro.distributions.torch_transform import TransformModule
+from pyro.distributions.util import copy_docs_from
+
+
+@copy_docs_from(Transform)
+class ConditionedMatrixExponential(Transform):
+    domain = constraints.real
+    codomain = constraints.real
+    bijective = True
+    event_dim = 1
+
+    def __init__(self, weights=None, iterations=8, spectral_norm=0.5):
+        super().__init__(cache_size=1)
+        assert iterations > 0
+        self.weights = weights
+        self.iterations = iterations
+        self.spectral_norm = spectral_norm
+
+    def _exp(self, x, M):
+        """
+        Performs power series approximation to the vector product of x with the
+        matrix exponential of M.
+        """
+        power_term = x.unsqueeze(-1)
+        y = x.unsqueeze(-1)
+        for idx in range(self.iterations):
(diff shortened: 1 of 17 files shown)


stefanwebb:
@fritzo could we please merge in #2544 before this one?

pyro/distributions/transforms/matrix_exponential.py
```diff
@@ -0,0 +1,174 @@
...
+        y = x.unsqueeze(-1)
+        for idx in range(self.iterations):
+            power_term = torch.matmul(M, power_term) / (idx + 1)
+            y = y + power_term
+
+        return y.squeeze(-1)
+
+    def _trace(self, M):
+        """
+        Calculate the trace of a matrix and is able to do broadcasting over batch
```
fritzo, line 51: nit: grammar. Calculate -> Calculates

pyro/distributions/transforms/matrix_exponential.py
```diff
@@ -0,0 +1,174 @@
...
+    [2] Tim Salimans, Diederik P. Kingma. Weight Normalization: A Simple
+        Reparameterization to Accelerate Training of Deep Neural Networks.
+        [arXiv:1602.07868]
+    [3] Takeru Miyato, Toshiki Kataoka, Masanori Koyama, Yuichi Yoshida. Spectral
+        Normalization for Generative Adversarial Networks. ICLR 2018.
+
+    """
+
+    domain = constraints.real
+    codomain = constraints.real
```
fritzo, line 148: nit: I guess we could use `constraints.real_vector` here. Do you know if that breaks anything?

stefanwebb: It doesn't seem to break anything as far as I'm aware... I should make this change then for the other `event_dim == 1` transforms

pyro/distributions/transforms/matrix_exponential.py
stefanwebb: Done!

pyro/distributions/transforms/matrix_exponential.py
stefanwebb: There's no `constraints.real_tensor` though for when `event_dim > 1`...

stefanwebb: I've made these changes for all transforms except for `GeneralizedChannelPermute`

PR #2557, opened 2020-07-14
Add helper to convert Bio.Phylo trees to PyTorch

fritzo:
Addresses #2426 

This adds a helper `bio_phylo_to_times()` to extract coalescent time statistics from a [Bio.Phylo](https://biopython.readthedocs.io/en/latest/api/Bio.Phylo.BaseTree.html#Bio.Phylo.BaseTree.Clade) tree object. This is convenient because `Bio.Phylo` provides parsers for most phylogeny tree formats, including Newick and NEXUS. Note this does not add a new dependency to Pyro: the code is duck typed and works whenever `biopython` is installed (it is not installed by default because `biopython` requires Python 3.6+).

I plan to document this helper in a new tutorial section #2558. I had already used a version of this helper in previous tutorial sketches.

## Tested
- [x] added unit tests
- [x] built docs locally.


fritzo:
Thanks for reviewing!

fritzo:
@eb8680 ready to merge?

PR #2561, opened 2020-07-15
Address cholesky issue with arrowhead mass trix

fehiepsi:
This has seen to be helpful in @martinjankowiak 's sir experiments, so I guess we can include it in the release. The idea is to scale non-diagonal terms down if Cholesky issue happens.

### DIFF
--- a/pyro/ops/arrowhead.py
+++ b/pyro/ops/arrowhead.py
@@ -28,12 +28,22 @@ def sqrt(x):
     # NB: the complexity is O(N * head_size^2)
     # ref: https://en.wikipedia.org/wiki/Schur_complement#Background
     Dsqrt = x.bottom_diag.sqrt()
-    B_Dsqrt = B / Dsqrt.unsqueeze(-2)  # shape: head_size x N
-    schur_complement = A - B_Dsqrt.matmul(B_Dsqrt.t())  # complexity: head_size^2 x N
-    # we will decompose schur_complement to U @ U.T (so that the sqrt matrix
-    # is upper triangular) using some `flip` operators:
-    #   flip(cholesky(flip(schur_complement)))
-    top_left = torch.flip(torch.flip(schur_complement, (-2, -1)).cholesky(), (-2, -1))
+
+    for i in range(6):
+        B_Dsqrt = B / Dsqrt.unsqueeze(-2)  # shape: head_size x N
+        schur_complement = A - B_Dsqrt.matmul(B_Dsqrt.t())  # complexity: head_size^2 x N
+        # we will decompose schur_complement to U @ U.T (so that the sqrt matrix
+        # is upper triangular) using some `flip` operators:
+        #   flip(cholesky(flip(schur_complement)))
+        try:
+            top_left = torch.flip(torch.cholesky(torch.flip(schur_complement, (-2, -1))), (-2, -1))
+            break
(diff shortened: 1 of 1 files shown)


pyro/ops/arrowhead.py
```diff
@@ -28,12 +28,22 @@ def sqrt(x):
...
     # ref: https://en.wikipedia.org/wiki/Schur_complement#Background
     Dsqrt = x.bottom_diag.sqrt()
-    B_Dsqrt = B / Dsqrt.unsqueeze(-2)  # shape: head_size x N
-    schur_complement = A - B_Dsqrt.matmul(B_Dsqrt.t())  # complexity: head_size^2 x N
-    # we will decompose schur_complement to U @ U.T (so that the sqrt matrix
-    # is upper triangular) using some `flip` operators:
-    #   flip(cholesky(flip(schur_complement)))
-    top_left = torch.flip(torch.flip(schur_complement, (-2, -1)).cholesky(), (-2, -1))
+
+    for i in range(6):
```
fritzo, line 32: nit: could you add a comment like
```py
# On cholesky error, retry with smaller tail part B.
num_attempts = 6
for i in range(num_attempts):
    ...
```

pyro/ops/arrowhead.py
```diff
@@ -28,12 +28,22 @@ def sqrt(x):
...
+        B_Dsqrt = B / Dsqrt.unsqueeze(-2)  # shape: head_size x N
+        schur_complement = A - B_Dsqrt.matmul(B_Dsqrt.t())  # complexity: head_size^2 x N
+        # we will decompose schur_complement to U @ U.T (so that the sqrt matrix
+        # is upper triangular) using some `flip` operators:
+        #   flip(cholesky(flip(schur_complement)))
+        try:
+            top_left = torch.flip(torch.cholesky(torch.flip(schur_complement, (-2, -1))), (-2, -1))
+            break
+        except RuntimeError:
+            B = B * 2 ** (-i - 1)
```
fritzo, line 42: Can you confirm that you want super-exponential growth here, not just exponential growth via `B = B / 2`?

fritzo: @martinjankowiak can you confirm this super-exponential backoff is the desired behavior?

martinjankowiak: i can confirm it's what i previously did experiments with

https://github.com/pyro-ppl/pyro/blob/2c709199d5932d48e1d409a9275a4ff67f6e6549/pyro/ops/arrowhead.py#L47

it's probably not the platonic solution but i guess it doesn't hurt... i suspect that in most cases where arrowhead is running into troubles here simplistic regularization will be moderately helpful at best

pyro/ops/arrowhead.py
```diff
@@ -28,12 +28,22 @@ def sqrt(x):
...
+            continue
+        raise RuntimeError("Singular schur complement in computing Cholesky of the input"
+                           " arrayhead matrix")
```
martinjankowiak, line 45: typo: arrayhead

pyro/ops/arrowhead.py
fehiepsi: Thanks for reviewing, Fritz! It is not the desired behavior. I will make the change.

PR #2565, opened 2020-07-20
Bump to version 1.4.0

fritzo:
Bump version strings from 1.3.1 -> 1.4.0

After this merges I'll create a release, upload to PyPi, update the tutorial website, and tweet.


fritzo:
> Do we need to update .travis.yml file to test against PyTorch 1.5.1?

Nope, we still allow `torch>=1.5.0` in setup.py, so there is no need to update .travis.yml

PR #2572, opened 2020-07-23
ConditionalSplineAutoregressive

stefanwebb:
The conditional version of the `SplineAutoregressive` transform.

I also ran `isort` on the whole source which is why there are changes to other files

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -6,7 +6,6 @@ import sys
 
 import sphinx_rtd_theme
 
-
 # import pkg_resources
 
 # -*- coding: utf-8 -*-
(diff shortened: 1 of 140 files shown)


fritzo:
I'm pretty sure `isort` on the whole source will break things due to import order dependencies. If you want, you could figure out how to disable isort on certain files and/or certain lines. In particular:

- pyro.distributions.__init__ has weird ordering dependencies due to some * imports being overridden by specific imports
- I believe for files that use matplotlib, we need to call `matplotlib.use('Agg')` or something before importing pyplot.

In any case, the global isort is appreciated but should be submitted as a separate PR in case we need to revert or git bisect to diagnose regression

stefanwebb:
I created a new PR for this

PR #2575, opened 2020-07-24
Add a horovod optimizer and example

### LINKED ISSUE #2570: FR distributed training via Horovod
What would it take to support multiple GPUs in SVI, as [requested on the forum](https://forum.pyro.ai/t/run-pyro-on-multiple-gpus/1066)?

See Horovod's [pytorch](https://github.com/horovod/horovod/blob/master/docs/pytorch.rst) [examples](https://github.com/horovod/horovod/blob/master/examples/pytorch_mnist.py).

## Tentative Tasks
- [x] Support some combination of `hvd.DistributedOptimizer` with `PyroOptimizer`. This probably requires extra initialization step to discover all parameters before inference. This definitely requires a static parameter set.
- [x] Add any required initialization logic to SVI
- [x] Add an example or tutorial

cc @mbabadi @dalessioluca

fritzo:
Resolves #2570 

This integrates with [Horovod](https://github.com/horovod/horovod) by implementing a `pyro.optim.HorovodOptimizer` that combines behavior of Pyro's `PyroOptim` and Horovod's `DistributedOptimizer`. The new `HorovodOptimizer` can wrap any other Pyro optimizer e.g. `optim = HorovodOptimizer(ClippedAdam({'lr': 1e-3}))`. I've also added an example demonstrating how to distribute training using the optimizer.

I have avoided adding horovod as a default dependency since it requires other system dependencies (MPI and maybe gloo). Instead I've added a `pyro[horovod]` extras require which points to `horovod[pytorch]`. I've also added a test decorator `requires_horovod` that skips on current CI.

## Tested
Tests with `--no-horovod` are run on CI. Other tests require `horovodrun` and are not run on CI. Because horovod must be run under a separate process, I have added only example tests and no pytest-driven unit tests.

- [x] smoke test `--no-horovod`
- [x] smoke test `--num-proc 1`
- [x] smoke test `--num-proc 2`
- [x] smoke test `--no-horovod` `--cuda`
- [x] smoke test `--num-proc 1` `--cuda`
- [x] smoke test `--num-proc 2` `--cuda`
- [x] verify convergence

### DIFF
--- a/examples/svi_horovod.py
+++ b/examples/svi_horovod.py
@@ -37,8 +37,8 @@ from pyro.nn import PyroModule
 from pyro.optim import Adam, HorovodOptimizer
 
 
-# We define a model as usual, with no reference to Horovod. This model is data
-# parallel and supports subsampling.
+# We define a model as usual, with no reference to Horovod.
+# This model is data parallel and supports subsampling.
 class Model(PyroModule):
     def __init__(self, size):
         super().__init__()
@@ -48,13 +48,19 @@ class Model(PyroModule):
         coeff = pyro.sample("coeff", dist.Normal(0, 1))
         bias = pyro.sample("bias", dist.Normal(0, 1))
         scale = pyro.sample("scale", dist.LogNormal(0, 1))
+
+        # Since we'll use a distributed dataloader during training, we need to
+        # manually pass minibatches of (covariates,data) that are smaller than
+        # the full self.size. In particular we cannot rely on pyro.plate to
+        # automatically subsample, since that would lead to all workers drawing
+        # identical subsamples.
         with pyro.plate("data", self.size, len(covariates)):
             loc = bias + coeff * covariates
             return pyro.sample("obs", dist.Normal(loc, scale),
                                obs=data)
 
 
-# The following is a standard training loop. To epmhasize the Horovod-specific
+# The following is a standard training loop. To emphasize the Horovod-specific
 # parts, we've guarded them by `if args.horovod:`.
 def main(args):
     # Create a model, synthetic data, and a guide.
--- a/examples/svi_horovod.py
+++ b/examples/svi_horovod.py
@@ -1,0 +1,141 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+# Distributed training via Horovod.
+#
+# This tutorial demonstrates how to distribute SVI training across multiple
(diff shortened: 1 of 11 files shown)


fritzo:
@fehiepsi @martinjankowiak do either of you want to review this?

examples/svi_horovod.py
```diff
@@ -0,0 +1,159 @@
...
+        coeff = pyro.sample("coeff", dist.Normal(0, 1))
+        bias = pyro.sample("bias", dist.Normal(0, 1))
+        scale = pyro.sample("scale", dist.LogNormal(0, 1))
+        with pyro.plate("data", self.size, len(covariates)):
+            loc = bias + coeff * covariates
+            return pyro.sample("obs", dist.Normal(loc, scale),
+                               obs=data)
+
+
+# The following is a standard training loop. To epmhasize the Horovod-specific
```
martinjankowiak, line 57: typo: epmhasize

pyro/optim/horovod.py
```diff
@@ -0,0 +1,47 @@
...
+    r"""
+    Distributed wrapper for a :class:`~pyro.optim.optim.PyroOptim` optimizer.
+
+    This class wraps a ``PyroOptim`` object similar to the way
+    :func:`horovod.torch.DistributedOptimizer` wraps a
+    :class:`torch.optim.Optimizer`.
+
+    .. note::
+
+        This requires :mod:`horovod.torch` is installed, e.g. via
```
martinjankowiak, line 19: grammar typo

examples/svi_horovod.py
```diff
@@ -0,0 +1,159 @@
...
+class Model(PyroModule):
+    def __init__(self, size):
+        super().__init__()
+        self.size = size
+
+    def forward(self, covariates, data=None):
+        coeff = pyro.sample("coeff", dist.Normal(0, 1))
+        bias = pyro.sample("bias", dist.Normal(0, 1))
+        scale = pyro.sample("scale", dist.LogNormal(0, 1))
+        with pyro.plate("data", self.size, len(covariates)):
```
martinjankowiak, line 51: so what happens if the user relies on `plate`-controlled subsampling? the same computation happens on all workers?

fritzo: Good point, I've added a couple sentences explaining requirements.

fritzo:
Thanks for reviewing!

PR #2576, opened 2020-07-27
removed explicit calls to `forward` in VAE tutorial

jamestwebber:
Calling the `forward` method bypasses anything that PyTorch/Pyro does in `__call__()` and so it's not recommended (I just learned 😎 ).

Removing it from this tutorial so that others don't make my mistake.


jamestwebber:
The semi-supervised VAE tutorial had the same thing going on so I removed that too. The rest of the examples were okay.

Issue #2580, opened 2020-07-29
FR Funsor-based Pyro backend


eb8680:
This issue is for tracking Funsor integration with Pyro in the form of a new `pyroapi` backend that shares lots of code with Pyro and lives in `pyro.contrib.funsor`, as implemented in the original PR #2307 and NumPyro PR https://github.com/pyro-ppl/numpyro/pull/572 and recently enabled by the removal of Python 3.5 support from PyTorch 1.6.

The first version of `pyro.contrib.funsor` hews very closely to Pyro's existing design and semantics, including replicating parallel enumeration and plate semantics exactly, preserving `trace` and `replay`, and making extensive use of Pyro's current test suite.  I'm hoping to progressively refine away much of this legacy design after the first version is merged.

Tentative sequence of PRs expected based on the contents of #2307 (mostly done already there):
- [x]  #2583 `make_handler` change, `funsor[torch]` optional dependency, and initial `pyro.contrib.funsor` library creation: `primitives`, `to_funsor`, `to_data`, `markov`, `pyroapi`, and `test_named.py`.
- [x] #2584 First introductory tutorial
- [x] #2588 `GlobalNamedMessenger` and a second tutorial
- [x] #2589 Exact-only `EnumMessenger`, `trace`, `replay`, and `test_valid_models.py`
- [x] #2593 Vectorized `PlateMessenger` and `test_valid_models_plate.py`
- [x] #2593 Support for subsampling and `pyro.subsample` in new `PlateMessenger`, and tests
- [x] #2594 Sequential `PlateMessenger` and `test_valid_models_sequential_plate.py`
- [x] #2595 `TraceTMC_ELBO`, sampling enumeration strategies, and `test_tmc.py`.
- [x] #2596 `TraceEnum_ELBO` and `test_enum.py`

Remaining tasks to wrap up the first version of `contrib.funsor` after #2596 is merged:
- [x] #2583 Support missing or globally inconsistent input specifications in `pyro.to_funsor` and `pyro.to_data`
- [x] #2583 xfailing `test_pyroapi.py`
- [x] #2610 Update domain syntax
- [x] #2611 JIT support for `TraceEnum_ELBO` and `TraceTMC_ELBO`
- [x] #2611 missing ELBOs necessary to make `test_pyroapi.py` pass (e.g. add a simple `Trace_ELBO`)
[...]

PR #2475, opened 2020-05-11
Initialize pyro.contrib.funsor with new primitives and some tests

eb8680:
This PR is the first in a series extracted from #2307 

It contains a new `contrib` module, `pyro.contrib.funsor`, that implements a new Pyro backend with two new primitives, `pyro.to_funsor` and `pyro.to_data`, and a reimplementation of `pyro.markov`.  These two primitives are effectful versions of `funsor.to_funsor` and `funsor.to_data` that maintain a globally consistent mapping between names and positional dimensions, allowing us to write Funsor-based inference algorithms that use enumeration.  A first tutorial that explains more about the design goals is available in a separate followup PR #2476 for easier review.

I would say the details of this PR are not too important, and can be revisited later because they are completely isolated behind the new primitives.  In particular, the `DimStack` design only has to satisfy two constraints: 
1. Like Pyro's existing enumeration machinery it must support dimension allocation in constant time (so enumeration in an HMM is still linear in the number of variables rather than quadratic)
2. Together with the new `pyro.markov` it must replicate as closely as possible the dimension allocation behavior of Pyro's existing enumeration machinery.  This is important for backwards compatibility and allowed me to write detailed tests that compare the results of enumeration in Pyro and `pyro.contrib.funsor`, as will become evident in followup PRs.

Tested:
- In this PR I have included some basic standalone tests for `pyro.to_funsor`/`pyro.to_data` and `pyro.markov`. 
- In #2307 this code is tested much more extensively and stringently by comparing the tensor shapes of enumeration reimplimented with `pyro.contrib.funsor` to enumeration in Pyro. I will be adding these tests in subsequent PRs (update: started in #2479).

PR #2583, opened 2020-08-04
Initialize contrib.funsor with new primitives, pyroapi backend, and some tests

eb8680:
Addresses #2580.  Rewrite of #2475.

This PR contains a new `contrib` module, `pyro.contrib.funsor`, that implements a new Pyro backend with two new primitives, `pyro.to_funsor` and `pyro.to_data`, and a reimplementation of `pyro.markov`.  These two primitives are effectful versions of `funsor.to_funsor` and `funsor.to_data` that maintain a globally consistent mapping between names and positional dimensions, allowing us to write Funsor-based inference algorithms that use enumeration.  A first tutorial that explains more about the design goals is available in a separate followup PR for easier review.

I would say the details of this PR are interesting but not too important, and can be revisited later because they are completely isolated behind the new primitives.  In particular, the `DimStack` design only has to satisfy two constraints: 
1. Like Pyro's existing enumeration machinery it must support dimension allocation in constant time (so enumeration in an HMM is still linear in the number of variables rather than quadratic)
2. Together with the new `pyro.markov` it must replicate as closely as possible the dimension allocation behavior of Pyro's existing enumeration machinery.  This is important for backwards compatibility and allowed me to write detailed tests that compare the results of enumeration in Pyro and `pyro.contrib.funsor`, as will become evident in followup PRs.

Tested:
- In this PR I have included some basic standalone tests for `pyro.to_funsor`/`pyro.to_data` and `pyro.markov`. 
- In #2307 this code is tested much more extensively and stringently by comparing the tensor shapes of enumeration reimplimented with `pyro.contrib.funsor` to enumeration in Pyro. I will be adding these tests in subsequent PRs.


pyro/poutine/handlers.py
```diff
@@ -98,27 +99,27 @@
...
     def handler(fn=None, *args, **kwargs):
-        if fn is not None and not callable(fn):
+        if fn is not None and not (callable(fn) or isinstance(fn, collections.abc.Iterable)):
             raise ValueError(
                 "{} is not callable, did you mean to pass it as a keyword arg?".format(fn))
         msngr = msngr_cls(*args, **kwargs)
         return functools.update_wrapper(msngr(fn), fn, updated=()) if fn is not None else msngr
 
-    return handler
+    handler_name = _re2.sub(
```
fritzo, line 112: nit: could you add a little comment explaining this string munging?

setup.py
```diff
@@ -116,6 +116,9 @@
             'yapf',
         ],
         'horovod': ['horovod[pytorch]>=0.19'],
+        'funsor': [
+            'funsor[torch] @ git+git://github.com/pyro-ppl/funsor.git@master',  # TODO update to release branch
+        ],
```
fritzo, line 121: It would be a little safer to pin to either a commit or a release. I'm worried that if we pin to a moving branch and that branch changes then we wouldn't be able to git bisect. What do you think of any of these alternatives:
- pinning to funsor 0.2.0
- releasing current master and pinning to 0.2.1
- creating a tag in the funsor repo named say `pyro-1.4.0` or maybe `pyro-1.4.1-pre0`, and continuing to increment these tags as you simultaneously develop `funsor` and `pyro`?

tests/contrib/funsor/conftest.py
```diff
@@ -0,0 +1,13 @@
...
+import pytest
+
+
+def pytest_collection_modifyitems(items):
+    for item in items:
+        if item.nodeid.startswith("tests/contrib/funsor"):
+            if "stage" not in item.keywords:
+                item.add_marker(pytest.mark.stage("funsor"))
+            if "init" not in item.keywords:
+                item.add_marker(pytest.mark.init(rng_seed=123))
```
fritzo, line 13: I'm curious whether and how pytest avoids import errors when collecting tests on a system on which funsor is not installed. Do we want to support e.g. `make test` which runs only `stage=unit` (no funsor tests) on a system when funsor is not installed?

One possible solution is, similar to my tests using BioPython,
```python
try:
    import funsor
except ImportError:
    pytestmark = pytest.mark.skip(reason="funsor is not installed")
```

.travis.yml
```diff
@@ -27,6 +27,7 @@ install:
     - pip install https://github.com/pyro-ppl/pyro-api/archive/master.zip
     - pip install torch==1.5.0+cpu torchvision==0.6.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
     - pip install .[test]
+    - pip install git+https://github.com/pyro-ppl/funsor.git@master#egg=funsor[torch]
```
fritzo, line 30: Can we move this down into the `funsor` stage?

tests/contrib/funsor/test_pyroapi_funsor.py
```diff
@@ -0,0 +1,20 @@
...
+
+from pyroapi import pyro_backend
+from pyroapi.tests import *  # noqa F401
+
+# triggers backend registration
+import pyro.contrib.funsor  # noqa: F401
+
+
+# TODO get this working again...
+pytestmark = pytest.mark.skip()
```
fritzo, line 14: Just curious, is there a reason to use `skip` rather than `xfail`?

setup.py
eb8680: I've just pinned it to a recent commit in funsor `master` for now, like NumPyro sometimes does with jax.

tests/contrib/funsor/conftest.py
eb8680: I tried something like that, we'll see if it works.

PR #2586, opened 2020-08-05
[WIP] Fix summations over repeated plate dims in RenyiELBO

sharrison5:
Potential fix for second issue in #2361 (i.e. summation over plates that share `dim`).

To discuss:
 + This simply removes repeated elements in `sum_dims`. My understanding of plate mechanics is that this is valid mathematically, but there could well be subtle issues (as per e.g. #2220) that I may have missed.
 + Should the regression tests introduced here and in #2541 be parameterised to cover more `ELBO` instances?

### DIFF
--- a/pyro/infer/util.py
+++ b/pyro/infer/util.py
@@ -106,14 +106,14 @@ def get_plate_stacks(trace):
 
 def get_dependent_plate_dims(sites):
     """
-    Return a list of dims for plates that are not common to all sites.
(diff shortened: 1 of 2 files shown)


fritzo:
> Should the ... tests ... cover more `ELBO` instances?

I think there is no need, since `get_dependent_plate_dims()` is used only by `RenyiELBO`.

sharrison5:
Great, thanks!

> Is there work that remains to be done?

Not as far as I'm concerned 👍

PR #2587, opened 2020-08-06
Reducing code duplication in minipyro

ciguaran:
Reducing code duplication in minipyro. 
Testing Elbo implementations
Found a slight difference in JIT Elbo for minipyro, wrote a test that highlights that.

### DIFF
--- a/pyro/contrib/minipyro.py
+++ b/pyro/contrib/minipyro.py
@@ -177,14 +177,18 @@ def apply_stack(msg):
     return msg
 
 
-def start_stack_call_or_apply(type, name, fn, obs, *args, **kwargs):
+# sample is an effectful version of Distribution.sample(...)
+# When any effect handlers are active, it constructs an initial message and calls apply_stack.
+def sample(name, fn, *args, **kwargs):
+    obs = kwargs.pop('obs', None)
+
     # if there are no active Messengers, we just draw a sample and return it as expected:
     if not PYRO_STACK:
         return fn(*args, **kwargs)
 
     # Otherwise, we initialize a message...
     initial_msg = {
-        "type": type,
+        "type": "sample",
         "name": name,
         "fn": fn,
         "args": args,
@@ -195,13 +199,6 @@ def start_stack_call_or_apply(type, name, fn, obs, *args, **kwargs):
     # ...and use apply_stack to send it to the Messengers
     msg = apply_stack(initial_msg)
     return msg["value"]
-
-
-# sample is an effectful version of Distribution.sample(...)
-# When any effect handlers are active, it constructs an initial message and calls apply_stack.
-def sample(name, fn, *args, **kwargs):
-    obs = kwargs.pop('obs', None)
-    return start_stack_call_or_apply("sample", name, fn, obs, *args, **kwargs)
 
 
 # param is an effectful version of PARAM_STORE.setdefault that also handles constraints.
@@ -228,7 +225,21 @@ def param(name, init_value=None, constraint=torch.distributions.constraints.real
         return constrained_value
 
     # if there are no active Messengers, we just draw a sample and return it as expected:
-    return start_stack_call_or_apply("param", name, fn, None, init_value, constraint)
+    if not PYRO_STACK:
+        return fn(init_value, constraint)
+
+    # Otherwise, we initialize a message...
+    initial_msg = {
+        "type": "param",
+        "name": name,
+        "fn": fn,
+        "args": (init_value, constraint),
+        "value": None,
+    }
+
+    # ...and use apply_stack to send it to the Messengers
+    msg = apply_stack(initial_msg)
+    return msg["value"]
 
 
 # boilerplate to match the syntax of actual pyro.plate:
@@ -264,14 +275,14 @@ class Adam:
 
 
 # This is a unified interface for stochastic variational inference in Pyro.
-# The actual construction of the loss is taken care of by `loss_factory`.
+# The actual construction of the loss is taken care of by `loss`.
 # See http://docs.pyro.ai/en/stable/inference_algos.html
 class SVI:
-    def __init__(self, model, guide, optim, loss_factory):
+    def __init__(self, model, guide, optim, loss):
         self.model = model
         self.guide = guide
         self.optim = optim
-        self.loss_factory = loss_factory
+        self.loss = loss
 
     # This method handles running the model and guide, constructing the loss
     # function, and taking a gradient step.
@@ -282,7 +293,7 @@ class SVI:
         with trace() as param_capture:
             # We use block here to allow tracing to record parameters only.
             with block(hide_fn=lambda msg: msg["type"] == "sample"):
-                loss = self.loss_factory(self.model, self.guide, *args, **kwargs)
+                loss = self.loss(self.model, self.guide, *args, **kwargs)
         # Differentiate the loss.
         loss.backward()
         # Grab all the parameters from the trace.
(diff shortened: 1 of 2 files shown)


pyro/contrib/minipyro.py
```diff
@@ -201,6 +197,13 @@ def sample(name, fn, *args, **kwargs):
     return msg["value"]
 
 
+# sample is an effectful version of Distribution.sample(...)
+# When any effect handlers are active, it constructs an initial message and calls apply_stack.
+def sample(name, fn, *args, **kwargs):
+    obs = kwargs.pop('obs', None)
+    return start_stack_call_or_apply("sample", name, fn, obs, *args, **kwargs)
+
+
```
fritzo, line 206: While code deduplication is nice in the main Pyro library, I feel that in minipyro it is advantageous to explicitly spell out exactly what `sample` and `param` do, with minimal abstraction. One of the original motivations of minipyro was to be less abstract because original-Pyro's abstractions made it difficult to read (even for Pyro devs 🙂).

pyro/contrib/minipyro.py
```diff
@@ -324,22 +313,19 @@ def elbo(model, guide, *args, **kwargs):
     # terms, this means our loss is constructed as an expectation w.r.t. the joint
     # distribution defined by the guide.
     model_trace = trace(replay(model, guide_trace)).get_trace(*args, **kwargs)
-    # We will accumulate the various terms of the ELBO in `elbo`.
+    # Return (-elbo) since by convention we do gradient descent on a loss and
+    # the ELBO is a lower bound that needs to be maximized. Given this,
+    # loop over all the sample sites in the model and add the corresponding
+    # - log p(z) term to the ELBO. Same thing for guide but adding + log p(z)
+    return elbo_terms(guide_trace) - elbo_terms(model_trace)
```
fritzo, line 320: Hmm, while this is fewer lines of code, it is more abstract. Again I think we need different [complexity metrics](https://en.wikipedia.org/wiki/Cyclomatic_complexity) for main Pyro and minipyro: In Pyro we should prefer well-factored code; and in minipyro we should [prefer monolithic blocks of code](http://number-none.com/blow/blog/programming/2014/09/26/carmack-on-inlined-code.html) and specifically blocks of code that look like minified versions of main Pyro.

pyro/contrib/minipyro.py
```diff
@@ -275,14 +264,14 @@ def __call__(self, params):
 
 
 # This is a unified interface for stochastic variational inference in Pyro.
-# The actual construction of the loss is taken care of by `loss`.
+# The actual construction of the loss is taken care of by `loss_factory`.
 # See http://docs.pyro.ai/en/stable/inference_algos.html
 class SVI:
-    def __init__(self, model, guide, optim, loss):
+    def __init__(self, model, guide, optim, loss_factory):
```
fritzo, line 270: I think we should keep the original name `loss` so as to support specification as kwarg. It is common to pass this as a kwarg as in `SVI(model, guide, optim=optim, loss=loss)`. I agree that `loss` is a confusing name for what is actually a factory or loss function, but OTOH I'd prefer not to break existing code and idioms.

pyro/contrib/minipyro.py
```diff
@@ -275,14 +264,14 @@ def __call__(self, params):
...
         self.model = model
         self.guide = guide
         self.optim = optim
-        self.loss = loss
+        self.loss_factory = loss_factory
```
fritzo, line 274: :+1: I think it's safe to rename the member. Is that what the main-Pyro SVI class does, so users won't be confused?

tests/contrib/test_minipyro.py
```diff
@@ -100,7 +115,7 @@ def guide(data):
     with pyro_backend(backend):
         Elbo = infer.JitTrace_ELBO if jit else infer.Trace_ELBO
         elbo = Elbo(ignore_jit_warnings=True)
-        assert_ok(model, guide, elbo, data)
+        assert_ok(model, guide, elbo, 2, data)
```
fritzo, line 118: nit: would you mind writing this in a self-documenting style
```py
steps = 2
assert_ok(model, guide, elbo, steps, data)
```
or maybe
```py
assert_ok(model, guide, steps=2, data=data)
```
or similar?

tests/contrib/test_minipyro.py
```diff
@@ -186,18 +201,47 @@ def guide():
...
-    def model():
-        locs = pyro.param("locs", torch.randn(3), constraint=constraints.real)
-        scales = pyro.param("scales", ops.exp(torch.randn(3)), constraint=constraints.positive)
-        p = torch.tensor([0.5, 0.3, 0.2])
-        x = pyro.sample("x", dist.Categorical(p))
-        pyro.sample("obs", dist.Normal(locs[x], scales[x]), obs=data)
+    with pyro_backend(backend):
+        Elbo = infer.JitTrace_ELBO if jit else infer.Trace_ELBO
+        elbo = Elbo(ignore_jit_warnings=True)
+        assert_ok(constrained_model, guide_constrained_model, elbo, 2, data)
```
fritzo, line 207: ditto
```py
steps = 2
assert_ok(..., steps, ...)
```

tests/contrib/test_minipyro.py
```diff
@@ -186,18 +201,47 @@ def guide():
...
+    SVI, JitTrace_ELBO yield
+    similar results for both pyro and minipyro.
+    This is a relaxed assertion over
+    the one reached in test_elbo_equivalence
+    Note: 15 is the minimun number for iterations for this
+    to happen.
+    """
+    pyro.set_rng_seed(0)
+    data = torch.tensor(0.5)
+    elbo_test_case(backend, True, 0.4780, data, 15)
```
fritzo, line 224: nit: ditto, self-docmenting or kwargs?
```py
elbo_test_case(backend, jit=True, expected_elbo=0.4780, data=data, steps=15)
```

tests/contrib/test_minipyro.py
```diff
@@ -186,18 +201,47 @@ def guide():
...
+
+@pytest.mark.parametrize(["backend", "jit"], [("pyro", True), ("pyro", False), ("minipyro", False)])
+def test_elbo_equivalence(backend, jit):
+    """
+    Given model and guide
+    Calculating the elbo yields equivalent results
+    """
+    pyro.set_rng_seed(0)
+    data = torch.tensor(0.5)
+    elbo_test_case(backend, jit, expected_elbo=0.4780, data=data)
```
fritzo, line 235: ditto

ciguaran:
> Thanks for noticing this and adding a test case!
:)!

> Do you have an opinion on a good backwards-compatible way we could resolve the interface difference? Maybe we could add a method to the main Pyro `Elbo` base class?
> 
> ```python
> class Elbo:
>     ...
>     def __call__(self, *args, **kwargs):
>         return self.differentiable_loss(*args, **kwargs)
> ```
> 
> I guess one issue is that `Elbo` subclasses distinguish `.loss()` from `.differentiable_loss()` whereas the minipyro only implements the differentiable version for simplicity.

Given that differentiable loss returns a tensor and loss a float (aka Liskov's substitution principle doesn't hold between them), I can't think of any other backward compatible solution. If we do what you suggested, we should add differentiable_loss as abstract method of the base class (and we would also need all subclases to implement it) I wonder why loss and differentiable loss aren't abstract method of the base class, are there substances that do not implement both?

fritzo:
> I wonder why loss and differentiable loss aren't abstract method

Historically the first two methods were `.loss()` and the in-place operating `.loss_and_grads()`. The `.differentiable_loss()` method was later added for compatibility with `torch.autograd.grad()`. In fact `.loss_and_grads()` can be cheaper than `.differentiable_loss()`. Some classes do not implement `.differentiable_loss()`, e.g. `TraceGraph_ELBO` and `RenyiELBO`.

ciguaran:
So one way of fixing this would be for minipyro's Trace_ELBO and JitTrace_ELBO to implement .loss. That would make things consistent at the expense of a new method implementation.
Another way, make elbo callable (but with loss() inside __call__ not differentiable_loss()) and make loss an abstract method of the base Elbo class.

fritzo:
@eb8680 do you have an opinion on the `Elbo` interface mismatch? I'm unsure whether it is more important for the interfaces to be consistent, or if we should keep the current balance between old-pyro being backwards compatible and minipyro being simple.

eb8680:
The current, simpler version seems fine - this is exactly the sort of unforunate Pyro implementation detail that Minipyro was designed to ignore in the interest of clarity.

martinjankowiak:
> The current, simpler version seems fine - this is exactly the sort of unforunate Pyro implementation detail that Minipyro was designed to ignore in the interest of clarity.

👍  yeah to my mind this was always primarily about being pedagogical

ciguaran:
Ok, so shall I rollback all changes but keep the tests, or close this PR altogether?

eb8680:
> Ok, so shall I rollback all changes but keep the tests, or close this PR altogether?

I'd be happy to keep the tests.  Generally, I think your intuition with this PR was correct, it's just that `minipyro` happens to have somewhat different design goals from Pyro itself.  I'd certainly be interested in any thoughts you might have on cleaning up the ELBO interfaces in Pyro itself.

tests/contrib/test_minipyro.py
ciguaran: Apologies I don't follow what's the suggestion in this case

ciguaran:
@eb8680 changes made!. Regarding the interfaces in pyro, do you have any doc or some initial idea? By what you wrote it seems you already have some thoughts on why cleanup may be worth it.

fritzo (approved):

PR #2592, opened 2020-08-11
Adding ch.1 of the book Probabilistic Programming and Bayesian Methods for Hackers

ucals:
This PR ports Chapter 1 of the book [Probabilistic Programming and Bayesian Methods for Hackers](https://camdavidsonpilon.github.io/Probabilistic-Programming-and-Bayesian-Methods-for-Hackers/) to Pyro.

I believe these introductory materials can significantly increase the adoption of the language.

Cheers!


eb8680:
@ucals thanks for porting this, I agree it will be very helpful for new users! Where do you think is the most appropriate place for this notebook to live?  It looks like the PyMC and TFP versions both live in the book's repository rather than in PyMC3 or TFP themselves, and I see you've created [a corresponding PR there](https://github.com/CamDavidsonPilon/Probabilistic-Programming-and-Bayesian-Methods-for-Hackers/pull/506) - would it be better to link to that on the Pyro tutorials web page?

ucals:
@eb8680 , yes, you are totally right: it's best for the notebooks to live in the book's repository. It would be great, though, as you pointed out, to link to that on the Pyro tutorials web page. I'll create a PR to that.. Cheers!

PR #2594, opened 2020-08-11
Implement sequential plate in contrib.funsor

eb8680:
Addresses #2580. Blocked by #2593.

This PR adds support for sequential (as opposed to vectorized) `pyro.plate`s in `pyro.contrib.funsor`, sticking to the semantics of Pyro's sequential plates.

It also contains more tests ported from `tests/infer/test_valid_models.py` and making use of the `assert_ok` utility added in #2589.

PR #2595, opened 2020-08-11
Implement sampling enumeration strategies and TraceTMC_ELBO in contrib.funsor

eb8680:
Addresses #2580. ~~Blocked by #2594.~~

This PR adds linear-time Monte Carlo enumeration strategies (diagonal and mixture, based on the ones already in Pyro) and a new implementation of `TraceTMC_ELBO` using the enumeration machinery in `pyro.contrib.funsor`.  I've intentionally kept the code looking similar to the one in Pyro (notably the continued use of `trace`/`replay` and the call to `sum_product`), but it's still a nice illustration of all the boilerplate cut out by `contrib.funsor` in the implementation of relatively complicated inference algorithms.  Funsor's simplified indexing behavior also trivially fixes the issues with the Monte Carlo strategies incorrectly sharing particle IDs across plate slices that I was worried about in #2258.

The tests in this PR are basically copied from `tests/infer/test_tmc.py`.  The shape behavior of the enumeration strategies is also tested more systematically in a followup PR containing `TraceEnum_ELBO`.

### DIFF
--- a/tests/contrib/funsor/test_tmc.py
+++ b/tests/contrib/funsor/test_tmc.py
@@ -1,0 +1,139 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import logging
+import math
+
+import pytest
+import torch
+from torch.autograd import grad
+from torch.distributions import constraints
+
+import funsor
+import pyro.contrib.funsor
+
+from pyroapi import infer, pyro, pyro_backend
+from pyroapi import distributions as dist
+from tests.common import assert_equal  # , xfail_param
+
+
+funsor.set_backend("torch")
+torch.set_default_dtype(torch.float32)
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.mark.parametrize("depth", [1, 2, 3, 4, 5])
+@pytest.mark.parametrize("num_samples", [None, 200])
+@pytest.mark.parametrize("max_plate_nesting", [2, 3])
+@pytest.mark.parametrize("tmc_strategy", ["diagonal", "mixture"])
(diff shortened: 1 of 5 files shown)


tests/contrib/funsor/test_tmc.py
```diff
@@ -0,0 +1,142 @@
...
+try:
+    import funsor
+    import pyro.contrib.funsor
+    from pyroapi import distributions as dist
+    from pyroapi import infer, pyro, pyro_backend
+    funsor.set_backend("torch")
+except ImportError:
+    pytestmark = pytest.mark.skip(reason="funsor is not installed")
+
+torch.set_default_dtype(torch.float32)
```
eb8680, line 24: I suspect this is the source of the mysterious nonlocal CI failures, but I may have to make a fix upstream in Funsor to remove it.

fritzo: Why is this required? We should be testing everything with double precision on CI, as specified at https://github.com/pyro-ppl/pyro/blob/dev/tests/conftest.py#L13 If you really want local tests to use low precision, you should add that as a fixture with a context manager or sth

eb8680:
Blocked by a bugfix PR upstream in Funsor: https://github.com/pyro-ppl/funsor/pull/349

After that PR lands, I'll need to bump the Funsor dependency commit in this PR in `setup.py` and elsewhere before this can be merged.

tests/contrib/funsor/test_tmc.py
eb8680: It should not be required, and was a workaround for a Funsor issue that should be fixed before this merges.

PR #2596, opened 2020-08-11
Implement fully functioning TraceEnum_ELBO in contrib.funsor

eb8680:
Addresses #2580.  ~~Blocked by #2595.~~

This PR, the last in the series extracted from #2307, adds a fully compliant implementation of `TraceEnum_ELBO` (or at least its `differentiable_loss` method, without sequential enumeration) to `contrib.funsor` that clocks in at under 100 lines of code in one file.  Together with #2595, it serves as a demonstration of how much cleaner and simpler the implementations of Pyro's advanced inference machinery can become using the tools in the new `contrib.funsor` backend #2580.  It is also essentially backend-agnostic thanks to Funsor, and could thus be copied to NumPyro with little or no changes just like the rest of `contrib.funsor`.

Other changes in this PR:
- Allow use of `plate` with `size` but no `name` or `dim` argument for vectorized sampling along a global, nonvisible dimension
- Factor out a `terms_from_trace` helper for collecting and scaling funsors from traces and reuse it in `TraceTMC_ELBO`

Nearly all of the code in this PR consists of tests copied verbatim from `tests/infer/test_enum.py`.  As such, it's not necessary to review them in detail.  The only change is to annotate most test functions with `pyro_backend("contrib.funsor")` decorator so that the tests run under the new backend.


pyro/contrib/funsor/infer/traceenum_elbo.py
```diff
@@ -0,0 +1,82 @@
...
+    return result
+
+
+@copy_docs_from(OrigTraceEnum_ELBO)
+class TraceEnum_ELBO(ELBO):
+
+    def _get_trace(self, *args, **kwargs):
+        raise ValueError("shouldn't be here")
+
+    def differentiable_loss(self, model, guide, *args, **kwargs):
```
eb8680, line 36: Note the similarity of this function to `funsor.minipyro.elbo`: https://github.com/pyro-ppl/funsor/blob/master/funsor/minipyro.py#L435

tests/contrib/funsor/test_enum.py
```diff
@@ -0,0 +1,1393 @@
...
+    _check_loss_and_grads(loss_iplate_iplate, loss_plate_plate)
+
+
+@pytest.mark.parametrize('guide_scale', [1])
+@pytest.mark.parametrize('model_scale', [1])
+@pytest.mark.parametrize('outer_vectorized,inner_vectorized,xfail',
+                         [(False, True, False)],
+                         # (True, False, xfail_param(True, reason="validation not yet implemented")),
+                         # (True, True, xfail_param(True, reason="validation not yet implemented"))],
+                         ids=['iplate-plate'])  # , 'plate-iplate', 'plate-plate'])
```
eb8680, line 1278: Why are these commented out? Check if they work again, they should

tests/contrib/funsor/test_enum.py
```diff
@@ -0,0 +1,1393 @@
...
+        pass
+
+    elbo = infer.TraceEnum_ELBO(max_plate_nesting=1)
+    auto_loss = elbo.differentiable_loss(auto_model, guide)
+    elbo = infer.TraceEnum_ELBO(max_plate_nesting=0)
+    hand_loss = elbo.differentiable_loss(hand_model, guide)
+    _check_loss_and_grads(hand_loss, auto_loss)
+
+
+@pytest.mark.xfail(reason="Not supported in regular Pyro")
```
eb8680, line 499: This test checks a failure case in regular Pyro. It's `xfail`ed here because I haven't included `TraceEnum_ELBO`'s validation logic in my implementation here.

pyro/contrib/funsor/infer/traceenum_elbo.py
```diff
@@ -0,0 +1,82 @@
...
+            plates=prod_vars, eliminate=(prod_vars | sum_vars) - frozenset(cost.inputs)
+        )
+        term = funsor.Integrate(log_prob, cost, sum_vars & frozenset(cost.inputs))
+        term = term.reduce(funsor.ops.add, prod_vars & frozenset(cost.inputs))
+        result += term
+    return result
+
+
+@copy_docs_from(OrigTraceEnum_ELBO)
+class TraceEnum_ELBO(ELBO):
```
eb8680, line 31: Copy draft tutorial from #2307 into this file as comments

pyro/contrib/funsor/infer/tracetmc_elbo.py
```diff
@@ -21,28 +22,25 @@ def _get_trace(self, *args, **kwargs):
         raise ValueError("shouldn't be here")
 
     def differentiable_loss(self, model, guide, *args, **kwargs):
-        with enum(first_available_dim=-self.max_plate_nesting-1):
+        with plate(size=self.num_particles) if self.num_particles > 1 else contextlib2.nullcontext(), \
```
fritzo, line 25: nit: Use an empty `ExitStack` as a `nullcontext` to avoid adding a dependency?
```py
try:
    from contextlib import nullcontext
except ImportError:
    from contextlib import ExitStack as nullcontext
```

setup.py
```diff
@@ -117,6 +117,7 @@
         ],
         'horovod': ['horovod[pytorch]>=0.19'],
         'funsor': [
+            'contextlib2>=0.6.0',
```
fritzo, line 120: nit: remove? (see above)

pyro/contrib/funsor/handlers/replay_messenger.py
```diff
@@ -13,7 +13,11 @@ class ReplayMessenger(OrigReplayMessenger):
     """
     def _pyro_sample(self, msg):
         name = msg["name"]
-        if self.trace is not None and name in self.trace:
+        msg["replay_active"] = True  # indicate replaying so importance weights can be scaled
```
fritzo, line 16: Can you explain the `replay_*` annotations to me? Is there a deep reason these are required in `contrib.funsor` but not in main Pyro? Or asked differently, where does the corresponding logic live in main Pyro? Is it spread between `Distribution.score_parts` and `Trace_ELBO.loss_and_grads()`?

eb8680: I added these to support `poutine.scale` and subsampling in `TraceEnum_ELBO` specifically, while also attempting to make `terms_from_trace` a pure function of a single trace that I could factor out of the ELBO computation.  They are only used within `terms_from_trace`, and are basically small hacks.

The corresponding logic for scaling in the original `TraceEnum_ELBO`, which I condensed but stuck to fairly closely, is spread across the various helpers in `pyro/infer/traceenum_elbo.py`, especially `compute_model_factors`.

The use of `replay_active` is in identifying *model* sites that depend on *model-enumerated* variables, in particular, distinguishing them from enumerated guide sites, since otherwise `terms_from_trace` does not distinguish between model and guide traces.  Within `TraceEnum_ELBO`, log-probabilities at these sites must be scaled for subsampling after all model-enumerated variables are contracted, rather than immediately.  `replay_active` could be replaced by an argument to `terms_from_trace`.

The use of `replay_skipped` is in identifying *model* sites that were enumerated or sampled in the *model*.  Since there is no corresponding guide site, the unnormalized importance weight for such a site is 1, and it would be incorrect to include a `log_prob` term in the model's `log_factors`.  

`replay_skipped` is an optimization. It could be removed by setting `node["funsor"]["log_prob"] = node["funsor"]["log_prob"] * 0.` or `node["funsor"]["log_prob"] = to_funsor(0.)` in `ReplayMessenger._pyro_post_sample`, but the former would be computationally wasteful without https://github.com/pyro-ppl/funsor/issues/235 and the latter would break some of the other tests in `contrib.funsor` that currently rely on checking the shapes of every `log_prob` tensor.

pyro/contrib/funsor/infer/tracetmc_elbo.py
```diff
@@ -21,28 +22,25 @@ def _get_trace(self, *args, **kwargs):
...
-                if name in guide_tr.nodes or node['is_observed']:
-                    log_factors.append(
-                        node["funsor"]["log_prob"] if role == "model" else -node["funsor"]["log_prob"])
-                if node["funsor"].get("log_measure", None) is not None:
-                    log_measures.append(node["funsor"]["log_measure"])
-                    measure_vars |= frozenset(node["funsor"]["log_measure"].inputs)
-                plate_vars |= frozenset(f.name for f in node["cond_indep_stack"] if f.vectorized)
-                measure_vars |= frozenset(node["funsor"]["log_prob"].inputs)
+        model_terms = terms_from_trace(model_tr)
+        guide_terms = terms_from_trace(guide_tr)
```
eb8680, line 31: It's not obvious what the right way to handle scaling is in `TraceTMC_ELBO` - perhaps subsampling-induced rescaling should be distinguished from `handlers.scale` applications?.  Note that the original implementation (and its tests) punts on this as well, so it's probably out of scope for this PR.

pyro/contrib/funsor/handlers/replay_messenger.py
fritzo: Thanks for explaining!

pyro/contrib/funsor/infer/traceenum_elbo.py
```diff
@@ -0,0 +1,99 @@
...
+        if node["type"] != "sample" or type(node["fn"]).__name__ == "_Subsample":
+            continue
+        # grab plate dimensions from the cond_indep_stack
+        terms["plate_vars"] |= frozenset(f.name for f in node["cond_indep_stack"] if f.vectorized)
+        # grab the log-measure, found only at sites that are not replayed or observed
+        if node["funsor"].get("log_measure", None) is not None:
+            terms["log_measures"].append(node["funsor"]["log_measure"])
+            # sum (measure) variables: the fresh non-plate variables at a site
+            terms["measure_vars"] |= frozenset(node["funsor"]["value"].inputs) | frozenset([name]) - terms["plate_vars"]
+        # grab the scale, assuming a common subsampling scale
```
fritzo, line 32: In Pyro we check for this with informative errors
https://github.com/pyro-ppl/pyro/blob/10c767f1e2ca9b2894ad84afe205eb83d346317f/pyro/infer/traceenum_elbo.py#L28-L32

What are your plans for error checking in `contrib.funsor`? I think it's fine to omit most error checking in the first pass, then we can add tests for `assert_warning` and `assert_error`.

eb8680: I'd like to try to centralize the error-checking logic and disentangle it from the elbo implementations, but beyond that I don't have specific plans yet and I'll probably leave it for the next redesign.

Issue #2598, opened 2020-08-13
FR: IWAE_DReG importance weighted VAE with doubly reparametrized gradients


fritzo:
This proposes adding an `IWAE_DReG` loss similar to the `ELBO` class but implementing the [doubly reparametrized gradient estimator](https://arxiv.org/abs/1810.04152). @iffsid points out https://github.com/pytorch/pytorch/issues/25783 that this would be much easier if `Distribution` objects had a `.detach()` method.

PR #2599, opened 2020-08-13
Implement a detach() helper that works on distributions

fritzo:
Addresses #2598, https://github.com/pytorch/pytorch/issues/25783

This implements a function `detach(-)` that creates a deep copy of an arbitrary Python object but with all tensors detached (and without copying underlying tensor data). Then main use case is in inference algorithms that require mixed detachment of gradients, e.g. `d = MyDist(param)` followed by `d.rsample()` but `detach(d).log_prob(x)`.

This works by overriding `deepcopy`'s behavior to use a modified memoizer dict. This seems like the cleanest way to apply arbitrary transforms on leaves of python objects (similar to JAX's PyTree). We could in principle use this trick for other leaf transforms.

## Tested
- [x] added unit tests
- [x] refactored `TraceGraph_ELBO` to use the new helper; should be covered by existing tests.

cc @iffsid

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,11 +1,10 @@
 [flake8]
 max-line-length = 120
 exclude = docs/src, build, dist, .ipynb_checkpoints
-extend-ignore = E741
+extend-ignore = E721,E741
 
 [isort]
 line_length = 120
-not_skip = __init__.py
 skip_glob = .ipynb_checkpoints
 known_first_party = pyro, tests
 known_third_party = opt_einsum, six, torch, torchvision
--- a/pyro/distributions/util.py
+++ b/pyro/distributions/util.py
@@ -1,6 +1,8 @@
 # Copyright (c) 2017-2019 Uber Technologies, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
+import copy
+import ctypes
 import functools
 import numbers
 import weakref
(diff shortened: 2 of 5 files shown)


setup.cfg
```diff
@@ -1,11 +1,10 @@
 [flake8]
 max-line-length = 120
 exclude = docs/src, build, dist, .ipynb_checkpoints
-extend-ignore = E741
+extend-ignore = E721,E741
 
 [isort]
 line_length = 120
-not_skip = __init__.py
```
fritzo, line 8: fixes a new error emitted by isort 5.0

setup.cfg
```diff
@@ -1,11 +1,10 @@
 [flake8]
 max-line-length = 120
 exclude = docs/src, build, dist, .ipynb_checkpoints
-extend-ignore = E741
```
fritzo, line 4: ignore errors about test assertions `assert type(x) is type(y)`

fritzo:
> Is this exercised by any of the existing `JitTraceGraph_ELBO` tests?

No, the `JitTraceGraph_ELBO` tests seem low value relative to planned follow-up tests that use `detach(-)` in `RenyiELBO` and a new `IWAE_ELBO` #2598

fritzo:
OK, I've added a low-level jit test to check that `detach(-)` can be called in jitted code.

PR #2611, opened 2020-08-21
Add a basic Trace_ELBO and JIT ELBOs to contrib.funsor

eb8680:
Addresses #2580. ~~Blocked by #2610.~~

This PR is intended to get the remaining `pyroapi` tests passing for `contrib.funsor`.  In addition to removing `xfail`s from `test/contrib/funsor/test_pyroapi.py`, it adds the following components to `contrib.funsor`:
1. A base `ELBO` that contains default implementations of `loss` and `loss_and_grads` methods so that the ELBOs that only implement `differentiable_loss` are automatically compatible with `pyro.infer.SVI`.
2. A basic funsor-based implementation of `Trace_ELBO`.
3. a helper `make_jit_elbo` for creating `JitTrace*_ELBO` implementations from `Trace*_ELBO`s
4. Generated versions of `JitTraceTMC_ELBO`, `JitTrace_ELBO`, `JitTraceEnum_ELBO`


pyro/contrib/funsor/handlers/plate_messenger.py
```diff
@@ -68,7 +68,7 @@ def __init__(self, name, size=None, subsample_size=None, subsample=None, dim=Non
             name, size, subsample_size, subsample, use_cuda, device)
         self.subsample_size = subsample_size
         self._full_size = size
-        self._scale = size / subsample_size
+        self._scale = size // subsample_size
```
fritzo, line 71: Don't we want floating point division here? E.g. if my dataset has size 100 but I sample minibatches of size 40 the scale should be 2.5, not 2.0.

pyro/contrib/funsor/infer/elbo.py
```diff
@@ -0,0 +1,52 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import weakref
+
+import pyro.ops.jit
+from pyro.infer import ELBO as OrigELBO
```
fritzo, line 7: nit: could you import this as `_ELBO` so it is not captured by `from pyro.infer.elbo import *`? Alternatively
```py
import pyro.infer

class ELBO(pyro.infer.ELBO):
    ...
```

pyro/contrib/funsor/infer/elbo.py
```diff
@@ -0,0 +1,52 @@
...
+    def loss(self, model, guide, *args, **kwargs):
+        return self.differentiable_loss(model, guide, *args, **kwargs).detach().item()
+
+    def loss_and_grads(self, model, guide, *args, **kwargs):
+        loss = self.differentiable_loss(model, guide, *args, **kwargs)
+        loss.backward()
+        return loss.item()
+
+
+def make_jit_elbo(elbo_cls):
```
fritzo, line 28: Would it be more pythonic to implement this as a base class? I think the following might work:
```py
class Jit_ELBO(ELBO):
    def differentiable_loss(...):
        ...

class JitTrace_ELBO(Jit_ELBO, Trace_ELBO):
    pass
```

fritzo (approved):

PR #2614, opened 2020-08-31
Make HMM examples compatible with contrib.funsor through pyroapi

eb8680:
Addresses #2580. ~~Blocked by #2611. Blocked by #2667~~

Current status:

The HMM examples work pretty well with and without the JIT, which as expected provides a substantial speedup for the funsor backend by compiling away all pattern-matching in Funsor - in fact, the runtimes after JIT compilation of the two backends are basically the same.  I've also gone through training most of the models and verified that they produce essentially the same results.  There are a lot of JIT warnings that refuse to be suppressed for some reason, though, and that seem to be increasing the memory usage of the JIT traces under the funsor backend.

The CJS example doesn't quite work, and needs some fixes in Funsor (adding a `Uniform` distribution, addressing casting issues with `Bernoulli`).  I've removed it from this PR for now.


examples/hmm.py
```diff
@@ -661,5 +672,15 @@ def main(args):
...
-    main(args)
+
+    if args.funsor:
+        import funsor
+        funsor.set_backend("torch")
+        PYRO_BACKEND = "contrib.funsor"
+    else:
+        PYRO_BACKEND = "pyro"
+
+    with ignore_jit_warnings(), pyro_backend(PYRO_BACKEND):
```
fritzo, line 685: How about keeping jit warnings in the Pyro version:
```py
if args.funsor:
    import funsor
    funsor.set_backend("torch")
    with ignore_jit_warnings(), pyro_backend("contrib.funsor"):
        main(args)
else:
    main(args)
```

examples/hmm.py
```diff
@@ -627,14 +638,14 @@ def main(args):
...
-    test_loss = elbo.loss(model, guide, sequences, lengths, args=args, include_prior=False)
+    test_loss = elbo.loss(model, guide, sequences, lengths, batch_size=sequences.shape[0], include_prior=False)
     logging.info('test loss = {}'.format(test_loss / num_observations))
 
     # We expect models with higher capacity to perform better,
     # but eventually overfit to the training set.
     capacity = sum(value.reshape(-1).size(0)
                    for value in pyro.get_param_store().values())
-    logging.info('{} capacity = {} parameters'.format(model.__name__, capacity))
+    logging.info('{} capacity = {} parameters'.format(model_name, capacity))
```
fritzo, line 648: nit: maybe change to `'model{} capacity ...'.format(args.model)` to avoid need to save `model_name` above?

examples/hmm.py
```diff
@@ -569,50 +574,56 @@ def main(args):
...
+        loss = svi.step(sequences, lengths, batch_size=args.batch_size)
         logging.info('{: >5d}\t{}'.format(step, loss / num_observations))
 
     if args.jit and args.time_compilation:
         logging.debug('time to compile: {} s.'.format(elbo._differentiable_loss.compile_time))
 
     # We evaluate on the entire training dataset,
     # excluding the prior term so our results are comparable across models.
-    train_loss = elbo.loss(model, guide, sequences, lengths, args, include_prior=False)
+    train_loss = elbo.loss(model, guide, sequences, lengths, batch_size=sequences.shape[0], include_prior=False)
```
fritzo, line 626: Why do default args no longer work?

examples/hmm.py
```diff
@@ -569,50 +574,56 @@ def main(args):
...
-            poutine.replay(poutine.enum(model, first_available_dim), guide_trace)).get_trace(
+        model_trace = handlers.trace(
+            handlers.replay(handlers.enum(model, first_available_dim), guide_trace)).get_trace(
             sequences, lengths, args=args, batch_size=args.batch_size)
         logging.info(model_trace.format_shapes())
 
+    # Appease the PyTorch JIT
+    model_name = model.__name__
+    model = functools.partial(model, args=args)
+    guide = functools.partial(guide, args=args)
```
fritzo, line 594: nit: more instruction less vent?
```py
# Bind non-PyTorch parameters to make these functions jittable.
model = functools.partial(model, args=args)
guide = functools.partial(guide, args=args)
```

tests/test_examples.py
```diff
@@ -270,3 +289,13 @@ def test_horovod(np, example):
     filename, args = example[0], example[1:]
     filename = os.path.join(EXAMPLES_DIR, filename)
     check_call(horovodrun.split() + [sys.executable, filename] + args)
+
+
+@requires_funsor
+@pytest.mark.parametrize('example', FUNSOR_EXAMPLES)
```
fritzo, line 295: Will we want to mark this as stage=funsor rather than stage=examples?

tests/common.py
```diff
@@ -78,6 +78,13 @@ def TemporaryDirectory():
 requires_horovod = pytest.mark.skipif(horovod is None,
                                       reason="horovod is not available")
 
+try:
+    import funsor
+except ImportError:
+    funsor = None
+requires_funsor = pytest.mark.skipif(funsor is None,
```
fritzo, line 85: nice!

fritzo (commented):
Hmm what are your thoughts on balancing education with generality? In particular, this is a basic tutorial in how to do enumeration in Pyro, so it seems like we should use standard idioms. Should we `from pyroapi import handlers as poutine`? Should Pyro rename `poutine` to `handlers` (if so we should do so at Pyro 2.0, not 1.5)?

examples/hmm.py
eb8680: I'm not sure where exactly the failure starts but I think there are a bunch of places in Funsor and possibly `contrib.funsor` that convert shapes to tuples and integers and screw up PyTorch JIT shape inference, and this causes a shape error here and further down at `test_loss =` after JIT compilation.

eb8680:
> Hmm what are your thoughts on balancing education with generality?

I can convert this into a new file if you prefer

fritzo:
> > Hmm what are your thoughts on balancing education with generality?

> I can convert this into a new file if you prefer

I think it would be best to either (a) `from pyroapi import handlers as poutine` or (ii) fork this file as `examples/contrib/funsor/hmm.py`.

fritzo (commented):
Thanks for refactoring, the cleanup looks great. I think this will allow us to cleanly fork additional examples into `examples/contrib/funsor` as we progress towards Pyro 2.

tests/test_examples.py
eb8680: I did this manually in `.travis.yml`, since it still makes sense to separate example smoke tests from unit tests

examples/contrib/funsor/hmm.py
```diff
@@ -0,0 +1,657 @@
...
+
+    if args.funsor:
+        import funsor
+        funsor.set_backend("torch")
+        PYRO_BACKEND = "contrib.funsor"
+    else:
+        PYRO_BACKEND = "pyro"
+
+    with pyro_backend(PYRO_BACKEND):
+        main(args)
```
fritzo, line 657: Should we hard-code funsor use here, since non-funsor use is covered by examples/hmm.py? I imagine we'll be editing these side-by-side in the future (e.g. with `vim -d`)

eb8680: It's useful for comparing results across the two backends

setup.py
```diff
@@ -120,7 +120,7 @@
         'horovod': ['horovod[pytorch]>=0.19'],
         'funsor': [
             # TODO update to release branch (currently using a recent commit on master)
-            'funsor[torch] @ git+git://github.com/pyro-ppl/funsor.git@e3eba15e2ad7d9db5dd5622c247e0aaa7e81e7dd',
+            'funsor[torch] @ git+git://github.com/pyro-ppl/funsor.git@778e70a567cabf4a40976bc971ea43a857003c2d',
```
fritzo, line 123: Update to 97c1a439d20b893bf877b959dae3ebdf2a051d94 ?

fritzo (approved):

tests/test_examples.py
```diff
@@ -208,6 +208,27 @@ def xfail_jit(*args):
                  marks=[requires_cuda]),
 ]
 
+FUNSOR_EXAMPLES = [
+    xfail_param('contrib/funsor/hmm.py --num-steps=1 --truncate=10 --model=0 --funsor',
+                reason="unreproducible recursion error on travis?"),
```
eb8680, line 213: I'm not sure what's going on here, this model worked before and now there's a `RecursionError` in the stack trace on Travis that I can't reproduce locally.

tests/test_examples.py
```diff
@@ -208,6 +208,27 @@ def xfail_jit(*args):
                  marks=[requires_cuda]),
 ]
 
+FUNSOR_EXAMPLES = [
```
fritzo, line 211: Why not divvy these into `CPU_EXAMPLES`, `CUDA_EXAMPLES` and `JIT_EXAMPLES`? It looks like you're breaking `test_coverage`

fritzo: Alternatively you could update `test_coverage`

examples/contrib/funsor/hmm.py
```diff
@@ -0,0 +1,657 @@
...
+    parser.add_argument("-b", "--batch-size", default=8, type=int)
+    parser.add_argument("-d", "--hidden-dim", default=16, type=int)
+    parser.add_argument("-nn", "--nn-dim", default=48, type=int)
+    parser.add_argument("-nc", "--nn-channels", default=2, type=int)
+    parser.add_argument("-lr", "--learning-rate", default=0.05, type=float)
+    parser.add_argument("-t", "--truncate", type=int)
+    parser.add_argument("-p", "--print-shapes", action="store_true")
+    parser.add_argument("--seed", default=0, type=int)
+    parser.add_argument('--cuda', action='store_true')
+    parser.add_argument('--jit', action='store_true')
```
fritzo, line 637: I think you'll need to either remove these options or add tests for them?

fritzo (approved):
feel free to merge as soon as tests pass

eb8680:
Ok, should finally be good to go...

fritzo (approved):

PR #2617, opened 2020-09-03
Fix LKJ sample shape bug

### LINKED ISSUE #2615: [bug] Samples from LKJCorrCholesky have the wrong shape
### Issue Description
The batch and sample shapes for `LKJCorrCholesky` are swapped when sampling.

### Environment
OS: Ubuntu 16.04
Python: 3.7.2
Torch version: 1.6.0
Pyro version: 1.4.0

### Code Snippet
Minimal example:

```python
>>> dist = pyro.distributions.LKJCorrCholesky(d=3, eta=torch.ones(())).expand(12)
>>> # batch shape and event shape are as you'd expect
>>> dist.batch_shape
torch.Size([12])
>>> dist.event_shape
torch.Size([3, 3])
>>> # samples have correct shape when sample_shape=()
>>> dist.shape(())
torch.Size([12, 3, 3])
>>> dist.sample().shape
torch.Size([12, 3, 3])
>>> # samples have the wrong shape when sample_shape is non-unit
>>> dist.shape((4,))  # as expected
torch.Size([4, 12, 3, 3])
>>> dist.sample((4,)).shape
torch.Size([12, 4, 3, 3])
```

I think this line is the culprit: https://github.com/pyro-ppl/pyro/blob/7c2c22c10dffda8a33ffbd593cc8d58819959e40/pyro/distributions/lkj.py#L67

```python
        y = self._gen.sample(sample_shape=self.batch_shape + sample_shape).detach()
        z = y.mul(2).add(-1.0)
        return _vector_to_l_cholesky(z)
```

should be 

```python
        y = self._gen.sample(sample_shape=sample_shape + self.batch_shape).detach()
        z = y.mul(2).add(-1.0)
        return _vector_to_l_cholesky(z)
```

fritzo:
Fixes #2615 

Thanks @PaperclipBadger for diagnosing and suggesting a fix!

## Tested
- [x] added a regression test

### DIFF
--- a/pyro/distributions/lkj.py
+++ b/pyro/distributions/lkj.py
@@ -64,7 +64,8 @@ class LKJCorrCholesky(TorchDistribution):
         super().__init__(torch.Size(), torch.Size((d, d)), validate_args=validate_args)
 
     def sample(self, sample_shape=torch.Size()):
-        y = self._gen.sample(sample_shape=self.batch_shape + sample_shape).detach()
+        with torch.no_grad():
+            y = self._gen.sample(sample_shape=sample_shape + self.batch_shape)
         z = y.mul(2).add(-1.0)
         return _vector_to_l_cholesky(z)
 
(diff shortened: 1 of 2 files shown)


fehiepsi (approved):
Thanks for the fix, @fritzo !

Issue #2618, opened 2020-09-08
FR Use funsor distributions directly in Pyro models


fritzo:
This issue proposes a low-level refactoring of `pyro.distribution` so that the basic object is no longer a `TorchDistribution` but a `Funsor` (or possibly a derived `class FunsorDistribution(Funsor, TorchDistribution)`). This low-level refactoring could precede #2580 insofar as it would be compatible with existing Pyro inference algorithms, and should be forwards compatible with #2580.

The motivation is to allow small funsor computations inside Pyro models using the `pyro.factor()` statement.

## Motivating Example 1: Beta-Binomial conjugacy

It would be nice to perform small Funsor computations directly in a Pyro model, using string notation:
```diff
  def model(data):
-     probs = pyro.sample("probs", dist.Beta(2, 2))
-     pyro.sample("obs", dist.Binomial(10, probs), obs=data)
+     prior = dist.Beta(2, 2).log_prob("probs")
+     likelihood = dist.Binomial(10, "probs").log_prob(data)
+     pyro.factor("obs", (prior + likelihood).reduce(ops.logaddexp))
```
or even simply
```diff
+     pyro.factor("obs", prior + likelihood)
```
where `pyro.factor` automatically calls `.reduce(ops.logaddexp)`, or could call `to_data()` which would in turn call `sum_product` which would `.reduce()` with either `ops.add` for plates or `ops.logaddexp` for non-plates.

## Motivating Example 2: Likelihoods vs Distributions

In some cases there is opportunity for significant caching of computation in a likelihood, e.g.
```py
class Model(PyroModule):
    def __init__(self, ...):
        data = torch.tensor(...)
        eager_param = torch.tensor(...)
        self.likelihood = lambda lazy_param: MyDistribution(eager_param, lazy_param).log_prob(data)
    def forward(self, ...):
        pyro.factor("likelihood", self.likelihood(lazy_param=torch.tensor(...))
```
A real world example is Pyro's [CoalescentTimesWithRate](http://docs.pyro.ai/en/stable/distributions.html#pyro.distributions.CoalescentTimesWithRate) distribution and the [CoalescentRateLikelihood](http://docs.pyro.ai/en/dev/contrib.epidemiology.html#pyro.distributions.CoalescentRateLikelihood). These two classes would most naturally be unified as a single `Funsor` class with different `eager` reduction rules depending on which arguments were lazy: (`coal_times` versus `rate_grid`)

## What would be required?
[...]

PR #2620, opened 2020-09-09
Implement a funsor-based poutine.collapse

fritzo:
Addresses #2618 
~~Blocked by https://github.com/pyro-ppl/funsor/pull/361~~
pair coded with @eb8680

This attempts to implement a `poutine.collapse()` using Funsors, in a way that is compatible with Pyro 1.0-style inference algorithms. This is a possible intermediate step towards Pyro 2.0-style inference algorithms as in `contrib.funsor` and #2580.

### DIFF
--- a/pyro/poutine/collapse_messenger.py
+++ b/pyro/poutine/collapse_messenger.py
@@ -16,6 +16,12 @@ except ImportError:
 
 
 class CollapseMessenger(TraceMessenger):
+    """
+    EXPERIMENTAL Collapses all sites in the context by lazily sampling and
+    attempting to use conjugacy relations. If no conjugacy is known this will
+    fail. Code using the results of sample sites must be written to accept
+    Funsors rather than Tensors. This requires ``funsor`` to be installed.
+    """
     def __init__(self, *args, **kwargs):
         import funsor
         funsor.set_backend("torch")
--- a/pyro/poutine/collapse_messenger.py
+++ b/pyro/poutine/collapse_messenger.py
@@ -1,6 +1,11 @@
-import pyro.distributions as dist
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import pyro
+from pyro.poutine.util import site_is_subsample
 
 from .runtime import _PYRO_STACK
+from .trace_messenger import TraceMessenger
 
 # TODO Remove import guard once funsor is a required dependency.
 try:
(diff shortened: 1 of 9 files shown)


martinjankowiak:
what are the semantics on this? it tries to integrate everything out in the collapse context?

eb8680:
> what are the semantics on this? it tries to integrate everything out in the collapse context?

It uses Funsor to lazily evaluate and analytically integrate out all latent variables in the context, hides them from any enclosing effect handlers, and introduces a single new `pyro.factor` statement containing the marginal likelihood of all observations in the context.  This final `factor` site is visible to and compatible with existing inference algorithms like HMC or `TraceEnum_ELBO`.

fritzo:
@eb8680 once we get the initial version of this working, let's try to use Funsor adjoints to sample and replay from the collapsed distributions, as @neerajprad implemented in [UncollapseConjugateMessenger](https://github.com/pyro-ppl/pyro/blob/1.4.0/pyro/contrib/conjugate/infer.py#L115) for reading the collapsed values during HMC.

pyro/poutine/collapse_messenger.py
```diff
@@ -0,0 +1,77 @@
...
+
+# TODO Remove import guard once funsor is a required dependency.
+try:
+    import funsor
+    from funsor.terms import Funsor
+except ImportError:
+    pass
+
+
+class CollapseMessenger(TraceMessenger):
```
fritzo, line 18: add docs

fritzo:
@eb8680 I believe the new tests are still not exercised on travis due to `pytest.imporortskip("funsir")`. I think those should be moved to the `funsor` stage; we'll need to mark the stage of those tests.

pyro/poutine/collapse_messenger.py
eb8680: What happened to the docstring here?

fritzo: oops, just pushed.

fritzo:
Well it looks like the tests are now [passing on travis](https://api.travis-ci.com/v3/job/386445321/log.txt):
```
...
tests/infer/test_gradient.py::test_collapse_beta_binomial tensor(0.5234) tensor(0.5234)
tensor(0.0356) tensor(0.0356)
PASSED
tests/infer/test_valid_models.py::test_collapse_beta_bernoulli XFAIL
tests/infer/test_valid_models.py::test_collapse_beta_binomial PASSED
tests/infer/test_valid_models.py::test_collapse_beta_binomial_plate XFAIL
```

eb8680 (approved):

PR #2621, opened 2020-09-10
Fix enumeration notebook tiny problem

xidulu:
The Dirhchlet prior should have length identical to the number of components instead of 3.


fritzo (approved):
Thanks for the fix!

PR #2624, opened 2020-09-13
[easy] fixed typo in error message

jamestwebber:
Noticed this while debugging


eb8680 (approved):
Thanks for fixing this!

PR #2626, opened 2020-09-14
Adding CVAE tutorial

ucals:
This PR contributes with a tutorial that implements the paper [Learning Structured Output Representation using Deep Conditional Generative Models](http://papers.nips.cc/paper/5775-learning-structured-output-representation-using-deep-conditional-generati), which introduced Conditional Variational Auto-encoders in 2015.
I implemented the paper's proof of concept using Pyro, and was able to achieve very similar qualitative and quantitative results to the ones shown in the paper.
Cheers!

View [notebook](https://nbviewer.jupyter.org/github/ucals/pyro/blob/cvae_tutorial/tutorial/source/cvae.ipynb)


karalets:
Hey,

This is cool, could you please double-check if there already is a model
called cvae for causal vae in the codebase? I remember Fritz adding sth
like that a while back.
If yes, it might be worth keeping the naming of those abbreviations
different for clarity.

Cheers

ucals:
Hi @karalets, thanks! The model you mentioned is called CEVAE, with an extra E.
This model is called CVAE, without the E. I could rename it to DCGM for Deep Conditional Generative Model, or just CGM for Conditional Generative Model, as the authors also use these names in the paper. However, they use CVAE more often. Pls let me know which is better, and whether I should rename it.. Cheers!

karalets:
Thanks for the clarification, I will leave those naming choices to others,
I just wanted to flag the potential for confusion which imho you clarified!

fehiepsi:
Haha, I also thought I saw the name before. 😄

fehiepsi:
@ucals Could you fix the lint issue?

fritzo:
Ha ha i also thought i saw the name before 😆. CVAE is a fine name though 👍

cc @stefanwebb who has implemented some conditional normalizing flows that
can be used in VAEs

examples/cvae/cvae.py
```diff
@@ -0,0 +1,188 @@
...
+            pyro.sample("z", dist.Normal(loc, scale).to_event(1))
+
+    def save(self, model_path):
+        torch.save({
+            'prior': self.prior_net.state_dict(),
+            'generation': self.generation_net.state_dict(),
+            'recognition': self.recognition_net.state_dict()
+        }, model_path)
+
+    def load(self, model_path, map_location=None):
```
martinjankowiak, line 123: @fritzo there's probably a better way to do this? use `PyroModule`?

examples/cvae/main.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def main(args):
+    device = torch.device("cuda:0" if torch.cuda.is_available() and args.cuda
+                          else "cpu")
+    results = []
+    columns = []
+
+    for num_quadrant_inputs in args.num_quadrant_inputs:  # [1, 2, 3]
+        maybes = 's' if num_quadrant_inputs > 1 else ''
```
martinjankowiak, line 20: what's `maybes` doing? could this use some comments?

examples/cvae/mnist.py
```diff
@@ -0,0 +1,83 @@
...
+
+class ToTensor:
+    def __call__(self, sample):
+        sample['original'] = functional.to_tensor(sample['original'])
+        sample['digit'] = torch.as_tensor(np.asarray(sample['digit']),
+                                          dtype=torch.int64)
+        return sample
+
+
+class MaskImages:
```
martinjankowiak, line 36: maybe this could use a few comments?

examples/cvae/cvae.py
```diff
@@ -0,0 +1,188 @@
...
+    def __init__(self, z_dim, hidden_1, hidden_2):
+        super().__init__()
+        self.fc1 = nn.Linear(784, hidden_1)
+        self.fc2 = nn.Linear(hidden_1, hidden_2)
+        self.fc31 = nn.Linear(hidden_2, z_dim)
+        self.fc32 = nn.Linear(hidden_2, z_dim)
+        self.relu = nn.ReLU()
+
+    def forward(self, x, y):
+        # put x and y together in the same image for simplification
```
martinjankowiak, line 24: what exactly is going on here?

examples/cvae/baseline.py
```diff
@@ -0,0 +1,117 @@
...
+                if i % 10 == 0:
+                    bar.set_postfix(loss='{:.2f}'.format(running_loss / num_preds),
+                                    early_stop_count=early_stop_count)
+
+            epoch_loss = running_loss / dataset_sizes[phase]
+            # deep copy the model
+            if phase == 'val':
+                if epoch_loss < best_loss:
+                    best_loss = epoch_loss
+                    best_model_wts = copy.deepcopy(baseline_net.state_dict())
```
martinjankowiak, line 102: is this sort-of-complex training mechanism necessary to get decent performance? if not it might be preferable to simplify?

martinjankowiak (commented):
this is great thanks for the tutorial!

comments/questions/suggestions:

- "maximize the conditional log-likelihood" => "maximize the conditional marginal log-likelihood"
- to me it seems some of your comments about neural networks e.g. "Deterministic neural networks are not suitable in modeling a distribution with multiple modes" are a bit strong and perhaps unnecessary for a tutorial? 
- "is the reparametrized latent variable product of the reparametrization trick" -> "is a Gaussian latent variable product"
- are you reporting log likelihoods or negative log likelihoods in the table?
- probably better to have less **bold text**?

ucals:
Thanks guys! @fehiepsi , I fixed the linting problem (making print compatible with Python 3.5) and also added CVAE smoke tests to `test_examples.py`, to avoid eroding test coverage.

Regarding Pyro specific code, it is in `examples/cvae/cvae.py`, lines 53 to 130. Before writing this tutorial, I've read almost all examples, so I think the code conform pretty much with the existing codebase. There is, though, 2 things I could have implemented in different ways, that I didn't see anywhere else (maybe I missed):

- As explained in the paper, a CVAE is composed by 4 models: a prior net, a recognition net, a generation net, and also a baseline net. The first 3 models have trainable parameters, while the last does not. I could implement that in 2 ways: i) using `with torch.no_grad()` in the `model` while computing `y_hat` (my choice, line 74), or ii) doing a loop in the CVAE `init` through all baseline net parameters, setting `param.requires_grad = False`. I preferred the former option as it could be done with one line instead of 2, and for my taste looked cleaner. But I couldn't find anywhere else in the codebase what should be the preferred way to deal with this situation, where some model parameters should be changed in the optimization process while other shouldn't

- I used a `loc.detach()` in line 97 to ensure I can do `.numpy()` when I sample from the predictive distribution later... I could also have achieved the same effect by using the context `with torch.set_grad_enabled(ys is not None)` while computing loc with the generation net, or by detaching later on (when using predictive). I preferred the fist choice as it took zero additional lines, while the others would have taken 1-2 additional lines

What do you think?

examples/cvae/baseline.py
```diff
@@ -0,0 +1,114 @@
...
+        self.reduction = reduction
+        self.masked_with = masked_with
+
+    def forward(self, input, target):
+        target = target.view(input.shape)
+        loss = F.binary_cross_entropy(input, target, reduction='none')
+        loss[target == self.masked_with] = 0
+        if self.reduction == 'sum':
+            return loss.sum()
+        elif self.reduction == 'mean':
```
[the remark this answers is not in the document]
ucals: You are right, I played with this functionality in my first implementations, but never used later... Removing it!

examples/cvae/baseline.py
```diff
@@ -0,0 +1,114 @@
...
+            bar = tqdm(dataloaders[phase],
+                       desc=f'NN Epoch {epoch} {phase}'.ljust(20))
+            for i, batch in enumerate(bar):
+                inputs = batch['input'].to(device)
+                outputs = batch['output'].to(device)
+
+                # zero the parameter gradients
+                optimizer.zero_grad()
+
+                # forward
```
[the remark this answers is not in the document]
ucals: Done!

examples/cvae/main.py
ucals: This just make sure the final table has headers ['1 quadrant', '2 quadrants', '3 quadrants'] (the first is singular, the 2nd and 3rd are plural). But I added a comment clarifying..

examples/cvae/mnist.py
ucals: Done!

examples/cvae/cvae.py
ucals: Let's say we are training the model using 1 quadrant of a MNIST digit as input X (bottom left), which means the remaining quadrants (top left, top right, bottom right) are the target output Y to be predicted. This means:
- In the input X, all pixels that are not in the bottom left quadrant are set to -1
- In the output Y, all pixels that are not in the top left, top right and bottom right quadrants (i.e. the pixels in the bottom left quadrant) are set to -1
- So X, and Y are complementary: at training time, they form perfectly an original MNIST digit, while at inference time, they form an approximate MNIST digit (at inference time, Y is not provided: as explained in the paper, at inference time we use the prior net to approximate Y_hat, using it instead of Y)

Lines 25 and 26 just put X and Y (or Y_hat) together in a single 28 x 28 image, to be fed either into the prior net or into the recognition net. 

There's another way to interpret & implement the paper: Instead of putting X and Y (or Y_hat) together as I did, I could have:
- Removed all pixels marked with -1 from X and Y (or Y_hat)
- Concatenated the flattened tensors and carried on

First choice seemed simpler & more elegant to me... But I can do the 2nd way as well, no problem :) (need to test it, but in theory it should work as well as the first approach)

examples/cvae/baseline.py
ucals: This is actually required for the early stopping, which is how the authors trained the models. These `if`'s are, imho, the simplest way to implement early stopping mechanism in PyTorch. They came from this PyTorch tutorial (https://pytorch.org/tutorials/beginner/transfer_learning_tutorial.html). 
Personally, I find a bit ugly as well, but I don't know how else to implement early stopping in PyTorch in as few lines as possible... pls advise!

martinjankowiak: i guess i'm asking if you still got ok performance without early stopping, e.g. if you just chose a reasonable number of epochs to train

ucals:
> this is great thanks for the tutorial!
> 
> comments/questions/suggestions:
> 
> * "maximize the conditional log-likelihood" => "maximize the conditional marginal log-likelihood"
> * to me it seems some of your comments about neural networks e.g. "Deterministic neural networks are not suitable in modeling a distribution with multiple modes" are a bit strong and perhaps unnecessary for a tutorial?
> * "is the reparametrized latent variable product of the reparametrization trick" -> "is a Gaussian latent variable product"
> * are you reporting log likelihoods or negative log likelihoods in the table?
> * probably better to have less **bold text**?

Thanks for your comments, @martinjankowiak !

1) Although the authors omitted the word *marginal*, I added them as you suggested, changing to *maximize the conditional marginal log-likelihood*
2) Also a choice from the authors, but changed as well, removing this sentence
3) Ok, better!
4) Negative
5) You are right, removed most of the bold text! Thanks!

examples/cvae/baseline.py
ucals: Yes, if I set the number of epochs to a reasonably high value, and different values for the NN and CVAE, yes I can get ok performance without early stopping. The NN usually stops improving around 20 epochs, while the CVAE trains for (way) longer, and if I increase the early stop patience, I could improve the results even further...

martinjankowiak: ok well as far as i'm concerned it's up to you if you want to keep the early stopping code. for simplicity it might be better to remove it but it's up to you

martinjankowiak (approved):
looks good unless @fritzo has any suggestions on how to improve model saving/loading?

one last nit: "Gaussian latent variable product" -> "Gaussian latent variable"
(accidentally included "product" in my previous review)

thanks again @ucals !

martinjankowiak:
@ucals i also wanted to suggest that after the table with results you might want to explicitly point out that "lower is better"

ucals:
Thanks for the inputs @martinjankowiak ! 
I followed your suggestions, changing to "Gaussian latent variable" and "lower is better"... Regarding removing the early stop code to simplify, I did some exercises trying to find a single magical number of epochs that could yield good performance for 1, 2 and 3 quadrants as inputs, but there isn't one... In order to remove the early stop code, I would have to use different numbers of epochs depending on the use case... we would simplify the training code, but complicate the `main` code... so, I left as it is...

Cheers!

examples/cvae/cvae.py
fritzo: I think `PyroModule` would be overkill for this model. However @ucals is it possible to simplify `save`/`load` as follows:
```py
def save(self, model_path):
    torch.save(self, model_path)

def load(self, model_path, map_location=None):
    loaded = torch.load(model_path, map_location=map_location)
    self.load_state_dict(loaded.state_dict())
    self.eval()
```
And if that works, I don't see why you really need custom `.save()`, `.load()` methods at all; user code could simply
```py
torch.save(model_path, cvae)
...
cvae = torch.load(model_path)
cvae.eval()
```

fritzo:
Beautiful tutorial! I have only a couple minor nits
- [ ] It looks like code in the notebook is entered in code cells rather than ```python ... ``` code blocks. It might be clearer to convert those code cells to text cells to make it clear that the notebook is not intended to be executed (e.g. it is missing imports). But feel free to keep as is if you prefer the current highlighting mode.
- [ ] Consider removing the `.save()` and `.load()` methods so as to focus more attention on the ML-relevant parts of your tutorial.

ucals:
> Beautiful tutorial! I have only a couple minor nits
> 
> * [X]  It looks like code in the notebook is entered in code cells rather than `python ... ` code blocks. It might be clearer to convert those code cells to text cells to make it clear that the notebook is not intended to be executed (e.g. it is missing imports). But feel free to keep as is if you prefer the current highlighting mode.
> * [X]  Consider removing the `.save()` and `.load()` methods so as to focus more attention on the ML-relevant parts of your tutorial.

Yes!! Converting code cells to text cells made it much better! Also, the `.save()` and `.load()` methods were completely unnecessary! A simple `torch.save(cvae_net.state_dict(), model_path)` and `cvae_net.load_state_dict(torch.load(model_path))` sufficed! 

Thanks @fritzo !!

examples/cvae/cvae.py
ucals: You're right! These custom methods were completely unnecessary, just tested! Thanks for the tip!

martinjankowiak (approved):

fritzo (approved):

eb8680:
Great work, @ucals!  We really appreciate the effort you've put in here and your other PRs - high-quality tutorials and documentation are the most valuable kind of contribution to almost all open source projects, including Pyro.

I don't know if you're planning to continue contributing, but we certainly hope so - happy to discuss further here or over Slack or Zoom.  If you're looking for ideas, one thing I know many users have requested is an expanded version of our [amortized LDA example](http://pyro.ai/examples/lda.html) with careful explanations and results on a real dataset, based primarily on [this paper referenced in the example script](https://arxiv.org/pdf/1703.01488).

ucals:
> Great work, @ucals! We really appreciate the effort you've put in here and your other PRs - high-quality tutorials and documentation are the most valuable kind of contribution to almost all open source projects, including Pyro.
> 
> I don't know if you're planning to continue contributing, but we certainly hope so - happy to discuss further here or over Slack or Zoom. If you're looking for ideas, one thing I know many users have requested is an expanded version of our [amortized LDA example](http://pyro.ai/examples/lda.html) with careful explanations and results on a real dataset, based primarily on [this paper referenced in the example script](https://arxiv.org/abs/1703.01488).

Thanks @eb8680 ! First of all, congrats to you and all Pyro team: it's an amazing project!
Yes, I'm planning to continue contributing, and get more involved! Currently, there are 3 areas/papers that I'm keen to explore/implement:

- Zero-shot learning: [One-Shot Generalization in Deep Generative Models](https://arxiv.org/abs/1603.05106) or something else in this area
- Disentangled representation: [Deep Convolutional Inverse Graphics Network](https://arxiv.org/abs/1503.03167) or something else in this area
- Trajectory prediction: [Real Time Trajectory Prediction Using Deep Conditional Generative Models](https://arxiv.org/abs/1909.03895) (actually I want to try this approach in the [new huge dataset](https://arxiv.org/abs/2006.14480) Lyft has just released)

At some point, I'd like to propose something new & publish... But I'd love to get your thoughts on which order do you think I should tackle these 4 areas/papers (the 3 I mentioned and the one you mentioned).. if you can, I'd love to talk over Zoom at some point over next weeks, pls let me know your availability!

PS: btw, do you have any suggestions of introductory material on effect handlers/effectful programming?

Cheers!

eb8680:
> Yes, I'm planning to continue contributing, and get more involved!

That's great!  It sounds like you have a few directions in mind already, so feel free to disregard my LDA suggestion.  The trajectory prediction paper looks like a good next step.  A more recent and more challenging/realistic disentangled representation learning paper you might try to reproduce after that is ["GENESIS: Generative Scene Inference and Sampling with Object-Centric Latent Representations"](https://arxiv.org/abs/1907.13052).

> do you have any suggestions of introductory material on effect handlers/effectful programming?

The [Pyro effect handler tutorial](http://pyro.ai/examples/effect_handlers.html) in Pyro is a good place to start, and there are some links to papers at the end that are good entry points to the literature, especially the review/position paper ["Handlers in action"](http://homepages.inf.ed.ac.uk/slindley/papers/handlers.pdf).

> if you can, I'd love to talk over Zoom at some point over next weeks, pls let me know your availability!

Sure, maybe sometime next week or the week after, shoot me an email at ebingham@broadinstitute.org and we can try to find a time.

PR #2619, opened 2020-09-08
Forecast DLM tutorial

edwinnglabs:
I just write a DLM tutorial.  Please let me know if I need to fix anything.

See [notebook](https://nbviewer.jupyter.org/github/edwinnglabs/pyro/blob/forecast-dlm-test/tutorial/source/forecasting_dlm.ipynb)

## Tasks

- [x] #2636 get `HaarReparam` to work with `pyro.contrib.forecast`
- [x] #2632 make it easier to examine posterior samples of latent variables

PR #2632, opened 2020-09-18
Support tracing of latents in contrib.forecast

fritzo:
Addresses #2619 

@edwinnglabs requested the ability to sample and forecast latent variables from the posterior. Previously this could only be accomplished by complex `poutine` logic. After this PR the same can be accomplished in one line:
```py
posterior = poutine.trace(forecaster).get_trace(data, covariates, num_samples)
```
The crux was to add a `poutine.block()` around the guide invocation so that an enclosing `poutine.trace()` sees only the replayed model.

## Tested
- [x] added a new test

### DIFF
--- a/pyro/contrib/forecast/forecaster.py
+++ b/pyro/contrib/forecast/forecaster.py
@@ -3,6 +3,7 @@
 
 import logging
 from abc import ABCMeta, abstractmethod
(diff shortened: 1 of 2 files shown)


eb8680 (approved):
LGTM

PR #2633, opened 2020-09-19
AutoGuideList documentation refers to deprecated method

randommm:
AutoGuideList documentation refers to deprecated method `add`


fehiepsi (approved):
Thanks, @randommm !

PR #2635, opened 2020-09-20
Add a block_plate() helper

fritzo:
Blocking #2636

This PR adds a primitive `pyro.block_plate()` that temporarily blocks a specified enclosing `pyro.plate()` so that plate-global variables can be sampled. The primitive is implemented via a more general `block_messenger()` helper that temporarily swaps out a messenger with a no-op `Messenger()` in the `_PYRO_STACK`.

The motivating use case is to support reparametrization over plated dimensions in `pyro.reparam()`, where the auxiliary variable needs to be outside of the plate context. See #2636 for demonstration.

## Tested
- [x] added to test_valid_models.py

### DIFF
--- a/pyro/poutine/messenger.py
+++ b/pyro/poutine/messenger.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2017-2019 Uber Technologies, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
+from contextlib import contextmanager
 from functools import partial
 
 from .runtime import _PYRO_STACK
@@ -133,15 +134,15 @@ class Messenger:
         Process the message by calling appropriate method of itself based
         on message type. The message is updated in place.
         """
-        method_name = "_pyro_{}".format(msg["type"])
-        if hasattr(self, method_name):
-            return getattr(self, method_name)(msg)
+        method = getattr(self, "_pyro_{}".format(msg["type"]), None)
+        if method is not None:
+            return method(msg)
         return None
 
     def _postprocess_message(self, msg):
-        method_name = "_pyro_post_{}".format(msg["type"])
-        if hasattr(self, method_name):
-            return getattr(self, method_name)(msg)
+        method = getattr(self, "_pyro_post_{}".format(msg["type"]), None)
+        if method is not None:
+            return method(msg)
         return None
 
     @classmethod
@@ -200,3 +201,27 @@ class Messenger:
             pass
 
         return fn
+
+
+@contextmanager
+def mute_messengers(predicate):
+    """
+    Context manager to temporarily remove matching messengers from the _PYRO_STACK.
+    Note this does not call the ``.__exit__()`` and ``.__enter__()`` methods.
+
+    This is useful to selectively block enclosing handlers.
+
+    :param callable predicate: A predicate mapping messenger instance to boolean.
+        This mutes all messengers ``m`` for which ``bool(predicate(m)) is True``.
+    :yields: A list of matched messengers that are muted.
+    """
+    muted = {}
+    try:
+        for i, messenger in enumerate(_PYRO_STACK):
+            if predicate(messenger):
+                muted[i] = messenger
+                _PYRO_STACK[i] = Messenger()  # trivial messenger
+        yield list(muted.values())
+    finally:
+        for i, messenger in muted.items():
+            _PYRO_STACK[i] = messenger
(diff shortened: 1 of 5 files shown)


pyro/poutine/messenger.py
```diff
@@ -133,15 +134,15 @@ def _process_message(self, msg):
         Process the message by calling appropriate method of itself based
         on message type. The message is updated in place.
         """
-        method_name = "_pyro_{}".format(msg["type"])
-        if hasattr(self, method_name):
-            return getattr(self, method_name)(msg)
+        method = getattr(self, "_pyro_{}".format(msg["type"]), None)
+        if method is not None:
+            return method(msg)
```
fritzo, line 139: This and the below change are micro-optimizations included in this PR to prevent any performance regression.

eb8680:
`pyro.unplate` seems reasonable for the use case in #2636 but I can imagine it becoming a foot-gun in the hands of users, especially when interacting with more complicated plate/enumeration structures. 

Would it be safer to expose `unplate` at individual sample sites as an `infer` option e.g. `sample(..., infer={"unplate": [...], ...})`, and perhaps require that those sites be marked as auxiliary variables e.g. `site["infer"]["is_auxiliary"] == True`?

fritzo:
Agreed that this can be a foot gun. How about we move it to a helper in `plate_messenger.py` and mark it experimental for now? I do prefer the composable syntax of context managers rather than the more manual `infer={...}` syntax or the more circuitous `@infer_config` syntax.

fritzo:
Re: interaction with enumeration: this should be fully compatible with enumeration.
**Proof:** any nested combination of `plate` and `unplate` can be normalized to a straight-line unplate-free form where each plate is created in the beginning and each sample statement is in a specific plate context. For example this function
```py
def model_1():
    a = pyro.sample("a", ...)
    with pyro.plate("plate", 2):
        model_1_1()

def model_1_1():
    b = pyro.sample("b", ...)
    with pyro.unplate("plate"):
        model_1_2()

def model_1_2():
    c = pyro.sample("c", ...)
```
can be normalized as the straight-line
```py
def model_2():
    plate = pyro.plate("plate")

    a = pyro.sample("a", ...)
    with plate:
        b = pyro.plate("b", ...)
    c = pyro.sample("c", ...)
```

eb8680:
>  How about we move it to a helper in `plate_messenger.py` and mark it experimental for now?

Sure, that seems like a reasonable compromise.

pyro/poutine/plate_messenger.py
```diff
@@ -19,3 +22,53 @@ def __enter__(self):
...
+    def predicate(messenger):
+        if not isinstance(messenger, PlateMessenger):
+            return False
+        if name is not None:
+            return messenger.name == name
+        if dim is not None:
+            return messenger.dim == dim
+
+    with mute_messengers(predicate) as matches:
+        if len(matches) != 1:
```
eb8680, line 72: When would `len(matches) > 1`?

tests/infer/test_valid_models.py
```diff
@@ -859,6 +860,69 @@ def guide():
...
+def test_unplate_name_ok():
+
+    def model():
+        a = pyro.sample("a", dist.Normal(0, 1))
+        assert a.shape == ()
+        with pyro.plate("plate", 2):
+            b = pyro.sample("b", dist.Normal(0, 1))
+            assert b.shape == (2,)
+            with unplate("plate"):
+                c = pyro.sample("c", dist.Normal(0, 1))
```
eb8680, line 872: The non-intuitive behavior I was concerned about is when the distribution here incorrectly depends on an enumerated variable inside `"plate"`. Do you think our existing enumeration checking logic in `TraceEnum_ELBO` is sufficient to help someone figure out that an `unplate` is breaking tractability?

eb8680 (commented):
Code and tests look good, just a couple of questions

pyro/poutine/plate_messenger.py
fritzo: I guess `len(matches) in {0,1}` as written. It would be easy to generalize `name` to a regular expression, in which case this check would be more fully exercised.

tests/infer/test_valid_models.py
fritzo: Yes, I believe any break in tractability possible with `block_plate` had already been addressed when we introduced reusable plates.

eb8680 (approved):

fritzo:
Thanks for reviewing!

PR #2636, opened 2020-09-20
Support HaarReparam in contrib.forecast

fritzo:
Addresses #2619

This adds support for `HaarReparam` and `DiscreteCosineReparam` in `pyro.contrib.forecast`, building on the batch dimension support in #2635. These reparametrizers are enabled by a kwarg to `Forecaster` and `HMCForecaster`.

These supersede the older `dct_gradients`+`DCTAdam` which attempt to leverage time-frequency transforms in the optimizer rather than in the variational distribution. It is advantageous to use these transforms at the higher level of distributions because this allows covariance structure in guides such as `AutoNormal` and `AutoLowRankMultivariateNormal`; we have seen that this covariance structure is beneficial during experiments with `contrib.epidemiology`.

## Tested
- [x] smoke tests for `Forecaster`
- [x] smoke tests for `HMCForecaster`
- [x] validate accuracy improvements: I have updated the [forecasting_i](https://nbviewer.jupyter.org/github/pyro-ppl/pyro/blob/forecast-haar/tutorial/source/forecasting_i.ipynb) notebook to use `time_reparam="dct"`, which seems to improve accuracy a bit:

<details>

### Before
image

### After
image

### Before
image

### After
image

</details>


pyro/infer/reparam/discrete_cosine.py
```diff
@@ -31,6 +31,8 @@ class DiscreteCosineReparam(UnitJacobianReparam):
         noise; when -1 this transforms violet noise to white noise; etc. Any
         real number is allowed. https://en.wikipedia.org/wiki/Colors_of_noise.
     """
-    def __init__(self, dim=-1, smooth=0.):
+    def __init__(self, dim=-1, smooth=0., *,
+                 experimental_allow_batch=False):
```
martinjankowiak, line 35: does the new flag need a docstring?

fritzo: no, I think it is fine undocumented because it is experimental. This kwarg simply controls whether the experimental new behavior is allowed.

fritzo:
any other review comments?

martinjankowiak (approved):

PR #2638, opened 2020-09-21
Fix model-directed subsampling in AutoGuide

fritzo:
This implements @fehiepsi's fix to `AutoGuide` from https://github.com/pyro-ppl/numpyro/pull/747

## Tested
- [x] added a regression test adapted from @fehiepsi's example

### DIFF
--- a/tests/infer/test_autoguide.py
+++ b/tests/infer/test_autoguide.py
@@ -697,6 +697,37 @@ def test_predictive(auto_class):
     assert len(samples) == len(samples_deser)
 
 
+@pytest.mark.parametrize("auto_class", [AutoDelta, AutoNormal])
+def test_subsample_model(auto_class):
+
+    def model(x, y=None, batch_size=None):
+        loc = pyro.param("loc", lambda: torch.tensor(0.))
+        scale = pyro.param("scale", lambda: torch.tensor(1.),
+                           constraint=constraints.positive)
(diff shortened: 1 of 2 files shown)


tests/infer/test_autoguide.py
```diff
@@ -697,6 +697,37 @@ def forward(self, x, y=None):
...
+
+    def model(x, y=None, batch_size=None):
+        loc = pyro.param("loc", lambda: torch.tensor(0.))
+        scale = pyro.param("scale", lambda: torch.tensor(1.),
+                           constraint=constraints.positive)
+        with pyro.plate("batch", len(x), subsample_size=batch_size):
+            batch_x = pyro.subsample(x, event_dim=0)
+            batch_y = pyro.subsample(y, event_dim=0) if y is not None else None
+            mean = loc + scale * batch_x
+            sigma = pyro.sample("sigma", dist.LogNormal(0., 1.))
```
fritzo, line 711: This is where the test failed prior to @fehiepsi's fix. It seems that local latent variables are needed to exercise the bug.

fehiepsi (approved):
LGTM! Thanks for adding the regression test.

tests/infer/test_autoguide.py
fehiepsi: I guess without this local latent variable, we can also test if different runs return different outputs.

PR #2640, opened 2020-09-24
Optimize logmatmulexp() and friends

fritzo:
This adds a tiny optimization to the numerical stabilizing logic in `logmatmulexp()` used in `DiscreteHMM` and enumeration.

The observation is that the numerical stabilizing `shift` does not need gradients, since we simply add and subtract it. Detaching `shift` leads to a small speedup, for example `python -O examples/hmm.py -m 7 -n 500 --jit` sees a ~15% speedup:

| | time |
|---|---|
| before | 16.9 sec | 
| after | 14.2 sec |

## Tested

### DIFF
--- a/pyro/distributions/hmm.py
+++ b/pyro/distributions/hmm.py
@@ -39,8 +39,8 @@ def _logmatmulexp(x, y):
     Numerically stable version of ``(x.log() @ y.log()).exp()``.
     """
     finfo = torch.finfo(x.dtype)  # avoid nan due to -inf - -inf
-    x_shift = x.detach().max(-1, keepdim=True).values.clamp(min=finfo.min)
-    y_shift = y.detach().max(-2, keepdim=True).values.clamp(min=finfo.min)
+    x_shift = x.detach().max(-1, keepdim=True).values.clamp_(min=finfo.min)
+    y_shift = y.detach().max(-2, keepdim=True).values.clamp_(min=finfo.min)
     xy = safe_log(torch.matmul((x - x_shift).exp(), (y - y_shift).exp()))
     return xy + x_shift + y_shift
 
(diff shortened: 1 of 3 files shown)


pyro/ops/einsum/torch_log.py
```diff
@@ -29,12 +29,12 @@ def einsum(equation, *operands):
...
     exp_operands = []
     for dims, operand in zip(inputs, operands):
-        shift = operand
+        shift = operand.detach()
         for i, dim in enumerate(dims):
             if dim not in output:
                 shift = shift.max(i, keepdim=True)[0]
         # avoid nan due to -inf - -inf
-        shift = shift.clamp(min=torch.finfo(shift.dtype).min)
+        shift.clamp_(min=torch.finfo(shift.dtype).min)
```
eb8680, line 37: Looks like this in-place op is triggering a bunch of CI failures:
```
RuntimeError: unsupported operation: more than one element of the written-to tensor refers to a single memory location. Please clone() the tensor before performing the operation.
```

eb8680 (commented):
Nice, that's quite a speedup.

pyro/distributions/hmm.py
```diff
@@ -39,8 +39,8 @@ def _logmatmulexp(x, y):
     Numerically stable version of ``(x.log() @ y.log()).exp()``.
     """
     finfo = torch.finfo(x.dtype)  # avoid nan due to -inf - -inf
-    x_shift = x.max(-1, keepdim=True).values.clamp(min=finfo.min)
-    y_shift = y.max(-2, keepdim=True).values.clamp(min=finfo.min)
+    x_shift = x.detach().max(-1, keepdim=True).values.clamp_(min=finfo.min)
+    y_shift = y.detach().max(-2, keepdim=True).values.clamp_(min=finfo.min)
```
fehiepsi, line 43: We also do this in NumPyro for some mysterious numerical issues. This trick is helpful, in `logsumexp` too.

pyro/ops/einsum/torch_log.py
fritzo: Thanks, reverted!

eb8680 (approved):

PR #2641, opened 2020-09-24
Fix cyclic import in DistributionMeta

fritzo:
Blocking https://github.com/pyro-ppl/funsor/pull/366

This fixes a cyclic import bug introduced in #2620 that is blocking Funsor development.

@fehiepsi @eb8680 maybe a better long-term solution is to remove [the line](https://github.com/pyro-ppl/funsor/blob/e3eba15e2ad7d9db5dd5622c247e0aaa7e81e7dd/funsor/__init__.py#L47)
```py
set_backend(get_backend())
```
from funsor entirely and remove the `FUNSOR_BACKEND` environment variable. Let's discuss at the next Pyro dev meeting.

## Tested
- [x] locally verified this fixes funsor tests

### DIFF
--- a/pyro/distributions/distribution.py
+++ b/pyro/distributions/distribution.py
@@ -5,21 +5,16 @@ from abc import ABCMeta, abstractmethod
 
 from pyro.distributions.score_parts import ScoreParts
 
+COERCIONS = []
 
-# TODO Remove import guard once funsor is a required dependency.
-try:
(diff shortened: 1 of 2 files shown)


eb8680 (approved):

fritzo:
Ready to merge so I can retest https://github.com/pyro-ppl/funsor/pull/366 ?

PR #2642, opened 2020-09-24
add scANVI example from single cell transcriptomics

martinjankowiak:
references:

[1] "Harmonization and Annotation of Single-cell Transcriptomics data with Deep Generative Models,"
    Chenling Xu, Romain Lopez, Edouard Mehlman, Jeffrey Regier, Michael I. Jordan, Nir Yosef.

[2] https://github.com/YosefLab/scvi-tutorials/blob/50dd3269abfe0c375ec47114f2c20725a016736f/seed_labeling.ipynb

example output:
[scanvi.pdf](https://github.com/pyro-ppl/pyro/files/5287423/scanvi.pdf)

PR #2643, opened 2020-09-25
add support for logits parameterization in zero-inflated distributions

martinjankowiak:
goal: add `gate_logits` parameterization to zero-inflated distributions and improve numerical stability of `log_prob`

- existing tests (expanded to use `gate_logits` where appropriate) pass
- the `gate_logits` parameterization is numerically stable when used in #2642 (and allows me to remove 3 lines of code)


pyro/distributions/zero_inflated.py
```diff
@@ -111,10 +127,11 @@ class ZeroInflatedNegativeBinomial(ZeroInflatedDistribution):
     arg_constraints = {"gate": constraints.unit_interval,
                        "total_count": constraints.greater_than_eq(0),
                        "probs": constraints.half_open_interval(0., 1.),
-                       "logits": constraints.real}
+                       "logits": constraints.real,
```
martinjankowiak, line 130: not sure what the best naming convention is for ZINB. there are two types of probabilities, which muddies the watter

pyro/distributions/zero_inflated.py
```diff
@@ -17,15 +17,24 @@ class ZeroInflatedDistribution(TorchDistribution):
...
     :param torch.Tensor gate: probability of extra zeros given via a Bernoulli distribution.
+    :param torch.Tensor logits: logits of extra zeros given via a Bernoulli distribution.
     :param TorchDistribution base_dist: the base distribution.
     """
-    arg_constraints = {"gate": constraints.unit_interval}
+    arg_constraints = {"gate": constraints.unit_interval,
+                       "logits": constraints.real}
 
-    def __init__(self, gate, base_dist, validate_args=None):
+    def __init__(self, base_dist, gate=None, logits=None, validate_args=None):
```
fritzo, line 26: nit: can you add a `*,` so that when breaking backwards compatibility we at least do so loudly?
```diff
- gate=None, logits=None, validate_args=None):
+ gate=None, *, logits=None, validate_args=None):
```

pyro/distributions/zero_inflated.py
```diff
@@ -79,18 +92,20 @@ class ZeroInflatedPoisson(ZeroInflatedDistribution):
     A Zero Inflated Poisson distribution.
 
     :param torch.Tensor gate: probability of extra zeros.
+    :param torch.Tensor logits: probability of extra zeros.
```
fritzo, line 95: "logits of extra zeros" ?

pyro/distributions/zero_inflated.py
```diff
@@ -79,18 +92,20 @@ class ZeroInflatedPoisson(ZeroInflatedDistribution):
...
     :param torch.Tensor rate: rate of poisson distribution.
     """
     arg_constraints = {"gate": constraints.unit_interval,
-                       "rate": constraints.positive}
+                       "rate": constraints.positive,
+                       "logits": constraints.real}
     support = constraints.nonnegative_integer
 
-    def __init__(self, gate, rate, validate_args=None):
+    def __init__(self, rate, gate=None, logits=None, validate_args=None):
```
fritzo, line 103: ditto
```diff
- gate=None, logits=None, validate_args=None):
+ gate=None, *, logits=None, validate_args=None):
```

pyro/distributions/zero_inflated.py
fritzo: yeah. I guess `gate_logits` makes the most sense. WDYT of using that for other distributions too, for consistency?

martinjankowiak: but then do i also change `probs` -> `gate_probs`?

fritzo: no, the pairs are `(gate, gate_logits)` and `(probs, logits)`. We shouldn't rename any existing parameters.

martinjankowiak:
@fritzo ok made requested changes. please check the `__init__(..., *, ***)` logic is to your liking

pyro/distributions/zero_inflated.py
```diff
@@ -17,15 +17,24 @@ class ZeroInflatedDistribution(TorchDistribution):
...
+        if (gate is None) == (gate_logits is None):
+            raise ValueError("Either `gate` or `gate_logits` must be specified, but not both.")
         if base_dist.event_shape:
             raise ValueError("ZeroInflatedDistribution expected empty "
                              "base_dist.event_shape but got {}"
                              .format(base_dist.event_shape))
+
+        if gate_logits is not None:
+            epsilon = 1.0e-7
+            gate = epsilon + (1.0 - 2.0 * epsilon) * gate_logits.sigmoid()
```
fritzo, line 36: I don't see the purpose of this PR. From what I can see, this PR breaks existing interfaces and introduces a new parameter `gate_logits` that is converted back to a parameter `gate` that would have worked fine with the old interface. Moreover the ad hoc conversion logic does not agree with the `Categorical` treatment of `logits`.

I think a better solution would be to move the `gate_logits -> gate` stabilization logic into user code.

fritzo (changes requested):

pyro/distributions/zero_inflated.py
martinjankowiak: well the real goal here is for this distribution to be stable without the user having to worry about it overly much. if a user has 5 distributions in their code, each with 2 parameters, that's a lot of parameterizations to worry about. non-expert users don't know about these things, get nans, get frustrated, and walk away from pyro. if possible we should provide them with stable distributions that don't require thinking about numerical nitty gritty.

the proper way to do this is presumably to rewrite `log_prob` in such a way that it directly uses `gate_logics` when available. i've tried to do so here (see changes) but had a bit of trouble with numerical stability. i suppose it's necessary to add some epsilons. @fehiepsi do you have a suggestion for how this `log_prob` method should be implemented?

pyro/distributions/zero_inflated.py
```diff
@@ -41,9 +50,18 @@ def log_prob(self, value):
...
             self._validate_sample(value)
 
-        gate, value = broadcast_all(self.gate, value)
-        log_prob = (-gate).log1p() + self.base_dist.log_prob(value)
-        log_prob = torch.where(value == 0, (gate + log_prob.exp()).log(), log_prob)
+        if 'gate' in self.__dict__:
+            gate, value = broadcast_all(self.gate, value)
+            log_prob = (-gate).log1p() + self.base_dist.log_prob(value)
+            log_prob = torch.where(value == 0, (gate + log_prob.exp()).log(), log_prob)
+        else:
```
martinjankowiak, line 57: how should this branch look?

pyro/distributions/zero_inflated.py
fehiepsi: I have faced this issue with many distributions (e.g. in Poisson, when `rate_log` is small, `rate` will become `0`, and log_prob will throw `nan` or something similar), especially when working on M5 competition. At that time, I also tried your way, i.e. to use `foo_log/logits` directly and reimplement the log_prob at various places. Eventually, I found that doing some truncation as in @fritzo's [bounded_exp](https://github.com/pyro-ppl/Pyro-M5-Starter-Kit/blob/master/model2.py#L42) is good enough for my purpose.

If you want to use `gate_logits` directly here, I think it is useful to use `torch.nn.softplus` for `-(-gate_logits).exp().log1p()`, which is not stable for small negative `gate_logits`

```
log_gate = - torch.nn.functional.softplus(-gate)
```
Similar for `zero_log_prob`.

martinjankowiak: @fehiepsi great thanks that was super helpful! let me know if you think it looks reasonable

martinjankowiak:
@fritzo ok this is ready for review again

pyro/distributions/zero_inflated.py
```diff
@@ -41,9 +51,17 @@ def log_prob(self, value):
...
+            gate_logits, value = broadcast_all(self.gate_logits, value)
+            log_gate = -softplus(-gate_logits)
+            log_one_minus_gate = log_gate - gate_logits
+            log_prob = log_one_minus_gate + self.base_dist.log_prob(value)
+            zero_log_prob = softplus(log_prob - log_gate) + log_gate
```
fehiepsi, line 63: We can save one operator here if we compute `log_prob_minus_log_gate` first then plus with loggate to compute logprob.

martinjankowiak: great, thanks! like this right?

fehiepsi: yup

tests/distributions/test_zero_inflated.py
```diff
@@ -27,30 +29,36 @@ def test_zid_shape(gate_shape, base_shape):
 @pytest.mark.parametrize("rate", [0.1, 0.5, 0.9, 1.0, 1.1, 2.0, 10.0])
 def test_zip_0_gate(rate):
     # if gate is 0 ZIP is Poisson
-    zip_ = ZeroInflatedPoisson(torch.zeros(1), torch.tensor(rate))
+    zip1 = ZeroInflatedPoisson(torch.tensor(rate), gate=torch.zeros(1))
```
fritzo, line 32: for shapes to match this should be `gate=torch.tensor(0.)` or `gate=torch.zeros(())`

tests/distributions/test_zero_inflated.py
```diff
@@ -64,28 +72,38 @@ def test_zip_mean_variance(gate, rate):
 @pytest.mark.parametrize("probs", [0.1, 0.5, 0.9])
 def test_zinb_0_gate(total_count, probs):
     # if gate is 0 ZINB is NegativeBinomial
-    zinb_ = ZeroInflatedNegativeBinomial(
-        torch.zeros(1), total_count=torch.tensor(total_count), probs=torch.tensor(probs)
+    zinb1 = ZeroInflatedNegativeBinomial(
+        total_count=torch.tensor(total_count), gate=torch.zeros(1), probs=torch.tensor(probs)
```
fritzo, line 76: ditto: `gate=torch.tensor(0.)` or `gate=torch.zeros(())`

tests/distributions/test_zero_inflated.py
```diff
@@ -64,28 +72,38 @@ def test_zip_mean_variance(gate, rate):
...
 
 
 @pytest.mark.parametrize("total_count", [0.1, 0.5, 0.9, 1.0, 1.1, 2.0, 10.0])
 @pytest.mark.parametrize("probs", [0.1, 0.5, 0.9])
 def test_zinb_1_gate(total_count, probs):
     # if gate is 1 ZINB is Delta(0)
-    zinb_ = ZeroInflatedNegativeBinomial(
-        torch.ones(1), total_count=torch.tensor(total_count), probs=torch.tensor(probs)
+    zinb1 = ZeroInflatedNegativeBinomial(
+        total_count=torch.tensor(total_count), gate=torch.ones(1), probs=torch.tensor(probs)
```
fritzo, line 95: ditto: `gate=torch.tensor(1.)` or `gate=torch.ones(())`

tests/distributions/test_zero_inflated.py
```diff
@@ -64,28 +72,38 @@ def test_zip_mean_variance(gate, rate):
 @pytest.mark.parametrize("probs", [0.1, 0.5, 0.9])
 def test_zinb_0_gate(total_count, probs):
     # if gate is 0 ZINB is NegativeBinomial
-    zinb_ = ZeroInflatedNegativeBinomial(
-        torch.zeros(1), total_count=torch.tensor(total_count), probs=torch.tensor(probs)
+    zinb1 = ZeroInflatedNegativeBinomial(
+        total_count=torch.tensor(total_count), gate=torch.zeros(1), probs=torch.tensor(probs)
+    )
+    zinb2 = ZeroInflatedNegativeBinomial(
+        total_count=torch.tensor(total_count), gate_logits=torch.tensor(-99.9), probs=torch.tensor(probs)
```
fritzo, line 79: Why not test with `gate_logits=torch.tensor(-math.inf)`? You test with `+math.inf` below? Shouldn't the clamping logic handle `-math.inf`?

fritzo (commented):
- Numerical stabilization logic looks good.
- Tests look good except for minor nits.
- I'm still concerned with interface breakage. @martinjankowiak would you like to eliminate breakage, or would you like me to create a PR into this branch to try to avoid breakage?

tests/distributions/test_zero_inflated.py
fritzo: (I guess this was also incorrect in the original version of this PR)

tests/distributions/test_zero_inflated.py
martinjankowiak: that nan's out

fritzo (approved):
LGTM, thanks for your patience.

This is indeed a breaking change, but I am willing to call it a "fix to an interface bug". Let's just take care to note this in the next release notes under Breaking Changes.

PR #2646, opened 2020-09-28
Add an optional .finalize() method to CompartmentalModel

fritzo:
This adds a `CompartmentalModel.finalize()` method that can declare likelihoods that depend on the entire model time series. The motivation is a phylogeographic likelihood ([private repo](https://github.com/broadinstitute/pyro-cov/blob/09c661f5588a864847e33660a915b69ed7ae1401/pyrophylo/models.py#L85)) that depends on the entire `I` time series in a way that does not factorize.

## Tested
- [x] added a smoke test

### DIFF
--- a/pyro/contrib/epidemiology/compartmental.py
+++ b/pyro/contrib/epidemiology/compartmental.py
@@ -8,7 +8,7 @@ import re
 import warnings
 from abc import ABC, abstractmethod
 from collections import OrderedDict
-from contextlib import ExitStack
+from contextlib import ExitStack, contextmanager
 from functools import reduce
 from timeit import default_timer
 
@@ -42,6 +42,20 @@ def _require_double_precision():
         warnings.warn("CompartmentalModel is unstable for dtypes less than torch.float64; "
                       "try torch.set_default_dtype(torch.float64)",
                       RuntimeWarning)
+
+
+@contextmanager
+def _disallow_latent_variables(section_name):
+    if not is_validation_enabled():
+        yield
+        return
+
+    with poutine.trace() as tr:
+        yield
+    for name, site in tr.trace.nodes.items():
+        if site["type"] == "sample" and not site["is_observed"]:
+            raise NotImplementedError("{} contained latent variable {}"
+                                      .format(section_name, name))
 
 
 class CompartmentalModel(ABC):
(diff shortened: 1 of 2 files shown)


pyro/contrib/epidemiology/compartmental.py
```diff
@@ -253,6 +267,20 @@ def transition(self, params, state, t):
         """
         raise NotImplementedError
 
+    def finalize(self, params, state):
+        """
+        Optional method for likelihoods that depend on the entire time series.
```
eb8680, line 272: Clarify here that this is only for non-factorized likelihoods and should be avoided if possible in favor of `transition`?

pyro/contrib/epidemiology/compartmental.py
```diff
@@ -660,6 +688,7 @@ def heuristic(self, num_particles=1024, ess_threshold=0.5, retries=10):
                 continue
 
         # Select the most probable hypothesis.
+        # Note this ignores the .finalize() likelihood.
```
eb8680, line 691: To clarify, is `finalize` ignored during initialization? If so, mention in the `finalize` docstring?

tests/contrib/epidemiology/test_models.py
```diff
@@ -495,6 +496,56 @@ def test_regional_smoke(duration, forecast, options, algo):
     assert samples["I"].shape == (num_samples, duration + forecast, num_regions)
 
 
+class RegionalSIRModelWithFinalize(RegionalSIRModel):
+    def finalize(self, params, state):
+        I = state["I"]
+        I_mean = I.mean(dim=[-1, -2], keepdim=True).expand_as(I)
+        with self.region_plate, self.time_plate:
```
eb8680, line 503: Can you explain the semantics of the `plate`s here since things are no longer independent?

fritzo: I'm simply using plates for convenience, to avoid a few `.unsqueeze(-2)`s.

pyro/contrib/epidemiology/compartmental.py
fritzo: done

pyro/contrib/epidemiology/compartmental.py
fritzo: done

eb8680 (approved):
LGTM

PR #2647, opened 2020-10-01
Fix exception chaining all over the codebase

akihironitta:
I would like to suggest an enhancement in the way that Python 3's exception chaining is used.

## Motivation
As described in detail in [this article](https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with), exception chaining ([PEP 3134](https://www.python.org/dev/peps/pep-3134/)) can be used to make exceptions more user-friendly, and in that case, the syntax `raise .. from ..` needs to be used.

When `raise .. from ..` is used, there is a line saying `The above exception was the direct cause of the following exception` between tracebacks. However, when implicitly chaining exceptions (meaning when `raise .. from ..` is not used), the message will be `During handling of the above exception, another exception occurred`, which can confuse users.

## Description of the change
Below is the before/after the change using the following code block:
```python
from pyro.infer.discrete import _make_ring
_make_ring(temperature=123, cache={}, dim_to_size={})
```
Before the change:
```sh
  File "/home/nitta/Dropbox/home/src/projects/pyro/pyro/infer/discrete.py", line 25, in _make_ring
    return _RINGS[temperature](cache=cache, dim_to_size=dim_to_size)
KeyError: 123

  File "/home/nitta/Dropbox/home/src/projects/pyro/pyro/infer/discrete.py", line 27, in _make_ring
    raise ValueError("temperature must be 0 (map) or 1 (sample) for now")
ValueError: temperature must be 0 (map) or 1 (sample) for now
```
After the change:
```sh
  File "/home/nitta/Dropbox/home/src/projects/pyro/pyro/infer/discrete.py", line 25, in _make_ring
    return _RINGS[temperature](cache=cache, dim_to_size=dim_to_size)
KeyError: 123

The above exception was the direct cause of the following exception:

  File "/home/nitta/Dropbox/home/src/projects/pyro/pyro/infer/discrete.py", line 27, in _make_ring
    raise ValueError("temperature must be 0 (map) or 1 (sample) for now") from e
ValueError: temperature must be 0 (map) or 1 (sample) for now
```

If this PR seems reasonable, I'd be happy to submit another PR for fixing exception chaining all over the codebase. Let me know your thoughts!

### DIFF
--- a/pyro/contrib/autoname/named.py
+++ b/pyro/contrib/autoname/named.py
@@ -213,9 +213,9 @@ class Dict(dict):
     def __getitem__(self, key):
         try:
             return super().__getitem__(key)
-        except KeyError:
+        except KeyError as e:
             if self._name is None:
-                raise RuntimeError("Cannot access an unnamed named.Dict")
+                raise RuntimeError("Cannot access an unnamed named.Dict") from e
             value = Object("{}[{!r}]".format(self._name, key))
             super(Object, value).__setattr__(
                 "_set_value", lambda value: self.__setitem__(key, value))
--- a/pyro/contrib/forecast/util.py
+++ b/pyro/contrib/forecast/util.py
@@ -211,8 +211,8 @@ def prefix_condition(d, data):
     """
     try:
         return d.prefix_condition(data)
-    except AttributeError:
-        raise NotImplementedError("prefix_condition() does not suport {}".format(type(d)))
+    except AttributeError as e:
+        raise NotImplementedError("prefix_condition() does not suport {}".format(type(d))) from e
 
 
 @prefix_condition.register(dist.Independent)
--- a/pyro/infer/discrete.py
+++ b/pyro/infer/discrete.py
@@ -23,8 +23,8 @@ _RINGS = {0: MapRing, 1: SampleRing}
 def _make_ring(temperature, cache, dim_to_size):
     try:
         return _RINGS[temperature](cache=cache, dim_to_size=dim_to_size)
-    except KeyError:
-        raise ValueError("temperature must be 0 (map) or 1 (sample) for now")
+    except KeyError as e:
+        raise ValueError("temperature must be 0 (map) or 1 (sample) for now") from e
 
 
 class SamplePosteriorMessenger(ReplayMessenger):
--- a/pyro/infer/mcmc/logger.py
+++ b/pyro/infer/mcmc/logger.py
@@ -174,8 +174,8 @@ class TqdmHandler(logging.StreamHandler):
             msg = self.format(record)
             self.flush()
             tqdm.write(msg, file=sys.stderr)
-        except (KeyboardInterrupt, SystemExit) as e:
-            raise e
+        except (KeyboardInterrupt, SystemExit):
+            raise
         except Exception:
             self.handleError(record)
 
@@ -205,8 +205,8 @@ class MCMCLoggingHandler(logging.Handler):
                 self.progress_bar.update()
             else:
                 self.log_handler.handle(record)
-        except (KeyboardInterrupt, SystemExit) as e:
-            raise e
+        except (KeyboardInterrupt, SystemExit):
+            raise
         except Exception:
             self.handleError(record)
 
--- a/pyro/infer/reparam/neutra.py
+++ b/pyro/infer/reparam/neutra.py
@@ -64,10 +64,10 @@ class NeuTraReparam(Reparam):
             # Sample a shared latent.
             try:
                 self.transform = self.guide.get_transform()
-            except (NotImplementedError, TypeError):
+            except (NotImplementedError, TypeError) as e:
                 raise ValueError("NeuTraReparam only supports guides that implement "
                                  "`get_transform` method that does not depend on the "
-                                 "model's `*args, **kwargs`")
+                                 "model's `*args, **kwargs`") from e
 
             z_unconstrained = pyro.sample("{}_shared_latent".format(name),
                                           self.guide.get_base_dist().mask(False))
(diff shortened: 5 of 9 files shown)


fritzo (commented):
LGTM, feel free to add all fixes in this PR.

I believe we've previously refactored to use `.with_traceback()`. Do you understand the relationship between `.with_traceback()` and `raise ... from ...`?

fritzo:
Actually yeah let's keep this as a one big atomic change in a single PR.

akihironitta:
@fritzo Thank you for reviewing the change very quickly.

Not sure why `with_traceback()` is used at some places.

As long as I understand, we use `with_traceback()` to include the previous exception's traceback in the exception raised in `except` clause. In Python 2, we had to manually propagate the traceback of the previous exception obtained with `sys.exc_info()` or whatever. However, in Python 3, as PEP 3134 was accepted, information of the previous exception is automatically propagated to the new exception.

Also, as pointed out in #1963, there are currently duplicate stack traces in the error message. This is because `with_traceback()` propagates the previous traceback and Python 3 feature also propagates the previous exception automatically (which is called "implicit exception chaining" in PEP 3134).

Two options I came up with to fix duplicate stack traces are:
1. to remove `with_traceback()` and use `raise .. from ..`
2. to keep `with_traceback()` and use `raise .. from None` to unchain and throw away the previous exception.

I hope the following minimal example will describe how `with_traceback()` and `raise .. from ..` affect in its error message.

<details><summary>with_traceback()</summary><p>

```python
import sys
try:
    raise Exception("a")
except Exception:
    exc_type, exc_value, exc_traceback = sys.exc_info()
    raise Exception("b").with_traceback(exc_traceback)
```
```shell
  File "test1.py", line 3, in <module>
    raise Exception("a")
Exception: a

  File "test1.py", line 6, in <module>
    raise Exception("b").with_traceback(exc_traceback)
  File "test1.py", line 3, in <module>
    raise Exception("a")
Exception: b
```

</p></details>


<details><summary>raise .. from ..</summary><p>

```python
try:
    raise Exception("a")
except Exception as e:
    raise Exception("b") from e
```

```shell
  File "test2.py", line 2, in <module>
    raise Exception("a")
Exception: a

The above exception was the direct cause of the following exception:

  File "test2.py", line 4, in <module>
    raise Exception("b") from e
Exception: b
```

</p></details>

<details><summary>with_traceback() & raise .. from ..</summary><p>

```python
import sys
try:
    raise Exception("a")
except Exception:
    exc_type, exc_value, exc_traceback = sys.exc_info()
    raise Exception("b").with_traceback(exc_traceback) from None
```

```shell
  File "test3.py", line 6, in <module>
    raise Exception("b").with_traceback(exc_traceback) from None
  File "test3.py", line 3, in <module>
    raise Exception("a")
Exception: b
```

</p></details>

fritzo:
Thanks for the detailed explanation @akihironitta ! I like this new `raise ... from ...` syntax and would be happy to drop the older `.with_traceback()` syntax, as in your suggestion 1. Would you be able to do that in this PR, or would you prefer a separate PR?

akihironitta:
Sure, I will complete all those fixes in this PR if that's ok!

akihironitta:
Just for the record, below are the lists of lines which needs checking.
### Lines where `.with_traceback()` is used
Run `grep "with_traceback" -r -n .`.
```
./pyro/poutine/trace_struct.py:196:                                         .format(name, exc_value, shapes)).with_traceback(traceback)
./pyro/poutine/trace_struct.py:221:                                         .format(name, exc_value, shapes)).with_traceback(traceback)
./pyro/poutine/trace_struct.py:248:                                     .format(name, exc_value, shapes)).with_traceback(traceback)
./pyro/poutine/trace_struct.py:382:                                 .format(site["name"], exc_value, shapes)).with_traceback(traceback)
./pyro/poutine/trace_messenger.py:170:                exc = exc.with_traceback(traceback)
```

### Lines where exceptions are `raise`d in `except` clause
Run [this script](https://gist.github.com/akihironitta/4223c1b32404b36c1b349d70c4c93b4d). Note that the list contains false positives.

```
examples/vae/utils/mnist_cached.py#L237                                                                                                          
examples/sparse_gamma_def.py#L190
pyro/ops/jit.py#L30
pyro/ops/integrator.py#L82
pyro/ops/contract.py#L352
pyro/ops/packed.py#L29
pyro/poutine/trace_struct.py#L195
pyro/poutine/trace_struct.py#L220
pyro/poutine/trace_struct.py#L247
pyro/poutine/trace_struct.py#L381
pyro/poutine/trace_messenger.py#L171
pyro/contrib/forecast/evaluate.py#L197
pyro/contrib/forecast/util.py#L215
pyro/contrib/epidemiology/compartmental.py#L695
pyro/contrib/autoname/named.py#L218
pyro/infer/discrete.py#L27
pyro/infer/reparam/neutra.py#L68
pyro/infer/mcmc/logger.py#L178
pyro/infer/mcmc/logger.py#L209
tests/contrib/funsor/test_valid_models_enum.py#L117
tests/contrib/funsor/test_valid_models_enum.py#L141
tests/contrib/funsor/test_valid_models_enum.py#L164
```

fritzo:
It looks like the removal of `.with_traceback()` is breaking some of our death tests. Is the error message preserved?

akihironitta:
I'm sure the error message holds all the information but in a different format. We need to update the death tests or keep `.with_traceback()` as it is.

Only applying `raise ... from ...` syntax is surely better than the current main branch. Since this PR originally tried to fix only explicit exception chaining, it might be reasonable to fix duplicate tracebacks by removing `.with_traceback()` at another time.

How does this sound?

fritzo:
> ... removing `.with_traceback()` at another time.

that sounds fine. Thanks for attempting the cleanup. I didn't expect it would become so intricate.

akihironitta:
@fritzo 
>  I didn't expect it would become so intricate.
Me, neither. I rolled back the commits and it seems the tests went well. Could you have a look at the changes?

fritzo (approved):

Issue #2571, opened 2020-07-22
[FR] Add ordered categorical distribution


ecotner:
### Issue Description
I'm going through the Statistical Rethinking textbook by McElreath and in chapter 12 the author discusses how to handle ordered categorical variables in a sensible way. I tried to look through the lists of distributions available in both `torch` and `pyro`, but could not find anything similar to the `dordlogit` (aka ordered-logit) distribution that he describes in the text. I realize that it's just a re-parametrization of the categorical distribution, but I think it might be a useful utility to include.

I tried to recreate the distribution myself; does this seem like a sensible implementation, and would it be something you would want to include in `pyro`?

### Code Snippet
I'm no expert in implementing torch/pyro distributions, but this seems to do the job, and I have used it successfully to do the trolley problem example in the textbook using NUTS. There are probably other methods/attributes that are missing for a full implementation.

```python
import torch
from torch.distributions import constraints
from pyro.distributions import TorchDistribution, Categorical

class OrderedCategorical(TorchDistribution):
    """
    Alternative parametrization of the distribution over a categorical variable.
    
    Instead of the typical parametrization of a categorical variable in terms
    of the probability mass of the individual categories ``p``, this provides an
    alternative that is useful in specifying ordered categorical models. This
    accepts a list of ``cutpoints`` which are a (potentially initially unordered)
    vector of real numbers denoting baseline cumulative log-odds of the individual
    categories, and a model vector ``phi`` which modifies the baselines for each
    sample individually.
    
    These cumulative log-odds are then transformed into a discrete cumulative
    probability distribution, that is finally differenced to return the probability
    mass function ``p`` that specifies the categorical distribution.
    """
    support = constraints.nonnegative_integer
    arg_constraints = {"phi": constraints.real, "cutpoints": constraints.real}
    has_rsample = False
    
    def __init__(self, phi, cutpoints):
        assert len(cutpoints.shape) == 1 # cutpoints must be 1d vector
        assert len(phi.shape) == 1 # model terms must be 1d vector of samples
        N, K = phi.shape[0], cutpoints.shape[0]+1
        cutpoints = torch.sort(cutpoints).values.reshape(1, -1)  # sort and reshape for broadcasting
        q = torch.sigmoid(cutpoints - phi.reshape(-1, 1))  # cumulative probabilities
        # turn cumulative probabilities into probability mass of categories
        p = torch.zeros((N, K)) # (batch/sample dim, categories)
        p[:,0] = q[:,0]
[...]

Issue #2630, opened 2020-09-16
List: move these things upstream to PyTorch, Funsor, etc.


fritzo:
This issue tracks Pyro components that should be moved to upstream dependencies.

## Move to PyTorch
- [ ] `dist.Delta` #1781
- [ ] `constraints.IndependentConstraint`
  - usage in `MultivariateNormal`
  - usage in `dist.Independent`
- [ ] `constraints.integer`
- [ ] `Geometric.log_prob()` as patched in pyro/distributions/torch.py
- [ ] `Uniform`'s `._unbroadcasted_*` logic

## Move to Funsor
See also https://github.com/pyro-ppl/funsor/issues/362

- [ ] `distributions.util.detach()`
- [ ] `ops.tensor_utils.haar_transform()` and `.inverse_haar_transform()`

RylanSchaeffer:
@fritzo out of curiosity is there a tutorial on the process for migrating code, updating docs and writing tests?

fritzo:
@RylanSchaeffer there is no tutorial on migrating distributions. You'd need to
- add code to torch/distributions/dirichlet_multinomial.py or torch/distributions/conjugate.py
- fix docstrings to follow numpy syntax rather than sphinx syntax
- import distributions in torch/distributions/__init__.py
- add tests to test/distributions/test_distributions.py
- add an entry in docs/source/distributions.rst
- cc @neerajprad and I on the pull request.

See for example #48798

PR #2651, opened 2020-10-03
added ordered logistic distribution

ecotner:
following up on issue #2571. I think my formatter automatically changed the `pyro/distributions/__init__.py` import statements around a bit, I hope that's alright


pyro/distributions/ordered_logistic.py
```diff
@@ -0,0 +1,49 @@
+import torch
+from torch.distributions import constraints
+from pyro.distributions.torch import Categorical
```
ecotner, line 3: I would like to point out that for some reason when running tests, I could not get anything to work when using `from pyro.distributions import Categorical`; kept getting an error that the `Categorical` class didn't exist...

It seems that `pyro/distributions/__init__.py` is importing the `Categorical` class from `pyro/distributions/torch.py` though, so hopefully this doesn't cause any problems.

fehiepsi (commented):
Thanks for the contribution, @ecotner! Could you add OrderedLogistic to this [list](https://github.com/pyro-ppl/pyro/blob/dev/tests/distributions/conftest.py#L268) to get more completed tests? You also need to add this class to [docs](https://github.com/pyro-ppl/pyro/blob/dev/docs/source/distributions.rst).

My main concern is the implementation only works for `batch_shape=(N,)`, which is the shape of `predictors`. You might revise the implementation a bit to support arbitrary batch_shape.

ecotner:
Yeah, I can add it to the tests/docs!

With regards to the batch shape, I also had the same thought initially, but according to the [pytorch documentation](https://pytorch.org/docs/stable/distributions.html#categorical), their implementation of `Categorical` only accepts 1D or 2D tensors... although now that you mention it, I went back to look at [the source](https://github.com/pytorch/pytorch/blob/master/torch/distributions/categorical.py#L47-L55) and it appears that arbitrary tensor shapes are acceptable! Perhaps I should also submit a PR with them to update the documentation :thinking:

pyro/distributions/ordered_logistic.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+
+    support = constraints.nonnegative_integer
+    arg_constraints = {
+        "predictors": constraints.real,
+        "cutpoints": constraints.real_vector,
+    }
+    has_rsample = False
+
+    def __init__(self, predictors, cutpoints):
```
fehiepsi, line 37: Could you add `validate_args=None` and modify the last line to `super(OrderedLogistic, self).__init__(p, validate_args=validate_args)`?

pyro/distributions/ordered_logistic.py
```diff
@@ -0,0 +1,57 @@
...
+        assert cutpoints.ndim == 1, "cutpoints must be 1d vector"
+        assert predictors.ndim >= 1, "predictors must have at least 1 dimension"
+        assert len(cutpoints) > 0
+        assert predictors.shape[-1] > 0
+        if len(cutpoints) > 1:
+            err_msg = "cutpoints must be monotonically increasing"
+            assert (cutpoints[1:] > cutpoints[:-1]).all().item(), err_msg
```
fehiepsi, line 44: Could you refactor this into an `ordered_vector` [constraint](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/constraints.py) so that in `arg_constraints`, you can set constraint for `cutpoints` to be `constraints.ordered_vector`?

pyro/distributions/ordered_logistic.py
```diff
@@ -0,0 +1,57 @@
...
+
+    :param Tensor predictors: A tensor of predictor variables of arbitrary
+        shape. The output shape of non-batched samples from this distribution will
+        be the same shape as ``predictors``.
+    :param Tensor cutpoints: A tensor of cutpoints that are used to determine the
+        cumulative probability of each entry in ``predictors`` belonging to a
+        given category. Must be 1-dimensional and monotonically increasing.
+    """
+
+    support = constraints.nonnegative_integer
```
fehiepsi, line 30: I think you can remove this line. The support for `Categorical` is an interger integral.

pyro/distributions/ordered_logistic.py
```diff
@@ -0,0 +1,57 @@
...
+    arg_constraints = {
+        "predictors": constraints.real,
+        "cutpoints": constraints.real_vector,
+    }
+    has_rsample = False
```
fehiepsi, line 35: nit: this line is not necessary

pyro/distributions/ordered_logistic.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __init__(self, predictors, cutpoints):
+        assert cutpoints.ndim == 1, "cutpoints must be 1d vector"
+        assert predictors.ndim >= 1, "predictors must have at least 1 dimension"
```
fehiepsi, line 39: Thanks for supporting `batch_shape` for `predictors`! Could you also support batch_shape for `cutpoints`? I would expect those parameters have shape
```
cutpoints.shape == batch_shape + event_shape
predictors.shape == batch_shape
```
where `event_shape = cutpoints.shape[-1:]` and `batch_shape` can be empty. You can broadcast those parameters as in [MVN implementation](https://github.com/pytorch/pytorch/blob/master/torch/distributions/multivariate_normal.py#L130).

fritzo:
> Perhaps I should also submit a PR with [PyTorch] to update the documentation

Good catch, an upstream PR would be appreciated!

pyro/distributions/ordered_logistic.py
fritzo: @ecotner FYI the advantage of expressing this as a `Constraint` object is that we can then define and register bijections into unconstrained space, thereby permitting inference over the cutpoints parameter.

ecotner:
> > Perhaps I should also submit a PR with [PyTorch] to update the documentation
> 
> Good catch, an upstream PR would be appreciated!

Done! The build is failing for some reason though... don't think it's the change I made to the docstring :man_shrugging:

pyro/distributions/ordered_logistic.py
ecotner: I think I found something a little more flexible and intuitive than fixing `cutpoints` and `predictors` to have the same `batch_shape` - as long as the first `N = cutpoints.ndim - 1` dimensions are broadcastable to the last `N` dimensions of `predictors`, then we can just use normal broadcasting semantics to handle it by appending an event dim to `predictors`:

```python
q = torch.sigmoid(cutpoints - predictors.view(predictors.shape + (1,)))
```

Check out the test I wrote for it: https://github.com/pyro-ppl/pyro/pull/2651/files#diff-f0cc7966b175f63150546ace313d0179R31-R42 - does that make sense?

pyro/distributions/ordered_logistic.py
ecotner: @fritzo that is interesting! I have seen the "bijection to unconstrained space" trick mentioned a couple times in the documentation an in other places, but did not realize the `Constraint` object is how that is defined. How exactly does `Constraint` define that bijection? It seems like the implementations in [constraints.py](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/constraints.py) only define methods to check whether an input satisfies the constraint or not. Is there some other place where the constraints are mapped to the appropriate bijective functions that are used in SVI and HMC?

pyro/distributions/constraints.py
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) 2017-2019 Uber Technologies, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
+import torch
 from torch.distributions.constraints import *  # noqa F403
 from torch.distributions.constraints import Constraint
 from torch.distributions.constraints import __all__ as torch_constraints
```
ecotner, line 7: Kind of off-topic, but I noticed there are a couple `Constraint` implementations here marked with `# TODO move this upstream to torch.distributions` that are several months old... perhaps that was forgotten about somewhere along the way? :laughing:

fehiepsi: Haha, if I remember correctly, then the blocker is it is not straightforward to add write tests when we move those upstream.

pyro/distributions/ordered_logistic.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(self, predictors, cutpoints, validate_args=None):
+        self.predictors = predictors
+        self.cutpoints = cutpoints
+        # get shape for input to Categorical dist
+        p_shape = predictors.shape + (cutpoints.shape[-1] + 1,)
+        # calculate cumulative probability for each predictor
+        q = torch.sigmoid(cutpoints - predictors.view(predictors.shape + (1,)))
+        # turn cumulative probabilities into probability mass of categories
+        p = torch.zeros(p_shape)
```
fehiepsi, line 45: I think you need to let `p` have the same device, dtype as `q`.

pyro/distributions/ordered_logistic.py
```diff
@@ -0,0 +1,49 @@
...
+        "cutpoints": constraints.ordered_vector,
+    }
+
+    def __init__(self, predictors, cutpoints, validate_args=None):
+        self.predictors = predictors
+        self.cutpoints = cutpoints
+        # get shape for input to Categorical dist
+        p_shape = predictors.shape + (cutpoints.shape[-1] + 1,)
+        # calculate cumulative probability for each predictor
+        q = torch.sigmoid(cutpoints - predictors.view(predictors.shape + (1,)))
```
fehiepsi, line 43: I think you can set `p_shape = q.shape[:-1] + (q.shape[-1] + 1,)`, which will also work for the case `predictor.shape = (3,)`, `cutpoints.shape = (10, 3, 5)`.

fehiepsi (commented):
Looks good to me overall. Thanks for making OrderedLogistic more flexible @ecotner! I have a few remaining comments:

pyro/distributions/ordered_logistic.py
fehiepsi: It makes sense to me. Thanks!

pyro/distributions/ordered_logistic.py
fritzo: @ecotner Transforms are defined by an extensible [registration mechanism](https://github.com/pytorch/pytorch/blob/master/torch/distributions/constraint_registry.py). See the bottom of that file for registrations. Note there are two generic transformers: `biject_to()` and `transform_to()`; the latter allows many-to-one mappings and can be cheaper.

pyro/distributions/constraints.py
fritzo: Yes, I have halted all my "move upstream" work #2630 until we split up test_distributions.py and can use PyTest https://github.com/pytorch/pytorch/issues/11578

fritzo (commented):
Very clean!

I beilive you'll need to implement a custom `.expand()` method for this to work with plates. As a test, you might add a simple example model to the smoke tests in tests/infer/test_valid_models.py

pyro/distributions/ordered_logistic.py
```diff
@@ -0,0 +1,49 @@
...
+    mass matrix ``p`` that specifies the categorical distribution.
+
+    :param Tensor predictors: A tensor of predictor variables of arbitrary
+        shape. The output shape of non-batched samples from this distribution will
+        be the same shape as ``predictors``.
+    :param Tensor cutpoints: A tensor of cutpoints that are used to determine the
+        cumulative probability of each entry in ``predictors`` belonging to a
+        given category. The first `cutpoints.ndim-1` dimensions must be
+        broadcastable to ``predictors``, and the -1 dimension is monotonically
+        increasing.
```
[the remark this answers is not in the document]
ecotner: thanks!

pyro/distributions/ordered_logistic.py
ecotner: Hmm the interpretation of `cutpoints.ndim > predictor.ndim + 1` seems confusing to me. Wouldn't that essentially mean that `len(cutpoints.event_shape) == 2` in your example? I think the underlying `Categorical` implementation can only handle `len(event_shape) <= 1`.

ecotner: Actually, I take that back, you could interpret it as batches of `cutpoints` being broadcast to `predictor`, while still retaining a single event dimension. That makes sense.

ecotner:
@fritzo sorry if I'm not familiar enough with the library to know this, but why is a custom `.expand()` method needed? It looks like `Categorical` has an `.expand()` method implemented, and `OrderedLogistic` subclasses from that. If it is needed, would it be as simple as copying the implementation in `TorchDistribution` like
```python
    return ExpandedDistribution(self, batch_shape)
```

fritzo:
@ecotner for `.expand()` each subclass needs to expand its attributes. I think you can get away with not expanding attributes but you'll want to call `Categorical.expand()` rather than creating an `ExpandedDistribution`: 
```py
    new = self._get_checked_instance(OrderedCategorical, _instance)
    return Categorical.expand(self, batch_shape, new)
```

ecotner:
> for `.expand()` each subclass needs to expand its attributes

Oh, I see, so like expand `predictor` and `cutpoints` in addition to just the underlying `p` that defines the `Categorical`?

pyro/distributions/ordered_logistic.py
```diff
@@ -12,38 +12,45 @@ class OrderedLogistic(Categorical):
...
         p[..., -1] = 1 - q[..., -1]
         super(OrderedLogistic, self).__init__(p, validate_args=validate_args)
+
+    def expand(self, batch_shape, _instance=None):
+        batch_shape = torch.Size(batch_shape)
+        new = self._get_checked_instance(OrderedLogistic, _instance)
+        predictor = self.predictor.expand(batch_shape)
+        cutpoints = self.cutpoints.expand(batch_shape + (self.cutpoints.shape[-1],))
+        new.__init__(predictor, cutpoints)
+        return new
```
fehiepsi: Hmm, the purpose of `expand` is to avoid recomputing the logics in the constructor again. I think it is better to do
```python
new.predictor = ...
new.cutpoints = ...
return super(OrderedLogistic, self).expand(batch_shape, new)
# or return Categorical.expand(self, batch_shape, new)
```
If you broadcast `predictor` and `cutpoints` here, then could you also do so in `__init__`?, kind of
```python
super(OrderedLogistic, self).__init__(p, validate_args=validate_args)
self.predictor = predictor.expand(self.batch_shape)
self.cutpoints = cutpoints.expand(self.batch_shape + cutpoints.shape[-1:])
```

fehiepsi: nit: you missed a new line at the end of this file

pyro/distributions/ordered_logistic.py
```diff
@@ -0,0 +1,56 @@
...
+        self.predictor = predictor
+        self.cutpoints = cutpoints
+        # calculate cumulative probability for each predictor
+        q = torch.sigmoid(cutpoints - predictor.view(predictor.shape + (1,)))
+        # turn cumulative probabilities into probability mass of categories
+        p_shape = q.shape[:-1] + (q.shape[-1] + 1,)
+        p = torch.zeros(p_shape, dtype=q.dtype, device=q.device)
+        p[..., 0] = q[..., 0]
+        p[..., 1:-1] = (q - torch.roll(q, 1, dims=-1))[..., 1:]
+        p[..., -1] = 1 - q[..., -1]
```
fritzo, line 47: I'd try to replace these mutations with a purely functional form
```py
p = torch.cat([q[..., 0],
               (q - torch.roll(q, 1, dims=-1))[..., 1:],
               1 - q[..., -1]], dim=-1)
```
Actually I'm surprised the in-place edits don't break autograd. Are we sure gradient is tested?

pyro/distributions/ordered_logistic.py
```diff
@@ -0,0 +1,56 @@
...
+    arg_constraints = {
+        "predictor": constraints.real,
+        "cutpoints": constraints.ordered_vector,
+    }
+
+    def __init__(self, predictor, cutpoints, validate_args=None):
+        self.predictor = predictor
+        self.cutpoints = cutpoints
+        # calculate cumulative probability for each predictor
+        q = torch.sigmoid(cutpoints - predictor.view(predictor.shape + (1,)))
```
fritzo, line 41: Avoid `.view()` in favor of `.reshape()` or `.unsqueeze()` which are safer:
```py
q = torch.sigmoid(cutpoints - predictor.unsqueeze(-1))
```

fritzo (commented):
Looks great, I have only minor PyTorch comments. Thanks for working through the long review cycle!

pyro/distributions/ordered_logistic.py
ecotner: I think the gradient still works; there's like a `SelectBackward` (or `InsertBackward`? can't remember which) grad function for those types of in-place mutations. I double-checked that the gradient still flows through the log prob with the current implementation and it looks like it works:

```python
>>> from pyro.distributions import OrderedLogistic
>>> import torch
>>> pred = torch.randn(5, requires_grad=True)
>>> cp = torch.sort(torch.randn(3)).values
>>> cp.requires_grad = True
>>> data = torch.tensor([0,1,2,3,0]).float()
>>> dist = OrderedLogistic(pred, cp)
>>> dist.log_prob(data).sum().backward()
>>> pred.grad
tensor([-0.8494, -0.8638, -0.6831,  0.7186, -0.7939])
>>> cp.grad
tensor([-2.7713,  5.0584,  0.1845])
```
I was playing around with your suggested implementation and it seems that one problem is that if you only have a single cutpoint along the last dimension (i.e. two categories) you can get the following error:
```python
>>> q = torch.tensor([1.])
>>> torch.cat([q[...,0], (q-torch.roll(q, 1, dims=-1))[...,1:], 1-q[...,-1]], dim=-1)
RuntimeError: zero-dimensional tensor (at position 0) cannot be concatenated
```
because the second element in the concat list is empty. But there's no problem assigning nothing to nothing
```python
>>> p = torch.tensor([0,0])
>>> p[...,1:-1] = (q-torch.roll(q, 1, dims=-1))[...,1:]
>>> p
tensor([0, 0])
```

pyro/distributions/ordered_logistic.py
ecotner: ahh yes, that makes sense

pyro/distributions/ordered_logistic.py
ecotner: actually, that's probably a good idea to add to the tests!

fritzo (dismissed):

pyro/distributions/ordered_logistic.py
```diff
@@ -0,0 +1,55 @@
...
+        p[..., 0] = q[..., 0]
+        p[..., 1:-1] = q[..., 1:] - q[..., :-1]
+        p[..., -1] = 1 - q[..., -1]
+        super(OrderedLogistic, self).__init__(p, validate_args=validate_args)
+
+    def expand(self, batch_shape, _instance=None):
+        batch_shape = torch.Size(batch_shape)
+        new = self._get_checked_instance(OrderedLogistic, _instance)
+        new.predictor = self.predictor.expand(batch_shape)
+        new.cutpoints = self.cutpoints.expand(batch_shape + (self.cutpoints.shape[-1],))
```
fehiepsi, line 54: nit: If we expand here, I think we should also do expanding in the constructor `self.predictor = predictor.expand(self.batch_shape) after `super(OrderedLogistic, self).init(p, validate_args=validate_args)` just for consistency.

fehiepsi (dismissed):
Awesome, @ecotner! I just have one minor comment. Thanks for adding this distribution to Pyro!

pyro/distributions/ordered_logistic.py
ecotner: I think it might be better to add the expansion beforehand since the batch shape can be determined from `q`. Saves two lines of code :wink:

fehiepsi (approved):

PR #2653, opened 2020-10-06
Replace torch.jit.script with torch.jit.script_if_tracing

fritzo:
Replaces #2387
See https://github.com/pytorch/pytorch/issues/45921

### DIFF
--- a/pyro/distributions/hmm.py
+++ b/pyro/distributions/hmm.py
@@ -6,7 +6,7 @@ from torch.distributions import constraints
 
 from pyro.distributions.torch import Categorical, Gamma, Independent, MultivariateNormal
 from pyro.distributions.torch_distribution import TorchDistribution
-from pyro.distributions.util import broadcast_shape
+from pyro.distributions.util import broadcast_shape, torch_jit_script_if_tracing
 from pyro.ops.gamma_gaussian import (GammaGaussian, gamma_and_mvn_to_gamma_gaussian, gamma_gaussian_tensordot,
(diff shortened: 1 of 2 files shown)


fehiepsi (approved):

mpariente:
I don't know if you fixed this since then, but `torch.jit.script_if_tracing` actually calls `torch.jit.script` in torch 1.6.0 because `torch.jit.is_tracing` returns the function instead of the bool. 

```python 
In [2]: torch.jit.is_tracing()
Out[2]: <function torch._C.PyCapsule._is_tracing>

In [3]: torch._C._is_tracing()
Out[3]: False

In [4]: torch._C._is_tracing
Out[4]: <function torch._C.PyCapsule._is_tracing>

In [5]: torch.jit.is_tracing??
Signature: torch.jit.is_tracing()
Source:   
def is_tracing():
    """
    Returns ``True`` in tracing (if a function is called during the tracing of
    code with ``torch.jit.trace``) and ``False`` otherwise.
    """
    return torch._C._is_tracing
```

fritzo:
Thanks @mpariente , do I understand correctly that you have found an upstream bug, and that Pyro is correctly using the interface provided by PyTorch? Thanks, we're doing our best :)

mpariente:
If you're using torch 1.6.0, `script_if_tracing` (and `_script_if_tracing`) don't work, they call `script` even when it's not in tracing mode. So no, I think Pyro is not loading `jit` lazily.

PR #2659, opened 2020-10-08
Adding Mean Field LDA draft example (yields bad results)

ucals:
Hey guys,
As [requested in the forum](https://forum.pyro.ai/t/params-inside-plates/2248/8), I'm opening a draft PR with the buggy Mean Field LDA code. Glad to talk if comments!
Cheers!
Carlos

### DIFF
--- a/examples/meanfield_lda.py
+++ b/examples/meanfield_lda.py
@@ -1,0 +1,175 @@
+import math
+import pandas as pd
+import pyro
+import pyro.distributions as dist
+import torch
+from torch.distributions import constraints
+from pyro.infer import SVI, Predictive, TraceEnum_ELBO
+from tqdm import trange
+import requests
+import tarfile
+from pathlib import Path
+from torch.nn.utils.rnn import pad_sequence
+from nltk.tokenize import RegexpTokenizer
+from nltk.stem.wordnet import WordNetLemmatizer
+
+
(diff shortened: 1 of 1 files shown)


examples/meanfield_lda.py
```diff
@@ -0,0 +1,175 @@
...
+
+    # Substitute words by their indexes
+    for doc_id in range(len(docs)):
+        df = pd.DataFrame(columns=['word'], data=docs[doc_id])
+        df = pd.merge(df, vocab[['index', 'word']], how='left', on='word').dropna()
+        docs[doc_id] = torch.from_numpy(df['index'].astype(int).values)
+
+    # Remove docs with zero length (2 occurrences) and pad docs with 0 index
+    docs = [doc for doc in docs if len(doc) > 0]
+    docs = pad_sequence(docs, batch_first=True, padding_value=0)
```
slayeren, line 134: are you sure that padding each document with zeros is a good idea? doesn't that affect the performance of the model?

eb8680: @slayeren this implementation uses dense tensors and Pyro's automatic enumeration to integrate out `zeta` and is indeed quite slow (hence the title :) ).  See the [ProdLDA tutorial](http://pyro.ai/examples/prodlda.html) for a less naive approach to topic modeling in Pyro.

Please use our [forum](https://forum.pyro.ai/) for any followup questions.

Issue #2247, opened 2020-01-08
[FR] pyro.barrier statement for SMC


fritzo:
This issue proposes a new statement in Pyro: a `pyro.barrier` statement to allow interaction between an inference algorithm and a model's local state. The interface is
```py
thing = pyro.barrier(thing)
```
where `thing` is a recursive data structure whose elements may be
1.  Python atoms, which will be unaffected
2.  Python collections, which will be recursed
    ```py
    barrier([x,y,z]) = [barrier(x), barrier(y), barrier(z)]
    ```
3. PyTorch tensors with named dimensions (or Funsors with named free variables). Note this feature crucially relies on PyTorch named dimensions to pass event_dim metadata through arithmetic computations in user code.

Using `pyro.barrier` we can re-implement SMC in a more Pyronic way, avoding restrictions on an `init`/`step` interface. For example we could implement SMC via effects acting on a `pyro.barrier` statement in the following annotated model:
```py
def model(data):
    z = pyro.sample("z_init", Normal(0,1))    # latent state
    k = 0 * z                                # control
    cost = 0 * z                             # cumulative cost of controller
    for t, x in enumerate(data):
        z,k,cost = pyro.barrier((z,k,cost))  # triggers resampling
        k = torch.where(z > 10, k + 1, k)
        k = torch.where(z < 10, k + 1, k)
        k = torch.where(-10 <= z & z <= 10, 0 * k, k)
        z = pyro.sample(f"z_{t}", Normal(z+k,1))
        pyro.sample(f"x_{t}", Normal(z,1), obs=x)
    return cost
```
While in Pyro this statement has limited power, supporting only a few features like SMC resampling and enumeration-delayed sampling, it becomes more powerful in a PPL with better support for lazy computation such as `funsor.minipyro`. I'd like to add this to the `pyroapi` so we can experiment with more inference algorithms in a generic way.

For details see [design doc](https://github.com/fritzo/barrier-design-doc/blob/master/main.pdf) and draft Funsor implementation https://github.com/pyro-ppl/funsor/pull/295

PR #2660, opened 2020-10-09
Implement first version of pyro.barrier

fritzo:
Addresses #2247
Pair coded with @eb8680 

This implements a first version of a `pyro.barrier()` statement. Further enhancements will require changes in [funsor](https://github.com/pyro-ppl/funsor).

## Tested
- [x] added a smoke test to test_valid_models.py


eb8680 (approved):
Should be good to merge, but won't be very useful yet without more functionality in Funsor. Do you want to include this in the 1.5 release or wait to merge until after?

fritzo:
Yeah let's include it in 1.5. I think it will be helpful to have the pair of experimental functions (`poutine.collapse`, `pyro.barrier`).

PR #2662, opened 2020-10-11
create OrderedTransform

ecotner:
Basically just copied the `OrderedTransform` implementation [in numpyro](http://num.pyro.ai/en/stable/_modules/numpyro/distributions/transforms.html#OrderedTransform) exactly. Was surprised the jacobian was so simple, I had to re-derive it myself just to make sure!

### DIFF
--- a/docs/source/distributions.rst
+++ b/docs/source/distributions.rst
@@ -692,10 +692,6 @@ neural_autoregressive
 ---------------------
 .. autofunction:: pyro.distributions.transforms.neural_autoregressive
 
-ordered
--------
-.. autofunction:: pyro.distributions.transforms.ordered
-
 permute
 -------
 .. autofunction:: pyro.distributions.transforms.permute
--- a/docs/source/distributions.rst
+++ b/docs/source/distributions.rst
@@ -382,6 +382,13 @@ LowerCholeskyAffine
     :undoc-members:
     :show-inheritance:
 
+OrderedTransform
+----------------
+.. autoclass:: pyro.distributions.transforms.OrderedTransform
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
 Permute
 -------
 .. autoclass:: pyro.distributions.transforms.Permute
@@ -684,6 +691,10 @@ matrix_exponential
 neural_autoregressive
 ---------------------
 .. autofunction:: pyro.distributions.transforms.neural_autoregressive
+
+ordered
+-------
+.. autofunction:: pyro.distributions.transforms.ordered
 
 permute
 -------
--- a/pyro/distributions/transforms/ordered.py
+++ b/pyro/distributions/transforms/ordered.py
@@ -6,10 +6,10 @@ from pyro.distributions import constraints
 class OrderedTransform(Transform):
     """
     Transforms a real vector into an ordered vector.
-    
+
     Specifically, enforces monotonically increasing order on the last dimension
     of a given tensor via the transformation :math:`y_0 = x_0`,
-    :math:`y_i = \sum_{1 \le j \le i} \exp(x_i)`
+    :math:`y_i = \\sum_{1 \\le j \\le i} \\exp(x_i)`
     """
     domain = constraints.real_vector
     codomain = constraints.ordered_vector
--- a/pyro/distributions/transforms/ordered.py
+++ b/pyro/distributions/transforms/ordered.py
@@ -8,8 +8,8 @@ class OrderedTransform(Transform):
     Transforms a real vector into an ordered vector.
     
     Specifically, enforces monotonically increasing order on the last dimension
-    of a given tensor via the transformation $y_0 = x_0$,
-    $y_i = \sum_{1 \le j \le i} \exp(x_i)$
+    of a given tensor via the transformation :math:`y_0 = x_0`,
+    :math:`y_i = \sum_{1 \le j \le i} \exp(x_i)`
     """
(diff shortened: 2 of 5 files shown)


ecotner:
secondary follow-up to #2571. @fehiepsi I want to create some tests for this, but am unsure of where to put them. I see that there is a [test_transforms.py](https://github.com/pyro-ppl/pyro/blob/dev/tests/distributions/test_transforms.py) but it seems like it requires a `TransformModule`? do I need to turn `OrderedTransform` into a `OrderedTransformModule` to get tests?

fehiepsi:
@ecotner I think you can create some tests in `test_ordered_logistic` that mimic transform tests in [test_lkj](https://github.com/pyro-ppl/pyro/blob/dev/tests/distributions/test_lkj.py).

ecotner:
@fehiepsi think that'll cover it?

pyro/distributions/transforms/ordered.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def _inverse(self, y):
+        x = (y[..., 1:] - y[..., :-1]).log()
+        return torch.cat([y[..., :1], x], dim=-1)
+
+    def log_abs_det_jacobian(self, x, y):
+        return torch.sum(x[..., 1:], dim=-1)
+
+
+def ordered():
```
fehiepsi, line 31: This lowercase pattern is particularly used for normalization flows, where the helpful simplies some underlying codes. It is not needed I believe.

tests/distributions/test_ordered_logistic.py
```diff
@@ -75,3 +80,48 @@ def test_autograd():
...
+    Recursively splits tensor ``X`` along its leading dimensions until we are
+    left with a vector, computes the jacobian of this vector under the
+    transformation ``func``, then stitches all the results back together using
+    ``torch.stack``.
+    """
+    assert X.ndim >= 1
+    if X.ndim == 1:
+        return jacobian(func, X)
+    else:
+        return torch.stack([cjald(func, X[i]) for i in range(X.shape[0])], dim=0)
```
fehiepsi, line 111: nice implementation!

fehiepsi (commented):
Looks great to me! Could you also register the bijective similar to [here](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/transforms/__init__.py#L54)? That way, if the inference code (e.g. in MCMC) sees the `constraint`, it will use the corresponding transform to transform a value in an unconstrained domain to a constrained domain.

pyro/distributions/transforms/ordered.py
```diff
@@ -0,0 +1,35 @@
...
+    Transforms a real vector into an ordered vector.
+
+    Specifically, enforces monotonically increasing order on the last dimension
+    of a given tensor via the transformation :math:`y_0 = x_0`,
+    :math:`y_i = \\sum_{1 \\le j \\le i} \\exp(x_i)`
+    """
+    domain = constraints.real_vector
+    codomain = constraints.ordered_vector
+    bijective = True
+    sign = +1
```
fehiepsi, line 17: Could you also add `event_dim = 1` here?

tests/distributions/test_ordered_logistic.py
ecotner: thanks, I thought it was pretty slick!

fritzo: Indeed slick! You could even condense :smile:
```py
torch.stack([cjald(func, x) for x in X.unbind(0)], dim=0)
```

docs/source/distributions.rst
```diff
@@ -685,6 +692,10 @@ neural_autoregressive
 ---------------------
 .. autofunction:: pyro.distributions.transforms.neural_autoregressive
 
+ordered
```
fritzo, line 695: It looks like docs is failing because you haven't defined a wrapper function `ordered` as many of the other transforms have defined.

fritzo (commented):
Could you add standard tests to tests/distributions/test_transforms.py, which has a nice framework built by @stefanwebb ? It should be a quick two-liner.

docs/source/distributions.rst
ecotner: Haha just noticed that. I did have one defined, but then @fehiepsi suggested I remove it because those helpers are typically reserved for normalizing flows. Forgot to take it out of the docs.

ecotner:
> Could you add standard tests to `tests/distributions/test_transforms.py`

Sure can! It appears the `_test_autodiff` method expects transformations to have learnable parameters (which `OrderedTransform` does not), so I've disabled it, but I do have some separate tests in `test_ordered_logistic.py` that makes sure autodiff propagates through the transform.

---
Also, FYI it looks like the inverse test for `test_conditional_generalized_channel_permute` is failing on line 109 with warning 
```
UserWarning: Mixed memory format inputs detected while calling the operator.
The operator will output contiguous tensor even if some of the inputs are in channels_last format.
(Triggered internally at  /pytorch/aten/src/ATen/native/TensorIterator.cpp:918.)
```

fehiepsi (approved):
LGTM if tests pass. Thanks, @ecotner!

ecotner:
hey @fehiepsi @fritzo I just realized I have enough commits this month to actually participate in hacktoberfest; would you consider marking this PR as `hacktoberfest-accepted`? Would be cool to get that t-shirt :sweat_smile:

PR #2671, opened 2020-10-15
Bump version to 1.5.0

fritzo:
- Add a `scripts/update_version.py` and a `make version` comand
- Add missing license header (I haven't figured out why `make lint` isn't catching these)
- Bump version to 1.5.0


fritzo:
@eb8680 when this merges I'll publish a release and PyPi release

eb8680 (approved):

PR #2673, opened 2020-10-16
Fix scanvi warning

fritzo:
This removes the `UserWarning: Found vars in model but not guide` from the scANVI example.

@martinjankowiak can you please confirm this preserves correct behavior?

## Tested
- [x] smoke test


examples/scanvi/scanvi.py
```diff
@@ -197,7 +197,8 @@ def model(self, x, y=None):
         # wrt the number of datapoints and genes
         with pyro.plate("batch", len(x)), poutine.scale(scale=self.scale_factor):
             z1 = pyro.sample("z1", dist.Normal(0, x.new_ones(self.latent_dim)).to_event(1))
-            y = pyro.sample("y", dist.OneHotCategorical(logits=x.new_zeros(self.num_labels)))
+            y = pyro.sample("y", dist.OneHotCategorical(logits=x.new_zeros(self.num_labels)),
+                            obs=y)
```
martinjankowiak, line 201: yikes

PR #2674, opened 2020-10-16
Add more collapse tests

fritzo:
1. Removes xfail markers from tests fixed by https://github.com/pyro-ppl/funsor/pull/378
2. Adds two normal-normal test, including one with a plate.

We should also be able to add these to NumPyro.


eb8680 (approved):
Nice!

PR #2676, opened 2020-10-19
Return funsor.Variable from pyro.sample inside collapse

fritzo:
Addresses https://github.com/pyro-ppl/funsor/issues/367#issuecomment-711450825

This changes `poutine.collapse` to more eagerly convert `msg["fn"]` and `msg["value"]` to Funsors, and thereby allows `pyro.sample` statements to return `funsor.Variable`s rather than strings. This is important e.g. in @fehiepsi's use case where he transforms one of those returned variables.

@fehiepsi I tried to get this working with NumPyro but NumPyro is a little behind the Pyro version. I think you might need the `._block` logic to avoid converting the final `pyro.factor` statement to a funsor expression, but I'm not sure why I couldn't get it working.

## Tested
- [x] added a type assertion
- [x] added smoke tests with an enclosing particle plate

### DIFF
--- a/pyro/poutine/collapse_messenger.py
+++ b/pyro/poutine/collapse_messenger.py
@@ -86,7 +86,7 @@ class CollapseMessenger(TraceMessenger):
     .. warning:: This is not compatible with automatic guessing of
         ``max_plate_nesting``. If any plates appear within the collapsed
         context, you should manually declare ``max_plate_nesting`` to your
-        inference algorithm (e.g. ``Trace_ELBO(max_plate_nesting=1)``.
+        inference algorithm (e.g. ``Trace_ELBO(max_plate_nesting=1)``).
     """
     _coerce = None
 
--- a/pyro/poutine/collapse_messenger.py
+++ b/pyro/poutine/collapse_messenger.py
@@ -4,8 +4,8 @@
 from functools import reduce, singledispatch
 
 import pyro
-from pyro.poutine.util import site_is_subsample
 from pyro.distributions.distribution import COERCIONS
+from pyro.poutine.util import site_is_subsample
 
 from .runtime import _PYRO_STACK
 from .trace_messenger import TraceMessenger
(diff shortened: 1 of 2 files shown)


pyro/poutine/collapse_messenger.py
```diff
@@ -109,8 +109,13 @@ def _process_message(self, msg):
     def _pyro_sample(self, msg):
         if self._block:
             return
+        dim_to_name = {f.dim: f.name for f in msg["cond_indep_stack"]}
```
fritzo, line 112: @eb8680 I'm not sure we'll have the complete `cond_indep_stack` here. Won't it be missing any plates enclosing the `poutine.collapse`? I'm not sure why tests are passing. Do you think we could handle this by promoting `collapse.__enter__()` to be effectful and recording those plates in a `IndepMessenger._pyro_collapse()` method?

eb8680: We won't have access to all the plates via `msg["cond_indep_stack"]` at this point, but I believe all the missing ones are already contained in `self.preserved_plates`?

eb8680: Also, what's the reason we need to compute `to_funsor(msg["fn"])` in `_pyro_sample` rather than `_pyro_post_sample`?  Is it just so that we can compute `domain`? If so, would it be easier to leave `msg["fn"]` to `_pyro_post_sample` and compute `domain` directly in `_pyro_sample`?

fehiepsi:
> might need the ._block logic to avoid converting the final pyro.factor statement to a funsor expression

Yes, adding `block` logic works for me. :)

pyro/poutine/collapse_messenger.py
fritzo: Good point, I've refactored `self.preserved_plates` to contain needed information, so this should now work.

Correct, the reason we compute `to_funsor(msg["fn"])` in `_pyro_sample()` is indeed to infer value domain. This actually seems like the cheapest surefire way to compute value domain, and we'll need to compute `to_funsor(msg["fn"])` anyway, so I'm happy with that computation living in `_pyro_sample()`. We could move it to `_pyro_post_sample()`, but that seems more complex and equivalent, since in `_pyro_sample()` we set `msg["stop"] = True` and immediately pass control to `_pyro_post_sample()`.

fehiepsi (approved):
Thanks for making this possible! LGTM pending further comments by @eb8680 .

eb8680 (approved):
LGTM

PR #2679, opened 2020-10-22
Fix TraceEnum_ELBO.compute_marginals() bug #2678

### LINKED ISSUE #2678: Elbo.compute_marginals throws an error for basic model
As @Bart mentioned in this forum question :
https://forum.pyro.ai/t/elbo-compute-marginals-throws-an-error-for-basic-model/2245

When observed variables has no enumerated ancestors, this error of compute_marginals will arise:
"ValueError: Number of einsum subscripts must be equal to the number of operands."

Now this error can be fixed by adding a dummy upstream discrete node, but a formal fix is needed.

fritzo:
Fixes #2678 

## Tested
- [x] added a regression test

### DIFF
--- a/pyro/infer/traceenum_elbo.py
+++ b/pyro/infer/traceenum_elbo.py
@@ -133,15 +133,13 @@ def _compute_model_factors(model_trace, guide_trace):
                 cost = site["packed"]["masked_log_prob"]
                 log_factors.setdefault(t, []).append(cost)
                 scales.append(site["scale"])
-    if log_factors:
-        for t, sites_t in enum_sites.items():
-            # TODO refine this coarse dependency ordering using time and tensor shapes.
-            if any(t <= u for u in log_factors):
-                for site in sites_t:
-                    logprob = site["packed"]["unscaled_log_prob"]
-                    log_factors.setdefault(t, []).append(logprob)
-                    scales.append(site["scale"])
-        scale = _get_common_scale(scales)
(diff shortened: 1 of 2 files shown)


pyro/infer/traceenum_elbo.py
```diff
@@ -133,15 +133,13 @@ def _compute_model_factors(model_trace, guide_trace):
                 cost = site["packed"]["masked_log_prob"]
                 log_factors.setdefault(t, []).append(cost)
                 scales.append(site["scale"])
-    if log_factors:
-        for t, sites_t in enum_sites.items():
-            # TODO refine this coarse dependency ordering using time and tensor shapes.
-            if any(t <= u for u in log_factors):
```
fritzo, line 139: The only change is removing line 139. @eb8680 do you have any idea what this line does?

eb8680: Isn't it getting rid of (some , hence the TODO) zero-expectation terms in guide enumeration?

fritzo: Yeah I think it was merely an optimization, but it looks like it can lead to breakage. I'm pretty sure it is correct to add it back. And tests pass 🤷

eb8680 (approved):
Ok, I'm happy to merge this since it seems like an edge case, we can always try to fix it more carefully if someone reports a mysterious guide enumeration regression.

PR #2682, opened 2020-10-28
Attempt to fix help(MyDistribution)

### LINKED ISSUE #2681: Init signature does not appear for distribution class
Currently, when creating an instance of `dist.Normal` in jupyter, the docstring will say
```
Init signature: dist.Normal(*args, **kwargs)
```
That infomation is not helpful. This new behavior happens because we define `DistributionMeta.__call__` for collapse messenger. It would be nice to find an alternative solution that preserves the signature of distributions' constructors.

fritzo:
Fixes #2681 

@fehiepsi this seems like a hack, but I believe this is an intended use of the [`.__wrapped__` attribute](https://docs.python.org/3.6/library/inspect.html#inspect.signature). WDYT?

This attempts to fix e.g. `help(dist.Normal)`. Before this PR:
```
$ python -c 'import pyro.distributions as d; help(d.Normal)' | head -n 4
Help on class Normal in module pyro.distributions.torch:

class Normal(torch.distributions.normal.Normal, pyro.distributions.torch_distribution.TorchDistributionMixin)
 |  Normal(*args, **kwargs)
```
After this PR:
```
$ python -c 'import pyro.distributions as d; help(d.Normal)' | head -n 4
Help on class Normal in module pyro.distributions.torch:

class Normal(torch.distributions.normal.Normal, pyro.distributions.torch_distribution.TorchDistributionMixin)
 |  Normal(loc, scale, validate_args=None)
```

### DIFF
--- a/pyro/distributions/distribution.py
+++ b/pyro/distributions/distribution.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2017-2019 Uber Technologies, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
+import functools
 from abc import ABCMeta, abstractmethod
 
 from pyro.distributions.score_parts import ScoreParts
@@ -15,6 +16,10 @@ class DistributionMeta(ABCMeta):
             if result is not None:
                 return result
         return super().__call__(*args, **kwargs)
+
+    @property
+    def __wrapped__(cls):
+        return functools.partial(cls.__init__, None)
 
 
 class Distribution(metaclass=DistributionMeta):


fehiepsi (approved):
Thanks for a quick fix, @fritzo! (I tried to play around with `__init__`, `__new__`, `__signature__`,... not none of them works) This also works for the autocomplete mechanism in jupyterlab. Do you think that it is worth to have a patch release for this? I think many users will suffer from this issue (it is inconvenient to remember the signature of all distributions).

fritzo:
I don't know whether this is worth a patch release. Maybe we wait a week and see if any other fixes come up?

fritzo:
But @fehiepsi feel free to release Pyro 1.5.1 if you think it is worth it 🙂

fritzo:
Actually, PyTorch 1.7 was just released. Why don't we fix any 1.7 bugs and then release Pyro 1.5.1?

PR #2684, opened 2020-10-30
Add log_likelihood helper utility

ecotner:
Opening this following a [short discussion on the forum](https://forum.pyro.ai/t/log-likelihood-helper-function/2328) with @fehiepsi. The goal is to provide a quick utility that can be used to calculate the log likelihood of a model without the user having to dive into the model trace, which can be used to calculate other quantities of interest, like WAIC, PSIS-LOO, etc. I tried to style it so that it would have a similar interface to `Predictive`, and takes some inspiration from a [similar calculation in `arviz`](https://github.com/arviz-devs/arviz/blob/master/arviz/data/io_pyro.py#L136) for vectorizable models (though this implementation should be able to handle non-vectorizable models too). I have a [demo notebook here](https://gist.github.com/ecotner/379f41698e51bfbaac266b99a287f9c3) to show how it behaves at the moment.

I'm sure there is more work to be done :wink: (tests to write, bugs to work out, design choices to be debated), but figured it would be best to get this out there now so I can start getting feedback!


fehiepsi (commented):
Sorry @ecotner, I missed your PR. Some of my first comments are:
+ This utility should be at `pyro.infer`. Probably at `pyro/infer/predictive.py` file? The `stats` module is used for tensor operators.
+ There are several assumptions that can be relaxed:
  - the likelihood is 2D tensor
  - there is only 1 observation site

Because you return a callable function, what about making `log_likelihood` a class like `Predictive`? I am not sure what is the best api here. In numpyro, we have a function [log_likelihood](http://num.pyro.ai/en/stable/utilities.html#numpyro.infer.util.log_likelihood) utility, which does the same thing but has simpler api (no `guide`, `num_samples`,...). We still not yet come up with a decision on what is a good predictive interface for Pyro/NumPyro. So we marked those helper "EXPERIMENTAL INTERFACE". :D

ecotner:
> This utility should be at pyro.infer. Probably at pyro/infer/predictive.py file? The stats module is used for tensor operators.

Yes, I can move it to the `infer` module, that makes sense.

> There are several assumptions that can be relaxed: 1. the likelihood is 2D tensor

So my initial thought on that was that when you calculate likelihood, it is typically on a per-sample basis, so that you would produce a tensor of shape `(number of posterior samples, number of independent data points)`. After looking at the implementation in `numpyro` you linked though, I do like the flexibility of going more granular by considering the likelihood for each observed site individually, and generalizing from `number of independent data points` -> `batch_shape`. So perhaps the best output would be a dictionary mapping observed site name to a log likelihood tensor of shape `(number posterior samples,) + batch_shape`?

It seems every time I suggest something for `pyro` there is already better implementation in `numpyro`; maybe I should just start using that package instead :laughing:. It is supposed to be faster too, right?

> 2. there is only 1 observation site

I think as written it should be able to handle multiple observed sites; I iterate over a dictionary of observed site names/values. It is just that right now I then _sum_ over all the observed sites to get a total likelihood rather than returning them for individual sites. My intended use-case was to generate input for `waic`, which accepts a total log-likelihood tensor as input, not per-site likelihood. But I agree with the sentiment that per-site likelihood would be valuable.

> what about making `log_likelihood` a class like `Predictive`

I do like the calling signature/API for `Predictive` (the clear separation of model arguments from the other input makes a lot of sense), so that is what I was trying to imitate, but stopped short of creating a class because it seems that the operation of extracting the likelihood is stateless. Once you provide the model, posterior samples (or guide), and model arguments/data, you can immediately calculate the likelihood, and there doesn't seem to be any interesting intermediate bits that you might want to keep around (besides maybe the posterior samples if you generate them from a guide rather than directly provide the samples). It seems that `Predictive` _almost_ suffers from the same problem as well; there is nothing going on in `Predictive.__init__` other than input validation, and `call` simply calls `forward` and does a bit of output formatting. It could all be turned into a single function. (Except for `get_vectorized_trace`; that clearly provides some additional functionality.)

Those are just some of my thoughts; I'm not 100% sure on the API either though.

fehiepsi:
Yeah, API for Predictive has several versions back and forth since the beginning. It has been used in many contrib modules (which needs predictions) since then. If I remember correctly, those utilities are needed when we tried to support Pyro/NumPyro for ArviZ (to get posterior predictive, prior predictive, waic,...), which is a pretty nice package for sample and model diagnostics. For `log_likelihood`, I think you don't have to worry much about the API for now. I'll ping some other devs to provide more opinions on it later. :)

> It is supposed to be faster too

Yes and no. For complicated models, if some errors happen, it is harder to debug in my experience.

ecotner:
@fehiepsi should I just close the PR and wait until we have a consensus on what the API should look like then?

pyro/infer/predictive.py
```diff
@@ -219,3 +220,71 @@ def get_vectorized_trace(self, *args, **kwargs):
...
+
+
+def log_likelihood(
+    model: Callable,
+    posterior_samples: Dict[str, torch.Tensor] = None,
+    guide: Callable = None,
+    num_samples: int = None,
+    parallel: bool = True,
+) -> Callable:
+    """Calculates the log likelihood of the data for a model with observed
```
fehiepsi, line 232: Could you add an EXPERIMENTAL flag here? (like in [this docs](http://docs.pyro.ai/en/stable/primitives.html#pyro.primitives.deterministic))

pyro/infer/predictive.py
```diff
@@ -219,3 +220,71 @@ def get_vectorized_trace(self, *args, **kwargs):
...
+            log_like = dict()
+            trace = predictive.get_vectorized_trace(*args, **kwargs)
+            for obs_name, obs_val in observed.items():
+                obs_site = trace.nodes[obs_name]
+                log_like[obs_name] = obs_site["fn"].log_prob(obs_val).detach().cpu()
+        # iterate over samples from posterior if model can't be vectorized
+        else:
+            all_samples = {
+                k: v for k, v in predictive(*args, **kwargs).items()
+                if k not in observed}
```
fehiepsi, line 272: @ecotner Sorry for late response!

I am a bit worried that the implementation is tracing the model 2 times (1 to get samples, 1 to get the trace of each sample). But this seems pretty clean to me. Please feel free to add a TODO here or if you want to address it, you can add an argument in `_predictive_sequential`, e.g. `collect_fn=None`, so that if `collect_fn` is not None, `_predictive` will return the collection of values obtained by `collect_fn`. For example, to get log_prob at observed sites, in `log_likelihood`, we can use
```python
collect_fn = lambda msg: msg["fn"].log_prob(msg["value"])
```
and replace line 43 of this file by
```python
collected.append({site: collect_fn(trace.nodes[site]) for site in return_site_shapes})
```

ecotner: ooh that's a great idea!

pyro/infer/predictive.py
```diff
@@ -255,36 +268,25 @@ def log_like_helper(*args, **kwargs) -> torch.Tensor:
...
-            # each sample is a tensor in a list; stack all the elements into
-            # a single tensor for each site
-            log_like = {
-                obs_name: torch.stack(ll_list, dim=0)
-                for obs_name, ll_list in log_like.items()
+            site_shapes = {
+                name: (num_samples,) + trace.nodes[name]["value"].shape
+                for name in observed
             }
+            log_like = _predictive_sequential(
```
fehiepsi, line 296: In case `guide` is not None (like in `Predictive` implementation), I think you need
```
posterior_samples = _predictive(guide, {}, num_samples, return_sites=None,
                                parallel=False, model_args=args, model_kwargs=kwargs)
```

It seems that you need to add some logic to get `num_samples` or assertations as in `Predictive` constructor. Do you think that supporting `guide`, `num_samples` makes things complicated?

pyro/infer/predictive.py
```diff
@@ -255,36 +268,25 @@ def log_like_helper(*args, **kwargs) -> torch.Tensor:
...
-                    obs_site = trace.nodes[obs_name]
-                    ll_sample = obs_site["fn"].log_prob(obs_val).detach().cpu()
-                    log_like[obs_name].append(ll_sample)
-            # each sample is a tensor in a list; stack all the elements into
-            # a single tensor for each site
-            log_like = {
-                obs_name: torch.stack(ll_list, dim=0)
-                for obs_name, ll_list in log_like.items()
+            site_shapes = {
+                name: (num_samples,) + trace.nodes[name]["value"].shape
```
fehiepsi, line 293: I think we need to use `name: (num_samples,) + trace.nodes[name]["fn"].batch_shape` or to make things safe, we need to broadcast `fn.batch_shape` and `value.shape[:len(value.shape) - fn.event_dim]`, where `value = trace.nodes[name]["value"]` and `fn = trace.nodes[name]["fn"]`. We need to add some smoke test for likelihoods with `event_dim != 0` (e.g. MultivariateNormal or dist.Normal(...).to_event(1)) to catch this issue.

ecotner: Does [this test](https://github.com/pyro-ppl/pyro/pull/2684/files#diff-8d729ab9d44de7289fe0dd334f312a5257b0f03ca91ccf1a5b6b549fdb6a3c1eR161-R181) cover that possibility? It seems like I didn't need to make use of the broadcasting trick to get it to work.

fehiepsi: Yes, it does. Thanks! Broadcasting is only useful in case
```
pyro.sample("a", dist.Normal(0, 1).expand([10]), obs=torch.tensor(1.))
```
like in the following program
```python
def model():
    x = pyro.sample("x", dist.Normal(0, 1).expand([10]))
    pyro.sample("obs", dist.Normal(x, 1), obs=torch.tensor(0.))

mcmc = MCMC(NUTS(model), 1000)
mcmc.run()
```

pyro/infer/predictive.py
```diff
@@ -219,3 +231,82 @@ def get_vectorized_trace(self, *args, **kwargs):
...
+
+    ```
+    x = torch.randn(10)
+    y = x + 0.05 * torch.randn(10)
+    cond_model = pyro.condition(model, data={"y": y})
+    ll_vals = log_likelihood(cond_model, guide=guide, num_samples)(x)
+    ```
+    """
+
+    _validate_predictive_inputs(posterior_samples, num_samples, guide, return_sites=None)
```
fehiepsi, line 271: Should we also return `posterior_samples` and `num_samples` here?

pyro/infer/predictive.py
```diff
@@ -219,3 +231,82 @@ def get_vectorized_trace(self, *args, **kwargs):
...
+        # use vectorized trace if parallelizable
+        if parallel:
+            log_like = dict()
+            # now trace it and extract the likelihood from observed sites
+            predictive = Predictive(
+                model, posterior_samples, guide, num_samples, (), parallel)
+            trace = predictive.get_vectorized_trace(*args, **kwargs)
+            for obs_name in observed.keys():
+                site = trace.nodes[obs_name]
+                log_like[obs_name] = site["fn"].log_prob(site["value"]).detach().cpu()
```
fehiepsi, line 293: nit: Probably we don't need to detach here (in NumPyro, we use `log_likelihood` in some algorithms that require taking `gradient`). But you can keep it and and a note for it in the docs and mention that this behavior might be changed in the future.

pyro/infer/predictive.py
```diff
@@ -219,3 +231,82 @@ def get_vectorized_trace(self, *args, **kwargs):
...
+    ll_vals = log_likelihood(cond_model, guide=guide, num_samples)(x)
+    ```
+    """
+
+    _validate_predictive_inputs(posterior_samples, num_samples, guide, return_sites=None)
+
+    # define the function to return the likelihood values
+    def log_like_helper(*args, **kwargs) -> torch.Tensor:
+        # trace model once to get observed nodes
+        trace = pyro.poutine.trace(model).get_trace(*args, **kwargs)
```
fehiepsi, line 276: You can avoid tracing the model to get the observed node by using
```
collect_fn=lambda msg: msg["fn"].log_prob(msg["value"]) if msg["is_observed"] else None
```

pyro/infer/predictive.py
ecotner: oh, I see what you mean now

pyro/infer/predictive.py
ecotner: I'm not sure that I can get rid of the trace in the parallel branch because it seems that `get_vectorized_trace` is returning `site["is_observed"] == True` even for sites that are not actually observed. I replaced
```python
for obs_name in observed.keys():
    site = trace.nodes[obs_name]
    log_like[obs_name] = site["fn"].log_prob(site["value"])
```
with
```python
for obs_name, site in trace.nodes.items():
    if site["type"] == "sample" and site["is_observed"]:
        log_like[obs_name] = site["fn"].log_prob(site["value"])
```
and my tests start failing because there are more observed nodes being returned than expected. It seems the problem is the following line in `_predictive`:
```python
if return_trace:
    trace = poutine.trace(poutine.condition(vectorize(model), reshaped_samples))\
        .get_trace(*model_args, **model_kwargs)
    return trace
```
where once the model is conditioned on the samples, then the trace thinks that they are now observed. Perhaps I can just extract the parts of `_predictive` that are necessary for getting the trace and modify this part?

ecotner: Also, in the serial branch, I think we need to do a trace before passing to `_predictive_sequential` so that we can get `site_shapes` to pass in as an argument. Maybe it might be better to unpack the functionality that we need from `_predictive_sequential` and rewrite this part too?

ecotner: It looks like each call to `get_vectorized_trace` already traces the model three times (in `_predictive`, once to guess plate nesting, once to prune subsamples, third time to actually return the trace); six times total if you provide a guide instead of samples. Maybe adding one more trace isn't so bad haha.

ecotner: re: my first comment - it looks like this behavior would also occur in `_predictive_sequential` as well; there are some lines like
```python
for i in range(num_samples):
    trace = poutine.trace(poutine.condition(model, samples[i])).get_trace(*model_args, **model_kwargs)
    if return_trace:
        collected.append(trace)
    else:
        collected.append({site: collect_fn(trace.nodes[site]) for site in return_site_shapes})
```
the only thing that prevents this from returning extra observed sites at the moment is that `return_site_shapes` only has keys from observed nodes (which are gathered from the trace in `log_likelihood_helper`) so the extras get filtered out in the last line

fehiepsi: Whoa, it is tricky... Thanks for explaining, @ecotner! Let's keep this as-is. :)

Sorry for making confusion!

ecotner: no worries; no confusion here. it is just a tough problem to solve haha. I think a genuine solution would involve rewriting large parts of `_predictive` and `_predictive_sequential` though, and I worry about accidentally breaking any downstream functionality that relies on this (like `Predictive` or utilities in `arviz`).

ecotner: oh, you know what would be a real simple fix? we could just have the user pass in the `observed` dictionary themselves as an optional argument. if it is not present, then we do the extra trace to generate it.

pyro/infer/predictive.py
ecotner: I added [some test](https://github.com/pyro-ppl/pyro/pull/2684/commits/52adf13cb9bfd61290fb9bb5d6efbdc9ae83fc8d#diff-8d729ab9d44de7289fe0dd334f312a5257b0f03ca91ccf1a5b6b549fdb6a3c1eR196-R216) to verify broadcasting works; it seems to work naturally for the `Normal` distribution.

pyro/infer/predictive.py
fehiepsi: I think it is better to keep the previous interface. There are several ways to get around this issue but it seems unnecessary here (in NumPyro, we also support computing log likelihood for 1 sample so avoiding tracing many times is needed - there, we use [substitute](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/infer/util.py#L615) rather than `condition`, so `is_observed` is not overwritten after substituted).

If you want to avoid tracing here, then you can use [infer_config_messenger](https://github.com/pyro-ppl/pyro/blob/dev/pyro/poutine/infer_config_messenger.py) to add `_is_observed` keyword to `msg["infer"]`. Then, in `collect_fn`, you only compute log prob if `msg["infer"].get("_is_observed") is True`. Similarly, in `parallel` branch, after getting the vectorized trace, you only compute log prob at sites that satisfy `msg["infer"].get("_is_observed") is True`. Does this make sense at all? I think it will work but not test it yet. Please let me know if you want me to try some code here. :)

ecotner: yeah that makes sense. is `substitute` basically the same thing as [`do`](http://docs.pyro.ai/en/stable/poutine.html#pyro.poutine.handlers.do)? could an alternative be to possibly replace the `condition` calls in `_predictive` and `_predictive_sequential` with `do`? yeah feel free to experiment if you want!

fehiepsi: Yes, it works like `do` but does not introduce a fresh sample site. I think it is a bit expensive here if we use `do`.

fritzo:
@ecotner @fehiepsi sorry this fell off the radar. Is it ready for review?

fehiepsi:
@fritzo @neerajprad The implementation looks good to me but we need to discuss the API. Currently, in this PR
+ `log_likelihood` is a function generator, rather than a class, with a similar signature as `Predictive`:
```
loglik_fn = log_likelihood(`predictive's args, kwargs`)
loglikelihood = loglik_fn(*model_args, **model_kwargs)
```
+ This also supports `guide` input, which will draw posterior samples as in Predictive, then compute the likelihoods for those posterior samples.
+ `obs_shapes` is used to avoid tracing the model to inspect shapes at observation sites.

Could you give some thoughts on this? The current API in NumPyro (put `model_args` to the input of `log_likelihood` to make it a function, rather than a function generator, no `guide`, no `obs_shapes`) looks good to me but I am happy to change it to match Pyro API if we agree with some parts of the API here.

ecotner:
@fritzo I'm sorry too; I got distracted with a [side project](https://ecotner.github.io/visualizations/quantum/2d_square_well/) and kind of forgot about this! I think in addition to the API design choice like @fehiepsi mentioned, there are some additional optimizations that might be possible in order to reduce the number of traces being made, but I think that could potentially require large changes to or parallel implementations of `_predictive` and/or `_predictive_sequential` because `log_likelihood` relies on them to do some of the heavy lifting, and they each generate multiple traces. Could be a case of "premature optimization" though; I haven't benchmarked this on any larger models, but it seems to run pretty quick on the small ones in the tests.

pyro/infer/predictive.py
fehiepsi: @ecotner I think it is better to use the previous API and leave comments for optimization in future PRs because they are not important if we use log_likelihood for a large number of samples (says 1000).

fritzo:
@ecotner there's no rush, I just want to make sure you're not blocked on review. Do you want me to mark this WIP and you can resume work whenever you want?

PR #2685, opened 2020-11-03
Use GPyTorch in SpanningTree.log_prob()

fritzo:
This uses GPyTorch's faster more stable approximate `.log_det()` computation in place of `torch.cholesky(-).diag().log().sum()`.

## Tested
- [x] ran locally with GPyTorch installed

### DIFF
--- a/pyro/distributions/spanning_tree.py
+++ b/pyro/distributions/spanning_tree.py
@@ -120,7 +120,11 @@ class SpanningTree(TorchDistribution):
         adjacency[grid[1], grid[0]] = edge_probs
         laplacian = adjacency.sum(-1).diag() - adjacency
         truncated = laplacian[:-1, :-1]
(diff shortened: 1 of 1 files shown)


fritzo:
@jacobrgardner we hope to use GPyTorch + KeOps for some scalable tree counting via Kirchhoff's theorem. Do you have any tips or gotchas we should look out for when using `.logdet()` on say a 10^5 x 10^5 matrix? Thanks!

eb8680 (approved):
LGTM. Do you want to add Gpytorch as an optional dependency?

jacobrgardner:
Shouldn't be any gotchas there. At 10000x10000 you're likely not even going to need KeOps yet.

fritzo:
@eb8680 I don't see a need to add an optional dependency.

PR #2686, opened 2020-11-03
Update contributing and release docs

fritzo:


fehiepsi (approved):
LGTM. Do you want to update the author's email in [setup](https://github.com/pyro-ppl/pyro/blob/dev/setup.py) file too?

fritzo:
@fehiepsi thanks, I've removed emails from setup.py also. Github seems like the right place for communication, rather than someone's personal email.

eb8680 (approved):

Issue #2688, opened 2020-11-03
[Doc] Inconsistent of param signature between AutoIAFNormal and affine_autoregressive


xidulu:
Doc for AutoIAFNormal (http://docs.pyro.ai/en/stable/infer.autoguide.html#autoiafnormal) says hidden_dims is an int
While http://docs.pyro.ai/en/stable/distributions.html?highlight=affine_autoregressive#affine-autoregressive claims that hidden_dims should be a list of int.

Is that a typo in AutoIAFNormal's doc?

fritzo:
cc @stefanwebb

stefanwebb:
@xidulu thanks for picking this up! Yes, I believe this is a bug in the docstring for `AutoIAFNormal`. It should be `list[int]`. If you passed in an `int` for `hidden_dims` to `AutoIAFNormal` it would raise an exception when it gets to this code in `ConditionalAutoregressiveNN`:

```
        # Hidden dimension must be not less than the input otherwise it isn't
        # possible to connect to the outputs correctly
        for h in hidden_dims:
            if h < input_dim:
                raise ValueError('Hidden dimension must not be less than input dimension.')
```
It's not being caught by the unit tests since they use the default argument of `hidden_dims=None`.

stefanwebb:
Are you able to submit a PR for this quick change?

stefanwebb:
@martinjankowiak you designed the unit tests for autoguides, right? Do you think it would be worth adjusting the unit tests slightly to test non-default keyword args?

martinjankowiak:
it seems like all our autoguide unit tests are integration tests; we could probably use some simple smoke tests like we have for [easyguide](https://github.com/pyro-ppl/pyro/blob/f3494ac9e2428b36d6840e4d3a6dbd31daaf4d30/tests/contrib/easyguide/test_easyguide.py#L102)

xidulu:
> Are you able to submit a PR for this quick change?

Sure，I can do that

xidulu:
Fixed by #2691

PR #2691, opened 2020-11-05
fix AutoIAF doc

xidulu:
As title
https://github.com/pyro-ppl/pyro/issues/2688


eb8680 (approved):
Thanks for fixing this!

PR #2692, opened 2020-11-06
Stabilize math in SpanningTree.log_partition_function

fritzo:
This makes the numerical stabilization math more fine-grained in `SpanningTree.log_partition_function`.

Previously we shifted by a max over all edges. After this PR we shift by column sums. This is advantageous because the resulting Laplacian matrix has constant 1's on the diagonal and is safer for the Cholesky factorization in `.logdet()`.

Could a reviewer please go over the math in detail to ensure gradients are correct? Thanks in advance!

## Tested
- [x] fixes nan issues in a gradient-based inference that uses `.log_partition_function`

### DIFF
--- a/pyro/distributions/spanning_tree.cpp
+++ b/pyro/distributions/spanning_tree.cpp
@@ -142,7 +142,7 @@ at::Tensor sample_tree_approx(at::Tensor edge_logits) {
   // the complete graph. The id of an edge (v1,v2) is k = v1+v2*(v2-1)/2.
   auto edge_ids = torch::empty({E}, at::kLong);
   // This maps each vertex to whether it is a member of the cumulative tree.
-  auto components = torch::zeros({V}, at::kByte);
+  auto components = torch::zeros({V}, at::kBool);
 
   // Sample the first edge at random.
   auto probs = (edge_logits - edge_logits.max()).exp();
--- a/pyro/distributions/spanning_tree.py
+++ b/pyro/distributions/spanning_tree.py
@@ -112,20 +112,21 @@ class SpanningTree(TorchDistribution):
         # use a Cholesky decomposition to compute the log determinant.
         # See https://en.wikipedia.org/wiki/Kirchhoff%27s_theorem
         V = self.num_vertices
-        grid = make_complete_graph(V)
-        shift = self.edge_logits.detach().max()
-        edge_probs = (self.edge_logits - shift).exp()
-        adjacency = torch.zeros(V, V, dtype=edge_probs.dtype)
-        adjacency[grid[0], grid[1]] = edge_probs
-        adjacency[grid[1], grid[0]] = edge_probs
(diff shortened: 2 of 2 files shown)


pyro/distributions/spanning_tree.cpp
```diff
@@ -142,7 +142,7 @@ at::Tensor sample_tree_approx(at::Tensor edge_logits) {
   // the complete graph. The id of an edge (v1,v2) is k = v1+v2*(v2-1)/2.
   auto edge_ids = torch::empty({E}, at::kLong);
   // This maps each vertex to whether it is a member of the cumulative tree.
-  auto components = torch::zeros({V}, at::kByte);
+  auto components = torch::zeros({V}, at::kBool);
```
fritzo, line 145: this avoids a deprecation warning in PyTorch 1.7

pyro/distributions/spanning_tree.py
```diff
@@ -112,20 +112,21 @@ def log_partition_function(self):
         # use a Cholesky decomposition to compute the log determinant.
         # See https://en.wikipedia.org/wiki/Kirchhoff%27s_theorem
         V = self.num_vertices
-        grid = make_complete_graph(V)
-        shift = self.edge_logits.detach().max()
-        edge_probs = (self.edge_logits - shift).exp()
```
martinjankowiak, line 117: these are normalized along each column/row?

pyro/distributions/spanning_tree.py
```diff
@@ -112,20 +112,21 @@ def log_partition_function(self):
...
-        adjacency[grid[1], grid[0]] = edge_probs
-        laplacian = adjacency.sum(-1).diag() - adjacency
+        v1, v2 = make_complete_graph(V).unbind(0)
+        logits = self.edge_logits.new_full((V, V), -math.inf)
+        logits[v1, v2] = self.edge_logits
+        logits[v2, v1] = self.edge_logits
+        log_diag = logits.logsumexp(-1)
+        # Numerically stabilize so that laplacian has 1's on the diagonal.
+        shift = 0.5 * log_diag
+        laplacian = torch.eye(V) - (logits - shift - shift[:, None]).exp()
```
martinjankowiak, line 122: looks right to me. what's the deal with the truncation?

fritzo: That's just how the theorem goes: you delete any single column and row, then compute the determinant.
See https://en.wikipedia.org/wiki/Kirchhoff%27s_theorem#An_example_using_the_matrix-tree_theorem

pyro/distributions/spanning_tree.py
fritzo: That's right, I'm normalizing by column, but since the matrix is symmetric, I split that normalization to be half along rows and half along columns. The diagonal doesn't care whether it's all-column or all-row or half-and-half.

martinjankowiak (approved):

PR #2693, opened 2020-11-06
Implement a GumbelSoftmaxReparam

fritzo:
This adds a reparametrizer for the Gumbel-Softmax trick.

The reparametrizer replaces a d-dimensional `RelaxedOneHotCategorical` with a d-dimensional `Uniform(0,1)` distribution whose samples are first transformed to be Gumbel distributed and then passed through the softmax trick.

The motivation is that, when unimodal autoguides try to learn the posterior of a R.O.H.C. distribution they are only able to capture a single mode, i.e. a single category. By instead learning the posterior in pre-softmax space, a single mode can then be softmax-transformed to a multimodal posterior, with one mode per category. Thus this reparametrizer allows naive autoguides like `AutoNormal` to capture multimodal posteriors.

## Tested
- [x] added moment tests
- [x] smoke tested in [a practical model](https://github.com/broadinstitute/pyro-cov/blob/816b8eba116bcc77a367aad7fdf736b09e87167e/pyrophylo/kirchhoff.py#L93)

### DIFF
--- a/docs/source/infer.reparam.rst
+++ b/docs/source/infer.reparam.rst
@@ -32,6 +32,15 @@ Loc-Scale Decentering
     :special-members: __call__
     :show-inheritance:
 
+Gumbel-Softmax
+--------------
+.. automodule:: pyro.infer.reparam.softmax
+    :members:
+    :undoc-members:
+    :member-order: bysource
+    :special-members: __call__
+    :show-inheritance:
+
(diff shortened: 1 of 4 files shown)


martinjankowiak (approved):

fritzo:
Thanks for reviewing!

PR #2697, opened 2020-11-12
Add a OneTwoMatching distribution

fritzo:
This implementation a distribution over perfect 1-2-matchings, i.e. matchings from 2*N sources to N destinations where each source matches exactly **one** destination and each destination matches exactly **two** sources.

Perfect 1-2-matchings are useful in phylogenetics, where random binary tree structures among N+1 leaves and N internal nodes with fixed times and genotypes reduce to random 1-2-matchings.

## Functionality in this PR
- [x] A `OneTwoMatchingConstraint` to validate samples
- [x] A general `.enumerate_support()` method, mostly for testing. (Warning this blows up quickly!)
- [x] An exact `.log_prob()` method when `bp_iters=None` (Warning this blows up quickly)
- [x] An approximate `.log_prob()` method when `bp_iters` is an int. This approximation uses loopy belief propagation to compute a Bethe approximation of the log partition function.
- [x] A `.mode()` method to find the maximum probability matching

## Deferred for future PRs
- An approximate `.sample()` method (I'm unsure how to implement this)

## Tests
- [x] unit tests for `.enumerate_support()`
- [x] unit tests for exact `.log_prob()`
- [x] approximate tests for Bethe `.log_prob()`
- [x] smoke tests for large computations, with autograd
- [x] unit tests for `.mode()`
- [x] unit tests for `.sample()`
- [x] trained in a real model using SVI ([private repo](https://github.com/broadinstitute/pyro-cov/pull/3))


fritzo:
@martinjankowiak I was able to improve accuracy by using the fractional free energy of [Chertkov & Yedidia (2013)](https://jmlr.org/papers/volume14/chertkov13a/chertkov13a.pdf). Most test cases now compute Z to within one nat :smile:

fritzo:
Yay, I've fixed the pairwise Bethe free energy and now all results are within one nat of truth, even for larger problem sizes that are too expensive for unit tests 😄

tests/distributions/test_one_two_matching.py
```diff
@@ -0,0 +1,113 @@
...
+@pytest.mark.parametrize("bp_iters", [None, BP_ITERS], ids=["exact", "bp"])
+def test_log_prob_full(num_destins, dtype, bp_iters):
+    num_sources = 2 * num_destins
+    logits = torch.randn(num_sources, num_destins, dtype=dtype) * 10
+    d = dist.OneTwoMatching(logits, bp_iters=bp_iters)
+    values = d.enumerate_support()
+    log_total = d.log_prob(values).logsumexp(0).item()
+    logging.info(f"log_total = {log_total:0.3g}, " +
+                 f"log_Z = {d.log_partition_function:0.3g}")
+    assert_close(log_total, 0., atol=min(num_destins, 1.0))
```
martinjankowiak, line 75: is this `min` as intended?

pyro/distributions/one_two_matching.py
```diff
@@ -0,0 +1,184 @@
...
+        p = (self.logits - shift).exp().clamp(min=finfo.tiny ** 0.5)
+        d = 2 / p.sum(0)
+        for _ in range(self.bp_iters):
+            s = 1 / (p @ d)
+            d = 2 / (s @ p)
+        b = s[:, None] * d * p
+
+        # Evaluate the Bethe free energy, adapting [4] Eqn 4 to one-two
+        # matchings.
+        b = b.clamp(min=finfo.tiny ** 0.5)
```
martinjankowiak, line 118: isn't `** 0.5` a bit strange?

martinjankowiak (approved):
lgtm although i didn't try to confirm every factor of 2. the tests are pretty convincing though!

tests/distributions/test_one_two_matching.py
fritzo: good catch, simplifed to 1.0.

pyro/distributions/one_two_matching.py
fritzo: not sure why `finfo.tiny ** 0.5` is strange? I wanted to clamp to something small such that its square is still nonzero. So `sqrt(finfo.tiny)` seemed like a reasonable choice. Also this avoided NAN grads in my model, whereas both the smaller `finfo.tiny` and the larger `finfo.eps` failed with NANs. Also I prefer to write `math.sqrt(x)` as `x ** 0.5` because it is shorter and doesn't require an import.

fritzo:
@martinjankowiak thanks for reviewing!  I'd like to add a `.mode()` method and a purturb-and-map implementation of `.sample()` in this PR before merging. I'd also like to try to wait for #2683 and Pyro's 1.5.1 release so that `OneTwoMatching` can mature a bit before release.

fritzo:
OK, ready to merge now that 1.5.1 has been released. Let me know if you want any further explanations.

martinjankowiak (approved):

PR #2701, opened 2020-11-23
Validate if __debug__ by default

### LINKED ISSUE #2700: Enable validation by default?
An issue I often see with new users is that they forget to enable validation when creating new models. Moreover I find that even in production models, validation is cheap enough to enable always. When speed matters, we usually JIT compile, and the JIT implementations generally disable validation, so enabling validation by default wouldn't slow down any of those users. Moreover all our tutorials specify `pyro.enable_validation() # Always a good idea.` so maybe we should act on that comment. Finally, when adding validation to our examples in #2701, I discovered a number of bugs, suggesting validation is indeed helpful.

Should we `enable_validation(True)` or `__debug__` by default in the next release?

cc @neerajprad @fehiepsi

fritzo:
Resolves #2700 

This effectively calls `pyro.enable_validation(__debug__)` so that validation is enabled by default unless Python is run in optimized mode (via `python -O`). This matches the behavior of the `assert` statement.

I have removed all uses of `pyro.enable_validation()` from examples and tutorials, since the default behavior is now preferred.

This caught a number of errors in examples and docs; this PR fixes those errors.
- type error in `poutine.scale` docs
- type error in mixed hmm example (as of recent PyTorch version)
- shape error in eight_schoos/svi.py


examples/eight_schools/svi.py
```diff
@@ -21,19 +21,20 @@ def model(data):
     y = data[:, 0]
     sigma = data[:, 1]
 
-    eta = pyro.sample('eta', dist.Normal(torch.zeros(J), torch.ones(J)))
-    mu = pyro.sample('mu', dist.Normal(torch.zeros(1), 10 * torch.ones(1)))
-    tau = pyro.sample('tau', dist.HalfCauchy(scale=25 * torch.ones(1)))
+    with pyro.plate("data", J):
```
fritzo, line 24: I've fixed this model by adding a plate.

examples/mixed_hmm/model.py
```diff
@@ -156,9 +156,10 @@ def model_generic(config):
 
                     # zero-inflation with MaskedMixture
                     step_zi = Vindex(step_zi_param)[..., y, :]
-                    step_zi_mask = pyro.sample("step_zi_{}".format(t),
-                                               dist.Categorical(logits=step_zi),
-                                               obs=(config["observations"]["step"][..., t] == MISSING))
+                    step_zi_mask = config["observations"]["step"][..., t] == MISSING
```
fritzo, line 159: I've fixed this model by distinguishing dtypes `bool` versus `long`

examples/neutra.py
```diff
@@ -43,12 +44,11 @@
 
 
 class BananaShaped(dist.TorchDistribution):
+    arg_constraints = {"a": constraints.positive, "b": constraints.real}
```
fritzo, line 47: I've fixed this model by adding an `.arg_constraints` class variable.

pyro/distributions/util.py
```diff
@@ -15,7 +15,8 @@
 
 from pyro.util import ignore_jit_warnings
 
-_VALIDATION_ENABLED = False
+_VALIDATION_ENABLED = __debug__
+torch_dist.Distribution.set_default_validate_args(__debug__)
```
fritzo, line 19: @neerajprad this line changes the default behavior of `torch.distributions`. It may lead to weird behavior since simply importing Pyro will turn on PyTorch validation. However this still seems like the best option to me.

pyro/poutine/scale_messenger.py
```diff
@@ -18,8 +18,7 @@ class ScaleMessenger(Messenger):
 
         >>> def model(x):
         ...     s = pyro.param("s", torch.tensor(0.5))
-        ...     z = pyro.sample("z", dist.Normal(x, s), obs=1.0)
-        ...     return z ** 2
+        ...     pyro.sample("z", dist.Normal(x, s), obs=torch.tensor(1.0))
```
fritzo, line 21: I've fixed this by ensuring the observation is a `Tensor`.

neerajprad (approved):

neerajprad:
The change looks good to me. My only minor concern would be that some users may find their earlier models slow down when they update Pyro, and might not know that they need to turn off validation explicitly to get the earlier behavior. This is more of an awareness problem that we can address as we go along. This will be a net benefit for our users, specially new ones. We should definitely highlight this in our changelog/release notes.

fritzo:
Thanks for reviewing @neerajprad !

PR #2706, opened 2020-12-03
Silence pyro.param warning if .training is False

### LINKED ISSUE #2705: PyroModule.requires_grad_(False).__call__() issues unwanted warning
After training a `PyroModule` model, I'd like to detach it and use it elsewhere in a "frozen" state. One obvious way to do so would be to call `my_module.requires_grad_(False)` to detach all parameters. Doing so results in the desired behavior, but has the side effect of triggering the following warning (many many times):
```
/Users/fobermey/github/pyro-ppl/pyro/pyro/primitives.py:361: UserWarning: linear.bias was not registered in the param store because requires_grad=False
  " requires_grad=False")
```

It is unclear to me how to resolve this. Should we provide a `detach_()` method for modules that converts all `PyroParam`s to `Tensor`s? Should we simply remove that warning if it occurs in a `PyroModule`? Should we be calling some other method like `my_module.train(False)`?

fritzo:
Fixes #2705 

This provides a mechanism to silence `pyro.param` warnings when running a detached `PyroModule` during training. It is a coarse-grained solution; this silences the warning for the module as a whole. To silence individual param warnings, users could convert from `PyroParam` to `Tensor`:
```py
value = my_module.my_param.detach()
del my_module.my_param  # deletes the PyroParam
my_module.my_param = value  # creates a new Tensor
```

### DIFF
--- a/pyro/primitives.py
+++ b/pyro/primitives.py
@@ -356,9 +356,10 @@ def module(name, nn_module, update_module_params=False):
 
             if param_value._cdata != returned_param._cdata:
                 target_state_dict[param_name] = returned_param
-        else:
-            warnings.warn("{} was not registered in the param store because".format(param_name) +
-                          " requires_grad=False")
+        elif nn_module.training:
+            warnings.warn(f"{param_name} was not registered in the param store "
+                          "because requires_grad=False. You can silence this "
+                          "warning by calling my_module.train(False)")
 
     if target_state_dict and update_module_params:
         # WARNING: this is very dangerous. better method?


fehiepsi (approved):
Looks reasonable to me!

> value = my_module.my_param.detach()
del my_module.my_param  # deletes the PyroParam
my_module.my_param = value  # creates a new Tensor

A bit verbose but I couldn't think of a better way.

fritzo:

PR #2710, opened 2020-12-07
Stabilize Bethe approximation in matching distributions

fritzo:
This improves numerical stability of gradient computations in `OneOneMatching` and `OneTwoMatching` by switching their Sinkhorn computations from linear coords to log coords (using `logsumexp`). This reduces the amount of clamping and empirically improves the gradients in large phylogenetic models.

## Tested
All matching tests still run in about ~15sec in a single process.
- [x] Narrowed tolerance of existing gradient tests
- [x] Added new gradient tests comparing BP gradients to exact gradients (this test can catch weird errors where BP gets lucky in computing a correct value but has arbitrarily bad gradients)
- [x] Tested new gradient computation in a bigger model, which resulted in: better loss, more plausible solutions, more plausible gradient trajectories (e.g. with many more zero crossings)

### DIFF
--- a/pyro/distributions/one_one_matching.py
+++ b/pyro/distributions/one_one_matching.py
@@ -108,20 +108,26 @@ class OneOneMatching(TorchDistribution):
         # We find that Sinkhorn iteration is more robust and faster than the
         # optimal belief propagation updates suggested in [1-4].
         finfo = torch.finfo(self.logits.dtype)
-        shift = self.logits.data.max(1, True).values
-        shift.clamp_(min=finfo.min, max=finfo.max)
-        p = (self.logits - shift).exp().clamp(min=finfo.tiny ** 0.5)
-        d = 1 / p.sum(0)
+        # Note gradients are more accurate when shift is not detached.
+        shift = self.logits.max(1, True).values
+        shift.data.clamp_(min=finfo.min, max=finfo.max)
+        logits = self.logits - shift
+        d = logits.logsumexp(0)
         for _ in range(self.bp_iters):
-            s = 1 / (p @ d)
-            d = 1 / (s @ p)
-        b = s[:, None] * d * p
+            s = (logits - d).logsumexp(-1, True)
+            d = (logits - s).logsumexp(0)
+        b = (logits - (d + s)).exp()
+
+        def log(x):
+            return x.clamp(min=finfo.tiny).log()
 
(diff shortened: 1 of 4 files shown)


pyro/distributions/one_one_matching.py
```diff
@@ -108,20 +108,26 @@ def log_partition_function(self):
         # We find that Sinkhorn iteration is more robust and faster than the
         # optimal belief propagation updates suggested in [1-4].
         finfo = torch.finfo(self.logits.dtype)
-        shift = self.logits.data.max(1, True).values
-        shift.clamp_(min=finfo.min, max=finfo.max)
-        p = (self.logits - shift).exp().clamp(min=finfo.tiny ** 0.5)
-        d = 1 / p.sum(0)
+        # Note gradients are more accurate when shift is not detached.
```
martinjankowiak, line 111: whoops i should have pointed this out

tests/distributions/test_one_one_matching.py
```diff
@@ -67,32 +67,44 @@ def test_log_prob_hard(dtype, bp_iters):
...
 
     def fn(logits):
         d = dist.OneOneMatching(logits, bp_iters=bp_iters)
         return d.log_partition_function
 
-    torch.autograd.gradcheck(fn, logits, atol=0.01, rtol=0.01)
+    torch.autograd.gradcheck(fn, logits, atol=1e-3, rtol=1e-3)
+
+
+def assert_grads_agree(logits):
```
martinjankowiak, line 79: :+1:

martinjankowiak (approved):

pyro/distributions/one_one_matching.py
fritzo: Oh is this well-known? Mathematically they're identical, and detaching is cheaper. In fact I got a lot of speedup in #2640 by detaching all over the place. I only luckily happened upon not-detaching here and I don't understand why it improves gradient accuracy 🤔

martinjankowiak: oh i guess you're right then i'm confused. should i merge anyway?

fritzo: Yes, ready to merge.

Empirically this has accurate gradients, but I would like to follow up trying to figure out why the gradients are better than the .detach'ed version. Thanks for reviewing!

PR #2712, opened 2020-12-08
Add rate parameter to CoalescentTimes distribution

fritzo:
This PR does two things.

1. Adds a `rate` parameter to the simple `CoalescentTimes` distribution so that models can learn a (time-homogeneous) effective population size. I'm using this to match behavior of Beast2 which can learn the coalescent rate while fixing substitution rate (i.e. mutation rate).

2. Adds support for gradients through `CoalescentTimes.log_prob()`. Previously this was difficult because `._weak_memoize()` relied on caching by id, which would have broken for mutable tensors. I've updated to check the `Tensor._version` counter to avoid cache hits if tensors have been mutated. @eb8680 I wonder if we could use this trick anywhere in Funsor.

## Tested
- [x] added test for the new rate parameter
- [x] tested in a phylogenetic inference problem

### DIFF
--- a/pyro/distributions/coalescent.py
+++ b/pyro/distributions/coalescent.py
@@ -34,13 +34,15 @@ class CoalescentTimesConstraint(constraints.Constraint):
 
 class CoalescentTimes(TorchDistribution):
     """
-    Distribution over coalescent times given irregular sampled ``leaf_times``.
+    Distribution over sorted coalescent times given irregular sampled
+    ``leaf_times`` and constant population size.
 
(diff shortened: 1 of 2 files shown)


eb8680 (approved):
LGTM

> I wonder if we could use this trick anywhere in Funsor

We could update `TensorMeta` to do something similar, but since the JAX backend doesn't allow mutation of underlying tensors, it would just be an extra layer of stability for cons-hashing in the PyTorch backend - we couldn't add mutation to the Funsor API in a backend-agnostic way.

PR #2716, opened 2020-12-17
TraceMarkovEnum_ELBO - a trace implementation of ELBO-based SVI that supports model side enumeration of Markov variables

ordabayevy:
WIP. This first commit has a draft implementation of the `TraceMarkovEnum_ELBO` with `modified_partial_sum_product`. Tests compare elbos for `pyro.markov` models computed using `TraceEnum_ELBO` and `pyro.vectorized_markov` models computed using `TraceMarkovEnum_ELBO`. Note: all models have an empty guide.

**Note**

`TraceMarkovEnum_ELBO` only supports model side Markov enumeration. All discrete variable with the same&higher ordinal must also be enumerated in the model.

### DIFF
--- a/pyro/contrib/funsor/infer/traceenum_elbo.py
+++ b/pyro/contrib/funsor/infer/traceenum_elbo.py
@@ -18,8 +18,10 @@ def terms_from_trace(tr):
     # data structure containing densities, measures, scales, and identification
     # of free variables as either product (plate) variables or sum (measure) variables
     terms = {"log_factors": [], "log_measures": [], "scale": to_funsor(1.),
-             "plate_vars": frozenset(), "measure_vars": frozenset()}
+             "plate_vars": frozenset(), "measure_vars": frozenset(), "plate_to_step": dict()}
     for name, node in tr.nodes.items():
+        if node["type"] == "markov_chain":
+            terms["plate_to_step"][node["name"]] = node["value"]
         if node["type"] != "sample" or type(node["fn"]).__name__ == "_Subsample":
             continue
         # grab plate dimensions from the cond_indep_stack
@@ -39,7 +41,67 @@ def terms_from_trace(tr):
         # grab the log-density, found at all sites except those that are not replayed
         if node["is_observed"] or not node.get("replay_skipped", False):
             terms["log_factors"].append(node["funsor"]["log_prob"])
+    for p in terms["plate_vars"]:
+        terms["plate_to_step"][p] = terms["plate_to_step"].get(p, {})
     return terms
+
+
+@copy_docs_from(_OrigTraceEnum_ELBO)
+class TraceMarkovEnum_ELBO(ELBO):
+
+    def differentiable_loss(self, model, guide, *args, **kwargs):
+
+        # get batched, enumerated, to_funsor-ed traces from the guide and model
+        with plate(size=self.num_particles) if self.num_particles > 1 else contextlib.ExitStack(), \
+                enum(first_available_dim=(-self.max_plate_nesting-1) if self.max_plate_nesting else None):
+            guide_tr = trace(guide).get_trace(*args, **kwargs)
+            model_tr = trace(replay(model, trace=guide_tr)).get_trace(*args, **kwargs)
+
+        # extract from traces all metadata that we will need to compute the elbo
+        guide_terms = terms_from_trace(guide_tr)
+        model_terms = terms_from_trace(model_tr)
+
+        # build up a lazy expression for the elbo
+        with funsor.interpreter.interpretation(funsor.terms.lazy):
+            # identify and contract out auxiliary variables in the model with partial_sum_product
+            contracted_factors, uncontracted_factors = [], []
+            for f in model_terms["log_factors"]:
+                if model_terms["measure_vars"].intersection(f.inputs):
+                    contracted_factors.append(f)
+                else:
+                    uncontracted_factors.append(f)
+            # incorporate the effects of subsampling and handlers.scale through a common scale factor
+            contracted_costs = [model_terms["scale"] * f for f in funsor.sum_product.modified_partial_sum_product(
+                funsor.ops.logaddexp, funsor.ops.add,
+                model_terms["log_measures"] + contracted_factors,
+                plate_to_step=model_terms["plate_to_step"],
+                eliminate=model_terms["measure_vars"] | frozenset({
+                    p for p, s in model_terms["plate_to_step"].items() if s})
+            )]
+
+            costs = contracted_costs + uncontracted_factors  # model costs: logp
+            costs += [-f for f in guide_terms["log_factors"]]  # guide costs: -logq
+
+            # finally, integrate out guide variables in the elbo and all plates
+            # dict1 | dict2 in python 3.9
+            plate_to_step = {**guide_terms["plate_to_step"], **model_terms["plate_to_step"]}
+            plate_vars = frozenset(plate_to_step.keys())
+            elbo = to_funsor(0, output=funsor.Real)
+            for cost in costs:
+                # compute the marginal logq in the guide corresponding to this cost term
+                log_prob = funsor.sum_product.modified_sum_product(
+                    funsor.ops.logaddexp, funsor.ops.add,
+                    guide_terms["log_measures"],
+                    plate_to_step=plate_to_step,
+                    eliminate=(plate_vars | guide_terms["measure_vars"]) - frozenset(cost.inputs)
+                )
+                # compute the expected cost term E_q[logp] or E_q[-logq] using the marginal logq for q
+                elbo_term = funsor.Integrate(log_prob, cost, guide_terms["measure_vars"] & frozenset(cost.inputs))
+                elbo += elbo_term.reduce(funsor.ops.add, plate_vars & frozenset(cost.inputs))
+
+        # evaluate the elbo, using memoize to share tensor computation where possible
+        with funsor.memoize.memoize():
+            return -to_data(funsor.optimizer.apply_optimizer(elbo))
 
 
 @copy_docs_from(_OrigTraceEnum_ELBO)
--- a/pyro/contrib/funsor/infer/traceenum_elbo.py
+++ b/pyro/contrib/funsor/infer/traceenum_elbo.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import contextlib
+from functools import reduce
 
 import funsor
 
@@ -89,12 +90,13 @@ class TraceMarkovEnum_ELBO(ELBO):
             elbo = to_funsor(0, output=funsor.Real)
             for cost in costs:
                 # compute the marginal logq in the guide corresponding to this cost term
-                log_prob = funsor.sum_product.modified_sum_product(
+                factors = funsor.sum_product.modified_partial_sum_product(
                     funsor.ops.logaddexp, funsor.ops.add,
                     guide_terms["log_measures"],
                     plate_to_step=plate_to_step,
                     eliminate=(plate_vars | guide_terms["measure_vars"]) - frozenset(cost.inputs)
                 )
+                log_prob = reduce(funsor.ops.add, factors, funsor.Number(funsor.ops.UNITS[funsor.ops.add]))
                 # compute the expected cost term E_q[logp] or E_q[-logq] using the marginal logq for q
                 elbo_term = funsor.Integrate(log_prob, cost, guide_terms["measure_vars"] & frozenset(cost.inputs))
                 elbo += elbo_term.reduce(funsor.ops.add, plate_vars & frozenset(cost.inputs))
--- a/pyro/contrib/funsor/infer/traceenum_elbo.py
+++ b/pyro/contrib/funsor/infer/traceenum_elbo.py
@@ -20,8 +20,8 @@ def terms_from_trace(tr):
     terms = {"log_factors": [], "log_measures": [], "scale": to_funsor(1.),
              "plate_vars": frozenset(), "measure_vars": frozenset(), "plate_to_step": dict()}
     for name, node in tr.nodes.items():
+        # add markov dimensions to the plate_to_step dictionary
         if node["type"] == "markov_chain":
-            # add markov dimensions to the plate_to_step dictionary
             terms["plate_to_step"][node["name"]] = node["value"]
         if node["type"] != "sample" or type(node["fn"]).__name__ == "_Subsample":
             continue
@@ -31,8 +31,7 @@ def terms_from_trace(tr):
         if node["funsor"].get("log_measure", None) is not None:
             terms["log_measures"].append(node["funsor"]["log_measure"])
             # sum (measure) variables: the fresh non-plate variables at a site
-            # terms["measure_vars"] |= (frozenset(node["funsor"]["value"].inputs) | {name}) - terms["plate_vars"]
-            terms["measure_vars"] |= (frozenset(node["funsor"]["log_prob"].inputs) | {name}) - terms["plate_vars"]
+            terms["measure_vars"] |= (frozenset(node["funsor"]["value"].inputs) | {name}) - terms["plate_vars"]
         # grab the scale, assuming a common subsampling scale
         if node.get("replay_active", False) and set(node["funsor"]["log_prob"].inputs) & terms["measure_vars"] and \
                 float(to_data(node["funsor"]["scale"])) != 1.:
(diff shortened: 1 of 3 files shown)


ordabayevy:
All current tests have empty guides, so I'll add some tests where markov variables are enumerated in the guide. @eb8680 do you have any suggestions on what kind of other tests need to be added?

eb8680:
> do you have any suggestions on what kind of other tests need to be added?

You've hit the edge of Pyro's enumeration functionality :)  You could port the model-guide pairs in [`test_elbo_hmm_in_model`](https://github.com/pyro-ppl/pyro/blob/dev/tests/infer/test_enum.py#L1108), [`test_elbo_hmm_in_guide`](https://github.com/pyro-ppl/pyro/blob/dev/tests/infer/test_enum.py#L1168), and [`test_hmm_enumerate_model_and_guide`](https://github.com/pyro-ppl/pyro/blob/dev/tests/infer/test_enum.py#L1264) from `tests/infer/test_enum.py` and compare the ELBO values for `markov` and `vectorized_markov` as in your updated `test_vectorized_markov`.

pyro/contrib/funsor/infer/traceenum_elbo.py
```diff
@@ -39,9 +42,70 @@ def terms_from_trace(tr):
...
+                factors = funsor.sum_product.modified_partial_sum_product(
+                    funsor.ops.logaddexp, funsor.ops.add,
+                    guide_terms["log_measures"],
+                    plate_to_step=plate_to_step,
+                    eliminate=(plate_vars | guide_terms["measure_vars"]) - frozenset(cost.inputs)
+                )
+                log_prob = reduce(funsor.ops.add, factors, funsor.Number(funsor.ops.UNITS[funsor.ops.add]))
+                # compute the expected cost term E_q[logp] or E_q[-logq] using the marginal logq for q
+                elbo_term = funsor.Integrate(log_prob, cost, guide_terms["measure_vars"] & frozenset(cost.inputs))
+                elbo += elbo_term.reduce(funsor.ops.add, plate_vars & frozenset(cost.inputs))
```
eb8680, line 102: Note these two lines will not be correct for markov plates; you should try to write a guide enumeration test where this causes the sequential and vectorized markov ELBO values to disagree. I believe some of the test cases I suggested would be sufficient.

You might need to rewrite this as another `modified_partial_sum_product` computation. Note that 
```py
funsor.Integrate(log_prob, integrand, reduced_vars)
```
is equivalent to
```py
(log_prob.exp() * integrand).reduce(ops.add, reduced_vars)
```

tests/contrib/funsor/test_vectorized_markov.py
```diff
@@ -327,6 +332,14 @@ def test_vectorized_markov(model, data, var, history, use_replay):
             expected_step |= frozenset({v_step})
         assert actual_step == expected_step
 
+    with pyro_backend("contrib.funsor"):
+        if model != model_5:
```
eb8680, line 336: nit: instead of putting everything here inside this `if` statement, add a standalone `if` that triggers an `xfail`, e.g.
```py
with pyro_backend("contrib.funsor"):
    if history > 1:
        pytest.xfail(reason="TraceMarkovEnum_ELBO does not yet support history > 1")

    elbo = ...
```

pyro/contrib/funsor/infer/traceenum_elbo.py
```diff
@@ -39,9 +42,70 @@ def terms_from_trace(tr):
         # grab the log-density, found at all sites except those that are not replayed
         if node["is_observed"] or not node.get("replay_skipped", False):
             terms["log_factors"].append(node["funsor"]["log_prob"])
+    for p in terms["plate_vars"]:
+        terms["plate_to_step"][p] = terms["plate_to_step"].get(p, {})
     return terms
 
 
+@copy_docs_from(_OrigTraceEnum_ELBO)
+class TraceMarkovEnum_ELBO(ELBO):
```
eb8680, line 51: In a future PR, once we're confident that `TraceMarkovEnum_ELBO` is correct, we can just merge it with `TraceEnum_ELBO` instead of keeping the implementations separate.

eb8680:
We'll need to update the Funsor dependency to the latest `master` before this can be merged.

pyro/contrib/funsor/infer/traceenum_elbo.py
ordabayevy: > Note these two lines will not be correct for markov plates

Ok. I haven't checked the math yet. So there might be some more errors.

> is equivalent to

Thanks for explaining this.

eb8680:
@ordabayevy tests should pass after you merge the latest `dev` branch

pyro/contrib/funsor/infer/traceenum_elbo.py
ordabayevy: After taking a closer look it seems like `modified_partial_sum_product` might not work for guide enumerated models. The reason is that cost (integrand) `logq(z)` of transition factors has to be split into the sum factors `logq(z_0) + logq(z_1|z_0) + ...` before integrating w.r.t. `q(z_0)q(z_1|z_0) ...`. I can explain this in more details at the Zoom meeting.

ordabayevy: In other words the `prod` operation in the integrand is `ops.add` and in the `log_prob.exp()` is `ops.mul`.

tests/contrib/funsor/test_vectorized_markov.py
```diff
@@ -457,3 +457,107 @@ def test_vectorized_markov_multi(model, weeks_data, days_data, vars1, vars2, his
...
+
+@pytest.mark.parametrize("model,guide,data,", [
+    (model_10, guide_10, torch.ones(5)),
+])
+def test_guide_enumerated_elbo(model, guide, data):
+
+    with pyro_backend("contrib.funsor"):
+        with pytest.raises(
+                NotImplementedError,
+                match="TraceMarkovEnum_ELBO does not yet support guide side enumeration"):
```
ordabayevy, line 555: `xfail` test for guide enumerated models.

pyro/contrib/funsor/infer/traceenum_elbo.py
```diff
@@ -39,9 +42,72 @@ def terms_from_trace(tr):
...
+            guide_tr = trace(guide).get_trace(*args, **kwargs)
+            model_tr = trace(replay(model, trace=guide_tr)).get_trace(*args, **kwargs)
+
+        # extract from traces all metadata that we will need to compute the elbo
+        guide_terms = terms_from_trace(guide_tr)
+        model_terms = terms_from_trace(model_tr)
+
+        # guide side enumeration is not supported
+        if any(guide_terms["plate_to_step"].values()):
+            raise NotImplementedError("TraceMarkovEnum_ELBO does not yet support guide side enumeration")
```
ordabayevy, line 67: raise `NotImplementedError`

ordabayevy (commented):
I also have tried to add a model with mixed enumeration, similar to the model that I'm using for my project. `m` is enumerated in the guide and `theta` markov variable is enumerated in the model. But then realized that `modified_partial_sum_product` has to eliminate all downstream variables first before eliminating markov variables. So it seems like guide side enumeration is needed for my model after all.
```py
def model_9(data, history, vectorized):
    theta_dim, m_dim = 3, 2
    theta_init = pyro.param("theta_init", lambda: torch.rand(theta_dim), constraint=constraints.simplex)
    theta_trans = pyro.param("theta_trans", lambda: torch.rand((theta_dim, theta_dim)), constraint=constraints.simplex)
    m_prior = pyro.param("m_prior", lambda: torch.rand((theta_dim, m_dim)), constraint=constraints.simplex)
    locs = pyro.param("locs", lambda: torch.rand(m_dim))

    with pyro.plate("targets", size=data.shape[-2], dim=-2) as targets:
        targets = targets[:, None]
        theta_prev = None
        markov_loop = \
            pyro.vectorized_markov(name="frames", size=data.shape[-1], dim=-1, history=history) if vectorized \
            else pyro.markov(range(data.shape[-1]), history=history)
        for i in markov_loop:
            theta_curr = pyro.sample(
                "theta_{}".format(i), dist.Categorical(
                    theta_init if isinstance(i, int) and i < 1 else theta_trans[theta_prev]),
            m = pyro.sample("m_{}".format(i), dist.Categorical(m_prior[theta_curr]))
            pyro.sample("y_{}".format(i), dist.Normal(Vindex(locs)[..., m], 1.), obs=Vindex(data)[targets, i])
            theta_prev = theta_curr


def guide_9(data, history, vectorized):
    m_dim = 2
    m_probs = pyro.param("m_probs",
                         lambda: torch.rand((data.shape[-2], data.shape[-1], m_dim)),
                         constraint=constraints.simplex)

    with pyro.plate("targets", size=data.shape[-2], dim=-2) as targets:
        targets = targets[:, None]
        markov_loop = \
            pyro.vectorized_markov(name="frames", size=data.shape[-1], dim=-1, history=history) if vectorized \
            else pyro.markov(range(data.shape[-1]), history=history)
        for i in markov_loop:
            pyro.sample("m_{}".format(i), dist.Categorical(Vindex(m_probs)[targets, i]),
```

pyro/contrib/funsor/infer/traceenum_elbo.py
eb8680: nit: change error message to `"TraceMarkovEnum_ELBO does not yet support guide side Markov enumeration"`

tests/contrib/funsor/test_vectorized_markov.py
```diff
@@ -457,3 +457,107 @@ def test_vectorized_markov_multi(model, weeks_data, days_data, vars1, vars2, his
...
+        if history > 1:
+            pytest.xfail(reason="TraceMarkovEnum_ELBO does not yet support history > 1")
+
+        elbo = infer.TraceEnum_ELBO(max_plate_nesting=4)
+        expected_loss = elbo.differentiable_loss(model, guide, data, history, False)
+
+        vectorized_elbo = infer.TraceMarkovEnum_ELBO(max_plate_nesting=4)
+        actual_loss = vectorized_elbo.differentiable_loss(model, guide, data, history, True)
+
+        assert_close(actual_loss, expected_loss)
```
eb8680, line 490: These tests look good, but can you also check gradients?
```py
expected_loss = ...
params = pyro.get_param_store().values()
expected_grads = torch.autograd.grad(expected_loss, params)

vectorized_elbo = ...
actual_loss = ...
actual_grads = torch.autograd.grad(actual_loss, params)

assert_close(actual_loss, expected_loss)
for actual_grad, expected_grad in zip(actual_grads, expected_grads):
    assert_close(actual_grad, expected_grad)
```

tests/contrib/funsor/test_vectorized_markov.py
```diff
@@ -457,3 +457,107 @@ def test_vectorized_markov_multi(model, weeks_data, days_data, vars1, vars2, his
...
+
+    with pyro_backend("contrib.funsor"):
+
+        elbo = infer.TraceEnum_ELBO(max_plate_nesting=4)
+        expected_loss = elbo.differentiable_loss(model, guide, weeks_data, days_data, history, False)
+
+        vectorized_elbo = infer.TraceMarkovEnum_ELBO(max_plate_nesting=4)
+        actual_loss = vectorized_elbo.differentiable_loss(model, guide, weeks_data, days_data, history, True)
+
+        assert_close(actual_loss, expected_loss)
```
eb8680, line 511: ditto: check gradients

tests/contrib/funsor/test_vectorized_markov.py
```diff
@@ -457,3 +457,107 @@ def test_vectorized_markov_multi(model, weeks_data, days_data, vars1, vars2, his
...
+                NotImplementedError,
+                match="TraceMarkovEnum_ELBO does not yet support guide side enumeration"):
+
+            elbo = infer.TraceEnum_ELBO(max_plate_nesting=4)
+            expected_loss = elbo.differentiable_loss(model, guide, data, False)
+
+            vectorized_elbo = infer.TraceMarkovEnum_ELBO(max_plate_nesting=4)
+            actual_loss = vectorized_elbo.differentiable_loss(model, guide, data, True)
+
+            assert_close(actual_loss, expected_loss)
```
eb8680, line 563: ditto: check gradients (even though this test is currently xfailed)

eb8680:
> it seems like guide side enumeration is needed for my model after all

Ok, I'll try to get the parallel ELBO computations working after this PR is merged.

tests/contrib/funsor/test_vectorized_markov.py
ordabayevy: Using `torch.autograd.grad` gives `RuntimeError`:

```
>       return Variable._execution_engine.run_backward(
            outputs, grad_outputs_, retain_graph, create_graph,
            inputs, allow_unused)
E       RuntimeError: One of the differentiated Tensors appears to not have been used in the graph. Set allow_unused=True if this is the desired behavior.
```

Do I have to add `clear_param_store()` in the models?

ordabayevy:
Also `model_9, guide_9` pair above runs without any errors but computes wrong ELBO. It's harder to catch it and I'm not sure whether that should happen in `modified_partial_sum_product` or `TraceMarkovEnum_ELBO`. Or we can leave it up to the user not to write such models.

tests/contrib/funsor/test_vectorized_markov.py
eb8680: You could add `clear_param_store()` to the beginning of the test body.

eb8680:
> model_9, guide_9 pair above runs without any errors but computes wrong ELBO. It's harder to catch the error

What do you think the ideal behavior here should be?  We have a bit of [validation logic](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/traceenum_elbo.py#L44) in `pyro.infer.TraceEnum_ELBO` that forbids related behavior (no guide enumeration can be in a more local plate context than any downstream model enumeration), but it's not immediately obvious to me how or whether to extend that logic to this case.

ordabayevy:
Probably either by extending/altering that validation logic or controlling `cond_indep_stack`. I believe this issue arises in the first place because we label `m` as conditionally independent across `frames` dimension when that's not entirely true and just a matter of convenience. So I would prefer the latter approach where `m` wouldn't have `frames` in its `cond_indep_stack`.

eb8680 (approved):
Code and tests for model enumeration look good to me. We can defer all guide enumeration issues to future PRs.

ordabayevy:
Hi @eb8680 . I was also trying to get the guide side enumeration working. It seems to work for `model_9` and `model_10` [here](https://github.com/pyro-ppl/pyro/compare/dev...ordabayevy:guide-enumeration) with one problem. The strategy is to compute forward sequental sum terms `alpha(z_t) = sum(z_0,...,z_t-1) [q(z_0) ... q(z_t|q(z_t-1)]` and backward sequential sum terms `beta(z_t) = sum(z_t+1,...,z_T) [q(z_t+1|z_t) ... q(z_T|z_T-1)]` in parallel and then integrate cost terms in parallel as well `sum(z_t,z_t-1) [alpha(z_t-1)*cost(z_t|z_t-1)*beta(z_t)]`. However, I only have the [naive implementation](https://github.com/pyro-ppl/funsor/compare/master...ordabayevy:sequential-integral#diff-a6946e3ddadce95cf27b089e404a84e1a94496702fa79017b87deb5478017ed3R503) of forward and backward term calculations (`naive_prefix_sum` and `naive_suffix_sum`). Forward terms are the terms in Eq 5 in [Temporal Parallelization of Bayesian Smoothers](https://arxiv.org/abs/1905.13002) paper and require the down sweep calculations in the algorithm in Figure 1. That's where I got stuck because I couldn't figure out how to do it with funsor Tensors.

eb8680:
@ordabayevy I'm not sure I follow, but I believe the necessary parallel forward-backward algorithms are already implemented in `funsor.adjoint`.  The main challenge is applying them in a way that addresses #2724.

ordabayevy:
@eb8680 I looked up what adjoint operators are. Now it all makes sense (if I understood it correctly)!! I'll try to update `compute_expectation` to handle markov variables too.

ordabayevy:
@eb8680 is there a reference or explanation what `funsor.adjoint` does? I'm still having a hard time understanding what it  does and how to use it.

eb8680:
> is there a reference or explanation what funsor.adjoint does?

Well, the implementation is currently a bit cluttered, but it computes derivatives of abstract sum-product expressions (including any expression computed with `modified_partial_sum_product`) with respect to input factors using reverse-mode automatic differentiation.  The API isn't all that different from `torch.autograd.grad`, except that it currently requires specifying a sum and product operation in addition to a root expression and input factors; there are some usage examples in the [tests](https://github.com/pyro-ppl/funsor/blob/master/test/test_adjoint.py).

As described in [this tutorial paper](https://www.aclweb.org/anthology/W16-5901.pdf), it turns out that these derivatives correspond to marginal distributions when the sum and product operations are `logaddexp` and `add` and the input factors are unnormalized log-probability tensors, and that computing multiple marginal distributions in this way is highly efficient because it maximizes the amount of intermediate memory and computation shared between marginals.

PR #2720, opened 2020-12-21
Fix some typos in the contrib.funsor docs

ordabayevy:
- It is mostly sphinx cross-references.
- Fix the range of slices in `VectorizedMarkovMessenger` docstring example

Html files looked alright after compiling them with `make docs`.


eb8680 (approved):
Thanks for fixing this!

PR #2723, opened 2020-12-25
updating svi tutorial 1 with cells instead of inline

Padarn:


fritzo:
@Padarn I'm undecided whether this improves readability. On the one hand users can now execute code cells everywhere in the notebook, not just at the end. On the other hand that executability requires extra machinery , and that extra machinery slightly steals attention from the salient parts of pyro being explained.

@ab-10 can you explain a bit more about how you'd like to execute the code in the tutorial, and how the final executable code block isn't accomplishing that?

@eb8680 do you have any opinion?

Padarn:
Hi, @fritzo thanks for taking a look. I'm actually in agreement with you that this doesn't really improve readability.  Possibly it would make sense to remove the final 'all in one' block', if the cells are executable.

Issue #2724, opened 2020-12-26
[bug] contrib.funsor.infer.TraceEnum_ELBO can compute inaccurate elbo for certain guide-enumerated models 


ordabayevy:
### Issue Description
Calculating marginal `logq` in the guide (`log_prob` in `TraceEnum_ELBO`) for some cost terms can lead to inaccurate elbo values. E.g., in the test below for cost terms `log p(c|b)` and `log q(c|b)` computing marginal of `logq(a)q(b|a)q(c|b)` integrates out `a` which shouldn't be done because `q(b|a)` plated factors are only conditionally independent given `a`.

### Environment
 - Pyro `dev` branch and Funsor `master` branch

### Code Snippet
Command:
`pytest tests/contrib/funsor/test_enum_funsor.py::test_elbo_enumerate_plate_9`

AssertionError:
```
>           assert (max_err <= prec), msg
E           AssertionError: Expected:
E           2.2369411746502923
E           Actual:
E           1.0018455541365452
```
`test_elbo_enumerate_plate_9`:

<details>

```py
@pyroapi.pyro_backend(_PYRO_BACKEND)
def test_elbo_enumerate_plate_9():
    #        Model   Guide
    #          a
    #  +-------|------------+
    #  | M=2   V            |
    #  |       b -> c -> d  |
    #  +--------------------+
    pyro.param("model_probs_a",
               torch.tensor([0.45, 0.55]),
               constraint=constraints.simplex)
    pyro.param("model_probs_b",
               torch.tensor([[0.3, 0.7], [0.6, 0.4]]),
               constraint=constraints.simplex)
    pyro.param("model_probs_c",
               torch.tensor([[0.3, 0.4, 0.3], [0.4, 0.4, 0.2]]),
               constraint=constraints.simplex)
    pyro.param("model_probs_d",
               torch.tensor([[0.1, 0.9], [0.2, 0.8], [0.4, 0.6]]),
               constraint=constraints.simplex)
    pyro.param("guide_probs_a",
               torch.tensor([0.45, 0.55]),
               constraint=constraints.simplex)
    pyro.param("guide_probs_b",
               torch.tensor([[0.3, 0.7], [0.8, 0.2]]),
               constraint=constraints.simplex)
    pyro.param("guide_probs_c",
               torch.tensor([[0.4, 0.5, 0.1], [0.3, 0.5, 0.2]]),
               constraint=constraints.simplex)
    data = torch.tensor([1, 0])

[...]

PR #2725, opened 2020-12-28
Fix expectation computations in TraceEnum_ELBO by using adjoints to calculate marginals

ordabayevy:
Addresses https://github.com/pyro-ppl/pyro/issues/2724.
<s>Blocked by https://github.com/pyro-ppl/funsor/pull/413.</s>

This implementation avoids exponentiation of individual terms as suggested by @eb8680  [here](https://github.com/pyro-ppl/pyro/issues/2724#issuecomment-751552777).


ordabayevy:
@eb8680 can you review this PR?

Notes:
- This PR fixes #2724 
- Marginals are calculated using `adjoint` module
- To get the required list of marginals dummy targets are created (similar to what you did in this [PR](https://github.com/pyro-ppl/pyro/pull/2726/commits/972bdc44ef26bcfe08807ddda2a7c36e2acab3cb))
- In order to work this PR also requires updated adjoint patterns https://github.com/pyro-ppl/funsor/pull/444

pyro/contrib/funsor/infer/traceenum_elbo.py
```diff
@@ -148,19 +148,33 @@ def differentiable_loss(self, model, guide, *args, **kwargs):
...
+                    eliminate=(plate_vars | guide_terms["measure_vars"])
                 )
-                # compute the expected cost term E_q[logp] or E_q[-logq] using the marginal logq for q
-                elbo_term = funsor.Integrate(log_prob, cost, guide_terms["measure_vars"] & frozenset(cost.inputs))
+            marginals = tape.adjoint(funsor.ops.logaddexp, funsor.ops.add, logzq, tuple(targets.values()))
+            # finally, integrate out guide variables in the elbo and all plates
+            elbo = to_funsor(0, output=funsor.Real)
+            for cost in costs:
+                target = targets[frozenset(cost.inputs.keys())]
+                log_prob = marginals[target] - logzq
```
eb8680, line 176: Hmm... I think you can divide by the plate-local normalizing constants here rather than the global `logzq`
```py
logzq_local = marginals[target].reduce(funsor.ops.logaddexp, cost.input_vars - plate_vars)
log_prob = marginals[target] - logzq
```
which should lead to lower gradient variance.  I believe this is what the main implementation of `TraceEnum_ELBO` does. More generally, this would benefit from eager normalization within the backward pass of `adjoint`.

pyro/contrib/funsor/infer/traceenum_elbo.py
```diff
@@ -148,19 +148,33 @@ def differentiable_loss(self, model, guide, *args, **kwargs):
...
+            # compute the marginal logq in the guide corresponding to each cost term
+            targets = dict()
             for cost in costs:
-                # compute the marginal logq in the guide corresponding to this cost term
-                log_prob = funsor.sum_product.sum_product(
+                input_vars = frozenset(cost.inputs.keys())
+                if input_vars not in targets:
+                    targets[input_vars] = funsor.Tensor(
+                        funsor.ops.new_zeros(
+                            funsor.tensor.get_default_prototype(), ()
```
eb8680, line 159: Use `cost` as a prototype here

tests/contrib/funsor/test_enum_funsor.py
```diff
@@ -1390,3 +1390,73 @@ def guide_iplate():
             elbo = infer.TraceEnum_ELBO(max_plate_nesting=1)
             actual_loss = elbo.differentiable_loss(model_iplate_iplate, guide_iplate)
     _check_loss_and_grads(expected_loss, actual_loss)
+
+
+@pyroapi.pyro_backend(_PYRO_BACKEND)
+def test_elbo_enumerate_plate_9():
```
eb8680, line 1396: Can you add a couple more tests that work along similar lines (comparing ELBO values in models with parallel and sequential plates)? e.g. some of the following:
```
a -> ( (cij) )

( ai -> ( bij ) )

a -> ( bi -> ( cij ) )

a -> ( bi )
|
V
( cj )
```
You might be able to copy some of these from existing tests.

eb8680 (commented):
Thanks for updating. Can you explain which specific changes to `adjoint` are necessary for this to work?

ordabayevy:
> Can you explain which specific changes to `adjoint` are necessary for this to work?

The minimal required changes are updating `adjoint_binary` and `adjoint_reduce` patterns as in pyro-ppl/funsor#444 and removing the lines

```py
            if not isinstance(v, Variable):
                target_adjs[v] = bin_op(target_adjs[v], v)
```

pyro/contrib/funsor/infer/traceenum_elbo.py
ordabayevy: All three of these work (same loss and grad):

```py
log_prob = marginals[target] 
log_prob = marginals[target] - logzq
log_prob = marginals[target] - logzq_local
```

even when compared against the main `TraceEnum_ELBO`:

```py
from pyro.infer import TraceEnum_ELBO
...
    pyro_elbo = TraceEnum_ELBO(max_plate_nesting=2)
    pyro_loss = pyro_elbo.differentiable_loss(model_plate, guide_plate)
    _check_loss_and_grads(actual_loss, pyro_loss)
```

I'm still trying to understand why. Is it because in these examples guide is always normalized?

pyro/contrib/funsor/infer/traceenum_elbo.py
ordabayevy: `cost` can be a lazy expression in which case i'm not sure how to retrieve its prototype.

pyro/contrib/funsor/infer/traceenum_elbo.py
eb8680: Because guides are always normalized and all of the integration in these tests is exact (i.e. there is no sampling).

eb8680 (commented):
LGTM after you fix merge conflicts and test against https://github.com/pyro-ppl/funsor/pull/445 . I think we'll have to wait until https://github.com/pyro-ppl/funsor/pull/445 lands in Funsor to merge this PR, though.

ordabayevy:
Thanks for reviewing @eb8680 !

eb8680:
Now that https://github.com/pyro-ppl/funsor/pull/445 has landed can you update the `funsor[torch]` dependency to the latest `master`?  Should be good to merge if tests pass after that, although performance may be disappointing due to regressions in https://github.com/pyro-ppl/funsor/pull/445.

setup.py
```diff
@@ -120,8 +120,7 @@
         'horovod': ['horovod[pytorch]>=0.19'],
         'funsor': [
             # This must be a released version when Pyro is released.
-            # 'funsor[torch] @ git+git://github.com/pyro-ppl/funsor.git@c685dde1746a8234b44a3aa0d6ab4e3a4d619c41',
-            'funsor[torch]==0.4.0',
+            'funsor[torch] @ git+git://github.com/pyro-ppl/funsor.git@master',
```
eb8680, line 123: Ah, sorry, to clarify, can you pin to the commit hash of the most recent commit on `master` (`5d737584ad055cdff8d56f38e0c3dd10d0e7e80e`), as in the commented-out line above? Pinning to a commit hash rather than a tag/branch keeps Funsor changes from causing test failures in other, unrelated Pyro PRs.

eb8680 (changes requested):

ordabayevy:
There is an unrelated to this PR `HTTPError: HTTP Error 403: Forbidden` from `cvae/main.py`.

fritzo (dismissed):
I'd like to postpone merging this until after Pyro's 1.6 release, so as to avoid the need to release funsor

eb8680 (approved):

PR #2726, opened 2020-12-28
Make replay work correctly with vectorized_markov

eb8680:
This PR fixes a bug in the interaction between `VectorizedMarkovMessenger`, enumeration and `replay` and strengthens the tests in `tests/contrib/funsor/test_vectorized_markov.py` so that they would have caught it.

Specifically, by setting `msg["infer"]["_do_not_trace"] = True` for all but the final parallel-enumerated site, `vectorized_markov` was guaranteeing that those sites would not have their values recorded in or replayed from a guide trace.  I have replaced `_do_not_trace` here with a different flag, `_do_not_score`, which ensures that these values are replayed correctly but still excluded from ELBO term computation.  There should be no change in the number of tensor ops used in `TraceMarkovEnum_ELBO`.

This was not being caught in the `vectorized_markov` test cases with `use_replay=True` because the test case models had hard-coded `infer={"enumerate": "parallel"}`, meaning that model sites that should have been replayed were enumerated in the model. Removing these hard-coded `infer` dicts causes those sites to be sampled from the prior, leading to incorrect factors and ELBOs.

### DIFF
--- a/pyro/contrib/funsor/handlers/plate_messenger.py
+++ b/pyro/contrib/funsor/handlers/plate_messenger.py
@@ -315,7 +315,8 @@ class VectorizedMarkovMessenger(NamedMessenger):
             assert msg["name"].endswith(str(self._indices))
             msg["name"] = msg["name"][:-len(str(self._indices))] + str(self._suffix)
         if str(self._suffix) != str(self._suffixes[-1]):
-            # do not trace auxiliary vars
+            # _do_not_score: record these sites when tracing for use with replay,
+            # but do not include them in ELBO computation.
             msg["infer"]["_do_not_score"] = True
             msg["infer"]["is_auxiliary"] = True
             # map auxiliary var to markov var name prefix
--- a/pyro/contrib/funsor/handlers/plate_messenger.py
+++ b/pyro/contrib/funsor/handlers/plate_messenger.py
@@ -316,7 +316,7 @@ class VectorizedMarkovMessenger(NamedMessenger):
             msg["name"] = msg["name"][:-len(str(self._indices))] + str(self._suffix)
         if str(self._suffix) != str(self._suffixes[-1]):
             # do not trace auxiliary vars
-            msg["infer"]["_do_not_trace"] = True
+            msg["infer"]["_do_not_score"] = True
             msg["infer"]["is_auxiliary"] = True
             msg["is_observed"] = False
             # map auxiliary var to markov var name prefix
--- a/pyro/contrib/funsor/handlers/plate_messenger.py
+++ b/pyro/contrib/funsor/handlers/plate_messenger.py
@@ -333,7 +333,7 @@ class VectorizedMarkovMessenger(NamedMessenger):
             return
         # if last step in the for loop
         if str(self._suffix) == str(self._suffixes[-1]):
-            funsor_log_prob = msg["funsor"]["log_prob"] if "log_prob" in msg["funsor"] else \
+            funsor_log_prob = msg["funsor"]["log_prob"] if "log_prob" in msg.get("funsor", {}) else \
                 to_funsor(msg["fn"].log_prob(msg["value"]), output=funsor.Real)
             # for auxiliary sites in the log_prob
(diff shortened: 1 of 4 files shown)


eb8680:
@ordabayevy can you review this PR?

tests/contrib/funsor/test_vectorized_markov.py
```diff
@@ -531,48 +538,59 @@ def model_10(data, vectorized):
...
+@pytest.mark.parametrize("model,guide,data,history", [
+    (model_0, _guide_from_model(model_0), torch.rand(3, 5, 4), 1),
+    (model_1, _guide_from_model(model_1), torch.rand(5, 4), 1),
+    (model_2, _guide_from_model(model_2), torch.ones((5, 4), dtype=torch.long), 1),
+    (model_3, _guide_from_model(model_3), torch.ones((5, 4), dtype=torch.long), 1),
+    (model_4, _guide_from_model(model_4), torch.ones((5, 4), dtype=torch.long), 1),
+    (model_5, _guide_from_model(model_5), torch.ones((5, 4), dtype=torch.long), 2),
+    (model_6, _guide_from_model(model_6), torch.rand(5, 4), 1),
+    (model_7, _guide_from_model(model_7), torch.ones((5, 4), dtype=torch.long), 1),
+    (model_10, guide_10, torch.ones(5), 1),
```
ordabayevy, line 573: Can `_guide_from_model` be used to generate `guide_10`?

ordabayevy (commented):
Everything looks good to me (except some KeyError in the test). I noticed this too when replay replay didn't work correctly with guide enumerated models. After replaying the model would have factors like `x_0,x_curr,time` instead of `x_prev,x_curr,time`. Adding `enumerate:parallel` to the model was a workaround fix. But this looks like a proper fix.

pyro/contrib/funsor/handlers/plate_messenger.py
```diff
@@ -315,10 +315,10 @@ def _pyro_sample(self, msg):
             assert msg["name"].endswith(str(self._indices))
             msg["name"] = msg["name"][:-len(str(self._indices))] + str(self._suffix)
         if str(self._suffix) != str(self._suffixes[-1]):
-            # do not trace auxiliary vars
-            msg["infer"]["_do_not_trace"] = True
+            # _do_not_score: record these sites when tracing for use with replay,
+            # but do not include them in ELBO computation.
+            msg["infer"]["_do_not_score"] = True
             msg["infer"]["is_auxiliary"] = True
```
ordabayevy, line 321: `is_auxiliary` can be removed too I think. The only reason it is there is because of the assertion statement in
https://github.com/pyro-ppl/pyro/blob/b5e4a7b10a445b947bc5d670bcc9eb43480fec8c/pyro/poutine/trace_messenger.py#L124

pyro/contrib/funsor/infer/traceenum_elbo.py
```diff
@@ -23,7 +23,8 @@ def terms_from_trace(tr):
         # add markov dimensions to the plate_to_step dictionary
         if node["type"] == "markov_chain":
             terms["plate_to_step"][node["name"]] = node["value"]
-        if node["type"] != "sample" or type(node["fn"]).__name__ == "_Subsample":
+        if node["type"] != "sample" or type(node["fn"]).__name__ == "_Subsample" or \
+                node["infer"].get("_do_not_score", False):
             continue
```
ordabayevy, line 28: After testing these changes I've found out that auxiliary sites are not added to the `measure_vars`. Something like this should fix it:

```py
        if node["type"] != "sample" or type(node["fn"]).__name__ == "_Subsample":
            continue
        if node["infer"].get("_do_not_score", False):
            if node["funsor"].get("log_measure", None) is not None:
                terms["measure_vars"] |= (frozenset(node["funsor"]["value"].inputs) | {name}) - terms["plate_vars"]
            continue
```

eb8680: Thanks, should be fixed.

fritzo (approved):

PR #2727, opened 2020-12-31
Add SpanningTree properties .mode, .edge_mean

fritzo:
This adds:
- An `.edge_mean` property that computes edge marginals. This has a different return shape than the usual `.mean` properties because `SpanningTree` returns a packed tensor rather than a binary array. Hence I've named it slightly differently.
- A `.mode` property to compute a maximum-weight spanning tree. This has nearly identical implementation to the greedy `sample_tree_approx()` method (here using argmax rather than sampling), hence I've simply forked that code for implementation.

## Tested
- [x] unit tests for `.edge_mean`
- [x] unit tests for `.mode`
- [x] larger smoke tests for `.mode`

### DIFF
--- a/pyro/distributions/spanning_tree.py
+++ b/pyro/distributions/spanning_tree.py
@@ -170,6 +170,14 @@ class SpanningTree(TorchDistribution):
         trees = enumerate_spanning_trees(self.num_vertices)
         return torch.tensor(trees, dtype=torch.long)
 
+    def mode(self):
+        """
+        :returns: The maximum weight spanning tree.
+        :rtype: Tensor
+        """
+        backend = self.sampler_options.get("backend", "python")
+        return find_best_tree(self.edge_logits, backend=backend)
+
     def edge_mean(self):
         """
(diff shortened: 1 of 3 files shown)


pyro/distributions/spanning_tree.py
```diff
@@ -170,6 +170,39 @@ def enumerate_support(self, expand=True):
...
+        return find_best_tree(self.edge_logits, backend=backend)
+
+    @property
+    def edge_mean(self):
+        """
+        Computes marginal probabilities of each edge being active.
+
+        ..note:: This is similar to other distributions' ``.mean()``
+            method, but with a different shape because this distribution's
+            values are note encoded as binary matrices.
```
martinjankowiak, line 189: typo: note?

martinjankowiak (dismissed):

pyro/distributions/spanning_tree.py
fritzo: thanks, fixed.

martinjankowiak (approved):

PR #2655, opened 2020-10-07
Adding Probabilistic Topic Modeling tutorial

ucals:
This PR contributes with a Probabilistic Topic Modeling Tutorial. Thank you guys for all the help to put it together... it took way longer than what I thought it would take, but I think it is a nice tutorial. :)
Please let me know your thoughts/inputs/comments.. Cheers!

View [notebook](https://nbviewer.jupyter.org/github/ucals/pyro/blob/prodlda_tutorial/tutorial/source/prodlda.ipynb)

PR #2729, opened 2021-01-02
add ProdLDA tutorial

martinjankowiak:
this tutorial has been slightly adapted from @ucals 's excellent work in #2655.

most of the changes are textual improvements but i also added smoke_test support for CI integration.


eb8680:
Thanks for updating this!
- Could we replace `LogNormal` with `Dirichlet` and drop the corresponding discussion about reparametrization, or is it better to stick closely to the content of the paper?
- Did you want to make the other changes you suggested in https://github.com/pyro-ppl/pyro/pull/2655#issuecomment-706618876?
- Use consistent names for the two neural networks throughout, in both text and code (e.g. always call them `encoder`/`decoder`), and avoid using "recognition net" for the decoder?

martinjankowiak:
thanks @eb8680 
- re: dirichlet i think it's best to stick to the paper (but i added a comment)
- tried to use encoder/decoder everywhere please check for consistency
- added more figure captions
- didn't add `poutine.scale` logic since it it's apparently stable

eb8680 (dismissed):

eb8680:
Can you also double-check that everything renders correctly in Sphinx?

martinjankowiak:
not sure my sphinx build dies on `mle_map.ipynb` 🤷‍♀️

eb8680:
Looks like a bug introduced in #2701, try with the change in #2730?

eb8680:
Can you also add an attribution to @ucals at the top, below the title? ("This tutorial was originally written by @ucals" or something to that effect)

martinjankowiak:
ok thanks @eb8680 merged your fix rendering looks fine

eb8680 (approved):

PR #2734, opened 2021-01-08
Fix measure_vars for guide enumerated models

ordabayevy:
This is a quick fix that ensures that "prev" vars are not added to the `model_terms["measure_vars"]`  for guide enumerated models.

### DIFF
--- a/tests/contrib/funsor/test_vectorized_markov.py
+++ b/tests/contrib/funsor/test_vectorized_markov.py
@@ -17,6 +17,7 @@ try:
     from pyroapi import distributions as dist
     funsor.set_backend("torch")
     from pyroapi import handlers, pyro, infer
+    from pyro.contrib.funsor.infer.traceenum_elbo import terms_from_trace
 except ImportError:
     pytestmark = pytest.mark.skip(reason="funsor is not installed")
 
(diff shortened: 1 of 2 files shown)


eb8680 (commented):
Can you add or modify a test so that this bug would have been caught?

tests/contrib/funsor/test_vectorized_markov.py
```diff
@@ -325,12 +326,20 @@ def test_enumeration(model, data, var, history, use_replay):
...
                      + tuple("{}_{}".format(v, slice(j, data.shape[-2]-history+j)) for j in range(history+1))
             expected_step |= frozenset({v_step})
+            # grab measure_vars, found only at sites that are not replayed
+            if not use_replay:
+                expected_measure_vars |= frozenset(v_step)
         assert actual_step == expected_step
 
+        # check measure_vars
+        actual_measure_vars = terms_from_trace(vectorized_trace)["measure_vars"]
+        assert actual_measure_vars == expected_measure_vars
```
ordabayevy, line 341: This checks `measure_vars` both when the model is replayed and not replayed.

eb8680 (approved):
Thanks for fixing this!

PR #2735, opened 2021-01-09
AffineBeta distribution

ordabayevy:
An implementation of the `AffineBeta` distribution.

**Distribution parameters**
There are other possible ways to parameterize `AffineBeta`, e.g. `low,high` instead of `loc,scale` or `mean,size` instead of `concentration1,concentration0`. Which approach is more preferred?

One way to achieve alternative parameterization is by subclassing `AffineBeta`:

<details>

```py
class NewAffineBeta(AffineBeta):
    """
    :param mean: mean of the distribution.
    :param size: size parameter of the Beta distribution.
    :param low: min parameter.
    :param high: max parameter.
    """

    arg_constraints = {
        "mean": constraints.dependent,
        "size": constraints.real,
        "low": constraints.real,
        "high": constraints.dependent,
    }

    def __init__(self, mean, size, low, high, validate_args=None):
        try:
            concentration1 = size * (mean - low) / (high - low)
            concentration0 = size * (high - mean) / (high - low)
        except ZeroDivisionError:
            # this is needed to work with funsor make_dist
            low = torch.tensor(0.0)
            high = torch.tensor(1.0)
            concentration1 = torch.tensor(1.0)
            concentration0 = torch.tensor(1.0)
        super(AffineBeta, self).__init__(
            concentration1,
            concentration0,
            low,
            high - low,
            validate_args=validate_args,
        )

        new = self._get_checked_instance(NewAffineBeta, _instance)
        return super(NewAffineBeta, self).expand(batch_shape, _instance=new)
```

</details>

**Clamping sample values**

To avoid  `NaN` and `Inf` values in the gradient sample values are clamped by  `torch.finfo(x.dtype).eps * self.scale`.

### DIFF
--- a/pyro/distributions/affine_beta.py
+++ b/pyro/distributions/affine_beta.py
@@ -69,6 +69,10 @@ class AffineBeta(TransformedDistribution):
         x = torch.max(torch.min(x, self.loc + eps), self.loc + self.scale - eps)
         return x
 
+    @constraints.dependent_property
+    def support(self):
+        return constraints.interval(self.low, self.high)
+
     @property
     def concentration1(self):
         return self.base_dist.concentration1
--- a/pyro/distributions/affine_beta.py
+++ b/pyro/distributions/affine_beta.py
@@ -52,7 +52,7 @@ class AffineBeta(TransformedDistribution):
                 x = transform(x)
             # eps = torch.finfo(x.dtype).eps
             eps = 1e-5
-            x = torch.max(torch.min(x, self.low + eps), self.high - eps)
+            x = torch.min(torch.max(x, self.low + eps), self.high - eps)
             return x
 
     def rsample(self, sample_shape=torch.Size()):
@@ -66,7 +66,7 @@ class AffineBeta(TransformedDistribution):
             x = transform(x)
         # eps = torch.finfo(x.dtype).eps
         eps = 1e-5
-        x = torch.max(torch.min(x, self.low + eps), self.high - eps)
+        x = torch.min(torch.max(x, self.low + eps), self.high - eps)
         return x
 
     @constraints.dependent_property
--- a/pyro/distributions/affine_beta.py
+++ b/pyro/distributions/affine_beta.py
@@ -1,6 +1,6 @@
 import torch
-from pyro.distributions.torch import TransformedDistribution
-from torch.distributions import Beta, constraints
+from pyro.distributions.torch import Beta, TransformedDistribution
+from torch.distributions import constraints
 from torch.distributions.transforms import AffineTransform
 
 
@@ -52,7 +52,7 @@ class AffineBeta(TransformedDistribution):
                 x = transform(x)
             # eps = torch.finfo(x.dtype).eps
             eps = 1e-5
-            x = torch.max(torch.min(x, self.loc + eps), self.loc + self.scale - eps)
+            x = torch.max(torch.min(x, self.low + eps), self.high - eps)
             return x
 
     def rsample(self, sample_shape=torch.Size()):
@@ -66,7 +66,7 @@ class AffineBeta(TransformedDistribution):
             x = transform(x)
         # eps = torch.finfo(x.dtype).eps
         eps = 1e-5
-        x = torch.max(torch.min(x, self.loc + eps), self.loc + self.scale - eps)
+        x = torch.max(torch.min(x, self.low + eps), self.high - eps)
         return x
 
     @constraints.dependent_property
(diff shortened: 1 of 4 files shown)


pyro/distributions/affine_beta.py
```diff
@@ -0,0 +1,107 @@
...
+    def concentration1(self):
+        return self.base_dist.concentration1
+
+    @property
+    def concentration0(self):
+        return self.base_dist.concentration0
+
+    @property
+    def size(self):
+        return self.concentration1 + self.concentration0
```
fritzo, line 83: This name of property seems a confusing, since in PyTorch `.size()` is often used as a discrete quantity. Would you mind omitting this one property, maybe defining it in your user code?

pyro/distributions/affine_beta.py
```diff
@@ -0,0 +1,107 @@
...
+
+    arg_constraints = {
+        "concentration1": constraints.positive,
+        "concentration0": constraints.positive,
+        "loc": constraints.real,
+        "scale": constraints.positive,
+    }
+
+    def __init__(self, concentration1, concentration0, loc, scale, validate_args=None):
+        base_dist = Beta(concentration1, concentration0)
```
fritzo, line 31:
```py
Beta(concentration1, concentration0, validate_args=validate_args)
```

pyro/distributions/affine_beta.py
```diff
@@ -0,0 +1,107 @@
...
+        Additionally clamps the output in order to avoid `NaN` and `Inf` values
+        in the gradients.
+        """
+        with torch.no_grad():
+            x = self.base_dist.sample(sample_shape)
+            for transform in self.transforms:
+                x = transform(x)
+            eps = torch.finfo(x.dtype).eps * self.scale
+            x = torch.min(torch.max(x, self.low + eps), self.high - eps)
+            return x
```
fritzo, line 54: I believe you can simply omit `.sample()`, since [Distribution.sample()](https://github.com/pytorch/pytorch/blob/0f412aa29392d75e33f7e37d8276be8b40b4ae73/torch/distributions/distribution.py#L123-L129) defaults to `torch.no_grad()(rsample)`.

fritzo (commented):
Looks great!

pyro/distributions/affine_beta.py
ordabayevy: If I rename it to `.sample_size` like here https://en.wikipedia.org/wiki/Beta_distribution#Mean_and_sample_size would it still be confusing? If that doesn't work either I'll remove it from here and just keep it in my code as you suggested.

tests/distributions/conftest.py
```diff
@@ -131,6 +131,20 @@ def __init__(self, loc, scale):
...
+            scipy_dist=sp.beta,
+            examples=[
+                {'concentration1': [2.4], 'concentration0': [3.6], 'loc': [-1.0], 'scale': [2.0],
+                 'test_data': [-0.4]},
+                {'concentration1': [[2.4, 2.4], [3.6, 3.6]], 'concentration0': [[2.5, 2.5], [2.5, 2.5]],
+                 'loc': [[-1.0, -1.0], [2.0, 2.0]], 'scale': [[2.0, 2.0], [1.0, 1.0]],
+                 'test_data': [[[-0.4, 0.4], [2.5, 2.6]]]},
+                {'concentration1': [[2.4], [3.7]], 'concentration0': [[3.6], [2.5]],
+                 'loc': [[-1.0], [2.0]], 'scale': [[2.0], [2.0]],
+                 'test_data': [[0.0], [3.0]]}
```
ordabayevy, line 144: Are there any criteria for setting `examples` values here? I just took values from `dist.Beta` and slightly changed them.

pyro/distributions/affine_beta.py
fritzo: Sure, `.sample_size` seems ok.

tests/distributions/conftest.py
fritzo: There are no specific criteria. It's fine if you simply copy `Beta`. We just want to test some nontrivial values and some batched values.

pyro/distributions/affine_beta.py
ordabayevy: `.sample()` is redefined in [`TransformedDistribution`](https://github.com/pytorch/pytorch/blob/0f412aa29392d75e33f7e37d8276be8b40b4ae73/torch/distributions/transformed_distribution.py#L78-L89). Maybe I should add it back.

fritzo: ok, this is fine then.

fritzo (approved):

ordabayevy:
Thanks for reviewing it!

Issue #2658, opened 2020-10-08
FR Expose testing tools in released library


fritzo:
This proposes to add a new `pyro.testing` module with a stable testing interface.

A common pattern in applied probabilistic modeling is to implement domain-specific distributions, constraints, transforms, and losses for individual data science projects. That is, it is common for users to extend Pyro/NumPyro with new components.

I believe usability would be improved if Pyro/NumPyro provided easy-to-use tools to help users test their custom components. These tools should allow one-line invocation so that it is easy for users to test custom components even in Jupyter notebooks. To enable one-line invocation, we may need to add (or fix bugs in) metadata provided with each component, e.g. `Distribution` objects already have lots of metadata, but `Constraints` and `Transforms` would need lots of fixes to make `event_dim` work correctly.

Additionally these testing helpers would ease the code review burden of new distribution submissions. In particular Pyro/PyTorch currently do not check for agreement of `.sample()` vs `.log_prob()`, so reviewers must verify math.

## Proposed test helpers

Adapt `test_distribution(MyDistribution, **options)` from [test_distributions.py](https://github.com/pyro-ppl/pyro/blob/dev/tests/distributions/test_distributions.py)
- [ ] smoke test shapes and batching
- [ ] test `.expand()`
- [ ] death test constraint checking
- [ ] test `.sample()` matches `.log_prob()`, via moment checking or KDE
- [ ] test `.rsample()` gradients match `.log_prob.()` gradients, via moment matching
- [ ] test `.mean`, `.variance` if available

Adapt `test_transform(MyTransform, **options)` from @stefanwebb's [test_transforms.py](https://github.com/pyro-ppl/pyro/blob/dev/tests/distributions/test_transforms.py).
- [ ] test approximate inverse
- [ ] test `.lob_abs_det_jacobian()`
- [ ] test caching

Add a `test_constraint(my_constraint, **options)`
- [ ] test `transform_to()` and `biject_to()`
- [ ] test `.check()`

Adapt a `test_loss(my_loss, **options)` from [test_valid_models.py](https://github.com/pyro-ppl/pyro/blob/dev/tests/infer/test_valid_models.py) and maybe [test_gradients.py](https://github.com/pyro-ppl/pyro/blob/dev/tests/infer/test_gradients.py).
- [ ] smoke test handling of various Pyro features

## Design requirements
- provide instructive failure messages in case of error, e.g. point to tutorials like tensor_shapes.ipynb or #2657
[...]

PR #2738, opened 2021-01-11
Add goodness of fit test helpers for testing distributions

fritzo:
Addresses #2658 

This adds a PyTorch port of the [goftests](https://github.com/posterior/goftests) library, enabling easy statistical testing between distributions' `.sample()` and `.log_prob()` methods. I've had good luck with these statistical tests in a number of past projects, including Loom and Gamalon's PPL.

The main function is `auto_goodness_of_fit(samples, probs)` which returns a p-value that should be Uniform(0,1)-distributed for good data but should be close to zero for bad data. This well-calibrated statistic allows precise balancing and tuning of failure rate across a suite of tests:
```py
TEST_FAILURE_RATE = 1 / 100  # For 1 in 100 chance of spurious failure.

def test_my_distribution():
    d = MyDistribution()
    samples = d.sample([10000])
    probs = d.log_prob(samples).exp()
    gof = auto_goodness_of_fit(samples, probs)
    assert gof > TEST_FAILURE_RATE
```

## Tested

The following tests take less than 4 seconds total:

- [x] added a generic test helper to `test_distributions.py`
- [x] added two tests to `test_von_mises.py` exercising strict submanifold behavior
- [x] ported the unit test for `chi2sf()`

### DIFF
--- a/pyro/distributions/testing/gof.py
+++ b/pyro/distributions/testing/gof.py
@@ -2,6 +2,8 @@
 # Copyright (c) 2015, Gamelan Labs, Inc.
 # Copyright (c) 2016, Google, Inc.
 # Copyright (c) 2019, Gamalon, Inc.
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0 AND MIT
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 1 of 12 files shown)


pyro/distributions/testing/gof.py
```diff
@@ -0,0 +1,280 @@
+# Copyright (c) 2014, Salesforce.com, Inc.  All rights reserved.
```
fritzo, line 1: This file is ported from https://github.com/posterior/goftests/blob/master/goftests/__init__.py
Differences include:
- this restricts to continuous distributions
- this ports to PyTorch and Python 3

pyro/distributions/testing/special.py
```diff
@@ -0,0 +1,107 @@
+# Copyright (c) 2014, Salesforce.com, Inc.  All rights reserved.
```
fritzo, line 1: This file is ported from https://github.com/posterior/goftests/blob/master/goftests/utils.py
Differences include:
- port to Python 3

pyro/distributions/testing/gof.py
```diff
@@ -0,0 +1,290 @@
...
+    counts,
+    *,
+    total_count=None,
+    plot=False,
+):
+    """
+    Pearson's chi^2 test, on possibly truncated data.
+    https://en.wikipedia.org/wiki/Pearson%27s_chi-squared_test
+
+    :param Tensor probs: Vector of probabilities.
```
neerajprad, line 92: nit: I think you may need to import `Tensor` or use `torch.Tensor` here?

neerajprad (approved):
Thanks for adding these, it is is a great addition to distribution testing utils! I have not checked the math here which should be fine given the lineage of the original codebase, but the tests and documentation look great.

fritzo:
Thanks for reviewing, @neerajprad !

neerajprad (approved):

PR #2739, opened 2021-01-14
Add a Distribution.infer_shapes() method to statically infer shapes

fritzo:
Addresses https://github.com/pyro-ppl/funsor/issues/412
Requires https://github.com/pytorch/pytorch/pull/50547 and https://github.com/pytorch/pytorch/pull/50581

This adds a `TorchDistribution.infer_shapes()` method to statically compute `(batch_shape, event_shape)` for use in Funsor. This method inputs `*args,**kwargs` with the same signature as `.__init__()` but inputting shapes rather than tensors. The following should be equivalent:
```py
# Eager version.
d = MultivariateNormal(loc, covariance_matrix)
batch_shape = d.batch_shape
event_shape = d.event_shape

# Lazy version.
batch_shape, event_shape = MultivariateNormal.infer_shapes(loc.shape, covariance_matrix.shape)
```

The default implementation uses new `.support.event_dim` logic from https://github.com/pytorch/pytorch/pull/50547, so I've included backports in this PR (we can delete next time PyTorch is released).

## Tested
- [x] unit tests for `.support.is_discrete`
- [x] unit tests for `.support.event_dim`
- [x] unit tests for `.infer_shapes()`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,11 @@ pyro/_version.py
 processed
 raw
 *.pkl
+baseline_net_q1.pth
+cvae_net_q1.pth
+cvae_plot_q1.png
+prep_seal_data.csv
+results.csv
 
 # Logs
 logs
--- a/pyro/distributions/affine_beta.py
+++ b/pyro/distributions/affine_beta.py
@@ -6,6 +6,7 @@ from torch.distributions import constraints
 from torch.distributions.transforms import AffineTransform
 
 from .torch import Beta, TransformedDistribution
+from .util import broadcast_shape
 
 
 class AffineBeta(TransformedDistribution):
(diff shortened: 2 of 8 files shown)


fehiepsi:
@fritzo Do we need to do the same for transforms?

fritzo:
> Do we need to do the same for transforms?

Indeed to implement `.infer_shapes()` for `TransformedDistribution` we will need some sort of static shape computation. @stefanwebb has been thinking about this in his [Bijectors design doc](https://docs.google.com/document/d/1dNI-PaJzTMrEI46bGNyYCpVXvqgkLx-CefszZX0WMHw/preview). I think we will be able to use that doc's suggested `.forward_shape()` method, but I'd like to wait to implement that until a later PR.

fritzo:
@fehiepsi this is now unblocked 😄

fehiepsi (approved):
LGTM!

fritzo:

Issue #2737, opened 2021-01-11
Allow size == 1 in funsor IndepMessenger


ordabayevy:
@eb8680 can you allow `size == 1` here:

https://github.com/pyro-ppl/pyro/blob/e39d36c758cf1ca9a16b75909c793b06a8dabdd6/pyro/contrib/funsor/handlers/plate_messenger.py#L30

I ran into this issue when `PlateMessenger` has `subsample_size=1` . Pyro `IndepMessenger` only [checks](https://github.com/pyro-ppl/pyro/blob/e39d36c758cf1ca9a16b75909c793b06a8dabdd6/pyro/poutine/indep_messenger.py#L57) for `size!=0`:

eb8680:
Resolved in #2741

PR #2741, opened 2021-01-16
Allow size 1 in funsor IndepMessenger; (and fix AffineBeta docstring)

ordabayevy:
Addresses #2737 and ran `make isort`.

Also fixes `AffineBeta` docsrting.

### DIFF
--- a/pyro/contrib/funsor/handlers/plate_messenger.py
+++ b/pyro/contrib/funsor/handlers/plate_messenger.py
@@ -6,17 +6,15 @@ from numbers import Number
 
 import funsor
 
+from pyro.contrib.funsor.handlers.named_messenger import DimRequest, DimType, GlobalNamedMessenger, NamedMessenger
+from pyro.contrib.funsor.handlers.primitives import to_data, to_funsor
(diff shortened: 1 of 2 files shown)


pyro/contrib/funsor/handlers/plate_messenger.py
```diff
@@ -50,7 +47,7 @@ def __enter__(self):
         name_to_dim = OrderedDict([(self.name, DimRequest(self.dim, self.dim_type))])
         indices = to_data(self._indices, name_to_dim=name_to_dim)
         # extract the dimension allocated by to_data to match plate's current behavior
-        self.dim, self.indices = -indices.dim(), indices.squeeze()
+        self.dim, self.indices = -indices.dim(), indices.reshape(-1)
```
ordabayevy, line 50: Changed to `.reshape(-1)` so that `self.indices` would have a vector shape even for `self.size=1`.  This makes it compatible with the behavior of the pyro backend.

eb8680 (approved):
Thanks for the fix!

PR #2745, opened 2021-01-21
Autoname messenger

ordabayevy:
This is a work in progress with some basic tests that pass.


ordabayevy:
Fixed the behavior as a context-manager and copied more tests from `test_scoping.py`. @eb8680 or @fritzo can you review this?

tests/contrib/autoname/test_autoname.py
```diff
@@ -0,0 +1,373 @@
...
+        f1()
+        # for i in autoname(name="loop")(range(3)): <- this works too
+        for i in autoname(range(3), name="loop"):
+            f1()
+            sample("x", dist.Bernoulli(0.5))
+        return sample(dist.Normal(0.0, 1.0))
+
+    tr = poutine.trace(f2).get_trace()
+    assert "model/f1/Bernoulli" in tr.nodes
+    assert "model/loop/f1/Bernoulli" in tr.nodes
```
ordabayevy, line 132: For loops it might be visually better to add suffixes starting from counter zero: `model/loop__0/f1`, `model/loop__1/f1` or like this `model/loop/0/f1`, `model/loop/1/f1`

eb8680 (approved):

PR #2751, opened 2021-01-25
Implement poutine.get_mask() and use it to optimize NeuTra etc

### LINKED ISSUE #2744: Delay log density computation in pyro.factor
In several places (e.g. [NeuTra reparam](https://github.com/pyro-ppl/pyro/blob/7a69a3cf8b62f398607b6180dda978029cde698c/pyro/infer/reparam/neutra.py#L88)) that we used `pyro.factor` or `dist.Delta`, log_density is computed explicitly. This computation is not needed, e.g. when making predictions. It would be nice to add a mechanism to defer this computation, only until some inference algorithms that need the log probability.

### Proposed solution

Allow [log_factor](http://docs.pyro.ai/en/stable/primitives.html#pyro.primitives.factor) to be a callable in `pyro.factor`. In `dist.Unit.log_prob`, if `log_factor` is a callable, we will call it to get the required value (similar to the behavior in `pyro.param`).

fritzo:
Resolves #2744

This adds a `poutine.get_mask()` function to gate expensive computations. This is implemented via an internal primitive `poutine._inspect()` that creates a single message, applies all effects, and returns the message. While in theory effects may apply to more things, in practice `_inspect()` pulls out exactly the `mask` we need, and also `cond_indep_stack` and `scale` that we often access via hackery and may want in the future.

I also needed to change `poutine.mask(mask=False)` to avoid coercing its `mask` arg to a `torch.Tensor`.

## Tested
- [x] non-masked case is covered by existing tests
- [x] unit test for masked case

### DIFF
--- a/pyro/contrib/funsor/handlers/trace_messenger.py
+++ b/pyro/contrib/funsor/handlers/trace_messenger.py
@@ -2,12 +2,19 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import funsor
-
-from pyro.poutine.subsample_messenger import _Subsample
-from pyro.poutine.trace_messenger import TraceMessenger as OrigTraceMessenger
+import torch
 
 from pyro.contrib.funsor.handlers.primitives import to_funsor
 from pyro.contrib.funsor.handlers.runtime import _DIM_STACK
+from pyro.poutine.subsample_messenger import _Subsample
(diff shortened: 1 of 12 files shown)


fehiepsi (commented):
The proposal change looks great to me! Thanks, @fritzo!

fehiepsi (approved):
Looks great to me!

PR #2753, opened 2021-01-27
Update dev branch to use PyTorch 1.8 prerelease

fritzo:
Addresses #2754 
Blocking #2739, #2766

- @fehiepsi , @neerajprad could one of you review general changes?
- @stefanwebb could you take a look at the flows changes?

This pins our dev branch to track the latest PyTorch 1.8 nightly build, and fixes Pyro to be compatible with PyTorch 1.8.

This may be inconvenient for advanced users who use Pyro's dev branch, but it will help minimize breakage around the upcoming PyTorch 1.8 release by allowing us to release Pyro faster.

## Tested
- added `test_distributions.py::test_support_shape`
- added `test_distributions.py::test_is_discrete` and `is_not_discrete`
- xfailed some `collapse` tests; these will be fixed by #2739

PR #2759, opened 2021-02-03
Fix Delta.arg_constraints and .support

fritzo:
Ports https://github.com/pyro-ppl/numpyro/pull/902 to Pyro.

Note Pyro does not use `support.is_discrete`, and `dependent_constraint` will not support `is_discrete` until #2753 .

### DIFF
--- a/pyro/distributions/constraints.py
+++ b/pyro/distributions/constraints.py
@@ -87,6 +87,7 @@ class _OrderedVector(Constraint):
 
 
 corr_cholesky_constraint = _CorrCholesky()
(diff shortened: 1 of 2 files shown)


fehiepsi:
I guess this is blocked by  #2753 to make `constraints.independent` available.

fritzo:
@fehiepsi we already have an `IndependentConstraint` in Pyro dev, so I've used that to unblock this PR.

fehiepsi (approved):
LGTM.

Issue #2756, opened 2021-01-28
[feature request] Stan like constraints


mandel:
Stan supports multiples kind of constraints on parameters (https://mc-stan.org/docs/2_26/reference-manual/vector-and-matrix-data-types.html). To translate Stan models into Pyro, it would be convenient to have the same constraints implemented in Pyro.

Here is the list of constraints (the marked ones are already supported):
- [x] lower bound (`<lower=...>`): greater or equal
- [x] upper bound (`<upper=...>`): lower or equal
- [x] interval (`<lower=..., upper=...>`)
- [x] unit simplex (`simplex`): vector with non-negative values whose entries sum to 1
- [x] unit vector (`unit_vector`): vector with a norm of one
- [x] ordered vector (`ordered`): vector whose entries are sorted in ascending order
- [x] positive, ordered vector (`positive_ordered`): ordered vector where all entries are constrained to be positive
- [x] covariance matrix (`cov_matrix`): symmetric and positive definite matrix
- [x] correlation matrix (`corr_matrix`): symmetric and positive definite matrix, that has entries between −1 and 1, and has a unit diagonal
- [x] cholesky factors of covariance matrices (`cholesky_factor_cov`): https://mc-stan.org/docs/2_26/reference-manual/vector-and-matrix-data-types.html#cholesky-factors-of-covariance-matrices
- [x] cholesky factors of correlation matrices (`cholesky_factor_corr`): https://mc-stan.org/docs/2_26/reference-manual/vector-and-matrix-data-types.html#cholesky-factors-of-correlation-matrices

fritzo:
Pyro's dev branch already has a few of those [constraints](http://docs.pyro.ai/en/dev/distributions.html#module-pyro.distributions.constraints):
- `greater_than_eq` and `greater_than` (corresponding to Stan's `lower` I believe)
- `corr_cholesky_constraint` soon to be renamed to `corr_cholesky` as in PyTorch
- `ordered_vector` (corresponding to Stan's `ordered`)
- `sphere` (corresponding to Stan's `unit_vector`; I plan to port this to NumPyro soon)

fehiepsi:
Thanks, @mandel! Beside some mentioned above by @fritzo, for NumPyro, there is
+ `corr_matrix` (to be ported to Pyro soon)

Currently, in your list, Pyro and NumPyro is missing:
+ `positive_ordered` (IIUC the corresponding transform will be `ExpTransform(OrderedTransform())`)

[...]

PR #2760, opened 2021-02-04
Add positive_ordered_vector and corr_matrix constraints

fehiepsi:
Addresses #2756. This adds the remaining missing constraints in Pyro. With those new constraints, we have a LKJ distribution and InvCholeskyTransform.


pyro/distributions/torch_patch.py
```diff
@@ -80,6 +80,16 @@ def _HalfCauchy_logprob(self, value):
     return log_prob
 
 
+# TODO fix batch_shape have an extra singleton dimension upstream
```
fehiepsi, line 83: this [upstream line](https://github.com/pytorch/pytorch/blob/master/torch/distributions/constraints.py#L467) is not correct IMO.

fritzo: yeah weird, I don't understand the `.unsqueeze(0)`.

tests/distributions/test_transforms.py
```diff
@@ -327,3 +328,40 @@ def test_sylvester(self):
     def test_normalize_transform(self):
         for p in [1., 2.]:
             self._test(lambda p: T.Normalize(p=p), autodiff=False)
+
+
+@pytest.mark.parametrize('batch_shape', [(), (7,), (6, 5)])
+@pytest.mark.parametrize('dim', [2, 3, 5])
+@pytest.mark.parametrize('domain', [dist.constraints.corr_cholesky_constraint,
+                                    dist.constraints.lower_cholesky])
+def test_inv_cholesky_transform(batch_shape, dim, domain):
```
fritzo, line 337: Is there a reason you're not adding a standard test `TestTransforms.test_inv_cholesky_transform`? We've put a lot of work into those standardized tests, and it's a shame to not exercise them.

fritzo (commented):
:tada It's great to see more STAN-compatible transforms!

pyro/distributions/transforms/__init__.py
```diff
@@ -54,12 +56,31 @@ def _transform_to_corr_cholesky(constraint):
...
     return OrderedTransform()
 
 
+@biject_to.register(positive_ordered_vector)
+@transform_to.register(positive_ordered_vector)
+def _transform_to_positive_ordered_vector(constraint):
+    return ComposeTransform([OrderedTransform(), ExpTransform()])
+
+
+# TODO: register biject_to when LowerCholeskyTransform is bijective
```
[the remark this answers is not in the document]
fehiepsi: In NumPyro, it is a bijective from a vector to a matrix. Probably we will use another class with different name in Pyro for that purpose. So this todo will remind us to make the change.

pyro/distributions/transforms/cholesky.py
```diff
@@ -79,3 +79,47 @@ def log_abs_det_jacobian(self, x, y):
...
+    Transform via the mapping :math:`y = x @ x.T`, where `x` is a lower
+    triangular matrix with positive diagonal.
+    """
+    bijective = True
+    sign = +1
+    event_dim = 2
+
+    def __init__(self, domain=constraints.lower_cholesky, cache_size=0):
+        # TODO: change corr_cholesky_constraint to corr_cholesky when it is available
+        assert domain in [constraints.lower_cholesky, corr_cholesky_constraint]
```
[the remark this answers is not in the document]
fehiepsi: Good point! Back then we don't have InverseTransform in NumPyro. :D

tests/distributions/test_transforms.py
fehiepsi: I think we can add another kind of Flatten to test the jacobian of those special matrices. Let me figure it out how to merge this jacobian test to the current TestTransforms.

fritzo: Note there's a `Reshape` transform in the latest PyTorch, in case that helps. We could temporarily port that to Pyro, then remove it next month.

fehiepsi: I would argue that those standardized test, with assumptions that the domain/codomain is real or real_vector, makes it pretty tricky to test for those positive_definite, corr_matrix domains. I can use some composed transforms to make tests pass but would also like to have separate tests that are more intuitive.

fehiepsi: @fritzo I made several attempts but we need to use the standard tests for something like `ComposeTransform([LowerCholeskyTransform(), CholeskyTransform().inv, CholeskyTransform()])`. But it is not clear to me that composing `t.inv` and `t` will actually assert that both `t._call` and `t._inverse` give correct shapes. I would like to accept the shame and have different tests for those cholesky transforms.

fritzo: @fehiepsi Sounds good thanks for attempting! I think this is valuable feedback for @stefanwebb and @feynmanliang as they create tests for https://github.com/stefanwebb/flowtorch

fritzo (approved):

fritzo:
Thanks for refactoring!

PR #2762, opened 2021-02-06
Docs: Add reference for effect handlers

FlorianWilhelm:
For me, it helped a lot to read [An Introduction to Algebraic Effects and Handlers](https://www.eff-lang.org/handlers-tutorial.pdf) by Matija Pretnar to understand what effect handlers are as they are frequently mentioned in Pyro's docs but never really explained. This might also help others that don't know the concept of effect handlers.


fritzo (approved):
Thanks for adding this!

PR #2764, opened 2021-02-10
Fix bitwise & for bool and Tensor in MaskMessenger

ordabayevy:
After the updates I got the error below for `self.mask & msg["mask"]` when `type(self.mask) == bool` and `type(msg["mask"]) == Tensor`

```
TypeError: unsupported operand type(s) for &: 'bool' and 'Tensor'
```

which seems can be fixed by rearranging the order of `msg["mask"] & self.mask`

Tested in Pytorch 1.7.1 and 1.8.0


fritzo (approved):
Thanks for fixing this!

PR #2767, opened 2021-02-19
Stabilize autoguide scale parameters via SoftplusTransform

### LINKED ISSUE #2766: Softplus transform for AutoNormal scales [feature request]
Hi @fritzo @fehiepsi 

Following the conversation in numpyro issues (https://github.com/pyro-ppl/numpyro/issues/855#issuecomment-778250255) about the usefulness of softplus transform for AutoNormal scales, adding this feature request for further discussion here.

Requires https://github.com/pytorch/pytorch/pull/52300

fritzo:
Resolves #2766 
~~Blocked by #2753~~
Ports https://github.com/pytorch/pytorch/pull/52300

This switches to using softplus transforms for autoguide scale *parameters* (point 2 in https://github.com/pyro-ppl/numpyro/issues/855#issuecomment-757512947), and adds relevant machinery:
- `constraints.softplus_positive` and `SoftplusTransform`
- `constraints.softplus_lower_cholesky` and `SoftplusLowerCholeskyTransform`

This does not use softplus transforms for *latent variables* that are scales (point 1 in https://github.com/pyro-ppl/numpyro/issues/855#issuecomment-757512947). While this PR adds machinery to declare and perform those transforms, it remains to detect which positive latent variables should be softplus-transformed rather than exp-transformed.

cc @vitkl @fehiepsi 

## Tested
- needed to tweak parameters in a couple inference tests (test now actually run a bit faster)


pyro/distributions/transforms/stable.py
```diff
@@ -0,0 +1,61 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+from torch.distributions import constraints
+from torch.distributions.transforms import Transform
+from torch.nn.functional import softplus
+
+
+def softplus_inv(y):
+    return y.expm1().log()  # FIXME stabilize
```
fehiepsi, line 10: `(-y).expm1().neg().log() + y`

pyro/distributions/transforms/stable.py
```diff
@@ -0,0 +1,61 @@
...
+        return isinstance(other, SoftplusTransform)
+
+    def _call(self, x):
+        return softplus(x)
+
+    def _inverse(self, y):
+        return softplus_inv(y)
+
+    def log_abs_det_jacobian(self, x, y):
+        return -(-x).exp().log1p()  # FIXME stabilize
```
fehiepsi, line 33: `-softplus(-x)`

fehiepsi:
Just curious, what makes you think this will be more stable?

fritzo:
> Just curious, what makes you think this will be more stable?

I don't understand why this PR would be more stable :shrug: I'm just trying to implement @vitkl's point 2 https://github.com/pyro-ppl/numpyro/issues/855#issuecomment-757512947 which he claimed was more stable. It's plausible I've misunderstood something, and also plausible that a better solution would be to use `ClippedAdam` or a lower learning rate.

@vitkl can you confirm (1) this PR implements roughly what you're requesting, (2) that you've shown it is more numerically stable, and (3) that you tried simpler solutions like `ClippedAdam` or tweaking optimizer parameters?

fritzo:
@fehiepsi do you think we should name these constraints `softplus_positive` and `softplus_lower_cholesky`? That seemed like mixing metaphors to me, but I suppose it's better than introducing yet another word 'stable'. I appreciate your help in choosing good names.

fehiepsi:
@fritzo I don't have a better solution. Those names *softplus_foo* looks good to me. In TFP wrapper, because distributions there do not have constraints, I created a class named `BijectorConstraint(bijector)`, which is similar here (:D) `Bijector <-> softplus`, `Constraint <-> positive`.

vitkl:
Hi @fehiepsi @fritzo 

Thanks for this.

1. This implements exactly what I requested - however in the numpyro issue (https://github.com/pyro-ppl/numpyro/issues/855#issuecomment-778645906) you also discussed making that an option rather than the default: 

```python
class AutoNormal(..., use_softplus=False):

    if use_softplus:
        _deep_setattr(self.scales, name,
                                PyroParam(init_scale, constraints.stable_positive, event_dim))
    else:
        _deep_setattr(self.scales, name,
                                PyroParam(init_scale, constraints.positive, event_dim))
```

2. It is more numerically stable for the cell2location model (tested in numpyro). Softplus is also used by pymc3.

3. We see that ClippedAdam and reduced learning rate (0.002 -> 0.0002) does not help (again tested in numpyro). We did not do an exhaustive search of training hyperparameters yet. 

I am planning to add a benchmark (to https://github.com/pyro-ppl/sandbox)that shows the impact of this modification of stability and accuracy for our model. To simplify the analysis, it would help to have the switch option `AutoNormal(..., use_softplus=False)` in both pyro and numpyro.

fritzo:
Hi @vitkl, I believe this should be ready to go.

> making that an option rather than the default

Hmm, I'm hesitant to add interface complexity to our so-called AutoGuides, especially since I suspect your experiments will show `softplus` should be the default.  At the same time I'd like to make it easy for you to run experiments.  WDYT of this compromise: How about we make the constraint hackable but not publicly configurable?  That is, we create a class level variable
```py
class AutoNormal(AutoGuide):
    scale_constraint = constraints.softplus_positive
    def __init__(self, model, *,
                 init_loc_fn=init_to_feasible,
                 init_scale=0.1,
                 create_plates=None):
        ...
```
Then in your experiments you can override this via
```py
guide_exp = AutoNormal(model)
guide_exp.scale_constraint = constraints.positive

guide_softplus = AutoNormal(model)
guide_softplus.scale_constraint = constraints.softplus_positive
```
Hope I'm not overthinking this, I'd just like to avoid interface bloat 😄

fehiepsi:
Sure! This only changes how we optimize the scale parameters. Probably this will affect the current inference code of some users but I think it is easy to fix... Btw, we should mention why we make this change in the next release notes. :)

fehiepsi (approved):
LGTM with passing tests and exposing those new transforms in docs. Thanks for addressing this issue, @fritzo!

vitkl:
Thanks a lot for implementing this. Class variable works great. We are in the middle of resubmitting the paper revisions this week - but I will try to do the testing soon!

fehiepsi (approved):
LGTM, thanks @fritzo! I'll port this to NumPyro soon.

fehiepsi (approved):

PR #2772, opened 2021-02-24
Support obs_mask kwarg in sample statements

### LINKED ISSUE #1106: Support masked sample statement for batched imputation
Pyro can currently impute data in sample statements via
```py
def model(x=None):
    x = pyro.sample("x", my_dist, obs=x)
    ...
```
where `x` is either sampled when calling `model()` or observed when calling `model(observed_value)`.

It would be nice if this supported batching, where within a single batch some values are observed and some need to be sampled, say specified by `obs_mask`. Pyro can already handle masked observe statements via `TorchDistribution.mask()`, which allow tensorized decision between whether an observe statement exists, i.e. whether the `log(p)` and `log(q)` terms exist. What is needed is a different tensorized decision between whether a statement is a sample statement or an observe statement, which determines only whether the `log(q)` exists.

A possible implementation could be
```py
def sample(name, fn, obs=None, obs_mask=None):
    ...
    value = fn()
    if obs_mask is not None and obs_mask.any():
        value[obs_mask] = obs[obs_mask]        # use obs where available; sample otherwise
        guide_site["fn"] = fn.mask(~obs_mask)  # exclude log(q) terms at observed entries
    else:
        guide_site["fn"] = fn
    model_site["fn"] = fn
    ...
```

fritzo:
Resolves #1106 

This implements `pyro.sample(..., obs=___, obs_mask=___)` by transforming that one statement into a pair of `poutine.mask`ed `sample` statements plus a final `deterministic` statement. This solution is simple and non-invasive, with changes being limited to the `pyro.sample()` function.

One disadvantage of this approach is that the new latent variable site is named `name + "_unobserved"` rather than simply `name`. However fixing this would require a much more invasive change. In particular it is unclear how handlers should treat a single site that is partially observed and partially sampled. I suspect this 3-site-name approach is actually optimal.

I think to get this to work with `AutoGuide`s we may need to add masking logic in autoguides, otherwise I think some entirely-local guides may diverge to infinite entropy and eventually NAN out. However since that is really a shortcoming of handling of `poutine.mask` rather than `obs_mask`, I have not addressed autoguides in this PR.

## Tested
- [x] ELBO smoke tests in tests/infer/test_valid_models.py
- [x] MCMC smoke tests in tests/infer/mcmc/test_valid_models.py

### DIFF
--- a/pyro/primitives.py
+++ b/pyro/primitives.py
@@ -72,14 +72,30 @@ def sample(name, fn, *args, **kwargs):
     :param fn: distribution class or function
     :param obs: observed datum (optional; should only be used in context of
         inference) optionally specified in kwargs
+    :param ~torch.Tensor obs_mask: EXPERIMENTAL Optional boolean tensor mask.
+        If provided, values with mask=True will be conditioned on ``obs`` and
+        remaining values will be imputed by sampling. This introduces a latent
+        sample site named ``name + "_unobserved"`` which should be used by
+        guides (this site name may change in a future version).
+    :type obs_mask: bool or ~torch.Tensor
     :param dict infer: Optional dictionary of inference parameters specified
         in kwargs. See inference documentation for details.
     :returns: sample
     """
+    # Transform obs_mask into multiple sample statements.
     obs = kwargs.pop("obs", None)
-    infer = kwargs.pop("infer", {}).copy()
-    # check if stack is empty
+    obs_mask = kwargs.pop("obs_mask", None)
+    if obs_mask is not None:
+        with poutine.mask(mask=obs_mask):
+            observed = sample(f"{name}_observed", fn, *args, **kwargs, obs=obs)
+        with poutine.mask(mask=~obs_mask):
+            unobserved = sample(f"{name}_unobserved", fn, *args, **kwargs)
+        value = observed.where(obs_mask, unobserved)
+        return deterministic(name, value)
+
+    # Check if stack is empty.
     # if stack empty, default behavior (defined here)
+    infer = kwargs.pop("infer", {}).copy()
     if not am_i_wrapped():
         if obs is not None and not infer.get("_deterministic"):
             warnings.warn("trying to observe a value outside of inference at " + name,
--- a/pyro/primitives.py
+++ b/pyro/primitives.py
@@ -7,6 +7,8 @@ from collections import OrderedDict
 from contextlib import ExitStack, contextmanager
 from inspect import isclass
 
+import torch
+
 import pyro.distributions as dist
 import pyro.infer as infer
 import pyro.poutine as poutine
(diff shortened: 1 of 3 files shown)


fehiepsi:
It is nice to have a simple api for users like this. My main concern is this will draw unnecessary masked latent statement in HMC. It might not be important because those extra values does not contribute to the log density, so its trajectory is just random walk (modulo some geometric quantity like u-turn condition).

Regarding the api, how does this new keyword interact with event_dim? IIUC users will need to mask all invalid event dimensions I guess.

fritzo:
> how does this new keyword interact with event_dim?

This interface does not support masking of events. Actually I think event masking would be nonsensical: each event should be either entirely observed or drawn jointly in entirety. It seems weird to interleave pieces of a so-called event. I think of 'event' as being an 'atomic' sample that cannot be split.

> My main concern is this will draw unnecessary masked latent statement in HMC

Agreed, this seems similar to the issue I mentioned with `AutoGuide`. And similarly I believe this is a more basic issue with `poutine.mask`, and not specific to `obs_mask`. I think an optimal strategy could be to transform masked sample statements to contiguous unmasked statements like
```py
# masked -> contiguous
contiguous_sample = masked_sample[mask]

# contiguous -> masked
masked_sample = zeros(...)
masked_sample[mask] = contiguous_sample
```
However this optimization might be invalid if the mask depends on upstream sample statements. I think this deserves further discussion in a separate issue.

fehiepsi:
> This interface does not support masking of events

Should we raise an error in this case then? I assume that users might write codes like
```python
with plate("a", A):
    sample("x", dist.Normal(...).to_event(1), obs=..., obs_mask=ambiguous_shape?)
```

> an optimal strategy could be to transform masked sample statements to contiguous unmasked statements

Sound like a proposal for MaskTransform... ;)

fritzo:
@fehiepsi good point. I've fixed handling of multivariate distributions, added a shape check and informative error message, and added death tests.

> MaskTransform

Interesting, we could even add a `MaskReparam` 🤔

pyro/primitives.py
```diff
@@ -61,6 +63,30 @@ def param(name, *args, **kwargs):
     return _param(name, *args, **kwargs)
 
 
+def _masked_observe(name, fn, obs, obs_mask, *args, **kwargs):
+    # Split into two auxiliary sample sites.
+    with poutine.mask(mask=obs_mask):
+        observed = sample(f"{name}_observed", fn, *args, **kwargs, obs=obs)
+    with poutine.mask(mask=~obs_mask):
+        unobserved = sample(f"{name}_unobserved", fn, *args, **kwargs)
```
martinjankowiak, line 71: cool so this will let downstream sample statements depend on the sampled `unobserved` values right (since they will appear in `value` together with `observed`)?

fritzo: Correct. And it took less than three years to figure out how to do it 😄

eb8680:
What is the expected behavior of `obs_mask`s under `Predictive`?

fritzo:
> What is the expected behavior of obs_masks under Predictive?

The `observed` part should just be provided by the user, the `unobserved` part should be sampled from the posterior, and finally these two parts are interleaved and returned.

eb8680:
> The observed part should just be provided by the user, the unobserved part should be sampled from the posterior, and finally these two parts are interleaved and returned.

It seems like we'd want some mechanism for optionally disregarding the observed part, though, perhaps via `poutine.uncondition` if not automatically in `Predictive`?  Maybe the current version of `uncondition` is sufficient? At any rate, I guess we can leave higher-level interfaces to future issues/PRs.

eb8680 (approved):
LGTM

fritzo:
@eb8680 @fehiepsi thanks for reviewing!

@eb8680 I agree there is more to think about wrt uncondition and posterior predictive etc.

fehiepsi (approved):
LGTM too!

PR #2773, opened 2021-03-03
Examples from Named tensor notation article

ordabayevy:
Attempting to re-write examples from the article using funsors

https://namedtensor.github.io/


eb8680:
@ordabayevy this is great, but I think it probably belongs in the Funsor repository, not the Pyro one.

eb8680:
You might also be interested in using `funsor.factory.Has`, just added in https://github.com/pyro-ppl/funsor/pull/472

ordabayevy:
Moved to https://github.com/pyro-ppl/funsor/pull/477

PR #2774, opened 2021-03-04
Revert to old default scale constraint in autoguides

fritzo:
This reverts autoguide scale constraints to the old exp-transformed versions until @vitkl can perform experiments confirming improved performance. If experiments succeed, we can simply revert this PR.  Note this PR preserves the ability to easily perform those experiments, via
```py
old_autonormal = AutoNormal(model)
old_autonormal.scale_constraint = constraints.positive

new_autonormal = AutoNormal(model)
new_autonormal.scale_constraint = constraints.softplus_positive
```

See also https://github.com/pyro-ppl/numpyro/pull/941

PR #2775, opened 2021-03-04
Fix MNIST url redirection

fritzo:
Blocking #2774 

This updates our rerouting of torchvision.datasets.mnist.MNIST after that class was moved in a recent torchvision. This also refactors from a sketchy `torch_patch` to a safer pyro-local subclass `MNIST` of the upstream class `MNIST`.

This also updates to use torch==1.8.0 and torchvision==0.9.0 test releases

### DIFF
--- a/pyro/contrib/examples/util.py
+++ b/pyro/contrib/examples/util.py
@@ -4,17 +4,12 @@
 import os
 import sys
 
-import torchvision
 import torchvision.datasets as datasets
 from torch.utils.data import DataLoader
 from torchvision import transforms
-from torchvision.datasets import MNIST
-
-from pyro.distributions.torch_patch import patch_dependency
 
 
-@patch_dependency('torchvision.datasets.MNIST', torchvision)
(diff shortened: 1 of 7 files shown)


fehiepsi:
@fritzo Probably we need to [pump `torchvision`](https://github.com/pyro-ppl/pyro/blob/dev/.travis.yml#L35) to recent nightly build due to the recent [mnist download issue](https://github.com/pytorch/vision/issues/3497).

pyro/contrib/examples/util.py
```diff
@@ -4,17 +4,12 @@
...
 from torch.utils.data import DataLoader
 from torchvision import transforms
-from torchvision.datasets import MNIST
 
-from pyro.distributions.torch_patch import patch_dependency
 
-
-@patch_dependency('torchvision.datasets.MNIST', torchvision)
-class _MNIST(getattr(MNIST, '_pyro_unpatched', MNIST)):
+class MNIST(datasets.MNIST):
```
neerajprad, line 12: Thanks for getting rid of this!

neerajprad (approved):

PR #2778, opened 2021-03-07
[docs] added note about graphviz in docs readme

jamestwebber:
Was trying to build the docs so I could submit a different PR and ran into this issue, so I just added a note.

Side question: is there a particular reason that `docs/` has a separate requirements file, as opposed to being another extra in `setup.py`?  I ask partially because there is a fair amount of overlap with the `dev` extra requirements, but not everything.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -10,6 +10,8 @@ make docs
 pip install -r requirements.txt
 ```
 
+Note that you will need to install [graphviz](https://www.graphviz.org/) separately.
+
 ## Workflow ##
 To change the documentation, update the `*.rst` files in `source`.
 


fritzo:
> is there a particular reason that docs/ has a separate requirements file

I believe docs/requirements.txt is needed for [builds on readthedocs.org](https://readthedocs.org/projects/pyro-ppl/builds/), which are triggered on update to the dev or master branch.

fritzo (approved):
Thanks for documenting this!

jamestwebber:
It turns out that I won't have a follow-up change because the thing I was going to add was a note about conda installation, but that's on the website and I don't want to try editing that.

PR #2786, opened 2021-03-23
Fix regression in unfold/Approximate pattern in funsor#488

fritzo:
pair debugged with @eb8680 

This works around a regression in `funsor.optimizer.apply_optimizer()` introduced in https://github.com/pyro-ppl/funsor/pull/488 whereby the `unfold_contraction_generic_tuple` appears to incorrectly interact with `Approximate`. This fix is required for funsor versions after that pull.

## Tested
- tested locally with funsor master branch c77a54e3d434
  `pytest tests/contrib/funsor/test_enum_funsor.py`

### DIFF
--- a/pyro/contrib/funsor/infer/traceenum_elbo.py
+++ b/pyro/contrib/funsor/infer/traceenum_elbo.py
@@ -13,6 +13,17 @@ from pyro.distributions.util import copy_docs_from
 from pyro.infer import TraceEnum_ELBO as _OrigTraceEnum_ELBO
 
 
+# Work around a bug in unfold_contraction_generic_tuple interacting with
+# Approximate introduced in https://github.com/pyro-ppl/funsor/pull/488 .
+# Once fixed, this can be replaced by funsor.optimizer.apply_optimizer().
+def apply_optimizer(x):
+    with funsor.interpretations.normalize:
+        expr = funsor.interpreter.reinterpret(x)
+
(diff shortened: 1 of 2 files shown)


eb8680:
Can you bump the Funsor dependency commit in `setup.py` to the latest `master`?

eb8680:
Will be unblocked by the fix in https://github.com/pyro-ppl/funsor/pull/508

fritzo:
> Can you bump the Funsor dependency commit

No, bumping funsor dependency is unrelated to this change and warrants a separate PR.  Keeping our PRs narrowly focused helps us later git bisect to diagnose errors.  Feel free to bump the funsor dependency in a separate PR.

eb8680:
> bumping funsor dependency is unrelated to this change

The current version of Funsor in `setup.py` is earlier than https://github.com/pyro-ppl/funsor/pull/488 which is why Travis didn't catch this bug earlier. The fix in this PR is thus not exercised on CI unless the Funsor version is bumped to the latest `master`.

fritzo:
As your fix https://github.com/pyro-ppl/funsor/pull/508 shows, bumping funsor involves more than this PR's fix.  Moreover this PR is backwards compatible with older versions of funsor.  Let's try to keep PRs small.

eb8680 (approved):
I don't think this should be merged untested, but in the interest of progress...

PR #2791, opened 2021-04-01
Add a SoftLaplace distribution

fritzo:
From the docs:

>    This distribution corresponds to the log-convex density::
>
>        z = (value - loc) / scale
>        log_prob = log(2 / pi) - log(scale) - logaddexp(z, -z)
>
>    Like the Laplace density, this density has the heaviest possible tails
>    (asymptotically) while still being log-convex. Unlike the Laplace
>    distribution, this distribution is infinitely differentiable everywhere,
>    and is thus suitable for constructing Laplace approximations.

```py
import torch
x = torch.linspace(-4, 4, 200)
plt.plot(x, dist.Laplace(0, 1).log_prob(x), "k--", label="Laplace")
plt.plot(x, dist.SoftLaplace(0, 1).log_prob(x), "r-", label="SoftLaplace")
plt.legend(loc="best");
```
download

## Tested

- [x] added to standard tests, including goftests
- [x] added a cdf-vs-icdf test that is now run on all continuous univariate distributions


pyro/distributions/softlaplace.py
```diff
@@ -0,0 +1,66 @@
...
+        new.scale = self.scale.expand(batch_shape)
+        super(SoftLaplace, new).__init__(batch_shape, validate_args=False)
+        new._validate_args = self._validate_args
+        return new
+
+    def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
+        z = (value - self.loc) / self.scale
+        return math.log(2 / math.pi) - self.scale.log() - torch.logaddexp(z, -z)
```
fehiepsi, line 52: Make senses to me: `logaddexp(z, -z) ~ |z|` in the tail

pyro/distributions/softlaplace.py
```diff
@@ -0,0 +1,66 @@
...
+    def rsample(self, sample_shape=torch.Size()):
+        shape = self._extended_shape(sample_shape)
+        u = self.loc.new_empty(shape).uniform_()
+        return self.icdf(u)
+
+    def cdf(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
+        z = (value - self.loc) / self.scale
+        return z.exp().atan().mul(2 / math.pi)
```
fehiepsi, line 63: Can we have tests for `cdf/icdf`? I think `icdf` is fine because it belongs to `gof` test.

fritzo: Good idea.  I've added a generic cdf-vs-icdf test that is run on all continuous univariate distributions.

fehiepsi (approved):
LGTM! It would be nice to include some [gradient tests](https://github.com/pyro-ppl/numpyro/blob/b93d10afd77d217348c16cbf4189c7b068e65c1d/test/test_distributions.py#L721), which you taught me previously, but let's do it when more distributions need the test.

fritzo:

PR #2792, opened 2021-04-03
assert in JitTraceMeanField_ELBO throws TracerWarning

ahoho:
I removed this assertion since it throws a `TracerWarning` when using `JitTraceMeanField_ELBO`---if there's a better solution, I'm happy to implement it.


pyro/infer/trace_mean_field_elbo.py
```diff
@@ -106,7 +106,6 @@ def _differentiable_loss_particle(self, model_trace, guide_trace):
                         kl_qp = kl_divergence(guide_site["fn"], model_site["fn"])
                         kl_qp = scale_and_mask(kl_qp, scale=guide_site["scale"], mask=guide_site["mask"])
                         if torch.is_tensor(kl_qp):
-                            assert kl_qp.shape == guide_site["fn"].batch_shape
```
fritzo, line 109: Rather than remove the assertion, could you gate by `torch._C.get_tracing_state()` as we do elsewhere in Pyro?
```py
if not torch._C.get_tracing_state():
    assert kl_qp.shape == guide_site["fn"].batch_shape
```

fritzo (changes requested):
Thanks for fixing, I just have a minor suggestion so the assertion still works outside of the jit.

ahoho:
Appreciate the fix, I just added that line

fritzo (dismissed):

ahoho:
Accidentally made some commits to the wrong branch, so reverted them

fritzo:
I don't understand why the build is failing, but it does seem to be caused by your change.

Do you think it would be too inconvenient to simply pass [ignore_jit_warnings=True](http://docs.pyro.ai/en/0.3.1/inference_algos.html#pyro.infer.trace_mean_field_elbo.JitTraceMeanField_ELBO) to your elbo instance?  If so, we could close this PR and avoid debugging those mysteriously failing builds.

ahoho:
Yeah, I think that's reasonable enough---I figured this was a straightforward fix but it appears otherwise---sorry for taking up time with something so trivial!

fritzo:
No worries, we're always happy to have small fixes!  Don't hesitate to submit other fixes you would like, this one just happened to be unusually subtle.

PR #2793, opened 2021-04-05
Allow saving of a subset of variables in MCMC

fritzo:
Addresses https://forum.pyro.ai/t/save-only-a-subset-of-latent-variables-in-mcmc/2753

This allows saving of only a subset of latent variables during MCMC.  This is useful in models with large nuisance variables that take up memory and screen space in `.summary()`.

## Tested
- [x] unit test with variables of different shape

### DIFF
--- a/pyro/infer/mcmc/api.py
+++ b/pyro/infer/mcmc/api.py
@@ -301,7 +301,8 @@ class MCMC:
     :param dict transforms: dictionary that specifies a transform for a sample site
         with constrained support to unconstrained space.
     :param List[str] save_params: Optional list of a subset of parameter names to
-        save during sampling and diagnostics. Defaults to saving all params.
+        save during sampling and diagnostics. This is useful in models with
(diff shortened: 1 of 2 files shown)


pyro/infer/mcmc/api.py
```diff
@@ -428,8 +435,8 @@ def model(data):
                 self.transforms = {}
 
         # transform samples back to constrained space
-        for name, transform in self.transforms.items():
-            z_acc[name] = transform.inv(z_acc[name])
+        for name, z in z_acc.items():
+            z_acc[name] = self.transforms[name].inv(z)
```
fehiepsi, line 439: I think you need to check if name in self.transforms to perform this action.

fehiepsi (commented):
LGTM. I just have the following comment

fehiepsi (approved):

fritzo:

PR #2796, opened 2021-04-07
Minor cleanup: removed try/except around numpy import

jamestwebber:
`numpy` is a required dependency of `pytorch` so this should be safe.

This change only affects `pyro.util` and the `minipyro` implementation. The tests for the latter fail, but that seems like a pre-existing condition.

### DIFF
--- a/pyro/contrib/minipyro.py
+++ b/pyro/contrib/minipyro.py
@@ -19,6 +19,7 @@ import warnings
 import weakref
 from collections import OrderedDict
 
+import numpy as np
 import torch
 
 from pyro.distributions import validation_enabled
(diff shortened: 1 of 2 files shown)


pyro/contrib/minipyro.py
```diff
@@ -119,15 +120,12 @@ def __init__(self, fn=None, rng_seed=None):
...
-        self.old_state = {'torch': torch.get_rng_state(), 'random': random.getstate()}
+        self.old_state = {
+            'torch': torch.get_rng_state(), 'random': random.getstate(), 'numpy': np.random.get_state()
+        }
         torch.manual_seed(self.rng_seed)
         random.seed(self.rng_seed)
-        try:
-            import numpy as np
-            np.random.seed(self.rng_seed)
-            self.old_state['numpy'] = np.random.get_state()
```
jamestwebber, line 128: This was maybe in the wrong order before?

fritzo (approved):
Thanks for cleaning up!

jamestwebber:
Turns out this was premature! Was reading the changelog for pytorch 1.9 and numpy is no longer listed as a dependency, so perhaps this should be reverted.

PR #2800, opened 2021-04-08
Add SoftLaplace.mean, .variance properties

fritzo:
Port of https://github.com/pyro-ppl/numpyro/pull/1002

## Tested
- [x] added a new `test_mean()` which is run on all standard distributions
- [x] added a new `test_variance()` which is run on all standard distributions


fehiepsi (approved):
LGTM!

PR #2802, opened 2021-04-13
Implement sparse Poisson.log_prob() computation

fritzo:
Similar to #1740,  https://github.com/pyro-ppl/numpyro/pull/1003

This adds an `is_sparse` constructor arg to `dist.Poisson` to trigger a cheaper computation if most data is zero.

## Tested
- [x] registered standard unit test


fehiepsi (approved):
LGTM!

PR #2812, opened 2021-04-21
Add an AutoStructured guide and StructuredReparam

fritzo:
Addresses #2813

This adds a flexible `AutoStructured` guide that allows a variety of distributions modeling each latent site (Delta, Normal, or MultivariateNormal), together with a mechanism to declare (link-)linear dependencies between latent variables.  As discussed with @fehiepsi this aims to (1) generalize guides with arrowhead covariance structure while (2) learning parameters that can be cheaply used to precondition NUTS via a reparameterizer `StructuredReparam`.

This also adds a simple `StructuredReparam` that uses a trained `AutoStructured` guide to precondition a model for use in HMC.  This new (guide,reparam) pair can be seen as a structured version of the monolithic (`AutoContinuous`,`NeuTraReparam`) pair in the same sense that `AutoNormal` is a structured version of the monolithic `AutoDiagonalNormal` guide.

My main motivation is to use this for high-dimensional models (e.g. 300000 latent variables) with a structured precision matrix, and then use that structured precision matrix as a preconditioner for NUTS.

(Note this does not implement [Automatic structured variational inference](https://arxiv.org/pdf/2002.00643.pdf), a variational family whose stricture is severely limited to dependencies in the model.  Nor does this first PR implement automatic suggestion of the guide structure as in [Faithful inversion of generative models for effective amortized inference](https://dl.acm.org/doi/10.5555/3327144.3327229).)

## Tested
- [x] unit tests for `AutoStructured`
- [x] unit tests for `StructuredReparam` 
- [x] test on a real world problem ([private repo](https://github.com/broadinstitute/pyro-cov/blob/master/pyrocov/mutrans.py) from which this was abstracted)


pyro/infer/autoguide/guides.py
```diff
@@ -1190,3 +1188,280 @@ def forward(self, *args, **kwargs):
...
+
+            # Accumulate upstream dependencies.
+            # Note: by accumulating upstream dependencies before updating the
+            # aux_values dict, we encode a block-sparse structure of the
+            # precision matrix; if we had instead accumulated after updating
+            # aux_values, we would encode a block-sparse structure of the
+            # covariance matrix.
+            # Note: these shear transforms have no effect on the Jacobian
+            # determinant, and can therefore be excluded from the log_density
+            # computation below, even for nonlinear dep().
```
fehiepsi, line 1413: Nice property!

pyro/infer/autoguide/guides.py
```diff
@@ -1190,3 +1188,280 @@ def forward(self, *args, **kwargs):
...
+    scale_tril_constraint = constraints.softplus_lower_cholesky
+
+    def __init__(
+        self,
+        model,
+        *,
+        conditionals: Dict[str, Union[str, Callable]] = "normal",
+        dependencies: Dict[str, Dict[str, Union[str, Callable]]] = "linear",
+        init_loc_fn=init_to_feasible,
+        init_scale=0.01,
```
fehiepsi, line 1268: This is different from other guides. Do we need a change here?

pyro/infer/autoguide/guides.py
```diff
@@ -1190,3 +1188,280 @@ def forward(self, *args, **kwargs):
...
+                )
+            elif conditional == "mvn":
+                # This overparametrizes by learning (scale,scale_tril),
+                # enabling faster learning of the more-global scale parameter.
+                scale = _deep_getattr(self.scales, name)
+                scale_tril = _deep_getattr(self.scale_trils, name)
+                scale_tril = scale[..., None] * scale_tril
+                aux_value = pyro.sample(
+                    name + "_aux",
+                    dist.MultivariateNormal(zero, scale_tril=scale_tril),
```
fehiepsi, line 1399: If we factor this out to `scale_tril @ Normal(0, 1)`, I guess HMC will be a bit happier.

fehiepsi (commented):
Looks clean to me! I can confirm that this is equivalent to if we learn arrowhead matrix directly:
```python
A = [[L @ L.t + w @ D @ w.T, w @ D], [D @ w.T, D]]
```
where L is scale_tril of `x_aux`, D is the variance of `y_aux`, w is dep.weight.

pyro/infer/autoguide/guides.py
fritzo: Done. :+1: for consistency

pyro/infer/autoguide/guides.py
fritzo: Great point, I guess that is equivalent to reparametrizing.  I've also changed `Normal(0,scale)` to `Normal(0,1) * scale` in the "normal" case.

pyro/infer/autoguide/guides.py
```diff
@@ -1382,23 +1382,24 @@ def get_deltas(self, save_params=None):
...
-                scale_tril = scale[..., None] * scale_tril
                 aux_value = pyro.sample(
                     name + "_aux",
-                    dist.MultivariateNormal(zero, scale_tril=scale_tril),
+                    dist.Normal(zero, 1).to_event(1),
                     infer={"is_auxiliary": True},
                 )
+                scale = _deep_getattr(self.scales, name)
+                scale_tril = _deep_getattr(self.scale_trils, name)
+                aux_value = aux_value @ scale_tril.T * scale
```
fehiepsi, line 1402: I think you will need to add [logdet](https://github.com/pyro-ppl/pyro/blob/6ec3be8186fb669cdb2dc797920444ba0e5904ca/pyro/distributions/transforms/lower_cholesky_affine.py#L66) of those affine transforms.

pyro/infer/autoguide/guides.py
fehiepsi: I think you will need to add [logdet](https://github.com/pyro-ppl/pyro/blob/6ec3be8186fb669cdb2dc797920444ba0e5904ca/pyro/distributions/transforms/lower_cholesky_affine.py#L66) of those affine transforms. How about using `dist.TransformedDistribution(dist.Normal(...), LowerCholeskyAffine(...))` so that we can use `TransformReparam` in the reparam?

fritzo: Thanks, I've added the logdet terms by hand here since it's simpler.  Does it look right now?

fehiepsi: Yes, it looks correct to me.

fehiepsi (approved):
This is great! I'll try to find some time next week to port this to NumPyro (hopefully it will be straightforward).

fritzo:
Thanks for reviewing @fehiepsi .  I'll try to add automated dependency structure with @eb8680 next week.

pyro/infer/autoguide/guides.py
fehiepsi: I believe two versions are equivalent... Not sure what's going on. Let me play with some tests to see if elbo is the same for the two versions.

fritzo: Thanks I'll do the same, at least to create a unit test I can run locally (not on some huge model on a GPU cloud machine)

fehiepsi: I think I find the issue. Here `log_density` is calculated as the sum over all dimensions of the site. However, the `ldj` term, which is used to calculate the logdet of unconstrained->constrained values, maintains the batch dimension. So sum of them will give wrong result if this site is under some plate. I guess we should use `pyro.factor` for those `log_density` terms. What do you think?

fritzo: @fehiepsi thanks, yes I now see the error.  I'll think about this and submit a fix ASAP.

PR #2814, opened 2021-04-22
Simplify param names in callable optim configs

fritzo:
This simplifies the interface of per-parameter callable optim configs, and updates it to make sense when `PyroModule` replaces `pyro.module` as the default interface. Basically:
```diff
- def optim_config(module_name, param_name):
+ def optim_config(param_name):
```

**BEFORE** this PR the callable took a pair (`module_name`, `param_name`), where:
- for parameters registered through `pyro.module`, `module_name` was the name passed to `pyro.module` and `param_name` was the fully-qualified name within that module.
- for parameters registered through `PyroModule...`, `module_name` == `param_name` == the fully-qualified name of the parameter, e.g. `Guide.locs.rate` (yes this is weird).

**AFTER** this PR the callable takes a single `param_name` which is the fully qualified name as in `nn.Module.named_parameters()`, e.g. `Guide.locs.rate`.  Crucially, this name remains the same even if we convert a submodule from an nn.Module to a PyroModule via `to_pyro_module_()` (and i've added a test for this).

This also fixes a bug in edge-case interactions between `PyroModule` and `to_pyro_module_()`.

## Backwards compatibility.

The old behavior is still supported but is no longer mentioned in documentation or examples.

## Tested
- [x] changed all occurrences in existing tests and examples
- [x] added test that names are not changed by `to_pyro_module_()`


fritzo:
- @martinjankowiak could you please review the changes to the optim interface?
- @fehiepsi could you please review the small change to `PyroModule`?

martinjankowiak (dismissed):

tests/optim/test_optim.py
```diff
@@ -231,3 +234,56 @@ def model():
...
+    adam(params)
+    assert param_names
+    actual_param_names = param_names.copy()
+    del adam, loss
+    param_names.clear()
+    pyro.clear_param_store()
+
+    print(expected_param_names)
+    print(actual_param_names)
+    assert actual_param_names == expected_param_names
```
fehiepsi, line 289: I saw some simplifications but still couldn't figure out what are the expected names here. Could you make this assertation more verbose like
```
assert acutal_param_names == ...
assert actual_param_names == expected_param_names
```

fritzo: Done.

fehiepsi (approved):
The change is a bit subtle but LGTM.

fritzo:
Thanks for reviewing!

PR #2821, opened 2021-04-24
Bivariate von Mises Distribution

OlaRonning:
This PR introduces the Sine Bivariate von Mises (BvM) distribution which is a distribution on the torus described by [Singh et al.](https://www.jstor.org/stable/4140614?seq=1). The sampling method works for the univariate case and follows the process developed by [Kent et al.](https://www.tandfonline.com/doi/full/10.1080/10618600.2017.1390468). 

In bioinformatics, BvM is well suited for molding torsion angles for peptide bonds. I'll submit an example using this distribution together with the [Sine Skewed distribution](https://github.com/aleatory-science/pyro/blob/feature/skewed_bivaritate_von_mises/pyro/distributions/bivariate_von_mises.py#L58-L83) in a separate PR.

Sampling may fail if the distribution is bimodal; I've added a [`raise Exception`](https://github.com/aleatory-science/pyro/blob/feature/BvM/pyro/distributions/bivariate_von_mises.py#L208-L209) if this occurs. However, I'm not sure of the preferred behavior in this case.


pyro/distributions/bivariate_von_mises.py
```diff
@@ -0,0 +1,257 @@
...
+        max_iter = SineBivariateVonMises.max_sample_iter
+
+        # flatten batch_shape
+        conc = conc.view(2, -1, 1)
+        eigmin = eigmin.view(-1, 1)
+        corr = corr.view(-1, 1)
+        eig = eig.view(2, -1)
+        b0 = b0.view(-1)
+        phi_den = log_I1(0, conc[1]).view(-1, 1)
+
```
fritzo, line 176: can you add a check to ensure this isn't jitted?
```py
assert not torch._C._get_tracing_state(), "jit not supported"
```

pyro/distributions/bivariate_von_mises.py
```diff
@@ -0,0 +1,257 @@
...
+                torch.arange(total).view(1, -1) >= start[missing > 0].view(-1, 1))
+
+            phi[:, phi_mask] = x[:, accepted]
+
+            start[missing > 0] += accepted.sum(-1)
+            missing[missing > 0] -= accepted.sum(-1)
+            max_iter -= 1
+
+        if max_iter == 0 or torch.any(missing > 0):
+            raise Exception("sample error")  # FIXME: what would be an appropriate exception?
```
fritzo, line 209: maybe a
```py
raise ValueError("maximum number of iterations exceeded; "
                 "try increasing `SineBivariateVonMises.max_sample_iter")
```

tests/distributions/test_bivariate_von_mises.py
```diff
@@ -0,0 +1,122 @@
+# Copyright Contributors to the Pyro project.
```
fritzo, line 1: Could you also add a standard test case to tests/distributions/conftest.py?  These tests check for standard things like shape errors and correct metadata.  You may need to add special case handling logic in a few places like test_gof; let me know if you'd like any help working through test errors.

fritzo (changes requested):
Hi @OlaRonning, nice contribution.  I'm curious what inference algorithms you'd like to support with this distribution.  As I understand, this will support MCMC because the `.log_prob()` is differentiable, but SVI would be trickier because `.sample()` is not reparametrizable.  Often when creating new distributions I mention in the docstring which inference algorithms are supported and what parameters can be learned.

OlaRonning:
Hi @fritzo,

Thanks for the review! I had not considered the issue with reparameterizing and intended to use SVI; however, HMC should still work. I'll update the docstring as you suggest and look into the possibility of a transform RxR -> S^1xS^1 (not for this PR :)).

pyro/distributions/bivariate_von_mises.py
```diff
@@ -0,0 +1,257 @@
...
+    seqs = (2 * lvalues[None, :, :] - lfactorials[None, None, :] - lgammas.gather(1, indices)[:, None, :]).logsumexp(-1)
+    assert seqs.shape == (orders, _flat_dim(vshape))
+
+    i1s = lvalues[..., :orders].T + seqs
+    assert i1s.shape == (orders, _flat_dim(vshape))
+    return i1s.view(-1, *vshape)
+
+
+class SineBivariateVonMises(TorchDistribution):
+    """ Distribution of two dependent angles on the 2-torus (S^1 ⨂ S^1) given by
```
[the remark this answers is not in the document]
OlaRonning: I'll check and correct the docstrings; thanks for the tip; I did not know raw strings worked like that.

pyro/distributions/bivariate_von_mises.py
```diff
@@ -0,0 +1,257 @@
...
+
+    :param torch.Tensor phi_loc: location of first angle
+    :param torch.Tensor psi_loc: location of second angle
+    :param torch.Tensor phi_concentration: concentration of first angle
+    :param torch.Tensor psi_concentration: concentration of second angle
+    :param torch.Tensor correlation: correlation between the two angles
+    :param torch.Tensor weighted_correlation: set correlation to
+    .. math:
+        w_\\rho\\sqrt{\\kappa_1\\kappa_2},
+    to avoid bimodality (see note).
```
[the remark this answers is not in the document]
OlaRonning: You maybe right, I'll check and correct as required.

pyro/distributions/bivariate_von_mises.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+
+    arg_constraints = {'phi_loc': constraints.real, 'psi_loc': constraints.real,
+                       'phi_concentration': constraints.positive, 'psi_concentration': constraints.positive,
+                       'correlation': constraints.real}
+    support = constraints.real
```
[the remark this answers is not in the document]
OlaRonning: Thanks that makes sense -- I'll update this.

fritzo:
Hi @OlaRonning,  for SVI you might also consider a [ProjectedNormal](https://docs.pyro.ai/en/dev/distributions.html#pyro.distributions.ProjectedNormal) prior together with a [ProjectedNormalReparam](https://docs.pyro.ai/en/dev/infer.reparam.html#pyro.infer.reparam.projected_normal.ProjectedNormalReparam) and an `AutoMultivariateNormal` guide.  The `ProjectedNormal` distribution and reparametrizer aims to support the reparametrization trick while working around the torus topology by working in a higher dimensional euclidean space and projecting down to spherical (or toroidal) space.  This combination supports learning of a bimodal posterior.

pyro/distributions/bivariate_von_mises.py
OlaRonning: Thanks for the suggestion; in my experience, when `max_iter` is hit, few to no samples are ever accepted, so the increase required is likely to be massive. 

I'll look into whether `max_iter` can be lower bounded or the growth can be estimated based on the distribution parameters, and maybe this estimate can be added to the error message.

tests/distributions/test_bivariate_von_mises.py
OlaRonning: I'll give it a shot and ping you if I can't figure out the cause myself.

OlaRonning:
That's neat, I'll definitely try this out! Thanks.

OlaRonning:
Using [`infer.Predictive`](https://github.com/aleatory-science/pyro/blob/example/ssbvm_mixture/examples/ssbvm_bmixture.py#L86), I could not use `autodiff` in `BvM._bfind`; I resolved this by manually computing the first and second-order derivates.

pyro/distributions/sine_bivariate_von_mises.py
```diff
@@ -0,0 +1,280 @@
...
+from pyro.distributions import constraints
+from pyro.distributions.torch_distribution import TorchDistribution
+from pyro.distributions.util import broadcast_shape
+
+
+def _flat_dim(size):
+    return torch.prod(torch.tensor(size), dim=0)
+
+
+def log_I1(orders: int, value: torch.Tensor, terms=250):
```
fritzo, line 22: If you think this might be used elsewhere, you could move it to [pyro.ops.special](https://github.com/pyro-ppl/pyro/blob/dev/pyro/ops/special.py).

pyro/distributions/sine_bivariate_von_mises.py
```diff
@@ -0,0 +1,280 @@
...
+import torch
+from torch.distributions import VonMises
+from torch.distributions.utils import broadcast_all
+
+from pyro.distributions import constraints
+from pyro.distributions.torch_distribution import TorchDistribution
+from pyro.distributions.util import broadcast_shape
+
+
+def _flat_dim(size):
```
fritzo, line 18: nit: replace `_flat_dim(shape)` with `shape.numel()` everywhere

pyro/distributions/sine_bivariate_von_mises.py
```diff
@@ -0,0 +1,280 @@
...
+    :param orders: orders of the log modified bessel function.
+    :param value: values to compute modified bessel function for
+    :param terms: truncation of summation
+    :return: 0 to orders modified bessel function
+    """
+    orders = orders + 1
+    if len(value.size()) == 0:
+        vshape = torch.Size([1])
+    else:
+        vshape = copy(value.shape)
```
fritzo, line 38: Just curious: I believe shapes are immutable, so the need for the `copy(-)` surprises me.  Do you recall the reason?

pyro/distributions/sine_bivariate_von_mises.py
```diff
@@ -0,0 +1,280 @@
...
+        self.psi_loc = psi_loc
+        self.phi_concentration = phi_concentration
+        self.psi_concentration = psi_concentration
+        self.correlation = correlation
+        self._norm_const = None
+        event_shape = torch.Size([2])
+        batch_shape = phi_loc.shape
+        super(SineBivariateVonMises, self).__init__(batch_shape, event_shape, validate_args)
+
+    @property
```
fritzo, line 143: I would recommend using `@lazy_property` here, so as to use existing distributions idioms.  That will also allow you to remove the `self._norm_const` member:
```py
from torch.distributions.utils import lazy_property

class SineBivariateVonMises(TorchDistribution):
    ...
    @lazy_property
    def norm_const(self):
        corr = self.correlation.view(1, -1) + 1e-8
        ...
        return norm_const.reshape(self.phi_loc.shape)
```

tests/distributions/conftest.py
```diff
@@ -305,19 +305,26 @@ def __init__(self, rate, *, validate_args=None):
...
                 {'concentration': [0., 0., 0.], 'test_data': [1., 0., 0.]},
                 {'concentration': [-1., 2., 3.], 'test_data': [0., 0., 1.]},
-                ]),
+            ]),
+    Fixture(pyro_dist=dist.SineBivariateVonMises,
+            examples=[
+                {'phi_loc': [0.], 'psi_loc': [0.], 'phi_concentration': [5.], 'psi_concentration': [6.],
+                 'correlation': [2.], 'test_data': [[1., 0.]]},
+                {'phi_loc': [0.], 'psi_loc': [0.], 'phi_concentration': [5.], 'psi_concentration': [6.],
+                 'weighted_correlation': [.5], 'test_data': [[1., 0.]]}
```
fritzo, line 326: Can you add one or two more examples that are batched, so as to test batch shape logic?

tests/distributions/test_distributions.py
```diff
@@ -108,7 +108,10 @@ def test_score_errors_non_broadcastable_data_shape(dist):
         if dist.get_test_distribution_name() == 'LKJCholesky':
             pytest.skip('https://github.com/pytorch/pytorch/issues/52724')
         shape = d.shape()
-        non_broadcastable_shape = (shape[0] + 1,) + shape[1:]
+        if dist.get_test_distribution_name() == 'SineBivariateVonMises':
+            non_broadcastable_shape = (shape[0] + 1,) + (shape[1] + 1,)
+        else:
+            non_broadcastable_shape = (shape[0] + 1,) + shape[1:]
```
fritzo, line 114: You shouldn't need to special case here. I believe this test is intended to find broadcasting errors among batch shapes, whereas the `(shape[1] + 1,)` is a more fundamental error in event shape.  If you'd like to test event shape errors, you could add a bespoke test.

tests/distributions/test_sine_bivariate_von_mises.py
```diff
@@ -0,0 +1,131 @@
...
+from tests.common import assert_equal
+
+
+def _unnorm_log_prob(value, loc1, loc2, conc1, conc2, corr):
+    if len(value.shape) == 1:
+        phi_val = value[0]
+        psi_val = value[1]
+    else:
+        phi_val = value[..., 0]
+        psi_val = value[..., 1]
```
fritzo, line 23: fyi: the else branch should work for both cases; there is no need for the if branch.

tests/distributions/test_sine_bivariate_von_mises.py
```diff
@@ -0,0 +1,131 @@
...
+    assert_equal(log_I1(10, tensor([[.6]])).shape, torch.Size([11, 1, 1]))
+    assert_equal(log_I1(10, tensor([.6, .2])).shape, torch.Size([11, 2]))
+    assert_equal(log_I1(0, tensor(.6)).shape, torch.Size((1, 1)))
+
+
+@pytest.mark.parametrize('order', [0, 1, 5, 10, 20])
+def test_log_I1(order):
+    expected = torch.tensor([iv(i, tensor((.6,)).numpy()) for i in range(order + 1)]).log()
+    actual = log_I1(order, tensor([.6]))
+    assert_equal(actual, expected)
```
fritzo, line 47: If you move `log_I1` to pyro.ops.special, you could move these tests to tests/ops/test_special.py.

I'd also recommend a small test checking values against scipy.

tests/distributions/test_sine_bivariate_von_mises.py
```diff
@@ -0,0 +1,131 @@
...
+            corr = b.sample((*batch_dim,))
+            if torch.all(torch.prod(conc, dim=0) > corr ** 2):
+                break
+
+        bmv = SineBivariateVonMises(locs[0], locs[1], conc[0], conc[1], corr)
+        assert_equal(bmv.batch_shape, torch.Size(batch_dim))
+        assert_equal(bmv.sample(sample_dim).shape, torch.Size((*sample_dim, *batch_dim, 2)))
+
+
+def test_mle_bvm():
```
fritzo, line 88: :+1: nice test!

pyro/distributions/sine_bivariate_von_mises.py
```diff
@@ -0,0 +1,280 @@
...
+
+        .. math::
+
+            \frac{\rho}{\kappa_1\kappa_2} \rightarrow 1
+
+        because the distribution becomes increasingly bimodal.
+
+    .. note:: The correlation and weighted_correlation params are mutually exclusive.
+
+    .. note:: This distribution does not work with :class:`~pyro.infer.svi.SVI`.
```
fritzo, line 96: I think you're underselling here.  Maybe say

> In the context of `SVI`, this distribution can be used as a likelihood but not for latent variables.

Even then, I don't understand why it cannot be used for latent variables.  I see that it has a slow sampler and is not reparametrized, but it should still lead to slow-but-correct inference.  Am I missing something?

pyro/distributions/sine_bivariate_von_mises.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def __init__(self, phi_loc, psi_loc, phi_concentration, psi_concentration, correlation=None,
+                 weighted_correlation=None, validate_args=None):
+
+        assert (correlation is None) != (weighted_correlation is None)
+
+        if weighted_correlation is not None:
+            correlation = weighted_correlation * (phi_concentration * psi_concentration).sqrt() + 1e-8
+
+        if torch.any(phi_concentration * psi_concentration <= correlation ** 2):
```
fritzo, line 124: Can you move this check after `super().__init__(...)` and gate it by `self._validate_args`?
```py
if self._validate_args and torch.any(...):
    warnings.warn(...)
```
I think that will be safer when using the PyTorch jit.

pyro/distributions/sine_bivariate_von_mises.py
```diff
@@ -0,0 +1,280 @@
...
+                                                                                            correlation)
+        self.phi_loc = phi_loc
+        self.psi_loc = psi_loc
+        self.phi_concentration = phi_concentration
+        self.psi_concentration = psi_concentration
+        self.correlation = correlation
+        self._norm_const = None
+        event_shape = torch.Size([2])
+        batch_shape = phi_loc.shape
+        super(SineBivariateVonMises, self).__init__(batch_shape, event_shape, validate_args)
```
fritzo, line 141: nit: simplify to `super().__init__(batch_shape, event_shape, validate_args))`

pyro/distributions/sine_bivariate_von_mises.py
```diff
@@ -0,0 +1,280 @@
...
+            return super(SineBivariateVonMises, self).expand(batch_shape)
+        except NotImplementedError:
+            validate_args = self.__dict__.get('_validate_args')
+            phi_loc = self.phi_loc.expand(batch_shape)
+            psi_loc = self.psi_loc.expand(batch_shape)
+            phi_conc = self.phi_concentration.expand(batch_shape)
+            psi_conc = self.psi_concentration.expand(batch_shape)
+            corr = self.correlation.expand(batch_shape)
+
+            return type(self)(phi_loc, psi_loc, phi_conc, psi_conc, corr, validate_args=validate_args)
```
fritzo, line 267: Could you use the more modern idiom, say [following SoftLaplace]():
```py
    new = self._get_checked_instance(SineBivariateVonMises, _instance)
    new.psi_loc = self.psi_loc.expand(batch_shape)
    new.phi_loc = self.phi_loc.expand(batch_shape)
    new.psi_conc = self.psi_conc.expand(batch_shape)
    new.phi_conc = self.phi_conc.expand(batch_shape)
    new.correlation = self.correlation.expand(batch_shape)
    super(SineBivariateVonMises, new).__init__(batch_shape, self.event_shape, validate_args=False)
    new._validate_args = self._validate_args
    return new
```

fritzo (changes requested):
Looks good!

I have checked coding idioms but not checked math.
1.  Can you please confirm that the the examples params in `tests/distributions/conftest.py` exercise a sufficiently diverse range of parameter space, such that `test_gof` can be trusted to verify agreement between `.log_prob()` and `.sample()`?  I see that this test passes
    ```sh
    pytest -vx tests/distributions/test_distributions.py::test_gof -k SineBiv
    ```
    but I want to make sure you're testing with some symmetry breaking parameters, e.g. randomly sampled from their supports, or otherwise asymmetric so as to detect math errors like writing `sin` instead of `cos`.
2.  Is there any math you'd like checked in particular?

tests/distributions/conftest.py
fritzo: Also as commented above, could you add some nontrivial examples of `phi_loc` and `psi_loc`?

pyro/distributions/sine_bivariate_von_mises.py
OlaRonning: Thanks; I've was wondering why there wasn't already a method like this.

pyro/distributions/sine_bivariate_von_mises.py
OlaRonning: It shows up in normalization constants for directional distribution (e.g. [Kent distribution](https://en.wikipedia.org/wiki/Kent_distribution)). I'll move it to `special`.

pyro/distributions/sine_bivariate_von_mises.py
OlaRonning: It was to make sure `vshape` wasn't overwritten; however, this is not necessary, as you correctly point out. I'll remove it.

pyro/distributions/sine_bivariate_von_mises.py
OlaRonning: I see, thanks. I'll change this.

pyro/distributions/sine_bivariate_von_mises.py
OlaRonning: I'll need to read up on how jit works in PyTorch to understand why this is necessary. For now, I've moved as requested.

pyro/distributions/sine_bivariate_von_mises.py
OlaRonning: Yes, of course. I'll update to 
``` python
  new = self._get_checked_instance(SineBivariateVonMises, _instance)
  for k in SineBivariateVonMises.arg_constraints.keys():
    setattr(new, k, getattr(self, k).expand(batch_shape))
  super(SineBivariateVonMises, new).__init__(batch_shape, self.event_shape, validate_args=False)
  new._validate_args = self._valdite_args
  return new
```

tests/distributions/test_sine_bivariate_von_mises.py
OlaRonning: I'll move this to `test_special.py`. By checking values against SciPy, do you mean checking more values than just .6 as `test_log_I1` does?

tests/distributions/test_distributions.py
OlaRonning: Thanks for clarifying; I'll move the test to `test_sine_bivariate_von_mises`.

pyro/distributions/sine_bivariate_von_mises.py
fritzo: Let me try to explanation of why this is necessary in more detail:
1. The PyTorch jit records tensor ops but not Python control flow. Therefore the computation in this check would be recorded, but the `if` decision would be frozen to the first iteration. Therefore the easiest solution is to simply disable this check during jitted inference.
2. The Pyro jit wrappers [disable validation globally](https://github.com/pyro-ppl/pyro/blob/dev/pyro/ops/jit.py#L92) for compiled code. Therefore we'll need to gate this line by the global validation state, which is inherited as `self._validate_args` but only after `super().__init__()` has been called.

tests/distributions/test_sine_bivariate_von_mises.py
fritzo: Oh, indeed I missed your test.  Yes, I'd test against a range of values e.g.
```py
torch.tensor([0.01, 0.1, 0.6, 1.0, 10.0, 100.0])
```

OlaRonning: That makes sense; I'll add this!

OlaRonning:
Thanks for the second review, @fritzo. I've added some more fixtures as you requested, and I believe these now sufficient exercise critical parameterizations (i.e., loc close to pi, high correlation, low correlation, and low concentration).

It would be nice to avoid the transcendental functions when sampling the second angle; however, this would involve reimplementing the von Mises distribution. [Kent et al.](https://www.tandfonline.com/doi/full/10.1080/10618600.2017.1390468) provide an R implementation under supplementary material, together with their article, I believe the maths is reasonably straight forward. 

The main difference between the R implementation and this is I [log transform](https://github.com/aleatory-science/pyro/blob/feature/BvM/pyro/distributions/sine_bivariate_von_mises.py#L162-L168]) the bound and [batch sampling](https://github.com/aleatory-science/pyro/blob/feature/BvM/pyro/distributions/sine_bivariate_von_mises.py#L173-L175), so those would be the critical points.

tests/distributions/conftest.py
OlaRonning: Yes, that makes sense. I'll add some more fixtures.

pyro/distributions/sine_bivariate_von_mises.py
OlaRonning: I was under the impression that the gradients estimations would be too noisy to converge to something trustworthy without reparameterization. However, I'm not particularly confident in this assertion.

fritzo: Sounds good, in that case I'd simply say something like
```
.. note:: This distribution does not generate reparameterized samples, and may
    result in slow convergence when used for latent variables in the context of SVI.
```

OlaRonning:
With the bug fix in 1789fb0, my mixture model is behaving as expected.

image

fritzo:
Hi @OlaRonning, could you try merging dev and pushing? I'm unsure why doc tests are failing.

OlaRonning:
Hi @fritzo, been slightly distracted by deadlines. I'll work out the problem with docs on Thursday if #2838 doesn't resolve it.

PR #2822, opened 2021-04-24
hmm example with vectorized time dimension

ordabayevy:
This PR adds `model_7` to the list of hmm models. `model_7` makes use of `pyro.vectorized_markov` to construct a model and `TraceMarkovEnum_ELBO` to construct the loss function. It is super fast!


fritzo (approved):
Thanks @ordabayevy this looks great!

PR #2823, opened 2021-04-25
Switch to softplus transforms for autoguide scales

fritzo:
@vitkl's experiments in https://github.com/pyro-ppl/sandbox/pull/14 suggest that softplus transforms are more numerically stable for scale parameters in guides, in agreement with anecdotal experience of @martinjankowiak.

This PR changes to use softplus transforms by default in autoguides.

### DIFF
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -433,9 +433,7 @@ class AutoNormal(AutoGuide):
         automatically as usual. This is useful for data subsampling.
     """
 
-    # TODO consider switching to constraints.softplus_positive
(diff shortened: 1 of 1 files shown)


fehiepsi (approved):

fritzo:
@fehiepsi this PR conflicts with https://github.com/pyro-ppl/pyro/pull/2812.  After you merge one of those, please let me update the other to resolve merge conflict (which git will not automatically detect).

fehiepsi:
@fritzo I just merged the other one.

fritzo:
OK, looks like all the guides use softplus transforms now.  Thanks for reviewing!

Issue #2813, opened 2021-04-21
Automatic structured variational preconditioners for HMC


fritzo:
This issue proposes a workflow speed up HMC in high dimensions: first train a structured Gaussian guide via SVI (which is cheap), then use the structured sqrt(covariance) as a preconditioner in HMC.

## Components

- [ ] #2812 An `AutoStructured` guide that allows users to create a structured Gaussian/Delta variational distribution that can be learned via SVI. Users specify the (1) learnable conditional distribution at each latent site, and (2) a set of learnable link-linear dependencies of each latent variable on a subset of other latent variables (this structure can be sparse and is not limited to structure present in the prior).
- [ ] A `StructuredReparam` analogous to `NeutraReparam` but for the simpler family of Gaussian/Delta distributions (we can basically copy the pattern from `Neutra`).  This is intended to be used by HMC and NUTS in lieu of full-rank mass matrix adaptation (which may be infeasible in high dimensions). 
- [ ] An "auto" option to automatically determine posterior dependency structure from dependencies in the model. These can be computed from
- [ ] Generalizations to allow not only Gaussian/Delta but also normalizing flows as conditional distributions.
- [ ] Generalizations to allow not only link-linear dependencies but also more general neural dependencies.

Issue #2550, opened 2020-07-06
Using Python type hints


jpchen:
Since Pyro supports exclusively Python 3, we should look into adding [type hints](https://docs.python.org/3/library/typing.html). This has a few advantages:
- pep8 linters can check for type errors
- easier for a reader to understand function signatures. useful for eg [funsor](https://github.com/pyro-ppl/funsor) where multiple dispatching is prevalent.
- sphinx docs can [autogenerate types](https://pypi.org/project/sphinx-autodoc-typehints/) of args
- IDE static analysis tools will work (better)

e.g.
```
from typing import List, Optional

def foo(bar: List, baz: Optional[List] = None):
    return []
```

fritzo:
I guess we should add a mypy test stage

fritzo:
This would also allow us to remove sphinx type annotations, if we were to use https://pypi.org/project/sphinx-autodoc-typehints/

kamathhrishi:
Can I work on this issue? Adding type hints to the existing codebase?

fritzo:
Hi @kamathhrishi, sure, we'd appreciate any help!  I especially like the idea of a bunch of small easy-to-review PRs that add type hints to one module at a time.

kamathhrishi:
Thank you @fritzo. I have made a PR for optim module.

fritzo:
It looks like we should also [add `py.types` stub files](https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker) to various directories to avoid errors like
```
infer.py:4: error: Skipping analyzing "pyro": module is installed, but missing library stubs or py.typed marker
infer.py:4: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
infer.py:5: error: Skipping analyzing "pyro.distributions": module is installed, but missing library stubs or py.typed marker
infer.py:8: error: Skipping analyzing "pyro.infer": module is installed, but missing library stubs or py.typed marker
infer.py:9: error: Skipping analyzing "pyro.infer.autoguide": module is installed, but missing library stubs or py.typed marker
infer.py:10: error: Skipping analyzing "pyro.infer.reparam": module is installed, but missing library stubs or py.typed marker
infer.py:11: error: Skipping analyzing "pyro.optim": module is installed, but missing library stubs or py.typed marker
```

willtai:
Can I take a look at parts of this? I can open a PR for `params`

PR #2824, opened 2021-04-26
Add automatic dependency tracking to AutoStructured guide

fritzo:
Addresses #2813 
`get_dependencies()` was pair coded with @eb8680 

This adds a helper `get_dependencies()` and uses that to automatically create a sparse dependency graph in the `AutoStructured` guide.

## Tasks
- [x] Implement a `get_dependencies()` helper
- [x] Integrate automatic guide structure in `AutoStructured`
- [x] Convert prior structure to posterior structure as in [(Webb et al. 2018)](https://dl.acm.org/doi/10.5555/3327144.3327229)

## Tested
- [x] tests for `get_dependencies()`
- [x] smoke tests for `AutoStructured` with default params
- [x] test of posterior guide structure


pyro/infer/autoguide/guides.py
```diff
@@ -1314,6 +1328,19 @@ def _setup_prototype(self, *args, **kwargs):
...
             init_locs[name] = init_loc.reshape(-1)
 
+        # Instantiate conditionals and dependencies as dictionaries.
+        if not isinstance(self.conditionals, dict):
+            self.conditionals = {name: self.conditionals for name in init_locs}
+        if not isinstance(self.dependencies, dict):
+            model_dependencies = poutine.block(get_dependencies)(
+                self._original_model[0], args, kwargs,
+            )
+            guide_dependencies = model_dependencies  # FIXME
```
fritzo, line 1338: @eb8680 next task is to replace this with the algorithm in (Webb et al. 2018), which will require additional tracing of observed sites.

fritzo:
@eb8680 ready for a second review.

pyro/infer/inspect.py
```diff
@@ -0,0 +1,191 @@
...
+    model_args: Optional[tuple] = None,
+    model_kwargs: Optional[dict] = None,
+) -> Dict[str, List[str]]:
+    r"""
+    Infers metadata about a conditioned model. Metadata includes
+
+    -   `prior_dependencies` is a dict mapping downstream latent and observed
+        sites to the dictionaries mapping upstream latent sites on which the
+        depend to sets of plates over which dependencies are independent.
+        Dependencies follow the original model order.
```
eb8680, line 58: It's probably worth adding an even simpler example of this data structure inline in these two bullet points and spelling out what it means, if only to save ourselves a bit of hassle if we want to refactor this in 6 months.

e.g.: "...original model order. For example, if `prior_dependencies == {"a": {"a": set()}, "b": {"b": {"i"}, "a": set()}}`, then `a` is a sample site with no ancestors and no plates, while each entry `b_i` of sample site `b` is conditionally independent along plate `i` given a value of `a`."

and correspondingly below, "Posterior dependencies are reversed from the model order. For example, if we condition on sample site `b` in the example above, then `posterior_dependencies == {"a": {"a": set(), "b": set()}}`: `b` no longer appears as a key because it is observed, while global variable `a` now depends on all entries of `b` in plate `i`."

pyro/infer/inspect.py
```diff
@@ -0,0 +1,191 @@
...
+
+    [1] S.Webb, A.Goliński, R.Zinkov, N.Siddharth, T.Rainforth, Y.W.Teh, F.Wood (2018)
+        "Faithful inversion of generative models for effective amortized inference"
+        https://dl.acm.org/doi/10.5555/3327144.3327229
+
+    :param callable model: A model.
+    :param tuple model_args: Optional tuple of model args.
+    :param dict model_kwargs: Optional tuple of model args.
+    :returns: A dictionary of metadata (see above).
+    :rtype: dict
```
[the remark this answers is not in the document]
fritzo: Good suggestion. They are necessary until we start using [sphinx-autodoc-typehints](https://pypi.org/project/sphinx-autodoc-typehints/). I've left a comment in #2550 to track this change.

fritzo:
Another motivating paper:
Manfred Opper, C´edric Archambeau (2009)
"The Variational Gaussian Approximation Revisited"
http://www0.cs.ucl.ac.uk/staff/c.archambeau/publ/neco_mo09_web.pdf

eb8680 (approved):
Looks good!

PR #2826, opened 2021-04-27
Sine Skewed Toridial distribution

OlaRonning:
This PR introduces the Sine Skewed Toridial distribution described by [Jose Ameijeiras-Alonso and Christophe Ley](https://arxiv.org/abs/1910.13293). 

The distribution enables skewing a base distribution on a d-dimensional torus, which is useful, for example, with dihedral angles (1-torus) of peptides as can be seen on a [Ramachandran plot](https://en.wikipedia.org/wiki/Ramachandran_plot#/media/File:Ramachandran_plot_general_100K.jpg).

~~_Missing prior suggestion in docstring; inference method and inferable params as suggested in #2821._~~


pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,80 @@
...
+
+        args_string = ', '.join(['{}: {}'.format(p, self.__dict__[p]
+                                if self.__dict__[p].numel() == 1
+                                else self.__dict__[p].size()) for p in param_names])
+        return self.__class__.__name__ + '(' + f'base_density: {self.base_density.__repr__()}, ' + args_string + ')'
+
+    def sample(self, sample_shape=torch.Size()):
+        bd = self.base_density
+        ys = bd.sample(sample_shape)
+        mask = Uniform(0, 1.).sample(sample_shape) < 1. + (self.skewness * torch.sin((ys - bd.mean) % (2 * pi))).sum(-1)
```
fehiepsi, line 69: I wonder if the uniform samples here should have the shape `sample_shape + batch_shape`?

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,80 @@
...
+    .. note:: ``skewness.abs().sum() <= 1.`` and ``(skewness.abs() <= 1).all()``.
+
+    ** References: **
+      1. Sine-skewed toroidal distributions and their application in protein bioinformatics
+         Ameijeiras-Alonso, J., Ley, C. (2019)
+
+    :param base_density: base density on the d-dimensional torus; event_shape must be [..., 2] where
+        ``prod(event_shape[:-1]) == d``.
+    :param skewness: skewness of the distribution; must have same shape as base_density.event_shape, all values
+        must be in [-1,1] and ``abs(skewness).sum() <= 1``.
```
fehiepsi, line 45: Should `skewness` also have batch_shape? In addition, why we need to support complicated `event_shape` here? Shouldn't it be enough to require `base_density.event_shape == (2,)`?

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
+        bd = self.base_density
+        return bd.log_prob(value) + torch.log(1 + (self.skewness * torch.sin((value - bd.mean) % (2 * pi))).sum(-1))
+
+    @classmethod
+    def infer_shapes(cls, **arg_shapes):
+        return arg_shapes['base_density'], arg_shapes['skewness']
```
fehiepsi, line 80: Not sure if this works but I think you can skip this.

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,80 @@
...
+    support = constraints.independent(constraints.real, 1)
+
+    def __init__(self, base_density: TorchDistribution, skewness, validate_args=None):
+        assert torch.all(skewness.abs() <= 1)
+        assert base_density.event_shape[-1] == 2
+        assert skewness.shape[-1] == 2
+        self.base_density = base_density
+        self.skewness = skewness
+        super().__init__(base_density.batch_shape, base_density.event_shape, validate_args=validate_args)
+
```
fehiepsi, line 57: I think you need to implement `expand` too.

pyro/distributions/__init__.py
```diff
@@ -2,10 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import pyro.distributions.torch_patch  # noqa F403
-from pyro.distributions.torch import *  # noqa F403
-
-# isort: split
-
```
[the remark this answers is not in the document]
OlaRonning: I see, it was moved after I ran `isort` which is why I kept it. I'll revert it.

pyro/distributions/sine_skewed.py
OlaRonning: > Should skewness also have batch_shape? 

Yes, you are right; I fixed this in code but forgot to update the docstring. 

> In addition, why we need to support complicated event_shape here? Shouldn't it be enough to require base_density.event_shape == (2,)?

This density works for any d-dimensional torus, so while the use cases I care/know are on the 1-torus, I figured it would be better to support the general case. What do you think? It's a minor change if you think it's appropriate.

pyro/distributions/sine_skewed.py
OlaRonning: Yes, that may be right. I'll add a test case and fix it as required.

pyro/distributions/sine_skewed.py
OlaRonning: I made a special case in [`test.test_distribution.test_infer_shapes`](https://github.com/pyro-ppl/pyro/blob/be91d9a76e51d210ac03c511ffff5884adddc5bf/tests/distributions/test_distributions.py#L47-L49) to handle this; however, I'll take it out if it's incorrect.

pyro/distributions/sine_skewed.py
OlaRonning: Yes, of course; thanks!

OlaRonning:
@fehiepsi, thanks for the review! :+1:

pyro/distributions/sine_skewed.py
fehiepsi: By d-dimensional torus, do you mean `SineSkewed(base_dist, skewness).to_event(d)` where `base_dist.event_shape == (2,)`?

pyro/distributions/__init__.py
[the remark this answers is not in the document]
fritzo: can you make sure you have `isort>=5.0`?

pyro/distributions/sine_skewed.py
fritzo: I believe `infer_shapes` currently works only with tensor arguments to distributions.  No need to add this method in your PR.

tests/distributions/test_distributions.py
```diff
@@ -44,8 +44,12 @@ def test_infer_shapes(dist):
...
     for idx in range(dist.get_num_test_data()):
         dist_params = dist.get_dist_params(idx)
-        arg_shapes = {k: v.shape if isinstance(v, torch.Tensor) else ()
-                      for k, v in dist_params.items()}
+        if 'SineSkewed' == dist.pyro_dist.__name__:
+            arg_shapes = {k: v.shape if isinstance(v, torch.Tensor) else v.batch_shape
+                          for k, v in dist_params.items()}
+        else:
+            arg_shapes = {k: v.shape if isinstance(v, torch.Tensor) else ()
+                          for k, v in dist_params.items()}
```
fritzo, line 52: It's probably best to revert this for now. At some future point we may add support for distribution arguments, but then we will do so more uniformly.

pyro/distributions/__init__.py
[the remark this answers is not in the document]
OlaRonning: I'm running version `isort==5.8`. Reverted manually and ran `isort`, and this time `# isort: split` stayed the same.

pyro/distributions/sine_skewed.py
OlaRonning: I see. Thanks for clarifying. I'll add more test cases and resolve this.

tests/distributions/test_distributions.py
OlaRonning: That makes sense; I'll revert this.

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,101 @@
...
+    .. note:: The base-distribution must be over a arbitrary dim torus.
+
+    .. note:: ``skewness.abs().sum() <= 1.`` and ``(skewness.abs() <= 1).all()``.
+
+    ** References: **
+      1. Sine-skewed toroidal distributions and their application in protein bioinformatics
+         Ameijeiras-Alonso, J., Ley, C. (2019)
+
+    :param base_density: base density on the d-dimensional torus; event_shape must be [..., 2] where
```
fehiepsi, line 42: For consistency with other distributions, pls use base_dist rather than base_density.

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,101 @@
...
+        ``prod(event_shape[:-1]) == d``.
+    :param skewness: skewness of the distribution; must have same shape as base_density.event_shape, all values
+        must be in [-1,1] and ``abs(skewness).sum() <= 1``.
```
fehiepsi, line 45: The last requirement is not compatible with pyro tensor senmatics. The condition should not involve batch dimensions.

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,101 @@
...
+    .. note:: The base-distribution must be over a arbitrary dim torus.
+
+    .. note:: ``skewness.abs().sum() <= 1.`` and ``(skewness.abs() <= 1).all()``.
+
+    ** References: **
+      1. Sine-skewed toroidal distributions and their application in protein bioinformatics
+         Ameijeiras-Alonso, J., Ley, C. (2019)
+
+    :param base_density: base density on the d-dimensional torus; event_shape must be [..., 2] where
+        ``prod(event_shape[:-1]) == d``.
```
fehiepsi, line 43: Is this distribution different from SineSkewed(...).to_event(...)?

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,101 @@
...
+        assert base_density.event_shape[-1] == 2
+        assert skewness.shape[-1] == 2
+
+        if base_density.mean.device != skewness.device:
+            raise ValueError(f"base_density: {base_density.__class__.__name__} and {self.__class__.__name__} "
+                             f"must be on same device.")
+
+        self.base_density = base_density
+        self.skewness = skewness
+        super().__init__(base_density.batch_shape, base_density.event_shape, validate_args=validate_args)
```
fehiepsi, line 61: batch_shape should be broadcastable w.r.t. skewness's batch_shape.

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,101 @@
...
+    :param skewness: skewness of the distribution; must have same shape as base_density.event_shape, all values
+        must be in [-1,1] and ``abs(skewness).sum() <= 1``.
+    """
+    arg_constraints = {'skewness': constraints.interval(-1., 1.)}
+    support = constraints.independent(constraints.real, 1)
+
+    def __init__(self, base_density: TorchDistribution, skewness, validate_args=None):
+        assert torch.all(skewness.abs() <= 1)
+        assert base_density.event_shape[-1] == 2
+        assert skewness.shape[-1] == 2
```
fehiepsi, line 53: I think you need to check for compatibility of skewness shapes with base_density shapes. I still don't know what's the requirement.

fehiepsi (commented):
Could you clarify the input shapes for base density and skewness? It is not clear to me what's the requirements so I couldn't check if the expand method and the reshape ops in sample and log_prob are valid.

pyro/distributions/sine_skewed.py
OlaRonning: Yes, sorry, you are right; the docstring for `skewness` is wrong. `skewness.shape == base_dist.shape()` should be the assertion. I will update both.

pyro/distributions/sine_skewed.py
OlaRonning: No, those are the same. Adding `prod(event_shape[:-1]) == d` was intended to clarify what is meant by d dimensional torus. I'll take it out as it seems to cause more confusion than good.

pyro/distributions/sine_skewed.py
OlaRonning: Hmm, I'm not sure I follow; the last condition (`abs(skewness).sum() <= 1`) should be changed to
``` python
assert torch.all(skewness.view((torch.prod(torch.tensor(base_dist.batch_shape), dim=0), -1)).sum(-1) <= 1.)
```
Is this what you mean?

OlaRonning:
Hi @fehiepsi,

Thanks for the review! Sorry for the input dimensions being vague. Let me clarify; the requirements are `base_dist.shape() == skewness.shape` and `base_dist.event_shape[-1] == 2'. So, 
``` python
ss = SineSkewed(base_dist=BaseDist(Size([3,3,2])).to_event(1), skewness=Size([3,3,2]))
assert ss.event_shape == (2)
assert ss.batch_shape == (3,3)

ss1 = SineSkewed(base_dist=BaseDist(Size([3,3,2])).to_event(1), skewness=Size([3,3,2])).to_event(1)
assert ss1.event_shape == (3,2)
assert ss1.batch_shape == (3,)

ss2 = SineSkewed(base_dist=BaseDist(Size([3,3,2])).to_event(2), skewness=Size([3,3,2])).to_event(1)
assert ss2.event_shape == (3,3,2)
assert ss2.batch_shape == tuple()

ss3 = SineSkewed(base_dist=BaseDist(Size([3,3,2])).to_event(1), skewness=Size([3,3,2])).to_event(2)
assert ss3.event_shape == (3,3,2)
assert ss3.batch_shape == tuple()
```
I would like for the construction `ss=SineSkewed(base_dist=BaseDist(Size([3,3,2])), skewness=Size([3,3,2])).to_event(1)` also to work; however, I cannot figure out how to ensure the to_event has been called on the `ss` object before it's used.

pyro/distributions/sine_skewed.py
fehiepsi: It seems like that (though it is a bit tricky to follow those shape manipulations). To be more clarified, I guess you'll need to add some assertation to the `__init__` method:
+ assert base_dist.batch_shape broadcastable with skewness batch_shape
+ assert base_dist.event_shape is valid
+ assert skewness event_shape is valid and compatible with base_dist.event_shape
+ make sure that `arg_constraints["skewness"]` has a correct event dimension. If the event dimension is not known, you can use `arg_constraints = {'skewness': constraints.dependent}`
+ assert skewness satisfies the condition (abs(skewness).sum(over event dimension?) <= 1) for all items in batch dimensions
+ if event_dim is not decided until the construction, you can use
```
    @constraints.dependent_property
    def support(self):
        return constraints.independent(constraints.real, self.event_dim)
```
+ mention the support of base_dist, is there any restriction?...
+ ...
+ add some comments to point to the corresponding formula in the reference

I guess the main thing that makes the implementation complicated is: this class supports arbitrary event_dim base distribution. I'm not sure why that's the case. If you have some applications that require arbitrary event_dim, then adding the above assertations might be enough. Otherwise, I would suggest to stick with a specific `event_dim=1` or `event_dim=2`,... My first impression is you need `event_dim=2` where `event_shape = (d, 2)` but I could be wrong...

OlaRonning: Thanks for taking the time to help me; I'll make a [separate branch](https://github.com/aleatory-science/pyro/tree/feature/mulit_dim_sine_skewed) for the d-torus (I may have an application modeling peptide side-chains) and update this branch for the fixed `event_dim=2` as you suggest. If the application works out, I'll make open a new PR with the arbitrary dimension version.

OlaRonning:
As suggested by @fehiepsi, I've restricted `event_dim<=2`.

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,92 @@
...
+         Ameijeiras-Alonso, J., Ley, C. (2019)
+
+    :param base_density: base density on a d-dimensional torus.
+    :param skewness: skewness of the distribution.
+    """
+    arg_constraints = {'skewness': constraints.independent(constraints.interval(-1., 1.), 1)}
+
+    support = constraints.independent(constraints.real, 1)
+
+    def __init__(self, base_density: TorchDistribution, skewness, validate_args=None):
```
fehiepsi, line 43: `base_density` -> `base_dist`

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,92 @@
...
+        assert base_density.event_shape[-1] == 2 and len(base_density.event_shape) <= 2
+        assert base_density.shape()[len(base_density.shape()) - len(skewness.shape):] == skewness.shape
+        assert (skewness.abs().sum(-1 if len(skewness.shape) == 1 else (-2, -1)) <= 1).all()
+
+        self.base_density = base_density
+        self.skewness = skewness.broadcast_to(base_density.shape())
```
fehiepsi, line 49: To make sure that batch shape agrees, you can add this logic
```
event_shape = base_dist.event_shape
event_dim = base_dist.event_dim
assert skewness.dim() >= event_dim
assert skewness.shape[-event_dim:] == event_shape
batch_shape = broadcast_shapes(base_dist.batch_shape, skewness.shape[:-event_dim]
self.skewness = skewness.broadcast_to(batch_shape + event_shape)
self.base_dist = base_dist.expand(batch_shape)
```

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,92 @@
...
+        must be less than or equal to one. See eq. 2.1 in [1].
+
+    ** References: **
+      1. Sine-skewed toroidal distributions and their application in protein bioinformatics
+         Ameijeiras-Alonso, J., Ley, C. (2019)
+
+    :param base_density: base density on a d-dimensional torus.
+    :param skewness: skewness of the distribution.
+    """
+    arg_constraints = {'skewness': constraints.independent(constraints.interval(-1., 1.), 1)}
```
fehiepsi, line 39: Unless you split this distribution into `SineSkewedCircular` and `SineSkewedTorus` for arbitrary event_dim, the constraint here is `constraints.dependent`. You can add some assertation to the constructor.

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,92 @@
...
+
+    support = constraints.independent(constraints.real, 1)
+
+    def __init__(self, base_density: TorchDistribution, skewness, validate_args=None):
+        assert base_density.event_shape[-1] == 2 and len(base_density.event_shape) <= 2
+        assert base_density.shape()[len(base_density.shape()) - len(skewness.shape):] == skewness.shape
+        assert (skewness.abs().sum(-1 if len(skewness.shape) == 1 else (-2, -1)) <= 1).all()
```
fehiepsi, line 46: Pls add some meaningful error message.

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,92 @@
...
+
+    :param base_density: base density on a d-dimensional torus.
+    :param skewness: skewness of the distribution.
+    """
+    arg_constraints = {'skewness': constraints.independent(constraints.interval(-1., 1.), 1)}
+
+    support = constraints.independent(constraints.real, 1)
+
+    def __init__(self, base_density: TorchDistribution, skewness, validate_args=None):
+        assert base_density.event_shape[-1] == 2 and len(base_density.event_shape) <= 2
```
fehiepsi, line 44: You can either use the arbitrary event_dim or split this into 2 classes: 1 for event_dim=1, and 1 for event_dim=2.

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,92 @@
...
+        assert base_density.shape()[len(base_density.shape()) - len(skewness.shape):] == skewness.shape
```
fehiepsi, line 45: Remove this assertation (see the broadcasting logic in a comment below). This assertation does not support broadcasting. For example, this will fail
```
base_dist.shape = batch_shape + event_shape = () + (100, 2)
skewness.shape = (1000, 100, 1000, 2)
```

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,92 @@
...
+
+        def model(...):
+            ...
+            skewness_phi = pyro.sample(f'skewness_phi', Uniform(skewness.abs().sum(), 1 - tots))
+            psi_bound = 1 - skewness_phi.abs()
+            skewness_psi = pyro.sample(f'skewness_psi', Uniform(-psi_bound, psi_bound)
+            skewness = torch.stack((skewness_phi, skewness_psi), dim=0)
+            ...
+
+    .. note:: An event in the base distribution must be on a d-torus, so the event_shape must be (d, 2) or (2,).
```
fehiepsi, line 27: Please add a note for why the last dimention is 2.

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,92 @@
...
+        return samples
+
+    def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
+        bd = self.base_density
+        bd_prob = bd.log_prob(value)
+        sine_prob = torch.log(
+            1 + (self.skewness * torch.sin((value - bd.mean) % (2 * pi))).reshape((-1, self.event_shape.numel())).sum(
+                -1))
```
fehiepsi, line 80: pls add reference for this formula

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def sample(self, sample_shape=torch.Size()):
+        bd = self.base_density
+        ys = bd.sample(sample_shape)
+        u = Uniform(0., torch.ones(torch.Size([]), device=self.skewness.device)).sample(sample_shape + self.batch_shape)
+
+        mask = u < 1. + (self.skewness * torch.sin((ys - bd.mean) % (2 * pi))).view(*(sample_shape + self.batch_shape),
+                                                                                    -1).sum(-1)
+        mask = mask.view(*sample_shape, *self.batch_shape, *(1 for _ in bd.event_shape))
+        samples = (torch.where(mask, ys, -ys + 2 * bd.mean) + pi) % (2 * pi) - pi
```
fehiepsi, line 70: pls add reference for this formula

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,92 @@
...
+        sine_prob = torch.log(
+            1 + (self.skewness * torch.sin((value - bd.mean) % (2 * pi))).reshape((-1, self.event_shape.numel())).sum(
+                -1))
+        return (bd_prob.view((-1)) + sine_prob).view(bd_prob.shape)
+
+    def expand(self, batch_shape, _instance=None):
+        batch_shape = torch.Size(batch_shape)
+        new = self._get_checked_instance(SineSkewed, _instance)
+        base_dist = self.base_density.expand(batch_shape, None)
+        new.base_density = base_dist
```
fehiepsi, line 87: what happens with new.skewness?

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,92 @@
...
+                                else getattr(self, p).size()) for p in self.arg_constraints.keys()])
+        return self.__class__.__name__ + '(' + f'base_density: {str(self.base_density)}, ' + args_string + ')'
+
+    def sample(self, sample_shape=torch.Size()):
+        bd = self.base_density
+        ys = bd.sample(sample_shape)
+        u = Uniform(0., torch.ones(torch.Size([]), device=self.skewness.device)).sample(sample_shape + self.batch_shape)
+
+        mask = u < 1. + (self.skewness * torch.sin((ys - bd.mean) % (2 * pi))).view(*(sample_shape + self.batch_shape),
+                                                                                    -1).sum(-1)
```
fehiepsi, line 68: I think you can use `pyro.distributions.util.sum_rightmost`
```
mask = u < 1. + sum_rightmost(self.skewness * torch.sin((ys - bd.mean) % (2 * pi)), self.event_dim)
```

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
+        bd = self.base_density
+        bd_prob = bd.log_prob(value)
+        sine_prob = torch.log(
+            1 + (self.skewness * torch.sin((value - bd.mean) % (2 * pi))).reshape((-1, self.event_shape.numel())).sum(
+                -1))
+        return (bd_prob.view((-1)) + sine_prob).view(bd_prob.shape)
```
fehiepsi, line 81: ditto: use `sum_rightmost`
```python
sine_prob = torch.log1p(sum_rightmost(self.skewness * torch.sin((value - bd.mean) % (2 * pi)), self.event_dim))
return bd_prob + sine_prob
```

tests/distributions/conftest.py
```diff
@@ -329,6 +330,11 @@ def __init__(self, rate, *, validate_args=None):
                 {'loc': [2.0, 50.0], 'scale': [4.0, 100.0],
                  'test_data': [[2.0, 50.0], [2.0, 50.0]]},
                 ]),
+    Fixture(pyro_dist=dist.SineSkewed,
+            examples=[{
+                'base_density': dist.Uniform(*tensor_wrap([-pi, -pi], [pi, pi])).to_event(1),
```
fehiepsi, line 335: What's the condition of base_dist? Does it need to be symmetric over a torus?...

tests/distributions/test_sine_skewed.py
```diff
@@ -0,0 +1,104 @@
...
+
+    loc = Normal(0., 1.).sample(base_dist.event_shape) % (2 * pi) - pi
+
+    base_prob = base_dist.log_prob(loc)
+    skewness = _skewness(base_dist.event_shape)
+
+    ss = SineSkewed(base_dist, skewness)
+    assert_equal(base_prob + torch.log(
+        1 + (skewness * torch.sin(loc - base_dist.mean)).view(*base_dist.batch_shape, -1).sum(-1)),
+                 ss.log_prob(loc))
```
fehiepsi, line 56: This is just a repeatation of the implementation. Could you come up with 2 formula for 2 `base_dist` (other than Uniform)?

fehiepsi (commented):
Thanks, @OlaRonning!

pyro/distributions/sine_skewed.py
OlaRonning: I'll split it into two distributions; I need to check that the distribution is proper when over an n-sphere. As you've already pointed out, this will make the implementation much more straightforward.

pyro/distributions/sine_skewed.py
OlaRonning: Thanks for the code snippet; I was not sure how to do this properly, so you've saved me quite a bit of time!

pyro/distributions/sine_skewed.py
OlaRonning: I'll add error messages to all of these.

pyro/distributions/sine_skewed.py
OlaRonning: As mentioned above, I think the simplest is to have two classes.

pyro/distributions/sine_skewed.py
OlaRonning: I see; I'll take it out.

pyro/distributions/sine_skewed.py
OlaRonning: These are the two angles that specify the position on the torus; I will update the docstring.

pyro/distributions/sine_skewed.py
OlaRonning: Cool, thanks. I'll try it out.

tests/distributions/conftest.py
OlaRonning: It should be symmetric, though it is not strictly necessary. I will also work over a circle; however, I've not checked if the distribution is proper a sphere.

tests/distributions/test_sine_skewed.py
OlaRonning: Yes, I'll update this. I can use the von Mises and BvM once  #2821 is merged.

OlaRonning:
Hi @fehiepsi,

I found a (fundamental and grave) error in my math when checking the n-sphere. A d-torus is over (-pi, pi]^d, so the `event_shape` is (d,) not (d,2) as I erroneously put. The good thing is this dramatically simplifies the implementation as this means `n_dim=1`. I'll make the necessary changes to the implementation and request another review from you once finished, if you don't mind.

fehiepsi:
Yeah, I can see it really simplifies the implementation and clarifies confusions. Please take your time.

OlaRonning:
Hi @fehiepsi, when you have time, will you help me with the error in CI for docs (commit 5cfee34)? Otherwise, I believe I've corrected the implementation now.

pyro/distributions/__init__.py
```diff
@@ -58,9 +58,11 @@
     RelaxedBernoulliStraightThrough,
     RelaxedOneHotCategoricalStraightThrough,
 )
+from pyro.distributions.sine_skewed import SineSkewed
 from pyro.distributions.softlaplace import SoftLaplace
 from pyro.distributions.spanning_tree import SpanningTree
 from pyro.distributions.stable import Stable
+from pyro.distributions.torch import *  # noqa F403
```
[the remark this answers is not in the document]
OlaRonning: I see; I'll take it out.

tests/distributions/test_cuda.py
```diff
@@ -15,6 +15,8 @@
 
 @requires_cuda
 def test_sample(dist):
+    if dist.pyro_dist.__name__ == 'SineSkewed':
+        pytest.xfail(reason="Fixture with distribution param not handled.")
```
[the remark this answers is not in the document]
fehiepsi: I see. I think you just need to make wrappers `_VonMisesSineSkewed` and `_UniformSineSkewed` to fixture and test them.

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,96 @@
...
+    .. note:: An event in the base distribution must be on a d-torus, so the event_shape must be (d,).
+
+    .. note:: For the skewness parameter, it must hold that the sum of the absolute value of its weights for an event
+        must be less than or equal to one. See eq. 2.1 in [1].
+
+    ** References: **
+      1. Sine-skewed toroidal distributions and their application in protein bioinformatics
+         Ameijeiras-Alonso, J., Ley, C. (2019)
+
+    :param base_dist: base density on a d-dimensional torus.
```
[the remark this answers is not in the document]
OlaRonning: Yes, thanks for the suggestion.

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,96 @@
...
+    the d-dimensional torus defined as ⨂^d S^1 where S^1 is the circle. So for example the 0-torus is a point, the
+    1-torus is a circle and the 2-tours is commonly associated with the donut shape (some may object to this simile).
+
+    The skewness parameter can be inferred using :class:`~pyro.infer.HMC` or :class:`~pyro.infer.NUTS`.
+    For example, the following will produce a uniform prior over skewness for the 2-torus,::
+
+        def model(...):
+            ...
+            skew_phi = pyro.sample(f'skew_phi', Uniform(-1., 1.))
+            psi_bound = 1 - skewness_phi.abs()
```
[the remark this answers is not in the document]
fritzo: Could you also provide a valid example demonstrating how to construct a `SineSkewed` distribution?

tests/distributions/test_cuda.py
```diff
@@ -77,6 +79,8 @@ def test_rsample(dist):
 
 @requires_cuda
 def test_log_prob(dist):
+    if dist.pyro_dist.__name__ == 'SineSkewed':
+        pytest.xfail(reason="Fixture with distribution param not handled.")
```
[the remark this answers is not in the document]
fritzo: Often for testing I'll create a test-only distribution whose distribution inputs are hard-coded and whose remaining inputs are all tensors, and add that distribution to the unit tests, e.g.
```py
# in tests/distributions/conftest.py
class SineSkewedUniform(SineSkewed):
    def __init__(self, skeweness, validate_args=None):
        base_dist = Uniform(torch.zeros(2), 1).to_event(1)
        super().__init__(base_dist, skewness, validate_args=validate_args)

    # ...you might also need a simple .expand() method...
```

pyro/distributions/sine_skewed.py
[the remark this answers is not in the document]
OlaRonning: Yes, I'll add an example with the `SineBivariateVonMises` distribution as base distribution.

tests/distributions/test_cuda.py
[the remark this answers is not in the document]
OlaRonning: Thanks for the example @fritzo!

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,136 @@
...
+            skew_psi = pyro.sample('skew_psi', Uniform(-1., 1.))
+            skewness = torch.stack((skew_phi, psi_bound * skew_psi), dim=-1)
+            assert skewness.shape == (num_mix_comp, 2)
+
+            with pyro.plate('obs_plate'):
+                sine = SineBivariateVonMises(phi_loc=phi_loc, psi_loc=psi_loc,
+                                            phi_concentration=1000 * phi_conc,
+                                            psi_concentration=1000 * psi_conc,
+                                            weighted_correlation=corr_scale)
+                        return pyro.sample('phi_psi', SineSkewed(sine, skewness), obs=obs)
```
fritzo, line 47: nit: is this indentation correct? it looks off on github

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,136 @@
...
+    def __repr__(self):
+        args_string = ', '.join(['{}: {}'.format(p, getattr(self, p)
+                                if getattr(self, p).numel() == 1
+                                else getattr(self, p).size()) for p in self.arg_constraints.keys()])
+        return self.__class__.__name__ + '(' + f'base_density: {str(self.base_dist)}, ' + args_string + ')'
+
+    def sample(self, sample_shape=torch.Size()):
+        bd = self.base_dist
+        ys = bd.sample(sample_shape)
+        u = Uniform(0., torch.ones(torch.Size([]), device=self.skewness.device)).sample(sample_shape + self.batch_shape)
```
fritzo, line 112: nit: `self.skewness.new_ones(())` would also get the correct dtype (float vs long)

fritzo (approved):
LGTM after a couple tiny nits

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,136 @@
...
+            phi_conc = pyro.sample('phi_conc', Beta(halpha_phi, beta_prec_phi - halpha_phi))
+            psi_conc = pyro.sample('psi_conc', Beta(halpha_psi, beta_prec_psi - halpha_psi))
+            corr_scale = pyro.sample('corr_scale', Beta(2., 5.))
+
+            # SS prior
+            skew_phi = pyro.sample('skew_phi', Uniform(-1., 1.))
+            psi_bound = 1 - skew_phi.abs()
+            skew_psi = pyro.sample('skew_psi', Uniform(-1., 1.))
+            skewness = torch.stack((skew_phi, psi_bound * skew_psi), dim=-1)
+            assert skewness.shape == (num_mix_comp, 2)
```
fehiepsi, line 40: It would be nice to have a constraint+transform for this (in a follow-up PR). I believe we can use [signed stick-breaking transform](https://github.com/pytorch/pytorch/blob/e66015dadf135f6e1cda69d80883505807b1bbed/torch/distributions/transforms.py#L760) here. This way users can define distributions over `skewness` (or just simply use `pyro.param` with correct constraint). Without that, it is a bit cumbersome for users to define correct `skewness` over general d-torus.

fehiepsi (approved):

pyro/distributions/sine_skewed.py
fehiepsi: With that, we can have correct `constraints` in the distribution definition. :D

pyro/distributions/sine_skewed.py
OlaRonning: I'll check that it looks right.

pyro/distributions/sine_skewed.py
OlaRonning: makes sense; I'll update it.

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,136 @@
...
+
+            with pyro.plate('obs_plate'):
+                sine = SineBivariateVonMises(phi_loc=phi_loc, psi_loc=psi_loc,
+                                            phi_concentration=1000 * phi_conc,
```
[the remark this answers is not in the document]
OlaRonning: Nice catch; thanks.

pyro/distributions/sine_skewed.py
OlaRonning: Ops, yes, are quite right.

pyro/distributions/sine_skewed.py
OlaRonning: That would be neat; I'll add it to the backlog.

fehiepsi (approved):

pyro/distributions/sine_skewed.py
```diff
@@ -0,0 +1,136 @@
...
+            skew_psi = pyro.sample('skew_psi', Uniform(-1., 1.))
+            skewness = torch.stack((skew_phi, psi_bound * skew_psi), dim=-1)
+            assert skewness.shape == (num_mix_comp, 2)
+
+            with pyro.plate('obs_plate'):
+                sine = SineBivariateVonMises(phi_loc=phi_loc, psi_loc=psi_loc,
+                                             phi_concentration=1000 * phi_conc,
+                                             psi_concentration=1000 * psi_conc,
+                                             weighted_correlation=corr_scale)
+                return pyro.sample(' phi_psi', SineSkewed(sine, skewness), obs=obs)
```
fehiepsi, line 47: nit: ' phi_psi' -> 'phi_psi'

fehiepsi (approved):

PR #2829, opened 2021-04-29
Fix AutoStructured guide

fritzo:
Addresses #2813 
Fixes the bug discussed in https://github.com/pyro-ppl/pyro/pull/2812#discussion_r622243256

This fixes a shape bug in `AutoStructured` that resulted in an extra factor of `batch_shape.numel()` being multiplied to the `scale` and `scale_tril` transformation `log_abs_det_jacobian`s, which then resulted in incorrect inference.

Note I've avoided using `pyro.factor` because it results in also-incorrect inference due to an implementation subtlety: `dist.Unit` is not reparametrized.  As discussed with @martinjankowiak, `pyro.factor` is dangerous when used in guides, so I've added a warning.

## Tested
- [x] regression tests for scalar models and `normal` and `mvn` guides
- [x] regression tests for batched models `normal` and `mvn` guides
- [x] strengthened shape tests
- [x] verified accuracy on a real-world model

### DIFF
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -1433,6 +1433,7 @@ class AutoStructured(AutoGuide):
             transform = biject_to(site["fn"].support)
             value = transform(unconstrained)
             if compute_density and conditional != "delta":
+                assert transform.codomain.event_dim == site["fn"].event_dim
                 log_density = log_density + transform.inv.log_abs_det_jacobian(
                     value, unconstrained
                 )
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -1420,12 +1420,12 @@ class AutoStructured(AutoGuide):
             aux_values[name] = aux_value
 
             # Shift by loc and reshape.
-            unconstrained = aux_value + loc
             batch_shape = torch.broadcast_shapes(
-                unconstrained.shape[:-1], self._batch_shapes[name]
+                aux_value.shape[:-1], self._batch_shapes[name]
             )
-            shape = batch_shape + self._unconstrained_event_shapes[name]
-            unconstrained = unconstrained.reshape(shape)
+            unconstrained = (aux_value + loc).reshape(
+                batch_shape + self._unconstrained_event_shapes[name]
+            )
             if not is_identically_zero(log_density):
                 log_density = log_density.reshape(batch_shape + (-1,)).sum(-1)
 
@@ -1463,9 +1463,7 @@ class AutoStructured(AutoGuide):
         result = {}
         for name, site in self._sorted_sites:
             loc = _deep_getattr(self.locs, name).detach()
-            shape = self._unconstrained_shapes[name]
-            if loc.shape != shape:
-                sample_shape = loc.shape[:-1]
-                loc = loc.reshape(sample_shape + shape)
+            shape = self._batch_shapes[name] + self._unconstrained_event_shapes[name]
+            loc = loc.reshape(shape)
             result[name] = biject_to(site["fn"].support)(loc)
         return result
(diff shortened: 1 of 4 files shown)


pyro/primitives.py
```diff
@@ -162,6 +162,13 @@ def factor(name, log_factor):
     Factor statement to add arbitrary log probability factor to a
     probabilisitic model.
 
+    .. warning:: Beware using factor statements in guides. Factor statements
+        assume ``log_factor`` is computed from non-reparametrized statements
+        such as observation statements ``pyro.sample(..., obs=...)``. If
+        instead ``log_factor`` is computed from e.g. the Jacobian determinant
+        of a transformation of a reparametrized variable, factor statements
+        in the guide will result in incorrect results.
```
martinjankowiak, line 170: "result in unnecessarily high variance gradients in the context of SVI"?

pyro/infer/autoguide/guides.py
```diff
@@ -1419,20 +1419,25 @@ def get_deltas(self, save_params=None):
...
-            # Shift by loc, reshape, and transform to constrained space.
-            unconstrained = aux_value + loc
-            shape = self._unconstrained_shapes[name]
-            if torch._C._get_tracing_state() or unconstrained.shape != shape:
-                sample_shape = unconstrained.shape[:-1]
-                unconstrained = unconstrained.reshape(sample_shape + shape)
+            # Shift by loc and reshape.
+            batch_shape = torch.broadcast_shapes(
+                aux_value.shape[:-1], self._batch_shapes[name]
+            )
```
fehiepsi, line 1425: It is not clear to me the relationship of `aux_value.shape[:-1]` and `fn.batch_shape`. Could you elaborate on this a bit? I feel that you need `batch_shape = aux_value.shape[:-1] + self._batch_shapes[name]` but I'm not sure.

pyro/primitives.py
fritzo: Gradient estimates would be not merely high variance but incorrect.  To make gradients correct users would need to additionally use non-reparametrized versions of distributions, triggering REINFORCE for that site (as we discussed via `fakes.NonreparametrizedNormal`).  Because this contract-at-a-distance is difficult to verify, I think we should recommend assuming gradients are incorrect.

pyro/infer/autoguide/guides.py
```diff
@@ -1419,20 +1419,25 @@ def get_deltas(self, save_params=None):
...
+                log_density = log_density.reshape(batch_shape + (-1,)).sum(-1)
+
+            # Transform to constrained space.
             transform = biject_to(site["fn"].support)
             value = transform(unconstrained)
-
-            # Create a Delta distribution.
             if compute_density and conditional != "delta":
-                ldj = transform.inv.log_abs_det_jacobian(value, unconstrained)
-                ldj = sum_rightmost(ldj, ldj.dim() - value.dim() + site["fn"].event_dim)
```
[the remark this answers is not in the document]
fritzo: This is no longer necessary now that we have `IndependentTransform` and that all distributions satisfy `fn.event_dim == fn.support.event_dim`.  In fact the code earlier in this file from which this code is copied is obsolete, but I have avoided cluttering this PR with unrelated changes.

To be safe I've added an assertion in case a user has created a custom distribution with invalid shape metadata.  I've also added a shape test with `LKJCholesky` which would have failed before our work on `IndependentTransform`.

pyro/infer/autoguide/guides.py
fritzo: - `self._batch_shapes[name]` is the batch shape of the site under the original model with original plates.
- `aux_value` has `event_dim = 1` so its batch shape is `aux_value.shape[:-1]`. This is the shape of the site with all original plates stripped away (converted to event dims) but possibly other plates added, e.g. `vectorize_particles`.

This computation combines the original plates with the new inference-specific plates.  Note we need to broadcast rather than concatenate because both batch shapes are aligned at the batch|event boundary.

fehiepsi (approved):
LGTM, thanks for fixing this issue and adding more coverage tests.

fritzo:
Thanks for reviewing this subtle code!

PR #2838, opened 2021-05-11
Fix docs issue for sphinx-4

fehiepsi:
Addresses a doc issue that is blocking current PRs in #2826.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -205,7 +205,7 @@ def skip(app, what, name, obj, skip, options):
 
 
 def setup(app):
-    app.add_stylesheet('css/pyro.css')
+    app.add_css_file('css/pyro.css')
 #     app.connect("autodoc-skip-member", skip)
 
 


docs/source/conf.py
```diff
@@ -205,7 +205,7 @@ def skip(app, what, name, obj, skip, options):
 
 
 def setup(app):
-    app.add_stylesheet('css/pyro.css')
+    app.add_css_file('css/pyro.css')
```
fehiepsi, line 208: `add_stylesheet` is renamed to `add_css_file` since version [1.8](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_css_file)

fehiepsi: I'm not sure. Probably you got some sort of circular import issues (e.g. `from ... import *` statement or type annotation...)

OlaRonning (approved):
LGTM! Thanks for a quick fix.

fritzo (approved):
Let's check readthedocs after this merges, to ensure `latest` is still built.

PR #2841, opened 2021-05-13
Remove overly-narrow validation check from AutoStructured

fritzo:
This PR allows a delta to depend on an upstream delta, which is reasonable if both deltas depend on other upstream latent variables.


fehiepsi (approved):
Makes sense to me.

fritzo:
Thanks @fehiepsi !

PR #2842, opened 2021-05-14
Stabilize DirichletMultinomial.log_prob() for large values

fritzo:
## Tested

- [x] added a unit test for `_log_beta_1()`
- [ ] tested on a real world example


fritzo:
Closing since this does not seem to solve my issue.

PR #2846, opened 2021-05-16
Fix pyro.poutine.do to avoid duplicate entries in cond_indep_stack

### LINKED ISSUE #2845: [bug] Runtime error during SVI inference when using poutine.do()
### Issue Description

Setting: a simple model with 2 latent Gaussians z1 and z2, giving rise to x ~ N( z1+z2, I).

In this setting p(z2 | x, z1) should be the same as p(z2 | x, do(z1)). 

I wanted to check whether the current Pyro interface reflects this and it seems it does not.

My initial thought is that there is a difference in how .do() and .condition() broadcast the constants across the plate context.

### Environment

 - OS and python version: MacOS 10.14.6, Python: 3.8.6
 - PyTorch version: 1.9.0.dev20210502 (nightly version)
 - Pyro version: 1.6.0.

### Code Snippet

Replication code:
https://pastebin.com/Ki2PYX7z

eb8680:
Resolves #2845 

This PR adds a one-line fix to `DoMessenger` to empty the `cond_indep_stack` of a duplicated sample site before the handler stack is reapplied. Without this fix, plates may appear twice in the new site's `cond_indep_stack`, which is incorrect in general and specifically breaks autoguide setup.

Tested:
- added a regression test adapted from @h1psta's example in #2845 that would have caught this

### DIFF
--- a/pyro/poutine/do_messenger.py
+++ b/pyro/poutine/do_messenger.py
@@ -66,6 +66,7 @@ class DoMessenger(Messenger):
 
             # split node, avoid reapplying self recursively to new node
             new_msg = msg.copy()
+            new_msg["cond_indep_stack"] = ()  # avoid entering plates twice
             apply_stack(new_msg)
 
             # apply intervention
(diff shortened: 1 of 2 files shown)


fehiepsi (approved):

fehiepsi (approved):
LGTM!

PR #2850, opened 2021-05-18
Implement AsymmetricLaplace distribution

fritzo:
## Tested
- [x] add to tests/distributions/conftest.py

### DIFF
--- a/pyro/distributions/asymmetriclaplace.py
+++ b/pyro/distributions/asymmetriclaplace.py
@@ -51,10 +51,23 @@ class AsymmetricLaplace(TorchDistribution):
 
     def rsample(self, sample_shape=torch.Size()):
         shape = self._extended_shape(sample_shape)
-        denom = self.left_scale + self.right_scale
-        shift = -self.left_scale / denom
-        u = self.loc.new_empty(shape).uniform_() - shift
+        total_scale = self.left_scale + self.right_scale
+        u = self.loc.new_empty(shape).uniform_() - self.left_scale / total_scale
         scale = torch.where(u < 0, -self.left_scale, self.right_scale)
-        u = u.abs().mul(denom / scale)
-        u = u.data.clamp(max=1 - torch.finfo(u.dtype).eps)
-        return u.neg().log1p().mul(scale)
+        u = u.mul(total_scale / scale)
+        u = u.clamp(max=1 - torch.finfo(u.dtype).eps)
+        return self.loc - u.neg().log1p().mul(scale)
+
+    @property
+    def mean(self):
+        total_scale = self.left_scale + self.right_scale
+        return self.loc + (self.right_scale ** 2 - self.left_scale ** 2) / total_scale
+
+    @property
+    def variance(self):
+        left = self.left_scale
+        right = self.right_scale
+        total = left + right
+        p = left / total
+        q = right / total
+        return p * left ** 2 + q * right ** 2 + p * q * total ** 2
--- a/pyro/distributions/asymmetriclaplace.py
+++ b/pyro/distributions/asymmetriclaplace.py
@@ -12,8 +12,8 @@ class AsymmetricLaplace(TorchDistribution):
     """
     Asymmetric version of Laplace distribution.
 
-    To the left of ``loc`` this acts like an ``-Exponential(left_scale)``;
-    to the right of ``loc`` this acts like an ``Exponential(right_scale)``.
+    To the left of ``loc`` this acts like an ``-Exponential(1 / left_scale)``;
+    to the right of ``loc`` this acts like an ``Exponential(1 / right_scale)``.
     The density is continuous so the left and right densities at ``loc`` agree.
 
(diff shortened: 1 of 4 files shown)


pyro/distributions/asymmetriclaplace.py
```diff
@@ -0,0 +1,73 @@
...
+        super(AsymmetricLaplace, new).__init__(batch_shape, validate_args=False)
+        new._validate_args = self._validate_args
+        return new
+
+    def log_prob(self, value):
+        if self._validate_args:
+            self._validate_sample(value)
+        z = value - self.loc
+        z = -z.abs() / torch.where(z < 0, self.left_scale, self.right_scale)
+        return z - (self.left_scale + self.right_scale).log()
```
fehiepsi, line 50: It took me a bit of time to see the continuity here (haha my mind is not working these days :D )
+ z -> 0- : -|z| / left_scale - log(left_scale + right_scale)
+ z -> 0+ : -|z| / right_scale - log(self_scale + right_scale)
both converges to `- log(self_scale + right_scale)`

pyro/distributions/asymmetriclaplace.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def rsample(self, sample_shape=torch.Size()):
+        shape = self._extended_shape(sample_shape)
+        total_scale = self.left_scale + self.right_scale
+        u = self.loc.new_empty(shape).uniform_() - self.left_scale / total_scale
+        scale = torch.where(u < 0, -self.left_scale, self.right_scale)
+        u = u.mul(total_scale / scale)
+        u = u.clamp(max=1 - torch.finfo(u.dtype).eps)
+        return self.loc - u.neg().log1p().mul(scale)
```
fehiepsi, line 59: I thought 
```
To the left of ``loc`` this acts like an ``-Exponential(left_scale)``;
to the right of ``loc`` this acts like an ``Exponential(right_scale)``
```
will lead to a simple sampler. But that's not the case here. I'll need to think a bit more about what "acts like" means. Somehow, it is reflected in the `log_prob` implementation.

fritzo: Thanks for reviewing!

The main trick with `.rsample()` is that we need to carefully split up the base `uniform(0,1)` such that 0 maps to -inf, 1 maps to inf, and the cdf is monotone. A naive implementation might accidentally map `icdf(0)=loc=icdf(1)` and need to travel through infinity to get from one side to the other.

fehiepsi: Shouldn't this be sampled by `Expon(1 / left_scale) - Expon(1 / right_scale)?`

pyro/distributions/asymmetriclaplace.py
```diff
@@ -0,0 +1,73 @@
...
+
+from .torch_distribution import TorchDistribution
+
+
+class AsymmetricLaplace(TorchDistribution):
+    """
+    Asymmetric version of Laplace distribution.
+
+    To the left of ``loc`` this acts like an ``-Exponential(left_scale)``;
+    to the right of ``loc`` this acts like an ``Exponential(right_scale)``.
```
fehiepsi, line 16: typo: `Exponential(1 / left_scale)`, `Exponential(1 / right_scale)`

pyro/distributions/asymmetriclaplace.py
fehiepsi: Looking like it is the case: take X ~ Expon(1 / right), Y ~ Expon(1 / left), then [the characteristic function](https://en.wikipedia.org/wiki/Laplace_distribution#Relation_to_the_exponential_distribution) of X - Y is
```python
1/(1 + it * right_scale) * 1/(1 - it * left_scale)
```
With the above log_prob, the corresponding characteristic function is
```python
integrate(e^(itx) e^(-|x|/right_scale) / (left_scale + right_scale), 0, inf) + ...
= right_scale / (left_scale + right_scale) * characteristic(Expon(1 / right_scale)) + ...
= (left_scale + right_scale) / (1/right_scale - it) + ...
= (left_scale + right_scale) * [1 / (1/right_scale - it) + 1 / (1/left_scale + it)]
= 1/(1 + it * left_scale) * 1/(1 - it * right_scale)
```

fritzo: Good point. I see we could implement:
```py
def rsample(self, sample_shape):
    return (self.loc - Exponential(1 / self.left_scale).rsample(sample_shape)
            + Exponential(1 / self.right_scale).rsample(sample_shape))
```
Pros: much simpler code.
Cons: reparametrized gradients are a bit noisier due to the extra latent variable.

@fehiepsi which do you think we should use?

fritzo: OK, I just pushed a much simplified and cheaper version directly using `Tensor.exponential_()`. Thanks for the suggestion!

fehiepsi (approved):
LGTM!

Issue #2803, opened 2021-04-13
[FR]: Example of Predictive with MCMC sampling


fonnesbeck:
There are currently no examples in the docs and tutorials of `Predictive` being used in MCMC sampling. Given its importance in returning deterministic values of interest from models, it would probably be useful for users to be able to see how they are used.

mtsokol:
Hi! 
Is it open as a good-first-issue? If yes then I would like to try it. 
I haven't used `pyro` (only TFP) so I would be grateful if you could point out which examples from http://pyro.ai/examples I can refer to and follow while writing it.

fritzo:
Hi @mtsokol, yes this would be a good first issue.

But wait, it looks like the [Bayesian Regression II tutorial](http://pyro.ai/examples/bayesian_regression_ii.html) and the [baseball example](http://pyro.ai/examples/baseball.html) use MCMC and `Predictive`.  @fonnesbeck can you suggest how we might improve these tutorials and examples to make `Predictive` easier to understand?

fonnesbeck:
My mistake--did not notice it in the baseball example. One minor suggestion is for important classes like `Predictive` to have links in the API docs to where it is used in the examples. I think Tensorflow does this (not sure if its done automatically or manually).

mtsokol:
Hi @fritzo! Sorry for the delay and thank you for a guidance!

I did a short (draft with possible errors and misunderstanding) notebook in a "tutorial" way that utilizes `deterministic`, `Predictive` and `MCMC` with the model you provided. But it's pretty short and it repeats content from existing tutorials so I don't think it's worth it. (However writing it was a great way to familiarize myself with the library!)

I can search for another issue in `pyro` or `numpyro` or continue this notebook according to your tips if you consider it worthwhile. What do you think?

Links to the notebook:
https://nbviewer.jupyter.org/gist/mtsokol/fe00f78bb0b925f986d241622c2ec019
https://gist.github.com/mtsokol/fe00f78bb0b925f986d241622c2ec019

(I focused on explaining `deterministic`, `rate` variable and `GammaPoisson` distribution)

fritzo:
Hi @mtsokol your notebook looks great!  Even though it is simple, I still think it is worth publishing it and adding to the index, so feel free to submit a PR.

I think the following additions would also help users:
- Include both words "Predictive" and "Deterministic" in the notebook title, to help users find content.
[...]

PR #2852, opened 2021-05-30
Predictive and Deterministic tutorial

mtsokol:
Hi @fritzo! 

Sorry for a delay, here's a PR with a `Predictive` and `deterministic` tutorial discussed in #2803.

As proposed it's extended with a SVI method (with user defined guide and `AutoGuide`).

As you suggested, in the next PR I will add "Advanced Usage" section with `poutine` routines instead of `Predictive` class for further comparison.

Here are links to the notebook from the PR:

https://gist.github.com/mtsokol/fe00f78bb0b925f986d241622c2ec019
https://nbviewer.jupyter.org/gist/mtsokol/fe00f78bb0b925f986d241622c2ec019

What should I improve/modify?

My questions:

1. When performing inference with SVI it was failing with:
```
ValueError: Error while computing log_prob at site 'counts':
The value argument must be within the support
```
Although `validate_args=False` in `GammaPoisson` fixed it, actual values, which I was inspecting in a debugger, seemed within `GammaPoisson`s support.

2. The guide I defined is rather arbitrary. Is it OK to use `Delta` this way or is it wrong?

Thank you for your help!

view [jupyter notebook](https://nbviewer.jupyter.org/github/mtsokol/pyro/blob/predictive-deterministic-tutorial/tutorial/source/predictive_deterministic.ipynb)


fritzo:
Hi @mtsokol this looks great, thanks for creating a tutorial!  (and sorry for my delay in reviewing.)  I have only minor comments:
- I didn't know about `%config InlineBackend.figure_format='retina'`. If this ends up looking good in your tutorial, we might want to change all other tutorials as well. (but let's check, since IIRC manually setting dpi=300 or svg output broke some tutorials in the past; we'll see on docs.pyro.ai once this merges).
- I think `batch_size` -> `batch_shape` and `event_size` -> `event_shape` would be more standard names.
- The model looks great, I like the use of sklearn's make_regression helper.
- Is there a reason you're using `validate_args=False` in the model?  I'd rather encourage new users to use validation whenever possible.  If you're simply trying to speed up MCMC, you might instead set `jit_compile=True`.
- nit: I think it would help to label the axes on all figures. In particular, it took me a while to figure out what the "rate mean" and "rate std" plots show.
- Wherever you `print(predictive.keys())` I think it would be helpful to instead print sizes as well, say:
  ```py
  for k, v in predictive.items():
      print(f"{k}: {tuple(v.shape)}")
  ```

mtsokol:
@fritzo Thanks for a review!

> Is there a reason you're using validate_args=False in the model?

Without `validate_args=False` SVI is failing with:

```
ValueError: Error while computing log_prob at site 'counts':
The value argument must be within the support
```

Here's a notebook with a full reproduction:
https://colab.research.google.com/drive/165osM9-jI2C8UbCppHTQe0ZpTc-S9WL-?usp=sharing

Is it a bug or my mistake in SVI part?

fritzo:
Hmm, are your counts `y_` actually integers? What if you
```diff
  y_ = torch.tensor((y**3)/100000. + 10., dtype=torch.float)
  y_.round_().clamp_(min=0)
```

mtsokol:
@fritzo Yes, it was my mistake. Now all working! I pushed changes with all fixes from review.

Also gist with a current version:
https://gist.github.com/mtsokol/39f6d679834432bf29e1c35e79f547cf

fritzo (approved):
Looks great, thanks for creating this tutorial!

Issue #2843, opened 2021-05-14
FR Streaming MCMC interface for big models


fritzo:
This issue proposes a streaming architecture for MCMC on models with large memory footprint.

@fehiepsi suggested creating a new MCMC class (say `StreamingMCMC`) with similar interface to `MCMC` and still independent of kernel (using either `HMC` or `NUTS`) but that an internal streaming architecture. Since these models usually run on GPU, it should be fine at first to avoid multiprocessing support in `StreamingMCMC`.

Along with the new `StreamingMCMC` class I think there should be a set of helpers to streamingly compute statistics from sample streams, e.g. mean, variance, covariance, r_hat statistics.

## Tasks
- [ ] Create a `StreamingMCMC` class
- [ ] Create streaming helpers
- [ ] Create a tutorial using `StreamingMCMC` on a big model

mtsokol:
Hi @fritzo!

I was searching for an issue and if this one's free I would like to try solving it.

So as I understand the main point here would be to implement `StreamingMCMC` that doesn't contain `get_samples` method and keeps in its state incrementally updated statistics (if all of them can be incrementally computed, can they?). Something like this:

```python3
class StreamingMCMC:
    def __init__(...):
        self.incremental_mean = ...
        self.incremental_variance = ...
        # and the rest of statistics that will be used by 'summary' method

    def run(self, *args, **kwargs):
        ...
        for x, chain_id in self.sampler.run(*args, **kwargs):
            num_samples += 1
            self.incremental_mean += (x - self.incremental_mean) / num_samples
            # ...and the rest of statistics
            del x
            ...

    def diagnostics(self):
        ...

    def summary(self, prob=0.9):
        # just returns computed incremental statistics
```

Also in `test_mcmc_api.py` additional tests should run both types of `MCMC` classes and compare final statistics.

As described, `StreamingMCMC` shouldn't support multiprocessing manually via `_Worker` because here `CUDA`, which is thought to be the main beneficiary of this new class, handles vectorization by itself. (is it correct?)

Follow up question: Should `StreamingMCMC` have `num_chains` argument and for `num_chains>1` just compute them sequentially or omit this argument?

---

It's seemingly straightforward, but I've just started looking at the source code. Are there any pitfalls that I should bear in mind?

fritzo:
Hi @mtsokol that sounds great and I'm happy to provide any review and guidance.

[...]

PR #2856, opened 2021-06-02
Implement pyro.ops.streaming module

fritzo:
Addresses #2843 

This implements a new module `pyro.ops.streaming` to streamingly track various statistics.  The first intended use case is the planned `StreamingMCMC` class which will track statistics rather than store samples.  There are other potential uses in high-dimensional inference, e.g. recording statistics of gradients during SVI and computing sample moments from predictive when the samples don't fit in memory.

## Design choices

The two basic operations are `.update()` and `.get()`. The third operation `.merge()` will be useful for multiple-chain MCMC and computing things like rhat.

I have restricted to the data type to dictionaries of tensors, which is the basic datatype in pyro.infer.mcmc and in much of NumPyro. We could easily generalize this to pytrees by adding classes `StatsOfList` and `StatsOfTuple`.

## Tested

- [x] tested commutativity of update-get
- [x] tested commutativity and associativity of update-merge-get
- [x] ran mypy locally

### DIFF
--- a/pyro/ops/streaming.py
+++ b/pyro/ops/streaming.py
@@ -148,6 +148,7 @@ class CountMeanVarianceStats(StreamingStats):
     def __init__(self):
         self.shape = None
         self.welford = WelfordCovariance(diagonal=True)
+        super().__init__()
 
     def update(self, sample: torch.Tensor):
         assert isinstance(sample, torch.Tensor)
--- a/pyro/ops/streaming.py
+++ b/pyro/ops/streaming.py
@@ -4,7 +4,7 @@
 import copy
 from abc import ABC, abstractmethod
 from collections import defaultdict
-from typing import Dict, Union
+from typing import Dict, Type, Union
 
 import torch
 
@@ -13,22 +13,24 @@ from pyro.ops.welford import WelfordCovariance
 
 class StreamingStats(ABC):
     """
-    Base class for streamable statistics of dictionaries of tensors.
+    Abstract base class for streamable statistics of trees of tensors.
 
     Derived classes must implelement :meth:`update`, :meth:`merge`, and
     :meth:`get`.
     """
 
     @abstractmethod
-    def update(self, sample: Dict[str, torch.Tensor]) -> None:
+    def update(self, sample) -> None:
         """
         Update state from a single sample.
 
         This mutates ``self`` and returns nothing. Updates should be
         independent of order, i.e. samples should be exchangeable.
 
-        :param dict sample: A sample dictionary mapping sample name to sample
-            value (a :class:`torch.Tensor`).
+        :param sample: A sample value which is a nested dictionary of
+            :class:`torch.Tensor` leaves. This can have arbitrary nesting and
+            shape shape, but assumes shape is constant across calls to
+            ``.update()``.
         """
         raise NotImplementedError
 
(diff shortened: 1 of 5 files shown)


fritzo:
hmm, I think I'll refactor a bit to make it easier to create new statistics

fritzo:
@eb8680 I've added you as a reviewer because these streaming classes create a new semigroup abstraction and you're the resident algebra expert.

fritzo:
@mtsokol I believe you can use the following statistics in #2843:
```
from pyro.ops.streaming import CountMeanVariance, StatsOfDict

stats = StatsOfDict(default=CountMeanVariance)
```
Let me know if it looks like you'll need any changes to this PR.

pyro/ops/streaming.py
```diff
@@ -0,0 +1,196 @@
...
+    def update(self, sample):
+        self.count += 1
+
+    def merge(self, other: "CountStats"):
+        assert isinstance(other, type(self))
+        result = CountStats()
+        result.count = self.count + other.count
+        return result
+
+    def get(self) -> Dict[str, Dict[str, Union[int, torch.Tensor]]]:
```
eb8680, line 76: Should this be `Dict[str, int]`?

pyro/ops/streaming.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __init__(self):
+        self.count = 0
+        super().__init__()
+
+    def update(self, sample):
+        self.count += 1
+
+    def merge(self, other: "CountStats"):
+        assert isinstance(other, type(self))
```
eb8680, line 71: nit: these assertions should no longer be necessary with type hints

pyro/ops/streaming.py
```diff
@@ -0,0 +1,196 @@
...
+    Statistics of samples that are dictionaries with constant set of keys.
+
+    :param default: Default type of statistics of values of the dictionary.
+        Defaults to :class:`CountStats`.
+    :param dict types: Dictionary mapping key to type of statistic that should
+        be recorded for values corresponding to that key.
+    """
+    def __init__(
+        self,
+        types: Dict[object, Type[StreamingStats]] = {},
```
eb8680, line 91: Can this be strengthened to `Dict[str, Type[StreamingStats]]`?

pyro/ops/streaming.py
```diff
@@ -0,0 +1,196 @@
...
+        assert isinstance(other, type(self))
+        result = CountStats()
+        result.count = self.count + other.count
+        return result
+
+    def get(self) -> Dict[str, Dict[str, Union[int, torch.Tensor]]]:
+        return {"count": self.count}
+
+
+class StatsOfDict(StreamingStats):
```
eb8680, line 80: Can you add a unit test exercising the `StatsOfDict` API more explicitly? That might clarify the return types of `get`

pyro/ops/streaming.py
```diff
@@ -0,0 +1,196 @@
...
+    def merge(self, other: "CountMeanStats"):
+        assert isinstance(other, type(self))
+        result = CountMeanStats()
+        result.count = self.count + other.count
+        p = self.count / max(result.count, 1)
+        q = other.count / max(result.count, 1)
+        result.mean = p * self.mean + q * other.mean
+        return result
+
+    def get(self) -> Dict[str, Dict[str, Union[int, torch.Tensor]]]:
```
eb8680, line 139: ditto: should this be `Dict[str, Union[int, torch.Tensor]]`?

pyro/ops/streaming.py
```diff
@@ -0,0 +1,196 @@
...
+        self.stats = defaultdict(default)
+        self.stats.update(**{k: v() for k, v in types.items()})
+        super().__init__()
+
+    def update(self, sample: dict):
+        for k, v in sample.items():
+            self.stats[k].update(v)
+
+    def merge(self, other):
+        result = copy.deepcopy(self)
```
eb8680, line 103: Does `deepcopy` trigger a tensor memory copy?

eb8680 (commented):
Neat API! I'm a little confused by some of the types. Did you try running mypy locally?

fritzo:
@eb8680 thanks for reviewing!

> I'm a little confused by some of the types. Did you try running mypy locally?

Sorry, I did not run mypy locally, and some of the types are stale after refactoring. I'll fix...

pyro/ops/streaming.py
fritzo: Good question! Yes, I just checked and it does trigger a memory copy, as intended in this PR.

pyro/ops/streaming.py
fritzo: Hmm, good point more generally.  However I'd like to argue that we should include both type hints and assertions until all common tools can leverage type hints.  My reasoning is that I'd really like to catch errors as early as possible, e.g. when users (like me) are working in a jupyter notebook.  I think until Jupyter dynamically checks types while editing we'll want extra guard rails especially for tricky interfaces like this.

pyro/ops/streaming.py
fritzo: done.

mtsokol:
Current #2857 draft isn't chain-aware and I'm wondering about it. It can be either handled by `pyro.ops.streaming`, e.g. 

```python3
class CountMeanStats(StreamingStats):
    def __init__(self, num_chains=1):
        self.counts = [0] * num_chains
        ...

    def update(self, sample: torch.Tensor, chain_index=0):
        ...

    def get(self, group_by_chain=True):
        # we can sum across chains
```

so the update would be easy: 

```python3
self._statistics.update({
    name: transformed_sample for name, transformed_sample in z_acc.items()
}, chain_index)
```

Otherwise it can be handled by `StreamingMCMC` with e.g. separate `CountMeanStats` for each chain that can be returned or merged into one if `group_by_chain=False` somewhere in summary.
WDYT?

pyro/ops/streaming.py
fritzo: I strengthened to `Hashable`, but I think we do want to support e.g. integer keys among chains.

fritzo:
@mtsokol I think it's best to keep chain logic in the `StreamingMCMC` class so as to keep stats classes as simple as possible (and hence easy to extend by creating new stats classes).  However I think with the latest couple commits you can easily separate by chain by making either a nested `StatsOfDict` or using keys of the form `(chain_id, site["name"])`.  Let me know if you have any ideas about changing the `StreamingStats` interface to make this easier.

pyro/ops/streaming.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def get(self) -> Dict[str, Union[int, torch.Tensor]]:
+        """
+        :returns: A dictionary with keys ``count: int`` and (if any samples
+            have been collected) ``samples: torch.Tensor``.
+        :rtype: dict
+        """
+        if not self.samples:
+            return {"count": 0}
+        return {"count": len(self.samples), "samples": torch.stack(self.samples)}
```
mtsokol, line 176: Could `"samples": ...` return a dict so we can do `sample_mean = samples['y'].mean()` in tests?

eb8680 (approved):
LGTM after merge conflicts are resolved

fritzo:
Thanks for reviewing @eb8680!  Looks like I'll be using this right away in my mutation models 😄

eb8680 (approved):

PR #2858, opened 2021-06-03
Incorporate mypy into lint stage

fritzo:
Addresses #2550 

@kamathhrishi does this PR look ok to you?  I thought it would be nice to:
1. Incorporate mypy into `make lint`, which we often use to quickly check the codebase.  Since we already have a travis stage to `make lint`, this PR moves mypy into the lint stage in .travis.yml.  I've also added mypy to the dev requirements in setup.py
2. Also check the `examples/` and `scripts/` directories.  Currently `examples/` fails so I've commented it out in the `make lint` command.  I guess we can fix examples in a later PR (lower priority than the library fixes you're working on)

note mypy was giving me weird errors for the `warn_incomplete_stub` changes:
```
setup.cfg: [mypy-pyro._version.*]: Per-module sections should only specify per-module flags (warn_incomplete_stub)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ jobs:
         - stage: lint
           name: lint
           python: 3.6
-          before_install: pip install flake8 isort>=5.0 nbstripout nbformat
+          before_install: pip install flake8 isort>=5.0 mypy nbstripout nbformat
           install:
           script:
               - make lint
@@ -55,13 +55,6 @@ jobs:
               - pip install -r docs/requirements.txt
               - make docs
               - make doctest
-        - name: type_check
-          python: 3.6
-          script:
-              - pip install mypy==0.812
-              - pwd
-              - ls
-              - mypy pyro
         - name: perf
           python: 3.6
           script:
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,9 @@ lint: FORCE
 	flake8
 	isort --check .
 	python scripts/update_headers.py --check
+	mypy pyro
+	# mypy examples  # FIXME
+	mypy scripts
 
 license: FORCE
 	python scripts/update_headers.py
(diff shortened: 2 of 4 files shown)


kamathhrishi (commented):
@fritzo Looks good to me. Adding it with the lint stage and making it part of dev requirements. The idea of checking scripts as part of mypy is a good idea for pythonic ones. 

Yes in the future annotating type hints for examples would be good. Easy for users to be able to understand various data types that can be accommodated by various modules.

kamathhrishi:
Shall I create issues for adding type hints to examples? Further, how about finding a way we mypy throws a warning each time we have ignored a module or used #type ignore somewhere. I wasn't able to find how to do it.

fritzo:
Thanks for reviewing @kamathhrishi!

> Shall I create issues ...

I think it will be easiest if you create a single general mypy issue with a checklist for all the tasks that remain.  I don't know how to give you edit rights to the original issue #2550.

> how about finding a way we mypy throws a warning ...

Sure, once we're done annotating most of the codebase we could write a little one-line script using e.g. `grep`.

eb8680 (approved):
LGTM

kamathhrishi:
> Thanks for reviewing @kamathhrishi!
> 
> > Shall I create issues ...
> 
> I think it will be easiest if you create a single general mypy issue with a checklist for all the tasks that remain. I don't know how to give you edit rights to the original issue #2550.
> 
> > how about finding a way we mypy throws a warning ...
> 
> Sure, once we're done annotating most of the codebase we could write a little one-line script using e.g. `grep`.

@fritzo Since you have writing permission for the repository, you can edit it. I will write a draft here which you can include there. So I thought you can just include a checklist like this:

Add type hints to modules
- [ ] contrib
- [ ] distributions
- [ ] infer
- [ ] nn
- [ ] ops
- [ ] optim
- [ ] params
- [ ] poutine
- [ ] scripts

Additional functionality for type hints include:

- [ ] Give warnings on  _#type ignore_ in codebase 
- [ ] Give warnings on  modules skipped by mypy 


Grey seems like a good idea for finding _#type ignores_

PR #2866, opened 2021-06-10
Enable LocScaleReparam for AsymmetricLaplace distribution

fritzo:
This makes `AsymmetricLaplace` compatible with `LocScaleReparam` by
1. changing `AsymmetricLaplace` distribution's interface to use parameters `(loc, scale, asymmetry)` where asymmetry is the square of the ratio of left to right scales (as in [wikipedia](https://en.wikipedia.org/wiki/Asymmetric_Laplace_distribution) and [scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.laplace_asymmetric.html))
2. changing `LocScaleReparam` to automatically infer `shape_params` of distributions like `StudentT` and `AsymmetricLaplace`.

## Tested
- [x] updated unit tests for `AsymmetricLaplace`
- [x] added unit tests for `LocScaleReparam`


pyro/distributions/asymmetriclaplace.py
```diff
@@ -12,32 +12,39 @@ class AsymmetricLaplace(TorchDistribution):
     """
     Asymmetric version of Laplace distribution.
 
-    To the left of ``loc`` this acts like an ``-Exponential(1 / left_scale)``;
-    to the right of ``loc`` this acts like an ``Exponential(1 / right_scale)``.
-    The density is continuous so the left and right densities at ``loc`` agree.
+    To the left of ``loc`` this acts like an
+    ``-Exponential(1/(asymmetry*scale))``; to the right of ``loc`` this acts
+    like an ``Exponential(asymmetry/scale)`` The density is continuous so the
```
fehiepsi, line 17: nit: missing `.` before ` The`.

fehiepsi (dismissed):
Looks clean to me! It is nice to be consistent with Wikipedia while still exposes easier interpretation properties `left_scale` and `right_scale`.

pyro/distributions/asymmetriclaplace.py
fritzo: Done.

fehiepsi:
@fritzo You'll need to merge dev branch to trigger CI.

fehiepsi (approved):

fritzo:
Thanks for reviewing @fehiepsi!

PR #2872, opened 2021-06-11
Implement a SoftAsymmetricLaplace distribution

### LINKED ISSUE #2868: FR an AsymmetricSoftLaplace distribution
It would be nice to have an `AsymmetricSoftLaplace` distribution, combining features of [AsymmetricLaplace](https://docs.pyro.ai/en/latest/distributions.html#asymmetriclaplace) and [SoftLaplace](https://docs.pyro.ai/en/latest/distributions.html#softlaplace):
- A (loc, scale) family amenable to `LocScaleReparam`
- Learnable skew
- Concave log density, leading to unimodal posteriors
- Infinitely differentiable, leading to smoother gradients in SVI and HMC and tighter Gaussian approximations

One approach is to use the density
```
p(x) = const / (exp(x/k) + exp(-k*x))
```
however according to [wolfram alpha](https://www.wolframalpha.com/input/?i=Integrate%5B1+%2F+%28e%5E%28x%2Fk%29+%2B+e%5E%28-k*x%29%29%2C+%7Bx%7D%5D), the normalizing `const` and `cdf` involve `torch.special.hyp2f1()` which is [not yet implemented](https://github.com/pytorch/pytorch/issues/50345).

Another approach might be to sum two Gamma random variables in opposite directions, with tunable smoothness (e.g. a shared `alpha` parameter). This has an easy `.rsample()` and I believe a tractable `.log_prob()`, but the density is not infinitely differentiable and requires a smoothness parameter.

A third possibility is to Gaussian-blur a standard `AsymmetricLaplace`.  This has tractable `.log_prob()` and `.rsample()` and a tunable smoothness parameter with a plausible default (i.e. half the variance is due to each of the asymmetric laplace and normal terms).

fritzo:
Resolves #2868

This adds a flexible univariate distribution with exponential tails and an infinitely differentiable concave log density.
image

## Numerical stability

This may be numerically unstable due to the `log(erfc(-))`, which will be replace by `torch.logerfc(-)` after https://github.com/pytorch/pytorch/issues/31945

## Tested
- [x] Added examples to tests/distributions/conftest.py

### DIFF
--- a/pyro/distributions/asymmetriclaplace.py
+++ b/pyro/distributions/asymmetriclaplace.py
@@ -193,4 +193,4 @@ def _logerfc(x):
         # Requires https://github.com/pytorch/pytorch/issues/31945
         return torch.logerfc(x)
     except AttributeError:
-        return x.clamp(min=-10, max=10).erfc().log()
+        return x.double().erfc().log().to(dtype=x.dtype)
--- a/pyro/distributions/asymmetriclaplace.py
+++ b/pyro/distributions/asymmetriclaplace.py
@@ -1,6 +1,8 @@
 # Copyright Contributors to the Pyro project.
 # SPDX-License-Identifier: Apache-2.0
 
+import math
+
 import torch
 from torch.distributions import constraints
 from torch.distributions.utils import broadcast_all, lazy_property
@@ -100,8 +102,8 @@ class SoftAsymmetricLaplace(TorchDistribution):
 
     :param loc: Location parameter, i.e. the mode.
     :param scale: Scale parameter = geometric mean of left and right scales.
-    :param asymmetry: Square of ratio of left to right scales.
-    :param softness: Scale parameter of the Gaussian smoother.
+    :param asymmetry: Square of ratio of left to right scales. Defaults to 1.
+    :param softness: Scale parameter of the Gaussian smoother. Defaults to 1.
     """
     arg_constraints = {"loc": constraints.real,
                        "scale": constraints.positive,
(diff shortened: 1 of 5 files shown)


pyro/distributions/asymmetriclaplace.py
```diff
@@ -74,3 +77,120 @@ def variance(self):
...
+    This has a smooth (infinitely differentiable) density with two asymmetric
+    asymptotically exponential tails, one on the left and one on the right. In
+    the limit of ``softness → 0``, this converges in distribution to the
+    :class:`AsymmetricLaplace` distribution.
+
+    This is equivalent to the sum of three random variables ``z - u + v`` where::
+
+        z ~ Normal(loc, scale * softness)
+        u ~ Exponential(1 / (scale * asymmetry))
+        v ~ Exponential(scale / asymmetry)
```
fehiepsi, line 96: typo: asymmetry / scale

pyro/distributions/asymmetriclaplace.py
```diff
@@ -74,3 +77,120 @@ def variance(self):
...
+
+    This is also equivalent the sum of two random variables ``z - a`` where::
```
fehiepsi, line 98: typo: z + a

pyro/distributions/asymmetriclaplace.py
```diff
@@ -74,3 +77,120 @@ def variance(self):
...
+        SS = S * S
+        S2 = S * math.sqrt(2)
+        Lx = L * x
+        Rx = R * x
+
+        # This is the sum of two integrals which are proportional to:
+        # left = Integrate[e^(-t/L - ((x+t)/S)^2/2)/sqrt(2 pi)/S, {t,0,Infinity}]
+        #      = 1/2 e^((2 L x + S^2)/(2 L^2)) erfc((L x + S^2)/(sqrt(2) L S))
+        # right = Integrate[e^(-t/R - ((x-t)/S)^2/2)/sqrt(2 pi)/S, {t,0,Infinity}]
+        #       = 1/2 e^((S^2 - 2 R x)/(2 R^2)) erfc((S^2 - R x)/(sqrt(2) R S))
```
fehiepsi, line 162: the approach sounds reasonable to me. I haven't checked the details but we can trust the tests I think.

fritzo:
@fehiepsi @neerajprad any ideas how to fix coveralls? Our config seems pretty different from the [parallel example](https://github.com/marketplace/actions/coveralls-github-action#complete-parallel-job-example) on github/marketplace/actions.

fehiepsi (approved):

fehiepsi:
The coveralls issue is not related to this PR so I just made a fix and merged to see if the dev branch passes CI. We can change what to put to `ci.yml` file later if needed.

fritzo:

PR #2874, opened 2021-06-16
Implement a SkewLogistic distribution

fritzo:
Addresses #2868 

This is another smooth asymmetric distribution with asymptotically exponential tails addressing #2868 but hopefully more numerically stable than `SoftAsymmetricLaplace` #2872 which is waiting on `torch.special.logerfc()` https://github.com/pytorch/pytorch/issues/31945.

image
image

## Tested
- [x] added to conftest (thank goodness for gof tests!)
- [x] confirmed numerical stability in a large real-world model

### DIFF
--- a/docs/source/distributions.rst
+++ b/docs/source/distributions.rst
@@ -337,6 +337,13 @@ SineSkewed
     :undoc-members:
     :show-inheritance:
 
+SkewLogistic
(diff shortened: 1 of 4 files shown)


fehiepsi (approved):
LGTM!!

PR #2876, opened 2021-06-17
Make ReparamMessenger aware of InitMessenger, simplifying initialization

fritzo:
Addresses #2878
pair coded with @eb8680 and @fehiepsi 

This aims to make initialization compatible with model reparametrization, for example
```py
def model():
    loc = pyro.sample("loc", dist.Normal(0, 1))
    scale = pyro.sample("scale", dist.LogNormal(0, 1))
    pyro.sample("x", dist.Normal(loc, scale))

reparam_model = pyro.reparam(model, {"x": LocScaleReparam()})
reparam_guide = AutoNormal(reparam_model, init_loc_fn=init_to_value({"x": torch.tensor(0.0)}))
assert reparam_guide()["x"] == 0
```

This involves a major interface change for reparametrizers.

The problem is that the `init_loc_fn` passed to autoguides and HMC are often customized to particular latent sites, but reparametrizers replace those sites with new auxiliary latent sites.  The solution proposed in this PR is to require each reparametrizer to know how to transform original values to latent values, and to use that in initialization.

## Tasks deferred to future PRs

- Support value initialization in `NeuTraReparam`
- Support value initialization in `StructuredReparam`

## Tested
- [x] tests for all new reparameterizers
- [x] tests for nesting and `msg["is_observed"]`
- [x] refactoring of `pyro.contrib.epidemiology` is covered by existing unit tests
- [x] add `check_init_reparam()` tests for all reparametrizers (ensuring coverage)


fehiepsi:
I like to have this feature implemented. This has been requested several times previously.

pyro/contrib/epidemiology/compartmental.py
```diff
@@ -734,13 +734,11 @@ def _heuristic(self, haar, **options):
...
         assert "auxiliary" in init_values, \
             ".heuristic() did not define auxiliary value"
-        if haar:
-            haar.user_to_aux(init_values)
         logger.info("Heuristic init: {}".format(", ".join(
             "{}={:0.3g}".format(k, v.item())
             for k, v in sorted(init_values.items())
             if v.numel() == 1)))
-        return init_to_value(values=init_values)
+        return init_to_value(values=init_values, fallback=None)
```
fritzo, line 741: I've added fallback logic to a few init strategies (hoping to port these to NumPyro and combine with https://github.com/pyro-ppl/numpyro/pull/1058), and set the fallback to `None` here, which would error if the new initialization logic were to fail.

pyro/infer/autoguide/initialization.py
```diff
@@ -54,10 +55,20 @@ def init_to_sample(site=None):
     return site["fn"].sample().detach()
 
 
-def init_to_median(site=None, num_samples=15):
+def init_to_median(
+    site=None,
+    num_samples=15,
+    *,
+    fallback: Optional[Callable] = init_to_feasible,
```
fritzo, line 62: I've added `fallback` kwargs to a few init strategies mainly to make it easier for users to catch unintended initialization.

pyro/infer/autoguide/initialization.py
```diff
@@ -194,4 +234,8 @@ def _pyro_sample(self, msg):
                     "{} provided invalid shape for site {}:\nexpected {}\nactual {}"
                     .format(self.init_fn, msg["name"], msg["value"].shape, value.shape))
         msg["value"] = value
-        msg["done"] = True
```
fritzo, line 197: @eb8680 can you confirm this is ok?  I think we want to avoid setting `done` here so as to enable subsequent reparametrization.

fritzo:
@eb8680 @fehiepsi let me know if you'd like to do a code walk-through.  This is a big refactoring and I'd like to eventually port it to NumPyro.

fritzo:
Looks like all tests are passing again, let me know if you have any questions.

pyro/infer/autoguide/initialization.py
```diff
@@ -51,16 +54,30 @@ def init_to_sample(site=None):
...
+    site=None,
+    num_samples=15,
+    *,
+    fallback: Optional[Callable] = init_to_feasible,
+):
     """
-    Initialize to the prior median; fallback to a feasible point if median is
-    undefined.
+    Initialize to the prior median; fallback to ``fallback`` (defaults to
+    :func:`init_to_feasible`` if mean is undefined.
```
fehiepsi, line 70: nit: missing the closing `)`.

pyro/infer/reparam/neutra.py
```diff
@@ -71,26 +84,31 @@ def __call__(self, name, fn, obs):
                                  "`get_transform` method that does not depend on the "
                                  "model's `*args, **kwargs`") from e
 
-            z_unconstrained = pyro.sample("{}_shared_latent".format(name),
-                                          self.guide.get_base_dist().mask(False))
+            with ExitStack() as stack:
+                for plate in self.guide.plates.values():
+                    stack.enter_context(block_plate(dim=plate.dim, strict=False))
```
fehiepsi, line 89: I just wonder why we need to `block_plate` here (and some other places). Could you elaborate a bit?

pyro/poutine/reparam_messenger.py
```diff
@@ -44,20 +54,64 @@ def _pyro_sample(self, msg):
...
-                msg["fn"]._validate_sample(value)
-        msg["fn"] = new_fn
+                msg["fn"]._validate_sample(new_msg["value"])
+
+            if msg["value"] is not None and msg["value"] is not new_msg["value"]:
+                # Check that overwritten initialization preserves shape.
+                if not torch._C._get_tracing_state():
+                    assert new_msg["value"].shape == msg["value"].shape
+
+                # Warn if a custom init method is overwritten by another init method.
```
fehiepsi, line 102: Sounds reasonable to me.

fehiepsi (commented):
The design looks great to me. Thanks for taking care of the user-defined strategies and backward compatibility. I just catch a small nit.

pyro/infer/autoguide/initialization.py
fritzo: thanks, fixed!

pyro/poutine/reparam_messenger.py
fritzo: Yeah, this is from personal experience where I spent about 6 hours hunting down a bug that ended up being overwritten initialization 😅

pyro/infer/reparam/neutra.py
fritzo: The problem addressed is with in models like
```py
@neutra.reparam()
def model():
    with pyro.plate("plate", 10):
        pyro.sample("x", dist.Normal(0,1))
```
where the `z_unconconstrained` shared latent is a single latent variable that is not independent along batch dimensions; rather it should have empty `batch_shape` (except for particle plates that are outside the guide and hence not blocked) and has `event_shape = (fn.shape().numel(),)`.  Without this `block` statement the `z_unconstrained` would have `event_shape = (10,)` but would then be broadcast up by an enclosing plate to `batch_shape + event_shape = (10, 10)`.  This this `block_plate` merely hoists that shared latent variable outside of all plates.

I believe this is what we want to do with `NeuTra`, but IIUC it prevents "batched neutra" in that we won't be able to share a single normalizing flow along a batch dimension and subsample.  I'm not sure whether that prevented use case is realistic though, since our HMC and NUTS do not support data subsampling.

eb8680 (approved):
LGTM after our walkthrough with extra tests and nits addressed

PR #2879, opened 2021-06-18
Implementation of rhat for StreamingMCMC

mtsokol:
@fritzo Hi! Here's a draft/proposition of `rhat` diagnostic for `StreamingMCMC`. I now that in #2843 `rhat` is stated as a `streaming.ops` member but I just wanted to share this idea (not necessarily stick to this approach and merge it).

Current `StreamingMCMC` implementation is chain-aware and new samples are added to streaming statistics by `(chain_id, name)` key (via `StatsOfDict`). This makes it impossible to create a streaming operator consuming all chains simultaneously (I think).

This implementation introduces `def diagnostics(self):` to `StreamingMCMC` which follows existing `MCMC` methods and relies on `mean` and `variance` ops that must be computed (therefore are required to be present). I followed the existing `gelman_rubin(-)` implementation. WDYT?

### DIFF
--- a/tests/infer/mcmc/test_mcmc_api.py
+++ b/tests/infer/mcmc/test_mcmc_api.py
@@ -219,25 +219,35 @@ def test_null_model_with_hook(run_mcmc_cls, kernel, model, jit, num_chains):
         assert iters == expected
 
 
+@pytest.mark.parametrize("run_mcmc_cls", [
+    run_default_mcmc,
+    run_streaming_mcmc
+])
 @pytest.mark.parametrize("num_chains", [
     1,
     2
 ])
 @pytest.mark.filterwarnings("ignore:num_chains")
-def test_mcmc_diagnostics(num_chains):
+def test_mcmc_diagnostics(run_mcmc_cls, num_chains):
     data = torch.tensor([2.0]).repeat(3)
     initial_params, _, transforms, _ = initialize_model(normal_normal_model,
                                                         model_args=(data,),
                                                         num_chains=num_chains)
     kernel = PriorKernel(normal_normal_model)
-    mcmc = MCMC(kernel, num_samples=10, warmup_steps=10, num_chains=num_chains, mp_context="spawn",
-                initial_params=initial_params, transforms=transforms)
+    if run_mcmc_cls == run_default_mcmc:
+        mcmc = MCMC(kernel, num_samples=10, warmup_steps=10, num_chains=num_chains, mp_context="spawn",
+                    initial_params=initial_params, transforms=transforms)
+    else:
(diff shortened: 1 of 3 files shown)


tests/infer/mcmc/test_mcmc_api.py
```diff
@@ -219,25 +219,35 @@ def test_null_model_with_hook(run_mcmc_cls, kernel, model, jit, num_chains):
...
+                             initial_params=initial_params, transforms=transforms)
     mcmc.run(data)
     if not torch.backends.mkl.is_available():
         pytest.skip()
     diagnostics = mcmc.diagnostics()
-    assert diagnostics["y"]["n_eff"].shape == data.shape
+    # TODO n_eff for streaming MCMC
+    # assert diagnostics["y"]["n_eff"].shape == data.shape
     assert diagnostics["y"]["r_hat"].shape == data.shape
+    print(diagnostics["y"])
```
fritzo, line 250: nit: keep old test around for existing MCMC:
```py
# TODO implement n_eff for StreamingMCMC.
if run_mcmc_cls == run_default_mcmc:
    assert diagnostics["y"]["n_eff"].shape == data.shape
```

fritzo (commented):
Hi @mtsokol this looks great: your solution is simpler and lower-memory than I had anticipated.  I think this is almost ready to merge as-is.  The only suggestions would be to (1) consider moving the functionality out into a helper, similar to the way MCMC is implemented:
```py
# in pyro/infer/mcmc/util.py
def diagnostics_from_stats(stats):
    ...

# in pyro.infer/mcmc/api.py
class StreamingMCMC(AbstractMCMC):
    def diagnostics(self):
        statistics = self._statistics.get()
        return diagnostics_from_stats(statistics)
```
and (2) maybe manipulate the dict a bit so that the multi-chain `.diagnostics()` result dict has the same structure when returned from MCMC versus StreamingMCMC (I'm unsure whether any further manipulation is needed).

tests/infer/mcmc/test_mcmc_api.py
mtsokol: Sure, done!

mtsokol:
@fritzo Thanks for a review! (1) I moved the implementation to `pyro/infer/mcmc/api.py`. (2) I checked diagnostics' structure between `MCMC` and `StreamingMCMC` in that test (also single and multichain output) and the only difference was that `MCMC` returns `OrderedDict`, so I added that.

Also I added abstract `def diagnostics(self):` method to `AbstractMCMC` as both implementations comply with it.

PR marked as ready!

fritzo (approved):
LGTM, thanks for implementing this!

PR #2880, opened 2021-06-20
Improve docstring example of PyroLRScheduler

gui11aume:


fritzo (commented):
Hi @gui11aume, thanks for the fix!  Just to clarify, does this PR fix previously incorrect code, or is it only a more extended example?

fritzo (commented):
Hi @gui11aume, thanks for the fix!  Just to clarify, does this PR fix previously incorrect code, or is it only a more extended example?

pyro/optim/lr_scheduler.py
```diff
@@ -21,8 +21,12 @@ class PyroLRScheduler(PyroOptim):
...
     Example::
 
-        optimizer = torch.optim.SGD
-        scheduler = pyro.optim.ExponentialLR({'optimizer': optimizer, 'optim_args': {'lr': 0.01}, 'gamma': 0.1})
+	optimizer = torch.optim.SGD
+	scheduler = pyro.optim.PyroLRScheduler(
+	      scheduler_constructor = torch.optim.lr_scheduler.ExponentialLR,
+	      optim_args = {'optimizer': optimizer, 'optim_args': {'lr': 0.01}, 'gamma': 0.1},
+	      clip_args = {'clip_norm': 5.} 
+	)
```
fritzo, line 29: it looks like you'll need to outdent two spaces in this block

Issue #2883, opened 2021-06-23
FR Automatic reparameterization strategies


fritzo:
As our available [reparameterization strategies](https://docs.pyro.ai/en/dev/infer.reparam.html) become increasingly complex and composable, it would be useful to provide some automatic strategies, possibly even enabled by default in autoguides and MCMC.

Maybe this could work with syntax like `poutine.reparam(config="auto")` or `poutine.reparam(config="full")`.

## Proposed automatic strategies
- *"minimal":* A minimal strategy that reparametrizes only sites that requires reparametrization for mathematical reasons, e.g. `StableReparam` for `Stable` distributions, `CircularReparam` #2740 for  `VonMises`, `ProjectedNormalReparam` for `ProjectedNormal` sites.
- *"full":* Complete reparametrization after which all `pyro.sample` sites are parameter-free, as required by #2768.  This should error on distributions with implicit reparameterization, e.g. `Gamma` with a learnable `concentration` parameter.
- *"auto":* Recommended reparametrization including something like
  - "minimal";
  - `LocScaleReparam` to all distributions with `.loc` and `.scale` attributes (will require unwrapping `Independent` and `Masked` distributions);
  - `SoftmaxReparam` for `Simplex`-constrained random variables.

eb8680:
> Should we automatically use these in AutoGuide and MCMC?

Would it also make sense to refactor the automatic transformation to unconstrained space in [MCMC](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/mcmc/util.py#L335) and [ADVI](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/autoguide/guides.py#L710) to use `poutine.reparam` and an automated `"unconstrain"` reparameterization strategy?

eb8680:
Also, the [enumeration strategies](https://github.com/pyro-ppl/pyro/blob/dev/pyro/poutine/enum_messenger.py#L55) are tantalizingly close to being `Reparam` classes - I wonder if we could actually rewrite them that way and make `config_enumerate` into a special case of `ReparamMessenger`.

fritzo:
> Would it also make sense to refactor the automatic transformation to unconstrained space...

[...]

Issue #2831, opened 2021-05-04
FR make Reparam objects nn.Modules for jitting, saving, etc.


fritzo:
As we move away from the global param store and towards closer integration with `torch.nn.Module`s, it would be good to refactor `Reparam` objects to be `nn.Module`s.

## Design questions

- [ ] Can we preserve backwards compatibility?
- [ ] Would a `class Reparam(nn.Module)` be jittable, even though its `.forward()` method inputs non-tensors, e.g. distributions?

## Tasks

- [ ] make `Reparam` objects `nn.Module`s
- [ ] replace `Reparam.__call__()` with `.forward()`
- [ ] replace `pyro.param` statements in `Reparam` objects with `getattr(self, -)` or `setattr(self, -, PyroParam(-))`

PR #2884, opened 2021-06-23
Add automatic reparametrization strategies

fritzo:
Addresses #2883

This adds two new automatic reparametrization strategies:
- `MinimalReparam` for weird distributions without `.log_prob` methods, e.g. `ProjectedNormal` and `Stable`
- `AutoReparam` for automatically applying recommended reparametrizers, e.g. loc-scale wherever possible.

### Design notes

- I've created classes rather than functions in anticipation of possible future promotion of `Reparam` objects to `nn.Module`s #2831. These classes would then also become `nn.Module`s.
- Ideally `AutoReparam` would only apply `LocScaleReparam` to sites with non-constant parameters, however this is blocked by the dependency tracking machinery being implemented in #2824. Once that merges, we can update `AutoReparam` to trace dependencies on the first model call, similar to what is done in `NeutraReparam` and `StructuredReparam`.

## Tested

- [x] unit tests to verify behavior
- [x] end-to-end tests to ensure these are compatible with other inference tools


pyro/infer/reparam/strategies.py
```diff
@@ -0,0 +1,180 @@
...
+
+def _loc_scale_reparam(name, fn):
+    if "_decentered" in name:
+        return  # Avoid infinite recursion.
+
+    # Check for location-scale families.
+    params = set(fn.arg_constraints)
+    if not {"loc", "scale"}.issubset(params):
+        return
+    assert not isinstance(fn, torch.distributions.LogNormal)
```
fritzo, line 126: TODO make this more generic by checking that support is constraints.real

pyro/infer/reparam/strategies.py
```diff
@@ -0,0 +1,180 @@
...
+
+    Derived classes must implement the :meth:`configure` method.
+
+    :ivar dict config: A dictionary configuration. This will be populated the
+        first time the model is run. Thereafter it can be used as an argument
+        to ``poutine.reparam(config=___)``.
+    """
+
+    def __init__(self):
+        self.config: Dict[str, Optional[Reparam]] = {}
```
fritzo, line 39: TODO comment that this can become a ModuleDict as part of #2831

pyro/infer/reparam/strategies.py
```diff
@@ -0,0 +1,180 @@
...
+    which is equivalent to::
+
+        @poutine.reparam(config=AutoReparam())
+        def model(...):
+            ...
+
+    .. warning:: This strategy may change behavior across Pyro releases.
+        To inspect or save a given behavior, extract the ``.config`` dict after
+        running the model at least once.
+    """
```
fritzo, line 160: TODO add a `centered` argument, so users could use `AutoReparam(centered=0.0)` for MCMC

pyro/infer/reparam/strategies.py
```diff
@@ -0,0 +1,180 @@
...
+
+from .loc_scale import LocScaleReparam
+from .projected_normal import ProjectedNormalReparam
+from .reparam import Reparam
+from .softmax import GumbelSoftmaxReparam
+from .stable import LatentStableReparam, StableReparam, SymmetricStableReparam
+from .transform import TransformReparam
+
+
+class Strategy(ABC):
```
martinjankowiak, line 27: should there be a `prefix` arg like in `AutoGuides` to avoid possible name mangling... i guess this would require changes on the `Reparam` end...

martinjankowiak (approved):

pyro/infer/reparam/strategies.py
fritzo: I guess we could add an `infix` arg to separate user-provided names from reparam suffixes, but I think that would best live at the level of base `Reparam` class and seems outside the scope of this PR. And maybe we should wait until we have at least one report of real-world name conflict before adding this bit of complexity.

martinjankowiak: sounds reasonable but should we add a note in the docs to the effect of 

"Note that reparametrizers introduce auxiliary sites with names like `my_latent_decentered`. In case of name conflict you should rename your sample sites accordingly."

martinjankowiak (approved):

pyro/infer/reparam/strategies.py
fritzo: SGTM, I'll address in a separate PR. I think we may be able to add some error checking as well.

PR #2893, opened 2021-07-03
Bug in `elbo` calculation in `contrib.funsor.infer.trace_elbo.Trace_ELBO`

ordabayevy:
I get wrong values for `elbo` (much larger absolute value compared to `pyro.infer.trace_elbo.Trace_ELBO`), presumably because `sum(log_factors, to_funsor(0.0))` in line 41 broadcasts terms in `log_factors` and then that leads to large absolute values after `elbo.reduce(funsor.ops.add, plate_vars)` summation.

Here I try to fix it by reducing each cost term individually similar to `TraceEnum_ELBO`. I'm also not sure if integration is needed here.

### DIFF
--- a/pyro/contrib/funsor/infer/trace_elbo.py
+++ b/pyro/contrib/funsor/infer/trace_elbo.py
@@ -29,19 +29,14 @@ class Trace_ELBO(ELBO):
         model_terms = terms_from_trace(model_tr)
         guide_terms = terms_from_trace(guide_tr)
 
-        log_measures = guide_terms["log_measures"] + model_terms["log_measures"]
-        log_factors = model_terms["log_factors"] + [
+        costs = model_terms["log_factors"] + [
             -f for f in guide_terms["log_factors"]
         ]
         plate_vars = model_terms["plate_vars"] | guide_terms["plate_vars"]
-        measure_vars = model_terms["measure_vars"] | guide_terms["measure_vars"]
 
-        elbo = funsor.Integrate(
-            sum(log_measures, to_funsor(0.0)),
-            sum(log_factors, to_funsor(0.0)),
-            measure_vars,
-        )
-        elbo = elbo.reduce(funsor.ops.add, plate_vars)
+        elbo = to_funsor(0.)
+        for cost in costs:
+            elbo += cost.reduce(funsor.ops.add, plate_vars & frozenset(cost.inputs))
 
         return -to_data(elbo)
 
--- a/pyro/contrib/funsor/infer/trace_elbo.py
+++ b/pyro/contrib/funsor/infer/trace_elbo.py
@@ -29,12 +29,10 @@ class Trace_ELBO(ELBO):
         model_terms = terms_from_trace(model_tr)
         guide_terms = terms_from_trace(guide_tr)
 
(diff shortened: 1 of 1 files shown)


pyro/contrib/funsor/infer/trace_elbo.py
```diff
@@ -29,19 +29,12 @@ def differentiable_loss(self, model, guide, *args, **kwargs):
...
-        elbo = funsor.Integrate(
-            sum(log_measures, to_funsor(0.0)),
-            sum(log_factors, to_funsor(0.0)),
-            measure_vars,
-        )
-        elbo = elbo.reduce(funsor.ops.add, plate_vars)
+
+        elbo = to_funsor(0.0)
+        for cost in costs:
+            elbo += cost.reduce(funsor.ops.add, plate_vars & frozenset(cost.inputs))
```
eb8680, line 37: Isn't this missing Dice factors included in `log_measures`? IIRC that was the reason for using `Integrate`.

ordabayevy: I copied the test from https://github.com/pyro-ppl/pyro/pull/2894 which has a simple model/guide pair. When running that model both `guide_terms["log_measures"]` and `model_terms["log_measures"]` are empty. I can't find Dice factors anywhere in `model_terms` or `guide_terms`.

eb8680: I guess they're not included because `Funsor.sample` isn't used in the evaluation of `Trace_ELBO`. I don't think `contrib.funsor.infer.Trace_ELBO` is tested extensively outside the `pyro-api` tests in [`tests/contrib/funsor/test_pyroapi_funsor.py`](https://github.com/pyro-ppl/pyro/blob/dev/tests/contrib/funsor/test_pyroapi_funsor.py), which is why this wasn't noticed before.

A more general Funsor-based implementation of `Trace_ELBO` is certainly possible and would look very similar to the guide-side enumeration handling logic in `TraceEnum_ELBO`. We might even be able to write a custom "enumeration" strategy that just called `Funsor.sample` and reuse `TraceEnum_ELBO` as the `Trace_ELBO` implementation.

I believe a completely general version might require variable elimination logic beyond what's currently in `funsor.sum_product` handling cases where the guide had plate structure incompatible with the restrictions there, although I can't immediately think of existing tests or examples where that would be the case.

pyro/contrib/funsor/infer/trace_elbo.py
```diff
@@ -21,29 +22,68 @@ def differentiable_loss(self, model, guide, *args, **kwargs):
         with enum(), plate(
             size=self.num_particles
         ) if self.num_particles > 1 else contextlib.ExitStack():
-            guide_tr = trace(config_enumerate(default="flat")(guide)).get_trace(
-                *args, **kwargs
-            )
+            guide_tr = trace(
+                config_enumerate(default="flat", num_samples=self.num_particles)(guide)
```
ordabayevy, line 26: Without `num_samples` argument here `config_enumerate` wasn't adding `{"enumerate": "flat"}` to `msg["infer"]`.

PR #2895, opened 2021-07-05
Add Logistic distribution

fritzo:
Addresses https://github.com/pytorch/pytorch/issues/7857

This aims to replace the less standard `SoftLaplace` distribution of #2791 and #2800 .

image

## Tested
- added test examples to tests/distributions/conftest.py


fehiepsi (approved):
LGTM! It is surprising that this distribution is not yet available in PyTorch.

PR #2904, opened 2021-07-20
Implement PositivePowerTransform

fritzo:
This implements a `PositivePowerTransform`.  Whereas `PowerTransform` allows arbitrary exponent and requires positive domain and codomain, this requires positive exponent and allows real domain and codomain.

The intended use case is in sparse regression problems where this can be combined with `TransformReparam` and `AutoNormal` / `AutoLowRankMultivariateNormal` guides, achieving a horseshoe-like prior while attempting to keep the transformed posterior unimodal:
```py
scale = pyro.sample("scale", LogNormal(0, 1))
sparsity = pyro.sample("sparsity", LogNormal(0, 1))
with poutine.reparam(config={"x": TransformReparam()}):
    x = pyro.sample(
        "x",
        TransformedDistribution(
            Normal(0, scale),
            PositivePowerReparam(sparsity),
        ),
    )
```

## Tested
- [x] added a transform test

### DIFF
--- a/pyro/distributions/transforms/power.py
+++ b/pyro/distributions/transforms/power.py
@@ -1,0 +1,59 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+from torch.distributions import Distribution, constraints
+from torch.distributions.transforms import Transform
+
+
+class PositivePowerTransform(Transform):
+    r"""
+    Transform via the mapping :math:`y = x^{\text{exponent}}`.
+
+    Whereas :class:`~torch.distributions.transforms.PowerTransform` allows
+    arbitrary ``exponent`` and restricts domain and codomain to postive values,
+    this class restricts ``exponent > 0`` and allows real domain and codomain.
+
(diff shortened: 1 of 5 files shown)


pyro/distributions/transforms/power.py
```diff
@@ -0,0 +1,59 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+from torch.distributions import Distribution, constraints
+from torch.distributions.transforms import Transform
+
+
+class PositivePowerTransform(Transform):
+    r"""
+    Transform via the mapping :math:`y = x^{\text{exponent}}`.
```
martinjankowiak, line 11: shouldn't this equation look more like line 47? under canonical interpretation this mapping can yield complex numbers, right?

martinjankowiak (approved):

pyro/distributions/transforms/power.py
fritzo: oh, like `y = sign(x) * abs(x) ** exponent`?  Sure I'll update the docs.

martinjankowiak (approved):

PR #2905, opened 2021-07-20
fix order of parameters in predictive

TradDog:
Fix predictive parameters. Please review. I am a perfect beginner in pyro and ppl in general.


TradDog:
Suggestion to extend the MCMC example to replicate the SVI 90% CI.

eb8680 (changes requested):
@TradDog thanks for the PR! Sorry we have been slow to review.

Can you rerun the notebook in a different environment such that all the new error/debugging messages in your latest commit are scrubbed and remove the new cells in the HMC section with regression line plots copied from part 1?

PR #2906, opened 2021-07-20
Cleanup coveralls in CI based on latest documentation

neerajprad:
I have simplified the earlier invocation to adhere to the latest recommendations for integration, such as not setting `GITHUB_REF`. I am still not sure if this will help with getting the correct badge on the dev branch until this PR is merged.


fritzo (approved):
Thanks for figuring this out @neerajprad !

neerajprad:
Thanks for reviewing, @fritzo. This should be fixed now but github could still be caching the old icon so you may need to clear that.

PR #2914, opened 2021-08-17
Add ParamStoreDict.scope() method

### LINKED ISSUE #2912: [discussion] Multiple guides/models in the same process
Following on from [this brief discussion](https://forum.pyro.ai/t/multiple-guides-models-in-the-same-process/3240) on the forum...

I would like to compare the results of using different guides for the same model (or the results from different models with different guides). Similar to this post ([Multiple guides for same model ](https://forum.pyro.ai/t/multiple-guides-for-same-model/1316)) it is of course possible to clear the global parameter store between runs of different models/guides, but that makes interactive comparison of the models/guides rather cumbersome. I’ve been using the following context manager to create multiple parameter stores (i.e. within the scope, a particular “local” parameter store is used rather than the “global” one).

Is this a sensible approach or is there something more straightforward that I’ve been missing? Is there interest in integrating this type of behaviour into the pyro codebase? This may be related to #502 in terms of global parameter stores sometimes being challenging.

```python
import pyro
from pyro.params.param_store import ParamStoreDict


class paramstore_scope:
    """
    Context manager for using multiple parameter stores within the same process.
    """
    GLOBAL_STORE = None
    
    def __init__(self, paramstore=None):
        self.paramstore = paramstore or ParamStoreDict()
        
    def __enter__(self):
        if self.GLOBAL_STORE is not None:
            raise RuntimeError('paramstore scopes cannot be nested')
        # Store the global paramstore.
        self.__class__.GLOBAL_STORE = pyro.poutine.runtime._PYRO_PARAM_STORE
        self._setup_paramstore(self.paramstore)
        return self
        
    def __exit__(self, *args):
        # Restore the global paramstore.
        self._setup_paramstore(self.GLOBAL_STORE)
        self.__class__.GLOBAL_STORE = None
        
    def _setup_paramstore(self, paramstore):
        # Replace all imports of the global param store with our version.
        pyro.poutine.runtime._PYRO_PARAM_STORE = \
            pyro.primitives._PYRO_PARAM_STORE = \
            pyro.nn.module._PYRO_PARAM_STORE = \
            paramstore
        # Replace static use of the imported param store.
        pyro.primitives._param = \
            pyro.poutine.runtime.effectful(paramstore.get_param, type="param")
```

An example looks like this.

```python
with paramstore_scope() as scope1:
    # Do your inference for model 1 (which will leave the global paramstore unchanged).

with paramstore_scope() as scope2:
    # Do your inference for model 2 (which will leave the global paramstore unchanged).

# Interactively explore the different models.
```

fritzo:
Resolves #2912 

This addresses @tillahoffman's idea of a context-manager-based way of handling multiple param store states in a single process, simplifying the manual use of `.get_state()`, `.clear()`, and `.set_state()`.

## Tested
- [x] unit tests

### DIFF
--- a/pyro/params/param_store.py
+++ b/pyro/params/param_store.py
@@ -244,21 +244,20 @@ class ParamStoreDict:
         Get the ParamStore state.
         """
         state = {
-            "params": self._params,
-            "constraints": self._constraints,
+            "params": self._params.copy(),
+            "constraints": self._constraints.copy(),
         }
         return state
 
     def set_state(self, state: dict):
         """
-        Set the ParamStore state using state from a previous get_state() call
+        Set the ParamStore state using state from a previous :meth:`get_state` call
         """
         assert isinstance(state, dict), "malformed ParamStore state"
         assert set(state.keys()) == set(
             ["params", "constraints"]
         ), "malformed ParamStore keys {}".format(state.keys())
 
-        self.clear()
         for param_name, param in state["params"].items():
             self._params[param_name] = param
             self._param_to_name[param] = param_name
@@ -271,7 +270,7 @@ class ParamStoreDict:
 
     def save(self, filename):
         """
-        Save parameters to disk
+        Save parameters to file
 
         :param filename: file name to save to
         :type filename: str
@@ -281,7 +280,7 @@ class ParamStoreDict:
 
     def load(self, filename, map_location=None):
         """
-        Loads parameters from disk
+        Loads parameters from file
 
         .. note::
 
@@ -330,9 +329,12 @@ class ParamStoreDict:
             state = {"params": {}, "constraints": {}}
         old_state = self.get_state()
         try:
+            self.clear()
             self.set_state(state)
             yield state
+            state.update(self.get_state())
         finally:
+            self.clear()
             self.set_state(old_state)
 
 
--- a/pyro/params/param_store.py
+++ b/pyro/params/param_store.py
@@ -4,6 +4,7 @@
 import re
 import warnings
 import weakref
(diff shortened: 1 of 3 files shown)


tests/params/test_param.py
```diff
@@ -149,3 +149,67 @@ def test_dict_interface():
     assert list(param_store.keys()) == []
     assert list(key for key, value in param_store.items()) == []
     assert len(list(param_store.values())) == 0
+
+
+def test_scope():
```
martinjankowiak, line 154: this doesn't seem to test the constraints?

martinjankowiak (dismissed):
looks great!

tests/params/test_param.py
fritzo: sure, new assertions added.

pyro/params/param_store.py
```diff
@@ -295,6 +299,44 @@ def load(self, filename, map_location=None):
...
+
+            # Now evaluate each, still avoiding name conflicts.
+            with param_store.scope(scope1):  # loads the first model's scope
+               # ...evaluate the first model...
+            with param_store.scope(scope2):  # loads the second model's scope
+               # ...evaluate the second model...
+        """
+        if state is None:
+            state = {"params": {}, "constraints": {}}
+        old_state = self.get_state()
```
tillahoffmann, line 330: Is there merit in raising an exception when scopes are nested? Or are nested scopes a valid use case?

fritzo: I think nested scopes are valid use cases

martinjankowiak (approved):

PR #2917, opened 2021-08-20
Clarify error message on model-side sequential enumeration

### LINKED ISSUE #2916: FR Warn that model-side sequential enumeration is not implemented
The [TraceEnum_ELBO](https://docs.pyro.ai/en/dev/inference_algos.html#pyro.infer.traceenum_elbo.TraceEnum_ELBO) docs state that guide-side enumeration allows either parallel or sequential enumeration, and model-side allows sequential. However if an unsuspecting user has not read this weak fine print and attempts to sequentially enumerate on the model side, the only warning is
```
UserWarning: Found vars in model but not guide: {'x'}
```
It would be helpful to explicitly warn that model-side sequential enumeration is not implemented.

See e.g. confusion arising on [this forum post](https://forum.pyro.ai/t/inferring-discrete-latent-variable/3172) with @martinjankowiak

fritzo:
Resolves #2916 

## Tested
- [x] added two new unit tests
- [x] updated `test_enum_sequential_in_model_error()` with more-specific error message

### DIFF
--- a/tests/infer/test_valid_models.py
+++ b/tests/infer/test_valid_models.py
@@ -1575,7 +1575,7 @@ def test_enum_sequential_in_model_error():
         model,
         guide,
         TraceEnum_ELBO(max_plate_nesting=0),
-        match="Found vars in model but not guide",
+        match="At site .*, model-side sequential enumeration is not implemented",
     )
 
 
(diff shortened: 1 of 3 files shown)


tests/infer/test_valid_models.py
```diff
@@ -1575,7 +1575,7 @@ def guide():
         model,
         guide,
         TraceEnum_ELBO(max_plate_nesting=0),
-        match="Found vars in model but not guide",
+        match="At site .*, model-side sequential enumeration is not implemented",
```
martinjankowiak, line 1578: ah i guess we even had a test of sorts....

martinjankowiak (approved):

PR #2918, opened 2021-08-24
Add pyro.util.deep_to() to recursively call .to() on Python data structures

fritzo:
Addresses https://github.com/pytorch/pytorch/issues/7795
Addresses https://github.com/pytorch/pytorch/issues/17234

Many users of torch.distributions claim to want a `.to()` method on `Distribution` objects. What they really want is a general `deep_to()` function that operates on arbitrary Python data structures. This is advantageous to users because they can use it on any Python data structure. This is advantageous to developers and maintainers of PyTorch and Pyro because they can avoid interface bloat. See [jax.device_put()](https://jax.readthedocs.io/en/latest/jax.html#jax.device_put) for similar functionality in JAX.

This PR implements a `deep_to()` helper based on `copy.deepcopy()`, the same trick used in #2599.

## Tested
- [x] added unit tests

### DIFF
--- a/pyro/distributions/util.py
+++ b/pyro/distributions/util.py
@@ -140,8 +140,10 @@ class _DeepToMemo(dict):
         if result is default:
             # Assume key is the id of another object, and look up that object.
             old = ctypes.cast(key, ctypes.py_object).value
-            if isinstance(old, (torch.Tensor, torch.nn.Module)):
+            if isinstance(old, torch.Tensor):
                 self[key] = result = old.to(*self._to_args, **self._to_kwargs)
+            elif isinstance(old, torch.nn.Module):
+                old.to(*self._to_args, **self._to_kwargs)
         return result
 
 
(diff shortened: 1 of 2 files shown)


pyro/distributions/util.py
```diff
@@ -129,6 +129,38 @@ def detach(obj):
...
+        super().__init__()
+        self._to_args = to_args
+        self._to_kwargs = to_kwargs
+
+    def get(self, key, default=None):
+        result = super().get(key, default)
+        if result is default:
+            # Assume key is the id of another object, and look up that object.
+            old = ctypes.cast(key, ctypes.py_object).value
+            if isinstance(old, (torch.Tensor, torch.nn.Module)):
```
eb8680, line 143: Can you add a test exercising the `nn.Module` path? Also, will this work as expected on `PyroModule`s with `PyroSample`s and `PyroParam`s?

eb8680 (commented):
This will be really useful!

pyro/distributions/util.py
fritzo: Hmm I believe distribution attributes of modules will be bypassed, since we assume `Module` objects handle themselves via `torch.nn.Module.to()`. I'll have to think more about this...

fritzo: OK, I've added tests for `nn.Module`, `PyroModule`, `PyroParam`, and `PyroSample`. The only way I could get this to work was to perform the behavior out-of-place, by deep-copying the `nn.Module`. IMO this is the best we can do, given PyTorch's early design choices about behavior of `.to()` on tensors versus modules 🤷

eb8680 (approved):
LGTM with new tests

PR #2925, opened 2021-09-08
Enable loading optimizer to a specific device

sjfleming:
While playing with saving and loading checkpoint files, I ended up saving a checkpoint on a CUDA-enabled machine and then trying to load it on a CPU-only device.
This is normally made possible by using `torch.load()`'s `map_location` parameter.

While the param_store loader has the ability to specify a `map_location`
https://github.com/pyro-ppl/pyro/blob/c8dc40a75cc4ff1f43c6ff9178d91c08155d7973/pyro/params/param_store.py#L281-L300

the loader for pyro.optim does not:
https://github.com/pyro-ppl/pyro/blob/c8dc40a75cc4ff1f43c6ff9178d91c08155d7973/pyro/optim/optim.py#L179-L188

I would propose to add the `map_location` parameter to pyro.optim's loader so that we can completely achieve CPU-loading of a CUDA-saved checkpoint.

### DIFF
--- a/pyro/optim/optim.py
+++ b/pyro/optim/optim.py
@@ -176,15 +176,17 @@ class PyroOptim:
         with open(filename, "wb") as output_file:
             torch.save(self.get_state(), output_file)
 
-    def load(self, filename: str) -> None:
+    def load(self, filename: str, map_location: Optional[str] = None) -> None:
         """
         :param filename: file name to load from
         :type filename: str
+        :param map_location: torch.load() map_location parameter
+        :type map_location: str
 
         Load optimizer state from disk
         """
         with open(filename, "rb") as input_file:
-            state = torch.load(input_file)
+            state = torch.load(input_file, map_location=map_location)
         self.set_state(state)
 
     def _get_optim(self, param: Union[Iterable[Tensor], Iterable[Dict[Any, Any]]]):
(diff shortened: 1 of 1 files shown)


pyro/optim/optim.py
```diff
@@ -176,15 +176,17 @@ def save(self, filename: str) -> None:
         with open(filename, "wb") as output_file:
             torch.save(self.get_state(), output_file)
 
-    def load(self, filename: str) -> None:
+    def load(self, filename: str, map_location: Optional[str] = None) -> None:
```
fritzo, line 179: It looks like `map_location` can be [more general](https://pytorch.org/docs/stable/_modules/torch/serialization.html#load) than a string, so I'd recommend omitting a type hint here. (But thanks for adding a type hint, we do try to add type hints to new code, this is just a particularly flexible case).

fritzo (changes requested):
Thanks for fixing this, I have only one minor comment.

pyro/optim/optim.py
sjfleming: Okay, I will make that change

sjfleming: Wow, you're right, `map_location` can be so many things!  :)

fritzo (approved):

PR #2927, opened 2021-09-09
Fix Cholesky error handling

dme65:
https://github.com/pytorch/pytorch/pull/63864 changes the error messages for Cholesky errors in PyTorch. Pyro relies on `"singular U"` being part of the error message, which isn't the case anymore on PyTorch master. 

Note that https://github.com/pytorch/pytorch/pull/63864 also introduces a bunch of new error messages for singular solves, eigendecompositions, etc. that Pyro may also rely so we may need to fix those as well.

### DIFF
--- a/pyro/ops/integrator.py
+++ b/pyro/ops/integrator.py
@@ -76,7 +76,7 @@ def potential_grad(potential_fn, z):
         potential_energy = potential_fn(z)
     # deal with singular matrices
     except RuntimeError as e:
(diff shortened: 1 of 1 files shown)


dme65:
cc: @neerajprad, @jpchen

fritzo (approved):

fritzo:
Thanks for the fix, @dme65!

PR #2929, opened 2021-09-13
Implement AutoGaussian guide

fritzo:
Addresses #2813 
Replaces #2926 #2928 
See [design doc](https://docs.google.com/document/d/1MVfG0po7M5UB2mNM2In6iKKq_OYNpa_I-FnJncUgXPI)

## Tasks
- [x] Parametrize sparse precision matrices. This is statistically optimal.
- [x] Implement "dense" backend. This is fully general but computationally expensive.
- [x] Implement "funsor" backend. This is asymptotically cheaper than the dense backend but has two restrictions (from TVE):
    - For any dependency `u --> d`, `plates(u) <= plates(d)`.
    - For any pair of transitive dependencies `u1 --> d <-- u2`, either `plates(u1) <= plates(u2) <= plates(d)` or `plates(u2) <= plates(u1) <= plates(d)`.

After this first implementation, I plan to switch to [forward_filter_backward_precondition()](https://github.com/pyro-ppl/funsor/pull/553) and either [compile or trace](https://github.com/pyro-ppl/funsor/pull/557) the funsor computation to reduce overhead.

## Tested
- [x] added to existing autoguide tests
- [x] smoke test on pyro-cov model variants
- [x] test equality of elbo to `AutoMultivariateNormal` (on a gaussian model)
- [x] test on lots of dependency structures

PR #2933, opened 2021-09-21
Fix get_dependencies(), add poutine.get_plates()

fritzo:
Blocking #2929 

This PR adds some fixes and machinery needed for #2929, aiming to simplify that growing PR.
* makes a number of fixes to `get_inspect()`, adding regression tests:
    * support for `pyro.factor()` statements
    * support for discrete variables
    * support for use inside `torch.no_grad()`, as in autoguide init
* adds a `poutine.get_plates()` and adds unit tests (also adds tests for `poutine.get_mask()`)
* fixes a typo in the `Logistic` distribution


eb8680 (approved):
Looks good

PR #2934, opened 2021-09-22
Fix multi-sample tests in contribfunsor by scaling samples manually

eb8680:
The Funsor pull request https://github.com/pyro-ppl/funsor/pull/549 changed the semantics of `Funsor.sample` so that it no longer rescales its result by the number of particles, which broke some TMC tests in `pyro.contrib.funsor`.

This PR performs the sample rescaling manually within Pyro, fixes the broken tests and pins Pyro's Funsor dependency to the `master` commit merging the above Funsor PR.


fritzo:
let me see if i can silence the coveralls failure

fritzo (approved):

Issue #2924, opened 2021-09-06
[FR] add an alternative parameterization for lower triangular cholesky factors


martinjankowiak:
add an alternative parameterization for lower triangular cholesky factors and consider using this in `AutoMultivariateNormal`:

instead of parameterizing a lower cholesky factor as an unconstrained strictly lower triangular piece and a positive diagonal we instead parameterize as

`L = unit_scale_tril @ scale_diag` 

where `unit_scale_tril` is lower triangular with ones along the diagonal and `scale_diag` is a positive diagonal matrix.

for more details see the [corresponding NumPyro PR](https://github.com/pyro-ppl/numpyro/pull/1146)

fritzo:
I agree we should improve the parametrization of `AutoNormal`. We did something similar to `AutoLowRankMultivariateNormal` #2127, and indeed `AutoLowRankMultivariateNormal` now works much better than `AutoMultivariateNormal`.

fritzo:
Note I've been using a similar overparametrization in the `AutoStructured` guide, since the raw MVN via `scale_tril` performed so poorly:
https://github.com/pyro-ppl/pyro/blob/c8dc40a75cc4ff1f43c6ff9178d91c08155d7973/pyro/infer/autoguide/guides.py#L1564-L1573

martinjankowiak:
interesting. note this isn't an overparametrization though: the diagonal on `unit_scale_tril` is all ones

PR #2946, opened 2021-10-14
Learn white_vec parameter of AutoGaussian guide

fritzo:
Addresses #2813 
~~Blocked by https://github.com/pyro-ppl/funsor/pull/553~~

This adds learnable parameters `white_vec` to each Gaussian factor in the AutoGaussian guide, so the total set of factors is: `loc`, `scale`, `white_vec`, and `prec_sqrt`. The parameters `(white_vec, prec_sqrt)` are important because they are closer to the natural parameters than `(loc, scale)`, and should in theory lead to faster learning.  The `scale` parameter is important because (i) being on a log scale it can quickly adapt to large or small magnitudes, and (ii) it performs row-size scaling of the precision matrix and leads to faster learning (see #2127, #2924). It appears that the `loc` parameter is also required.

## Tested
- [x] strengthened a dispatch test
- [x] pass existing tests


fritzo:
@eb8680 this isn't super high priority, but it would be nice to get it in for the first release of `AutoGaussian`. Nothing has changed since we discussed this a month or two ago.

eb8680 (approved):
Looks good, merging

PR #2951, opened 2021-10-26
Support multivariate dependencies in AutoStructured

### LINKED ISSUE #2950: Support for multiple dependency/parent nodes in AutoStructured [feature request]
Often in hierarchical models, the prior value of a site depends on the value of several parent nodes via a specific computation (such as matrix multiplication).

For example in this case, the value of `w_sf` depend on `k_sr` and `q_rf`:
```
d_sg ~ Poisson(mu_sg)
mu_sg = sum_f w_sf * g_fg
w_sf ~ Gamma(10, 10 / mu_sf)
mu_sf = sum_r k_sr * q_rf
k_sr ~ Gamma(1, 1)
q_rf ~ Gamma(1, 1)
```
Would be great if it was possible to not only say that `w_sf` depend on `k_sr` and `q_rf` but specify a function that gives a hierarchical component of `w_sf`:
``` 
AutoStructured(dependencies={'w_sf': {['k_sr', 'q_rf']: lambda k_sr, q_rf: k_sr @ q_rf})
```
This probably also needs the site to be available in a non-flattened format.

fritzo:
Resolves #2950 

## Tasks
- [x] support multivariate dependencies
- [ ] support unflattened coordinates?  Can we make this backwards compatible?
- [ ] support constrained coordinates?  Is this a good idea?

## Tested
- [ ] unit tests in test_autoguide.py

### DIFF
--- a/pyro/infer/autoguide/structured.py
+++ b/pyro/infer/autoguide/structured.py
@@ -4,7 +4,7 @@
 from collections import OrderedDict, defaultdict
 from contextlib import ExitStack
 from types import SimpleNamespace
-from typing import Callable, Dict, Optional, Union
+from typing import Callable, Dict, Optional, Tuple, Union
 
 import torch
 from torch.distributions import biject_to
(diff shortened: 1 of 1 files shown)


fritzo:
Replaced by #2953

PR #2955, opened 2021-10-30
Implement AutoHierarchicalNormalMessenger

vitkl:
Based on #2953 

The mean-field posterior at any site is a transformed normal distribution, the mean of which depends on the value of that site given its dependencies in the model:

```python
loc = loc + transform.inv(prior.mean) * weight
```

Where the value of `prior.mean` is conditional on upstream sites in the model. This approach doesn't work for distributions that don't have the mean, as discussed here https://github.com/pyro-ppl/pyro/pull/2953#discussion_r739671482.

Currently, for the sites that don't have upstream sites, the guide is representing/learning deviation from the prior. Would it be useful to auto-detect sites that do not have upstream sites? Alternatively, hierarchical sites could be specified by the user. @fritzo WDYT?

## Tasks
- [x] add a section to infer.autoguide.rst as for `AutoNormalMessenger`
- [x] add import in `pyro/infer/autoguide/__init__.py` and add an entry in that file's `__all__`
- [x] add to tests in tests/infer/test_autoguide.py, following `AutoNormalMessenger`


pyro/infer/autoguide/effect.py
```diff
@@ -0,0 +1,396 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Callable, Dict, Union
```
fritzo, line 4: `from typing import Optional`

fritzo (commented):
Looks great @vitkl!  Before merging, could you also add docs, tests, and imports?  I'll add a task list to the PR description.

pyro/infer/autoguide/effect.py
```diff
@@ -0,0 +1,396 @@
...
+        except AttributeError:
+            pass
+
+        # Initialize.
+        with poutine.block(), torch.no_grad():
+            transform = biject_to(prior.support)
+            event_dim = transform.domain.event_dim
+            constrained = self.init_loc_fn({"name": name, "fn": prior}).detach()
+            unconstrained = transform.inv(constrained)
+            init_loc = self._remove_outer_plates(unconstrained, event_dim)
```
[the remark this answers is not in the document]
vitkl: This makes sense, thank you for suggesting!

pyro/infer/autoguide/effect.py
```diff
@@ -0,0 +1,396 @@
...
+        finally:
+            self._computing_median = False
+
+    def _get_posterior_median(self, name, prior):
+        transform = biject_to(prior.support)
+        loc, scale = self._get_params(name, prior)
+        return transform(loc)
+
+
+class AutoHierarchicalNormalMessenger(AutoNormalMessenger):
```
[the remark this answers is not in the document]
vitkl: I don't have many ideas except to remove `Messenger` from the name. Presumably `AutoHierarchicalNormal` will never be implemented using AutoNormal non-messenger style.

vitkl:
Would it make sense to constrain the weight between 0 and 1?
```python
loc = loc * (1 - weight) + transform.inv(prior.mean) * weight
```

fritzo:
> Would it make sense to constrain the weight between 0 and 1?

Yes, I think that would make sense. OTOH weights outside of that range should not result in NAN, and my intuition is that it should be easy to learn weight, so I'd opt for the cheaper solution of unconstrained weight.

pyro/infer/autoguide/effect.py
```diff
@@ -0,0 +1,409 @@
...
+        # Initialize.
+        with poutine.block(), torch.no_grad():
+            transform = biject_to(prior.support)
+            event_dim = transform.domain.event_dim
+            constrained = self.init_loc_fn({"name": name, "fn": prior}).detach()
+            unconstrained = transform.inv(constrained)
+            init_loc = self._remove_outer_plates(unconstrained, event_dim)
+            init_scale = torch.full_like(init_loc, self._init_scale)
+            # weight is a single value parameter
+            init_weight = torch.full((), self._init_weight)
```
fritzo, line 292: Re: plates, can you say more about your single-value design decision?  I can see why you might want to exclude `event_shape` from this, but [Gorinova et al. (2019)](https://arxiv.org/abs/1906.03028) observe that element-wise weights empirically outperformed single global weights in a similar context.  I can also see how you might want to amortize over a particular plate so as to support subsampling, but for that I think we could provide a more flexible mechanism such as a user specified `amortized_plates` kwarg to the `AutoMessenger.__init__()` (and then use the logic in `._remove_outer_plates()`), or even automatically amortize over any plate that has been subsampled.

docs/source/infer.autoguide.rst
```diff
@@ -125,6 +125,38 @@ AutoGaussian
...
+
+AutoNormalMessenger
+-------------------
+.. autoclass:: pyro.infer.autoguide.AutoNormalMessenger
+    :members:
+    :undoc-members:
+    :member-order: bysource
+    :show-inheritance:
+
+AutoHierarchicalNormal
```
fritzo, line 144: Could you call this `AutoHierarchicalMessenger` or other `Auto*Messenger` to distinguish it from other `Auto*` guides?  I think it's important to add `Messenger` to the name because all these `GuideMessenger` guides require different inference machinery (`Effect_ELBO` rather than `Trace_ELBO`) and it's easiest to remind users "if it ends with `Messenger`, use `Effect_ELBO`"?  Sorry for flip flopping!

pyro/infer/autoguide/effect.py
vitkl: My intuition was that element-wise weights are over-parameterised because you can get the same loc by either adding `loc_independent` to `prior.mean` or by weighting `prior.mean`:

```python
loc = loc_independent + transform.inv(prior.mean) * weight
20 = 0 + 10 * 2
20 = 10 + 10 * 1
```

WDYT about adding several options that can be chosen by changing class attributes?
 1. Scalar weight
 2. Element-wise weights (same shape as init_loc). In this case, `weights` should be amortised exactly the same way as `loc` and `scales`.
 3. Exclude `event_shape`. I don't know how to do this.

docs/source/infer.autoguide.rst
vitkl: Ok, that makes sense too. I renamed it `AutoHierarchicalNormalMessenger` because it is a fully descriptive name.

fritzo:
@vitkl you should now be able to merge the `guide-messenger` branch and then the `dev` branch (I'd recommend doing so in that order to minimize merge conflicts due to our squash-merge policy)

pyro/infer/autoguide/effect.py
```diff
@@ -211,6 +211,153 @@ def _get_posterior_median(self, name, prior):
         return transform(loc)
 
 
+class AutoHierarchicalNormalMessenger(AutoNormalMessenger):
+    """
+    EXPERIMENTAL Automatic :class:`~pyro.infer.effect_elbo.GuideMessenger` ,
+    intended for use with :class:`~pyro.infer.effect_elbo.Effect_ELBO` or
```
fritzo, line 217: update docstring and check links (see other AutoMessengers for updated example)

pyro/infer/autoguide/effect.py
```diff
@@ -211,6 +211,153 @@ def _get_posterior_median(self, name, prior):
...
+    The mean-field posterior at any site is a transformed normal distribution,
+    the mean of which depends on the value of that site given its dependencies in the model:
+
+        loc = loc + transform.inv(prior.mean) * weight
+
+    Where the value of `prior.mean` is conditional on upstream sites in the model.
+    This approach doesn't work for distributions that don't have the mean.
+
+    Derived classes may override particular sites and use this simply as a
+    default, see AutoNormalMessenger documentation for example.
```
fritzo, line 229: add a sphinx link:
```
see :class:`AutoNormalMessenger` documentation for details
```

fritzo (commented):
Logic looks great, I have only minor comments about docs & tests.

Tests: could you make a second pass at `test_autoguide.py` and try adding `AutoHierarchicalNormalMessenger` to every test supported by `AutoNormalMessenger`? In particular `test_exact*` tests are good at catching math errors. You may need to add logic to avoid jit elbos in those tests, following `AutoRegressiveMessenger` logic.

vitkl:
Adding tests now. Everything seems to work (no need to avoid jit) except `test_subsample_model` and I don't think I understand the source of this error:
```python
___________________________________________________________ test_subsample_model_amortized[AutoHierarchicalNormalMessenger] ___________________________________________________________

auto_class = <class 'pyro.infer.autoguide.effect.AutoHierarchicalNormalMessenger'>

    @pytest.mark.parametrize(
        "auto_class",
        [
            AutoNormalMessenger,
            AutoHierarchicalNormalMessenger,
            AutoRegressiveMessenger,
        ],
    )
    def test_subsample_model_amortized(auto_class):
        def model(x, y=None, batch_size=None):
            loc = pyro.param("loc", lambda: torch.tensor(0.0))
            scale = pyro.param(
                "scale", lambda: torch.tensor(1.0), constraint=constraints.positive
            )
            with pyro.plate("batch", len(x), subsample_size=batch_size):
                batch_x = pyro.subsample(x, event_dim=0)
                batch_y = pyro.subsample(y, event_dim=0) if y is not None else None
                mean = loc + scale * batch_x
                sigma = pyro.sample("sigma", dist.LogNormal(0.0, 1.0))
                return pyro.sample("obs", dist.Normal(mean, sigma), obs=batch_y)

        guide1 = auto_class(model)
        guide2 = auto_class(model, amortized_plates=("batch",))

        full_size = 50
        batch_size = 20
        pyro.set_rng_seed(123456789)
        x = torch.randn(full_size)
        with torch.no_grad():
            y = model(x)
        assert y.shape == x.shape

        for guide in guide1, guide2:
            pyro.get_param_store().clear()
            pyro.set_rng_seed(123456789)
            svi = SVI(model, guide, Adam({"lr": 0.02}), Trace_ELBO())
            for step in range(5):
>               svi.step(x, y, batch_size=batch_size)

tests/infer/test_autoguide.py:1131:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pyro/infer/svi.py:145: in step
pyro/infer/trace_elbo.py:140: in loss_and_grads
    for model_trace, guide_trace in self._get_traces(model, guide, args, kwargs):
pyro/infer/elbo.py:182: in _get_traces
    yield self._get_trace(model, guide, args, kwargs)
pyro/infer/trace_elbo.py:57: in _get_trace
    model_trace, guide_trace = get_importance_trace(
pyro/infer/enum.py:57: in get_importance_trace
    guide(*args, **kwargs)
pyro/nn/module.py:636: in cached_fn
    return fn(self, *args, **kwargs)
pyro/infer/autoguide/effect.py:46: in __call__
    return super().__call__(*args, **kwargs)
pyro/poutine/guide.py:45: in __call__
    self.model(*args, **kwargs)
tests/infer/test_autoguide.py:1112: in model
    sigma = pyro.sample("sigma", dist.LogNormal(0.0, 1.0))
pyro/primitives.py:164: in sample
    apply_stack(msg)
pyro/poutine/runtime.py:213: in apply_stack
    frame._process_message(msg)
pyro/poutine/messenger.py:154: in _process_message
    return method(msg)
pyro/poutine/guide.py:62: in _pyro_sample
    posterior = self.get_posterior(msg["name"], prior)
pyro/infer/autoguide/effect.py:276: in get_posterior
    loc, scale, weight = self._get_params(name, prior)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = AutoHierarchicalNormalMessenger(), name = 'sigma', prior = LogNormal()

    def _get_params(self, name: str, prior: Distribution):
        try:
            loc = deep_getattr(self.locs, name)
            scale = deep_getattr(self.scales, name)
            if (self._hierarchical_sites is None) or (name in self._hierarchical_sites):
                weight = deep_getattr(self.weights, name)
                return loc, scale, weight
            else:
                return loc, scale
        except AttributeError:
            pass

        # Initialize.
        with torch.no_grad():
            transform = biject_to(prior.support)
            event_dim = transform.domain.event_dim
            constrained = self.init_loc_fn({"name": name, "fn": prior}).detach()
            unconstrained = transform.inv(constrained)
            init_loc = self._remove_outer_plates(unconstrained, event_dim)
            if self.weight_type == "scalar":
                # weight is a single value parameter
                init_weight = torch.full((), self._init_weight)
            if self.weight_type == "element-wise":
                # weight is element-wise
                init_weight = torch.full_like(init_loc, self._init_weight)
            # if site is hierarchical substract contribution of dependencies from init_loc
            if (self._hierarchical_sites is None) or (name in self._hierarchical_sites):
>               init_loc = init_loc - init_weight * transform.inv(prior.mean)
E               RuntimeError: The size of tensor a (50) must match the size of tensor b (20) at non-singleton dimension 0

pyro/infer/autoguide/effect.py:315: RuntimeError
```

I don't understand why this is happening because the logic is the same as `AutoNormalMessenger` and the new guide works for cell2location model.

fritzo:
@vitkl can you push so we can inspect in a debugger? happy to pair code through the error

fritzo:
Aha, I think you just need to `self._remove_outer_plates(prior.mean)`:
```diff
- init_loc = init_loc - init_weight * transform.inv(prior.mean)
+ mean = self._remove_outer_plates(transform.inv(prior.mean))
+ init_loc = init_loc - init_weight * mean
```

vitkl:
I will add the code with other changes soon. I see - but why should the
fail for non-amortised test? In that test, the subsampling plate is not
supposed to be removed, right?

fritzo:
> subsampling plate is not supposed to be removed, right?

`._remove_outer_plates()` does a little more than subsampling now, it also fixes subsampled plates. Feel free to rename it to `._adjust_plates()` or something.

vitkl:
I added the changes you requested and renamed `._adjust_plates()` - but I don't know why the docs test fails.

pyro/infer/autoguide/effect.py
```diff
@@ -312,7 +313,8 @@ def _get_params(self, name: str, prior: Distribution):
                 init_weight = torch.full_like(init_loc, self._init_weight)
             # if site is hierarchical substract contribution of dependencies from init_loc
             if (self._hierarchical_sites is None) or (name in self._hierarchical_sites):
-                init_loc = init_loc - init_weight * transform.inv(prior.mean)
+                init_prior_mean = self._adjust_plates(prior.mean, event_dim)
+                init_loc = init_loc - init_weight * transform.inv(init_prior_mean)
```
fritzo, line 317: you'll need to transform before applying `self._adjust_plates` because the transform could introduce batch dimensions e.g. for an `AffineTransform`

pyro/infer/autoguide/effect.py
```diff
@@ -213,20 +213,20 @@ def _get_posterior_median(self, name, prior):
 
 class AutoHierarchicalNormalMessenger(AutoNormalMessenger):
     """
-    EXPERIMENTAL Automatic :class:`~pyro.infer.effect_elbo.GuideMessenger` ,
-    intended for use with :class:`~pyro.infer.effect_elbo.Effect_ELBO` or
-    similar.
+    :class:`AutoMessenger` with mean-field normal posterior conditional on all dependencies.
 
     The mean-field posterior at any site is a transformed normal distribution,
     the mean of which depends on the value of that site given its dependencies in the model:
```
fritzo, line 219: double colon before a code block:
```
the mean of ... model::

    loc_total = ...
```

pyro/infer/autoguide/effect.py
```diff
@@ -213,20 +213,20 @@ def _get_posterior_median(self, name, prior):
...
 
-        loc = loc + transform.inv(prior.mean) * weight
+        loc_total = loc + transform.inv(prior.mean) * weight
 
-    Where the value of `prior.mean` is conditional on upstream sites in the model.
+    Where the value of `prior.mean` is conditional on upstream sites in the model,
```
fritzo, line 223: double ticks around code snippets
```
``prior.mean``
```

fritzo:
Hmm, looks like docs build is failing due to a failure to build the `lap` wheel, I'll look into it...

pyro/infer/autoguide/effect.py
vitkl: Sorry, still learning how to use pandoc syntax

fritzo:
Blocked by https://github.com/pypa/setuptools/issues/2849

fritzo (approved):
Looks great!  Thanks for adding the new tests.

Issue #2957, opened 2021-11-01
[FR] Backport numpyro.render_model()


fritzo:
Replaces #1502

The old `pyro.util.save_visualization()` has languished, but users seem to like [numpyro.render_model](http://num.pyro.ai/en/stable/tutorials/model_rendering.html).  Let's port this functionality back to Pyro.

## Tasks
- [x] Support discrete latent variables in [pyro.inspect.get_dependencies()](https://docs.pyro.ai/en/latest/infer.util.html#pyro.infer.inspect.get_dependencies). @ordabayevy created a great solution to this in funsor's [ProvenanceTensor](https://github.com/pyro-ppl/funsor/blob/a6eeef75875da195d645396de4c0f4331e3c71b1/funsor/torch/provenance.py#L7); it should be easy to refactor `get_dependencies()` to use this rather than autograd to determine dependency structure.  #2959 
- [ ] Implement `pyro.render_model()` based on [pyro.inspect.get_dependencies()](https://docs.pyro.ai/en/latest/infer.util.html#pyro.infer.inspect.get_dependencies) #2962

PR #2959, opened 2021-11-03
Support discrete variables in get_dependencies() via ProvenanceTensor

fritzo:
Addresses #2957 
pair coded with @ordabayevy 

This refactors `get_dependencies()` to use @ordabayevy's `ProvenanceTensor` instead of `torch.autograd.grad`. This results in a cubic -> linear speedup, and support for tracking dependencies through discrete latent variables and through `.detach()` (fixing a previously known failure mode).

## Tested
- [x] fixed a known failure due to `.detach()`
- [x] added a new discrete test


pyro/ops/provenance.py
```diff
@@ -0,0 +1,71 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+
+
+class ProvenanceTensor(torch.Tensor):
+    """
+    Provenance tracking implementation in Pytorch.
```
eb8680, line 9: Maybe add slightly more documentation here explaining what extra metadata is attached and a reference to 'Wingate, David, et al. "Nonstandard Interpretations of Probabilistic Programs for Efficient Inference." NIPS. 2011'?

eb8680 (commented):
Looks reasonable. Would this ever interact with the JIT in the process of autoguide initialization? If so, is that exercised by any existing tests of e.g. `AutoStructured`?

Also, are there any noteworthy edge cases of `__torch_function__` that would produce incorrect results or cause an error? For example, does it work with most built-in `nn.Module`s (e.g. `nn.Linear`) in `torch.nn`? Maybe add a test case with the Bayesian regression example to make sure it plays nicely with `nn.Linear` and `PyroModule`?

pyro/ops/provenance.py
fritzo: done

fritzo:
> Would this ever interact with the JIT in the process of autoguide initialization?

`get_dependencies()` is called during autoguide initialization in `AutoStructured`  `AutoGaussian` which are already heavily tested with the jit and with nn.Modules in tests/infer/test_autoguide.py and tests/infer/autoguide/test_gaussian.py

> Also, are there any noteworthy edge cases of __torch_function__ that would produce incorrect results or cause an error?

@ordabayevy said the `.__torch_function__()` code is young. I expect there will be edge cases, and we can address them as they arise, but there are no know failures.

> does it work with most built-in nn.Modules ... ?

The `__torch_function__` layer is below and independent of the `nn.Module` layer. This is already tested in test_autoguide.py

fritzo:
Blocked by https://github.com/pypa/setuptools/issues/2849

eb8680 (approved):
Lgtm

fritzo:
Thanks for reviewing @eb8680 !

PR #2960, opened 2021-11-04
Work around setuptools breakage

fritzo:
Works around https://github.com/pypa/setuptools/issues/2849
Blocking #2955, #2959


fehiepsi (approved):

fritzo:
Thanks, @fehiepsi !

PR #2963, opened 2021-11-09
Change coordinatization of AutoMultivariateNormal

fritzo:
Addresses #2924 

This adds a row-wise `.scale` parameter to `AutoLowRankMultivariateNormal`. The resulting overparametrized triple (loc,scale,scale_tril) has been observed to speed up learning in `AutoLowRankMultivariateNormal` and in `AutoStructured`.

This also fixes `.get_base_dist()` and adds tests of use in `NeuTraReparam`, and moves two tests from tests/contrib/autoguide to tests/infer/autoguide.

## Tested
- [x] `.get_base_dist()` covered by a new regression test


martinjankowiak:
is there any reason to think that this works better than the non-overparameterized version i implemented in https://github.com/pyro-ppl/numpyro/pull/1146? i would be mildly surprised if that were the case

fritzo:
> is there any reason to think that this works better than the non-overparameterized version

I would expect this PR to work about the same as the non-overreparametrized version, but with only a fraction of the coding effort.  Feel free to implement the fancier version in a subsequent PR.

pyro/infer/autoguide/guides.py
```diff
@@ -874,27 +875,31 @@ def _setup_prototype(self, *args, **kwargs):
...
         self.scale_tril = PyroParam(
-            eye_like(self.loc, self.latent_dim) * self._init_scale,
-            self.scale_tril_constraint,
+            eye_like(self.loc, self.latent_dim), self.scale_tril_constraint
         )
 
     def get_base_dist(self):
         return dist.Normal(
-            torch.zeros_like(self.loc), torch.zeros_like(self.loc)
+            torch.zeros_like(self.loc), torch.ones_like(self.loc)
```
fehiepsi, line 887: Hmm, what a bug!? It is surprised to me that we didn't catch this earlier.

fritzo: well, apparently nobody uses `AutoMultivariateNormal` with `NeuTraReparam`.  My motivation for this PR is to write a tutorial about autoguides, so hopefully they'll get more exposure

pyro/infer/autoguide/guides.py
```diff
@@ -874,27 +875,31 @@ def _setup_prototype(self, *args, **kwargs):
...
         """
         Returns a MultivariateNormal posterior distribution.
         """
-        return dist.MultivariateNormal(self.loc, scale_tril=self.scale_tril)
+        scale_tril = self.scale[..., None] * self.scale_tril
+        return dist.MultivariateNormal(self.loc, scale_tril=scale_tril)
 
     def _loc_scale(self, *args, **kwargs):
-        return self.loc, self.scale_tril.diag()
+        return self.loc, self.scale * self.scale_tril.diag()
```
martinjankowiak, line 902: does `scale` need an unsqueeze here?

pyro/infer/autoguide/guides.py
```diff
@@ -862,6 +862,7 @@ class AutoMultivariateNormal(AutoContinuous):
         (unconstrained transformed) latent variable.
     """
 
+    scale_constraint = constraints.softplus_positive
     scale_tril_constraint = constraints.softplus_lower_cholesky
```
martinjankowiak, line 866: i'm confused. it looks like we rename torch's `lower_cholesky to `softplus_lower_cholesky` even though the former uses a an `exp` transform for the diagonal?  or am i missing something?

pyro/infer/autoguide/guides.py
fritzo: nope, `scale` is already the correct shape

pyro/infer/autoguide/guides.py
fritzo: [softplus_lower_cholesky](https://github.com/pyro-ppl/pyro/blob/07193b8a21eb5ff4c117097fed470024c6a794e1/pyro/distributions/constraints.py#L110) is @vitkl's more numerically stable constraint, which dispatches to a [SoftplusLowerCholeskyTransform](https://github.com/pyro-ppl/pyro/blob/07193b8a21eb5ff4c117097fed470024c6a794e1/pyro/distributions/transforms/softplus.py#L36) which uses a `torch.nn.functional.softplus()` under the hood. @vitkl found this to be more numerically stable after extensive experiments. See #2823 and [supporting experiments](https://github.com/pyro-ppl/sandbox/tree/master/2021-03-softplus_scales).

pyro/infer/autoguide/guides.py
martinjankowiak: oh i see this is only for marginals

martinjankowiak (approved):

fritzo:
Thanks for reviewing @martinjankowiak!

PR #2964, opened 2021-11-09
Fix provenance of kwargs

fritzo:
This extends `ProvenanceTensor` to handle `kwargs`, which is needed in PyTorch 1.10 because `torch.nn.functional.linear`'s `bias` input is passed as a named `kwarg` rather than a positional `arg`.  To clarify the code I've also refactored to use singledispatch with two new functions `track_provenance()` and `extract_provenance()`.

## Tested
- [x] refactoring is covered by existing tests (test_inspect.py, test_autoguide.py, test_gaussian.py)
- [x] tested locally with PyTorch 1.10

cc @ordabayevy

### DIFF
--- a/pyro/ops/provenance.py
+++ b/pyro/ops/provenance.py
@@ -1,6 +1,9 @@
 # Copyright Contributors to the Pyro project.
 # SPDX-License-Identifier: Apache-2.0
 
+from functools import singledispatch
+from typing import Tuple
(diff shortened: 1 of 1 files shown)


eb8680 (approved):

fritzo:
Thanks for reviewing @eb8680!

PR #2966, opened 2021-11-10
Use dependencies in the AutoReparam strategy

fritzo:
This attempts to apply `LocScaleReparam` in `AutoReparam` only for latent variables whose prior distributions depends on upstream variables. In case the prior has fixed parameters, there is no need to reparametrize, and reparametrizing introduces unnecessary complexity and obfuscation.

This first attempt is blocked by nuances in the interaction of `get_dependencies()` with `AutoReparam()`, since the model's latent variables are determined only dynamically, but they want to use dependencies to determine strategy. The fix may involve subtler use of `ProvenanceTensor` 🤔

### DIFF
--- a/pyro/infer/reparam/strategies.py
+++ b/pyro/infer/reparam/strategies.py
@@ -8,14 +8,16 @@ accessed by name via ``poutine.reparam(config=name_of_strategy)`` .
 See :func:`~pyro.poutine.handlers.reparam` for usage.
 """
 
-from abc import ABC, abstractmethod
-from typing import Callable, Dict, Optional, Union
+from abc import abstractmethod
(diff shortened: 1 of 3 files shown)


pyro/infer/reparam/strategies.py
```diff
@@ -200,10 +219,6 @@ def _loc_scale_reparam(name, fn, centered):
     if not _is_unconstrained(fn.support):
         return
 
-    # TODO reparametrize only if parameters are variable. We might guess
-    # based on whether parameters are differentiable, .requires_grad. See
-    # https://github.com/pyro-ppl/pyro/pull/2824
-
```
fritzo, line 206: This PR resolves this TODO

PR #2972, opened 2021-11-27
add UnitLowerCholeskyTransform; change default parameterization for AutoMultivariateNormal

martinjankowiak:
addresses #2963 

questions for reviewers:
- did i put things in the right files?
- do i need additional tests?
- do i need to change any additional autoguide defaults?
- do i need to do anything on the doc side of things for `unit_lower_cholesky`?

### DIFF
--- a/docs/source/distributions.rst
+++ b/docs/source/distributions.rst
@@ -522,6 +522,13 @@ SoftplusLowerCholeskyTransform
     :undoc-members:
     :show-inheritance:
 
+UnitLowerCholeskyTransform
+--------------------------
+.. autoclass:: pyro.distributions.transforms.UnitLowerCholeskyTransform
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
 SoftplusTransform
 -----------------
 .. autoclass:: pyro.distributions.transforms.SoftplusTransform
--- a/docs/source/distributions.rst
+++ b/docs/source/distributions.rst
@@ -522,16 +522,16 @@ SoftplusLowerCholeskyTransform
     :undoc-members:
     :show-inheritance:
 
-UnitLowerCholeskyTransform
---------------------------
-.. autoclass:: pyro.distributions.transforms.UnitLowerCholeskyTransform
+SoftplusTransform
+-----------------
+.. autoclass:: pyro.distributions.transforms.SoftplusTransform
     :members:
     :undoc-members:
     :show-inheritance:
 
-SoftplusTransform
------------------
-.. autoclass:: pyro.distributions.transforms.SoftplusTransform
+UnitLowerCholeskyTransform
+--------------------------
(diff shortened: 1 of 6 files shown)


pyro/distributions/transforms/unit_cholesky.py
```diff
@@ -0,0 +1,32 @@
...
+    codomain = unit_lower_cholesky
+
+    def __eq__(self, other):
+        return isinstance(other, UnitLowerCholeskyTransform)
+
+    def _call(self, x):
+        return x.tril(-1) + torch.eye(x.size(-1), device=x.device, dtype=x.dtype)
+
+    def _inverse(self, y):
+        return y.tril(-1)
```
fritzo, line 27: It's cheaper to simply `return y`.

docs/source/distributions.rst
```diff
@@ -522,6 +522,13 @@ SoftplusLowerCholeskyTransform
     :undoc-members:
     :show-inheritance:
 
+UnitLowerCholeskyTransform
```
fritzo, line 525: It looks like we're trying to keep this file alphabetical, can you try to preserve that convention?

fritzo (changes requested):
LGTM except for minor comments.

> did i put things in the right files?
:+1: yes

> do i need additional tests?
Tests already look sufficient.

> do i need to change any additional autoguide defaults?
Nope, it looks like `AutoMultivariateNormal` is the only use of `lower_cholesky`.

> do i need to do anything on the doc side of things for unit_lower_cholesky?
I don't think so, but can you `make docs` and check that `unit_lower_cholesky` shows up?

pyro/distributions/transforms/unit_cholesky.py
martinjankowiak: done

docs/source/distributions.rst
martinjankowiak: done

martinjankowiak:
unfortunately i'm unable to build the docs:

```
Warning, treated as error:
autodoc: failed to import module 'multi_mnist' from module 'pyro.contrib.examples'; the following exception was raised:
.........
  File "/Users/mjankowi/pyro/pyro/contrib/examples/util.py", line 15, in MNIST
    ] + datasets.MNIST.mirrors
AttributeError: type object 'MNIST' has no attribute 'mirrors'
```

tried to upgrade sphinx etc but that didn't help...

fritzo:
> i'm unable to build the docs:

Try upgrading torchvision?
```
pip install -U torch torchvision
```

martinjankowiak:
ah yeah torchvision was it. and with the addition to `__all__` i can confirm the docs build and the appropriate entry shows up

fritzo (approved):
:+1: Thanks for checking docs!

PR #2979, opened 2021-12-06
Add tutorial with simple SARS-CoV-2 lineage growth models

fritzo:
[open in colab](https://colab.research.google.com/github/pyro-ppl/pyro/blob/logistic-growth/tutorial/source/logistic-growth.ipynb)


fritzo (commented):
Looks great @martinjankowiak!  Comments for latest draft:
- model progression looks great
- it'd be nice to add a final model regressing against mutations
- `pyro.render_model()` gets the plates wrong, we should either fix it or remove the plots
- it'd be nice to add `%%time` magics to each svi cell
- `Alpha` denotes `B.1.1.7` + `Q.*` descendents; `Delta` denotes `B.1.617.2` + `AY.*` descendents.
- we should avoid ever plotting absolute counts, and plot only fractions of strains (this is because genetic sequence collection totals have no connection with case counts, and readers are liable to be confused)

PR #2982, opened 2021-12-07
Add logistic growth tutorial

martinjankowiak:
[open in colab](https://colab.research.google.com/github/pyro-ppl/pyro/blob/add-logistic-growth-tut/tutorial/source/logistic-growth.ipynb)


martinjankowiak:
our notebooks assume that pyro is installed which is not the case for colab. do we just make a second copy of the notebook that has the `pip install` at the top?

martinjankowiak:
- typo: +to => "Next let's extract sublineages corresponding two of the"
- arguably `rate_scale` isn't doing much in the last model since there are only two lineages
- typo: generalization section => "usepyro.plate"

generally looks great and i think this will be a good fit for the workshop

fritzo (approved):
Thanks for making this @martinjankowiak!

martinjankowiak:
thank for your substantial improvements!

PR #2984, opened 2021-12-10
remove scanvi.rst and fix docs conflict

martinjankowiak:
confirmed doc build is successful.


fritzo (approved):
Thanks for fixing this!

PR #2988, opened 2021-12-11
Implementation of quantiles for messenger guides [WIP]

vitkl:
This PR contributes the implementation of quantiles for messenger guides - which resembles how this is done in AutoNormal. I see certain errors that I don't know how to address - namely, the messenger guide tries to expand returned quantiles `new.v = self.v.expand(batch_shape + self.event_shape)`. This can potentially be addressed like this:

```python
def quantiles(self, quantiles, *args, **kwargs):
        self._computing_quantiles = True
        try:
            for i, q in enumerate(quantiles):
                self._quantile_values = q
                if
                _quantiles = {k: [v] for k, v in _quantiles.items()}

            return 
        finally:
            self._computing_quantiles = False
```

### DIFF
--- a/pyro/infer/autoguide/effect.py
+++ b/pyro/infer/autoguide/effect.py
@@ -157,12 +157,16 @@ class AutoNormalMessenger(AutoMessenger):
         self.init_loc_fn = init_loc_fn
         self._init_scale = init_scale
         self._computing_median = False
+        self._computing_quantiles = False
+        self._quantile_values = None
 
     def get_posterior(
         self, name: str, prior: Distribution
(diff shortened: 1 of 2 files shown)


fritzo (commented):
Thanks for sketching this @vitkl. The approach seems reasonable, but I'm unsure why the shapes are erroring.

vitkl:
Is it possible to restart the checks? Maybe the issue resolved itself with new changes to pyro/pytorch?

fritzo:
not sure how to restart in github actions, but you could merge Pyro dev branch and push, that should rerun ci

Issue #2970, opened 2021-11-18
🐛 LocScaleReparam and enumeration with NUTS


pietrolesci:
## Context

Hi there,

While trying to reproduce the [annotators.py](https://github.com/pyro-ppl/numpyro/blob/master/examples/annotation.py) example - originally written in numpyro - in pyro, I've come across what might be a bug. Discussion on pyro forum [here](https://forum.pyro.ai/t/issue-when-combining-reparametrization-and-automatic-enumeration/3561). As suggested by @fehiepsi, I am opening this issue.

## Description

`LogScaleReparam` seems to create problems when coupled with the automatic enumeration of discrete variables in the context of MCMC inference with NUTS. The same issue is not present when the reparametrization is done manually.
Below I report the code to reproduce the issue.

## Code

```python
import pyro
import pyro.distributions as dist
import torch
import torch.nn.functional as F
from pyro.infer import MCMC, NUTS
from pyro.infer.reparam import LocScaleReparam
from pyro.ops.indexing import Vindex
from pyro.poutine import reparam


def get_data():
    """
    :return: a tuple of annotator indices and class indices. The first term has shape
        `num_positions` whose entries take values from `0` to `num_annotators - 1`.
        The second term has shape `num_items x num_positions` whose entries take values
        from `0` to `num_classes - 1`.
    """
    # NB: the first annotator assessed each item 3 times
    positions = torch.tensor([1, 1, 1, 2, 3, 4, 5])
    annotations = torch.tensor(
        [
            [1, 1, 1, 1, 1, 1, 1],
            [3, 3, 3, 4, 3, 3, 4],
            [1, 1, 2, 2, 1, 2, 2],
            [2, 2, 2, 3, 1, 2, 1],
            [2, 2, 2, 3, 2, 2, 2],
            [2, 2, 2, 3, 3, 2, 2],
        ]
    )
    # we minus 1 because in Python, the first index is 0
    return positions - 1, annotations - 1


def hierarchical_dawid_skene(positions: torch.Tensor, annotations: torch.Tensor) -> None:
    """
    This model corresponds to the plate diagram in Figure 4 of reference [1].
    """
    num_annotators = positions.unique().numel()
    num_classes = annotations.unique().numel()
    num_items, num_positions = annotations.shape
    # print(f"{num_classes=}, {num_annotators=}, {num_items=}, {num_positions=}")

    with pyro.plate("class", num_classes):
[...]

PR #2989, opened 2021-12-11
Add a failing test for using init messenger with reparam inside plate

fehiepsi:
Added a test to catch the issue in #2970.

To resolve the issue, we need to apply `reparam` outside of plate statements. It would be nice to add a warning to help the users fix their model, but I'm not sure where we need to add such check.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ tutorial: FORCE
 
 lint: FORCE
 	flake8
-	black --extend-exclude=\.ipynb --check .
+	black --extend-exclude=\.ipynb --check . || black --exclude=\.ipynb --check .
 	isort --check .
 	python scripts/update_headers.py --check
 	mypy pyro
@@ -30,7 +30,7 @@ license: FORCE
 	python scripts/update_headers.py
 
 format: license FORCE
-	black --extend-exclude=\.ipynb .
+	black --extend-exclude=\.ipynb --check . || black --exclude=\.ipynb --check .
 	isort .
 
 version: FORCE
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -30,7 +30,7 @@ license: FORCE
 	python scripts/update_headers.py
 
 format: license FORCE
-	black --extend-exclude=\.ipynb .
+	black --extend-exclude=\.ipynb --check . || black --exclude=\.ipynb --check .
```
fehiepsi, line 33: In the new version, there is no flag `--extend-exclude`. So I added this additional commend in case the former failed.

fritzo: I think it's better to pin black version than to use this workaround. The reason is that, if the first `black --extend-exclude...` legitimately fails with an important error message, that message will then be converted to an `unknown command line option` error message, confusing devs.

Makefile
```diff
@@ -19,7 +19,7 @@ tutorial: FORCE
 
 lint: FORCE
 	flake8
-	black --extend-exclude=\.ipynb --check .
+	black --extend-exclude=\.ipynb --check . || black --exclude=\.ipynb --check .
```
fritzo, line 22: ditto, it's better to pin black version

fritzo (changes requested):
LGTM on the test. We may want to add an issue to discuss and point the xfail to that issue.

fritzo:
How about instead of pinning black we simply list files to be checked?
```sh
black *.py pyro examples tests scripts profiler
```

Makefile
fehiepsi: That's a good point! I didn't think about what happens when there are errors.

fehiepsi:
Thanks for the suggestions, @fritzo! I have addressed your comments.

fritzo (approved):

PR #3004, opened 2022-01-12
Clarified comments in `minipyro.py` to fix #3003

### LINKED ISSUE #3003: minipyro documentation clarification.
I really like that minipyro is in here for those interested in the internals of pyro. Super awesome!

Some of the wording in the module, particularly regarding `PYRO_STACK`, is confusing to me, though.

For example:
https://github.com/pyro-ppl/pyro/blob/dc7b96471b3e2dd32f6f39eb2e10ccedc2a4a8a3/pyro/contrib/minipyro.py#L48

If we run:
```python
trace(replay(model, guide_trace)).get_trace(...)
```

We end up (somewhere during the execution of the above line) with a `PYRO_STACK` with the first item referencing `trace`, and the second item referencing `replay`. But when we hit `apply_stack`, because of the reverse order in which the handlers are applied, `replay` is applied first. So, the comments should read, `# Handlers *later* in the PYRO_STACK are applied first.`, correct?

Similarly, in
https://github.com/pyro-ppl/pyro/blob/dc7b96471b3e2dd32f6f39eb2e10ccedc2a4a8a3/pyro/contrib/minipyro.py#L164

Shouldn't 
> When a Messenger sets the "stop" field of a message,
> it prevents any Messengers above it on the stack from being applied.

really be:
> When a Messenger sets the "stop" field of a message,
> it prevents any Messengers *below* it on the stack from being applied.

And
> A Messenger that sets msg["stop"] == True also prevents application
> of postprocess_message by Messengers above it on the stack
> via the pointer variable from the process_message loop

should really be:
> A Messenger that sets msg["stop"] == True also prevents application
> of postprocess_message by Messengers *below* it on the stack
> via the pointer variable from the process_message loop

I could be convinced to interpret all this otherwise, but I guess this was my default interpretation. Am I way off-base?

luiarthur:
Applied changes proposed in #3003.

### DIFF
--- a/pyro/contrib/minipyro.py
+++ b/pyro/contrib/minipyro.py
@@ -45,7 +45,7 @@ class Messenger:
         self.fn = fn
 
     # Effect handlers push themselves onto the PYRO_STACK.
-    # Handlers earlier in the PYRO_STACK are applied first.
+    # Handlers later in the PYRO_STACK are applied first.
     def __enter__(self):
         PYRO_STACK.append(self)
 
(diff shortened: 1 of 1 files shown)


pyro/contrib/minipyro.py
```diff
@@ -162,17 +162,19 @@ def __iter__(self):
 # apply_stack is called by pyro.sample and pyro.param.
 # It is responsible for applying each Messenger to each effectful operation.
 def apply_stack(msg):
+    # PYRO_STACK is reversed so that effect handlers later
```
fritzo, line 165: ditto: later -> higher?

fritzo (commented):
Thanks for submitting a PR!  Looks great, I just have one question about terminology

PR #3006, opened 2022-01-17
Implement Bayesian regression example from NumPyro in Pyro

arijc76:
This PR contributes the implementation of a Bayesian regression example / tutorial from NumPyro to Pyro.
There is one issue that I am not able to address yet - I have to run the below cell before running the cells for Model 2 and Model 3. I did not face this issue with the NumPyro tutorial.

```
# Run NUTS
kernel = NUTS(model)
num_samples = 2000
mcmc = MCMC(kernel, num_samples=num_samples, warmup_steps=200)
```
Please let me know what you think @eb8680


eb8680:
@arijc76 sorry I am so slow this week, great work! Here are some minor comments:

**Notebook**
- Could you use the same model function for SVI and MCMC, instead of having a separate `model_svi`?
- Could you add a bit of descriptive text in the section on SVI? Perhaps a sentence or two before each of cell 20, 21 and 22 explaining what each cell is about to do and what the results mean
- Typo: "For this we one of" -> "For this we use one of"

**Rendering**
- Could you add your notebook's filename to `tutorial/source/index.rst` under the "Other inference algorithms" header?
- Could you run `cd tutorial && make html` locally and check that everything in the notebook (e.g. math, images) is rendered correctly on your machine? The generated HTML files should appear in `tutorial/build/html/` - start a temporary local HTTP server in the `tutorial` directory (e.g. via `python -m http.server`), open `index.html` in a browser and check that your example appears under the "other inference algorithms" header on the sidebar, then click the link to make sure it works and scroll through the generated HTML of your notebook and look for obvious visual errors

eb8680:
> I have to run the below cell before running the cells for Model 2 and Model 3.

There may be some difference between the `MCMC` objects in Pyro and NumPyro. Could you try creating separate instances for each model rather than reusing the same one?

eb8680:
The custom `predict_fn` in your updated "Predictive Utility With Effect Handlers" section can be simplified - the `pyro.plate("samples", 2000)` you added plays the same role as Jax's `vmap`, so you should be able to write something like this that more directly follows the original structure:
```python
def predict(post_samples, model, *args, **kwargs):
    conditioned_model = poutine.condition(model, post_samples)
    model_trace = poutine.trace(conditioned_model).get_trace(*args, **kwargs)
    return model_trace.nodes["obs"]["value"]

def predict_fn(post_samples, model, *args, **kwargs):
    with pyro.plate("samples", num_samples):
        return predict(post_samples, model, *args, **kwargs)
```

arijc76:
> The custom `predict_fn` in your updated "Predictive Utility With Effect Handlers" section can be simplified - the `pyro.plate("samples", 2000)` you added plays the same role as Jax's `vmap` in this case, so you should be able to write something like this in cell 12 that more directly follows the original structure:
> 
> ```python
> def predict(post_samples, model, *args, **kwargs):
>     conditioned_model = poutine.condition(model, post_samples)
>     model_trace = poutine.trace(conditioned_model).get_trace(*args, **kwargs)
>     return model_trace.nodes["obs"]["value"]
> 
> def predict_fn(post_samples):
>     with pyro.plate("samples", num_samples):
>         return predict(post_samples, model, marriage=torch.tensor(dset.MarriageScaled.values, dtype=torch.float))
> ```

@eb8680 Thanks for this suggestion.

arijc76:
> * `cd tutorial && make html`

@eb8680 When I run `cd tutorial && make html`, the build stops at `waiting for workers...`. Then after some time I get the error  as shown below. Can you let me know what I need to do for this? Thanks.

```
Running Sphinx v4.4.0
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 78 source files that are out of date
updating environment: [new config] 78 added, 0 changed, 0 removed
reading sources... [100%] tensor_shapes .. working_memory                                                                     
waiting for workers...

Warning, treated as error:
/Users/arijeetchatterjee/Documents/github_personal_projects/pyro/tutorial/source/ss-vae.ipynb:7:Unexpected indentation.
make: *** [html] Error 2
```

eb8680:
> Can you let me know what I need to do for this?

I can't reproduce your error, but you can tell Sphinx not to treat warnings as errors by overriding the `SPHINXOPTS` environment variable used in our Makefile:
```
SPHINXOPTS="-E -j 8" make html
```

arijc76:
> > Can you let me know what I need to do for this?
> 
> I can't reproduce your error, but you can tell Sphinx not to treat warnings as errors by overriding the `SPHINXOPTS` environment variable used in our Makefile:
> 
> ```
> SPHINXOPTS="-E -j 8" make html
> ```

Thanks @eb8680 
Sorry about the delay. I was facing some issues with `make html` not working locally, but that's solved now (made a couple of changes in `conf.py` along with above suggested change for SPHINXOPTS in the Makefile). 
I have completed the suggested changes and now when I run `make html` to render the HTML, the example appears under the "other inference algorithms" header on the sidebar. The generated HTML of the notebook does not show any visual errors. 
Can I commit only the updated version of the notebook for a review?

PR #3010, opened 2022-01-23
add log normal negative binomial distributions

martinjankowiak:
a 3-parameter generalization of the negative binomial distribution.

### DIFF
--- a/pyro/distributions/log_normal_negative_binomial.py
+++ b/pyro/distributions/log_normal_negative_binomial.py
@@ -134,10 +134,6 @@ class LogNormalNegativeBinomial(TorchDistribution):
         new._validate_args = self._validate_args
         return new
 
-    @constraints.dependent_property
-    def support(self):
-        return self.nb_dist.support
-
     @lazy_property
     def mean(self):
         return torch.exp(
--- a/pyro/distributions/log_normal_negative_binomial.py
+++ b/pyro/distributions/log_normal_negative_binomial.py
@@ -4,7 +4,8 @@ from numpy.polynomial.hermite import hermgauss
 import torch
 from torch.distributions import constraints
 
-from pyro.distributions import NegativeBinomial, TorchDistribution
+from pyro.distributions.torch_distribution import TorchDistribution
+from pyro.distributions.torch import NegativeBinomial
 from pyro.distributions.util import broadcast_shape
 
 
--- a/pyro/distributions/log_normal_negative_binomial.py
+++ b/pyro/distributions/log_normal_negative_binomial.py
@@ -1,24 +1,14 @@
 # Copyright Contributors to the Pyro project.
 # SPDX-License-Identifier: Apache-2.0
 
-import numpy as np
 import torch
-from numpy.polynomial.hermite import hermgauss
 from torch.distributions import constraints
 from torch.distributions.utils import broadcast_all, lazy_property
 
 from pyro.distributions.torch import NegativeBinomial
 from pyro.distributions.torch_distribution import TorchDistribution
 from pyro.distributions.util import broadcast_shape
-
-
-def get_quad_rule(num_quad, prototype_tensor):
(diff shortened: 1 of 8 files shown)


pyro/distributions/log_normal_negative_binomial.py
```diff
@@ -0,0 +1,58 @@
+import numpy as np
+from numpy.polynomial.hermite import hermgauss
+
+import torch
+from torch.distributions import constraints
+
+from pyro.distributions import NegativeBinomial, TorchDistribution
```
fritzo, line 7: import these from their specific modules:
```py
from pyro.distributions.torch import NegativeBinomial
```
That should resolve the circular import issues. The problem was that `pyro.distributions` imports this module and vice versa. Does that make sense?

martinjankowiak: yup thanks!

pyro/distributions/log_normal_negative_binomial.py
```diff
@@ -0,0 +1,162 @@
...
+import torch
+from torch.distributions import constraints
+from torch.distributions.utils import lazy_property, broadcast_all
+
+from pyro.distributions.torch_distribution import TorchDistribution
+from pyro.distributions.torch import NegativeBinomial
+from pyro.distributions.util import broadcast_shape
+
+
+def get_quad_rule(num_quad, prototype_tensor):
```
martinjankowiak, line 13: should i put this in pyro ops or the like and add tests?

pyro/distributions/log_normal_negative_binomial.py
```diff
@@ -0,0 +1,162 @@
...
+        super().__init__(batch_shape, event_shape, validate_args)
+
+    def log_prob(self, value):
+        nb_log_prob = self.nb_dist.log_prob(value.unsqueeze(-1))
+        return torch.logsumexp(self.log_weights + nb_log_prob, axis=-1)
+
+    def sample(self, sample_shape=torch.Size()):
+        raise NotImplementedError
+
+    def expand(self, batch_shape, _instance=None):
```
martinjankowiak, line 126: this method always confuses me...

martinjankowiak:
@fritzo this should be ready for review, at your convenience

martinjankowiak:
docs preview 
docs

pyro/distributions/log_normal_negative_binomial.py
```diff
@@ -0,0 +1,164 @@
...
+        "multiplicative_noise_scale": constraints.positive,
+    }
+    support = constraints.nonnegative_integer
+
+    def __init__(
+        self,
+        total_count,
+        logits,
+        multiplicative_noise_scale,
+        num_quad_points=8,
```
fritzo, line 91: nit: I'd insert a `*, ` above this line to ensure the default kwargs are name-only; that makes it safer to add kwargs later.

pyro/distributions/log_normal_negative_binomial.py
fritzo: me too. this looks good though

fritzo (changes requested):
Can you append one or two test cases to the end of [this list](https://github.com/pyro-ppl/pyro/blob/65225ae2ac925db404e8355e35a99f39790b0db5/tests/distributions/conftest.py#L662 in `tests/distributions/conftest.py`?

pyro/distributions/log_normal_negative_binomial.py
fritzo: It's up to you. If you foresee reusing it, feel free to move it. Otherwise I'd rename it to `_get_quad_rule()` to ensure it's private to this file.

martinjankowiak:
@fritzo this should be good to go unless additional tests are required (don't think so)

fritzo (approved):

PR #3011, opened 2022-01-24
Support 4D ProjectedNormal distribution

fritzo:
Addresses @geoffwoollard's use case discussed [on the forum](https://forum.pyro.ai/t/von-mises-fisher-scoring-and-sampling/3788/6).

## Tested
- added a new test case which is covered by existing gof tests

### DIFF
--- a/pyro/distributions/projected_normal.py
+++ b/pyro/distributions/projected_normal.py
@@ -170,3 +170,24 @@ def _log_prob_3(concentration, value):
     ).log()
 
     return para_part + perp_part
+
+
+@ProjectedNormal._register_log_prob(dim=4)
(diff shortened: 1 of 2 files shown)


geoffwoollard:
wow that is an amazingly fast response! thanks! I'm usually doing `! pip3 install pyro-ppl `. How long will it take this patch to get through to that. Can I get it before that by installing from source. I've been developing in a colab notebook.

fritzo:
In colab you can
```py
!pip install https://github.com/pyro-ppl/pyro/archive/projected-normal-4.zip
```

martinjankowiak (approved):

PR #3012, opened 2022-01-25
Add make tutorial to docs CI workflow

eb8680:
Motivated by off-target Sphinx build errors in #3006. Can't remember if there was a reason we weren't doing this, feel free to close if there was one...


docs/requirements.txt
```diff
@@ -7,3 +7,8 @@ opt_einsum>=2.3.2
 pyro-api>=0.1.1
 tqdm>=4.36
 funsor[torch]
+nbformat
+nbsphinx>=0.3.2
+nbstripout
+pypandoc
+ninja
```
fritzo, line 14: I worry this will make our readthedocs build slower and flakier, which is a problem because the rtd debugging interface is cumbersome.  Could we instead add these dependencies to the docs stage in ci.yml, so they are triggered in github's easier to debug workflow?

eb8680:
Not sure what's going on, but the tutorial HTML build is taking forever on the GitHub CI machines. I might have to move this to a separate stage to avoid a big increase in CI turnaround time.

fritzo:
> taking forever

that's probably the reason we didn't add this to ci

eb8680:
Still, even on my slow personal laptop `make tutorial` runs in seconds with no cached builds - there's no way GitHub should be taking 25+ minutes.

.github/workflows/ci.yml
```diff
@@ -62,6 +62,40 @@ jobs:
         run: |
           make docs
           make doctest
+  tutorials-html:
```
eb8680, line 65: I moved the tutorial build to a separate stage to limit the impact on CI time

fritzo (approved):
Thanks for getting this to work @eb8680 !

Issue #3026, opened 2022-02-28
[Feature Request] Better Errors when passing numpy array instead of torch.tensor


nipunbatra:
I was trying to write a small example for GP classification.

```python
import pyro
import pyro.contrib.gp as gp
import numpy as np
import torch

X = torch.randn(200, 2)
Y = torch.logical_xor(X[:, 0] > 0, X[:, 1] > 0).double()

kernel = gp.kernels.RBF(input_dim=2)
likelihood = gp.likelihoods.Binary()
model = gp.models.VariationalGP(X, Y, kernel, likelihood=likelihood, whiten=True)
```

The above code works. However, as expected if instead of `X = torch.tensor`, I got an error which wasn't very obvious. 

```python
rng = np.random.RandomState(0)

X = rng.randn(200, 2)

likelihood = gp.likelihoods.Binary()
model = gp.models.VariationalGP(X, Y, kernel, likelihood=likelihood, whiten=True)
```

Error Trace
```python

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [5], in <module>
      3 X = rng.randn(200, 2)
      5 likelihood = gp.likelihoods.Binary()
----> 6 model = gp.models.VariationalGP(X, Y, kernel, likelihood=likelihood, whiten=True)

File ~/miniforge3/lib/python3.9/site-packages/pyro/contrib/gp/models/vgp.py:74, in VariationalGP.__init__(self, X, y, kernel, likelihood, mean_function, latent_shape, whiten, jitter)
     63 def __init__(
     64     self,
     65     X,
   (...)
     72     jitter=1e-6,
     73 ):
---> 74     super().__init__(X, y, kernel, mean_function, jitter)
     76     self.likelihood = likelihood
     78     y_batch_shape = self.y.shape[:-1] if self.y is not None else torch.Size([])

File ~/miniforge3/lib/python3.9/site-packages/pyro/contrib/gp/models/model.py:93, in GPModel.__init__(self, X, y, kernel, mean_function, jitter)
     91 def __init__(self, X, y, kernel, mean_function=None, jitter=1e-6):
     92     super().__init__()
---> 93     self.set_data(X, y)
     94     self.kernel = kernel
     95     self.mean_function = (
     96         mean_function if mean_function is not None else _zero_mean_function
     97     )

File ~/miniforge3/lib/python3.9/site-packages/pyro/contrib/gp/models/model.py:189, in GPModel.set_data(self, X, y)
[...]

PR #3036, opened 2022-03-04
Added type assertion for better code clarity

GautamV234:
I am working with Prof. @nipunbatra and I am currently working on this PR to solve #3026.


fritzo:
Can you `make lint` and ensure your code is well formatted? I believe you'll need to change the assertion to
```python
assert isinstance(X, torch.Tensor)
```

fehiepsi:
I thought that it was recommended to use `torch.is_tensor(x)` but looking at its docs again, it recommends to use `is_instance(...)` pattern.

GautamV234:
> Can you `make lint` and ensure your code is well formatted? I believe you'll need to change the assertion to
> 
> ```python
> assert isinstance(X, torch.Tensor)
> ```

Sure I will do that and resubmit

GautamV234:
I have modified all the files (added assertion checks) having this problem in the pyro/contrib/gp/models folder

Below is an example that illustrates the assertion error.

```py 
import pyro
import pyro.contrib.gp as gp
import numpy as np
import torch 
X = np.random.randn(200, 2)
Y = np.logical_xor(X[:, 0] > 0, X[:, 1] > 0).astype(np.float64) 
kernel = gp.kernels.RBF(input_dim=2)
likelihood = gp.likelihoods.Binary()
model = gp.models.VariationalGP(X, Y, kernel, likelihood=likelihood, whiten=True)

AssertionError                            Traceback (most recent call last)
Input In [6], in <cell line: 11>()
      9 kernel = gp.kernels.RBF(input_dim=2)
     10 likelihood = gp.likelihoods.Binary()
---> 11 model = gp.models.VariationalGP(X, Y, kernel, likelihood=likelihood, whiten=True)

File C:\work\Pyro\pyro\pyro\contrib\gp\models\vgp.py:74, in VariationalGP.__init__(self, X, y, kernel, likelihood, mean_function, latent_shape, whiten, jitter)     63 def __init__(
     64     self,
     65     X,
   (...)
     72     jitter=1e-6,
     73 ):
---> 74     assert isinstance(
     75         X, torch.Tensor
     76     ), "X needs to be a torch Tensor instead of a {}".format(type(X))
     77     assert isinstance(
     78         y, torch.Tensor
     79     ), "y needs to be a torch Tensor instead of a {}".format(type(y))
     80     super().__init__(X, y, kernel, mean_function, jitter)

AssertionError: X needs to be a torch Tensor instead of a <class 'numpy.ndarray'>
```
Please let me know if some tests need to be added pertaining to the updated code.

GautamV234:
Hi there! I have been facing some issues while running the make lint command and would be great if you could look into it and let me know how I should proceed. PFA the screenshot of the issue. Thanks in advance.
image


> Thanks for adding the checks! I think no need to add tests. Pyro is based on PyTorch, not numpy, so we would expect all inputs are torch tensors.
> 
> Could you fix the lint issue? You can run `make lint`, `make format` to catch and fix the issues.

GautamV234:
> Run `make format` and push

Thanks for the help!

fehiepsi (approved):
Thanks, @GautamV234 !

Issue #3023, opened 2022-02-25
[Feature Request] 1) More details in `render_model` (esp. for `pyro.params`); 2) Returning dictionary from render_mode


nipunbatra:
Hi,
From the MLE-MAP tutorial, we have the following models

### MLE model
```python
def model_mle(data):
    f = pyro.param("latent_fairness", torch.tensor(0.5), 
                   constraint=constraints.unit_interval)
    with pyro.plate("data", data.size(0)):
        pyro.sample("obs", dist.Bernoulli(f), obs=data)
```

If we render this, we get something like the following image
<img width="333" alt="Screenshot 2022-02-25 at 4 30 40 PM" src="https://user-images.githubusercontent.com/60985/155704377-629009af-267b-4c11-b8d0-820497f39d5c.png">

### MAP model

```python
def original_model(data):
    f = pyro.sample("latent_fairness", dist.Beta(10.0, 10.0))
    with pyro.plate("data", data.size(0)):
        pyro.sample("obs", dist.Bernoulli(f), obs=data)
```

If we render this, we get something like the following image

<img width="395" alt="Screenshot 2022-02-25 at 4 30 56 PM" src="https://user-images.githubusercontent.com/60985/155704513-49fe7409-57f8-47f5-a132-04e56230d6c7.png">

### Coin toss graphical model images from the popular [Maths for ML](https://mml-book.github.io/book/mml-book.pdf) book

This is from Figure 8.10

<img width="1531" alt="Screenshot 2022-02-25 at 4 34 45 PM" src="https://user-images.githubusercontent.com/60985/155704648-707a5839-dd7e-4a75-a00b-eb10e5014c5d.png">

We'd expect our MLE model render to look like 8.10 b) and our MAP model to look like 8.10 c)

So, when we have `latent_fairness` as a parameter, it should perhaps just be written as `latent_fairness` and under the MAP model, it should be parameterised by the `Beta` distribution.

From the pyro render of the MLE model, it is not easily visible how observations are related to `latent_fairness`. 

### Feature Requests
So, I have two questions/requests

1. Would it make sense to have `pyro.params` also show in renders. The difference in the renders between `pyro.sample` and `pyro.parameter` would be the associated distribution (and thus hyperparams) in `pyro.sample`
[...]

PR #3039, opened 2022-03-08
improved `render_model`

karm216:
As discussed in #3023, I made the following changes in [pyro.infer.inspect.py](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/inspect.py)

* added `_pyro_post_param()` method for provenance tracking for `pyro.param`.
* added `bool: render_param` argument in `render_model()` to display param in graph.
* added `sample_param` and `param_constraint` in dictionary returned by `get_model_relations()`
* modified `generate_graph_specification()` and `render_graph()` to show param in graph returned by `render_model()`

@fritzo, Can you please suggest which type of test cases needs to be changed/added to test the above changes?


karm216:
Hi @fritzo, I tried to study why this check is failing, But I'm unable to get an exact reason. It looks like adding provenance tracking for `pyro.param` conflicts with other methods. Can you please help me to get started to solve this problem?

karm216:
Hi @fritzo, This PR is ready for review.

pyro/infer/inspect.py
```diff
@@ -276,17 +282,34 @@ def model(data):
...
+    param_constraint = {}
     plate_sample = defaultdict(list)
     observed = []
+
+    def _get_type_from_frozenname(frozen_name):
+        return trace.nodes[frozen_name]["type"]
+
     for name, site in trace.nodes.items():
+        if site["type"] == "param":
+            param_constraint[name] = site["kwargs"]["constraint"]
```
fritzo, line 296: Could we store a string representation of the constraint here, rather than the `Constraint` object?  For example:
```py
param_constraint[name] = str(site["kwargs"]["constraint"])
```
The reason is that the constraint object may contain tensors, and storing the tensor values in the dependencies dict may leak memory. This dict is intended to be pure-Python with no PyTorch data structures.

pyro/infer/inspect.py
```diff
@@ -339,6 +366,16 @@ def generate_graph_specification(model_relations: dict) -> dict:
         rv for rv in model_relations["sample_sample"] if rv not in plate_rvs
     ]  # RVs which are in no plate
 
+    params = set()
```
fritzo, line 369: nit: let's give this a consistent type, either set or list. we're moving towards checking with mypy and we want to make sure this code won't cause later type errors

pyro/infer/inspect.py
```diff
@@ -339,6 +366,16 @@ def generate_graph_specification(model_relations: dict) -> dict:
...
+    if render_params:
+        for rv, params_list in model_relations["sample_param"].items():
+            for param in params_list:
+                params.add(param)
+        params = list(params)
```
fritzo, line 374: I think you can simply delete this line, right?

fritzo (commented):
Hi @karm216, thanks for your patience.  Your code looks great!  I have only minor comments.

Re: testing, I think since this is a visual output and we have no existing tests for `get_model_relations()` it is enough to simply rerun tutorial/source/intro_long.ipynb and commit the updated output. I think your improved rendering will improve that tutorial. Feel free to add a sentence or two to that tutorial describing the difference between sample and param nodes, if you think that's needed.

pyro/infer/inspect.py
karm216: Oh thanks for pointing out, I have corrected it.

pyro/infer/inspect.py
karm216: Yes, I removed that line.

pyro/infer/inspect.py
karm216: Yes, I resolved it by keeping set only.

fritzo:
Hi @karm216 it looks like your changes include accidental running of `black` on notebooks. Can you please revert that part of the PR?

fritzo (approved):
Loogs great, thanks for updating the notebooks!

karm216:
Hi @fritzo , Some checks are failed. It looks like there is a problem in installing dependencies. Please let me know if there is a problem at my side.

PR #3042, opened 2022-03-11
change sample to param statements in guides

dhudsmith:
This change only effects the SVI Part IV: Tips and Tricks tutorial. In the section on parameter initialization. The param statements are incorrectly written as sample statements.


fehiepsi (approved):
Nice catch! Thanks, @dhudsmith!

PR #3045, opened 2022-03-16
Update to PyTorch 1.11.0

fritzo:
This is long overdue: before this PR Pyro's dev branch was still on torch==1.9.0.

- Bump Python version to 3.7, drops support for 3.6 (which reached end of life [last December](https://endoflife.date/python))
- Bump PyTorch to 1.11.0, torchvision to 0.12.0
  - Convert `torch.triangular_solve()` -> `torch.linalg.solve_triangular()`
  - Ignore irrelevant `__floordiv__` warnings
  - Fix catching of singular matrices in hmc
  - Resolves #3032 (meshgrid workaround)
  - Resolves #3035
  - Remove an XPASSing jit test
  - Xfailed some funsor tests, these are collected in #3046
  - Work around a pandas>=1.3 bug in the baseball example
- Update to latest `black`

I've tried to keep each git commit narrowly focused on a single change.

PR #3047, opened 2022-03-17
Drop more references to Python 3.6

fritzo:
This drops more references to Python 3.6, following up on #3045 

@jamestwebber one reference that remains is the [docker/Makefile](https://github.com/pyro-ppl/pyro/blob/dev/docker/Makefile), which seems to use python==3.6 and an ubuntu image from 2016. Yikes actually, the [Dockerfile](https://github.com/pyro-ppl/pyro/blob/dev/docker/Dockerfile) is still on Python 2.7. I guess we should update this to Python 3.7 and a 2020 image?  Or should just we wait until users ask for an update?

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,9 +34,6 @@ For additional blog posts, check out work on [experimental design](https://eng.u
 ### Installing a stable Pyro release
 
 **Install using pip:**
-
-Pyro supports Python 3.6+.
-
 ```sh
 pip install pyro-ppl
 ```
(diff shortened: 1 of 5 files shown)


jamestwebber:
Is the `docker` folder used for development, or is it just intended as a tutorial? If the latter--in my experience, the trickier part is installing pytorch w/ GPU support, and pyro is easy.

Unless it's being used for CI or something, seems like it could just be removed. But maybe some people on the team are using it for development? I'm curious if they have their own updated version.

fritzo:
@jamestwebber I really have no idea who uses the docker file. It might be an old holdover from @neerajprad's days at Uber 🤷

fehiepsi (approved):

PR #3050, opened 2022-03-21
Implement DiscreteHMM.sample()

fritzo:
Inspired by a [forum question](https://forum.pyro.ai/t/modelling-mutivariate-switching-dynamics-advice-on-getting-started/4017/4) whose answer would have been easier if we had implemented `DiscreteHMM.sample()` for generating data. In addition, [Bayesian workflow](https://arxiv.org/abs/2011.01808) recommends sanity checking models by generating data.

## Tested
- [x] added lots of shape tests

### DIFF
--- a/pyro/distributions/hmm.py
+++ b/pyro/distributions/hmm.py
@@ -15,6 +15,7 @@ from pyro.ops.gaussian import (
     matrix_and_mvn_to_gaussian,
     mvn_to_gaussian,
 )
+from pyro.ops.indexing import Vindex
 from pyro.ops.special import safe_log
 from pyro.ops.tensor_utils import cholesky, cholesky_solve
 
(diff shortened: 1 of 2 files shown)


pyro/distributions/hmm.py
```diff
@@ -423,6 +424,32 @@ def filter(self, value):
...
+        trans_shape = self.batch_shape + (self.duration, S, S)
+        trans_logits = self.transition_logits.expand(trans_shape)
+        xs = []
+        for t in range(self.duration):
+            x = Categorical(logits=Vindex(trans_logits)[..., t, x, :]).sample()
+            xs.append(x)
+        x = torch.stack(xs, dim=-1)
+
+        # Sample observations conditioned on hidden states.
+        obs_shape = self.batch_shape + (self.duration, S)
```
martinjankowiak, line 447: shouldn't we index and then sample?

tests/distributions/test_hmm.py
```diff
@@ -55,6 +55,15 @@ def check_expand(old_dist, old_data):
     assert new_dist.log_prob(new_data).shape == new_batch_shape
 
 
+def check_sample_shape(d):
```
martinjankowiak, line 58: shouldn't we have a more comprehensive test? e.g. create a `duration=1` hmm with posterior mass that concentrates on `x=0` and test accordingly?

martinjankowiak (commented):
generally looks good but this `sample` method is non-parallel in contrast to all (?) other `*HMM` code. should at least comment in the docstring, which purports to `parallelize over time`?

pyro/distributions/hmm.py
fritzo: I've added a comment explaining why the efficient approach is hard (see also #3052).

tests/distributions/test_hmm.py
fritzo: Thanks for keeping me honest! I've added a test of distribution with a simple example as you suggest.

martinjankowiak (approved):

fritzo:
Thanks for reviewing @martinjankowiak!

PR #3053, opened 2022-03-22
Parallelized `DiscreteHMM.sample()`

ordabayevy:
This is a follow up PR to #3050 that proposes to use sampling approach based on parallel-scan algorithm.

### DIFF
--- a/pyro/distributions/hmm.py
+++ b/pyro/distributions/hmm.py
@@ -273,8 +273,10 @@ class HiddenMarkovModel(TorchDistribution):
 class DiscreteHMM(HiddenMarkovModel):
     """
     Hidden Markov Model with discrete latent state and arbitrary observation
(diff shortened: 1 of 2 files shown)


fritzo (changes requested):
Nice work @ordabayevy!  The algorithm looks good to me.

Could you update the docstring to reflect complexity changes, in particular the two paragraphs

>    This uses [1] to parallelize over time, achieving O(log(time)) parallel
>    complexity for computing :meth:`log_prob` and :meth:`filter`.

and

>    The :meth:`sample` method is sequential (not parallized), slow, and memory
>    inefficient. It is intended for data generation only and is not recommended
>    during inference.

fritzo (approved):

ordabayevy:
Thanks for reviewing @fritzo !

Issue #3046, opened 2022-03-17
Some pyro.contrib.funsor tests are failing in torch==1.11.0


fritzo:
This issue records tests of `pyro.contrib.funsor` that started failing in #3045. In that PR I've xfailed the following tests so as to unblock that PR.
- [ ] `python examples/contrib/funsor/hmm.py --num-steps=1 --truncate=10 --model=5 --funsor`
- [ ] `pytest -x tests/contrib/funsor/test_enum_funsor.py::test_elbo_enumerate_plate_1`
- [ ] `pytest -x tests/contrib/funsor/test_enum_funsor.py::test_elbo_enumerate_plate_4`
- [ ] `pytest -x tests/contrib/funsor/test_enum_funsor.py::test_elbo_enumerate_plate_6`
- [ ] `pytest -x tests/contrib/funsor/test_enum_funsor.py::test_elbo_enumerate_plate_7`
- [ ] `pytest -x tests/contrib/funsor/test_enum_funsor.py::test_elbo_enumerate_plates_2`
- [ ] `pytest -x tests/contrib/funsor/test_enum_funsor.py::test_elbo_enumerate_plates_3`
- [ ] `pytest -x tests/contrib/funsor/test_enum_funsor.py::test_elbo_enumerate_plates_4`
- [ ] `pytest -x tests/contrib/funsor/test_enum_funsor.py::test_elbo_enumerate_plates_5`
- [ ] `pytest -x tests/contrib/funsor/test_enum_funsor.py::test_elbo_enumerate_plates_6`
- [ ] `pytest -x tests/contrib/funsor/test_enum_funsor.py::test_elbo_enumerate_plates_7`
- [ ] `pytest -x tests/contrib/funsor/test_enum_funsor.py::test_elbo_enumerate_plates_8`

cc @eb8680 @ordabayevy

fritzo:
I suspect this is due to a torch==1.11.0 issue with `ProvenanceTensor` which we fixed on the Pyro side via
```diff
diff --git a/pyro/ops/provenance.py b/pyro/ops/provenance.py
index b1ec32d9..51a1ec9e 100644
--- a/pyro/ops/provenance.py
+++ b/pyro/ops/provenance.py
@@ -46,9 +46,7 @@ class ProvenanceTensor(torch.Tensor):
         assert not isinstance(data, ProvenanceTensor)
         if not provenance:
             return data
-        instance = torch.Tensor.__new__(cls)
[...]

PR #3048, opened 2022-03-18
Debugging torch=1.11 related pyro.contrib.funsor tests

ordabayevy:
Resolves #3046

PR #3054, opened 2022-03-22
Bump version to 1.8.1

fritzo:
In preparation for 1.8.1 release:
- update version checks in notebooks to 1.8.1
- update funsor pinned version to 0.4.3
- xfailed a bunch of pyro.contrib.funsor enumeration tests #3046

### DIFF
--- a/examples/air/main.py
+++ b/examples/air/main.py
@@ -272,7 +272,7 @@ def main(**kwargs):
 
 
 if __name__ == "__main__":
-    assert pyro.__version__.startswith("1.8.0")
+    assert pyro.__version__.startswith("1.8.1")
     parser = argparse.ArgumentParser(
         description="Pyro AIR example", argument_default=argparse.SUPPRESS
     )
--- a/examples/baseball.py
+++ b/examples/baseball.py
@@ -392,7 +392,7 @@ def main(args):
 
 
 if __name__ == "__main__":
-    assert pyro.__version__.startswith("1.8.0")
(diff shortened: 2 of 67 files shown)


fritzo:
Looks like we're seeing #3046 again with `funsor==0.4.3`. @ordabayevy to try to get tests to pass I've reverted #3048 (which had simply removed xfails)

fritzo:
Indeed git bisect shows the offending commit is @eb8680's and my https://github.com/pyro-ppl/funsor/pull/584

eb8680:
It looks like re-running the tests made them pass? There were at least a couple tests off by a factor of 2 rather than numerical error, which is odd. I'll look into the failures this evening. I don't think it's worth blocking a Pyro release on `contrib.funsor` bugs though - I'd be OK with either releasing as-is with some tests marked xfail or reverting the change in Funsor. I'll look into the failures tonight.

fritzo:
@eb8680 the tests passed because I xfailed them 😊  LMK if you want to pair debug on our funsor PR. I'm ok either reverting that PR or releasing pyro with some contrib.funsor bugs, but I'd prefer to do a quick bugfix release of funsor 0.4.4 if we can find & fix the error. It took so long to get that PR merged...

fritzo:
@fehiepsi, @eb8680 are ok merging this with the xfailed tests in contrib.funsor and we can look into fixing those tests later. I think the PyTorch update is higher priority.

martinjankowiak (approved):

PR #3059, opened 2022-03-31
Updating Docker file to use Ubuntu 18.04

Jayanth-kumar5566:
this update uses ubuntu 18.04 and python3 

As the latest version PyTorch requires CUDA version > 10 -- which I have modified in the Makefile

PR #3060, opened 2022-03-31
Updated CUDA and base OS version

Jayanth-kumar5566:


fritzo (approved):
Should this be merged into #3059? That is, are the two dependent? It seems simpler to combine these into a single PR.

Jayanth-kumar5566:
> #3059

Yes, they are dependent. Will do create a single PR next time.

PR #3064, opened 2022-04-10
Fix docstrings of distributions rendered on sphinx

fehiepsi:
Currently, our [readthedocs](https://docs.pyro.ai/en/stable/distributions.html) shows `(*args, **kwargs)` for all distributions. This PR tries to fix that issue.

Tested:
image


fritzo (approved):
Thanks @fehiepsi, looks great!

PR #3065, opened 2022-04-10
Add test applying AutoNormal to two different models

fritzo:
This simply verifies the behavior recommended in https://forum.pyro.ai/t/paramstore-not-correctly-updated/4091


fehiepsi (approved):
LGTM! The behavior in the test is expected.

PR #3066, opened 2022-04-13
Fix link in AIR tutorial

dilaragokay:
Amortization is covered in https://github.com/pyro-ppl/pyro/blob/dev/tutorial/source/svi_part_ii.ipynb


fritzo (approved):
Thanks for the fix!

Issue #3082, opened 2022-05-05
[bug] Unable to render the model in Dirichlet Process Mixture Models in Pyro


dilaragokay:
### Issue Description
I tried to render the model in https://pyro.ai/examples/dirichlet_process_mixture.html using `pyro.render_model(model, model_args=(data))`. However, I got the following error

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/tmp/ipykernel_29904/3865686802.py in <module>
     38         pyro.sample("obs", MultivariateNormal(mu[z], torch.eye(2)), obs=data)
     39 
---> 40 pyro.render_model(model, model_args=(data))

~/thesis/pyro/pyro/infer/inspect.py in render_model(model, model_args, model_kwargs, filename, render_distributions)
    492     :rtype: graphviz.Digraph
    493     """
--> 494     relations = get_model_relations(model, model_args, model_kwargs)
    495     graph_spec = generate_graph_specification(relations)
    496     graph = render_graph(graph_spec, render_distributions=render_distributions)

~/thesis/pyro/pyro/infer/inspect.py in get_model_relations(model, model_args, model_kwargs)
    274     with torch.random.fork_rng(), torch.no_grad(), pyro.validation_enabled(False):
    275         with TrackProvenance():
--> 276             trace = poutine.trace(model).get_trace(*model_args, **model_kwargs)
    277 
    278     sample_sample = {}

~/thesis/pyro/pyro/poutine/trace_messenger.py in get_trace(self, *args, **kwargs)
    196         Calls this poutine and returns its trace instead of the function's return value.
    197         """
--> 198         self(*args, **kwargs)
    199         return self.msngr.get_trace()

~/thesis/pyro/pyro/poutine/trace_messenger.py in __call__(self, *args, **kwargs)
    172             )
    173             try:
--> 174                 ret = self.fn(*args, **kwargs)
    175             except (ValueError, RuntimeError) as e:
    176                 exc_type, exc_value, traceback = sys.exc_info()

TypeError: model() takes 1 positional argument but 200 were given
```

### Environment
 - OS: Ubuntu 18.04.3 LTS
 - Python version: 3.7.11
 - PyTorch version: 1.11.0
 - Pyro version: 1.8.0+5ab7da2a

### Code Snippet
```python
import matplotlib.pyplot as plt
import numpy as np

import torch
import torch.nn.functional as F

import pyro
from pyro.distributions import *
from pyro.infer import SVI
[...]

PR #3083, opened 2022-05-05
Add type assertion for `model_args` and `model_kwargs` of `render_model`

dilaragokay:
If a model contains only one parameter, then a user might attempt to pass an argument that is not a `tuple`. In that case, the user is likely to get an error message which does not point out the exact reason (see #3082). Therefore, adding this assertion would help them pinpoint the reason faster.


fritzo (approved):
Thanks this is great!  WDYT of adding an assertion for `model_kwargs` as well?

fritzo (approved):

PR #3086, opened 2022-05-12
Remove `batch_expand` helper function in air example

ordabayevy:
Removing an old helper function that was used back in PyTorch 0.2.0 times.


fritzo (approved):
Thanks!

Issue #3098, opened 2022-06-01
MultivariateStudentT distribution initializes batch_shape as tuple instead of torch.Size


flo-schu:
https://github.com/pyro-ppl/pyro/blob/2b653e0c92d4e0840f2aafd81659d2b5ca6b8cf9/pyro/distributions/multivariate_studentt.py#L39

While this is no problem in itself, this created an Error when using `MultiVariateStudentT` distribution in a higher level package, because it required batch_shape to be of type `torch.Size`

I don't see any issue with assigning the variable to a `torch.Size` object. Did I miss something?

```python
batch_shape = torch.Size(broadcast_shape(df.shape, loc.shape[:-1], scale_tril.shape[:-2]))
```

PR #3099, opened 2022-06-01
store batch_shape as torch.Size object in MultivariateStudentT distribution

flo-schu:
resolves https://github.com/pyro-ppl/pyro/issues/3098
according to suggestions by @fritzo 

Since this is my first contribution here, I'd be happy if you could point out, things I should improve in the PR

### DIFF
--- a/pyro/distributions/multivariate_studentt.py
+++ b/pyro/distributions/multivariate_studentt.py
@@ -36,7 +36,9 @@ class MultivariateStudentT(TorchDistribution):
         assert scale_tril.shape[-2:] == (dim, dim)
         if not isinstance(df, torch.Tensor):
             df = loc.new_tensor(df)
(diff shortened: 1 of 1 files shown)


fritzo (dismissed):

flo-schu:
> Could you `make format` to fix the long line?

I assume there is a shortcut to what I did?

fritzo:
> I assume there is a shortcut to what I did?

All I see is the [failed lint stage](https://github.com/pyro-ppl/pyro/runs/6696392609?check_suite_focus=true) (LMK if you have trouble accessing this) where `black` complains.  I don't know what exactly it's complaining about (guessing a long line), but whatever it's complaining about you should be able to fix via `make format` or simply `black --check .`.  We do mention formatting in [CONTRIBUTING.md](https://github.com/pyro-ppl/pyro/blob/dev/CONTRIBUTING.md#testing) and if you can offer any improvements to contributing docs, we welcome those contributions too 🙂

PR #3101, opened 2022-06-02
Fix bound partial for python 3.10

### LINKED ISSUE #3018: [Support for Python 3.10] MCMC example in documentation does not work: `AttributeError: __enter__`
I am following the MCMC example in the documentation: <https://docs.pyro.ai/en/stable/mcmc.html#nuts>

```python
import pyro
from pyro.infer import MCMC, NUTS
import torch

true_coefs = torch.tensor([1., 2., 3.])
data = torch.randn(2000, 3)
dim = 3
labels = dist.Bernoulli(logits=(true_coefs * data).sum(-1)).sample()

def model(data):
    coefs_mean = torch.zeros(dim)
    coefs = pyro.sample('beta', dist.Normal(coefs_mean, torch.ones(3)))
    y = pyro.sample('y', dist.Bernoulli(logits=(coefs * data).sum(-1)), obs=labels)
    return y

nuts_kernel = NUTS(model, adapt_step_size=True)
mcmc = MCMC(nuts_kernel, num_samples=500, warmup_steps=300)
mcmc.run(data)
```

But it raises an error:

```
  File "/home/ayaka/Projects/test/main.py", line 19, in <module>
    mcmc.run(data)
  File "/home/ayaka/venv/lib/python3.10/site-packages/pyro/poutine/messenger.py", line 11, in _context_wrap
    with context:
AttributeError: __enter__
```

Pyro version: 1.8.0

### LINKED ISSUE #3095: `poutine.block` does not work in Python 3.10
### Reproducible code

```python
from pyro import poutine

class A:
    @poutine.block
    def run(self):
        return 1

a = A()
a.run()  # error
```

This causes the issue https://github.com/pyro-ppl/pyro/issues/3018

fehiepsi:
Fixes #3095 and resolves #3018.

In Python <= 3.9, `_bound_partial(_bound_partial(f)).func != f` but for some reasons, they are the same in Python 3.10. This PR tries to address such issue by registering `func` as a slot of `_bound_partial`.


fehiepsi:
For some reason, the change is failing for `infer_discrete` with `temperature=1` and funsor backend. I have been debugging this but still haven't figured it out.

fritzo (approved):
LGTM Valiant effort solving this @fehiepsi!

One thing to keep in mind is that classes with `__slots__` might not be pickleable (they weren't in Python 2, and I'm not sure if that's been fixed in Python 3).  Let's keep an eye out for new pickling errors this might introduce.

Balandat:
Are there plans to put out a new release with these changes anytime soon? Looks like the latest release 1.8.1. still doesn't support python 3.10.

fritzo:
Hi @Balandat sure we can push out a release this week. Thanks for the ping

fritzo:
@Balandat just released Pyro 1.8.2.

Balandat:
Woohoo, thanks! Great customer service! Where do I leave my yelp review?

PR #3114, opened 2022-07-08
Numerically stabilize autocorrelation()

fritzo:
This sets the autocorrelation of constant series to be 1.0 rather than NAN.

I believe this is the proper behavior for MCMC testing, where series with no variation should be seen as having effective sample size of 1.

My use case is in computing autocorrelation of discrete time series, where I'm using autocorrelation as a diagnostic to detect lack of mixing.  In that case I believe this PR is correct.

## Tested
- [x] added two unit tests
- [x] avoided NANs locally in [this notebook](https://github.com/fritzo/notebooks/blob/master/potts-locally-informed-mh.ipynb)

### DIFF
--- a/tests/ops/test_stats.py
+++ b/tests/ops/test_stats.py
@@ -114,6 +114,27 @@ def test_autocorrelation():
     )
 
 
+def test_autocorrelation_trivial():
+    x = torch.zeros(10)
+    with xfail_if_not_implemented():
+        actual = autocorrelation(x)
+    assert_equal(actual, torch.ones(10), prec=0.01)
+
+
+def test_autocorrelation_vectorized():
+    # make a mostly noisy x with a couple constant series
+    x = torch.randn(3, 4, 5)
+    x[1, 2] = 0
+    x[2, 3] = 1
+
+    actual = autocorrelation(x, dim=-1)
+    expected = torch.tensor([[autocorrelation(xij).tolist() for xij in xi] for xi in x])
+    assert_equal(actual, expected)
+
+    assert (actual[1, 2] == 1).all()
+    assert (actual[2, 3] == 1).all()
+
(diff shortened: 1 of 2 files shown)


martinjankowiak (commented):
is this the right fix? under what circumstances does `autocorr[..., :1]` get small?

fritzo:
Yeah I don't know if this is the right fix.  The issue is when one value doesn't vary at all, so in that case I think the autocorrelation should be 1.

martinjankowiak:
would it make sense to e.g. add an explicit test that checks that the resulting effective sample size is small for a variable with zero variance?

fritzo:
OK I've added some tests.

> add an explicit test that checks that the resulting effective sample size is small for a variable with zero variance

hmm, it looks like our `effective_sample_size()` function suffers from the same problem because it [divides by the variance](https://github.com/pyro-ppl/pyro/blob/dc1351a8d44581aca8e32a29939c81777e981c2e/pyro/ops/stats.py#L188) which may be zero.

tests/ops/test_stats.py
```diff
@@ -114,6 +114,27 @@ def test_autocorrelation():
     )
 
 
+def test_autocorrelation_trivial():
+    x = torch.zeros(10)
+    with xfail_if_not_implemented():
```
martinjankowiak, line 119: i guess this xfail is no longer required?

martinjankowiak (dismissed):

tests/ops/test_stats.py
fritzo: sure, let me try removing it :+1:

martinjankowiak (approved):

fritzo:
Thanks for reviewing!

PR #3115, opened 2022-07-08
Fix multinomial_goodness_of_fit printing

fritzo:
This fixes the `multinomial_goodness_of_fit(..., plot=True)` printing after `torch.Tensor` removed the `.__round__()` method at some past release.

## Tested
- [x] added a regression test
- [x] tested locally in [this notebook](https://github.com/fritzo/notebooks/blob/master/potts-locally-informed-mh.ipynb)

### DIFF
--- a/tests/distributions/testing/test_gof.py
+++ b/tests/distributions/testing/test_gof.py
@@ -1,0 +1,20 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+import torch.distributions as dist
+
+from pyro.distributions.testing.gof import multinomial_goodness_of_fit
+
+
+def test_multinomial_goodness_of_fit():
+    N = 100000
+    K = 20
+    logits = torch.randn(K)
+    probs = (logits - logits.logsumexp(-1)).exp()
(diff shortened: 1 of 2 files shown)


tests/distributions/testing/test_gof.py
```diff
@@ -0,0 +1,20 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+import torch.distributions as dist
+
+from pyro.distributions.testing.gof import multinomial_goodness_of_fit
+
+
+def test_multinomial_goodness_of_fit():
```
martinjankowiak, line 10: i'm confused about this test? does `print_histogram` not get called in existing tests?

fritzo: That's right printing is [disabled by default](https://github.com/pyro-ppl/pyro/blob/dc1351a8d44581aca8e32a29939c81777e981c2e/pyro/distributions/testing/gof.py#L86) (I think because it was filling up logs), and this new test explicitly enables printing to make sure it works.

martinjankowiak (approved):

PR #3116, opened 2022-07-09
Implement NanMaskedNormal, NanMaskedMultivariateNormal

fritzo:
This implements two distributions to serve as likelihoods for partially observed data, where unobserved elements are specified as NAN values.  This is new functionality beyond `pyro.mask()` and `Distribution.mask()` in that it allows NAN values _within_ an event of MultivariateNormal; in this case we can analytically marginalize out the missing value.  The `NanMaskedNormal` is similar to `Normal.mask(...)`, but I've included it for easier compatibility with the nontrivial `NanMaskedMultivariateNormal`.

My motivating example is a Bayesian multivariate linear regression model with learned multivariate noise distribution and partially observed response as specified in a pandas dataframe.  Each of the response columns is differently partially observed.

## Tested
- [x] unit test of `NanMaskedNormal`
- [x] unit test of `NanMaskedMultivariateNormal`
- [x] end-to-end smoke test of `NanMaskedMultivariateNormal`

### DIFF
--- a/pyro/distributions/nanmasked.py
+++ b/pyro/distributions/nanmasked.py
@@ -1,0 +1,99 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+
+from .torch import MultivariateNormal, Normal
+
+
+class NanMaskedNormal(Normal):
+    """
+    Wrapper around :class:`~pyro.distributions.Normal` to allow partially
+    observed data as specified by NAN elements in :meth:`log_prob`; the
+    ``log_prob`` of these elements will be zero. This is useful for likelihoods
+    with missing data.
+
+    Example::
+
+        from math import nan
+        data = torch.tensor([0.5, 0.1, nan, 0.9])
+        with pyro.plate("data", len(data)):
+            pyro.sample("obs", NanMaskedNormal(0, 1), obs=data)
+    """
+
+    def log_prob(self, value: torch.Tensor) -> torch.Tensor:
+        ok = value.isfinite()
+        if ok.all():
+            return super().log_prob(value)
+
+        # Broadcast all tensors.
+        value, ok, loc, scale = torch.broadcast_tensors(value, ok, self.loc, self.scale)
+        result = value.new_zeros(value.shape)
+
+        # Evaluate ok elements.
+        if ok.any():
+            marginal = Normal(loc[ok], scale[ok], validate_args=False)
+            result[ok] = marginal.log_prob(value[ok])
+        return result
+
+
+class NanMaskedMultivariateNormal(MultivariateNormal):
+    """
+    Wrapper around :class:`~pyro.distributions.MultivariateNormal` to allow
+    partially observed data as specified by NAN elements in the argument to
+    :meth:`log_prob`. The ``log_prob`` of these events will marginalize over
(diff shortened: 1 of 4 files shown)


pyro/distributions/nanmasked.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Example::
+
+        from math import nan
+        data = torch.tensor([
+            [0.1, 0.2, 3.4],
+            [0.5, 0.1, nan],
+            [0.6, nan, nan],
+            [nan, 0.5, nan],
+            [nan, nan, nan],
```
fehiepsi, line 55: I'm curious on what log prob of this all-nan element is?

fehiepsi:
Nice, I remember that this is requested by many forum users.

pyro/distributions/nanmasked.py
fritzo: it will be zero since all values will be marginalized out.

fritzo: here, I just pushed a test for this fact 🙂

pyro/distributions/nanmasked.py
```diff
@@ -0,0 +1,99 @@
...
+        n = result_shape.numel()
+        p = value.shape[-1]
+        value = value.reshape(n, p)
+        ok = ok.reshape(n, p)
+        loc = loc.reshape(n, p)
+        cov = cov.reshape(n, p, p)
+        result = value.new_zeros(n)
+
+        # Evaluate ok elements.
+        for pattern in sorted(set(map(tuple, ok.tolist()))):
```
martinjankowiak, line 86: oh i thought you were computing one big marginalized covariance with 0s/1s where appropriate so that everything could be vectorized (no for loop)

pyro/distributions/nanmasked.py
```diff
@@ -0,0 +1,99 @@
...
+            if not any(pattern):
+                continue
+            # Marginalize out NAN elements.
+            col_mask = torch.tensor(pattern)
+            row_mask = (ok == col_mask).all(-1)
+            ok_value = value[row_mask][:, col_mask]
+            ok_loc = loc[row_mask][:, col_mask]
+            ok_cov = cov[row_mask][:, col_mask][:, :, col_mask]
+            marginal = MultivariateNormal(ok_loc, ok_cov, validate_args=False)
```
martinjankowiak, line 95: do these invocation not need `covariance_matrix=`?

martinjankowiak: i guess one nice thing about this pattern is that you don't need to worry about factors of log 2pi explicitly...

martinjankowiak (approved):
lgtm. obviously there are various ways the computation could be sped-up in different regimes but since this is probably most useful in the relatively low dimensional setting anyway...

pyro/distributions/nanmasked.py
fritzo: 😄 that's beyond my linear algebra skills / patience.  In practice I'm working with 3 columns so there are at most 7 patterns.

pyro/distributions/nanmasked.py
fritzo: covariance_matrix is the [default first argument](https://pytorch.org/docs/stable/distributions.html#multivariatenormal), so no kwarg is necessary.

Issue #3104, opened 2022-06-07
Variance reduction for (Renyi)ELBO


OlaRonning:
From the forum [thread](https://forum.pyro.ai/t/iwae-elbo-in-pyro-numpyro/3604/2). [Tighter Variational Bounds are Not Necessarily Better](https://arxiv.org/pdf/1802.04537.pdf) demonstrates that tighter VBs can lead to deteriorating gradient estimates for VAEs trained with [IWAE](https://arxiv.org/pdf/1509.00519.pdf) due to the variance decreasing slower than the expected value. [DReG](https://arxiv.org/pdf/1810.04152.pdf) is a variance reduction technique that can alleviate this issue. 

@fritzo and @eb8680 have a DReG draft for RenyiELBO (#2605), from feature request #2598. Is it a good place to continue from?

@martinjankowiak you mentioned there may be other ways to revamp the ELBOs, did you have something particular in mind?

martinjankowiak:
@OlaRonning well i believe the current renyi implementation is suboptimal (or incorrect, depending on your point of view) for a local latent variable model like a VAE. basically it has to do with order of summation/logarithm. if there are plates that enclose the entire model then the corresponding summations should be moved out but as it is i believe the implementation effectively assumes that there is a single global latent variable and doesn't make effective use of conditional independence

not sure if you should start from [2605](https://github.com/pyro-ppl/pyro/pull/2605 ) but you should definitely look at it carefully :)

davidaknowles:
@martinjankowiak just to clarify is the RenyiELBO implementation just "incorrect" in terms of being higher variance (for a fixed `num_particles`) than it should be? i.e. it will still correctly estimate the true marginal likelihood for sufficiently large `num_particles`?

martinjankowiak:
@davidaknowles  yes, that's right. but in practice the variance is expected to be very high and `num_particles` would need to be very very large to compensate. as you can see there is an open PR but it was never merged...

davidaknowles:
Thanks that's helpful to know. I mostly just want it for getting test marginal likelihood where I can just throw a few thousand `num_particles` at it, rather than for training.

martinjankowiak:
yes though in that case i'd probably do it one data point at a time if that is feasible

PR #3123, opened 2022-08-02
Support inference with RenyiELBO for local latent variable models 

OlaRonning:
Addresses #3104

## TODO
- [x] Add Local latent variable model test case
- [ ] Correct RenyiELBO `logsumexp` reduction dim for LLV model

### DIFF
--- a/tests/infer/test_inference.py
+++ b/tests/infer/test_inference.py
@@ -1005,3 +1005,74 @@ def test_non_nested_plating_sum():
         loss = svi.step(data, weights)
         if step % 20 == 0:
             logger.info("step {} loss = {:0.4g}".format(step, loss))
+
+
+@pytest.mark.stage("integration", "integration_batch_2")
+class OneWayNormalRandomEffects(TestCase):
+    def setUp(self) -> None:
+        self.n_groups = 3
+        self.n_experiments = 5
+        self.data = torch.tensor(
+            [
+                [4.1, 3.5, 0.2, -3.3, 3.3],
+                [2.4, -6.5, -0.7, 4.4, -4.8],
+                [1.1, -0.6, 1.3, -1.3, -1.1],
+            ]
+        )
+        self.group_locs = torch.tensor([[3.0], [-2.0], [0.0]])
+        self.group_prec = torch.tensor([[0.2], [0.1], [0.3]])
+        self.obs_prec = torch.tensor(6.0)
+        obs_prec = self.obs_prec * self.n_experiments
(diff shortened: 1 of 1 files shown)


OlaRonning:
Should I ignore notebooks when using black? When running `black[jupyter]==22.6.0`, I get 48 files changed.

fritzo:
Yes could you please ignore notebooks?  The easiest way to do so is to run black through `make format`.  Alternatively we could set up a pyproject.toml file with the black exclusion list

OlaRonning:
> Yes could you please ignore notebooks?  

Sure, no problem.

> The easiest way to do so is to run black through `make format`.  Alternatively we could set up a pyproject.toml file with the black exclusion list.

Looking at the format target in the Makefile, it seems it's only black that needs a section in the TOML file. I'll configure a pyproject.toml and update the Makefile.

tests/infer/test_inference.py
```diff
@@ -1005,3 +1005,72 @@ def guide(data, weights):
         loss = svi.step(data, weights)
         if step % 20 == 0:
             logger.info("step {} loss = {:0.4g}".format(step, loss))
+
+
+@pytest.mark.stage("integration", "integration_batch_2")
+class OneWayNormalRandomEffects(TestCase):
```
martinjankowiak, line 1011: how much time do these tests take?

OlaRonning: > how much time do these tests take?

In CI :
|                                 | Time [sec] |
|---------------------------------|------------|
| `test_renyi_nonreparameterized` | 50.67         |
| `test_renyi_reparameterized`    | 36.59         |
| `test_renyi_vectorized`         | 36.32        |

So all are in the top ten slowest for batch 2.

OlaRonning:
@martinjankowiak, can you suggest a plating structure where we need to compute the dependence (i.e., `all_dims - sum_dims !=  num_particles_vectorized`)? I've been playing around with the current implementation but cannot come up with one.

martinjankowiak (commented):
@OlaRonning will this still give results for a local latent variable model like a VAE with a data `plate`? why doesn't this logic sum out the data plate dimension?

OlaRonning:
@martinjankowiak, sorry for the late reply. 
 >  why doesn't this logic sum out the data plate dimension?

From my understanding, this corresponds to including the data plate dimension in [`sum_dims`](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/renyi_elbo.py#L107), or have I misunderstood something?

Issue #3124, opened 2022-08-03
[feature request] handler for parameter substitution


thisiscam:
### Issue Description

It would be nice if pyro implements a handler/poutine that substitutes parameters by fixed constants, similar to `numpyro.handler.substitute`.
The motivation is to allow cleaner code for ancestral sampling from the model with fixed parameters.

For more details, see also: https://forum.pyro.ai/t/equivalent-of-numpyro-substitute-in-pyro/4465

### Code Snippet
See snippet below for my current temporary solution by using `lift` and `dist.Delta`:

```python
import torch
import pyro
from pyro import distributions as dist
from pyro import poutine
from torch.distributions import constraints


def model(N):
	a = pyro.param("a", 
			init_tensor=torch.tensor([0.]), 
			constraint=constraints.unit_interval)
	with pyro.plate("N", N):
		x = pyro.sample("x", dist.Bernoulli(probs=a))
	return x


def generate(rng_seed, N, a):
	with poutine.seed(rng_seed=rng_seed):
		# Is there a better alternative than below?
	    return poutine.lift(
	        model,
	        prior={
	            "a": dist.Delta(a),
	        })(N)
```

PR #3125, opened 2022-08-04
add substitute handler

thisiscam:
Attempt at #3124, largely mimics the `lift` handler


fritzo (commented):
Looks great and I especially like the name checking for usability!

1. Could you explain the need for the _data_cache?  Would this PR break if that were missing, or is it merely for speed?
2. Could you add some unit tests in tests/poutine/test_poutines.py, perhaps forking `ConditionHandlerTests`?
3. Could you add an entry in docs/source/pyro.poutine.rst (inserting in alphabetical order please 😉)
4. Looks like you'll need to run `make format` to fix coding style (see [CONTRIBUTING.md](https://github.com/pyro-ppl/pyro/blob/dev/CONTRIBUTING.md))

thisiscam:
Hi @fritzo, 

My guess is: It's probably ok to drop `_data_cache`, assuming nothing other than concrete values are passed into the substitution. E.g. we never substitute a distribution into a variable. Otherwise, we need to make sure that if there are multiple `param` calls, they are substituted to point to the same object. 

TBH, I have not looked into Pyro's code too much, and was largely mimicing the lift handler. 

For the test case, I'm not sure how `ConditionHandlerTests ` is relevant. Could you explain? It might be easier to add a simple standalone test case, though.

thisiscam:
@fritzo I added some tests, but I don't know how to run the test locally... Sorry
```
make test
``` 
produces:
 
```
/pyro/poutine/handlers.py:docstring of pyro.poutine.handlers._make_handler.<locals>.handler:9:Unexpected indentation.

```

fritzo:
I'm not sure why your local `make test` failed, but it might be failing linting rather than hitting the real test.  To run your particular tests you could try locally `pytest tests/poutine/test_poutines.py`

fritzo:
Hi @thisiscam, sorry this seems to have fallen off the radar and CI for some reason didn't trigger. Could you try merging in the dev branch and re-pushing?

thisiscam:
@fritzo Hi, sorry about the delay. I just merged `dev` and pushed to my branch. 
Just posting here in case GitHub didn't send a notification.

fritzo (approved):
Looks great, thanks for working through the CI issues.

PR #3131, opened 2022-08-28
Fix memory leak in TraceEnumELBO

### LINKED ISSUE #3014: [bug] Memory leak on GPU
### Issue Description
After successfully following the SS-VAE example in pyro I have moved over to training a variant of it on a GPU. A memory leak has arisen that has been boiled down to a minimum working example that replicates the error. I am at a loss as to how to proceed in rectifying this issue, any help or insight would be much appreciated!

### Environment
 - OS and python version. 
   - OS: Red Hat Enterprise Linux Workstation release 7.9 (Maipo) 
   - Python: 3.9.7
   - CUDA Version: 11.2
 - PyTorch version: 1.10.0
 - Pyro version: 1.7.0

### Code Snippet
``` python
from datetime import datetime

import pyro
import torch
import torch.nn as nn
from pyro.infer import SVI, TraceEnum_ELBO
from pyro.optim import Adam


class SSVAE(nn.Module):
    def __init__(self):
        super().__init__()
        self.accept_ratio = .5
        ones = torch.ones((), device='cuda')
        ratio = 1 / (self.accept_ratio + 1)
        self.alpha_prior = torch.stack((ones - ratio, ones - self.accept_ratio * ratio)).T
        self.cuda()

    def model(self):
        pyro.sample("y", dist.OneHotCategorical(self.alpha_prior))

    def guide(self):
        pass


def sizeof_fmt(num, suffix="B"):
    for unit in ["", "Ki", "Mi", "Gi", "Ti", "Pi", "Ei", "Zi"]:
        if abs(num) < 1024.0:
            return f"{num:3.1f}{unit}{suffix}"
        num /= 1024.0
    return f"{num:.1f}Yi{suffix}"


if __name__ == '__main__':
    torch.cuda.empty_cache()
    ssvae = SSVAE()

    optimizer = Adam({"lr": 1e-4})

    svi = SVI(ssvae.model, ssvae.guide, optimizer, loss=TraceEnum_ELBO())

    for i in range(10):
        print(f"{datetime.now().strftime('%H:%M:%S')} - Starting epoch #{i + 1}")
        print(
            f"Current GPU memory allocation:"
            f" {sizeof_fmt(torch.cuda.memory_allocated())}/{sizeof_fmt(torch.cuda.max_memory_allocated())}")

        # get the losses for an epoch
        for _ in range(100):
            svi.step()
```

### LINKED ISSUE #3068: Memory leak using TraceEnum_ELBO
I noticed a major memory leak when training SVI using `TraceEnum_ELBO`.
I initially noticed this in a custom model we are developing but then I found it seems a more general bug.

For example, it affects even the Pyro tutorials GMM example [here](https://pyro.ai/examples/gmm.html). Where memory usage rapidly goes from a couple of hundred MBs to a many **GB**s very quickly!

I have run this Macbook Pro 2019 running MacOS 10.15. To replicate the issue is enough running the notebook linked.

I have tried to comment out the following lines and add a garbage collector call, that reduces the entity of the memory accumulation of one order of magnitude but does not solve the problem completely, which becomes particularly severe for large datasets.

```python
# Register hooks to monitor gradient norms.
# gradient_norms = defaultdict(list)
# for name, value in pyro.get_param_store().named_parameters():
#     value.register_hook(lambda g, name=name: gradient_norms[name].append(g.norm().item()))

import gc
losses = []
for i in range(200000):
    loss = svi.step(data)
    #losses.append(loss)
    gc.collect()
``` 

(from [this forum post](https://forum.pyro.ai/t/memory-usage-grows-with-iterations-in-svi-step/4103))

fehiepsi:
Fixes #3068 and fixes #3014. I'm not sure what changes in PyTorch caused the issue but replacing `x[...]` with `x.clone()` fixes the memory leak.


pyro/ops/einsum/torch_log.py
```diff
@@ -23,7 +23,7 @@ def einsum(equation, *operands):
 
     inputs, output = equation.split("->")
     if inputs == output:
-        return operands[0][...]  # create a new object
+        return operands[0].clone()  # create a new object
```
fritzo, line 26: Could we add a comment here indicating that we avoid shallow copy to avoid a pytorch bug, so somebody doesn't cleverly revert this line in an effort to reduce memory footprint?

tests/ops/einsum/test_adjoint.py
```diff
@@ -114,3 +115,24 @@ def test_marginal(equation):
...
+
+def test_require_backward_memory_leak():
+    # Using clone resolves memory leak.
+    for i in range(10):
+        x = torch.tensor(0.0)
+        require_backward(x)
+        x._pyro_backward.process(x.clone())
+
+    tensors = [o for o in gc.get_objects() if torch.is_tensor(o)]
+    assert len(tensors) <= 5
```
fritzo, line 128: Could we grab the initial tensor count and use relative assertions in this test? Im afraid the current absolute assertions might fail if someome adds new global tensors, e.g. in a `@pytest.mark.parametrize` list or in a global cache

fritzo (commented):
Great sleuthing!  How did you find the offending line? Did you log tensor count or use any detective tooling code?

pyro/ops/einsum/torch_log.py
fritzo: Maybe this is why ci is failing

tests/ops/einsum/test_adjoint.py
fritzo: Maybe this is why ci is failing

pyro/ops/einsum/torch_log.py
fehiepsi: Done

tests/ops/einsum/test_adjoint.py
fehiepsi: That's a good point! I just updated the test to incorporate the logic. CI seems to be failing due to a deprecation warning caused by `torch.is_tensor(o)` during gc.

fehiepsi:
> How did you find the offending line?

Hi Fritz, I used your trick [here](https://forum.pyro.ai/t/a-trick-to-debug-tensor-memory/556) to check for memory leaks in [this example](https://github.com/pyro-ppl/pyro/issues/3014#issue-1142627375). After removing some gpu codes, I found that the leak also happens in CPU. Then I first copied the TraceEnum_ELBO implementation and removed line by line to see where is the offending code. That leads to  `Dice.compute_expectations` and then [this line](https://github.com/pyro-ppl/pyro/blob/d7f64747dbc387c3bc95c59bd95b4ec11b5df35b/pyro/infer/util.py#L298) with `root._pyro_backward()` call. When I print out the `root`, and `root._pyro_backward` I found that they are pretty simple expressions that only invoke `_EinsumBackward` and `_LeafBackward`, which are independent of the rest of pyro codes. I tried to play a little bit with `_LeafBackward` but couldn't replicate the leak so I guessed that the leak happened in some tensor operator of `torch_marginal._EinsumBackward`. It is lucky that only `pyro.ops.einsum.torch_log.einsum` is called and inside it, only `operands[0][...]` is called, so I found the leak. This process took me several hours though. (-:

fritzo (approved):
Thanks for your debugging story!

qinqian:
Thank you @fehiepsi for the detailed fixing history!

PR #3140, opened 2022-09-25
fix_reverse_out_bound_quadratic_spline

LiaoShiqi97:
Issue Description
quadratic spline in pyro.distributions.transforms.Spline does not consider outside_interval input at reverse.

According to the Neural spline flow(https://arxiv.org/abs/1906.04032), monotonic rational-quadratic transforms sets the boundary derivatives as 1 to match the linear tails. However, when it takes the y(sample from the target distribution ) to reverse back to the x(sample from the base distribution) in the function _monotonic_rational_spline, it does not consider the scenario where y is out of bound, which would render a negative discriminant and return Error.

Environment

python version: 3.8.7
PyTorch version: 1.12.1+cpu
Pyro version: 1.8.2
Solution
I replace a fake discriminant for the out-of-bound input.
Origin:
discriminant = b.pow(2) - 4 * a * c
assert (discriminant >= 0).all()

Changed:
discriminant = b.pow(2) - 4 * a * c
discriminant[outside_interval_mask] = 0 # added to make sure outputs[outside_interval_mask] = inputs[outside_interval_mask]
assert (discriminant >= 0).all()


pyro/distributions/transforms/spline.py
```diff
@@ -254,6 +254,7 @@ def _monotonic_rational_spline(
             c = -input_delta * (inputs - input_cumheights)
 
             discriminant = b.pow(2) - 4 * a * c
+            discriminant[outside_interval_mask] = 0 # added to make sure outside_interval input can be reversed as identity.
```
fritzo, line 257: I think you'll need two spaces before the comment, and possibly to move  the line comment to a separate line to satisfy max line length requirements.  You can locally run `make lint` to see lint errors.

Also you might need to use a non-inplace version like
```python
discriminant = discriminant.masked_fill(outside_interval_mask, 0)
```

fritzo (changes requested):

pyro/distributions/transforms/spline.py
LiaoShiqi97: Thanks for your comment. I am new to pulling a request. So in this case, it is because of my non-standard comment that my request fails, right?

fritzo: I believe so, we use the `black` automatic style checker to ensure all code follows the same style.  This helps everyone collaborate by making the code more uniform.

LiaoShiqi97: I get it. Thanks. I will edit it soon

LiaoShiqi97: I have re-pulled it. Please check it.

fritzo (approved):

PR #3122, opened 2022-07-30
Custom loss documentation

e-pet:
As proposed in #3121, these are some minor changes to the [custom objectives example page](https://pyro.ai/examples/custom_objectives.html).

- Renamed from "Custom SVI Objectives" to "Customizing objectives and training loops", also changed the file name.
- Extended the lower-level pattern to what I hope is a full replacement of `svi.step()`.

PR #3141, opened 2022-09-30
rename custom_objectives_training.ipynb -> custom_objectives.ipynb

martinjankowiak:
i noticed that the custom objectives tutorial is no longer visible on the website because #3122 renamed the `.ipynb` without accompanying changes in the `.rst` (and i didn't notice). this unrolls that change so that the tutorial will show up again upon next release


fritzo (approved):
Thanks @martinjankowiak!

PR #3147, opened 2022-10-20
Pin visdom version to unblock CI

fritzo:
Blocking #3140


martinjankowiak (dismissed):

fritzo:
@martinjankowiak thanks for approving, could you merge?

martinjankowiak (approved):

PR #3149, opened 2022-10-23
Add an option to stop PyroModules from sharing parameters

eb8680:
Replaces #2996 

This PR adds two small related features for easier Pyro-PyTorch integration:
1. A `__call__` method for the base `pyro.infer.elbo.ELBO` that binds `ELBO` instances to specific `nn.Module` model/guide pairs in a `Module` that exposes their PyTorch parameters
2. A global setting (off by default for full backwards compatibility) that prevents `PyroModule` instances from sharing parameter values with one another through the global Pyro parameter store, and a primitive and context manager for toggling it. One context where this is useful is for workflows that involve multiple models and autoguides with overlapping parameter names.

An edge case I haven't handled here is the behavior under the new local parameter setting of regular `pyro.param` statements (as opposed to `PyroParam`) within a `PyroModule` that don't have their data associated with any underlying `nn.Module`. I'm inclined to add some sort of warning or error rather than attempt to get this working, since I think it's usually a `PyroModule` programming anti-pattern to mix global and local parameter states in this way.

I am also hopeful that these changes will simplify the use of Pyro with the PyTorch JIT and other PyTorch compilers, but I have left testing this for future work, since I suspect it will require additional engineering that is out of scope for this PR.

Tested:
- Added test cases to some existing `PyroModule` tests in `tests/nn/test_module.py`
- Added a new test that checks the behavior of these two features together

### DIFF
--- a/pyro/infer/elbo.py
+++ b/pyro/infer/elbo.py
@@ -5,11 +5,24 @@ import logging
 import warnings
 from abc import ABCMeta, abstractmethod
 
+import torch
+
 import pyro
 import pyro.poutine as poutine
 from pyro.infer.util import is_validation_enabled
 from pyro.poutine.util import prune_subsample_sites
 from pyro.util import check_site_shape
+
+
+class _ELBOModule(torch.nn.Module):
+    def __init__(self, model, guide, elbo):
+        super().__init__()
+        self.model = model
+        self.guide = guide
+        self.elbo = elbo
+
+    def forward(self, *args, **kwargs):
+        return self.elbo.differentiable_loss(self.model, self.guide, *args, **kwargs)
 
 
 class ELBO(object, metaclass=ABCMeta):
@@ -85,6 +98,9 @@ class ELBO(object, metaclass=ABCMeta):
         self.ignore_jit_warnings = ignore_jit_warnings
         self.jit_options = jit_options
         self.tail_adaptive_beta = tail_adaptive_beta
+
+    def __call__(self, model, guide):
+        return _ELBOModule(model, guide, self)
 
     def _guess_max_plate_nesting(self, model, guide, args, kwargs):
         """
--- a/pyro/infer/elbo.py
+++ b/pyro/infer/elbo.py
@@ -15,7 +15,7 @@ from pyro.util import check_site_shape
 
 
 class _ELBOModule(torch.nn.Module):
-    def __init__(self, model, guide, elbo):
+    def __init__(self, model: torch.nn.Module, guide: torch.nn.Module, elbo: "ELBO"):
         super().__init__()
         self.model = model
         self.guide = guide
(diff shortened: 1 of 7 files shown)


pyro/infer/elbo.py
```diff
@@ -5,13 +5,26 @@
...
+import torch
+
 import pyro
 import pyro.poutine as poutine
 from pyro.infer.util import is_validation_enabled
 from pyro.poutine.util import prune_subsample_sites
 from pyro.util import check_site_shape
 
 
+class _ELBOModule(torch.nn.Module):
```
fritzo, line 17: nit: consider going ahead and name this publicly `ELBOModule`. I expect users may want to `torch.save` and inspect these things. And looking back I wish we had just made the `torch.constraints` classes public, since people now treat them as public even though there are underescores.

pyro/infer/elbo.py
```diff
@@ -5,13 +5,26 @@
...
+    def __init__(self, model: torch.nn.Module, guide: torch.nn.Module, elbo: "ELBO"):
+        super().__init__()
+        self.model = model
+        self.guide = guide
+        self.elbo = elbo
+
+    def forward(self, *args, **kwargs):
+        return self.elbo.differentiable_loss(self.model, self.guide, *args, **kwargs)
```
fritzo, line 25: Regarding your concern

 > Note that Pyro's global parameter store may cause this new interface to behave unexpectedly relative to standard PyTorch when working with :class:`~pyro.nn.PyroModule` s.

have you considered using a scope here?  I haven't thought this through, but I suppose we could use a param store scope, like this
```python
class ELBOModule(...):
    def __init__(self, ...):
        ...
        self.param_scope = {}

    def forward(self, ...):
        with pyro.params.param_store.scope(self.param_scope):
            return self.elbo.differentiable_loss(...)
```
For example, could a scope here simplify or replace the global `_PYRO_MODULE_LOCAL_PARAM` machinery?  Would making `self.param_scope = nn.ParamDict()` solve your issue of `pyro.param` being undefined in this new setting?

pyro/infer/elbo.py
```diff
@@ -86,6 +133,15 @@ def __init__(
         self.jit_options = jit_options
         self.tail_adaptive_beta = tail_adaptive_beta
 
+    def __call__(
+        self, model: torch.nn.Module, guide: torch.nn.Module
+    ) -> torch.nn.Module:
```
fritzo, line 138: nit: refine the return type to `ELBOModule`?

fritzo (commented):
Nice rethinking towards more idiomatic PyTorch!

pyro/infer/elbo.py
eb8680: Good point, done

pyro/infer/elbo.py
eb8680: I added a persistent parameter scope state to the `pyro.nn._Context` class, it seems to work as expected.

pyro/infer/elbo.py
eb8680: Done

pyro/nn/module.py
```diff
@@ -178,15 +182,23 @@ def __init__(self):
         self.active = 0
         self.cache = {}
         self.used = False
+        if _is_module_local_param_enabled():
+            self.param_state = {"params": {}, "constraints": {}}
 
     def __enter__(self):
+        if not self.active and _is_module_local_param_enabled():
+            self._param_ctx = pyro.get_param_store().scope(state=self.param_state)
```
eb8680, line 190: Persisting `self.param_state` like this (and in `_pyro_set_supermodule` below) seems to be a reasonable solution for the behavior of vanilla `pyro.param` statements. Values of these parameters are now local to the outermost `PyroModule` in a nested `PyroModule` instance.

pyro/nn/module.py
```diff
@@ -409,6 +421,8 @@ def named_pyro_params(self, prefix="", recurse=True):
             yield elem
 
     def _pyro_set_supermodule(self, name, context):
+        if _is_module_local_param_enabled():
+            context.param_state.update(self._pyro_context.param_state)
```
eb8680, line 425: This ensures values of vanilla `pyro.param` statements are preserved by nesting of `PyroModule` instances. This update could probably be refined to include only the vanilla values, but that's just an optimization.

eb8680: Updated further to only include the vanilla `param` values.

fritzo (commented):
Hey thanks for your patience in reviewing this subtle PR.  The `ELBOModule` changes look clean.  I'm still working through understanding the module_local_param changes...

tests/conftest.py
```diff
@@ -34,6 +34,7 @@ def pytest_runtest_setup(item):
     if test_initialize_marker:
         rng_seed = test_initialize_marker.kwargs["rng_seed"]
         pyro.set_rng_seed(rng_seed)
+    pyro.settings.set(module_local_params=False)
```
fritzo, line 37: I believe if we are careful to use context managers in tests, then the default settings should always be active.

tests/nn/test_module.py
```diff
@@ -532,7 +643,9 @@ def randomize(model):
     assert_identical(actual, expected)
 
 
-def test_torch_serialize_attributes():
+@pytest.mark.parametrize("local_params", [True, False])
+def test_torch_serialize_attributes(local_params):
+    pyro.settings.set(module_local_params=local_params)
```
fritzo, line 648: Could we use a `pyro.settings.context` here so as not to leak across other tests?

fritzo (commented):
LGTM after minor comment on `.set()` vs `.context()` in tests

tests/conftest.py
eb8680: Done

tests/nn/test_module.py
eb8680: Done

fritzo (approved):

PR #3151, opened 2022-10-28
Add jitter to Cholesky factorization in Gaussian ops

fritzo:
This numerically stabilizes Gaussian parallel-scan operations for use in very long chains, say > 10000 items.  This use case is important for heterogeneous-length batching, where a batch of sequences can be concatenated and operated on as a single chain.  In this setting I was seeing Cholesky errors in both single and double precision.

The fix in this PR is to modify `pyro.ops.tensor_utils.cholesky()` to add a small adaptive amount of jitter to all computations.  This appears to both pass all existing tests (which are quite strong, it doesn't work to add a non-adaptive jitter), and to make Gaussian parallel-scan filtering work for very long sequences. :tada:

## Tested
- [x] added tests for very long sequences
- [x] tried on a real-world example of length up to 1e6

### DIFF
--- a/pyro/ops/tensor_utils.py
+++ b/pyro/ops/tensor_utils.py
@@ -7,7 +7,7 @@ import torch
 from torch.fft import irfft, rfft
 
 _ROOT_TWO_INVERSE = 1.0 / math.sqrt(2.0)
-CHOLESKY_RELATIVE_JITTER = 1.0  # in units of finfo.eps
+CHOLESKY_RELATIVE_JITTER = 4.0  # in units of finfo.eps
 
 
 def as_complex(x):
@@ -403,7 +403,7 @@ def safe_cholesky(x):
     if CHOLESKY_RELATIVE_JITTER:
         # Add adaptive jitter.
         x = x.clone()
-        x_max = x.data.reshape(*x.shape[:-2], -1).abs().max(-1, True).values
+        x_max = x.data.abs().max(-1).values
         jitter = CHOLESKY_RELATIVE_JITTER * torch.finfo(x.dtype).eps * x_max
         x.data.diagonal(dim1=-1, dim2=-2).add_(jitter)
 
--- a/pyro/ops/tensor_utils.py
+++ b/pyro/ops/tensor_utils.py
@@ -2,34 +2,11 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import math
-import warnings
-from contextlib import contextmanager
-from typing import Optional
 
 import torch
 from torch.fft import irfft, rfft
 
 _ROOT_TWO_INVERSE = 1.0 / math.sqrt(2.0)
-JITTER = 1e-12
-
-
-@contextmanager
-def settings(*, jitter: float):
-    """
-    Context manager to set global settings.
-
-    :param float jitter: Constant added to matrix diagonals before
-        performing Cholesky decompositions, to improve stability.
-    """
-    global JITTER
-    jitter = float(jitter)
-    assert jitter >= 0
-    old = JITTER
-    try:
-        JITTER = jitter
-        yield
-    finally:
-        JITTER = old
 
 
 def as_complex(x):
@@ -416,32 +393,17 @@ def inverse_haar_transform(x):
     return x
 
 
-def cholesky(x, *, jitter: Optional[float] = None):
-    if jitter is None:
-        jitter = JITTER
-
-    # Handle simple scalar case.
+def cholesky(x, *, safe: bool = True):
     if x.size(-1) == 1:
-        if jitter:
-            x = x.clamp(min=0)
+        if safe:
+            x = x.clamp(min=torch.finfo(x.dtype).tiny)
         return x.sqrt()
 
-    if jitter > 0:
-        # First try without jitter.
-        result, info = torch.linalg.cholesky_ex(x)
-        if not info.any():
-            return result
-
-        # Try adding increasing amounts of jitter where needed.
-        warnings.warn("Singular matrix in cholesky(); adding jitter.")
+    if safe:
         x = x.clone()
-        x_diag = x.diagonal(dim1=-1, dim2=-2)
-        while jitter < 1:
-            x_diag[info > 0] += jitter
-            result, info = torch.linalg.cholesky_ex(x)
-            if not info.any():
-                return result
-            jitter *= 3
+        x_max = x.data.reshape(*x.shape[:-2], -1).abs().max(-1, True).values
+        jitter = x_max * torch.finfo(x.dtype).eps
+        x.data.diagonal(dim1=-1, dim2=-2).add_(jitter)
 
     return torch.linalg.cholesky(x)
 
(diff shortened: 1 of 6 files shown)


pyro/ops/tensor_utils.py
```diff
@@ -395,7 +395,15 @@ def inverse_haar_transform(x):
 
 def cholesky(x):
     if x.size(-1) == 1:
+        x = x.clamp(min=torch.finfo(x.dtype).tiny)
         return x.sqrt()
+
+    # Add adaptive jitter.
+    x = x.clone()
```
martinjankowiak, line 402: are you sure you need/want clones? fwiw i do [this](https://github.com/BasisResearch/millipede/blob/ecafe4194e216846600ff4f87dc02f65ce4428e2/millipede/util.py#L16) in millipede, which is similar to what's done in gpytorch

fritzo: Yes this `.clone()` is needed because we're mutating the matrix.

Thanks, I did try various gpytorch-style lazy tactics that avoid adding jitter until a failure has occurred.  I found that since the Cholesky error happens late in the filtering process, by that point the filter state had already been corrupted by nearly-singular matrices that just barely didn't trigger an error.  The best solution I've found so far is to add a tiny amount of noise to all matrices so that error doesn't build up during filtering.  Other solutions include using `svd` or `pinv` or `ldl_factor`, but they were more expensive.

Note the core piece of linear algebra is `.marginalize()` that is repeatedly called in the filter pass of `sequential_gaussian_filter_sample()`.  This has the nice property that if we add a little bit of jitter before Cholesky factorizing, the resulting `precision` becomes only better-conditioned.  Empirically this allows us to get away with much smaller jitter than would be required if we waited for an error to occur:
```python
P_aa = self.precision[..., a, a]
P_ba = self.precision[..., b, a]
P_bb = self.precision[..., b, b]
P_b = safe_cholesky(P_bb)                       # Note if we add a little jitter here...
P_a = triangular_solve(P_ba, P_b, upper=False)  # ...then this is smaller...
P_at = P_a.transpose(-1, -2)
precision = P_aa - matmul(P_at, P_a)            # ...so this is even better conditioned.
```

BTW it looks like you could speed up millipede by switching from `try: c = cholesky()` to the faster `c, info = cholesky_ex(); if not info.any(): return c`, which is used in gpytorch.  The only reason I'm not using `cholesky_ex()` here is that I found the decision-based version was too unstable.

martinjankowiak (approved):
lgtm. but `CHOLESKY_JITTER` isn't actually easily toggle-able is it?

pyro/ops/tensor_utils.py
```diff
@@ -393,15 +394,15 @@ def inverse_haar_transform(x):
...
+def safe_cholesky(x):
     if x.size(-1) == 1:
         x = x.clamp(min=torch.finfo(x.dtype).tiny)
         return x.sqrt()
 
     # Add adaptive jitter.
     x = x.clone()
     x_max = x.data.reshape(*x.shape[:-2], -1).abs().max(-1, True).values
-    jitter = x_max * torch.finfo(x.dtype).eps
+    jitter = CHOLESKY_JITTER * torch.finfo(x.dtype).eps * x_max
```
fehiepsi: Adding jitter relatives to a specific "norm" (or something else) of the matrix makes sense to me. Probably renaming `CHOLESKY_JITTER` into `RELATIVE_CHOLESKY_JITTER`?

fehiepsi (approved):

pyro/ops/tensor_utils.py
```diff
@@ -393,9 +394,19 @@ def inverse_haar_transform(x):
     return x
 
 
-def cholesky(x):
+def safe_cholesky(x):
     if x.size(-1) == 1:
+        if CHOLESKY_JITTER:
```
fehiepsi, line 399: Do you intend to use `CHOLESKY_JITTER * sqrt(finfo(x.dtype).eps)` here?

fritzo: My intention was to scale by about `finfo(x.dtype).eps * x.max()` so that the jitter was just barely detectable by the largest matrix entry before Cholesky factorizing.  That way if we set `RELATIVE_CHOLESKY_JITTER = 1/2` then jitter will only affect matrix entries less than half the size of the max.  And it kindof makes sense to me that each additional bit of precision would mean we would need to add half as much jitter, thus jitter would be proportional to `finfo(x.dtype).eps`.  Mostly the proportional think helps us keep a constant `RELATIVE_CHOLESKY_JITTER` across float32 and float64.

What's your intuition behind the square here, is that to keep constant error post-Cholesky-factorization?

fritzo:
> @martinjankowiak CHOLESKY_JITTER isn't actually easily toggle-able is it?

That's correct, I'm hoping we won't actually need to toggle it and I'm intending the global variable as an emergency switch in case I need to change something post-release in a production model and want to avoid monkey patching.

Actually we ought to have some sort of standard interface for all of Pyro's global settings, similar to GPyTorch's [settings](https://docs.gpytorch.ai/en/stable/settings.html).  Mind if I attempt that in a follow-up PR?

pyro/ops/tensor_utils.py
fehiepsi: I missed the `x_max` term in the above comment. Using square seems to be more consistent w.r.t. to cases `x.size(-1) > 1` - but I like your clamp by `tiny` better.

Re `x_max`: using global `max` makes sense, but I feel that it might be better to use `max` of rows instead, e.g. considering the diagonal matrix `[0.0001, 10000]`, the global jitter is large w.r.t. the first diagonal term.

fritzo: Nice idea!  I've switched to using a row-wise max.  This required increasing `CHOLESKY_RELATIVE_JITTER` from 1.0 to 4.0, but this way still seems better :+1:

fehiepsi (approved):

fritzo:
Thanks for reviewing!

pyro/ops/tensor_utils.py
martinjankowiak: is this really preferred? this changes the eigenvalues and eigenvectors as opposed to jitter that is proportional to the identity (which only changes eigenvalues)

fritzo: @martinjankowiak that's a good point, I didn't know about eigenvalue preservation.  I'd be ok with either version.

One thing I like about @fehiepsi's solution is that users can on their side rotate the system before performing Gaussian ops, e.g. I'm approximately diagonalizing via QR
```python
evals, evecs = torch.linalg.eig(transition_matrix)
Q, R = torch.linalg.qr(evecs.real)
transition_matrix = Q.T @ transition_matrix @ Q
```
which shrinks my diagonal perturbations
```diff
- [5.45, 1.81, 0.86, 0.76] * eps * CHOLESKY_RELATIVE_JITTER
+ [4.91, 1.67, 0.21, 0.17] * eps * CHOLESKY_RELATIVE_JITTER
```

PR #3157, opened 2022-11-22
Adds min_stepsize and max_stepsize to HMC, avoiding infinite loop

### LINKED ISSUE #3156: [Bug] Infinite loop in `HMC._find_reasonable_step_size`
### Issue Description
In some edge cases, the while loop in `HMC._find_reasonable_step_size` never exits and becomes an infinite loop. This happens because the `direction` never changes and the `step_size` eventually becomes `0`. The below printout of `t` & `step_size` shows an example of this. 

Screen Shot 2022-11-18 at 11 34 59 PM

### Environment
 - OS and python version: CentOS 8, Python 3.8
 - PyTorch version, or if relevant, output of `pip freeze`. Latest main.
 - Pyro version: output of `python -c 'import pyro; print pyro.__version__'` 1.5.2 (checked that the error is not fixed in latest)

### Code Snippet
Unfortunately, I do not have a clean code snippet that reproduces the issue. I have extracted the `HMC` object and the input `z` that produces the infinite loop and used `torch.save` to save them. With the attached output, run
```
import torch
save_dict = torch.load("/home/saitcakmak/pyro_debug.pt")  # update the file path
save_dict["hmc"]._find_reasonable_step_size(save_dict["z"])
```
[pyro_debug.pt.zip](https://github.com/pyro-ppl/pyro/files/10060315/pyro_debug.pt.zip)

### Quick fix
For a hot-fix, I updated the following line
https://github.com/pyro-ppl/pyro/blob/8b7e5641228b664b3c9b674a079e95a478f3e2a0/pyro/infer/mcmc/hmc.py#L193
with `while direction_new == direction and step_size > 1e-100:` which prevents the infinite loop.

cc @dme65, @balandat

fritzo:
Fixes #3156

### DIFF
--- a/pyro/infer/mcmc/hmc.py
+++ b/pyro/infer/mcmc/hmc.py
@@ -66,6 +66,8 @@ class HMC(MCMCKernel):
         step size, hence the sampling will be slower and more robust. Default to 0.8.
     :param callable init_strategy: A per-site initialization function.
         See :ref:`autoguide-initialization` section for available functions.
+    :param min_stepsize (float): Lower bound on stepsize in adaptation strategy.
+    :param max_stepsize (float): Upper bound on stepsize in adaptation strategy.
 
     .. note:: Internally, the mass matrix will be ordered according to the order
         of the names of latent variables, not the order of their appearance in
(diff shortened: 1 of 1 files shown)


martinjankowiak (approved):
lgtm

PR #3159, opened 2022-11-24
Allow rendering of semisupervised models

fritzo:
Addresses [this forum post](https://forum.pyro.ai/t/how-to-render-a-semi-supervised-model/4751)

This uses color gradients.  It doesn't look perfect, but it's a good first pass. For example:
```python
def model(x, y=None):
    with pyro.plate("N", 2):
        z = pyro.sample("z", dist.Normal(0, 1))
        y = pyro.sample("y", dist.Normal(0, 1), obs=y)
        pyro.sample("x", dist.Normal(y + z, 1), obs=x)

pyro.render_model(
    model,
    model_kwargs=[
        {"x": torch.zeros(2)},
        {"x": torch.zeros(2), "y": torch.zeros(2)},
    ]
)
```
image

## Tested
- [x] added a new section to the [rendering tutorial](https://github.com/pyro-ppl/pyro/blob/render-semisupervised/tutorial/source/model_rendering.ipynb)


fehiepsi:
This looks pretty cool!! I'll take a pass soon

pyro/infer/inspect.py
```diff
@@ -435,6 +437,31 @@ def generate_graph_specification(
...
+    if len(things) == 1:
+        return things[0]
+
+    # Recurse into dicts.
+    if isinstance(things[0], dict):
+        result = {}
+        for thing in things:
+            for key, value in thing.items():
+                if key not in result:
+                    result[key] = _deep_merge([t[key] for t in things])
```
fehiepsi, line 450: The logic is a little bit complicated to follow so it would be nice to add a test for `_deep_merge`.

fehiepsi (approved):
Thanks, Fritz!

PR #3163, opened 2022-12-10
add GroupedNormalNormal distribution

martinjankowiak:
Conjugate distribution for Normally distributed data that is controlled by a latent mean that is also controlled by a Normal distribution. Supports ragged grouped data.

### DIFF
--- a/pyro/distributions/grouped_normal_normal.py
+++ b/pyro/distributions/grouped_normal_normal.py
@@ -1,3 +1,6 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
 import math
 
 import torch
--- a/pyro/distributions/grouped_normal_normal.py
+++ b/pyro/distributions/grouped_normal_normal.py
@@ -24,15 +24,29 @@ class GroupedNormalNormal(TorchDistribution):
 
     See e.g. Eqn. (55) in ref. [1] for relevant expressions in a simpler case with scalar `obs_scale`.
 
+    Example:
+
+    >>> num_groups = 3
+    >>> num_data = 4
+    >>> prior_loc = torch.randn(num_groups)
+    >>> prior_scale = torch.rand(num_groups)
+    >>> obs_scale = torch.rand(num_data)
+    >>> group_idx = torch.tensor([1, 0, 2, 1]).long()
+    >>> values = torch.randn(num_data)
+    >>> gnn = GroupedNormalNormal(prior_loc, prior_scale, obs_scale, group_idx)
+    >>> assert gnn.log_prob(values).shape == ()
+
     References:
     [1] "Conjugate Bayesian analysis of the Gaussian distribution," Kevin P. Murphy.
 
-    Args:
-        prior_loc: (num_groups,)
-        prior_scale: (num_groups,)
-        obs_scale: (num_data,)
-        group_idx: (num_data,)
-        value: (num_data,)
+    :param torch.Tensor prior_loc: Tensor of shape `(num_groups,)` specifying the prior mean of the latent
+        of each group.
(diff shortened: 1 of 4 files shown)


pyro/distributions/grouped_normal_normal.py
```diff
@@ -0,0 +1,142 @@
...
+        obs_scale_sq_inv_sum = torch.zeros_like(self.prior_loc).scatter_add(
+            0, self.group_idx, obs_scale_sq_inv
+        )
+        precision = prior_scale_sq_inv + obs_scale_sq_inv_sum
+        scaled_value_sum = torch.zeros_like(self.prior_loc).scatter_add(
+            0, self.group_idx, value * obs_scale_sq_inv
+        )
+
+        loc = (scaled_value_sum + self.prior_loc * prior_scale_sq_inv) / precision
+        scale = precision.pow(-0.5)
```
fritzo, line 108: nit: I believe `.rsqrt()` is cheaper

pyro/distributions/grouped_normal_normal.py
```diff
@@ -0,0 +1,142 @@
...
+        self.num_data_per_batch = prior_loc.new_zeros(self.num_groups).scatter_add(
+            0, self.group_idx, prior_loc.new_ones(self.group_idx.shape)
+        )
+        super().__init__(batch_shape, validate_args=validate_args)
+
+    def expand(self, batch_shape, _instance=None):
+        raise NotImplementedError
+
+    def sample(self, sample_shape=()):
+        raise NotImplementedError
```
fritzo, line 88: Gosh it seems a shame not to implement at least a generator for test data
```python
has_rsample = True
    group_part = Normal(self.prior_loc, self.prior_scale).rsample(sample_shape)
    obs_part = Normal(0, self.obs_scale).rsample(sample_shape)
    return group_part[..., self.group_idx] + obs_part
```
but I guess it's fine to omit it until someone needs it 🤔

tests/distributions/test_grouped_normal_normal.py
```diff
@@ -0,0 +1,43 @@
...
+
+import math
+
+import torch
+
+from pyro.distributions import GroupedNormalNormal, Normal
+from tests.common import assert_close
+
+
+def test_grouped_normal_normal(num_groups=3, num_samples=10 * 6):
```
fritzo, line 12: did you mean `10 ** 6`?  would that allow tighter `atol` below?

pyro/distributions/grouped_normal_normal.py
```diff
@@ -0,0 +1,142 @@
...
+        "prior_scale": constraints.positive,
+        "obs_scale": constraints.positive,
+    }
+    support = constraints.real
+
+    def __init__(
+        self, prior_loc, prior_scale, obs_scale, group_idx, validate_args=None
+    ):
+        prior_loc, prior_scale = broadcast_all(prior_loc, prior_scale)
+        obs_scale, group_idx = broadcast_all(obs_scale, group_idx)
```
fritzo, line 62: Could you add some shape assertions to clarify requiremets?  I'm guessing:
```python
assert prior_loc.dim() == 1
assert obs_scale.dim() == 1
```

fritzo (commented):
Looks good, just a couple minor comments

pyro/distributions/grouped_normal_normal.py
martinjankowiak: yeah the thing is the only distribution that a user is actually likely to care about is the posterior predictive and not the prior predictive

tests/distributions/test_grouped_normal_normal.py
martinjankowiak: good catch 👍

martinjankowiak:
thanks should be good to go now 👍

fritzo (approved):

Issue #3172, opened 2023-01-12
Metaclass conflict while importing AutoGuideList with PyTorch nightly


saitcakmak:
### Issue Description
Due to a change ([commit](https://github.com/pytorch/pytorch/commit/d0bfd79f3d1bbf8885b00acb6d72db0bc16f1995)) in the inheritance of `torch.nn.ModuleList`, attempting to import pretty much anything from Pyro leads to the following error:

```
    import pyro.distributions as base_distributions
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/pyro/__init__.py:8: in <module>
    from pyro.primitives import (
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/pyro/primitives.py:13: in <module>
    import pyro.infer as infer
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/pyro/infer/__init__.py:11: in <module>
    from pyro.infer.mcmc.api import MCMC
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/pyro/infer/mcmc/__init__.py:5: in <module>
    from pyro.infer.mcmc.api import MCMC, StreamingMCMC
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/pyro/infer/mcmc/api.py:28: in <module>
    from pyro.infer.mcmc.hmc import HMC
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/pyro/infer/mcmc/hmc.py:13: in <module>
    from pyro.infer.autoguide import init_to_uniform
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/pyro/infer/autoguide/__init__.py:10: in <module>
    from pyro.infer.autoguide.gaussian import AutoGaussian
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/pyro/infer/autoguide/gaussian.py:24: in <module>
    from .guides import AutoGuide
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/pyro/infer/autoguide/guides.py:183: in <module>
    class AutoGuideList(AutoGuide, nn.ModuleList):
E   TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

[...]

PR #3173, opened 2023-01-12
Fix metaclass conflict with ModuleList

saitcakmak:
- https://github.com/pytorch/pytorch/pull/89135 changed the inheritance of `torch.nn.ModuleList`. This changed the metaclass of `ModuleList` from `type` to `abc.ABCMeta` (print `type(ModuleList)` before & after the change).
- `PyroModule` has its own metaclass `_PyroModuleMeta`. Attempting to create a class (`AutoGuideList`) with bases that have different metaclasses leads to: `TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases`
- The solution for this is to create a new metaclass that inherits from both metaclasses, which is the `_PyroModuleListMeta` introduced here. 
- Note that changing `_PyroModuleMeta` to inherit from `ABCMeta` does not work as this leads to a different `TypeError` (something with MRO conflict). If this worked, it'd be a more general solution.

Fixes https://github.com/pyro-ppl/pyro/issues/3172

### DIFF
--- a/pyro/infer/autoguide/guides.py
+++ b/pyro/infer/autoguide/guides.py
@@ -32,7 +32,7 @@ from pyro.distributions import constraints
 from pyro.distributions.transforms import affine_autoregressive, iterated
 from pyro.distributions.util import eye_like, sum_rightmost
 from pyro.infer.enum import config_enumerate
-from pyro.nn.module import PyroModule, PyroParam
+from pyro.nn.module import PyroModule, PyroModuleList, PyroParam
 from pyro.ops.hessian import hessian
 from pyro.ops.tensor_utils import periodic_repeat
 from pyro.poutine.util import site_is_subsample
@@ -180,7 +180,7 @@ class AutoGuide(PyroModule):
         raise NotImplementedError
 
 
-class AutoGuideList(AutoGuide, nn.ModuleList):
+class AutoGuideList(AutoGuide, PyroModuleList):
     """
     Container class to combine multiple automatic guides.
 
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -262,7 +262,7 @@ class _PyroModuleMeta(type):
         return result
 
 
-class _PyroModuleListMeta(_PyroModuleList, ABCMeta): pass
+class _PyroModuleListMeta(_PyroModuleMeta, ABCMeta): pass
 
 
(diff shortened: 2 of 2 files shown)


saitcakmak:
Creating a new metaclass & a PyroModuleList fixed it. `import pyro` now works with pytorch nightly.

Balandat:
Could you add some comments in code why this is needed? I doubt this will be clear to people just from the code itself.

fritzo:
Thanks for suggesting a fix.  Can you please explain:
- the error this fixes,
- why `PyroModule[torch.nn.ModuleList]` fails as expected, and
- what torch versions trigger the error?

In particular I'd like to understand whether it would be possible to implement a more general solution to fix the `PyroModule[-]` type operator.

saitcakmak:
I'm not an expert in metaclasses but here's my understanding from working on this fix:
- https://github.com/pytorch/pytorch/pull/89135 changed the inheritance of `torch.nn.ModuleList`. This changed the metaclass of `ModuleList` from `type` to `abc.ABCMeta` (print `type(ModuleList)` before & after the change).
- `PyroModule` has its own metaclass `_PyroModuleMeta`. Attempting to create a class (`AutoGuideList`) with bases that have different metaclasses leads to: `TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases`
- The solution for this is to create a new metaclass that inherits from both metaclasses, which is the `_PyroModuleListMeta` introduced here. 
- Note that changing `_PyroModuleMeta` to inherit from `ABCMeta` does not work as this leads to a different `TypeError` (something with MRO conflict). If this worked, it'd be a more general solution.

saitcakmak:
https://github.com/pytorch/pytorch/pull/89135 is being reverted, so the fix is not needed anymore.

fritzo:
@saitcakmak thanks for explaining!  Glad this was averted.  I vaguely recall some challenges in `PyroModule`'s metaclass hierarchy: IIRC we tried to inherit from `ABCMeta` at some point but found that caused more headaches than it resolved, so we offload that work to subclasses, e.g. [EasyGuide](https://github.com/pyro-ppl/pyro/blob/b7ba915b851/pyro/contrib/easyguide/easyguide.py#L22-L26).  And thanks for the PR, this will at least serve as starting point next time we face a similar metaclass issue.

PR #3177, opened 2023-02-02
fix: typo in log-absolute-derivative of exponential transform (normalizing flow tutorial) closes #3176

### LINKED ISSUE #3176: Small typo in tutorial notebook on Normalizing Flows
Hi,

there seems to be a small typo in the [tutorial notebook on normalizing flows](https://github.com/pyro-ppl/pyro/blob/dev/tutorial/source/normalizing_flows_i.ipynb) in section "Fixed Univariate Transforms in Pyro".

The log-absolute derivative of the exponential transform $y = g(x) = \exp(x)$ is stated as $\log(\vert\frac{dg}{dx}\vert)=y$ while it should rather be $\log(\vert\frac{dg}{dx}\vert)=x$ (see also the implementation of [torch.distributions.transforms.ExpTransform](https://pytorch.org/docs/master/_modules/torch/distributions/transforms.html#ExpTransform)).

felixschltr:
This PR addresses Issue #3176

I ran `make test` and `make test-examples` locally just to be sure, but since only a single character inside a notebook markdown cell is changed, there shouldn't be bigger problems.

### DIFF
--- a/tutorial/source/normalizing_flows_i.ipynb
+++ b/tutorial/source/normalizing_flows_i.ipynb
@@ -126,7 +126,7 @@
     "\\begin{align}\n",
     "g(x) &= \\text{exp(x)}\\\\\n",
     "g^{-1}(y) &= \\log(y)\\\\\n",
-    "\\log\\left(\\left|\\frac{dg}{dx}\\right|\\right) &= y.\n",
+    "\\log\\left(\\left|\\frac{dg}{dx}\\right|\\right) &= x.\n",
     "\\end{align}\n",
     "\n",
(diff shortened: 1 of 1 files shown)


fritzo (approved):
Thanks for the fix!

felixschltr:
Sorry about the failing lint check. Though I'm kind of surprised by this given the small change. Should I request a revert?

ordabayevy:
I think linting is failing because of the new version of the black formatter.

martinjankowiak:
yes let's see if #3178 fixes it

PR #3199, opened 2023-04-22
Deprecate `CorrLCholeskyTransform` in favor of  upstream `CorrCholeskyTransform`

### LINKED ISSUE #3181: Difference in event_shape between pyro's LKJCholesky and torch's
### Issue Description
Event shape of inverse transform of `LKJCholesky` seems to differ from torch's, and of samples from distribution.

### Environment
 - WSL2 Ubuntu 20.04
 - Python 3.9
 - torch 1.13.1
 - pyro 1.8.4+9ed468d

### Code Snippet
```python
from torch.distributions import TransformedDistribution, LKJCholesky
from torch.distributions import biject_to

# torch
torch_lkj = LKJCholesky(2, 0.5)
torch_inv = biject_to(torch_lkj.support).inv
torch_transformed = TransformedDistribution(torch_lkj, torch_inv)

# pyro
from pyro.distributions import LKJCholesky as PyroCholesky
from pyro import __version__

print(__version__)

pyro_lkj = PyroCholesky(2, 0.5)
pyro_inv = biject_to(pyro_lkj.support).inv
pyro_transformed = TransformedDistribution(pyro_lkj, pyro_inv)

assert pyro_transformed.event_shape == torch_transformed.event_shape
```
Stepping through the code, it seems as though `pyro.distributions.transforms.cholesky.CorrLCholeskyTransform` doesn't implement the `inverse_shape` method, whereas torch does with `torch.distributions.transforms.CorrCholeskyTransform`. This results in pyro returning an event shape corresponding to `torch.Size([2])`, wheras torch returns `torch.Size([1])`.

ordabayevy:
Closes #3181

### DIFF
--- a/pyro/distributions/transforms/__init__.py
+++ b/pyro/distributions/transforms/__init__.py
@@ -161,7 +161,6 @@ __all__ = [
     "ConditionalRadial",
     "ConditionalSpline",
     "ConditionalSplineAutoregressive",
-    "CorrLCholeskyTransform",
     "CorrMatrixCholeskyTransform",
     "DiscreteCosineTransform",
     "ELUTransform",
--- a/pyro/distributions/transforms/__init__.py
+++ b/pyro/distributions/transforms/__init__.py
@@ -30,7 +30,11 @@ from .affine_coupling import (
 from .basic import ELUTransform, LeakyReLUTransform, elu, leaky_relu
 from .batchnorm import BatchNorm, batchnorm
 from .block_autoregressive import BlockAutoregressive, block_autoregressive
-from .cholesky import CholeskyTransform, CorrMatrixCholeskyTransform
+from .cholesky import (
+    CholeskyTransform,
+    CorrCholeskyTransform,
+    CorrMatrixCholeskyTransform,
+)
 from .discrete_cosine import DiscreteCosineTransform
 from .generalized_channel_permute import (
     ConditionalGeneralizedChannelPermute,
--- a/pyro/distributions/transforms/__init__.py
+++ b/pyro/distributions/transforms/__init__.py
@@ -33,6 +33,7 @@ from .block_autoregressive import BlockAutoregressive, block_autoregressive
 from .cholesky import (
     CholeskyTransform,
     CorrCholeskyTransform,
+    CorrLCholeskyTransform,
     CorrMatrixCholeskyTransform,
 )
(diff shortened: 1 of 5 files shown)


pyro/distributions/transforms/__init__.py
```diff
@@ -171,7 +165,6 @@ def iterated(repeats, base_fn, *args, **kwargs):
     "ConditionalRadial",
     "ConditionalSpline",
     "ConditionalSplineAutoregressive",
-    "CorrLCholeskyTransform",
```
fritzo, line 174: Shouldn't the new CorrCholeskyTransform apper in this list?

pyro/distributions/transforms/__init__.py
```diff
@@ -32,7 +32,7 @@
 from .block_autoregressive import BlockAutoregressive, block_autoregressive
 from .cholesky import (
     CholeskyTransform,
-    CorrLCholeskyTransform,
```
fritzo, line 35: I'd lean towards preserving the old import so as to avoid breaking existing code, but I'll defer to your judgement if you think we should delete both.  (I think of "deprecation" as meaning "it still works but we no longer recommend it", whereas removing from `__init__.py` would be a truly breaking change)

pyro/distributions/transforms/cholesky.py
```diff
@@ -2,89 +2,21 @@
...
+import warnings
 
 import torch
-from torch.distributions.transforms import Transform
+from torch.distributions.transforms import CorrCholeskyTransform, Transform
 
 from .. import constraints
 
 
-def _vector_to_l_cholesky(z):
```
fritzo, line 12: Woo hoo less math to maintain 📉

fritzo (commented):
Looks good after addressing a couple comments 👍

pyro/distributions/transforms/__init__.py
ordabayevy: It should be added with other PyTorch transforms below:

```
__all__.extend(torch_transforms)
```

ordabayevy (commented):
Thanks for reviewing @fritzo . Addressed your comments.

ordabayevy:
Hi @fritzo. Can you have another look at this PR?

fritzo (approved):

PR #3206, opened 2023-05-05
fix saving the param store with PyTorch 2 (closes #3201)

### LINKED ISSUE #3201: [bug] param store's `.save()` method fails with pytorch 2.0.0
### Issue Description
The param store dict's save method seems to have been messed up by pytorch version 2.0.0

with
```
pyro-ppl  1.8.4+dd4e0f81
torch     1.13.0
```
if I run
```python
import pyro
import torch


def model():
    x_scale = pyro.param("x_scale", torch.tensor(1.0),
    z = pyro.sample("z", pyro.distributions.Normal(0, 1))
    return pyro.sample("x", pyro.distributions.Normal(z, x_scale), obs=torch.tensor(0.1))


def guide():
    z_loc = pyro.param("z_loc", torch.tensor(0.0))
    z_scale = pyro.param(
     "z_scale", torch.tensor(0.5), constraint=constraints.positive
    )
    pyro.sample("z", pyro.distributions.Normal(z_loc, z_scale))


optimizer = pyro.optim.ClippedAdam({"lr": 0.01})
svi = pyro.infer.SVI(model, guide, optimizer, pyro.infer.Trace_ELBO())
for _ in range(10):
    svi.step()

store = pyro.get_param_store()
store.save("tmp.pt")
```
I see no error, and the param store can be saved just fine.

However, with 
```
pyro-ppl  1.8.4+dd4e0f81
torch     2.0.0
```
if I run the same thing, I get
```pytb
  File "/Users/sfleming/opt/anaconda3/envs/cellbender38/lib/python3.8/site-packages/pyro/params/param_store.py", line 279, in save
  File "/Users/sfleming/opt/anaconda3/envs/cellbender38/lib/python3.8/site-packages/torch/serialization.py", line 441, in save
    _save(obj, opened_zipfile, pickle_module, pickle_protocol)
  File "/Users/sfleming/opt/anaconda3/envs/cellbender38/lib/python3.8/site-packages/torch/serialization.py", line 653, in _save
    pickler.dump(obj)
TypeError: cannot pickle 'weakref' object
```

### Environment
 - macOS Monterey
 - Python 3.8
 - pytorch and pyro versions as above

ilia-kats:
In ParamStoreDict.__getitem__, we add a weakref to the unconstrained value to the transformed parameter. Apparently, when the constrained is Real(), transform_to returns the same object that it was passed, so we end up with weakrefs in ParamStoreDict._params. This was fine with PyTorch1, but under Pytorch2 an exception is raised when saving a Parameter with a weakref. So let's explicitly remove the .unconstrained attribute from all parameters to be saved

### DIFF
--- a/pyro/params/param_store.py
+++ b/pyro/params/param_store.py
@@ -243,8 +243,9 @@ class ParamStoreDict:
         """
         Get the ParamStore state.
         """
+        params = {name: param[...] for name, param in self._params.items()}
         state = {
-            "params": self._params.copy(),
+            "params": params,
             "constraints": self._constraints.copy(),
         }
         return state


fritzo:
Thanks for fixing this!

Hmm, I'm concerned that if we delete the `.unconstrained()` from the parameters, it will corrupt the param store. While this will work for the workflow `save; restart; load`, it will break the checkpointing workflow `train; save; train; save;...; restart; load`.

WDYT of this slight modification of your fix
```python
# Create shallow copies of parameters to remove .unconstrained() attributes.
params = {name: param[...] for name, param in self._params.items()}
```

fritzo (approved):

fritzo:
Hmm, looks like some tests are still failing. Let me look into it...

fritzo:
Hey Thanks for your help fixing this @ilia-kats.  Following your lead I found more weakref issues and fixed them all at once in #3212. Close this PR, but it #3212 wouldn't have happened without your help!

PR #3192, opened 2023-03-20
Run CI with PyTorch 2.0

eb8680:
This is a dummy PR to trigger a test run with the new PyTorch 2.0 release and see what breaks.

Log from running `pytest -v --stage=unit --tb=no -n auto` on my local machine (Ubuntu 22.04, Python 3.10.9, `torch==2.0.0`):
<details>

```
=========================================================================================================== warnings summary ============================================================================================================
pyro/generic.py:9: 24 warnings
  /home/eli/development/pyro/pyro/generic.py:9: DeprecationWarning: pyro.generic has moved to the pyroapi package
    warnings.warn("pyro.generic has moved to the pyroapi package", DeprecationWarning)

tests/distributions/test_stable.py:8: 24 warnings
  /home/eli/development/pyro/tests/distributions/test_stable.py:8: DeprecationWarning: Please use `IntegrationWarning` from the `scipy.integrate` namespace, the `scipy.integrate.quadpack` namespace is deprecated.
    from scipy.integrate.quadpack import IntegrationWarning

tests/ops/test_contract.py: 165 warnings
  /home/eli/development/pyro/pyro/ops/contract.py:433: DeprecationWarning: 'ubersum' is deprecated, use 'pyro.ops.contract.einsum' instead
    warnings.warn(

tests/infer/test_predictive.py::test_posterior_predictive_svi_auto_delta_guide[False]
tests/infer/test_predictive.py::test_posterior_predictive_svi_auto_delta_guide[True]
tests/infer/test_predictive.py::test_posterior_predictive_svi_auto_diag_normal_guide[False]
tests/infer/test_predictive.py::test_posterior_predictive_svi_one_hot
  /home/eli/development/pyro/pyro/infer/predictive.py:284: DeprecationWarning: The method `.get_samples` has been deprecated in favor of `.forward`.
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
======================================================================================================== short test summary info ========================================================================================================
FAILED tests/distributions/test_rejector.py::test_rejector[0.25-0.5] - AssertionError: bug in .rsample()
FAILED tests/distributions/test_stable.py::test_additive[1.01-0.1-0.9--0.5--0.9] - assert 0.04126178292409227 > 0.05
FAILED tests/distributions/test_stable.py::test_additive[0.99-0.1-0.9--0.5--0.5] - assert 0.03966362560959423 > 0.05
[...]

PR #3212, opened 2023-05-17
Clear .unconstrained weakrefs before pickling; rebuild them more often

### LINKED ISSUE #3201: [bug] param store's `.save()` method fails with pytorch 2.0.0

fritzo:
Resolves #3201 
Replaces #3206

This builds on @ilia-kats's diagnosis of why `torch.save` is failing with various Pyro objects in torch>=2, namely that the `.unconstrained` attributes of tensors are breaking pickling.  This PR:
- removes `.unconstrained` in one other place,
- rebuilds `.unconstrained` in one place,
- fixes `__getstate__` in a number of places to avoid serializing `weakref.ref`s
- fixes one module_local_param comment.

## Tested
- ran serialization tests locally under torch==2.0.1

### DIFF
--- a/pyro/contrib/util.py
+++ b/pyro/contrib/util.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2017-2019 Uber Technologies, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
+import weakref
 from collections import OrderedDict
 
 import torch
@@ -75,3 +76,22 @@ def iter_plates_to_shape(shape):
     # Go backwards (right to left)
     for i, s in enumerate(shape[::-1]):
         yield pyro.plate("plate_" + str(i), s)
+
+
+def check_no_weakref(obj, path="", avoid_ids=None):
+    """Attempts to check that an object has no weakrefs."""
+    if avoid_ids is None:
+        avoid_ids = {id(obj)}
+
+    if isinstance(obj, weakref.ref):
+        raise ValueError(f"Weakref found at {path}")
+    elif isinstance(obj, dict):
+        for k, v in obj.items():
+            if id(v) not in avoid_ids:
+                check_no_weakref(v, path + f"[{k}]")
+    elif isinstance(obj, (list, tuple)):
+        for i, v in enumerate(obj):
+            if id(v) not in avoid_ids:
+                check_no_weakref(v, path + f"[{i}]")
+    elif hasattr(obj, "__dict__"):
+        check_no_weakref(obj.__dict__, path)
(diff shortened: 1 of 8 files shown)


pyro/contrib/util.py
```diff
@@ -75,3 +76,22 @@ def iter_plates_to_shape(shape):
     # Go backwards (right to left)
     for i, s in enumerate(shape[::-1]):
         yield pyro.plate("plate_" + str(i), s)
+
+
+def check_no_weakref(obj, path="", avoid_ids=None):
```
fritzo, line 81: This is just a helper to debug similar issues in the future.

eb8680 (approved):
LGTM, thanks for tracking this down. This might also be sufficient to fix #3192

PR #3220, opened 2023-05-25
make ProvenanceTensor behave more like a Tensor (closes #3218)

### LINKED ISSUE #3218: `ProvenanceTensor` bug when used with `torch.set_default_device`
When trying to debug the code from the [forum post](https://forum.pyro.ai/t/running-pyro-plate-on-gpu/5097/8) I came across this bug:

```py
import torch
from pyro.ops.provenance import ProvenanceTensor

device = torch.device("cuda")
torch.set_default_device(device)
x = torch.tensor([1., 2., 3.])
y = ProvenanceTensor(x, frozenset(["x"]))
print(torch.as_tensor(y))
```
returns `tensor([], device='cuda:0')`

**Two observations:**
1. This doesn't happen if `torch.set_default_device` is not used: `print(torch.as_tensor(y.cuda()))` works fine
2. When `torch.set_default_device` is used then the following code gets invoked
https://github.com/pytorch/pytorch/blob/main/torch/utils/_device.py#L72-L76
where `func(*args, **kwargs)` returns `tensor([], device='cuda:0')`. This doesn't happen when `.cuda()` is used like in 1).

ilia-kats:
The data is now stored in the Tensor itself instead of an attribute. This fixes torch.to_tensor returning empty tensors when called with a ProvenanceTensor and and a device as arguments.

This is super hacky, but I couldn't come up with a cleaner way. Note that this is the only way to use `pyro.infer.inspect.get_dependencies` when training on GPUs (I'm using it in a custom Messenger guide), since the `log_prob` function of some distributions (for example Gamma) calls `torch.to_tensor`.

### DIFF
--- a/pyro/ops/provenance.py
+++ b/pyro/ops/provenance.py
@@ -59,6 +59,8 @@ class ProvenanceTensor(torch.Tensor):
         if isinstance(data, ProvenanceTensor):
             provenance |= data._provenance
         self._provenance = provenance
+        if hasattr(data, "unconstrained"):
+            self._t.unconstrained = data.unconstrained
 
     def __repr__(self):
         return "Provenance:\n{}\nTensor:\n{}".format(self._provenance, self._t)
--- a/pyro/ops/provenance.py
+++ b/pyro/ops/provenance.py
@@ -47,7 +47,7 @@ class ProvenanceTensor(torch.Tensor):
         if not provenance:
             return data
         ret = data.view(data.shape)
-        ret._t = data.view(data.shape)  # this makes sure that detach_provenance always
+        ret._t = data  # this makes sure that detach_provenance always
         # returns the same object. This is important when
         # using the tensor as key in a dict, e.g. the global
         # param store
--- a/pyro/ops/provenance.py
+++ b/pyro/ops/provenance.py
@@ -46,18 +46,15 @@ class ProvenanceTensor(torch.Tensor):
         assert not isinstance(data, ProvenanceTensor)
         if not provenance:
             return data
-        ret = data.view(data.shape)
+        ret = data.as_subclass(cls)
         ret._t = data  # this makes sure that detach_provenance always
         # returns the same object. This is important when
         # using the tensor as key in a dict, e.g. the global
         # param store
-        ret.__class__ = cls
         return ret
 
     def __init__(self, data, provenance=frozenset()):
         assert isinstance(provenance, frozenset)
-        if isinstance(data, ProvenanceTensor):
-            provenance |= data._provenance
         self._provenance = provenance
 
     def __repr__(self):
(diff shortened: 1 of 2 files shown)


pyro/ops/provenance.py
```diff
@@ -46,15 +46,21 @@ def __new__(cls, data: torch.Tensor, provenance=frozenset(), **kwargs):
         assert not isinstance(data, ProvenanceTensor)
         if not provenance:
             return data
-        return super().__new__(cls)
+        ret = data.view(data.shape)
+        ret._t = data.view(data.shape)  # this makes sure that detach_provenance always
```
fritzo, line 50: Could this line be simplified to
```python
ret._t = data
```
or would that break something?

fritzo (commented):
Thanks for this subtle fix!  This looks good to me (after one minor comment), but I'm unsure how this will interact with other subclasses of tensor.

@ordabayevy could you also take a look as you've thought about this before?

fritzo:
Would you be able to add a regression test and decorate it with @requires_cuda?  It won't run on CI, but it might help future maintainers of `ProvenanceTensor` preserve your intent in the face of future changes.

pyro/ops/provenance.py
ilia-kats: Right, thanks. Took me about four tries to get all the tests to pass, this was still a remnant of an earlier attempt.

fritzo (approved):
LGTM, thanks for adding a test.

I'll leave this up a couple days before merging in case @ordabayevy has any comments.

ordabayevy:
Thanks for holding it up. I'll have a look at this later tonight.

ordabayevy:
@ilia-kats thanks for fixing this!

What about trying to use `torch.Tensor._make_subclass` as is used by [torch.nn.Parameter](https://pytorch.org/docs/stable/_modules/torch/nn/parameter.html#Parameter):

```diff
class ProvenanceTensor(torch.Tensor):
         assert not isinstance(data, ProvenanceTensor)
         if not provenance:
             return data
-        return super().__new__(cls)
+        return torch.Tensor._make_subclass(cls, data)
```

And I believe we can remove instance check from the `__init__` method since `data` being a `ProvenanceTensor` is prohibited in the `__new__` method anyways? WDYT @fritzo @ilia-kats ?

```diff
     def __init__(self, data, provenance=frozenset()):
         assert isinstance(provenance, frozenset)
-        if isinstance(data, ProvenanceTensor):
-            provenance |= data._provenance
-            data = data._t
         self._t = data
         self._provenance = provenance
```

ilia-kats:
@ordabayevy Thanks for the comment. I actually `Tensor.as_subclass` in the public API, which also seems to work.

fritzo (approved):

fritzo:
@ordabayevy ready to merge? I'll release today or tomorrow and will include this PR in the release

ordabayevy:
Yeah, lgtm.

PR #3223, opened 2023-05-28
Refactor `ProvenanceTensor` to use `pytree`

ordabayevy:
This PR:

- Removes `__init__` method
- Refactors the code to use `pytree` to track/extract provenance


fritzo (commented):
Do we have any tests of `track_provenance()`?  It'd be nice to have tests before we refactor it.

fritzo (approved):
Thanks for adding tests!

PR #3224, opened 2023-06-01
Create FUNDING.yml

fritzo:
Add @fritzo to github funding targets. For motivation see https://news.ycombinator.com/item?id=36153293


fritzo:
@martinjankowiak @eb8680 @ordabayevy feel free to add your github handles to this list. If you do you'll also need to enroll as a github sponsor, I believe here: https://github.com/sponsors/accounts

PR #3227, opened 2023-06-08
Remove DataParallel container in SS-VAE model

martinrohbeck:
This PR removes the usage of the `DataParallel` container, because it seems to cause issues.

1) The issues only come with cuda enabled, because otherwise the `DataParallel` is not used. When running `python ss_vae_M2.py --cuda` memory is allocated on more than one GPU, but nothing seems to happen.
However, after dropping the `--cuda`, i.e. running `python ss_vae_M2.py`, everything works fine. Code also works with `CUDA_VISIBLE_DEVICE=1 python ss_vae_M2.py --cuda`, hence the multi-gpu training create the trouble.

2) On the other hand, it is recommended to use [DistributedDataParallel](https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html#torch.nn.parallel.DistributedDataParallel), see [here](https://pytorch.org/docs/stable/generated/torch.nn.DataParallel.html).

3) I think the lines can be dropped, since MNIST is not a dataset where multi-GPU training is needed anymore ;).

I installed pyro from the latest dev branch (v1.8.5) and pytorch v2.0.1.
PR also contains minor housekeeping.

### DIFF
--- a/examples/vae/ss_vae_M2.py
+++ b/examples/vae/ss_vae_M2.py
@@ -63,7 +63,7 @@ class SSVAE(nn.Module):
         self.aux_loss_multiplier = aux_loss_multiplier
 
         # define and instantiate the neural networks representing
-        # the paramters of various distributions in the model
+        # the parameters of various distributions in the model
         self.setup_networks()
 
     def setup_networks(self):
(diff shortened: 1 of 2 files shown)


fritzo (approved):
Thanks for cleaning up!

PR #3228, opened 2023-06-14
Updated ZeroInflatedDistribution Class 

StatMixedML:
# Summary
- Changed the import of the NegativeBinomial from pyro to torch, since pyro doesn't have the NegativeBinomial anymore. 
- Updated ZeroInflatedDistribution Class to allow for 0 values in y (=value) that have support only for $(0,\infty]$. 

# Description
This is need for `log_prob = (-gate).log1p() + self.base_dist.log_prob(value)`, since otherwise `self.base_dist.log_prob(value)` returns `inf`.

### DIFF
--- a/pyro/distributions/zero_inflated.py
+++ b/pyro/distributions/zero_inflated.py
@@ -11,9 +11,9 @@ from torch.distributions.utils import (
 )
 from torch.nn.functional import softplus
 
-from torch.distributions import NegativeBinomial, Poisson
-from pyro.distributions import TorchDistribution
+from pyro.distributions import NegativeBinomial, Poisson, TorchDistribution
 from pyro.distributions.util import broadcast_shape
+from pyro.distributions.util import is_identically_one, is_identically_zero
 
 
 class ZeroInflatedDistribution(TorchDistribution):
@@ -76,24 +76,24 @@ class ZeroInflatedDistribution(TorchDistribution):
         zero_idx = (value == 0)
 
         if hasattr(support, "lower_bound"):
-            if support.lower_bound == 0.0:
+            if is_identically_zero(getattr(support, "lower_bound", None)):
                 value = value.clamp_min(epsilon)
 
         if hasattr(support, "upper_bound"):
-            if support.upper_bound == 1.0:
+            if is_identically_one(getattr(support, "upper_bound", None)) & (value.max() == 1.0):
                 value = value.clamp_max(1 - epsilon)
 
         if "gate" in self.__dict__:
             gate, value = broadcast_all(self.gate, value)
-            log_prob = (-gate).log1p() + self.base_dist.log_prob(value)
-            log_prob = torch.where(zero_idx == True, (gate + log_prob.exp()).log(), log_prob)
+            log_prob = torch.log1p(1 - gate) + self.base_dist.log_prob(value)
+            log_prob = torch.where(zero_idx, torch.log1p(gate), log_prob)
         else:
             gate_logits, value = broadcast_all(self.gate_logits, value)
             log_prob_minus_log_gate = -gate_logits + self.base_dist.log_prob(value)
             log_gate = -softplus(-gate_logits)
             log_prob = log_prob_minus_log_gate + log_gate
             zero_log_prob = softplus(log_prob_minus_log_gate) + log_gate
-            log_prob = torch.where(zero_idx == True, zero_log_prob, log_prob)
+            log_prob = torch.where(zero_idx, zero_log_prob, log_prob)
         return log_prob
 
     def sample(self, sample_shape=torch.Size()):
--- a/pyro/distributions/zero_inflated.py
+++ b/pyro/distributions/zero_inflated.py
@@ -11,7 +11,8 @@ from torch.distributions.utils import (
 )
 from torch.nn.functional import softplus
 
-from pyro.distributions import NegativeBinomial, Poisson, TorchDistribution
+from torch.distributions import NegativeBinomial, Poisson
+from pyro.distributions import TorchDistribution
 from pyro.distributions.util import broadcast_shape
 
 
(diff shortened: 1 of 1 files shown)


pyro/distributions/zero_inflated.py
```diff
@@ -70,17 +71,29 @@ def log_prob(self, value):
...
+
+        if hasattr(support, "upper_bound"):
+            if support.upper_bound == 1.0:
+                value = value.clamp_max(1 - epsilon)
+
         if "gate" in self.__dict__:
             gate, value = broadcast_all(self.gate, value)
             log_prob = (-gate).log1p() + self.base_dist.log_prob(value)
-            log_prob = torch.where(value == 0, (gate + log_prob.exp()).log(), log_prob)
+            log_prob = torch.where(zero_idx == True, (gate + log_prob.exp()).log(), log_prob)
```
fritzo, line 89: nit: simplify `zero_idx == True` to `zero_idx`

pyro/distributions/zero_inflated.py
```diff
@@ -70,17 +71,29 @@ def log_prob(self, value):
...
         else:
             gate_logits, value = broadcast_all(self.gate_logits, value)
             log_prob_minus_log_gate = -gate_logits + self.base_dist.log_prob(value)
             log_gate = -softplus(-gate_logits)
             log_prob = log_prob_minus_log_gate + log_gate
             zero_log_prob = softplus(log_prob_minus_log_gate) + log_gate
-            log_prob = torch.where(value == 0, zero_log_prob, log_prob)
+            log_prob = torch.where(zero_idx == True, zero_log_prob, log_prob)
```
fritzo, line 96: ditto

pyro/distributions/zero_inflated.py
```diff
@@ -70,17 +71,29 @@ def log_prob(self, value):
         if self._validate_args:
             self._validate_sample(value)
 
+        support = self.support
+        epsilon = abs(torch.finfo(value.dtype).eps)
```
fritzo, line 75: Just curious, what does the `abs(-)` here do?

pyro/distributions/zero_inflated.py
```diff
@@ -70,17 +71,29 @@ def log_prob(self, value):
...
+        zero_idx = (value == 0)
+
+        if hasattr(support, "lower_bound"):
+            if support.lower_bound == 0.0:
+                value = value.clamp_min(epsilon)
+
+        if hasattr(support, "upper_bound"):
+            if support.upper_bound == 1.0:
+                value = value.clamp_max(1 - epsilon)
```
fritzo, line 84: Let's see, is this sufficiently general?  🤔 

I believe `.upper_bound` and `.lower_bound` can be tensors, which would break this logic.  I guess one workaround would be to use our `is_identically_*()` helpers
```python
from pyro.distributions.util import is_identically_one, is_identically_zero

if is_identically_zero(getattr(support, "lower_bound", None)):  ...
if is_identically_one(getattr(support, "upper_bound", None)):  ...
```
I'm not sure how ZeroInflatedDistribution is used, but I guess you could even allow other scalar bounds via
```python
lb = getattr(support, "lower_bound", None)
if isinstance(lb, (float, int)):
    value = value.clamp_min(lb + epsilon)
ub = ...
```

Another thought is that these zero-inflated distributions should only add a new point at zero (e.g. changing support from `(0,1)` to `[0,1)` or `(-1,0)` to `(-1,0]`), so maybe we should use `is_identically_zero` for both the lower and upper bound checks?

WDYT?

pyro/distributions/zero_inflated.py
```diff
@@ -11,7 +11,8 @@
 )
 from torch.nn.functional import softplus
 
-from pyro.distributions import NegativeBinomial, Poisson, TorchDistribution
+from torch.distributions import NegativeBinomial, Poisson
+from pyro.distributions import TorchDistribution
```
fritzo, line 15: Can you explain more about your motivation here?  Was something breaking?

My concern is that the pyro imports are actually [wrappers](https://github.com/pyro-ppl/pyro/blob/727aff741e105715840bfdafee5bfeda7e8b65e8/pyro/distributions/torch.py#L350) around the torch versions that add a [mixin class](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/torch_distribution.py#L18) providing a bunch of extra functionality.  While I'm unsure how much of this extra functionality is exercised by users of `ZeroInflated` distributions, I'm hesitant to demote these distributions without good reason.

fritzo (changes requested):
Thanks for fixing this!  The bounding logic looks like a reasonable approach, but I'm not sure about the import change.

pyro/distributions/zero_inflated.py
StatMixedML: Yes good one, I change it from `zero_idx == True` to `zero_idx`.

Actually, the reason why I changed the import is that my IDE (Dataspell) indicated that it cannot import the `NegativeBinomial` from pyro. But I just tried it in the console and it is working, so I move back to the initial imports.

Thanks for pointing me to these issues.

pyro/distributions/zero_inflated.py
StatMixedML: changed

pyro/distributions/zero_inflated.py
StatMixedML: I just want to make sure that `epsilon` is positive. Say, for example for a `Beta` distribution, the bounds are (0,1). So if `value=0` and `epsilon<0` then we have a negative `value`.

pyro/distributions/zero_inflated.py
StatMixedML: see my comment above

pyro/distributions/zero_inflated.py
StatMixedML: > from pyro.distributions.util import is_identically_one, is_identically_zero
> if is_identically_zero(getattr(support, "lower_bound", None)):  ...
> if is_identically_one(getattr(support, "upper_bound", None)):  ...

These are actuall good suggestions. But the problem is that not all distributions have a lower/upper bound in their support, e.g., Normal. So we first need to check if it has such a support. This is true for, e.g., the `Beta`.

The same goes for your second comment.

pyro/distributions/zero_inflated.py
fritzo: Well `finfo(...).eps` is always positive, but I guess an `abs(-)` doesn't hurt

StatMixedML:
For the updated version, I slightly changed the `log_prob`. I have also added `if is_identically_one(getattr(support, "upper_bound", None)) & (value.max() == 1.0): ...`. 

I also tried using `is_identically_one(value.max())` but it doesn't seem to work. Here is a small example.

````python
import torch
from pyro.distributions.util import is_identically_one

torch_tens = torch.randn((1000,1))
torch_tens = abs(torch_tens / sum(torch_tens))
torch_tens[200] = 1

if is_identically_one(torch_tens.max()):
    print("True")
else:
    print("False")
False
```

fritzo:
Can you run `black` or `make format` to fix the lint errors?

StatMixedML:
I need to further test the zero-inflated class. So I close the PR for now and potentially re-open it.

PR #3229, opened 2023-06-21
Fix jit ELBO sign error

### LINKED ISSUE #2798: [bug] Sign error in JitTraceGraph_ELBO
Trying out this class and ended up with a rather strange looking loss curve (I was using a LR scheduler which explains the waviness):

elbo_bug

It looks like the training losses have a sign error, although they are still optimizing in the right direction. This bug doesn't appear in `TraceGraph_ELBO` or `JitTrace_ELBO`.

Looking at the code, I think this is just because the JITed class overrides `loss_and_grads` and lacks a negative sign which appears in the non-JIT version. The validation curve is generated via `SVI.evalulate_loss` which calls `TraceGraph_ELBO.loss`  and includes the negation.

So it seems like a simple fix, but before I made a PR I wanted to flag the issue in case I'm missing something. There seem to be negations in a couple places and it seems easy to mess up

fritzo:
Fixes #2798 
Closes #2799

This fixes a sign error in `JitTraceGraph_ELBO` and adds a regression test for other ELBOs.

## Tested
- [x] added a regression test with a golden value ELBO computed via Trace_ELBO with 1e8 particles.

### DIFF
--- a/pyro/ops/provenance.py
+++ b/pyro/ops/provenance.py
@@ -43,6 +43,9 @@ class ProvenanceTensor(torch.Tensor):
     :param frozenset provenance: An initial provenance set.
     """
 
+    _t: torch.Tensor
+    _provenance: frozenset
+
     def __new__(cls, data: torch.Tensor, provenance=frozenset(), **kwargs):
         assert not isinstance(data, ProvenanceTensor)
         if not provenance:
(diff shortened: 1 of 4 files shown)


pyro/ops/provenance.py
```diff
@@ -43,6 +43,9 @@ class ProvenanceTensor(torch.Tensor):
     :param frozenset provenance: An initial provenance set.
     """
 
+    _t: torch.Tensor
+    _provenance: frozenset
+
```
fritzo, line 48: fix for mypy

martinjankowiak (approved):
lgtm. looks like we chose some poor names for some of these functions...

fritzo:
@martinjankowiak Thanks for reviewing. Merge?

Issue #3232, opened 2023-06-22
Incorrect scale calculation for AutoMultivariateNormal


davidaknowles:
I could be wrong but I think the scale calculation for the AutoMultivariateNormal is incorrect. According to
https://docs.pyro.ai/en/stable/_modules/pyro/infer/autoguide/guides.html#AutoMultivariateNormal
it is `self.scale * self.scale_tril.diag()`. However, the scale should be the square root of the diagonal of the covariance, which would be something more like `(self.scale[..., None] * self.scale_tril)**2).sum(-1).sqrt()`. Note these are not equivalent since the latter depends on the off-diagonal elements of scale_tril whereas the former does not.

martinjankowiak:
@davidaknowles can you clarify what you mean by incorrect? from what i recall, `scale` is just a parameter in a non-user-facing parameterization. nowhere is it claimed, e.g., that scale is a vector of marginal square root covariances. i guess your expectation is that it should be?

martinjankowiak:
incidentally we use this parameterization because we found it tends to work quite a bit better than a vanilla `chol @ chol.T` (presumably because it reduces dependence on the ordering)

martinjankowiak:
although i think you're right that the `_loc_scale` method may be implemented incorrectly (which feeds into e.g. the quantiles method)

davidaknowles:
The parameterization seems very reasonable, but I would expect `scale` to correspond to the marginal for each variable (and presumably the marginal is what people expect from e.g. `quantiles`. I'm not sure what the current implementation corresponds to, maybe something like the standard deviation of the posterior for that variable conditioning on all other variables in the MVN (I don't think that's true even: it is the correct marginal calculation for the first variable in the MVN).

fritzo:
@davidaknowles thanks for pointing this out.  This bug may have crept in as we changed parametrizations over the years.  IIRC the original `AutoMultivariateNormal` was nearly unusable due to poor parametrization.

I can add this to my TODO list: adding stricter tests for the internal `._loc_scale()` and fixing any bugs those tests witness.  I'm a little surprised the existing `.quantiles()` tests didn't catch these, so I can try tightening their tolerances and maybe add more correlated examples.

[...]

PR #3233, opened 2023-06-23
fix _loc_scale method in AutoMultivariateNormal

martinjankowiak:
addresses #3232 

cc @fritzo

### DIFF
--- a/tests/infer/test_autoguide.py
+++ b/tests/infer/test_autoguide.py
@@ -527,9 +527,9 @@ def AutoGuideList_x(model):
 @pytest.mark.parametrize("Elbo", [Trace_ELBO, TraceGraph_ELBO, TraceEnum_ELBO])
 def test_quantiles(auto_class, Elbo):
     def model():
-        pyro.sample("x", dist.Normal(0.0, 1.0))
         pyro.sample("y", dist.LogNormal(0.0, 1.0))
         pyro.sample("z", dist.Beta(2.0, 2.0).expand([2]).to_event(1))
+        pyro.sample("x", dist.Normal(0.0, 1.0))
 
     guide = auto_class(model)
     optim = Adam({"lr": 0.05, "betas": (0.8, 0.99)})
@@ -542,6 +542,13 @@ def test_quantiles(auto_class, Elbo):
 
     if auto_class is AutoLaplaceApproximation:
         guide = guide.laplace_approximation()
+
+    if hasattr(auto_class, "get_posterior"):
+        posterior = guide.get_posterior()
+        posterior_scale = posterior.variance[-1].sqrt()
+        q = guide.quantiles([0.158655, 0.8413447])
+        quantile_scale = 0.5 * (q["x"][1] - q["x"][0])  # only x is unconstrained
+        assert_close(quantile_scale, posterior_scale, atol=1.0e-6)
 
     quantiles = guide.quantiles([0.1, 0.5, 0.9])
     median = guide.median()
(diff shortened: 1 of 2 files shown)


tests/infer/test_autoguide.py
```diff
@@ -527,9 +527,9 @@ def AutoGuideList_x(model):
 @pytest.mark.parametrize("Elbo", [Trace_ELBO, TraceGraph_ELBO, TraceEnum_ELBO])
 def test_quantiles(auto_class, Elbo):
     def model():
-        pyro.sample("x", dist.Normal(0.0, 1.0))
         pyro.sample("y", dist.LogNormal(0.0, 1.0))
         pyro.sample("z", dist.Beta(2.0, 2.0).expand([2]).to_event(1))
+        pyro.sample("x", dist.Normal(0.0, 1.0))
```
martinjankowiak, line 532: move "x" so that it's not in the trivial top-left of `scale_tril` position

martinjankowiak: otherwise this test would still pass in master : )

fritzo (commented):
Wouldn't we need a correlated model to test this change?  E.g.
```python
def model():
    w = pyro.sample("w", dist.Normal(0.0, 1.0))
    x = pyro.sample("x", dist.Normal(x, 1.0))
```

tests/infer/test_autoguide.py
```diff
@@ -543,6 +543,13 @@ def model():
     if auto_class is AutoLaplaceApproximation:
         guide = guide.laplace_approximation()
 
+    if hasattr(auto_class, "get_posterior"):
+        posterior = guide.get_posterior()
+        posterior_scale = posterior.variance[-1].sqrt()
+        q = guide.quantiles([0.158655, 0.8413447])
```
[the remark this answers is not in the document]
martinjankowiak: i can replace with this if you prefer
```
guide.quantiles(dist.Normal(0, 1).cdf(torch.tensor([-1.0, 1.0])).data.numpy().tolist())
```

clunky because `quantiles` wraps in a tensor which raises a user warning if the input is a tensor....

```
UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
```

martinjankowiak:
> Wouldn't we need a correlated model to test this change? E.g.
> 
> ```python
> def model():
>     w = pyro.sample("w", dist.Normal(0.0, 1.0))
>     x = pyro.sample("x", dist.Normal(x, 1.0))
> ```

i don't think so. we just need to make sure we're computing the correct diagonal element of the full covariance matrix. so as long as we're doing that in a case where that computation is non-trivial (because e.g. it involves off-diagonal elements of `scale_tril`) we'll be good. as noted in the above comment, it's enough if `x` is not in the top-left of `scale_tril`

martinjankowiak:
@fritzo afaict only `quantiles` was affected and only the `AutoMultivariateNormal` autoguide so afaik this PR is complete, pending further comments

fritzo:
>> Wouldn't we need a correlated model to test this change?

> that computation is non-trivial (because e.g. it involves off-diagonal elements of `scale_tril`)

My concern is that in the existing test, all those off-diagonal elements happen to be zero, so the old and new formulae happen to be identical.  Am I missing something?

martinjankowiak:
the `covariance_matrix` is not diagonal since we do 100 svi steps (except in the Laplace case where I believe the full hessian is not completed). for example:

```
[[ 0.96436582  0.01351931 -0.01359293  0.10933437]
 [ 0.01351931  1.28534454  0.07159596  0.01766826]
 [-0.01359293  0.07159596  1.14718362  0.0288711 ]
 [ 0.10933437  0.01766826  0.0288711   0.96723489]]
```

fritzo (approved):

PR #3243, opened 2023-07-19
Copied comment-docs to torch.distributions wrappers from corresponding sources. 

rtviii:
Was surprised not to find a little handy annotations, had to drill down to source make sure i'm using the parameter correctly.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -38,6 +38,7 @@ sys.path.insert(0, os.path.abspath("../.."))
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
+    'sphinxcontrib.napoleon',
     "sphinx.ext.intersphinx",  #
     "sphinx.ext.todo",  #
     "sphinx.ext.mathjax",  #
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -38,7 +38,6 @@ sys.path.insert(0, os.path.abspath("../.."))
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
-    'sphinxcontrib.napoleon',
     "sphinx.ext.intersphinx",  #
     "sphinx.ext.todo",  #
     "sphinx.ext.mathjax",  #
@@ -48,6 +47,7 @@ extensions = [
     "sphinx.ext.graphviz",  #
     "sphinx.ext.autodoc",
     "sphinx.ext.doctest",
+    'sphinx.ext.napoleon',
 ]
 
 # Disable documentation inheritance so as to avoid inheriting
--- a/pyro/distributions/torch.py
+++ b/pyro/distributions/torch.py
@@ -334,6 +334,13 @@ class Uniform(torch.distributions.Uniform, TorchDistributionMixin):
         return constraints.interval(self._unbroadcasted_low, self._unbroadcasted_high)
 
 
+def doctest_disable(docstring):
+    _ = ""
+    for line in docstring.splitlines():
+        _ += line + "#doctest: +DISABLE"
+    return _
+
+
 # Programmatically load all distributions from PyTorch.
 __all__ = []
 for _name, _Dist in torch.distributions.__dict__.items():
@@ -346,18 +353,25 @@ for _name, _Dist in torch.distributions.__dict__.items():
 
     try:
         _PyroDist = locals()[_name]
+        torchDistDocstring = _Dist.__doc__
+
     except KeyError:
         _PyroDist = type(_name, (_Dist, TorchDistributionMixin), {})
         _PyroDist.__module__ = __name__
         locals()[_name] = _PyroDist
+        torchDistDocstring = None
 
     _PyroDist.__doc__ = """
     Wraps :class:`{}.{}` with
     :class:`~pyro.distributions.torch_distribution.TorchDistributionMixin`.
+
     """.format(
         _Dist.__module__, _Dist.__name__
+    ) + (
+        "\n\n" + doctest_disable(torchDistDocstring)
+        if torchDistDocstring is not None
+        else ""
     )
-
     __all__.append(_name)
 
 


fritzo (changes requested):
Good point that it would be helpful to remove that layer of indirection between Pyro wrappers and their PyTorch distributions.  However I have a couple concerns with the proposed change:
1. We'd prefer to programmatically copy docs directly torch via `@copy_docs_from` rather than manually embed those docs once. The programmatic approach would allow Pyro's wrappers to receive upstream updates from the docs and would reduce our team's maintenance burden.
2. Pyro uses sphinx-style docstrings whereas PyTorch uses NumPy/Google style docstrings. This unfortunate design choice prevents Pyro's doc building system from directly embedding PyTorch docstrings in Pyro wrappers (and I suspect this PR's docs build will fail even after we `make lint` to fix `black` errors).

I think we can accomplish both tasks with little effort by
1. Updating the loop at the end of this file to insert the torch docstrings into our wrappers,
2. Adding [napoleaon](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) to our sphinx dependencies to start supporting NumPy/Google style docstrings as in PyTorch.

@rtviii would you like to try this alternative approach, or would you like me to do it?  Either way, thanks for the nudge to fix our redirected docstrings 🙂

rtviii:
Certainly, i can try. Took a brief look.. 

- soo [that loop at bottom of `torch.py`](https://github.com/pyro-ppl/pyro/blob/891880f/pyro/distributions/torch.py#L339-L361) ought to grab the `_Dist`'s `__doc__` and set the `_PyroDist`'s `__doc__` attr to the same value? Is that how you see it?

- `napoleon` is easy to add to sphinx's `conf.py`.. never used it though, do i have to invoke it explicitly in the make routines?

fritzo:
* in the loop I think I'd keep the existing docstring and append the new docstring to it.  Our current docstring points to the original, and while I agree with you it's helpful to have the original docstring, it's also useful to keep a link to the base distribution.
* I believe we can just add napoleon to our dev requirements in [setup.py](https://github.com/pyro-ppl/pyro/blob/1.8.5/setup.py#L120-L137) and [docs/requirements.txt](https://github.com/pyro-ppl/pyro/blob/1.8.5/docs/requirements.txt). Let's see 🙂 
* Also you can format and lint our python code via `make format` and `make lint` respectively. That should help avoid these CI linting errors

rtviii:
Hey, Fritz. 

A few things.. [This](https://github.com/pyro-ppl/pyro/blob/ae124d51a9e88068b177656f430e33a41c7a39d4/pyro/contrib/forecast/forecaster.py#L73-L74) fails(asking for nonidented lines) seemingly after `sphinx.ext.napoleon` is enabled in `conf.py`. That's the way [napoleon site recommends](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/) it be used by the way, it comes prepackaged with `sphinx` now, they say.

I'm still getting the error (below) thought which seems to have do with the identation characteristic of the numpy/google docstrings i'm pulling from pytorch. Maybe you know what's going on there? 

Furthermore, i had to disable the doctest ( see `doctest_disable` in *1* below) for the build to succeed for now (not sure if that can be a part of the problem). 

<details>
  <summary>1. Updated docbuilding loop in `torch.py`</summary>

```python3 

def doctest_disable(docstring):
    _ = ""
    for line in docstring.splitlines():
        _ += line + "#doctest: +DISABLE"
    return _


# Programmatically load all distributions from PyTorch.
__all__ = []
for _name, _Dist in torch.distributions.__dict__.items():
    if not isinstance(_Dist, type):
        continue
    if not issubclass(_Dist, torch.distributions.Distribution):
        continue
    if _Dist is torch.distributions.Distribution:
        continue

    try:
        _PyroDist = locals()[_name]
        torchDistDocstring = _Dist.__doc__

    except KeyError:
        _PyroDist = type(_name, (_Dist, TorchDistributionMixin), {})
        _PyroDist.__module__ = __name__
        locals()[_name] = _PyroDist
        torchDistDocstring = None

    _PyroDist.__doc__ = """
    Wraps :class:`{}.{}` with
    :class:`~pyro.distributions.torch_distribution.TorchDistributionMixin`.

    """.format(
        _Dist.__module__, _Dist.__name__
    ) + (
        "\n\n" + doctest_disable(torchDistDocstring)
        if torchDistDocstring is not None
        else ""
    )
    __all__.append(_name)


# Create sphinx documentation.
__doc__ = "\n\n".join(
    [
        """
    {0}
    ----------------------------------------------------------------
    .. autoclass:: pyro.distributions.{0}
    """.format(
            _name
        )
        for _name in sorted(__all__)
    ]
)
```

</details>

<details>
  <summary>2. Error from building with pytorch docstrings (`napoleon` enabled)</summary>


```bash
ᢹ saeta.rtviii[ ~/pyro ]  make format                                                                                                                                                                             [dev]
python scripts/update_headers.py
ruff check --fix .
black *.py pyro examples tests scripts profiler
Skipping .ipynb files as Jupyter dependencies are not installed.
You can fix this by running ``pip install "black[jupyter]"``
reformatted /Users/rtviii/pyro/pyro/contrib/forecast/forecaster.py

All done! ✨ 🍰 ✨
1 file reformatted, 616 files left unchanged.


ᢹ saeta.rtviii[ ~/pyro ]  make test                                                                                                                                                                               [dev]
ruff check .
black --check *.py pyro examples tests scripts profiler
Skipping .ipynb files as Jupyter dependencies are not installed.
You can fix this by running ``pip install "black[jupyter]"``
All done! ✨ 🍰 ✨
617 files would be left unchanged.
python scripts/update_headers.py --check
mypy --install-types --non-interactive pyro scripts
Success: no issues found in 319 source files
/Library/Developer/CommandLineTools/usr/bin/make -C docs html
Running Sphinx v6.2.1
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
loading intersphinx inventory from https://pytorch.org/docs/master/objects.inv...
loading intersphinx inventory from http://funsor.pyro.ai/en/stable/objects.inv...
loading intersphinx inventory from https://optimized-einsum.readthedocs.io/en/stable/objects.inv...
loading intersphinx inventory from https://docs.scipy.org/doc/scipy/reference/objects.inv...
loading intersphinx inventory from https://biopython.org/docs/latest/api/objects.inv...
loading intersphinx inventory from https://horovod.readthedocs.io/en/stable/objects.inv...
loading intersphinx inventory from https://graphviz.readthedocs.io/en/stable/objects.inv...
intersphinx inventory has moved: http://funsor.pyro.ai/en/stable/objects.inv -> https://funsor.pyro.ai/en/stable/objects.inv
intersphinx inventory has moved: https://docs.scipy.org/doc/scipy/reference/objects.inv -> https://docs.scipy.org/doc/scipy/objects.inv
building [mo]: targets for 0 po files that are out of date
writing output...
building [html]: targets for 32 source files that are out of date
updating environment: [new config] 32 added, 0 changed, 0 removed
reading sources... [100%] testing

Warning, treated as error:
/Users/rtviii/pyro/pyro/distributions/torch.py:docstring of pyro.distributions.torch.Categorical:6:Inline interpreted text or phrase reference start-string without end-string.
make[1]: *** [html] Error 2
make: *** [docs] Error 2
```

</details>

fritzo:
Hey @rtviii i worked out the errors you discovered and I think things are working in #3246. Feel free to merge those changes into this PR or we can close this PR and merge that, either way.

rtviii:
Closing in favor of #3246

PR #3249, opened 2023-07-25
SWA implementation

yuanqing-wang:


yuanqing-wang:
Addresses https://forum.pyro.ai/t/using-stochastic-weight-averaging-swa-in-pyro/5290/2

fritzo:
Hi @yuanqing-wang, I'd recommend taking a look at our [CONTRIBUTING.md](https://github.com/pyro-ppl/pyro/blob/dev/CONTRIBUTING.md), and in particular:
- `make format` and `make lint` to work past linting errors
- add torchcontrib to the EXTRAS_REQUIRE list in setup.py (and ideally lower bound a version so we know it has SWA)
- add a test, probably in a new file tests/optim/test_swa.py
- probably add a [pytest.importorskip("torchcontrib")](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip) to your test
- add a new automodule section to https://github.com/pyro-ppl/pyro/blob/dev/docs/source/pyro.optim.txt

PR #3253, opened 2023-08-02
Fix GammaPoisson distribution from ignoring validate_args param

austinv11:
Currently the GammaPoisson distribution doesn't propagate the validate_args parameter.


fritzo (approved):
Thanks for fixing this!

Issue #3134, opened 2022-09-06
Render deterministic functions of random variables


Eric-Bradford:
I would like to show variables explicitly using "render_model" that are functions of other variables. I tried using "deterministic" or "Delta distribution" for this purpose, but this seems to not propagate like other distributions, e.g. 

```python
def model(data):
    m = pyro.sample("m", dist.Normal(0, 1))
    sd = pyro.sample("sd", dist.Delta(m**2))
    with pyro.plate("N", len(data)):
        pyro.sample("obs", dist.Normal(m, sd), obs=data)
data = torch.ones(10)
pyro.render_model(model, model_args=(data,))
```

![image](https://user-images.githubusercontent.com/32679792/188665033-092838b2-ddf6-4240-809b-d3d6f6cc2212.png)

If I define sd as a normal distribution it shows the connection between sd and obs correctly, but with "Delta" the connection is not shown. How can I define sd as a function of m that would show this connection?

fritzo:
Hi @Eric-Bradford this looks like a bug to me, let me take a look at what might be causing it...

fritzo:
OK, this is not a bug, merely a historical design choice (pyro.infer.inspect originated as an inference tool that was concerned with only latent variables, hence it ignored deterministic dependencies).  I've started sketching support for this feature [in a branch](https://github.com/pyro-ppl/pyro/tree/render-deterministic) but it's a little more complex than I expected due to low-level provenance tracking issues.  Working on it...

fritzo:
See also [this forum discussion](https://forum.pyro.ai/t/how-to-render-a-semi-supervised-model/4751/8)

r3v1:
Have been any advances on this?

fritzo:
I believe there has been no progress, but @eb8680 was looking into it around 6 months ago.

r3v1:
Is there any plan to implement this in the short time? I've tried patching it somehow but I got lost...

eb8680:
@r3v1 no, sorry, it's unlikely I/we will have the bandwidth to fix this ourselves any time soon. You're welcome to play around with the branch linked above and see if you can get it working yourself.

r3v1:
Thanks anyway. I tried but it was quite difficult for me to understand what is going on behind the scenes, I would take it more deeply if I had some documentation, by the way, do you have any resources about?

PR #3259, opened 2023-08-08
Render deterministic functions

r3v1:
Addresses #3134.


fritzo (commented):
Hi @r3v1, thanks for taking on this task!  Could you describe your approach a bit in the PR description? 🙏 

It's been almost a year since I looked into this, but I recall one issue was that Pyro's visualization reuses inference machinery for tracking dependency, and the goals of these two uses are a bit at odds: in inference one wants to ignore deterministic dependencies and simply track stochastic latent variables because only those stochastic latent variables can be freely varied during inference. By contrast in visualization one often wants to inspect the model, including in many cases deterministic dependencies or even compute graphs.

Now again it's been a while but I think where I [left off on this task](https://github.com/pyro-ppl/pyro/compare/dev...render-deterministic), it seemed challenging to separate these two concerns, like we might need to fork some of the provenance tracking and maybe add some conditional flags so one could optionally trace deterministic nodes (for visualization) or optionally elide them (for inference). I'm not sure how you PR might address these issues, but I think it's worth examining before making changes that affect inference. 🙂

r3v1:
Of course @fritzo, I could explain the process I have gone through.

First, I started by studying the unit tests and seeing what the `get_model_relations` function was expected to return depending on the model, which returned a dictionary representing a graph. Next, I started debugging inside that method the logic you had implemented and started to understand how the different parts "exchange messages" (dictionaries) and I noticed that everything resided inside the `apply_stack` call. 

When calling [`_post_process_message()`](https://github.com/r3v1/pyro/blob/537f7a071ec020855c3f56fe635ec10272261801/pyro/poutine/runtime.py#L221), it checks if the TrackProvenance object has a method `f"_pyro_post_{}".format(msg["type"])`, but as being `deterministic`, it was not implemented yet, so I need to [implement it](https://github.com/r3v1/pyro/blob/537f7a071ec020855c3f56fe635ec10272261801/pyro/infer/inspect.py#L63). To do it, I just remembered watching the same structured code somewhere in [Numpyro's code](https://github.com/pyro-ppl/numpyro/blob/6b2353427c21792b5b4bfcb01192bb408ee9944f/numpyro/infer/inspect.py#L363), so that's what I [applied](https://github.com/r3v1/pyro/blob/537f7a071ec020855c3f56fe635ec10272261801/pyro/infer/inspect.py#L63). In the same way, the [`_pyro_post_deterministic`](https://github.com/r3v1/pyro/blob/537f7a071ec020855c3f56fe635ec10272261801/pyro/poutine/trace_messenger.py#L140) of `TraceMessenger` was implemented.

Of course, the primitive `deterministic` should be modified to match the above. I just took another look to Numpyro's [`deterministic`](https://github.com/pyro-ppl/numpyro/blob/6b2353427c21792b5b4bfcb01192bb408ee9944f/numpyro/primitives.py#L283) to take some ideas. I noticed that their implementation differs, so I gave a try and implemented something similar. I ended up with the idea submitted, emulating the same functionality.

Finally, I apply some corrections to the core of `get_model_relations` to recognize deterministic node's parents and ended applying the same visualization (dashed) to the rendered graph as Numpyro. 

That`s it. Hope it works, or at least, it do the job on my project :nerd_face: 

About the merge intention, I just take the render-deterministic branch to continue the work done. It cannot be merged into the `dev` branch, can it? Or can it?

r3v1:
Hi, any advances on this?

eb8680:
@r3v1 could you revert your changes to the body of `pyro.deterministic` and refactor your new handler methods for `deterministic` to fit within handler methods for `sample`? Changing the `type` of `deterministic` will break some things elsewhere in Pyro and downstream.

You can just use a helper function to check within a `_pyro_sample`/`_pyro_post_sample` if a `sample` site came from `deterministic`:
```python
def site_is_deterministic(msg: dict) -> bool:
    return msg["type"] == "sample" and msg["infer"].get("_deterministic", False)

...

class ...:
    ...
    def _pyro_sample(self, msg):
        if site_is_deterministic(msg):
            # new logic for handling deterministic sites here
            ...
            return

        # old logic for handling sample sites here
        ...
```

eb8680:
@r3v1 I also just changed the base branch to `dev`, can you resolve the merge conflicts and push so that CI can run tests?

PR #3265, opened 2023-08-30
Fix #3255 (draft)

### LINKED ISSUE #3255: Improve user experience in inference with partially observed discrete variables
### Summary
Inference for partially observed discrete variables occasionally produces some counter-intuitive results. Those are not bugs but users may waste a lot of time dealing with them or trying to understand them. The behavior has been tested on Pyro 1.8.5 and 1.8.6.

### A simple example with coins

The example below is meant to show in which kind of context the issues appear. It is artificial and has no practical applications, but it is inspired from real examples I stumbled upon. In the model, we flip a fair coin and do not show the result; if it lands 'heads' we flip a coin with bias 0.05; if it lands 'tails' we flip a coin with bias 0.95. We always observe the result of the biased coin (but not which coin was flipped). In the guide, we simply sample the unbiased coin.

```python
import pyro
import torch

def model(X):
   with pyro.plate("tosses", X.shape[0]):
      c = pyro.sample("unbiased", dist.Categorical(torch.ones(1, 2)))
      pyro.sample("obs", dist.Bernoulli(torch.tensor([.05,.95])[c]), obs=X)

def guide(X):
   post_p = pyro.param("post_p", torch.ones(X.shape[0], 2),
      constraint = torch.distributions.constraints.simplex)
   with pyro.plate("tosses", X.shape[0]):
      pyro.sample("unbiased", dist.Categorical(post_p))

svi = pyro.infer.SVI(model, guide, pyro.optim.Adam({"lr": 0.01}), pyro.infer.Trace_ELBO())
for step in range(2000):
   svi.step(torch.tensor([1.])) # X = 1. stands for "tails".
print(pyro.param("post_p"))

# tensor([[0.0552, 0.9448]], grad_fn=<DivBackward0>)
```
The result is correct, the second coin landed 'tails' so the posterior probability that the unbiased coin landed 'tails' is 0.95.

### Issue 1: Code failure when masking

If the second coin is sometimes observed, we can introduce an observation mask for the `obs` sample. Let us modify the code and run the same example, _i.e._, we specify that the second coin landed 'tails' and this is observed.

```python
import pyro
import torch

# Add `mask` argument, together with `obs_mask=mask` in sample "obs".
def model(X, mask):
   with pyro.plate("tosses", X.shape[0]):
      c = pyro.sample("unbiased", dist.Categorical(torch.ones(1, 2)))
      pyro.sample("obs", dist.Bernoulli(torch.tensor([.05,.95])[c]), obs=X, obs_mask=mask)

def guide(X, mask):
   post_p = pyro.param("post_p", torch.ones(X.shape[0], 2),
      constraint = torch.distributions.constraints.simplex)
   with pyro.plate("tosses", X.shape[0]):
      pyro.sample("unbiased", dist.Categorical(post_p))

svi = pyro.infer.SVI(model, guide, pyro.optim.Adam({"lr": 0.01}), pyro.infer.Trace_ELBO())
for step in range(2000):
   svi.step(torch.tensor([1.]), torch.tensor([True])) # True means observed.
print(pyro.param("post_p"))
```

The code fails with the error below.
```
(...)/pyro/util.py:303: UserWarning: Found vars in model but not guide: {'obs_unobserved'}
  warnings.warn(f"Found vars in model but not guide: {bad_sites}")
  File "(...)", line 18, in <module>
    svi.step(torch.tensor([1.]), torch.tensor([True]))
  File "(...)/pyro/infer/svi.py", line 145, in step
  File "(...)/pyro/infer/trace_elbo.py", line 141, in loss_and_grads
    loss_particle, surrogate_loss_particle = self._differentiable_loss_particle(
  File "(...)/pyro/infer/trace_elbo.py", line 106, in _differentiable_loss_particle
    log_r = _compute_log_r(model_trace, guide_trace)
  File "(...)/pyro/infer/trace_elbo.py", line 27, in _compute_log_r
    log_r_term = log_r_term - guide_trace.nodes[name]["log_prob"]
KeyError: 'obs_unobserved'
```

First there is a warning for missing site in the guide and then a `KeyError` for the same reason. This is counter-intuitive: either guides with missing sites should be allowed (warn only), or they should not (raise an error for every guide with missing sites).

The error comes from a part of the code that evaluates the loss using the REINFORCE estimator (_i.e._, when the reparametrization trick cannot be used, as in the case of discrete random variables). Line 27 in `trace_elbo.py` assumes that every unobserved site in the model also exists in the guide. The user may not be aware that the site `obs_unobserved` is created in the model (but not in the guide) as soon as the argument `obs_mask` is not `None`.

The solution is to define the sample `obs_unobserved` in the guide (see how below), but there are barely any mentions of this, so we cannot assume that users will do it. If guides with missing sites are allowed, line 27 in `trace_elbo.py` should be replaced with a fail-safe version. Ideally, a message could point users in the right direction if Pyro creates an `_unobserved` site that is not in the guide.

### Issue 2: Counter-intuitive gradient

Now if the unbiased coin is sometimes observed, we can introduce an observation mask for the `unbiased` sample, together with some observations when they are available. As mentioned above, we need to add a site in the guide called `unbiased_unobserved` explaining what to do when the coin is not observed (_i.e._, sample it as we were doing until now). We have to sample the whole tensor; Pyro will automatically mix in observed and sampled values for us as needed.

Some values in `unbiased_unobserved` are sampled for nothing: they will be replaced with the observed values if they are available. In this case, the sampled have no effect on the inference, but just to be sure, we are going to mask them in the guide to set their `log_prob` terms to 0. We do this by using `poutine.mask` where we invert the observation mask with `~mask`.

```python
import pyro
import torch

# Add `coin_obs` argument; now `obs_mask` applies to the unbiased coin.
def model(X, coin_obs, mask):
   with pyro.plate("tosses", X.shape[0]):
      c = pyro.sample("unbiased", dist.Categorical(torch.ones(1, 2)), obs=coin_obs, obs_mask=mask)
      pyro.sample("obs", dist.Bernoulli(torch.tensor([.05,.95])[c]), obs=X)

def guide(X, coin_obs, mask):
   post_p = pyro.param("post_p", torch.ones(X.shape[0], 2), 
      constraint = torch.distributions.constraints.simplex)
   with pyro.plate("tosses", X.shape[0]):
      with pyro.poutine.mask(mask=~mask): # Invert the observation mask.
         pyro.sample("unbiased_unobserved", dist.Categorical(post_p))

svi = pyro.infer.SVI(model, guide, pyro.optim.Adam({"lr": 0.01}), pyro.infer.Trace_ELBO())
for step in range(2000):
   # Second coin: tails, heads, tails. Unbiased coin: tails, heads, ?
   svi.step(torch.tensor([1., 0., 1.]), torch.tensor([1, 0, 0]), torch.tensor([True, True, False]))
print(pyro.param("post_p"))

# tensor([[0.3058, 0.6942],
#         [0.7498, 0.2502],
#         [0.0502, 0.9498]], grad_fn=<DivBackward0>)
```

In this example, we observed the first two flips of the unbiased coin, but not the third. We set the value to `heads` with `0` but this is irrelevant because the value is never used throughout the inference. The inference is correct for the third flip and there is nothing to infer for the first two flips because the values were observed... So why did the values of `post_p` change from the initial `0.5` and what do the current values represent?

As far as I understand, the values have no special meaning. There is nothing to infer anyway. So why did they change? Once again, this has to do with the way Pyro evaluates the loss using the REINFORCE estimator. Internally, it keeps track of a `log_prob` term and a `score_function` term for the sites of the guide. The `log_prob` terms are masked but not the `score_function` terms, so all the values of `unbiased_unobserved` contribute to the gradient, even those that are overwritten by observed values.

I don't think that this has side effects, so this is not really a bug. The issue here is that Pyro is difficult enough to debug, and erratic behaviors make it harder. It would help if parameters that have no effect on the inference have gradient 0, so that the user gets alerted when there is an error in the model (_e.g._, when values that should have no effect on the inference do in fact have an effect).

gui11aume:


gui11aume:
The purpose of this pull request is to harmonize the behavior of masking between the different ways of estimating the gradient of the ELBO (most notably when `has_r_sample` is `False`). See the discussion on issue #3255.

I added some tests inspired from those that were already present. Just ignoring the fact that some sites of the model are not present in the guide gave the correct results right away. The main difficulties were:
1. Having missing sites in the guide triggered a warning that made the tests fail.
2. The memory blew up with 50,000 samples in one particular case (going above 256 GB).

I had to disable user warnings for `test_mask_gradient` and break down sampling in 10 series of 5000 each.

The changes are otherwise as discreet as possible.

ordabayevy:
@gui11aume  can you please fix lint issues from `make lint` output and push changes? You can first run `make format` to automatically fix formatting and then manually fixing any leftover linting issues.

gui11aume:
Hi @ordabayevy! Thanks for taking the time to help me with this. There were issues in the file `tests/infer/test_gradient.py` (mostly in the code for the new tests) and running `make format` seemed to fix them all. Let me know if some issues remain.

pyro/infer/trace_elbo.py
```diff
@@ -23,7 +23,7 @@ def _compute_log_r(model_trace, guide_trace):
     for name, model_site in model_trace.nodes.items():
         if model_site["type"] == "sample":
             log_r_term = model_site["log_prob"]
-            if not model_site["is_observed"]:
+            if not model_site["is_observed"] and name in guide_trace.nodes:
```
fritzo, line 26: I may be forgetting something, but I thought `Trace_ELBO` requires the guide to include all model sites that are not observed. If that's the case, we wouldn't want to keep the old version where `Trace_ELBO` errors. Can you explain when a model site should not be in the guide?

pyro/distributions/score_parts.py
```diff
@@ -25,6 +25,6 @@ def scale_and_mask(self, scale=1.0, mask=None):
         :type mask: torch.BoolTensor or None
         """
         log_prob = scale_and_mask(self.log_prob, scale, mask)
-        score_function = self.score_function  # not scaled
+        score_function = scale_and_mask(self.score_function, 1.0, mask)  # not scaled
```
fritzo, line 28: Can you point out exactly the scenario that is fixed by this one line change?  IIRC, `score_function` would always be multiplied by another tensor that is masked, so the mask here would be redundant.

pyro/infer/tracegraph_elbo.py
```diff
@@ -218,6 +218,8 @@ def _compute_elbo(model_trace, guide_trace):
     # we include only downstream costs to reduce variance
     # optionally include baselines to further reduce variance
     for node, downstream_cost in downstream_costs.items():
+        if node not in guide_trace.nodes:
+            continue
```
fritzo, line 222: ditto: this should never happen

pyro/infer/trace_mean_field_elbo.py
```diff
@@ -108,7 +108,7 @@ def _differentiable_loss_particle(self, model_trace, guide_trace):
             if model_site["type"] == "sample":
                 if model_site["is_observed"]:
                     elbo_particle = elbo_particle + model_site["log_prob_sum"]
-                else:
+                elif name in guide_trace.nodes:
```
fritzo, line 111: ditto: this should never happen

tests/infer/test_gradient.py
```diff
@@ -214,6 +215,102 @@ def guide(subsample):
     assert_equal(actual_grads, expected_grads, prec=precision)
 
 
+# Not including the unobserved site in the guide triggers a warning
+# that can make the test fail if we do not deactivate UserWarning.
+@pytest.mark.filterwarnings("ignore::UserWarning")
```
fritzo, line 220: ditto: this should never happen

pyro/infer/trace_elbo.py
gui11aume: I agree with you and I cannot think of useful cases where this is the case. The point I make in the [issue](https://github.com/pyro-ppl/pyro/issues/3255) is that this triggers a warning when `has_rsample` is true and an error when it is false. I think they should both trigger a warning or both trigger an error. The suggested changes try to make the behavior consistent with the case `has_rsample = True`. But maybe it makes more sense to trigger an error everywhere?

pyro/distributions/score_parts.py
gui11aume: Here it's again for consistency. When a variable is partially observed, some parameters are created for the missing observations, and some dummy parameters are created for the non-missing ones. When `has_rsample` is true, the dummy parameters are not updated: they retain their initial values because they do not contribute to the gradient. When `has_rsample` is false, the gradient "leaks" through this line and the dummy parameters are updated during learning. (but I found that inference on the non-dummy parameters was correct in the cases I checked). As above, this line does not really fix any bug, it just tries to make the behavior consistent between `has_rsample = True` and `has_rsample = False`.

pyro/infer/trace_elbo.py
fritzo: Ah, thanks for explaining, I think I better understand now.

I'd like to hear what other folks think (@martinjankowiak @eb8680 @fehiepsi).  One argument for erroring more often is that there is a lot code in Pyro that tacitly assumes all sites are either observed or guided.  I'm not sure what that code is, since we've only tacitly made that assumption, but it's worth thinking about: reparametrizers, `Predictive`, `AutoGuideMessenger`.

One argument for allowing "partial" guides is that it's just more general.  But if we decide to support "partial" guides throughout Pyro, I think we'll need to adopt importance sampling semantics, so we'd need to replace pyro's basic `Trace` data structure with a weighted trace, and replace sample sets with weighted sets of samples in a number of places e.g. `Predictive`.  This seems like a perfectly fine design choice for a PPL, but it is different from much of Pyro's existing usage, and I think we would need to make many small changes throughout the codebase including tutorials. 🤔

pyro/distributions/score_parts.py
fritzo: Thanks for explaining, I think I better understand now.

PR #3266, opened 2023-09-14
Render deterministic functions

r3v1:
Addresses #3134 and pull request #3259.


pyro/poutine/trace_messenger.py
```diff
@@ -137,6 +137,9 @@ def _pyro_post_sample(self, msg):
     def _pyro_post_param(self, msg):
         self.trace.add_node(msg["name"], **msg.copy())
 
+    def _pyro_post_deterministic(self, msg):
```
eb8680, line 140: This method is left over from the previous version #3259 and can be removed

pyro/infer/inspect.py
```diff
@@ -510,11 +528,23 @@ def render_graph(
...
                 shape = "ellipse"
+                rv_label = rv
 
             # For param_nodes - No shape
             else:
                 shape = "plain"
-            cur_graph.node(rv, label=rv, shape=shape, style="filled", fillcolor=color)
+                rv_label = rv.replace(
+                    "$params", ""
+                )  # incase of neural network parameters
```
eb8680, line 538: Can you revert this change, since it's independent of the main goal of this PR? You're welcome to open a separate issue or PR with suggested changes to rendering settings like this one.

pyro/infer/inspect.py
```diff
@@ -43,8 +43,20 @@ def is_sample_site(msg):
...
+
 class TrackProvenance(Messenger):
     def _pyro_post_sample(self, msg):
+        if site_is_deterministic(msg):
+            provenance = frozenset({msg["name"]})  # track only direct dependencies
+            value = detach_provenance(msg["value"])
+            msg["args"] = (*msg["args"], msg["value"],)
+            msg["kwargs"] = {}
+            msg["value"] = ProvenanceTensor(value, provenance)
+            # msg["fn"] = lambda x: x
```
eb8680, line 58: nit: remove commented line

pyro/infer/inspect.py
```diff
@@ -301,10 +311,16 @@ def _get_type_from_frozenname(frozen_name):
         if site["type"] != "sample" or site_is_subsample(site):
             continue
 
+        provenance = get_provenance(
+            site["fn"].log_prob(site["value"])
+            if not site_is_deterministic(site)
+            else site["args"][-1]
+        )
```
eb8680, line 318: Using the site value like this will not be correct when `site["fn"].event_dim > 0`.

At `deterministic` sites, the distribution is `Delta(v, event_dim).mask(False)`. The `.mask` messes up the provenance of `log_prob`, but if you use `site["fn"].base_dist` to access the underling `Delta` using `log_prob` should work as expected:
```python
provenance = get_provenance(
    site["fn"].log_prob(msg["value"])
    if not site_is_deterministic(site)
    else site["fn"].base_dist.log_prob(msg["value"])
)
```

pyro/infer/inspect.py
```diff
@@ -43,8 +43,20 @@ def is_sample_site(msg):
     return True
 
 
+def site_is_deterministic(msg: dict) -> bool:
+    return msg["type"] == "sample" and msg["infer"].get("_deterministic", False)
+
+
 class TrackProvenance(Messenger):
     def _pyro_post_sample(self, msg):
+        if site_is_deterministic(msg):
```
eb8680, line 52: Can you add a boolean flag `self.include_deterministic` to `TrackProvenance` as in #3259 and check it here:
```python
if self.include_deterministic and site_is_deterministic(msg):
```

eb8680 (changes requested):
@r3v1 thanks for doing this! I left some comments requesting changes below. Can you also add the unit tests from #3259?

eb8680:
@r3v1 can you also run `make format` and `make lint` to appease the linter?

pyro/infer/inspect.py
```diff
@@ -43,8 +43,20 @@ def is_sample_site(msg):
...
+            provenance = frozenset({msg["name"]})  # track only direct dependencies
+            value = detach_provenance(msg["value"])
+            msg["args"] = (*msg["args"], msg["value"],)
+            msg["kwargs"] = {}
```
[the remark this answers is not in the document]
r3v1: In fact, I use `args`  to track provenance in line 314, may be not the best way but it's working

pyro/infer/inspect.py
r3v1: Yes, but as I define `rv_label` in the `if` clause, it should also be defined in the `else` clause. Otherwise, it will be undefined

pyro/infer/inspect.py
r3v1: Understood

pyro/infer/inspect.py
r3v1: Should it be also extended to TrackProvenance calling (and in its uses in `get_dependencies`, `get_model_relations`, etc)?

eb8680: Yes, as discussed in #3259 we want to be able to switch between the current behavior of `TrackProvenance` that only tracks stochastic dependencies for use in inference and the new behavior in this PR that tracks deterministic and stochastic dependencies for use in visualization.

pyro/infer/inspect.py
eb8680: You can just revert to using `rv`, once you revert this chance there's no need for `rv_label`.

eb8680: @r3v1 looks like this change did not get reverted

pyro/infer/inspect.py
```diff
@@ -560,6 +597,7 @@ def render_model(
     filename: Optional[str] = None,
     render_distributions: bool = False,
     render_params: bool = False,
+    include_deterministic: bool = False,
```
eb8680, line 600: nit: rename this keyword argument to `render_deterministic` to match the others

eb8680 (changes requested):
@r3v1 can you add a unit test like `test_get_model_relations` from #3259 that exercises your new functionality, and can you also add a quick example at the end of the [model rendering tutorial](https://github.com/pyro-ppl/pyro/blob/dev/tutorial/source/model_rendering.ipynb) to demonstrate how to use it?

For the example, you can just copy the model that's already defined in the "Rendering the parameters" section and add a couple of new deterministic variables, e.g.:
```python
def model_deterministic(data):
    sigma = pyro.param("sigma", torch.tensor([1.]), constraint=constraints.positive)
    mu = pyro.param("mu", torch.tensor([0.]))
    x = pyro.sample("x", dist.Normal(mu, sigma))
    log_y = pyro.sample("y", dist.Normal(x, 1))
    y = pyro.deterministic("y", log_y.exp())
    with pyro.plate("N", len(data)):
        eps_z_loc = pyro.sample("eps_z_loc", dist.Normal(0, 1))
        z_loc = pyro.deterministic("z_loc", eps_z_loc + x, event_dim=0)
        pyro.sample("z", dist.Normal(z_loc, y), obs=data)

data = torch.ones(10)
pyro.render_model(
    model_deterministic,
    model_args=(data,),
    render_params=True,
    render_distributions=True,
    render_deterministic=True
)
```

eb8680 (commented):
> Do I push the change?

Sorry, I think there was a problem with the diff I was looking at on GitHub. I was specifically requesting the removal of the `.replace("$params", "")` method call, which I now see you actually did already. The most recent commit as of this review produces the second of your two figures, right?

Other than a couple minor comments I left below, I think this is good to go as a first PR assuming tests pass since it retains the current `dev` behavior of `get_dependencies` and `render_model` as the default. We can change that behavior in a followup PR.

pyro/test.py
eb8680, line 1: Also remove this file

tutorial/source/model_rendering.ipynb
eb8680, line 1: You can also remove one of the two new examples from this notebook, just one is enough.

pyro/infer/inspect.py
```diff
@@ -23,29 +23,42 @@
...
+        while hasattr(fn, "base_dist"):
+            fn = fn.base_dist
+        if type(fn).__name__ == "Delta":
+            return False
 
     return True
 
 
+def site_is_deterministic(msg: dict) -> bool:
+    return msg["type"] == "sample" and msg["infer"].get("_deterministic", False)
```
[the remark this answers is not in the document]
r3v1: I test this and introduces some errors in the flow, I would mantain the requested one

pyro/test.py
r3v1: My bad

r3v1:
> > Do I push the change?
> 
> Sorry, I think there was a problem with the diff I was looking at on GitHub. I was specifically requesting the removal of the `.replace("$params", "")` method call, which I now see you actually did already. The most recent commit as of this review produces the second of your two figures, right?


The most recent commit [6fbb58c](https://github.com/pyro-ppl/pyro/commit/6fbb58c5160828595b502c8b3c178ac42707e2b1) produces the first picture, so I'm pushing the fix along the other suggestions in the next commit.

pyro/infer/inspect.py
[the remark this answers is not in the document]
eb8680: OK, thanks for trying, I'll see if I can get it working myself in a followup PR.

eb8680 (approved):
LGTM, outstanding issues can be addressed in followup PRs. Thanks again for taking this on!

r3v1:
Hi @eb8680, do you plan to make a new release with this change?

eb8680:
@r3v1 yes, we'll do a minor release sometime soon. In the meantime you can follow the instructions from the README to install the `dev` branch, where this PR landed: https://github.com/pyro-ppl/pyro#installing-pyro-dev-branch

Issue #3234, opened 2023-06-23
consider fixing possible setuptools vulnerability


martinjankowiak:
https://github.com/pyro-ppl/pyro/security/dependabot/1

> To resolve the issue add a supported lockfile (Pipfile.lock, pyproject.lock or poetry.lock).

> fritz: oh man we're using an old setuptools because lap requires it. I guess we can drop lap support; I doubt anyone but me ever used it, but I'd still prefer to follow a responsible deprecation process.

fritzo:
We may be able to replace lap with scipy's [maximum_bipartite_matching](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csgraph.maximum_bipartite_matching.html)

PR #3267, opened 2023-09-15
Replace lap.lapjv() with scipy.optimize.linear_sum_assignment()

fritzo:
Addresses #3234 

This replaces the unmaintained and difficult-to-install [lap](https://pypi.org/project/lap/) dependency with [scipy.optimize.linear_sum_assignment](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linear_sum_assignment.html#scipy.optimize.linear_sum_assignment).  This is important because `lap` requires an old version of `setuptools`, and @dependabot has recommended pinning to a newer version of `setuptools` that fixes a vulnerability.  This PR does not attempt to pin setuptools yet.

This PR also fixes a bug in `assert_equal()` for integer tensors whereby some assertions were not being checked.

## Tested
- [x] removed `pytest.skip` from affected tests
- [x] fixed a bug in `assert_equal()` for integer tensors

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -7,4 +7,4 @@ opt_einsum>=2.3.2
 pyro-api>=0.1.1
 tqdm>=4.36
 funsor[torch]
-setuptools<60
+setuptools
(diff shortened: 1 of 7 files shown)


eb8680 (approved):
LGTM, thanks for fixing this!

Issue #3268, opened 2023-09-20
Wrong codomain in LeakyReLUTransform


fariedabuzaid:
### Issue Description
The codomain of a LeakyReLUTransform is set to **positive** but should be **real**.

### Environment
 - Pyro version: 1.8.5

### Code Snippet
From [pyro/distributions/transforms/basic.py](https://github.com/pyro-ppl/pyro/blob/dev/pyro/distributions/transforms/basic.py)
```python
class LeakyReLUTransform(Transform):
    """
    Bijective transform via the mapping :math:`y = \text{LeakyReLU}(x)`.
    """
    domain = constraints.real
    codomain = constraints.positive

```

### Suggestion 
```python
class LeakyReLUTransform(Transform):
    """
    Bijective transform via the mapping :math:`y = \text{LeakyReLU}(x)`.
    """
    domain = constraints.real
    codomain = constraints.real
```

PR #3270, opened 2023-09-20
Fix: Wrong codomain in LeakyReLUTransform

MeenuyD:
The codomain constraint of the LeakyReLUTransform has been updated from 'constraints. positive' to 'constraints. real' to accurately reflect the Leaky ReLU activation function's behavior. This will allow the transform to produce both positive and negative values when applied to different inputs.
#3268


fritzo (approved):
Thanks for the fix!

Issue #3276, opened 2023-10-05
[bug] mypy global python version too old


cyianor:
### Problem description

I forked pyro from the dev branch at 01ccf3647c1c1031b8a188488f1330b16fb5fd31 and installed all dependencies with `make install` as in `CONTRIBUTING.md`. I also needed to manually install ruff and mypy to be able run `make format` and `make lint`.

During `make lint` I encountered the following error
```bash
mypy --install-types --non-interactive pyro scripts
/Users/xxx/.pyenv/versions/3.11.5/lib/python3.11/site-packages/numpy/_typing/_nested_sequence.py:60: error: Positional-only parameters are only supported in Python 3.8 and greater  [syntax]
Found 1 error in 1 file (errors prevented further checking)
make: *** [lint] Error 2
```
I saw that in [`setup.cfg`](https://github.com/pyro-ppl/pyro/blob/c00bcc3fb701327b07e94de88754da49b7f29ebe/setup.cfg#L38) it says that `python_version = 3.7` which seems to be the problem. As soon as I change this to `python_version = 3.8` all works as expected.

I assume that pyro is still meant to be usable with Python 3.7 and that is why this line is there? However, linting will then always fail when performed with newer versions of Python.

### Setup

macOS 13.6
Python 3.11.5
mypy 1.5.1
numpy 1.26.0

PR #3277, opened 2023-10-05
Address the issues in #3274

### LINKED ISSUE #3274: [bug] MixtureOfDiagNormals
### Problem description

I recently tried to use [`MixtureOfDiagNormals`](https://github.com/pyro-ppl/pyro/blob/c00bcc3fb701327b07e94de88754da49b7f29ebe/pyro/distributions/diag_normal_mixture.py#L15) and ran across two issues while using it together with an `AutoNormal` guide.

1. The distributions [`MixtureOfDiagNormal`](https://github.com/pyro-ppl/pyro/blob/c00bcc3fb701327b07e94de88754da49b7f29ebe/pyro/distributions/diag_normal_mixture.py#L15) and [`MixtureOfDiagNormalsSharedCovariance`](https://github.com/pyro-ppl/pyro/blob/c00bcc3fb701327b07e94de88754da49b7f29ebe/pyro/distributions/diag_normal_mixture_shared_cov.py#L15) do not implement the `support` property, which leads to problems when using the distribution for latent variables and e.g. the `AutoNormal` guide. This can be solved easily by adding `support = constraints.real_vector` at the beginning of both classes. I assume that `real_vector` is the correct constraint here since `D=1` is not supported anyways.
2. There seems to be a bug in the [`forward`](https://github.com/pyro-ppl/pyro/blob/c00bcc3fb701327b07e94de88754da49b7f29ebe/pyro/distributions/diag_normal_mixture.py#L134) method of `_MixDiagNormalSample`. Specifically, when creating the white noise
   ```python
       white = locs.new(noise_shape).normal_()
   ```
   This creates a new tensor of size `(len(noise_shape),)` instead of expected size `noise_shape`. This is actually done correctly in [`_MixDiagNormalSharedCovarianceSample`](https://github.com/pyro-ppl/pyro/blob/c00bcc3fb701327b07e94de88754da49b7f29ebe/pyro/distributions/diag_normal_mixture_shared_cov.py#L137C30-L137C30) with a call to `torch.randn` instead.
   The correct way to do it would be
   ```python
       white = locs.new(*noise_shape).normal_()
   ```
   However, since `torch.Tensor.new` does not appear in the PyTorch documentation (at least I didn't find it), I assume it is recommended to use a method like `torch.Tensor.new_empty` instead.
   ```python
       white = locs.new_empty(noise_shape).normal_()
   ```
   works as expected. Alternatively, the same approach as for `MixtureOfDiagNormalSharedCovariance` is possible.

### Package versions

PyTorch: version 2.0.1
Pyro: version 1.8.6

cyianor:
Fixes #3274 

* Added `support = constraints.real_vector` to `MixtureOfDiagNormals` and `MixtureOfDiagNormalsSharedCovariance`
* Fixed the white noise sampling bug in the `forward` method of `_MixDiagNormalSample`
* Same call in `_MixDiagNormalSample` and `_MixDiagNormalSharedCovarianceSample` to generate white noise
* Harmonized tensor shape error messages between `MixtureOfDiagNormals` and `MixtureOfDiagNormalsSharedCovariance`
* Added the correct class name in tensor shape errors for `MixtureOfDiagNormalsSharedCovariance`

`make format` went through without problems. `make lint` runs with the change noted in issue #3276 but fails with some type errors in pyro/optim.  This is not related to any changes I made here. The rest of the linting goes through without a problem.

### DIFF
--- a/pyro/distributions/diag_normal_mixture.py
+++ b/pyro/distributions/diag_normal_mixture.py
@@ -46,13 +46,15 @@ class MixtureOfDiagNormals(TorchDistribution):
         "coord_scale": constraints.positive,
         "component_logits": constraints.real,
     }
+    support = constraints.real_vector
 
     def __init__(self, locs, coord_scale, component_logits):
         self.batch_mode = locs.dim() > 2
         assert coord_scale.shape == locs.shape
-        assert (
-            self.batch_mode or locs.dim() == 2
-        ), "The locs parameter in MixtureOfDiagNormals should be K x D dimensional (or B x K x D if doing batches)"
+        assert self.batch_mode or locs.dim() == 2, (
+            "The locs parameter in MixtureOfDiagNormals should be K x D dimensional "
+            "(or ... x B x K x D if doing batches)"
+        )
         if not self.batch_mode:
             assert (
                 coord_scale.dim() == 2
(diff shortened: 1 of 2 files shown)


fritzo (approved):
Thanks for fixing this!

Issue #3269, opened 2023-09-20
Drop Python 3.7 support


ordabayevy:
Version 3.7 is no longer supported by Python and it probably makes sense to drop the support for it in Pyro as well:

https://devguide.python.org/versions/

ordabayevy:
According to scientific python (https://scientific-python.org/specs/spec-0000/#support-window) support for Python 3.8 can also be dropped already.

PR #3278, opened 2023-10-06
Drop Python 3.7 support

ordabayevy:
Addresses #3276 and #3269

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -137,8 +137,6 @@ setup(
         "horovod": ["horovod[pytorch]>=0.19"],
         "lightning": ["lightning"],
         "funsor": [
-            # This must be a released version when Pyro is released.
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -152,7 +152,6 @@
         "License :: OSI Approved :: Apache Software License",
         "Operating System :: POSIX :: Linux",
         "Operating System :: MacOS :: MacOS X",
-        "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
```
fritzo, line 156: Feel free to add 3.10, 3.11, 3.12 or whatever are supported by PyTorch

fritzo (dismissed):
Thanks for fixing this!

setup.py
ordabayevy: It's not clear if Python 3.12 is supported by PyTorch so skipped it for now.

fritzo (approved):

PR #3283, opened 2023-10-19
Fix `pyro.poutine.handlers`

ordabayevy:
Two fixes:

- `PlateMessenger` is not used in this file and not imported from it
- `MarkovMessenger` from the `_msngrs` list is used to add `markov` variable to the `locals()` but then `markov` is redefined. So I believe `MarkovMessenger` can be removed from `_msngrs`


fritzo (approved):
Seems reasonable, I doubt anyone is importing `PlateMessenger` from `pyro.poutine.handlers`.

Issue #3284, opened 2023-10-19
Mypy has issues with adding classes to the `locals()` in `pyro.poutine.handlers` module


ordabayevy:
In the `pyro.poutine.handlers` submodule there is a `_make_handler` function that converts the classes from the `_msngrs` list to handler functions and adds them to the `locals()` (e.g., `BlockMessenger` -> `block`). Then these are imported in the `pyro.poutine.__init__` (i.e., `from .handlers import block`). Mypy doesn't like this. One solution could be to rename these messengers like `BlockMessenger` -> `block` and make them callable like in NumPyro.

PR #3285, opened 2023-10-22
Add handlers explicitly

ordabayevy:
Addresses #3284 

This should also improve the documentation by displaying handler args in https://docs.pyro.ai/en/stable/poutine.html


fritzo (approved):
Looks like a great improvement to usability!

PR #3288, opened 2023-10-30
Type annotate `pyro.poutine.runtime`

ordabayevy:


pyro/poutine/guide.py
```diff
@@ -112,7 +112,7 @@ def get_posterior(
         """
         raise NotImplementedError
 
-    def upstream_value(self, name: str) -> torch.Tensor:
+    def upstream_value(self, name: str):
```
ordabayevy, line 115: Temporarily removed the return type (changed to Any) so that mypy wouldn't complain.

pyro/poutine/runtime.py
```diff
@@ -1,21 +1,48 @@
...
     ParamStoreDict,
 )
 
+if TYPE_CHECKING:
+    from pyro.poutine.indep_messenger import CondIndepStackFrame
+    from pyro.poutine.messenger import Messenger
+
 # the global pyro stack
-_PYRO_STACK = []
+_PYRO_STACK: List[Messenger] = []
```
fritzo, line 22: This is great!

pyro/poutine/runtime.py
```diff
@@ -24,26 +51,25 @@ class _DimAllocator:
...
-    def allocate(self, name, dim):
+    def allocate(self, name: str, dim: Optional[int]) -> int:
         """
         Allocate a dimension to an :class:`plate` with given name.
         Dim should be either None for automatic allocation or a negative
         integer for manual allocation.
         """
         if name in self._stack:
-            raise ValueError('duplicate plate "{}"'.format(name))
+            raise ValueError(f"duplicate plate '{name}'")
```
fritzo, line 65: No need to change, but I've been using the notation `f"duplicate plate {name!r}"` which I find easier to read.

setup.cfg
```diff
@@ -77,11 +78,9 @@ warn_unused_ignores = True
 ignore_errors = True
 warn_unused_ignores = True
 
-[mypy-pyro.optm.*]
-warn_unused_ignores = True
-
-[mypy-pyro.poutine.*]
+[mypy-pyro.optim.*]
 ignore_errors = True
+warn_unused_ignores = True
```
fritzo, line 83: Can we add `warn_unused_ignores` to the top level `[mypy]` config?  I've found that some `# type: ignore`s mask real errors on the same line, so I've started using `warn_unused_ignores` globally, but maybe this practice is too pedantic for a user base with diverse mypy versions.

fritzo (approved):
LGTM, I'll try to keep review latency low...

PR #3290, opened 2023-11-04
Type annotate `pyro.poutine.messenger`

ordabayevy:


pyro/poutine/messenger.py
```diff
@@ -61,17 +74,15 @@ class Messenger:
     Most inference operations are implemented in subclasses of this.
     """
 
-    def __call__(self, fn):
+    def __call__(self, fn: Callable) -> Callable:
```
fritzo, line 77: Should we be more precise and allow passing of type information?
```python
_F = TypeVar("_F", bound=Callable)

class Messenger:
    def __call__(self, fn: _F) -> _F:
        ...
```

pyro/poutine/messenger.py
```diff
@@ -227,7 +250,9 @@ def unregister(cls, fn=None, type=None):
 
 
 @contextmanager
-def block_messengers(predicate):
+def block_messengers(
+    predicate: Callable[[Messenger], bool]
```
fritzo, line 254: This is great!

fritzo (approved):

PR #3292, opened 2023-11-12
Type annotate `pyro.primitives` & `poutine.block_messenger`

ordabayevy:


pyro/primitives.py
```diff
@@ -6,18 +6,22 @@
 from collections import OrderedDict
 from contextlib import ExitStack, contextmanager
 from inspect import isclass
+from typing import Callable, Dict, Iterator, Optional, Sequence, Union
 
 import torch
+from torch.distributions import constraints
```
fritzo, line 12: I see lots of new dependencies here.  Historically we've had a number of cyclic dependency issues in Pyro.  One thing we might consider to try to avoid cyclid dependencies is to guard these with an `if TYPE_CHECKING`:
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from torch.distributions import contstraints
    from pyro.distributions import TorchDistribution
    from pyro.params.param_store import ParamStoreDict
    from pyro.poutine.runtime import Message
```
Or maybe we can just use that trick next time we need to fix a cyclic dependency. Either way.

pyro/distributions/torch_distribution.py
```diff
@@ -51,7 +51,7 @@ def __call__(self, sample_shape=torch.Size()):
         )
 
     @property
-    def event_dim(self):
+    def event_dim(self) -> int:
```
ordabayevy, line 54: In this file I type annotated only methods needed for `pyro.primitives`

pyro/poutine/runtime.py
```diff
@@ -252,15 +266,31 @@ def apply_stack(initial_msg: Message) -> None:
...
+@overload
+def effectful(
+    fn: Callable[P, T] = ..., type: Optional[str] = ...
+) -> Callable[..., Union[T, torch.Tensor, None]]:
+    ...
+
+
+def effectful(
+    fn: Optional[Callable[P, T]] = None, type: Optional[str] = None
+) -> Callable:
```
ordabayevy, line 293: This is the best I could do at being specific with callable types.

pyro/primitives.py
```diff
@@ -102,7 +122,15 @@ def _masked_observe(name, fn, obs, obs_mask, *args, **kwargs):
     return deterministic(name, value)
 
 
-def sample(name, fn, *args, **kwargs):
+def sample(
+    name: str,
+    fn: TorchDistribution,
```
ordabayevy, line 125: Is this correct or does `fn` can be any callable returning `torch.Tensor`?

pyro/primitives.py
```diff
@@ -102,7 +122,15 @@ def _masked_observe(name, fn, obs, obs_mask, *args, **kwargs):
...
+    *args,
+    obs: Optional[torch.Tensor] = None,
+    obs_mask: Optional[torch.Tensor] = None,
+    infer: Optional[Dict[str, Union[str, bool]]] = None,
```
ordabayevy, line 129: Made these args explicit.

pyro/primitives.py
```diff
@@ -374,7 +406,9 @@ def __init__(self, *args, **kwargs):
 
 
 @contextmanager
-def plate_stack(prefix, sizes, rightmost_dim=-1):
+def plate_stack(
+    prefix: str, sizes: Sequence[int], rightmost_dim: int = -1
```
ordabayevy, line 408: The docstring says `sizes` is iterable, however, iterable is not reversible.

pyro/primitives.py
```diff
@@ -462,7 +498,7 @@ def module(name, nn_module, update_module_params=False):
                         param_name
                     ] = target_state_dict[_name]
                 else:
-                    nn_module._parameters[mod_name] = target_state_dict[_name]
+                    nn_module._parameters[mod_name] = target_state_dict[_name]  # type: ignore[assignment]
```
ordabayevy, line 499: `nn_module._parameters`'s type is `nn.Parameter` and `target_state_dict[_name]` is `torch.Tensor`.

pyro/primitives.py
ordabayevy: Great idea. However, in this case it turns out that `make docs` fails without `TorchDistribution` and `ParamStoreDict` imported.  And `constraints` is used in actual code.

fritzo (approved):
Looks great!

PR #3298, opened 2023-11-27
Typo in the documentation example

pavleb:
The parameter value in the code is ``a``, where as in the example was ``s``.


fritzo (approved):

fritzo:
Thanks for the fix!

PR #3299, opened 2023-12-03
Type annotate TraceMessenger

ordabayevy:


pyro/poutine/guide.py
```diff
@@ -21,22 +21,22 @@ class GuideMessenger(TraceMessenger, ABC):
...
+    def __getstate__(self) -> Dict[str, object]:
         # Avoid pickling the trace.
-        state = super().__getstate__()
-        state.pop("trace")
+        state = self.__dict__.copy()
+        del state["trace"]
         return state
 
-    def __call__(self, *args, **kwargs) -> Dict[str, torch.Tensor]:  # type: ignore[override]
+    def __call__(self, *args, **kwargs) -> Dict[str, Optional[torch.Tensor]]:  # type: ignore[override]
```
fritzo, line 39: I'm surprised this is `Optional[Tensor]` rather than just `Tensor`. It might make more sense to assert or filter in the samples loop below, so callers know they are getting tensors.

fritzo (commented):
Looks great!  I have just one comment on `GuideMessenger.__call__()`

pyro/poutine/guide.py
ordabayevy: Added assert statements in the for loop (a bit longer code but probably better than filtering with if statement)

Issue #3297, opened 2023-11-24
Recommend an external normalizing flow library


francois-rozet:
Hello :wave: As the development of Pyro's normalizing flow API has stopped for quite some time, I have been discussing with @fritzo about recommending an external normalizing flow library that would be compatible with Pyro.

I believe that [Zuko](https://github.com/probabilists/zuko) would be a great recommendation, but disclaimer I am its maintainer.

### Why Zuko?

I have extensively used and developed normalizing flows for my research and I always preferred the approach Pyro takes to define parametrized and conditional transformations over other libraries ([nflows](https://github.com/bayesiains/nflows), [FrEIA](https://github.com/vislearn/FrEIA), [FlowTorch](https://github.com/stefanwebb/flowtorch), ...). In particular, I liked that Pyro's distributions/transformations shared the interface of PyTorch distributions/transformations, which is not the case of other libraries.

Hence, I started (2 years ago) coding my own normalizing flows for my research and at some point (1 year ago) decided to publish it as an open-source library: [Zuko](https://github.com/probabilists/zuko). The project has grown and now features many flow architectures (MAF, NSF, CNF, NAF, UMNN, GF, ...), automatic tests, a [fully documented API](https://zuko.readthedocs.io) and tutorials. The repository does not have as many stars as other libraries (mainly because I never advertised it), but actually has [many users](https://www.pepy.tech/projects/zuko).

Most importantly, Zuko is very compatible with Pyro as it relies heavily on PyTorch's distributions interface. In Zuko, **a flow is a distribution**, with `sample`, `log_prob`, `even_shape`, etc.

> Zuko is a reference to a powerful firebender (pyromancer) in the [Avatar](https://wikipedia.org/wiki/Avatar:_The_Last_Airbender) cartoon.

### Roadmap

If the Pyro team agrees with recommending Zuko, I propose to

- [ ] Update the current [normalizing flow tutorial](https://github.com/pyro-ppl/pyro/blob/dev/tutorial/source/normalizing_flows_i.ipynb) to mention that the flow API is not actively developed anymore.
- [ ] Add a simple normalizing flow tutorial with Zuko. This would probably take inspiration from [Zuko's tutorials](https://zuko.readthedocs.io/en/stable/tutorials.html).
- [ ] Add a SVI tutorial with a normalizing flow guide.
- [ ] Add a VAE tutorial with a normalizing flow prior.

WDYT?

martinjankowiak:
[...]

PR #3302, opened 2023-12-05
Add tutorials using normalizing flows

francois-rozet:
This PR adds tutorials (SVI and VAE) using normalizing flows. The normalizing flow implementations are loaded from the [Zuko](https://github.com/probabilists/zuko) library. By the way, I think it could be interesting to add a "compatibility with external tools" section to the examples.

In addition (as discussed in #3297), this PR also updates the normalizing flow tutorial to mention that Pyro's built-in flow API is not developed anymore. I also fixed a mistake ($\det | \frac{\partial f}{\partial x} |$ should be $| \det \frac{\partial f}{\partial x} |$) in the math and cleaned up a bit.


martinjankowiak:
@francois-rozet isn't `rsample_and_log_prob` zuko-specific?

francois-rozet:
Indeed, but the class still works if `dist` does not implement `rsample_and_log_prob`. So it is not actually necessary to import `zuko`. Do you have a preference for `pyro.distributions` or `contrib.zuko` ?

BTW, `Transform.call_and_ladj` (also Zuko-specific) and `Distribution.rsample_and_log_prob` could be worth moving upstream (to torch itself), as they allow sampling and scoring without cache.

tutorial/source/normalizing_flows_intro.ipynb
```diff
@@ -4,10 +4,12 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Normalizing Flows - Introduction (Part 1)\n",
+    "# Normalizing Flows - Introduction\n",
+    "\n",
+    "This tutorial introduces Pyro's built-in normalizing flows. It is independent of most of Pyro, but users may want to read about distribution shapes in the [Tensor Shapes Tutorial](http://pyro.ai/examples/tensor_shapes.html).\n",
+    "\n",
+    "> The development of Pyro's built-in flows has stopped in favor of external libraries. We recommend [Zuko](https://zuko.readthedocs.io) as it is compatible with Pyro, implements many flow architectures and is [well documented](https://zuko.readthedocs.io).\n",
```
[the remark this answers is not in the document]
francois-rozet: I agree. I wanted to do that initially, but some libraries are unmaintained (`nflows`, `FlowTorch`) and some I don't know how they integrate with Pyro (`FrEIA`, `normflows`) as they don't follow PyTorch's distribution API. So should I just put all of them? I can also mention the [awesome-normalizing-flows](https://github.com/janosh/awesome-normalizing-flows) repo.

For the NF tuto, I only made minor edits, which either fix the error mentioned previously or simplify math notations (removing parentheses when clear from context). For example,

image


became

image

or

image


became

image

My editor also removed trailing white-spaces, but nothing else changed.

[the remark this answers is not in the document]
martinjankowiak: i defer to your judgment as to which libraries are worth mentioning, though i'd err on the side of more. you can just make an explicit caveat like "although we note that some of these libraries may no longer be actively maintained and may have APIs that are not directly compatible with Pyro (and thus may require a custom wrapper for use in Pyro)."

martinjankowiak:
@fritzo do you have a preference between `pyro.distributions` and `contrib.zuko`? i do not. the former is probably more discoverable...

martinjankowiak:
note you can put a test for your contrib code in `tests/contrib`

pyro/contrib/zuko.py
```diff
@@ -0,0 +1,46 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+from torch import Size, Tensor
+
+import pyro
+
+
+class Zuko2Pyro(pyro.distributions.TorchDistribution):
+    r"""Wraps a Zuko (or PyTorch) distribution as a Pyro distribution."""
```
martinjankowiak, line 11: can you expand this docstring? link to zuko? etc

also the `(or PyTorch)` seems confusing since we have `pyro.distributions.TorchDistribution` for that

pyro/contrib/zuko.py
```diff
@@ -0,0 +1,46 @@
...
+        elif self.has_rsample:
+            x = self.dist.rsample(shape)
+        else:
+            x = self.dist.sample(shape)
+
+        return x
+
+    def log_prob(self, x: Tensor) -> Tensor:
+        if x in self.cache:
+            return self.cache[x]
```
martinjankowiak, line 41: @fritzo can you please take a look at the caching logic?

tests/contrib/test_zuko.py
```diff
@@ -0,0 +1,56 @@
...
+import torch
+
+import pyro
+from pyro.contrib.zuko import Zuko2Pyro
+from pyro.infer import SVI, Trace_ELBO
+from pyro.optim import Adam
+
+
+@pytest.mark.parametrize("multivariate", [True, False])
+def test_Zuko2Pyro(multivariate: bool):
```
martinjankowiak, line 15: @fritzo anything else worth testing for?

tutorial/source/normalizing_flows_intro.ipynb
```diff
@@ -4,10 +4,12 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Normalizing Flows - Introduction (Part 1)\n",
+    "# Normalizing Flows - Introduction\n",
+    "\n",
+    "This tutorial introduces Pyro's built-in normalizing flows. It is independent of most of Pyro, but users may want to read about distribution shapes in the [Tensor Shapes Tutorial](http://pyro.ai/examples/tensor_shapes.html).\n",
+    "\n",
+    "> The development of Pyro's built-in flows has stopped in favor of external libraries, such as [Zuko](https://github.com/probabilists/zuko), [nflows](https://github.com/bayesiains/nflows), [normflows](https://github.com/VincentStimper/normalizing-flows) or [FlowTorch](https://github.com/stefanwebb/flowtorch). Some of these libraries may have interfaces that are not directly compatible with Pyro. See the [SVI with flow guide](svi_flow_guide.ipynb) and [VAE with flow prior](vae_flow_prior.ipynb) tutorials for example usages of [Zuko](https://github.com/probabilists/zuko) within Pyro.\n",
```
martinjankowiak, line 11: think this should point to flowtorch.ai?

tutorial/source/normalizing_flows_intro.ipynb
```diff
@@ -4,10 +4,12 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Normalizing Flows - Introduction (Part 1)\n",
+    "# Normalizing Flows - Introduction\n",
+    "\n",
+    "This tutorial introduces Pyro's built-in normalizing flows. It is independent of most of Pyro, but users may want to read about distribution shapes in the [Tensor Shapes Tutorial](http://pyro.ai/examples/tensor_shapes.html).\n",
```
martinjankowiak, line 9: can you add something like this to the top of both example notebooks?

"In this notebook we leverage normalizing flows implemented in Zuko, but similar functionality could also be obtained with another PyTorch-based normalizing library."

tutorial/source/svi_flow_guide.ipynb
```diff
@@ -0,0 +1,236 @@
...
+    "        pyro.sample(\"z\", Zuko2Pyro(flow(x.flatten(-2))))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## SVI\n",
+    "\n",
+    "We train our guide with a standard stochastic variational inference (SVI) pipeline. We use 16 particles to reduce the variance of the ELBO and clip the norm of the gradients to make training more stable."
```
martinjankowiak, line 113: quite aggressive clipping?

tutorial/source/svi_flow_guide.ipynb
```diff
@@ -0,0 +1,236 @@
...
+    "    ], dim=-1).unflatten(-1, (2, 2))\n",
+    "\n",
+    "    return pyro.distributions.MultivariateNormal(mu, cov)\n",
+    "\n",
+    "def model(x: Tensor):\n",
+    "    with pyro.plate(\"batch\", len(x)):\n",
+    "        z = pyro.sample(\"z\", prior)\n",
+    "\n",
+    "        with pyro.plate(\"obs\", 5):\n",
+    "            pyro.sample(\"x\", likelihood(z), obs=x.transpose(0, 1))"
```
martinjankowiak, line 62: are all these transposes necessary?

martinjankowiak:
@fritzo do we need to add zuko to `EXTRAS_REQUIRE`?

pyro/contrib/zuko.py
francois-rozet: I don't think `TorchDistribution` does the same think. `Zuko2Pyro` takes any **instance** of distribution and makes it a valid Pyro distribution.

I'll add a page in the docs.

pyro/contrib/zuko.py
francois-rozet: Note that I don't clear the cache as I consider that `Zuko2Pyro` objects should be temporary.

tutorial/source/svi_flow_guide.ipynb
francois-rozet: That's a value I use quite often, but I don't have much intuition about it.

tutorial/source/svi_flow_guide.ipynb
francois-rozet: Either there or in the guide.

francois-rozet:
@martinjankowiak I have addressed your comments (at least the ones I could). I don't think it is necessary to add zuko in `EXTRAS_REQUIRE` as even `contrib/zuko.py` does not rely on it. However, it seems like `doctest` tries to run the example I added in the docstring and fails because it does not import `zuko`. I can modify the example to be a `code-block`, which should not be executed.

tutorial/source/svi_flow_guide.ipynb
martinjankowiak: it seems like this could be re-organized more elegantly to reduce the number of transposes

martinjankowiak (dismissed):
lgtm. @fritzo could you please take a final pass, especially w.r.t. caching logic?

tutorial/source/svi_flow_guide.ipynb
francois-rozet: I think I can reduce the number of `transpose` from 2 to 1.

fritzo (commented):
@francois-rozet thank you for cleaning up our existing tutorials and for adding this new tutorial! I think these will help lots of users provide exposure for Zuko. And thanks for your patience in this very slow holiday review cycle. I think we can push a Pyro 1.9 release soon after this merges.

LGTM, subject to @martinjankowiak's review and a couple nits.

pyro/contrib/zuko.py
```diff
@@ -0,0 +1,77 @@
...
+`tutorial/vae_flow_prior.ipynb`.
+"""
+
+import torch
+from torch import Size, Tensor
+
+import pyro
+
+
+class Zuko2Pyro(pyro.distributions.TorchDistribution):
```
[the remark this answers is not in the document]
francois-rozet: Ahah I thought it was "cool" :laughing:, will fix

pyro/contrib/zuko.py
francois-rozet: Note that the `cache` is only used when `x` is sampled with `rsample_and_log_prob`. But, yes I will mention it in the docstring.

tests/contrib/test_zuko.py
francois-rozet: I can add a "fake" `rsample_and_log_prob` method to the `dist` object and check that it runs with that.

```python
def rsample_and_log_prob(self, shape):
    x = self.rsample(shape)
    return x, self.log_prob(x)

dist.rsample_and_log_prob = rsample_and_log_prob
```

This would at least test that `ZukoToPyro` works as expected.

francois-rozet:
I addressed your comments, and I think the tests have passed (went to bed before the end).

tests/contrib/test_zuko.py
martinjankowiak: sounds like a reasonable solution to me

martinjankowiak:
lgtm apart from possibly adding a `rsample_and_log_prob` test

martinjankowiak (approved):

francois-rozet:
> apart from possibly adding a rsample_and_log_prob test

I added it in the test.

tests/contrib/test_zuko.py
francois-rozet: It is done!

martinjankowiak:
sorry i missed that thanks

francois-rozet:
🎉

martinjankowiak:
thanks @francois-rozet !!

Issue #3301, opened 2023-12-05
Possible bug in gate_logits of ZeroInflatedDistribution


reemagit:
When I use the `gate_logits` parameter of ZeroInflatedPoisson I get unexpected behavior in the number of zeros sampled by the distribution. This doesn't happen if I use the `gate` parameter.

```python
N=1000 # num samples
rate=torch.tensor(100.) ## arbitrary rate
p = torch.tensor(0.3) # inflation probability
logit = torch.log(p / (1-p)) # logit of inf. prob.

x1 = (dist.ZeroInflatedPoisson(rate=rate), gate_logits=logit).sample([N])==0).sum()/N 
x2 = (dist.ZeroInflatedPoisson(rate=rate), gate_logits=logit).sample([N])==0).sum()/N # prints 1, i.e. all zeros

print(x1) # prints approximately 0.3
print(x2) # prints 1.
```

I would expect that both `x1` and `x2` are close to 0.3 (the inflation probability). Am I missing something or is this a bug?

fritzo:
@reemagit is there a typo? x1 and x2 appear to be identical?

reemagit:
@fritzo I apologize, I corrected a couple of typos, thanks.

ordabayevy:
@reemagit can you check and see if #3303 fixes the bug?

reemagit:
I currently don't have the development version installed, so I can't check, sorry. However, looking at the fix I think that that was the problem, since I've had similar issues with `logits_to_probs` and `probs_to_logits` in the past.

ordabayevy:
Closed via #3303

PR #3303, opened 2023-12-06
Fix `gate` & `gate_logits` in `ZeroInflatedDistribution`

ordabayevy:
Addresses #3301


fritzo (approved):
Yikes, another bug due to logits name ambiguity.

PR #3306, opened 2023-12-16
Type annotate poutines

ordabayevy:
- escape_messenger
- indep_messenger
- infer_config_messenger
- lift_messenger
- mask_messenger
- uncondition_messenger

### DIFF
--- a/pyro/poutine/enum_messenger.py
+++ b/pyro/poutine/enum_messenger.py
@@ -22,7 +22,7 @@ def _tmc_mixture_sample(msg: Message) -> torch.Tensor:
     # find batch dims that aren't plate dims
     batch_shape = [1] * len(dist.batch_shape)
     for f in msg["cond_indep_stack"]:
-        if f.vectorized:
+        if f.dim is not None:
             batch_shape[f.dim] = f.size if f.size > 0 else dist.batch_shape[f.dim]
     batch_shape_tuple = tuple(batch_shape)
 
@@ -71,7 +71,7 @@ def _tmc_diagonal_sample(msg: Message) -> torch.Tensor:
     # find batch dims that aren't plate dims
     batch_shape = [1] * len(dist.batch_shape)
     for f in msg["cond_indep_stack"]:
-        if f.vectorized:
+        if f.dim is not None:
             batch_shape[f.dim] = f.size if f.size > 0 else dist.batch_shape[f.dim]
     batch_shape_tuple = tuple(batch_shape)
 
--- a/pyro/poutine/enum_messenger.py
+++ b/pyro/poutine/enum_messenger.py
@@ -22,7 +22,8 @@ def _tmc_mixture_sample(msg: Message) -> torch.Tensor:
     # find batch dims that aren't plate dims
     batch_shape = [1] * len(dist.batch_shape)
     for f in msg["cond_indep_stack"]:
-        if f.dim is not None:
+        if f.vectorized:
+            assert f.dim is not None
             batch_shape[f.dim] = f.size if f.size > 0 else dist.batch_shape[f.dim]
     batch_shape_tuple = tuple(batch_shape)
 
@@ -71,7 +72,8 @@ def _tmc_diagonal_sample(msg: Message) -> torch.Tensor:
     # find batch dims that aren't plate dims
     batch_shape = [1] * len(dist.batch_shape)
     for f in msg["cond_indep_stack"]:
-        if f.dim is not None:
+        if f.vectorized:
+            assert f.dim is not None
             batch_shape[f.dim] = f.size if f.size > 0 else dist.batch_shape[f.dim]
     batch_shape_tuple = tuple(batch_shape)
 
--- a/pyro/poutine/indep_messenger.py
+++ b/pyro/poutine/indep_messenger.py
@@ -2,7 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import numbers
-from typing import Iterator, NamedTuple, Optional, Tuple
+from dataclasses import dataclass
+from typing import Iterator, Optional, Tuple
 
 import torch
 from typing_extensions import Self
@@ -12,7 +13,8 @@ from pyro.poutine.runtime import _DIM_ALLOCATOR, Message
 from pyro.util import ignore_jit_warnings
 
 
-class CondIndepStackFrame(NamedTuple):
+@dataclass
+class CondIndepStackFrame:
     name: str
     dim: Optional[int]
     size: int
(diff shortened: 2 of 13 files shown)


pyro/poutine/enum_messenger.py
```diff
@@ -71,7 +71,7 @@ def _tmc_diagonal_sample(msg: Message) -> torch.Tensor:
     # find batch dims that aren't plate dims
     batch_shape = [1] * len(dist.batch_shape)
     for f in msg["cond_indep_stack"]:
-        if f.vectorized:
+        if f.dim is not None:
```
ordabayevy, line 74: `f.vectorized` just check if `f.dim is not None` so I took the liberty to replace it here. Alternatively I can change it back as follows:
```py
if f.vectorized:
    assert f.dim is not None
```

pyro/poutine/indep_messenger.py
```diff
@@ -2,40 +2,47 @@
...
 
-from .messenger import Messenger
-from .runtime import _DIM_ALLOCATOR
 
+@dataclass
+class CondIndepStackFrame:
+    name: str
+    dim: Optional[int]
+    size: int
+    counter: int
```
ordabayevy, line 21: Initially I tried to use `class CondIndepStackFrame(NamedTuple)`, however, there is code in other places that tries to assign `frame.full_size` attribute which throws an error because `NamedTuple` is immutable(?).

pyro/poutine/indep_messenger.py
```diff
@@ -59,7 +66,13 @@ class IndepMessenger(Messenger):
...
     """
 
-    def __init__(self, name=None, size=None, dim=None, device=None):
+    def __init__(
+        self,
+        name: str,
+        size: int,
+        dim: Optional[int] = None,
+        device: Optional[str] = None,
+    ):
```
ordabayevy, line 75: Here it doesn't make sense for `name` and `size` to be `None` (e.g. `self.size` is used to create `torch.arange` on line 137). So I decided to remove their default `None` values. Let me know if I should add them back.

pyro/poutine/enum_messenger.py
fritzo: yeah I'd aim for the more verbose version because it makes intent clearer

pyro/poutine/indep_messenger.py
fritzo: Hmm, immutability is a great property to reason about when reading & maintaining a codebase. I'd be a shame to lose that property even as we get type hints. Is there some way we could keep this class clean and immutable, even if that hack gets uglier? Even something like this:
```py
@dataclass(frozen=True, slots=True)  # frozen + slots is close to namedtuple
class CondIndepStackFrame:
    name: str
    dim: Optional[int]
    size: int
    counter: int
    _full_size: Optional[int] = None  # hack: this is actually mutable

    @property
    def full_size(self) -> int:
        return self.size if self._full_size is None else self._full_size

    @full_size.setter
    def full_size(self, value: int) -> None:
        object.__setattr__(self, "_full_size", value)

    # ... __eq__ etc. that ignore .full_size
```

pyro/poutine/trace_struct.py
```diff
@@ -424,7 +424,7 @@ def symbolize_dims(self, plate_to_symbol: Optional[Dict[int, str]] = None) -> No
         self.plate_to_symbol = plate_to_symbol
         self.symbol_to_dim = symbol_to_dim
 
-    def pack_tensors(self, plate_to_symbol: Optional[Dict[int, str]] = None) -> None:
+    def pack_tensors(self, plate_to_symbol: Optional[Dict[str, str]] = None) -> None:
```
fritzo, line 427: yikes good catch! looks like this made it past my last review 😊

fritzo (changes requested):
It's great to see such careful maintenance work, kudos!

I just have a couple comments about code clarity and immutability.

pyro/poutine/indep_messenger.py
ordabayevy: I just tried this approach, however, python complained that you cannot use setattr with `frame.full_size = size` if `frozen=True`. Also `slots` is new in Python 3.10. I'll think more about other ways of keeping immutability, don't have any at the moment.

fritzo: What if we kept `CondIndepStackFrame` as `NamedTuple` but moved `FullSize` to a global `WeakKeyDictionary`?  There are only 4 code locations that use the hacky `.full_size` attribute. We could either change those to access the `WeakKeyDictionary` or make a property that does it for us:

<details>

class CondIndepStackFrame(NamedTuple):
    name: str
    dim: Optional[int]
    size: int
    counter: int

    @property
    def full_size(self) -> Optional[int]:
        return COND_INDEP_FULL_SIZE.get(self)

    @full_size.getter
    def full_size(self, value: int) -> None:
        COND_INDEP_FULL_SIZE[self] = value

COND_INDEP_FULL_SIZE: WeakKeyDictionary[CondIndepStackFrame, int] = WeakKeyDictionary()

</details>

fritzo: And apologies for that `full_size` hack 😊

ordabayevy: I assume you meant `@full_size.setter`. Using `NamedTuple` and `WeakKeyDictionary` I get this error:

```
  File "/mnt/disks/dev/repos/pyro/pyro/poutine/subsample_messenger.py", line 154, in _process_message
    frame.full_size = self.size  # Used for param initialization.
  File "/mnt/disks/dev/repos/pyro/pyro/poutine/indep_messenger.py", line 28, in full_size
    COND_INDEP_FULL_SIZE[self] = value
  File "/home/yordabay/anaconda3/envs/pyro/lib/python3.8/weakref.py", line 396, in __setitem__
    self.data[ref(key, self._remove)] = value
TypeError: cannot create weak reference to 'CondIndepStackFrame' object
```

It seems like you cannot create weakrefs to tuples (https://stackoverflow.com/questions/58312618/is-there-a-way-to-support-weakrefs-with-collections-namedtuple)

I also tried it with a dataclass and using `frozen=True` but then it doesn't allow to use setter:

```
  File "/mnt/disks/dev/repos/pyro/pyro/poutine/subsample_messenger.py", line 154, in _process_message
    frame.full_size = self.size  # Used for param initialization.
  File "<string>", line 4, in __setattr__
dataclasses.FrozenInstanceError: cannot assign to field 'full_size'
```

ordabayevy: It seems like Python doesn't really want to allow setting attributes on immutable objects. It is also strange that an old `namedtuple` had no problems with that.

fritzo: Mind if I push a fix?

fritzo (approved):
@ordabayevy your changes LGTM. Could you PTAL at my `.full_size` solution? Feel free to merge when tests pass.

PR #3307, opened 2023-12-17
Numpy Import Convention

RatishT:
Dear Pyro contributors,

I've taken the initiative to address and resolve several linting warnings pointed out by a pylint plugin called dslinter. The code suggestions that it gave were:

1. **torch.log() Variable Clamping:**
   - Enclosed the variable in the torch.log() operation with torch.clamp() to establish a minimum value. This modification ensures a safeguard against potential NaN errors, particularly when the input variable is exceptionally small.
   - Before:
     ```python
     result = torch.log(input_var)
     ```
   - After:
     ```python
     result = torch.log(torch.clamp(input_var, min=1e-10))
     ```
2. **Numpy Import Convention:**
   - I added "as np" to a numpy import, matching standard convention and consistency in code

I invite your valuable insights and feedback on these modifications to guarantee that they align seamlessly with our shared coding standards and project requirements.

### DIFF
--- a/pyro/contrib/bnn/hidden_layer.py
+++ b/pyro/contrib/bnn/hidden_layer.py
@@ -94,7 +94,7 @@ class HiddenLayer(TorchDistribution):
         KL_A = torch.pow(self.A_mean / self.A_prior_scale, 2.0).sum()
         KL_A -= self.dim_X * self.dim_H
         KL_A += torch.pow(self.A_scale / self.A_prior_scale, 2.0).sum()
-        KL_A -= 2.0 * torch.log(torch.clamp(self.A_scale / self.A_prior_scale, 1e-10)).sum()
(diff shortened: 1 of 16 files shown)


martinjankowiak:
thanks but i don't think the `log` change is necessary. in any case if we were to do something like that we would use `torch.finfo` and not some ad hoc number like `1e-10`.

the `np` change looks fine to me

RatishT:
> thanks but i don't think the `log` change is necessary. in any case if we were to do something like that we would use `torch.finfo` and not some ad hoc number like `1e-10`.
> 
> the `np` change looks fine to me

Thank you for your reply! I have reverted the torch.clamp in the torch.log calls
For future reference, would it be better to use ```variable.clamp(min=torch.finfo(variable.dtype).tiny)```?

ordabayevy (approved):

PR #3309, opened 2024-01-02
Type annotate messengers

ordabayevy:
- plate_messenger
- reentrant_messenger
- reparam_messenger

### DIFF
--- a/pyro/distributions/torch_distribution.py
+++ b/pyro/distributions/torch_distribution.py
@@ -3,6 +3,7 @@
 
 import warnings
 from collections import OrderedDict
+from typing import Callable
 
 import torch
 from torch.distributions.kl import kl_divergence, register_kl
@@ -15,7 +16,7 @@ from .score_parts import ScoreParts
 from .util import broadcast_shape, scale_and_mask
 
 
-class TorchDistributionMixin(Distribution):
+class TorchDistributionMixin(Distribution, Callable):
     """
     Mixin to provide Pyro compatibility for PyTorch distributions.
 
--- a/pyro/poutine/reparam_messenger.py
+++ b/pyro/poutine/reparam_messenger.py
@@ -2,16 +2,33 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import warnings
-from typing import Callable, Dict, Union
+from typing import (
+    TYPE_CHECKING,
+    Callable,
+    Dict,
+    Generic,
+    List,
+    Optional,
+    TypeVar,
+    Union,
+)
 
 import torch
+from typing_extensions import ParamSpec
 
-from .messenger import Messenger
-from .runtime import effectful
+from pyro.distributions.torch_distribution import TorchDistributionMixin
+from pyro.poutine.messenger import Messenger
+from pyro.poutine.runtime import Message, effectful
+
+if TYPE_CHECKING:
+    from pyro.infer.reparam.reparam import Reparam
+
+P = ParamSpec("P")
+T = TypeVar("T")
 
 
 @effectful(type="get_init_messengers")
-def _get_init_messengers():
+def _get_init_messengers() -> List[Messenger]:
     return []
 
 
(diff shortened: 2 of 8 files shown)


pyro/distributions/torch_distribution.py
```diff
@@ -15,7 +16,7 @@
 from .util import broadcast_shape, scale_and_mask
 
 
-class TorchDistributionMixin(Distribution):
+class TorchDistributionMixin(Distribution, Callable):
```
ordabayevy, line 19: This is a hack so that mypy understands `assert isinstance(msg["fn"], TorchDistributionMixin)`

pyro/poutine/runtime.py
```diff
@@ -64,26 +63,26 @@ class InferDict(TypedDict, total=False):
...
+class Message(TypedDict, Generic[P, T], total=False):
     type: str
     name: Optional[str]
-    fn: Union[Callable, TorchDistributionMixin]
+    fn: Callable[P, T]
     is_observed: bool
     args: Tuple
     kwargs: Dict
-    value: Optional[torch.Tensor]
+    value: Optional[T]
```
ordabayevy, line 73: I've learned this neat trick with `Generic` and `TypeVar` where the type of `value` can be inferred from the `Callable` signature. I also fixed `effectful` so that it gives the correct signature for the decorated function when the return type is diferent from `torch.Tensor` (e.g. `reparam_messenger._get_init_messengers`).

pyro/poutine/runtime.py
```diff
@@ -368,6 +365,7 @@ def _fn(
             )
             # apply the stack and return its return value
             apply_stack(msg)
+            assert msg["value"] is not None
```
ordabayevy, line 368: Is this always correct? All tests have passed

pyro/poutine/reparam_messenger.py
```diff
@@ -80,10 +102,10 @@ def _pyro_sample(self, msg):
         # similarly be safely applied twice, with the second application
         # avoiding overwriting the original application.
         for m in _get_init_messengers():
```
ordabayevy, line 104: E.g. here it is not expected that `_get_init_messengers()` can return `None`. See the comment below about the return type of the `effectful`.

pyro/poutine/runtime.py
```diff
@@ -1,13 +1,12 @@
 # Copyright (c) 2017-2019 Uber Technologies, Inc.
 # SPDX-License-Identifier: Apache-2.0
 
-from __future__ import annotations
```
ordabayevy, line 4: I came to a conclusion that it is better not to use this

pyro/poutine/runtime.py
fritzo: @eb8680 ?

fritzo (commented):
Looks great, just a couple nits.

pyro/distributions/torch_distribution.py
ordabayevy: Using `@Callable.register` doesn't seem to work (the code after `assert isinstance(msg["fn"], TorchDistributionMixin)` is unreachable). Maybe this has to do with `typing.Callable` vs `collections.abc.Callable`?

It is strange that `assert isinstance(msg["fn"], TorchDistributionMixin)` doesn't work in the first place because `assert isinstance(msg["fn"], Distribution)` works fine. Seems like a Mypy bug.

In the following PR I want to add `warn_unreachable=True` mypy flag that can catch cases like this, otherwise unreachable code is not checked by mypy and fails silently.

pyro/poutine/runtime.py
```diff
@@ -339,18 +338,18 @@ def effectful(
...
-        *args: P.args,
+        *args: _P.args,
         name: Optional[str] = None,
         infer: Optional[InferDict] = None,
-        obs: Optional[torch.Tensor] = None,
-        **kwargs: P.kwargs,
-    ) -> Union[T, torch.Tensor, None]:
+        obs: Optional[_T] = None,
+        **kwargs: _P.kwargs,
+    ) -> Optional[_T]:
```
ordabayevy, line 346: Changed this back to return `Optional` and removed the `assert  msg["value"] is not None` line. One concern I have is that if `_T` itself is `None` then it will raise an assertion error.

ordabayevy (commented):
@fritzo can you have another look? I have addressed your comments.

pyro/distributions/torch_distribution.py
fritzo: Well thanks for checking!

fritzo (approved):
LGTM, Thanks for the ping!

PR #3320, opened 2024-02-04
Port implementation of SimplexToOrderedTransform from numpyro

peblair:
This pull request adds support to Pyro for [`SimplexToOrderedTransform`](https://github.com/pyro-ppl/numpyro/blob/01089cfad4078e21e062c02147bcc636331548f1/numpyro/distributions/transforms.py#L919). The implementation used is a port of the `numpyro` implementation of the same class.

I was working on implementing a model based on [this numpyro guide](https://num.pyro.ai/en/stable/tutorials/ordinal_regression.html#Principled-prior-with-Dirichlet-Distribution) using Pyro, and this was the one thing which was necessary to reimplement.

Because this class is meant to be used on values greater than zero, I have additionally added the ability to specify the family of distribution which should be used in the transformation tests; this new transformation uses a Dirichlet distribution for its tests.

### DIFF
--- a/pyro/distributions/transforms/simplex_to_ordered.py
+++ b/pyro/distributions/transforms/simplex_to_ordered.py
@@ -1,0 +1,80 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+from torch.distributions.transforms import Transform
+
+from .. import constraints
+
+
+def expit(x):
+    return 1.0 / (1 + torch.exp(-x))
+
+
+def logit(x):
+    return torch.log(x / (1.0 - x))
+
+
+# This class is a port of https://num.pyro.ai/en/stable/_modules/numpyro/distributions/transforms.html#SimplexToOrderedTransform
+class SimplexToOrderedTransform(Transform):
+    """
+    Transform a simplex into an ordered vector (via difference in Logistic CDF between cutpoints)
+    Used in [1] to induce a prior on latent cutpoints via transforming ordered category probabilities.
+
+    :param anchor_point: Anchor point is a nuisance parameter to improve the identifiability of the transform.
+        For simplicity, we assume it is a scalar value, but it is broadcastable x.shape[:-1].
+        For more details please refer to Section 2.2 in [1]
+
+    **References:**
+
+    1. *Ordinal Regression Case Study, section 2.2*,
+       M. Betancourt, https://betanalpha.github.io/assets/case_studies/ordinal_regression.html
+
+    """
+
+    domain = constraints.simplex
+    codomain = constraints.ordered_vector
+
+    def __init__(self, anchor_point=None):
+        super().__init__()
+        self.anchor_point = (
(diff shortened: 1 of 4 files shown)


pyro/distributions/transforms/simplex_to_ordered.py
```diff
@@ -0,0 +1,80 @@
...
+
+from .. import constraints
+
+
+def expit(x):
+    return 1.0 / (1 + torch.exp(-x))
+
+
+def logit(x):
+    return torch.log(x / (1.0 - x))
```
fritzo, line 15: Could you replace these with `torch.special.expit()` and `torch.specical.logit()` which should be more numerically stable?

pyro/distributions/transforms/simplex_to_ordered.py
```diff
@@ -0,0 +1,80 @@
...
+        s = expit(y)
+        # x0 = s0, x1 = s1 - s0, x2 = s2 - s1,..., xn = 1 - s[n-1]
+        # add two boundary points 0 and 1
+        s = torch.concat(
+            [torch.zeros_like(s)[..., :1], s, torch.ones_like(s)[..., :1]], dim=-1
+        )
+        x = s[..., 1:] - s[..., :-1]
+        return x
+
+    def log_abs_det_jacobian(self, x, y, intermediates=None):
```
fritzo, line 60: I believe `intermediates` is a numpyro-specific arg and can be removed in Pyro.

pyro/distributions/transforms/simplex_to_ordered.py
```diff
@@ -0,0 +1,80 @@
...
+        # |dp/dc| = |dx/dy| = prod(ds/dy) = prod(expit'(y))
+        # we know log derivative of expit(y) is `-softplus(y) - softplus(-y)`
+        J_logdet = (
+            torch.nn.functional.softplus(y) + torch.nn.functional.softplus(-y)
+        ).sum(-1)
+        return J_logdet
+
+    def tree_flatten(self):
+        return (self.anchor_point,), (("anchor_point",), dict())
```
fritzo, line 69: I believe the `tree_flatten()` method is numpyro-specific and can be removed in Pyro.

fritzo (changes requested):
:+1: Looks great, thanks for porting this!  I'll take a closer review of the math on second pass.

Thanks for adding tests and a docstring!

Could you also register this in [docs/source/distributions.rst](https://github.com/pyro-ppl/pyro/blob/dev/docs/source/distributions.rst#transforms) in the Transforms section? We try to keep that section alphabetical. I usually `make docs` and check that the new class does show up in the generated docs.

pyro/distributions/transforms/simplex_to_ordered.py
peblair: Of course! I hadn't realized these were available.

peblair:
@fritzo Thanks for giving this a look! The issues you flagged should be fixed now. Let me know if there's anything else this needs.

fritzo (approved):
Looks great, thanks for making the changes.

PR #3321, opened 2024-02-05
Annotate handlers & add py.typed

ordabayevy:


pyro/poutine/handlers.py
```diff
@@ -115,82 +136,375 @@ def handler(fn=None, *args, **kwargs):
...
+    fn: Optional[Callable[_P, _T]] = None,
+    *args: Any,
+    **kwargs: Any,
+) -> Union[CollapseMessenger, Callable[_P, _T]]: ...
+
+
+@overload
+def condition(
+    data: Union[Dict[str, "torch.Tensor"], "Trace"],
+) -> ConditionMessenger: ...
```
ordabayevy, line 226: Some handlers have arguments that are not optional so `fn` has to be required in the signature. This is the trick with overloading to have `fn` as optional. Based on the signature mypy can figure out which type annotations to use.

pyro/primitives.py
```diff
@@ -94,14 +94,14 @@ def _masked_observe(
     name: str,
     fn: TorchDistributionMixin,
     obs: Optional[torch.Tensor],
-    obs_mask: torch.Tensor,
```
fritzo, line 97: Nice!  Is `torch.cuda.BoolTensor` a subclass of `torch.BoolTensor`?

fritzo (commented):
Looks great, I have just a couple questions.

pyro/poutine/handlers.py
```diff
@@ -115,82 +136,375 @@ def handler(fn=None, *args, **kwargs):
...
+    expose_all: bool = False,
+    hide: Optional[List[str]] = None,
+    expose: Optional[List[str]] = None,
+    hide_types: Optional[List[str]] = None,
+    expose_types: Optional[List[str]] = None,
+) -> Callable[_P, _T]: ...
+
+
+@_make_handler(BlockMessenger)
+def block(  # type: ignore[empty-body]
```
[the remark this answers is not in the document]
ordabayevy: Just checked it, it gives the same empty-body mypy error:

```
(pyro) yordabay@yo-dl-dev:/mnt/disks/dev/repos/pyro$ git diff
diff --git a/pyro/poutine/handlers.py b/pyro/poutine/handlers.py
index c54da6cf..404ae667 100644
--- a/pyro/poutine/handlers.py
+++ b/pyro/poutine/handlers.py
@@ -165,7 +165,7 @@ def block(
 
 
 @_make_handler(BlockMessenger)
-def block(  # type: ignore[empty-body]
+def block(
     fn: Optional[Callable[_P, _T]] = None,
     hide_fn: Optional[Callable[["Message"], Optional[bool]]] = None,
     expose_fn: Optional[Callable[["Message"], Optional[bool]]] = None,
@@ -175,7 +175,8 @@ def block(  # type: ignore[empty-body]
     expose: Optional[List[str]] = None,
     hide_types: Optional[List[str]] = None,
     expose_types: Optional[List[str]] = None,
-) -> Union[BlockMessenger, Callable[_P, _T]]: ...
+) -> Union[BlockMessenger, Callable[_P, _T]]:
+    pass
 
 
 @overload
(pyro) yordabay@yo-dl-dev:/mnt/disks/dev/repos/pyro$ make lint
ruff check .
black --check *.py pyro examples tests scripts profiler
Skipping .ipynb files as Jupyter dependencies are not installed.
You can fix this by running ``pip install "black[jupyter]"``
All done! ✨ 🍰 ✨
621 files would be left unchanged.
python scripts/update_headers.py --check
mypy --install-types --non-interactive pyro scripts
pyro/poutine/handlers.py:168: error: Missing return statement  [empty-body]
Found 1 error in 1 file (checked 321 source files)
make: *** [Makefile:24: lint] Error 1
```

pyro/primitives.py
```diff
@@ -94,14 +94,14 @@ def _masked_observe(
...
+    obs_mask: torch.BoolTensor,
     *args,
     **kwargs,
 ) -> torch.Tensor:
     # Split into two auxiliary sample sites.
     with poutine.mask(mask=obs_mask):
         observed = sample(f"{name}_observed", fn, *args, **kwargs, obs=obs)
-    with poutine.mask(mask=~obs_mask):
+    with poutine.mask(mask=~obs_mask):  # type: ignore[call-overload]
```
ordabayevy, line 104: Here, unfortunately, `~torch.BoolTensor` returns a `torch.Tensor`

pyro/primitives.py
fritzo: Yeah, it should be ok. Let's just keep in mind that `assert isinstance(obs_mask, torch.BoolTensor)` would fail in runtime.

fritzo (approved):

pyro/primitives.py
ordabayevy: Yeah. That's why I try to use `assert x is not None` when `x: Optional[torch.Tensor]`

PR #3325, opened 2024-02-10
Fix bug in CVAE example

### LINKED ISSUE #3273: Bug in `CVAE` example
There is a bug in `CVAE` example. The `target` input to the `MaskedBCELoss` is not binary (it has values of -1, 0, 1). This was discovered by the PyTorch 2.1 update which started to validate the inputs of `F.binary_cross_entropy_loss`.

> FAILED tests/test_examples.py::test_cpu[cvae/main.py --num-quadrant-inputs=1 --num-epochs=1] - subprocess.CalledProcessError: Command '['/opt/hostedtoolcache/Python/3.8.18/x64/bin/python', '/home/runner/work/pyro/pyro/examples/cvae/main.py', '--num-quadrant-inputs=1', '--num-epochs=1']' returned non-zero exit status 1.
= 1 failed, 148 passed, 97 skipped, 26558 deselected, 2 warnings in 1948.89s (0:32:28) =

FDUguchunhui:
Fixes #3273 

The bug is caused by input validation (only accept value between 0 or 1 otherwise raise error) of F.binary_cross_entropy_loss after Pytorch 2.1 update. The fix is to manually mask the input of the loss function to avoid invalid value rather than removing the loss of invalid input from the final loss result. Please refer to:https://github.com/pyro-ppl/pyro/issues/3273 for the initial report of the bug

### DIFF
--- a/examples/cvae/baseline.py
+++ b/examples/cvae/baseline.py
@@ -34,8 +34,12 @@ class MaskedBCELoss(nn.Module):
 
     def forward(self, input, target):
         target = target.view(input.shape)
-        loss = F.binary_cross_entropy(input, target, reduction="none")
-        loss[target == self.masked_with] = 0
+        # only calculate loss on target pixels (value = -1)
+        loss = F.binary_cross_entropy(
+            input[target != self.masked_with],
+            target[target != self.masked_with],
+            reduction="none",
+        )
         return loss.sum()
 
 
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -123,10 +123,7 @@ CPU_EXAMPLES = [
     "vae/ss_vae_M2.py --num-epochs=1 --enum-discrete=sequential",
     "vae/vae.py --num-epochs=1",
     "vae/vae_comparison.py --num-epochs=1",
-    pytest.param(
-        "cvae/main.py --num-quadrant-inputs=1 --num-epochs=1",
-        marks=pytest.mark.skip(reason="https://github.com/pyro-ppl/pyro/issues/3273"),
-    ),
+    "cvae/main.py --num-quadrant-inputs=1 --num-epochs=1",
     "contrib/funsor/hmm.py --num-steps=1 --truncate=10 --model=0 ",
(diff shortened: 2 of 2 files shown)


tests/test_examples.py
```diff
@@ -123,10 +123,6 @@
     "vae/ss_vae_M2.py --num-epochs=1 --enum-discrete=sequential",
     "vae/vae.py --num-epochs=1",
     "vae/vae_comparison.py --num-epochs=1",
-    pytest.param(
-        "cvae/main.py --num-quadrant-inputs=1 --num-epochs=1",
-        marks=pytest.mark.skip(reason="https://github.com/pyro-ppl/pyro/issues/3273"),
-    ),
```
fritzo, line 129: Could you replace these four lines with just the string
```
"cvae/main.py --num-quadrant-inputs=1 --num-epochs=1",
```
The pytest.mark.skip is just to skip the test, and we now want it to run.

examples/cvae/baseline.py
```diff
@@ -34,8 +34,8 @@ def __init__(self, masked_with=-1):
 
     def forward(self, input, target):
         target = target.view(input.shape)
-        loss = F.binary_cross_entropy(input, target, reduction="none")
-        loss[target == self.masked_with] = 0
+        # only calculate loss on target pixels (value = -1)
+        loss = F.binary_cross_entropy(input[target != self.masked_with], target[target != self.masked_with], reduction='none')
```
fritzo, line 38: Could you run `make lint` to fix the lint error? I think it's just a line too long.

fritzo (changes requested):
Thanks for the fix, just a couple of suggestions.

tests/test_examples.py
FDUguchunhui: Added the line for the test.

examples/cvae/baseline.py
FDUguchunhui: Had checked the format using `make lint` and fix the lint error.

fritzo (approved):

PR #3326, opened 2024-02-14
Add .github/workflows/publish.yml to publish to pypi.org

fritzo:
This aims to automate part of pyro's release process, namely uploading the pyro-ppl package to pypi.

I don't know how to test this script. I've copied it from [github docs](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#publishing-to-package-registries) (which seems to be newer than the [starter workflow](https://github.com/actions/starter-workflows/blob/main/ci/python-publish.yml); see [all versions here](https://github.com/pypa/gh-action-pypi-publish)).

I've also created an org-wide secret `PYPI_API_TOKEN`. I'm unsure whether there's anything else I need to do to make this secret available to the workflow.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,35 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+# GitHub recommends pinning actions to a commit SHA.
+# To get a newer version, you will need to update the SHA.
+# You can also reference a tag or branch, but the action may change without warning.
+
+name: Upload Python Package
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  deploy:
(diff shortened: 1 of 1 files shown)


eb8680:
> I don't know how to test this script

I'm not sure if it will run from a branch other than the designated main branch (`dev`), but I think you can merge this into `dev` and cut a test release that won't be installed by people who don't explicitly request it by doing the following (possibly redundant) steps:
1. Including a pre-release suffix in the `setup.py` semver and commit tag (e.g. `1.8.6-alpha0`)
2. Marking the GitHub test release as a pre-release and as not the latest release
3. Yanking or deleting the release from PyPI and GitHub after it is published

eb8680 (approved):
Should be correct as long as the PyPI API token is valid. Feel free to merge and cut a test release to exercise the workflow, unless there's a way to test it from this branch.

PR #3329, opened 2024-02-16
Fix action link for publishing to PyPI

fehiepsi:
I tested the workflow in https://github.com/pyro-ppl/numpyro/pull/1736 and found the link issue
https://github.com/pyro-ppl/numpyro/actions/runs/7935986924/job/21670294954


fritzo (approved):

fritzo:
Thanks @fehiepsi !

Issue #3332, opened 2024-02-26
VS Code mypy warnings due to Pyro's type hints


fritzo:
Parent issue #2550 

This issue is a running list of new mypy warnings when using Pyro>=1.9 which has a py.typed.
- [x] `poutine.trace(...).get_trace()` has no attribute "get_trace" #3334 
- [x] `dist.Categorical`: Module has no attribute "Categorical" #3333 
- [ ] `PyroParam` Missing positional arguments "constraint", "event_dim" in call to "PyroParam"  [call-arg]
- [ ]  `self.W_kg.T` "PyroParam" has no attribute "T"  [attr-defined]

## Environment
```
% python --version
Python 3.11.5

% pip freeze | grep mypy
mypy==1.8.0
mypy-extensions==1.0.0

% pip freeze | grep pyro
pyro-api==0.1.2
pyro-ppl==1.9.0
```

ordabayevy:
I get a lot of errors related to `PyroParam`.

PR #3333, opened 2024-02-26
Explicitly import distributions from torch

fritzo:
Addresses #3332 #2550 

This attempts to resolve mypy attribute undefined errors for `dist.Normal` and others. After these fixes, I am able to use VS Code to (1) hover over `dist.Normal` to read docs, and (2) detect type errors in e.g. `dist.Abnormal`.

1. Add `--config-settings editable_mode=strict` to `make install` which helps mypy
2. Explicitly add `py.typed` to `package_data` for setuptools<69
3. Explicitly add distributions to `__all__` in pyro.distributions.torch
4. Explicitly add distributions to `__all__` in pyro.distributions
5. Runs `make format` with the latest `black`

- [x] Question: will this break anyone's installs? E.g. if torch renames a distribution, we might get an import error in pyro.distributions.
  **Answer:** No, since we're using both `import *` and `try...except` around explicit imports.


ordabayevy (approved):
I like having explicit imports. If a distribution gets renamed or a new distribution gets added maybe we can try to mitigate it by putting `try/except` guards to work around import errors?

fritzo:
@ordabayevy Great, I've added `try..except` around the explicit imports, and also used the same implicit+explicit pattern in pyro.distributions.transforms and pyro.distributions.constraints.

ordabayevy (approved):
Looks great!

akotlar:
We just ran into this; gentle suggestion to cut a patch release with this commit.

PR #3337, opened 2024-03-09
Type annotate `pyro.nn.module`

ordabayevy:


pyro/nn/module.py
```diff
@@ -83,29 +108,37 @@ def forward(self):
...
+    ) -> "PyroParam":
         assert self.init_value is None
         return PyroParam(init_value, self.constraint, self.event_dim)
 
 
-PyroParam.__new__.__defaults__ = (None, constraints.real, None)
-
-
-class PyroSample(namedtuple("PyroSample", ("prior",))):
+@dataclass
```
fritzo, line 140: Can we use `@dataclass(slots=True, frozen=True)` here, to remain closer to namedtuple? Of course we'll then need to use `object._setattribute__(self, "foo", bar)` rather than `self.foo = bar` in `__post_init__`.

pyro/nn/module.py
```diff
@@ -136,24 +169,30 @@ def forward(self):
         object.
     """
 
-    def __init__(self, prior):
+    prior: Union[
+        "TorchDistributionMixin", Callable[["PyroModule"], "TorchDistributionMixin"]
+    ]
+
+    def __post_init__(self) -> None:
         super().__init__()
```
fritzo, line 177: I believe `super().__init__()` should be omitted because it should be called in the autogenerated `__init__` method.

pyro/nn/module.py
```diff
@@ -579,7 +635,11 @@ def __getattr__(self, name):
 
         return result
 
-    def __setattr__(self, name, value):
+    def __setattr__(
+        self,
+        name: str,
+        value: Union[torch.Tensor, torch.nn.Module, PyroParam, PyroSample],
```
fritzo, line 641: `value` should have type `Any`, since we fall back to `super.__setattr__(name, value)` at the last line of this method.

setup.cfg
```diff
@@ -63,10 +63,6 @@ warn_unused_ignores = True
 ignore_errors = True
 warn_unused_ignores = True
 
-[mypy-pyro.nn.*]
-ignore_errors = True
-warn_unused_ignores = True
```
fritzo, line 68: 🎉

fritzo (changes requested):
Looks great! This must have been a challenging puzzle; that `PyroModule` metaclass stuff is quite intricate.

Just a couple minor comments.

pyro/nn/module.py
ordabayevy: https://docs.python.org/3/library/dataclasses.html#post-init-processing says that `super().__init__()` should be called by the `__post_init__`. But since `PyroSample` isn't subclassing another class there is no need to call `super().__init__()`, right? So I removed it anyways.

pyro/nn/module.py
ordabayevy: Interesting. Is that why mypy thinks `torch.nn.Module`s attributes are either `Tensor` or `Module` because they annotated them as such 🤔 ? That always confused me.

pyro/nn/module.py
ordabayevy: `slots` seems to be new in Python 3.10

fritzo (approved):

PR #3339, opened 2024-03-11
Introducing PyroModuleList, because torch.nn.ModueList reinitializies modules when slice-indexing

### LINKED ISSUE #3341: Using PyroModule with torch.nn.ModueList fails for nested modules
Hi,

I already have made a [PR](https://github.com/pyro-ppl/pyro/pull/3339) but it seems like it just does not get any attention without an issue related to it. So here is the issue to it:

While using "nested" `PyroModule`'s: A `PyroModule` with a `PyroModule[torch.nn.ModuleList]` argument, containing another such module (see the test implementation in the PR for a more detailed example), I encountered some errors related to the existence of multiple sample sites with the same name.

I spend some time on tracing the issue and found that the `RuntimeError` was caused by an unlucky combination of `PyroModule` and `torch.nn.ModuleList`:
First: why using `torch.nn.ModuleList`? Well I wanted to model a `PyroModule` that owns a list of sub-pyromodules. My idea was to replace `linear = PyroModule[Linear](5, 2)` from the [modules example](https://pyro.ai/examples/modules.html) by `PyroModule[torch.nn.ModuleList](...)`.
That works fine if there is a
```python
class MyPyroModule(PyroModule):
    def __init__(self, some_module: PyroModule):
        self.my_submodule = PyroModule[torch.nn.ModuleList]([some_module for _ in range(3)])
```
but **can** fail if there is a nested structure, like
```python
my_nested_pyro_module = MyPyroModule(MyPyroModule(...))
```
The "can fail" could be resolved to the following different types if *accessing* the `self.my_submodule` argument:
+ using a index-positioned access, like `self.my_submodule[0](...)` worked fine, even for nested modules
+ using slice-indexing access, like `self.my_submodule[:-1](...)` only works if there is a single `MyPyroModule()` and not a `MyPyroModule(MyPyroModule(...))`

The cause is [this](https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/container.py#L293) line in the `torch.nn.ModuleList` class:
```python
@_copy_to_script_wrapper
    def __getitem__(self, idx: Union[int, slice]) -> Union[Module, 'ModuleList']:
        if isinstance(idx, slice):
            return self.__class__(list(self._modules.values())[idx])
        else:
            return self._modules[self._get_abs_string_index(idx)]
```
which calls `self.__class__`, which, means that for an object of type `PyroModule[torch.nn.ModuleList]`, it calls the `PyroModule.__init__` function, without the context of the parent module. This results in overwriting the names of the module's submodules and their `._pyro_name` attributes, and because of that, during sampling, sample sites may not be unique anymore.

I see two possible fixes for this:
1. I have introduced a `pyro.nn.PyroModuleList` class in this PR that inherits from `torch.nn.ModuleList` can overwrites the problematic `__getitem__` function
2. We may add some example/documentation on this issue, explicitly warning about using `PyroModule[torch.nn.ModuleList]` in combination with slice-indexing (feels a little unsafe to me)
3. Maybe some has another (potentially better) idea of how to deal with this?

I know that this is kind of a "special purpose" usage and may not affect basic `pyro` usage, but in particular the way the [modules example](https://docs.pyro.ai/en/stable/nn.html) motivates the usage of `PyroModule[torch.nn.Something]` could, imho, quickly lure other users into this (and it took me quite some time to find the root issue).

Please let me know what you think about this PR, and whether it needs updates or clarification.

Best regards
Martin

MartinBubel:
Fixes #3341

Hi,

While using "nested" `PyroModule`'s: A `PyroModule` with a `PyroModule[torch.nn.ModuleList]` argument, containing another such module (see the test implementation in the PR for a more detailed example), I encountered some errors related to the existence of multiple sample sites with the same name.

I spend some time on tracing the issue and found that the `RuntimeError` was caused by an unlucky combination of `PyroModule` and `torch.nn.ModuleList`:
First: why using `torch.nn.ModuleList`? Well I wanted to model a `PyroModule` that owns a list of sub-pyromodules. My idea was to replace `linear = PyroModule[Linear](5, 2)` from the [modules example](https://pyro.ai/examples/modules.html) by `PyroModule[torch.nn.ModuleList](...)`.
That works fine if there is a
```python
class MyPyroModule(PyroModule):
    def __init__(self, some_module: PyroModule):
        self.my_submodule = PyroModule[torch.nn.ModuleList]([some_module for _ in range(3)])
```
but **can** fail if there is a nested structure, like
```python
my_nested_pyro_module = MyPyroModule(MyPyroModule(...))
```
The "can fail" could be resolved to the following different types if *accessing* the `self.my_submodule` argument:
+ using a index-positioned access, like `self.my_submodule[0](...)` worked fine, even for nested modules
+ using slice-indexing access, like `self.my_submodule[:-1](...)` only works if there is a single `MyPyroModule()` and not a `MyPyroModule(MyPyroModule(...))`

The cause is [this](https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/container.py#L293) line in the `torch.nn.ModuleList` class:
```python
@_copy_to_script_wrapper
    def __getitem__(self, idx: Union[int, slice]) -> Union[Module, 'ModuleList']:
        if isinstance(idx, slice):
            return self.__class__(list(self._modules.values())[idx])
        else:
            return self._modules[self._get_abs_string_index(idx)]
```
which calls `self.__class__`, which, means that for an object of type `PyroModule[torch.nn.ModuleList]`, it calls the `PyroModule.__init__` function, without the context of the parent module. This results in overwriting the names of the module's submodules and their `._pyro_name` attributes, and because of that, during sampling, sample sites may not be unique anymore.

I see two possible fixes for this:
1. I have introduced a `pyro.nn.PyroModuleList` class in this PR that inherits from `torch.nn.ModuleList` can overwrites the problematic `__getitem__` function
2. We may add some example/documentation on this issue, explicitly warning about using `PyroModule[torch.nn.ModuleList]` in combination with slice-indexing (feels a little unsafe to me)
3. Maybe some has another (potentially better) idea of how to deal with this?

I know that this is kind of a "special purpose" usage and may not affect basic `pyro` usage, but in particular the way the [modules example](https://docs.pyro.ai/en/stable/nn.html) motivates the usage of `PyroModule[torch.nn.Something]` could, imho, quickly lure other users into this (and it took me quite some time to find the root issue).

Please let me know what you think about this PR, and whether it needs updates or clarification.

Best regards
Martin

### DIFF
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -14,9 +14,10 @@ the :class:`PyroSample` struct::
 """
 import functools
 import inspect
+from typing import Union
 import weakref
 from collections import OrderedDict, namedtuple
-
+from torch._jit_internal import _copy_to_script_wrapper
 import torch
 from torch.distributions import constraints, transform_to
 
@@ -826,3 +827,30 @@ class _FlatWeightsDescriptor:
 
 
 PyroModule[torch.nn.RNNBase]._flat_weights = _FlatWeightsDescriptor()
+
+
+# pyro module list
+# using pyro.nn.PyroModule[torch.nn.ModuleList] can cause issues when
+# slice-indexing nested PyroModuleLists, so we define a separate PyroModuleList
+# class that overwrites the __getitem__ method to return a torch.nn.ModuleList
+# to not use self.__class__ in __getitem__, as that would call the
+# PyroModule.__init__ without the parent module context, leading to a loss
+# of the parent module's _pyro_name, and eventually, errors during sampling
+# as parameter names may not be unique anymore
+# The scenario is rare but happend.
+# The fix could not be applied in torch directly, which is why we have to deal
+# with it here, see https://github.com/pytorch/pytorch/issues/121008
+class PyroModuleList(PyroModule, torch.nn.ModuleList):
+    def __init__(self, modules):
+        PyroModule.__init__(self)
+        torch.nn.ModuleList.__init__(self, modules)
+
+    @_copy_to_script_wrapper
+    def __getitem__(
+        self, idx: Union[int, slice]
+    ) -> Union[torch.nn.Module, "PyroModuleList"]:
+        if isinstance(idx, slice):
+            # return self.__class__(list(self._modules.values())[idx])
+            return torch.nn.ModuleList(list(self._modules.values())[idx])
+        else:
+            return self._modules[self._get_abs_string_index(idx)]
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -28,6 +28,7 @@ except ImportError:
     def _copy_to_script_wrapper(fn):
         return fn
 
+
 from collections import OrderedDict
 from dataclasses import dataclass
 from types import TracebackType
@@ -915,6 +916,7 @@ class _FlatWeightsDescriptor:
 
 
 PyroModule[torch.nn.RNNBase]._flat_weights = _FlatWeightsDescriptor()  # type: ignore[attr-defined]
+
 
 # pyro module list
 # using pyro.nn.PyroModule[torch.nn.ModuleList] can cause issues when
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -14,10 +14,10 @@ the :class:`PyroSample` struct::
 """
 import functools
 import inspect
-from typing import Iterable, Union
+import warnings
 import weakref
 from collections import OrderedDict, namedtuple
-import warnings
+from typing import Union
 
 try:
     from torch._jit_internal import _copy_to_script_wrapper
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -17,7 +17,20 @@ import inspect
 from typing import Union
 import weakref
 from collections import OrderedDict, namedtuple
-from torch._jit_internal import _copy_to_script_wrapper
+import warnings
+
+try:
+    from torch._jit_internal import _copy_to_script_wrapper
+except ImportError:
+    warnings.warn(
+        "Cannot find torch._jit_internal._copy_to_script_wrapper", ImportWarning
+    )
+
+    # Fall back to trivial decorator.
+    def _copy_to_script_wrapper(fn):
+        return fn
+
+
 import torch
 from torch.distributions import constraints, transform_to
 
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -867,3 +867,6 @@ class PyroModuleList(PyroModule, torch.nn.ModuleList):
             return torch.nn.ModuleList(list(self._modules.values())[idx])
         else:
             return self._modules[self._get_abs_string_index(idx)]
+
+
+_PyroModuleMeta._pyro_mixin_cache[torch.nn.ModuleList] = PyroModuleList
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -14,7 +14,7 @@ the :class:`PyroSample` struct::
 """
 import functools
 import inspect
-from typing import Union
+from typing import Iterable, Union
 import weakref
 from collections import OrderedDict, namedtuple
 import warnings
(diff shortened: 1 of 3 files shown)


pyro/nn/module.py
```diff
@@ -14,9 +14,10 @@
 """
 import functools
 import inspect
+from typing import Union
 import weakref
 from collections import OrderedDict, namedtuple
-
+from torch._jit_internal import _copy_to_script_wrapper
```
fritzo, line 20: I'm a little worried about depending on an internal function here. What do you think of limiting damage scope by guarding this import into a try-catch block:
```py
import warnings

try:
    from torch._jit_internal import _copy_to_script_wrapper
except ImportError:
    warnings.warn("Cannot find torch._jit_internal._copy_to_script_wrapper", ImportWarning)

    # Fall back to trivial decorator.
    def _copy_to_script_wrapper(fn):
        return fn
```

pyro/nn/module.py
```diff
@@ -826,3 +827,30 @@ def __set__(self, obj, value):
...
+# slice-indexing nested PyroModuleLists, so we define a separate PyroModuleList
+# class that overwrites the __getitem__ method to return a torch.nn.ModuleList
+# to not use self.__class__ in __getitem__, as that would call the
+# PyroModule.__init__ without the parent module context, leading to a loss
+# of the parent module's _pyro_name, and eventually, errors during sampling
+# as parameter names may not be unique anymore
+# The scenario is rare but happend.
+# The fix could not be applied in torch directly, which is why we have to deal
+# with it here, see https://github.com/pytorch/pytorch/issues/121008
+class PyroModuleList(PyroModule, torch.nn.ModuleList):
```
fritzo, line 843: This looks great. I think we can add a line to register your new module
```py
_PyroModuleMeta._pyro_mixin_cache[torch.nn.ModuleList] = PyroModuleList
```
That way users can simply type `PyroModule[torch.nn.ModuleList]`, and your fix will be applied to all existing uses of `PyroModule[torch.nn.ModuleList]` including when use in components of existing modules via `to_pyro_module_(...)`.

pyro/nn/module.py
```diff
@@ -826,3 +827,30 @@ def __set__(self, obj, value):
...
+    def __init__(self, modules):
+        PyroModule.__init__(self)
+        torch.nn.ModuleList.__init__(self, modules)
```
fritzo, line 846: Can we omit the `__init__` method by swapping inheritance order to be closer to `PyroModule[-]`'s order (defined [here](https://github.com/pyro-ppl/pyro/blob/490a7ff332f65700dd1b43802502f506b8390da3/pyro/nn/module.py#L251-L255))?
```diff
- class PyroModuleList(PyroModule, torch.nn.ModuleList):
+ class PyroModuleList(torch.nn.ModuleList, PyroModule):
```
I think that will be safer with `super.__init__()` handling of subclasses. This will also inherit `torch.nn.ModuleList`'s type hints and default args (`modules` defaults to `None`).

fritzo (changes requested):
Thanks for this subtle fix, thanks for writing thorough tests, and thanks for your patience with our review! I have just a few comments.

fritzo:
BTW lint errors can be detected with `make lint` and can often be fixed via `make format`.

pyro/nn/module.py
MartinBubel: that is good foresight, I added that in fae8ba6

pyro/nn/module.py
MartinBubel: I adressed that in fae8ba6.
I added the first recommendation at the end of the Module.py file, it that was the intention.
I also added the recommended assertion in the tests. The question is whether we keep the tests, which I used to demonstrate the behavior, or if we remove the part that served as a failing example.

pyro/nn/module.py
MartinBubel: Sound good, I applied it in 360fa10

pyro/nn/module.py
MartinBubel: that is good foresight, fixed it via 403f0d4

pyro/nn/module.py
MartinBubel: Updated it via 7516493.
I added the
```python
_PyroModuleMeta._pyro_mixin_cache[torch.nn.ModuleList] = PyroModuleList
```
at the end of the module.py file, if that was the intention.

I also added the assertion to the tests. However, I guess we can remove the "failing test" from the PR (as with the above change, it does not fail anymore). Or should we keep it as a demonstration of usage?

pyro/nn/module.py
MartinBubel: Applied in 5ffd8f7.

MartinBubel:
Linting is applied in b23177f.

Thanks @fritzo for the helpful review! I hope I included your suggestions as intended. If not, please let me know.

fritzo:
LGTM. I think you just need to `make format`

MartinBubel:
Seems like something got lost in the linting commit, should all be good now.

fritzo (approved):

PR #3340, opened 2024-03-14
Add function for calculating quantiles of weighed samples.

BenZickel:
# The Problem

When calculating distribution quantiles based on samples sampled from a guide it is often desirable to correct for the non-uniform gap between the model log-probability and the guide log-probability (this gap is already computed by the `pyro.infer.importance.Importance` module).

Currently there is only support for the below:
- Calculating quantiles from equally weighed samples by using `pyro.ops.stats.quantile` which interpolates quantiles over the sorted samples.
- Sampling with non-uniform weights using `pyro.distributions.empirical.Empirical` which does not support interpolation.

# The Proposed Solution

Add a function `pyro.ops.stats.weighed_quantile` which supports calculating interpolated quantiles with non-uniform weights. The implementation is similar to `pyro.ops.stats.quantile` but with different weights generated for each index of the input, according to the indices that sort that specific index.

### DIFF
--- a/pyro/ops/stats.py
+++ b/pyro/ops/stats.py
@@ -270,16 +270,15 @@ def weighed_quantile(input, probs, log_weights, dim=0):
     :param torch.Tensor log_weights: sample weights tensor.
     :param int dim: dimension to take quantiles from ``input``.
     :returns torch.Tensor: quantiles of ``input`` at ``probs``.
-    
+
     Example:
     >>> from pyro.ops.stats import weighed_quantile
-    >>> from torch import Tensor
-    >>> input = Tensor([[10, 50, 40], [20, 30, 0]])
-    >>> probs = Tensor([0.2, 0.8])
-    >>> log_weights = Tensor([0.4, 0.5, 0.1]).log()
-    >>> weighed_quantile(input, probs, log_weights, -1)
-    tensor([[40.4000, 47.6000],
-            [ 9.0000, 26.4000]])
+    >>> import torch
+    >>> input = torch.Tensor([[10, 50, 40], [20, 30, 0]])
+    >>> probs = torch.Tensor([0.2, 0.8])
+    >>> log_weights = torch.Tensor([0.4, 0.5, 0.1]).log()
+    >>> result = weighed_quantile(input, probs, log_weights, -1)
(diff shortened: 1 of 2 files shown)


BenZickel:
@fehiepsi, I've fixed the testing errors and now `make format` and `make test` pass successfully.
Sorry for not getting it right earlier.

pyro/ops/stats.py
```diff
@@ -261,6 +261,61 @@ def quantile(input, probs, dim=0):
     return quantiles if probs.shape != torch.Size([]) else quantiles.squeeze(dim)
 
 
+def weighed_quantile(input, probs, log_weights, dim=0):
```
fritzo, line 264: Would you mind adding type hints for new code?
```py
from typing import List

def weighted_quantile(
    input: torch.Tensor,
    probs: List[float],
    log_weights: torch.Tensor,
    dim: int = 0,
) -> torch.Tensor:
    ...
```

fritzo (changes requested):
Hi @BenZickel looks great! I wish we had designed Pyro for more native support of weighted sets of samples, but this is a step in the right direction.

Could you add a simple unit test of this function, similar to [test_quantiles()](https://github.com/pyro-ppl/pyro/blob/490a7ff332f65700dd1b43802502f506b8390da3/tests/ops/test_stats.py#L47-L62) in tests/ops/test_stats.py?

pyro/ops/stats.py
BenZickel: Thank you for the review @fritzo! I've added testing and type hints.

fritzo (changes requested):
Looks great, just one little comment on `Tuple[float]` vs `Tuple[float, ...]`

pyro/ops/stats.py
```diff
@@ -261,6 +262,66 @@ def quantile(input, probs, dim=0):
     return quantiles if probs.shape != torch.Size([]) else quantiles.squeeze(dim)
 
 
+def weighed_quantile(
+    input: torch.Tensor,
+    probs: Union[List[float], Tuple[float], torch.Tensor],
```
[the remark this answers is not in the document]
BenZickel: Thanks for catching this. I'm relatively new to type hints in Python. I've fixed to `Tuple[float, ...]`.

fritzo (approved):

PR #3345, opened 2024-03-22
Introducing pyro.infer.predictive.WeighedPredictive which reports weights along with predicted samples

BenZickel:
# The Problem

When sampling from the posterior predictive distribution we are often using a guide as an approximation for the posterior. As mentioned in #3340 it is often desirable to correct for the non-uniform per sample gap between the model log-probability and the guide log-probability. This gap is essentially the weight that should be assigned to each sample.

The current implementation of `pyro.infer.predictive.Predictive` does not support calculation of these weights.

# The Proposed Solution

Add `pyro.infer.predictive.WeighedPredictive` which supports calculation of per sample weights.

The implementation relies on three objects:
- Model which samples from priors and observations (same as in instantiation of `pyro.infer.predictive.Predictive`).
- Guide which approximates the posterior given observations (same as in instantiation of `pyro.infer.predictive.Predictive`).
- Model with observations constrained to be the actual observations. This model was used in creating the guide and is provided to instances of `pyro.infer.predictive.WeighedPredictive` when called as the keyword argument `model_guide` (as in the model that was used when creating the guide).

The `model_guide` is what enables calculation of the weights. If not provided we use the model provided at instantiation of `pyro.infer.predictive.WeighedPredictive` as the `model_guide` (in this case the model provided at instantiation is usually already with observations constrained to be the actual observations).

# Design Considerations

- Maintain backwards compatibility of `pyro.infer.predictive.Predictive`.
- Reuse as much as possible from `pyro.infer.predictive.Predictive` when implementing `pyro.infer.predictive.WeighedPredictive`.

### DIFF
--- a/pyro/infer/predictive.py
+++ b/pyro/infer/predictive.py
@@ -9,6 +9,7 @@ import torch
 
 import pyro
 import pyro.poutine as poutine
+from pyro.infer.util import plate_log_prob_sum
 from pyro.poutine.trace_struct import Trace
 from pyro.poutine.util import prune_subsample_sites
 
@@ -64,6 +65,9 @@ def _predictive_sequential(
     )
 
 
+_predictive_vectorize_plate_name = "_num_predictive_samples"
+
+
 def _predictive(
     model,
     posterior_samples,
@@ -72,11 +76,12 @@ def _predictive(
     parallel=False,
     model_args=(),
     model_kwargs={},
+    mask=True
 ):
-    model = torch.no_grad()(poutine.mask(model, mask=False))
+    model = torch.no_grad()(poutine.mask(model, mask=False) if mask else model)
     max_plate_nesting = _guess_max_plate_nesting(model, model_args, model_kwargs)
     vectorize = pyro.plate(
-        "_num_predictive_samples", num_samples, dim=-max_plate_nesting - 1
+        _predictive_vectorize_plate_name, num_samples, dim=-max_plate_nesting - 1
     )
     model_trace = prune_subsample_sites(
         poutine.trace(model).get_trace(*model_args, **model_kwargs)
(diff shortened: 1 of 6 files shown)


BenZickel:
Thank you for your [comments](https://github.com/pyro-ppl/pyro/pull/3345#issuecomment-2018416581) @fritzo. See below my feedback:

1. I've added more mathematical details to the docstring.
2. Yes, `WeighedPredictive` returns the exact same samples returned by `Predictive`, namely `p(x|z)q(z)`, but accompanies each sample with its weight `p(Xobs,z)/q(z)` where `p(Xobs,z)=p(Xobs|z)p(z)`.
3. I agree that we need to have some way to convert weighed samples to unweighed samples, but I believe this should be added in another pull request as there are several considerations related to multi-dimensional event samples and interpolation methods ([SMCFilter](https://github.com/pyro-ppl/pyro/blob/81def9c535113d2affd3d4a111a96d5a89fec52f/pyro/infer/smcfilter.py#L167) does not do interpolation). For now, we can calculate quantiles of weighed samples using `weighed_quantile` introduced in #3340.
4. I've created as much shared machinery between `WeighedPredictive` and [pyro.infer.Importance](https://docs.pyro.ai/en/dev/inference_algos.html#pyro.infer.importance.Importance) as I could in [026cae2](https://github.com/pyro-ppl/pyro/pull/3345/commits/026cae27510a0be08bf01d3e416e380a2bddbf3e). They do indeed share concepts and code.
5. As I see it [pyro.infer.ReweightedWakeSleep](https://docs.pyro.ai/en/dev/inference_algos.html#module-pyro.infer.rws) is a strategy to create a guide from your model and observations. As the created guide is usually not perfectly proportional to the model you would want to use `WeighedPredictive` in order to obtain the true distribution quantiles.

pyro/infer/predictive.py
```diff
@@ -31,53 +34,58 @@ def _guess_max_plate_nesting(model, args, kwargs):
     return max_plate_nesting
 
 
+class _predictiveResults(NamedTuple):
```
fritzo, line 37: Thanks for the cleanup! This will also help us with #2550

fritzo (approved):
The code looks great. Thanks for adding tests!

PR #3349, opened 2024-03-27
GCC-8 CI error test

ordabayevy:


ordabayevy:
Error message from CI:

```
/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/torch/include/c10/util/C++17.h:16:2: error: #error "You're trying to build PyTorch with a too old version of GCC. We need GCC 9 or later."
```

fritzo:
I guess we'll need to bump gcc version [here](https://github.com/pyro-ppl/pyro/blob/dev/.github/workflows/ci.yml#L10-L11)?

ordabayevy:
Do we need to upgrade g++ as well @fritzo ?

PR #3351, opened 2024-03-28
Add effective sample size analytics to WeighedPredictive results

BenZickel:
1. Allows calculation of the effective sample size of weighed sample results (returned by ``pyro.infer.predictive.WeighedPredictive``) by calling the ``.get_ESS`` method.
2. Code is shared with ``pyro.infer.importance.Importance``.

### DIFF
--- a/pyro/infer/importance.py
+++ b/pyro/infer/importance.py
@@ -15,7 +15,59 @@ from .enum import get_importance_trace
 from .util import plate_log_prob_sum
 
 
-class Importance(TracePosterior):
+class WeightAnalytics:
+    def get_log_normalizer(self):
+        """
+        Estimator of the normalizing constant of the target distribution.
+        (mean of the unnormalized weights)
+        """
+        # ensure list is not empty
+        if len(self.log_weights) > 0:
+            log_w = (
+                self.log_weights
+                if isinstance(self.log_weights, torch.Tensor)
+                else torch.tensor(self.log_weights)
+            )
+            log_num_samples = torch.log(torch.tensor(log_w.numel() * 1.0))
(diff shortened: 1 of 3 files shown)


pyro/infer/importance.py
```diff
@@ -15,7 +15,59 @@
 from .util import plate_log_prob_sum
 
 
-class Importance(TracePosterior):
+class WeightAnalytics:
```
fritzo, line 18: WDYT of making this a little clearer:
- Maybe renaming to `WeightAnalyticsMixin` or `LogWeightsMixin` or something?
- Adding a simple docstring like """Mixin class to compute analytics from a `.log_weights` attribute."""
- Adding just type declaration (but no initialization):
  ```py
  class WeightAnalytics:
      log_weights: Union[List[float], torch.Tensor]
      ...
  ```

fritzo (commented):
Looks great, just a couple nits.

fritzo:
And thanks for sharing code across modules, it helps us with maintenance!

pyro/infer/importance.py
BenZickel: Thank you for your comments @fritzo! I completely agree with them and have integrated them.

pyro/infer/predictive.py
```diff
@@ -317,16 +318,20 @@ def get_vectorized_trace(self, *args, **kwargs):
 
 
 class WeighedPredictiveResults(NamedTuple):
```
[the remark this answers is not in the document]
BenZickel: I didn't know ``dataclass`` but I agree it is a better alternative than ``namedtuple`` in this case, as the return values are not a natural tuple.

fritzo (approved):
Thanks for addressing my nits!

PR #3352, opened 2024-04-05
Resampler for weighed samples

BenZickel:
# Background

This pull request introduces a resampler for weighed samples that creates equally weighed samples from the distribution specified by the generator of the weighed samples (see item 3 [here](https://github.com/pyro-ppl/pyro/pull/3345#issuecomment-2018416581)). 

# Implementation

- Based on the Metropolis-Hastings algorithm.
- The ``guide`` is used as a proposal distribution which is independent of the current sample.
- Sample are generated and returned via the [WeighedPredictiveResults](https://github.com/pyro-ppl/pyro/blob/58080f81b662bd9575cdf4b466ab3d87236c95df/pyro/infer/predictive.py#L323) class.

# Notes

- The implementation guarantees that:
    - Samples are independent.
    - Samples will not be repeated (up to the probability that the random number generators generate identical samples) unlike sampling with replacement from a fixed set of weighed samples.
- In case the ``guide`` perfectly tracks the ``model`` this sampler will do nothing as the Metropolis-Hastings algorithm will have an acceptance probability of one for new samples.
- This resampler can be an efficient and fast way to correct "relatively small" errors in the ``guide`` that would otherwise take many optimization iterations to converge or not converge at all (this is demonstrated in the tests of this module).
- Calculating weighed samples quantiles by using resampling will usually have less variance then direct calculation using weighed quantiles (see [notes in the documentation](https://benzickel-pyro.readthedocs.io/en/add_mh_resampler/inference_algos.html#mhsampler-behavior)).


fritzo:
Hi @BenZickel, thanks for your patience. I'm not sure I understand how `MHResampler` is intended to be used. Do you intend this new resampler to be used as a component in a larger training algorithm, such that the whole algorithm will be akin to Reweighted Wake Sleep? Or would the resampler be mainly used for prediction? Gosh in either case it might help to have example usage in the docstring.

Also, do you understand the relationship betwen your `MHResampler` and the `importance_resample` utility [we](https://github.com/pyro-ppl/pyro/pull/3345#issuecomment-2018416581) [discussed](https://github.com/pyro-ppl/pyro/pull/3345#issuecomment-2019012614) in your [PR](https://github.com/pyro-ppl/pyro/pull/3345) from a couple weeks ago?

> @fritzo: ...it might be nice to add some sort of utility importance_resample: WeightedSamples -> Samples to convert from the weighted representation back to an unweighted representation...
> @BenZickel: I agree that we need to have some way to convert weighed samples to unweighed samples, but I believe this should be added in another pull request as there are several considerations related to multi-dimensional event samples and interpolation methods...

My thinking then was that it would be nice to bridge the two worlds (weighted vs uniform samples). I figured a simple way convert weighted -> unweighted would be to add a method `.resample()` method that just called [_systematic_resample](https://github.com/pyro-ppl/pyro/blob/725b135d68e2514299f8c6735f2c940bee134fc5/pyro/infer/smcfilter.py#L167) under the hood, and converted types. I'm not sure that has anything to do with your current PR other than the word 'resample' 😄

BenZickel:
Thx @fritzo for the review! The intended use of `MHResampler` is mainly for prediction and I've [added an example](https://benzickel-pyro.readthedocs.io/en/add_mh_resampler/inference_algos.html#pyro.infer.predictive.MHResampler) that reflects that (the example is basically copied from the combined test for `MHResampler` and `WeighedPredictive`). Although `MHResampler` is mainly used for prediction, it actually creates posterior predictive samples that are independent of the guide, and therefore could produce accurate results with fewer `SVI` iterations and reduced overall running time (this is actually tested in [this](https://github.com/pyro-ppl/pyro/pull/3352/files#diff-0d5a410678b9897d3ccb449a9346619ceba72a8a8f4aa7e20e9b23cbf9b1125eR50) test configuration where `SVI` iteration count is reduced from 5000 to 1000).

Regarding your second point, `MHResampler` is a way to do `importance_resample`, but not as a method `.resample` of `WeightedPredictiveResult`. The reason is that resampling from a fixed set of samples (by a method `.resample` for example) creates correlation and high variance of computed quantities, whereas `MHResampler` continuously creates new samples and selects which current samples will be replaced by the new samples. Due to this `MHResampler` needs access to the callable that creates new `WeightedPredictiveResult` when called (usually an instance of `WeighedPredictive` but other callables will work as well).

Lastly, you mentioned [_systematic_resample](https://github.com/pyro-ppl/pyro/blob/725b135d68e2514299f8c6735f2c940bee134fc5/pyro/infer/smcfilter.py#L167) which does resampling from a fixed set of samples (which is not what we want as explained in the previous paragraph). Resampling from a fixed set of samples is a necessity in [Sequential Monte Carlo](https://pyro.ai/examples/smcfilter.html) methods as the samples are time (sequence) dependent and therefore new samples for the current time cannot be generated without starting from time zero.

pyro/infer/predictive.py
```diff
@@ -450,3 +450,181 @@ def forward(self, *args, **kwargs):
...
+
+        # Fit guide
+        elbo = Trace_ELBO(num_particles=100, vectorize_particles=True)
+        svi = SVI(conditioned_model, guide, optim.Adam(dict(lr=3.0)), elbo)
+        for i in range(num_svi_steps):
+            svi.step()
+
+        # Create callable that returns weighed samples
+        posterior_predictive = WeighedPredictive(model,
+                                                 guide=beta_guide,
```
fritzo, line 500: nit: `beta_guide` -> `guide`?

pyro/infer/predictive.py
```diff
@@ -450,3 +450,181 @@ def forward(self, *args, **kwargs):
...
+
+        weighed_samples = posterior_predictive(model_guide=conditioned_model)
+        # Calculate quantile directly from weighed samples
+        weighed_samples_quantile = weighed_quantile(weighed_samples.samples['_RETURN'],
+                                                    [prob],
+                                                    weighed_samples.log_weights)[0]
+
+        resampler = MHResampler(posterior_predictive)
+        for resampling_count in range(10):
+            resampled_weighed_samples = resampler(model_guide=conditioned_model)
```
fritzo, line 515: Do I understand correctly that this discards the first 9 return values and keeps only the 10th? If so, consider adding a comment explaining what the number 10 is, or just naming it `mh_steps = 10`.

pyro/infer/predictive.py
```diff
@@ -450,3 +450,181 @@ def forward(self, *args, **kwargs):
             guide_log_prob=guide_log_prob,
             model_log_prob=model_log_prob,
         )
+
+
+class MHResampler(torch.nn.Module):
+    r"""
+    Resampler for weighed samples that generates equally weighed samples from the distribution
```
fritzo, line 457: Gosh I guess I missed this in the previous review, but shouldn't we use the standard term `weighted` rather than `weighed`? I interpret `weighted` as "individually annotated by weights", whereas `weighed` sounds like the state of a slab of fish just before the fishmonger stamps on its price tag.

pyro/infer/predictive.py
```diff
@@ -450,3 +450,181 @@ def forward(self, *args, **kwargs):
...
+    The above is the Metropolis-Hastings algorithm with the new sample candidate
+    proposal distribution being equal to the ``guide`` and independent of the
+    current sample such that :math:`g(x')=g(x' \mid x)`.
+
+    :param callable sampler: When called returns :class:`WeighedPredictiveResults`.
+    :param slice source_samples_slice: Select source samples for storage (default is `slice(0)`, i.e. none).
+    :param slice stored_samples_slice: Select output samples for storage (default is `slice(0)`, i.e. none).
+
+    The typical use case of :class:`MHResampler` would be to convert weighed samples
+    generated by :class:`WeighedPredictive` into equally weighed samples from the target distribution.
```
fritzo, line 479: Could you add a phrase or sentence regarding the intended iterative use of this function? Is it intended to be called repeatedly? What happens if you call it only once? (see my question below)

fritzo (changes requested):
Thanks for adding the example! I have just a couple more clarifying questions, whose answers I think could make `MHResampler` easier for users to understand.

pyro/infer/predictive.py
BenZickel: I think `weighed` is good if you look at it as derived from the verb to `weigh`, which is to ascertain the weight or importance of something,, so you have `weighed` samples, i.e. samples for which their weights have been ascertained.

pyro/infer/predictive.py
BenZickel: Yes, you understood correctly, the first 9 sets of samples are discarded in order to allow the resampling process to converge to the true posterior predictive distribution.

pyro/infer/predictive.py
BenZickel: I've added the below to the docs:

Each time an instance of `MHResampler` is called it returns a new set of samples, with the samples generated by the first call being distributed according to the ``guide``, and with each subsequent call the distribution of the samples becomes closer to that of the posterior predictive disdtribution. It might take some experimentation in order to find out in each case how many times one would need to call an instance of `MHResampler` in order to be close enough to the posterior predictive distribution.

pyro/infer/predictive.py
BenZickel: I've replaced 10 with `num_mh_steps`.

pyro/infer/predictive.py
BenZickel: You're right! I missed that one.

fritzo (approved):

PR #3354, opened 2024-04-05
Fix ruff version to 0.3.4 in order to overcome issue with version 0.3.5.

BenZickel:
- Linting with ruff version 0.3.5 fails as can be seen [here](https://github.com/pyro-ppl/pyro/actions/runs/8572360312/job/23494720392?pr=3352#step:5:16).
- The failure is due to ruff version 0.3.5 not respecting [this](https://github.com/pyro-ppl/pyro/blob/58080f81b662bd9575cdf4b466ab3d87236c95df/pyro/distributions/torch.py#L350) directive.
- The problem does not occur with ruff version 0.3.4.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip wheel 'setuptools!=58.5.*,<60'
-        pip install ruff black mypy nbstripout nbformat
+        pip install ruff==0.3.4 black mypy nbstripout nbformat
     - name: Lint
       run: |
         make lint


fritzo (approved):
Thanks @BenZickel! Do you have any idea how to get ruff>=0.3.5 to respect `__all__`?

BenZickel:
Hi @fritzo! I think the issue is that ruff 0.3.5 is not respecting the `# noqa: F822` directive for all lines of a multiline expression, while ruff 0.3.4 does. I will open an issue in the ruff repo and based on the feedback I get we can decide how to fix this.
This pull request is just a hotfix in the meantime.

Issue #3358, opened 2024-04-19
[FR] Predictive with deterministic site in the guide


OlaRonning:
Hi,

I'm working on a project where we would like to access the output of an NN in the guide when using `Predictive`. We've implemented it using a deterministic site in the guide. The program boils down to the following.

```
import pyro
from pyro.infer import Predictive
from pyro.distributions import Normal
import torch

def model():
    pyro.deterministic('m_deter', torch.tensor(1.))
    pyro.sample('x', Normal(torch.zeros(()), torch.ones(())))

def guide():
    pyro.deterministic('g_deter', torch.tensor(1.))
    pyro.sample('x', Normal(torch.zeros(()), torch.ones(())))


Predictive(
  model=model, 
  guide=guide, 
  return_sites=('model_site', 'guide_site', 'x'), 
  num_samples=1)() # Includes m_deter but not g_deter
```
We would like for both `m_deter` and `g_deter` to be included. It looks like Predictive currently only considers [model sites](https://github.com/pyro-ppl/pyro/blob/dev/pyro/infer/predictive.py#L108) for return sites. Would it be possible to expand it so we can include deterministic sites from the guide?

fritzo:
This looks reasonable to me. Since deterministic guide sites are usually ignored (e.g. in AutoGuides) I think we may want to gate this new behavior by an arg like `return_deterministic_guide_sites: bool` or something.

OlaRonning:
A guard makes sense. I'll give an implementation a shot.

OlaRonning:
@SarthakNikhal absolutely. Feel free to look at my WIP PR. I wrote the unittest relatively fast; you can probably develop a more suitable one.

OlaRonning:
I would make the test cover four cases:
 1. return_deterministic is true and no return_sites. returned samples should include all deterministic sites in the guide.
 2. return_deterministic is true and return_sites includes one of two deterministic sites in the guide. returned samples should only include the deterministic guide site in return_sites.
 3. return_determininistic is true and there are no deterministic sites in the guide. returned samples should be the same as when return_deterministic is false.
 4. return_deterministic is false and there is a deterministic site in the guide. the returned samples should not include the deterministic site from the guide.

You'd probably want to check both that sites are included in the returned samples and that their values are as expected. I believe you can work directly on the aleatory_science branch.

OlaRonning:
Hi @ Andrei-Treil,
[...]

PR #3361, opened 2024-04-22
Allow returning deterministic sites from guide in Predictive

OlaRonning:
Related to  #3358.

### DIFF
--- a/pyro/util.py
+++ b/pyro/util.py
@@ -266,6 +266,12 @@ def check_model_guide_match(model_trace, guide_trace, max_plate_nesting=math.inf
         if site["type"] == "sample"
         if site["infer"].get("is_auxiliary")
     )
+    det_vars = set(
+        name
+        for name, site in guide_trace.nodes.items()
+        if site["type"] == "sample"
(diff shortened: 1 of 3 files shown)


pyro/util.py
```diff
@@ -284,7 +290,7 @@ def check_model_guide_match(model_trace, guide_trace, max_plate_nesting=math.inf
         warnings.warn(
             "Found auxiliary vars in the model: {}".format(aux_vars & model_vars)
         )
-    if not (guide_vars <= model_vars | aux_vars):
+    if not (guide_vars <= model_vars | aux_vars | det_vars):
```
OlaRonning, line 293: @fritzo, it looks like deterministic sites are currently treated as (auxiliary) sample sites in the guide when comparing the model to the guide. Do you know if this is the intended behavior? I assumed it wasn't and sketched the following solution.

fritzo: Hi @OlaRonning, I'm unsure as I seldom use `predictive`, preferring custom predictive utilities.

@fehiepsi do you have an opinion?  It would be nice to keep similar behavior in pyro and numpyro.

PR #3362, opened 2024-04-22
Add the energy score scoring rule for comparing multivariate distributions

BenZickel:
# Background

In order to compare a predicted univariate distribution to a true univariate distribution Pyro already has ``pyro.ops.stats.crps_empirical`` which implements the Continuous Ranked Probability Score (CRPS) scoring rule for univariate distributions.

For multivariate distributions we can only use the CRPS to compare marginals but not the full distribution, and therefore two predicted distributions with the same marginals but different full distribution will have the same per marginal CRPS score relative to the true distribution.

# Proposal

The energy score scoring rule implemented in ``pyro.ops.stats.energy_score_empirical`` can be used to compare a predicted multivariate distribution to a true multivariate distribution, such that the score is minimized only if the full predicted multivariate distribution is equal to the full true multivariate distribution, i.e. it is not sufficient for the marginals of the predicted and true distributions to be the same in order to minimize the score (see more details [here](https://benzickel-pyro.readthedocs.io/en/add_energy_score/ops.html#pyro.ops.stats.energy_score_empirical)).

### DIFF
--- a/pyro/ops/stats.py
+++ b/pyro/ops/stats.py
@@ -508,3 +508,54 @@ def crps_empirical(pred, truth):
     weight = weight.reshape(weight.shape + (1,) * (diff.dim() - 1))
 
     return (pred - truth).abs().mean(0) - (diff * weight).sum(0) / num_samples**2
+
+
+def energy_score_empirical(pred, truth):
+    """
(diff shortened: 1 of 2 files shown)


fritzo (approved):
Looks great!

PR #3364, opened 2024-05-03
Replace torch._utils._accumulate with numpy.cumsum.

BenZickel:
This fixes failures in testing of `examples/contrib/mue/FactorMuE.py` and `examples/contrib/mue/ProfileHMM.py` as `torch._utils._accumulate` was removed in PyTorch 2.3.

### DIFF
--- a/examples/contrib/mue/ProfileHMM.py
+++ b/examples/contrib/mue/ProfileHMM.py
@@ -92,7 +92,7 @@ def main(args):
         indices = torch.randperm(sum(data_lengths), device=device).tolist()
         dataset_train, dataset_test = [
             torch.utils.data.Subset(dataset, indices[(offset - length) : offset])
-            for offset, length in zip(pyro.util._accumulate(data_lengths), data_lengths)
+            for offset, length in zip(np.cumsum(data_lengths), data_lengths)
         ]
     else:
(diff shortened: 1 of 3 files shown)


examples/contrib/mue/ProfileHMM.py
```diff
@@ -92,9 +92,7 @@ def main(args):
         indices = torch.randperm(sum(data_lengths), device=device).tolist()
         dataset_train, dataset_test = [
             torch.utils.data.Subset(dataset, indices[(offset - length) : offset])
-            for offset, length in zip(
-                torch._utils._accumulate(data_lengths), data_lengths
-            )
+            for offset, length in zip(pyro.util._accumulate(data_lengths), data_lengths)
```
fritzo, line 95: Why not just `np.cumsum(data_lengths)`?

BenZickel: Yes, that's a much better solution. I've updated the PR accordingly. Thx!

BenZickel:
@martinjankowiak, I am not aware of any other tests/code that is broken with PyTorch 2.3. The tests mentioned in this PR are the only ones that failed with PyTorch 2.3.

PR #3369, opened 2024-05-20
Add Stable distribution with numerically integrated log-probability calculation (StableWithLogProb).

### LINKED ISSUE #3280: Trouble with basic parameter estimation with the Stable distribution
[Moved the discussion in [this forum thread](https://forum.pyro.ai/t/trouble-with-basic-parameter-estimation-with-the-stable-distribution/5567) to here.]

Doing MLE on the skewness parameter of Stable distribution does not recover the original parameter. The skewness tends to stay around 0 regardless initial values.

```python
# adapted from https://github.com/fritzo/notebooks/blob/master/stable_mle.ipynb
import math
import torch
import pyro
from pyro.distributions import constraints
from pyro.infer.reparam import MinimalReparam
from pyro.infer.autoguide import AutoNormal, AutoGaussian

torch.set_default_dtype(torch.float64)
pyro.set_rng_seed(20230928)

# Define true parameters and number of datapoints
alpha = 1.5
beta = 0.8
c = 1.0
mu = 0.0
n = 10000

# sample data
data = dist.Stable(alpha, beta, c, mu).sample((n,))

@MinimalReparam()
def model(data):    
    alpha = 1.5  # pyro.param("alpha", torch.tensor(1.99), constraint=constraints.interval(0, 2))
    beta = pyro.param("beta", torch.tensor(0.5), constraint=constraints.interval(-1, 1))
    c = 1.0  # pyro.param("c", torch.tensor(1.0), constraint=constraints.positive)
    mu = 0.0  # pyro.param("mu", torch.tensor(0.0), constraint=constraints.real)
    with pyro.plate("data", data.shape[0]):
        pyro.sample("obs", dist.Stable(alpha, beta, c, mu), obs=data)

def train(model, guide, num_steps=1001, lr=0.1):
    pyro.clear_param_store()
    pyro.set_rng_seed(20230928)

    # set up ELBO, and optimizer
    elbo = Trace_ELBO()
    elbo.loss(model, guide, data=data)
    optim = pyro.optim.Adam({"lr": lr})
    svi = SVI(model, guide, optim, loss=elbo)

    # optimize
    losses = []
    for i in range(num_steps):
        loss = svi.step(data) / data.numel()
        losses.append(loss)
        if i % 100 == 0:
            print(f"step {i} loss = {loss:0.6g}")

    print(f"Parameter estimates (n = {n}):")
    print(f"beta: Estimate = {pyro.param('beta')}, true = {beta}")
    return losses

guide = AutoNormal(model)
train(model, guide);
```
gives us something like
```
step 0 loss = 57.484
step 100 loss = 2.82283
step 200 loss = 2.64291
step 300 loss = 2.57734
step 400 loss = 2.57825
step 500 loss = 2.55413
step 600 loss = 2.58021
step 700 loss = 2.55925
step 800 loss = 2.56627
step 900 loss = 2.55161
step 1000 loss = 2.54688
Parameter estimates (n = 10000):
beta: Estimate = -0.0007896144629401247, true = 0.8
```

BenZickel:
This fixes #3280 by adding `pyro.distributions.StableWithLogProb` which is based on `pyro.distributions.Stable` with an additional `log_prob` method (I opted for not modifying the `pyro.distributions.Stable` distribution at this stage).

Code is based on combining https://github.com/pyro-ppl/pyro/issues/3280#issuecomment-1758461760 by @mawright with the existing Stable distribution Pyro code base, with the following modifications:
- Make the calculation stable at and near an `alpha` value of one, and values at and near zero.
- Eliminate dependency on the `torchquad` package.
- Cache integration points (not sure if `torchquad` does this but overall speed is 25% faster than the reference implementation based on `torchquad`).

Per iteration duration is about 5 times slower than with reparameterization but overall convergence is much faster, and includes cases which do not converge with reparameterization (like skew `beta` estimation).

The log-probability calculation is based on integration over a uniformly distributed random variable $u$ such that $P(x) = \int du P(x|u) P(u)$. The integral can be converted to a reparameterization where we first sample $u$ with probability density $P(u)$ or $g(u)$ when approximating the posterior distribution by a guide, and secondly sampling or observing $x$ with the distribution $P(x|u)$. Initial tests indicate this reparameterization works but is still slower than estimating the log-probability by integration.

A [usage example](https://github.com/BenZickel/pyro/blob/stable_with_log_prob/tutorial/source/stable.ipynb) with real life data has been added to the last section of the Stable distribution tutorial.


fritzo (commented):
It's great to see this implemented!

I don't trust my review of the math, but I would trust some sort of density test. One option is to use [goftests.density_goodness_of_fit](https://github.com/posterior/goftests/blob/5fa60aed79c471f8c9b4c8dec93c81fd9ee2154f/goftests/__init__.py#L170), something like:
```py
@pytest.mark.parametrize(...)
def test_density(stability, skew, loc, scale):
    d = StableWithLogProb(stability, skew, loc, scale)
    samples = d.sample(1000)
    probs = d.log_prob(samples).exp()
    gof = goftests.density_goodness_of_fit(samples, probs)
    assert gof > 1e-2
```
Another option is to check against a reference implementation, say something in scipy. WDYT?

(btw thanks for your patience!)

fritzo (commented):
Looks great!

I think it would actually be cleaner to implement `Stable.log_prob()` rather than a separate `StableWithLogProb` class (but thank you for drafting a non-invasive solution!). Do you see any blockers to simply merging Stable <-> StableWithLogProb in this PR? I think the only change will be the need to update your tutorial's summary:
```diff
 ## Summary
-- [Stable.log_prob()](http://docs.pyro.ai/en/stable/distributions.html#stable) is undefined.
+- [Stable.log_prob()](http://docs.pyro.ai/en/stable/distributions.html#stable) is very expensive.
```
and simply the reparam stuff from your new section. The single `Stable` solution is nice in that users will at least be able to use default SVI and HMC, and the older StableReparam machinery can become an approximate cost-saving tool.

EDIT I guess we'd need to revise some `pytest.raises` checks in the tests, which might be easiest by adding an internal distribution `pyro.distributions.testing.fakes.StableWithoutLogProb`.

BTW I am a big fan of the Levy Stable distribution, and am delighted to see Pyro improving its support for heavy-tailed inference.

fritzo (approved):
Happy to merge as-is, but I think this `.log_prob()` deserves to be in the main `Stable` distribution. LMK if you want me to merge now, or if you want to combine `Stable` <-> `StableWithLogProbs`.

BenZickel:
Thx @fritzo for the review!

I also think heavy-tailed inference is much needed and I really appreciate all the work done on this so far.

It might be better to combine `StableWithLogProb` and `Stable` but I'd do it in a separate pull request (if at all). The advantage of keeping them separate is that users will be made explicitly aware of both the high cost of the log-probability calculation and the possibility of reducing that cost at the expense of accuracy by reparameterization. If we do combine the two we also need to figure out if the behavior of [MinimalReparam](https://github.com/BenZickel/pyro/blob/ca36025a3502c0160395b53145d2e95b56eaf15f/pyro/infer/reparam/strategies.py#L83) needs to be modified when handling the `Stable` distribution.

BenZickel:
One more option that comes to mind is to keep both `Stable` and `StableWithLogProb` and add the `.log_prob` method to `Stable`.  This way a user can enforce no reparameterization by using `StableWithLogProb` instead of `Stable`.

PR #3370, opened 2024-05-29
Add log_prob method to Stable (same one that already exists in StableWithLogProb)

BenZickel:
- Implements https://github.com/pyro-ppl/pyro/pull/3369#issuecomment-2134406814
- Added broadcasting and validation checks.


fritzo (approved):
Looks great! LMK if you'd like to make any other changes before our next release.

PR #3376, opened 2024-06-23
Support for deterministic dependent samples in PyroSample [enhancement]

BenZickel:
Make `pyro.nn.PyroSample(prior)` statements sample deterministic values when the `prior` is a function that returns a `torch.Tensor` value, as described in the below example:

```
from pyro.nn import PyroSample, PyroModule
from pyro import distributions as dist

class Location(PyroModule):
    def __init__(self):
        super().__init__()
        # Independent priors
        self.radius = PyroSample(dist.LogNormal(0,1))
        self.theta = PyroSample(dist.Normal(0.5, 0.1))
        # Dependent deterministic
        self.true_x = PyroSample(lambda self: self.radius * self.theta.cos())
        self.true_y = PyroSample(lambda self: self.radius * self.theta.sin())
        # Dependent samples
        self.observed_x = PyroSample(lambda self: dist.Normal(self.true_x, 0.05))
        self.observed_y = PyroSample(lambda self: dist.Normal(self.true_y, 0.05))
        
    def forward(self):
        return self.true_x, self.true_y, self.observed_x, self.observed_y
```

In the above code the dependent deterministic samples can be converted to dependent samples by assigning them `pyro.nn.PyroSample(prior)` statements with the `prior` being a function returning a proper distribution, allowing simple exploration of various model complexity levels for the same problem.

The updated docs can be reviewed [here](https://benzickel-pyro.readthedocs.io/en/pyro_module_deterministic_dependent_sample/nn.html?highlight=pyrosample#pyro.nn.module.PyroSample).

### DIFF
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -163,6 +163,7 @@ class PyroSample:
         assert isinstance(my_module, PyroModule)
         my_module.x = PyroSample(Normal(0, 1))                    # independent
         my_module.y = PyroSample(lambda self: Normal(self.x, 1))  # dependent
+        my_module.z = PyroSample(lambda self: self.y ** 2)        # deterministic dependent
 
     or EXPERIMENTALLY as a decorator on lazy initialization methods::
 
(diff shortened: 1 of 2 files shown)


fehiepsi:
Thanks, @BenZickel! The feature looks great.

fehiepsi (approved):

PR #3378, opened 2024-06-27
Fix pyro parameter nodes not showing up in graph of pyro deterministic nodes [bugfix]

### LINKED ISSUE #3371: [discussion] Parameters not rendering when feeding into `pyro.deterministic`
### Issue Description
Parts of my programs are deterministic, and these parts still have latent parameters I would like to make inferences of. But when I try to render my program, my registered parameters are not showing up. It looks like this only happens when these parameters are feeding strictly deterministic variables.

I suspect they're not rendering because parameters that feed only into deterministic "variables" will not actually be part of optimization. Is this true?

### Environment
 - OS:  macOS Sonoma 14.5
 - Python version:  3.11.4
 - PyTorch version:  2.0.0
 - Pyro version: 1.9.0+f02dfb9

### Code Snippet
```
import pyro
import torch


data = torch.tensor([1.0, 2.0, 3.0])

def deterministic_model(data):
    value = pyro.param("param", torch.tensor(0.))
    with pyro.plate("plate", len(data)):
        pyro.deterministic("deterministic", data + value)

pyro.render_model(deterministic_model, model_args=(data,), render_params=True)
```

```
def probabilistic_model(data):
    value = pyro.param("param", torch.tensor(0.))
    with pyro.plate("plate", len(data)):
        pyro.sample("probabilistic", dist.Normal(value, 1), obs=data)

pyro.render_model(probabilistic_model, model_args=(data,), render_params=True)
```

BenZickel:
- Fixes #3371.
- Also had to instruct ruff to ignore newly introduced object redefinition checks.

### DIFF
--- a/pyro/distributions/hmm.py
+++ b/pyro/distributions/hmm.py
@@ -1091,7 +1091,7 @@ class LinearHMM(HiddenMarkovModel):
         self.transforms = transforms
 
     @constraints.dependent_property(event_dim=2)
-    def support(self):
+    def support(self):  # noqa: F811
         return constraints.independent(self.observation_dist.support, 1)
 
     def expand(self, batch_shape, _instance=None):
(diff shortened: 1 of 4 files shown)


fehiepsi (approved):
LGTM. Thanks for the fix, @BenZickel!

PR #3383, opened 2024-07-13
Update mappings to Sphnix documentation of other projects [bugfix]

BenZickel:
Fixes [this CI failure](https://github.com/pyro-ppl/pyro/actions/runs/9867350497/job/27247602355).


fritzo (approved):
Thanks, @BenZickel!

PR #3387, opened 2024-07-23
Fix models not rendering after application of the equalize effect handler

BenZickel:
# Problem

After using the `pyro.poutine.equalize` effect handler model rendering with `pyro.render_model` does not work due to no `base_dist` attribute in `Delta` sampled sites.

# Solution

Add masking so that the sampling function has the `base_dist` attribute as expected by `pyro.render_model` for `Delta` sampled sites.


BenZickel:
Also added a fix for this [lint failure](https://github.com/pyro-ppl/pyro/actions/runs/10053766446/job/27787126851?pr=3387) due to a `mypy` update.

fritzo (approved):

PR #3389, opened 2024-07-27
Support PyTorch 2.4 in tests

BenZickel:
Fixes CI failures:
- Due to `torch.load` which requires explicit specification of weights_only=False (see https://github.com/pyro-ppl/pyro/actions/runs/10118757885/job/27986180729).
- Altered warning messgae (see https://github.com/pyro-ppl/pyro/actions/runs/10121842111/job/27993274686).

### DIFF
--- a/tests/ops/einsum/test_adjoint.py
+++ b/tests/ops/einsum/test_adjoint.py
@@ -117,7 +117,7 @@ def test_marginal(equation):
         assert_equal(expected, actual)
 
 
-@pytest.mark.filterwarnings("ignore:.*reduce_op` is deprecated")
+@pytest.mark.filterwarnings("ignore:.*reduce_op`? is deprecated")
 def test_require_backward_memory_leak():
     tensors = [o for o in gc.get_objects() if torch.is_tensor(o)]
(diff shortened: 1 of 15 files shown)


tests/ops/einsum/test_adjoint.py
```diff
@@ -117,7 +117,7 @@ def test_marginal(equation):
         assert_equal(expected, actual)
 
 
-@pytest.mark.filterwarnings("ignore:.*reduce_op is deprecated")
+@pytest.mark.filterwarnings("ignore:.*reduce_op` is deprecated")
```
fritzo: Can we replace the "`" with a ".?" or something so as to cover both versions of the deprecation warning?

fritzo (commented):
Thanks so much for maintaining the tests. I have just one small comment, otherwise good to go!

tests/ops/einsum/test_adjoint.py
BenZickel: Yes, I've changed to "ignore:.*reduce_op`? is deprecated" which matches both versions.

fritzo (approved):

PR #3397, opened 2024-09-18
Use more numerically stable `log1p` in examples/scanvi/scanvi.py

kit1980:
Found using TorchFix https://github.com/pytorch-labs/torchfix/

PR #3398, opened 2024-09-19
Maintenance fixes

ordabayevy:
- Deprecated API usages found by TorchFix linter (https://github.com/pyro-ppl/pyro/pull/3397#issuecomment-2357456164)
- Remove tests for deprecated `torch.triangular_solve`
- Replace `deep_getattr` function with Python's `operator.attrgetter`

### DIFF
--- a/pyro/distributions/omt_mvn.py
+++ b/pyro/distributions/omt_mvn.py
@@ -68,7 +68,7 @@ class _OMTMVNSample(Function):
         diff_L_ab = 0.5 * sum_leftmost(g_ja * epsilon_jb + g_R_inv * z_ja, -2)
 
         Sigma_inv = torch.mm(R_inv, R_inv.t())
-        V, D, _ = torch.svd(Sigma_inv + jitter)
+        V, D, _ = torch.linalg.svd(Sigma_inv + jitter)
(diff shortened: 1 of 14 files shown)


martinjankowiak (approved):
lgtm thanks!

kit1980:
You can add TorchFix to the CI to prevent regressions and get future improvements.
It would be very easy if you run flake8, but I see you recently removed flake8 from the CI...

ordabayevy:
Yes, we moved from flake8 to ruff because it is much faster. Do you plan to support ruff in the future?

kit1980:
For ruff everything needs to be rewritten completely, in Rust. Maybe someone from ruff community will do that.
TorchFix already uses LibCST as the main library, which is written in Rust, so should be fast already.
Maybe you want to re-enable flake8 just for TorchFix?

Also there is a standalone `torchfix` script besides the flake8 interface, but currently it doesn't have support for running in CI.

martinjankowiak:
thanks @kit1980 !

we probably don't want to bring `flake8` back into our CI, but i've made an issue to track this in case future versions of `torchfix` can be brought into our CI more easily: #3399

PR #3401, opened 2024-09-23
Fix EqualizeMessenger type annotations

BenZickel:


pyro/poutine/handlers.py
```diff
@@ -306,14 +306,16 @@ def escape(  # type: ignore[empty-body]
 def equalize(
     sites: Union[str, List[str]],
     type: Optional[str],
-) -> ConditionMessenger: ...
+    keep_dist: Optional[bool],
```
ordabayevy, line 309: I think `keep_dist` should have just a `bool` type, right? It should have the same type as in the `EqualizeMessenger` class definition.

BenZickel: You are right they should have the same type, so I changed the type in the `EqualizeMessenger` class definition to `Optional[bool]`, as `keep_dist` is optional.

ordabayevy (approved):
LGTM!

PR #3402, opened 2024-09-24
Add batched calculation option to `energy_score_empirical` in order to reduce memory consumption

BenZickel:
# The Problem

When calculating the energy score of large samples with `pyro.ops.stats.energy_score_empirical` the memory consumption can be very high which is prohibitive for some applications.

# The Solution

Add an option to do the energy score calculation in batches by providing an additional optional parameter `pred_batch_size` which specifies the sample batch to use in the calculation. The calculation time is longer but the result stays the same up to numerical accuracy.

### DIFF
--- a/pyro/ops/stats.py
+++ b/pyro/ops/stats.py
@@ -570,13 +570,13 @@ def energy_score_empirical(
             pred = pred[..., pred_batch_size:, :]
         # Calculate predictions distance to truth
         retval = (
-            torch.cat(
+            torch.stack(
                 [
-                    torch.cdist(pred_batch, truth).sum(dim=-2, keepdim=True)
+                    torch.cdist(pred_batch, truth).sum(dim=-2)
                     for pred_batch in pred_batches
                 ],
-                dim=-2,
-            ).sum(dim=-2)
+                dim=0,
+            ).sum(dim=0)
             / pred_len
         )
         # Calculate predictions self distance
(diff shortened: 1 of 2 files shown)


pyro/ops/stats.py
```diff
@@ -552,10 +556,43 @@ def energy_score_empirical(pred: torch.Tensor, truth: torch.Tensor) -> torch.Ten
...
+            pred = pred[..., pred_batch_size:, :]
+        # Calculate predictions distance to truth
+        retval = (
+            torch.cat(
+                [
+                    torch.cdist(pred_batch, truth).sum(dim=-2, keepdim=True)
+                    for pred_batch in pred_batches
+                ],
+                dim=-2,
+            ).sum(dim=-2)
```
fritzo: I'm surprised by the `cat(..., dim=-2).sum(dim=-2)` rather than a `sum(....sum(dim=-2))` or `functools.reduce(operator.add, [....sum(dim=-2)])`. If the cat-then-sum is intentional, could you add a comment explaining why you're not reducing early?

fritzo (dismissed):
LGTM after explaining or replacing the cat-then-add

pyro/ops/stats.py
BenZickel: My initial thought was to use torch native operations, but I didn't stay consistent and used `sum(...)` later on in the function.

After further thought and research I think it would be best to use `stack(...,dim=0).sum(dim=0)` instead of `sum(..)` at both locations, as this allows this function to be JIT compiled, despite being slightly less memory efficient (could also be that `stack(...,dim=0).sum(dim=0)` is more memory efficient if you consider the torch computational tree and gradient calculations).

I have added a commit to that effect and also added a test for JIT compilation.

Nevertheless, I think both options can work so I'm happy to change this if you think otherwise.

fritzo (approved):

PR #3405, opened 2024-10-21
fix autodetermination of plate device when running in a device contexf

ilia-kats:
Apparently, the torch.Tensor() constructor ignores device contexts, e.g. running 

```
with torch.device("cuda"):
    print(torch.Tensor().device)
````
still returns `cpu`.


ilia-kats:
The CI is failing in files that this PR does not touch. Let me know if I should still fix it.

ordabayevy:
Thanks for the fix! Could you merge the dev branch that should fix the CI errors.

ordabayevy (approved):

PR #3415, opened 2025-01-21
Add a centered variance option to the ClippedAdam optimizer

BenZickel:
# Problem

When using the [ClippedAdam](https://docs.pyro.ai/en/dev/optimization.html#pyro.optim.clipped_adam.ClippedAdam) optimizer with highly imbalanced parameter gradients stability, the convergence rate of parameters with stable gradients is slower than what it could be.

# Solution

Add an option to use the centered variance in the denominator of the step size calculation. Parameters with stable gradients will have a lower centered variance, than the current uncentered variance, and therefore will have a larger step size and higher convergence rate.

# Testing

The improvement in convergence rate is shown below (taken from the test function run with plotting enabled):
- The first plot shows the number of iterations needed in order to reach convergence, where convergence is defined as the ultimate loss plus a small threshold.
- The second plot shows the convergence rate which is the mean per iteration improvement of the gap between the loss and the ultimate loss, which is roughly proportional to the inverse of the number of iterations needed in order to reach convergence. One can also notice that the convergence rate is less sensitive to changes in the learning rate when using the centered variance, compared to the uncentered variance case.
- The third plot shows the ultimate loss reached, which shows that for regular Adam with uncentered variance the best possible loss, which is zero in our case, is not attained for small learning rates, within the allotted number of iterations.
image

### DIFF
--- a/pyro/optim/clipped_adam.py
+++ b/pyro/optim/clipped_adam.py
@@ -23,17 +23,17 @@ class ClippedAdam(Optimizer):
 
     Small modification to the Adam algorithm implemented in torch.optim.Adam
     to include gradient clipping and learning rate decay and an option to use
-    the centered variance.
+    the centered variance (see equation 2 in [2]).
 
-    References
+    **References**
 
-    `A Method for Stochastic Optimization`, Diederik P. Kingma, Jimmy Ba
-    https://arxiv.org/abs/1412.6980
+    [1] `A Method for Stochastic Optimization`, Diederik P. Kingma, Jimmy Ba
+        https://arxiv.org/abs/1412.6980
 
-    `A Two-Step Machine Learning Method for Predicting the Formation Energy of Ternary Compounds`,
-    Varadarajan Rengaraj, Sebastian Jost, Franz Bethke, Christian Plessl,
-    Hossein Mirhosseini, Andrea Walther, Thomas D. Kühne
-    https://doi.org/10.3390/computation11050095
+    [2] `A Two-Step Machine Learning Method for Predicting the Formation Energy of Ternary Compounds`,
+        Varadarajan Rengaraj, Sebastian Jost, Franz Bethke, Christian Plessl,
+        Hossein Mirhosseini, Andrea Walther, Thomas D. Kühne
+        https://doi.org/10.3390/computation11050095
     """
 
     def __init__(
--- a/pyro/optim/clipped_adam.py
+++ b/pyro/optim/clipped_adam.py
@@ -22,12 +22,18 @@ class ClippedAdam(Optimizer):
     :param centered_variance: use centered variance (default: False)
 
     Small modification to the Adam algorithm implemented in torch.optim.Adam
-    to include gradient clipping and learning rate decay.
+    to include gradient clipping and learning rate decay and an option to use
+    the centered variance.
 
-    Reference
+    References
 
     `A Method for Stochastic Optimization`, Diederik P. Kingma, Jimmy Ba
     https://arxiv.org/abs/1412.6980
+
+    `A Two-Step Machine Learning Method for Predicting the Formation Energy of Ternary Compounds`,
+    Varadarajan Rengaraj, Sebastian Jost, Franz Bethke, Christian Plessl,
+    Hossein Mirhosseini, Andrea Walther, Thomas D. Kühne
+    https://doi.org/10.3390/computation11050095
     """
 
     def __init__(
--- a/pyro/optim/clipped_adam.py
+++ b/pyro/optim/clipped_adam.py
@@ -19,6 +19,7 @@ class ClippedAdam(Optimizer):
     :param weight_decay: weight decay (L2 penalty) (default: 0)
     :param clip_norm: magnitude of norm to which gradients are clipped (default: 10.0)
     :param lrd: rate at which learning rate decays (default: 1.0)
+    :param centered_variance: use centered variance (default: False)
 
     Small modification to the Adam algorithm implemented in torch.optim.Adam
     to include gradient clipping and learning rate decay.
@@ -38,6 +39,7 @@ class ClippedAdam(Optimizer):
         weight_decay=0,
         clip_norm: float = 10.0,
         lrd: float = 1.0,
+        centered_variance: bool = False,
     ):
         defaults = dict(
             lr=lr,
@@ -46,6 +48,7 @@ class ClippedAdam(Optimizer):
             weight_decay=weight_decay,
             clip_norm=clip_norm,
             lrd=lrd,
+            centered_variance=centered_variance,
         )
         super().__init__(params, defaults)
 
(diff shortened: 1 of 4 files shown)


pyro/optim/clipped_adam.py
```diff
@@ -87,7 +90,8 @@ def step(self, closure: Optional[Callable] = None) -> Optional[Any]:
 
                 # Decay the first and second moment running average coefficient
                 exp_avg.mul_(beta1).add_(grad, alpha=1 - beta1)
-                exp_avg_sq.mul_(beta2).addcmul_(grad, grad, value=1 - beta2)
+                grad_var = (grad - exp_avg) if group["centered_variance"] else grad
```
martinjankowiak, line 99: i haven't seen this kind of algorithm in the literature but presumably you want bias correction?

BenZickel: The Adam optimizer uses the sqrt of the mean squared gradient in the denominator of the step size, which is the same as the sqrt of the variance of the gradient if the mean gradient is zero (this is the case most of the times). However, I've seen many optimizations in which for some parameter elements the mean gradient is not zero which significantly slows down optimization. The idea is that if the gradient is relatively stable, i.e. with small centered variance, you can take bigger steps in its direction. I use the mean gradient estimator in the centered variance calculation.

martinjankowiak:
@BenZickel can you please explain your figure? i don't know how a convergence rate is computed, and i can't tell if the differences in the second plot are significant given the scale

pyro/optim/clipped_adam.py
martinjankowiak: still not sure why you're not applying bias correction? `exp_avg` is biased down early in the algo

martinjankowiak:
a bit of googling led me here. the same algo in essence? a 2 second scan suggests they do bias correction

https://edoc.hu-berlin.de/server/api/core/bitstreams/14960a8d-4c35-4d08-86d7-1e130ecd42c8/content

BenZickel:
Thanks for the review @martinjankowiak.
- I've added an additional figure and explanations about the plots. I think the first plot is the most important, as it shows the number of iterations needed in order to reach convergence.
- Regarding the [reference](https://edoc.hu-berlin.de/server/api/core/bitstreams/14960a8d-4c35-4d08-86d7-1e130ecd42c8/content) you provided, the algorithm described is indeed the same so I've added it to the references section. They describe the algorithm as using the true variance instead of the second moment of the gradient ("One has to note that Adam does not actually calculate the variance but the second moment instead. If E[gt] = 0, both definitions are identical, and fewer operations are required to calculate the second moment."). As for bias correction, they reach the conclusion that it does not need to change ("Finally, from the bias-corrected expression for vt, we conclude that the bias correction for cAdam is the same as that of Adam.").

BenZickel:
I've added the option to use the centered variance option in the Latent Dirichlet Allocation (LDA) example. When running some tests I've noticed that the centered variance option improve both the convergence rate, and ultimate loss, for a wide range of learning rates. Additionally, the same phenomena, as seen in the above example, of reduced sensitivity of the convergence rate to changes in the learning rate, can be observed when using the centered variance option.

The centered variance option in the LDA example can be used by running
```
python pyro-ppl/examples/lda.py -cv True
```

pyro/optim/clipped_adam.py
```diff
@@ -19,14 +19,21 @@ class ClippedAdam(Optimizer):
     :param weight_decay: weight decay (L2 penalty) (default: 0)
     :param clip_norm: magnitude of norm to which gradients are clipped (default: 10.0)
     :param lrd: rate at which learning rate decays (default: 1.0)
+    :param centered_variance: use centered variance (default: False)
 
     Small modification to the Adam algorithm implemented in torch.optim.Adam
-    to include gradient clipping and learning rate decay.
+    to include gradient clipping and learning rate decay and an option to use
+    the centered variance.
```
martinjankowiak: can you point to the ref here?

tests/optim/test_optim.py
```diff
@@ -435,3 +435,105 @@ def step(svi, optimizer):
         actual.append(step(svi, optimizer))
 
     assert_equal(actual, expected)
+
+
+def test_centered_clipped_adam(plot_results=False):
```
martinjankowiak: how long does this test take?

tests/optim/test_optim.py
```diff
@@ -435,3 +435,105 @@ def step(svi, optimizer):
...
+        )
+        for count in range(num_iter):
+            optim.zero_grad()
+            loss = loss_fn(p)
+            loss.backward()
+            optim.step()
+            loss_vec.append(loss)
+        return torch.Tensor(loss_vec)
+
+    def calc_convergence(loss_vec, tail_len=100, threshold=0.01):
```
martinjankowiak, line 473: comment what is being computed?

tests/optim/test_optim.py
```diff
@@ -435,3 +435,105 @@ def step(svi, optimizer):
...
+        ultimate_loss = loss_vec[-tail_len:].mean()
+        convergence_iter = (loss_vec < (ultimate_loss + threshold)).nonzero().min()
+        convergence_vec = loss_vec[:convergence_iter] - ultimate_loss
+        convergence_rate = (convergence_vec[:-1] / convergence_vec[1:]).log().mean()
+        return ultimate_loss, convergence_rate, convergence_iter
+
+    def get_convergence_vec(lr_vec, centered_variance):
```
martinjankowiak, line 486: comment what is being computed?

martinjankowiak:
thanks @BenZickel ! the motivation makes sense, and i can imagine how this might help, though i'm perhaps somewhat surprised by the size of the effect, though i guess your `w` has quite a range in magnitude, perhaps a larger range than we might expect to see in most scenarios

martinjankowiak (dismissed):

tests/optim/test_optim.py
BenZickel: The test runs for about 20 seconds on my laptop.

pyro/optim/clipped_adam.py
BenZickel: Done

tests/optim/test_optim.py
BenZickel: I've added a comment.

tests/optim/test_optim.py
BenZickel: I've added a comment.

tests/optim/test_optim.py
martinjankowiak: cool i think it might be better to simplify the test to make it faster. you convinced me the algorithm can work well. we don't need to relitigate that every time CI runs. either: i) a basic smoke_test; or ii) just showing that the algorithm achieves an approximate minimum, should be sufficient?

BenZickel: I've reduced the number of learning rates tested so now the test runs in slightly less than 5 seconds. I've also added an option to run the full test with plots by running
```
pytest tests/optim/test_optim.py::test_centered_clipped_adam --plot True
```

martinjankowiak (approved):
thanks @BenZickel !

PR #3420, opened 2025-02-05
Update intro_long.ipynb

yannickeich:
Updated the name of the second argument in pyro.param from "init" to "init_tensor" in the long introduction


fritzo (approved):
Thanks for the fix!

PR #3422, opened 2025-02-28
don't log to the root logger

ilia-kats:
logging to the root logger creates a StreamHandler on the root logger, leading to duplicated log messages if there are any other loggers with handlers and propagate=True


ilia-kats:
The CI is failing due to HTTP errors unrelated to this PR

fritzo (approved):

PR #3425, opened 2025-03-11
fix datapath bart data

juanitorduz:
I was trying to simply run

```python
from pyro.contrib.examples.bart import load_bart_od

dataset = load_bart_od()
```

and I was getting a timeout error. I saw this in NumPyro and this was fixed in https://github.com/pyro-ppl/numpyro/pull/1996 by @fehiepsi . I tested the change locally (same strategy) and works as expected 🙏 .

### DIFF
--- a/pyro/contrib/examples/bart.py
+++ b/pyro/contrib/examples/bart.py
@@ -123,7 +123,9 @@ def load_bart_od():
     if os.path.exists(pkl_file):
         return torch.load(pkl_file, weights_only=False)
 
-    filenames = multiprocessing.Pool(len(SOURCE_FILES)).map(_load_hourly_od, SOURCE_FILES)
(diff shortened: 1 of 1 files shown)


fritzo:
Thanks for the fix! LGTM except you'll need to `make lint` or `black` or something. Looks like your editor accidentally reformatted some lines to exceed this codebase's max line length.

juanitorduz:
Fixed in https://github.com/pyro-ppl/pyro/pull/3425/commits/e3856029458ea60dd5646aff60a0db4a6809e254 (and https://github.com/pyro-ppl/pyro/pull/3425/commits/ba6b5b0fb8031fe9cfb8fe343f1a940b4c3502aa) :)

fritzo (approved):

PR #3427, opened 2025-04-05
Use datasets from GitHub instead of CloudFront.

BenZickel:
# Problem

The CloudFront storing most of the datasets used by examples is down and this causes many CI tests to fail.

# Solution

Switch to using datasets from GitHub.


fritzo (dismissed):
Thanks, @BenZickel!

fritzo (approved):

PR #3440, opened 2025-07-09
correct svi_part_ii docs

shiraamitchell:
I was talking with @martinjankowiak in this thread, where he said to correct the doc:

https://forum.pyro.ai/t/it-is-always-safe-to-assume-dependence-to-event-vs-plate/8858/6?u=shiraamitchell


martinjankowiak (approved):
thanks!

PR #3452, opened 2026-05-18
Stabilize Gaussian mixture backward passes

jburroni:
## Summary

This PR fixes numerical instabilities in the backward passes for Pyro's Gaussian mixture distributions:

- `MixtureOfDiagNormals`

- `GaussianScaleMixture`

- `MixtureOfDiagNormalsSharedCovariance`

The affected backward paths can materialize very small density/product terms in probability space. In float32, these terms can underflow to zero, which can then produce nonfinite gradients during backpropagation.

## Changes

- Adds regression tests that assert finite gradients in numerically sensitive high-dimensional / small-scale cases.

- Updates the affected backward computations to avoid unstable probability-space divisions/products where possible.

- Format code to pass lint.

## Tests

Added regression coverage for:

- small-scale `MixtureOfDiagNormals`

- small-scale `GaussianScaleMixture`

- high-dimensional `MixtureOfDiagNormalsSharedCovariance`

Each test verifies that samples, losses, and parameter gradients remain finite.

### DIFF
--- a/pyro/distributions/gaussian_scale_mixture.py
+++ b/pyro/distributions/gaussian_scale_mixture.py
@@ -153,14 +153,15 @@ class _GSMSample(Function):
         r_sqr_j = r_sqr / component_scale_sqr  # l j
         log_coord_scale_product = coord_scale.log().sum()
         log_component_scale_power = component_scale.log() * float(dim)
-        
+
         log_gaussian_normalizer = (
             0.5 * math.log(2.0 * math.pi) * float(dim)
             + log_coord_scale_product
             + log_component_scale_power
-        ).unsqueeze(0)  # 1 j
+        )
+        log_gaussian_normalizer = log_gaussian_normalizer.unsqueeze(0)  # 1 j
 
-        log_q_j = -0.5 * r_sqr_j - log_gaussian_normalizer # l j
+        log_q_j = -0.5 * r_sqr_j - log_gaussian_normalizer  # l j
         log_q_tot = torch.logsumexp(pis.log() + log_q_j, dim=-1, keepdim=True)  # l 1
         posterior_j = torch.exp(pis.log() + log_q_j - log_q_tot)  # l j
         Phi_j = torch.exp(-0.5 * r_sqr_j)  # l j
@@ -179,12 +180,13 @@ class _GSMSample(Function):
             extra_term = (
                 coeffs[-1]
                 * math.sqrt(0.5 * math.pi)
-                * torch.erfc(r_sqr_j.sqrt() / root_two))  # l j
+                * torch.erfc(r_sqr_j.sqrt() / root_two)
+            )  # l j
             Phi_j += extra_term * torch.pow(r_sqr_j, -0.5 * float(dim))
 
         logits_grad = (z.unsqueeze(-2) * Phi_j.unsqueeze(-1) * g).sum(-1)  # l j
-        
-        log_logits_scale = -log_gaussian_normalizer - log_q_tot # l j
+
+        log_logits_scale = -log_gaussian_normalizer - log_q_tot  # l j
         logits_grad = logits_grad * torch.exp(log_logits_scale)  # l j
         logits_grad = sum_leftmost(logits_grad, -1)
         logits_grad = pis * logits_grad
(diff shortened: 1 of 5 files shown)


martinjankowiak:
thanks @jburroni! your diff changed a bunch of other files so you'll need to fix that if you want this reviewed and merged

jburroni:
Hi @martinjankowiak!
The commits for this PR are ebd8d57 and 8384f6f.
The third one, which modified many files, was made by running `make lint`.

The GH actions seem to be broken, and I don't know how to proceed. What do you suggest?

jburroni:
Hi @martinjankowiak, I've removed everything that was not strictly related to the PR.

martinjankowiak:
@jburroni thanks! i think we'll want to wait for #3453 to merge, then rebase, and then we can review, and then we should be good to go

martinjankowiak:
@jburroni fyi #3453 merged

tests/distributions/test_gaussian_mixtures.py
```diff
@@ -211,6 +211,81 @@ def test_gsm_log_prob():
     )
 
 
+def test_mix_of_diag_normals_backward_small_scales_finite_grads():
+    torch.manual_seed(0)
```
martinjankowiak: please either remove seed or keep local to this test with fork_rng

tests/distributions/test_gaussian_mixtures.py
```diff
@@ -211,6 +211,81 @@ def test_gsm_log_prob():
...
+
+    loss.backward()
+
+    assert torch.isfinite(locs.grad).all()
+    assert torch.isfinite(coord_scale.grad).all()
+    assert torch.isfinite(component_logits.grad).all()
+
+
+def test_gsm_backward_small_coord_scale_finite_grads():
+    torch.manual_seed(0)
```
martinjankowiak: please either remove seed or keep local to this test with fork_rng

pyro/distributions/gaussian_scale_mixture.py
```diff
@@ -151,21 +151,34 @@ def backward(ctx, grad_output):
         epsilons_sqr = torch.pow(epsilons, 2.0)  # l i
         r_sqr = epsilons_sqr.sum(-1, keepdim=True)  # l
         r_sqr_j = r_sqr / component_scale_sqr  # l j
-        coord_scale_product = coord_scale.prod()
-        component_scale_power = torch.pow(component_scale, float(dim))
+        log_coord_scale_product = coord_scale.log().sum()
+        log_component_scale_power = component_scale.log() * float(dim)
+
+        log_q_j = (
```
martinjankowiak: this diff is complex. i didn't look carefully but claude is suspicious:
```
The logits_grad computation is suspicious. The old code:
pythonlogits_grad /= q_tot
logits_grad = sum_leftmost(...) * (2π)^(-dim/2)
logits_grad = pis * logits_grad / (component_scale_power * coord_scale_product)
The new code removes these normalization factors entirely and applies exp(log_normalizer) only inside the if dim % 2 == 1 branch (the extra_term). For even dim, log_normalizer is computed but never used — the normalization factors /q_tot, *(2π)^(-dim/2), /(component_scale_power * coord_scale_product) are simply absent from logits_grad.
The new centering term logits_grad = logits_grad - logits_grad.sum() * pis does implement the correct softmax Jacobian (d_loss/d_logits_k = pis_k*(d_loss/d_pis_k - Σ_j pis_j * d_loss/d_pis_j)), but only if the incoming logits_grad at that point is already d_loss/d_pis. Given the missing normalization for even dim, I'm not convinced it is.
```

martinjankowiak (commented):
thanks @jburroni !

jburroni:
Thank you for your feedback @martinjankowiak 

I made a mistake in my previous version. I'm sorry. I’ve now improved my changes to `gaussian_scale_mixture.py`.
In particular, I introduced a variable, `log_gaussian_normalizer`, made the code closer to the original, and used `torch.erfc` instead of `1-torch.erf`.

I also removed manual seeding from the new tests.

pyro/distributions/gaussian_scale_mixture.py
martinjankowiak, line 180: does `test_mean_gradient` pass with `D=3`? unclear to me if we're testing this branch properly. this would be an existing issue, but good to check?

martinjankowiak (approved):
looks good, thanks for doing this!

pyro/distributions/gaussian_scale_mixture.py
jburroni: I'm sorry for the delay.
I've included a comment before the if statement to clarify its relation to the equation in the paper. 
I've also added a test for the case D=3.

martinjankowiak:
@jburroni please pull in main after #3459 so lint works again

jburroni:
Done

martinjankowiak:
thanks @jburroni !!

PR #3469, opened 2026-08-19
Fix PyroModule registration of frozen parameters

### LINKED ISSUE #3438: Pyro sets all Parameters to requires_grad=True, when using PyroModule
### Issue Description
Hi all,
apparently the bug initially seen in #1778 reappears, when using a `PyroModule` instead and running a simple inference. 

Meaning that parameters which have `requires_grad=False` set will still be added to the `param_store` and additionally be set to `requires_grad=True`, making them learnable, even when they should be not.

### Environment
 - Debian, Python 3.11.6
 - PyTorch version: 2.2.1
 - Pyro version: 1.9.1

### Code Snippet

Here a quick example based on the already existing test for the previous issue:
```python
import torch
from torch.nn import Parameter

import pyro
from pyro.nn import PyroModule
from pyro.infer.autoguide import AutoDelta

# Create a generic PyroModule with some non-learnable parameters
class net_class(PyroModule):
    def __init__(self):
        super().__init__()
        self.x = Parameter(torch.zeros(1))
        self.y = Parameter(torch.zeros(1), requires_grad=False)
    
    def forward(self, s, y=None):
        s += self.x
        s += self.y
        
        return s

# Initialize model
net = net_class()

# Create guide
guide = AutoDelta(net)

# Create optimizer
optimizer = torch.optim.Adam
scheduler = pyro.optim.CosineAnnealingLR({'optimizer': optimizer, 'optim_args': {'lr': 1e-3}, 'T_max': 250})

# Create loss
loss_method = Trace_ELBO(
    num_particles=1,
)

# Create svi object
svi = SVI(net, guide, scheduler, loss=loss_method)

# Run an eval step to populate the param_store
net.eval()
loss = svi.evaluate_loss([1,2], [1,1])

# Test for parameters in param_store
print('Parameters in Pyro param_store:')
for key in pyro.get_param_store().keys():
    print(key, pyro.get_param_store()[key])

assert "x" in pyro.get_param_store().keys()
assert "y" not in pyro.get_param_store().keys()
```
This will print:
```
Parameters in Pyro param_store:
x Parameter containing:
tensor([0.], requires_grad=True)
y Parameter containing:
tensor([0.], requires_grad=True)
```

Which means that `y` is now learnable and part of the `param_store`, even though it should be neither.

Potentially a similar workaround as in #1779 could work here as well?

tandede:
## Proposed changes

`PyroModule` currently emits a `pyro.param` statement whenever a regular `torch.nn.Parameter` is read during a module call. `ParamStoreDict` makes registered tensors trainable, so this silently changes parameters created with `requires_grad=False` to `requires_grad=True` and adds them to the parameter store.

This change:

- skips parameter statements for frozen regular parameters, matching the registration contract of `pyro.module()`;
- applies the same rule when a regular parameter is assigned inside an active `PyroModule` context;
- preserves the existing global and module-local behavior for trainable parameters.

The regression test covers both parameter-store modes, parameters created during initialization, and parameters assigned dynamically during `forward()`.

Fixes #3438.

## Tests

- `pytest -q tests/nn/test_module.py` (69 passed, 1 xfailed)
- `pytest -q tests/params/test_module.py` (7 passed)
- `pytest -q tests/infer/test_autoguide.py -k 'test_exact and not AutoGaussianFunsor'` (29 passed)
- `pytest -q tests/distributions/test_spanning_tree.py` (66 passed, 60 skipped, 6 xfailed)
- `ruff check .`
- `black --check pyro/nn/module.py tests/nn/test_module.py`
- `python scripts/update_headers.py --check`
- `mypy --warn-unused-ignores pyro/nn/module.py tests/nn/test_module.py`

### DIFF
--- a/pyro/nn/module.py
+++ b/pyro/nn/module.py
@@ -629,12 +629,14 @@ class PyroModule(torch.nn.Module, metaclass=_PyroModuleMeta):
         result = super().__getattr__(name)
 
         # Regular nn.Parameters trigger pyro.param statements.
-        if isinstance(result, torch.nn.Parameter) and not name.endswith(
-            "_unconstrained"
+        if (
+            isinstance(result, torch.nn.Parameter)
+            and not name.endswith("_unconstrained")
+            and self._pyro_context.active
         ):
-            if self._pyro_context.active and not _is_module_local_param_enabled():
+            if result.requires_grad and not _is_module_local_param_enabled():
                 pyro.param(self._pyro_get_fullname(name), result)
-            elif self._pyro_context.active and _is_module_local_param_enabled():
+            elif result.requires_grad:
                 # fake param statement to ensure any handlers of pyro.param are applied,
                 # even though we don't use the contents of the local parameter store
                 fullname = self._pyro_get_fullname(name)
@@ -740,7 +742,11 @@ class PyroModule(torch.nn.Module, metaclass=_PyroModuleMeta):
                 delattr(self, name)
             except AttributeError:
                 pass
-            if self._pyro_context.active and not _is_module_local_param_enabled():
+            if (
+                self._pyro_context.active
+                and value.requires_grad
+                and not _is_module_local_param_enabled()
+            ):
                 fullname = self._pyro_get_fullname(name)
                 value = pyro.param(fullname, value)
                 if not isinstance(value, torch.nn.Parameter):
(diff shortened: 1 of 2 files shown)


tandede:
Hi maintainers, I wanted to follow up on this PR in case it was overlooked. It fixes registration of frozen parameters in PyroModule and is ready for review. Please let me know if any changes or additional validation would be useful. Thank you!

PR #3470, opened 2026-08-27
Migrate packaging metadata to pyproject.toml

ordabayevy:
## Summary
- migrate project metadata and dependencies from `setup.py`/`setup.cfg` to `pyproject.toml`
- use `pyro.__version__` as the canonical version through setuptools dynamic metadata
- update release documentation and version/header tooling for the new metadata layout
- remove the obsolete `setup.py`/`setup.cfg` packaging entry points

## Versioning note
This drops the legacy `1.2.3+sha` development-version scheme previously implemented in `setup.py`. Development builds now use the declared package version without appending the current commit SHA. Given Pyro’s slower development cadence, keeping development builds aligned with the declared release version is acceptable.

## Verification
- `ruff check .`
- Black formatting check
- setuptools metadata generation reports version `1.9.1`
- Python import and TOML parsing checks

The full test suite is left to CI.


pyro/__init__.py
```diff
@@ -27,14 +27,7 @@
 
 from . import settings
 
-# After changing this, run scripts/update_version.py
```
fritzo, line 30: Please restore this comment. See [update_version.py](https://github.com/pyro-ppl/pyro/blob/dev/scripts/update_version.py) for behavior. Try running it.

fritzo (changes requested):
Thanks for doing this!

Just one requested change, reverting a comment deletion that is necessary to update tutorial version pins, which saved us work in the past when users ran tutorials with the wrong pyro version.

Could you also add to the PR description that this drops our old 1.2.3+sha version scheme that was implemented in setup.py? I think this is probably fine as Pyro development has slowed.

pyro/__init__.py
ordabayevy: Done

fritzo (approved):
Thanks for cleaning this up!

PR #3472, opened 2026-08-31
Add InverseWishart distribution

### LINKED ISSUE #3448: [feature request] New Distribution: Inverse Wishart
Add this distribution, it sometime basic for econometric modelling.

webzuweb:
## Summary

Closes #3448.

Adds the inverse Wishart distribution, which is useful for econometric modelling (e.g., covariance priors).

## Implementation

`InverseWishart` is implemented as a `TransformedDistribution` wrapping `Wishart(df, precision_matrix=scale)` under a matrix-inverse transform. This relies on the identity: if `X ~ Wishart(df, Psi^{-1})` then `X^{-1} ~ InverseWishart(df, Psi)`.

The matrix inverse map on the space of `p x p` symmetric matrices has Jacobian determinant `|X|^{-(p+1)}`, which is applied via the (private) `_MatrixInverseTransform`.

## Tests

Added `tests/distributions/test_inverse_wishart.py` covering:
- `log_prob` checked against the closed-form density (via `torch.special.multigammaln`)
- analytic `mean` and `mode`
- sample shape, positive-definiteness and symmetry of samples
- the relationship `X^{-1} ~ Wishart` up to the Jacobian
- `expand`

All 13 tests pass locally.

### DIFF
--- a/pyro/distributions/__init__.py
+++ b/pyro/distributions/__init__.py
@@ -98,6 +98,7 @@ from pyro.distributions.improper_uniform import ImproperUniform
 
 if "InverseGamma" not in locals():  # Use PyTorch version if available.
     from pyro.distributions.inverse_gamma import InverseGamma
+from pyro.distributions.inverse_wishart import InverseWishart
 from pyro.distributions.lkj import LKJ, LKJCorrCholesky
 from pyro.distributions.log_normal_negative_binomial import LogNormalNegativeBinomial
 from pyro.distributions.logistic import Logistic, SkewLogistic
@@ -190,6 +191,7 @@ __all__ = [
     "Independent",
     "IndependentHMM",
     "InverseGamma",
+    "InverseWishart",
     "Kumaraswamy",
     "LKJ",
     "LKJCholesky",
--- a/pyro/distributions/inverse_wishart.py
+++ b/pyro/distributions/inverse_wishart.py
@@ -1,0 +1,86 @@
+# Copyright Contributors to the Pyro project.
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+from torch.distributions import constraints
+from torch.distributions.transforms import Transform
+
+from pyro.distributions.torch import TransformedDistribution, Wishart
+
+
+class _MatrixInverseTransform(Transform):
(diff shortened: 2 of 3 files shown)


fritzo (commented):
Hi @webzuweb thanks for suggesting this. Could we discuss design before I thoroughly review?

I believe one of the reasons delaying this implementation was numerical stability, and that many Bayesian practitioners seem to prefer the LKJ distribution over inverse Wishart. Regarding numerical stability, do you think it would be more stable to parametrize by a scale_tril matrix / lower Cholesky factor as in [MultivariateNormal](https://docs.pyro.ai/en/stable/distributions.html#pyro.distributions.MultivariateNormal)?

webzuweb:
Thanks for the thoughtful review @fritzo!

On LKJ vs InverseWishart — fully agree LKJ is more flexible and increasingly the default for covariance priors. Still, InverseWishart is the textbook conjugate prior for the covariance of a multivariate normal and shows up in a lot of Bayesian modeling code, so I think a minimal, correct implementation is worth having even if the docs keep recommending LKJ as the more flexible option.

On stability — good point, and you're right that the weak spot is the matrix inverse. The current transform computes `torch.linalg.inv(x)` (dense inverse, O(p^3), can be ill-conditioned). A more stable construction computes the inverse via Cholesky: `L = cholesky(x)`, then `X^-1 = cholesky_inverse(L)` (or triangular solves), with the log-abs-det Jacobian from `2 * sum(log(diag(L)))` instead of `slogdet`. That avoids dense inversion entirely.

Re: parametrizing by `scale_tril` — that would be more stable for *constructing* the base Wishart, but `torch.distributions.Wishart` only exposes `covariance_matrix`/`precision_matrix` (no `scale_tril`), so I'd still have to rebuild `Psi = L @ L.T` and pass it through, which doesn't preserve the triangular structure at the transform level. Two options I see:

1. keep `scale_matrix` as the public param (parity with scipy/torch) and switch the transform to a Cholesky-based inverse for stability, or
2. add an internal Cholesky path in the transform.

Which fits pyro's conventions better? Happy to adjust either way.

PR #3474, opened 2026-09-04
Switch formatting from Black to Ruff

ordabayevy:
## Summary

- Replace Black formatting commands and dependency with Ruff formatter.
- Configure Ruff formatting to 88 characters.
- Reformat the affected Python files and update contributor documentation.
- Preserve Black-compatible long comment/string behavior by ignoring E501.

## Validation

- `make lint`
- `pytest -q tests/test_settings.py`
- `git diff --check`

### DIFF
--- a/examples/air/air.py
+++ b/examples/air/air.py
@@ -209,12 +209,12 @@ class AIR(nn.Module):
             )
 
     def guide(self, data, batch_size, **kwargs):
-        pyro.module("rnn", self.rnn),
-        pyro.module("predict", self.predict),
-        pyro.module("encode", self.encode),
-        pyro.module("embed", self.embed),
-        pyro.module("bl_rnn", self.bl_rnn),
-        pyro.module("bl_predict", self.bl_predict),
+        (pyro.module("rnn", self.rnn),)
+        (pyro.module("predict", self.predict),)
+        (pyro.module("encode", self.encode),)
+        (pyro.module("embed", self.embed),)
+        (pyro.module("bl_rnn", self.bl_rnn),)
+        (pyro.module("bl_predict", self.bl_predict),)
         pyro.module("bl_embed", self.bl_embed)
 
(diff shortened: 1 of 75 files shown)


examples/air/air.py
```diff
@@ -209,12 +209,12 @@ def model(self, data, batch_size, **kwargs):
...
-        pyro.module("encode", self.encode),
-        pyro.module("embed", self.embed),
-        pyro.module("bl_rnn", self.bl_rnn),
-        pyro.module("bl_predict", self.bl_predict),
+        (pyro.module("rnn", self.rnn),)
+        (pyro.module("predict", self.predict),)
+        (pyro.module("encode", self.encode),)
+        (pyro.module("embed", self.embed),)
+        (pyro.module("bl_rnn", self.bl_rnn),)
+        (pyro.module("bl_predict", self.bl_predict),)
```
fritzo: nit: This seems weird, I think those commas should not be there. They are no-ops creating an immediately discarded tuple. Either is correct, but this PR amplifies the old noise

fritzo (approved):
Thanks for cleaning up!

examples/air/air.py
ordabayevy: Nice catch!

fritzo (approved):

ordabayevy:
I see this warning for `docs` job:

```
[docs (3.12)]
(https://github.com/pyro-ppl/pyro/actions/runs/33728918193/job/101139783933#step:11:2)
Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/checkout@v2, actions/setup-python@v2. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
```

I added a commit to update github actions version.

fritzo (approved):

